PR #889, opened 2023-08-13
Add more bounding box utilities

sethrj:
This changes the bounding box so that it's in a null state when any lower coordinate is higher than the upper coordinate. This change simplifies all the logic with unions and intersections: the intersection of a default bbox and another bbox is the default null bbox; the union of a null bbox and a valid bbox is the valid bbox. A bounding box being "null" is different than being "empty": empty can mean the lower and upper extents match to make a plane (finite or infinite); or an empty bbox could be a single point. Null means no point exists in the bbox.

### DIFF
--- a/src/orange/BoundingBox.hh
+++ b/src/orange/BoundingBox.hh
@@ -22,8 +22,9 @@ namespace celeritas
 /*!
  * Axis-aligned bounding box.
  *
- * The bounding box can be constructed in an unassigned state, in which lower
- * and upper cannot be called.
+ * The default bounding box is "degenerate", which encloses no space and
+ * evaluates to "false". A degenerate bounding box still has the ability to be
+ * unioned and intersected with other bounding boxes with the expected effect.
  */
 template<class T>
 class BoundingBox
@@ -39,8 +40,8 @@ class BoundingBox
     static inline CELER_FUNCTION BoundingBox from_infinite();
 
     // Construct from unchecked lower/upper bounds
-    static inline CELER_FUNCTION BoundingBox from_unchecked(Real3 const& lower,
-                                                            Real3 const& upper);
+    static CELER_CONSTEXPR_FUNCTION BoundingBox
+    from_unchecked(Real3 const& lower, Real3 const& upper);
 
     // Construct in unassigned state
     CELER_CONSTEXPR_FUNCTION BoundingBox();
@@ -96,7 +97,7 @@ CELER_FUNCTION BoundingBox<T> BoundingBox<T>::from_infinite()
  * "degenerate" implementation of the bounding box.
  */
 template<class T>
-CELER_FUNCTION BoundingBox<T>
+CELER_CONSTEXPR_FUNCTION BoundingBox<T>
 BoundingBox<T>::from_unchecked(Real3 const& lo, Real3 const& hi)
 {
     return BoundingBox<T>{std::true_type{}, lo, hi};
@@ -125,7 +126,7 @@ CELER_CONSTEXPR_FUNCTION BoundingBox<T>::BoundingBox()
 
 //---------------------------------------------------------------------------//
 /*!
- * Create a valid bounding box from two points.
+ * Create a nondegenerate bounding box from two points.
  *
  * The lower and upper points are allowed to be equal (an empty bounding box
  * at a single point) but upper must not be less than lower.
--- a/src/orange/BoundingBox.hh
+++ b/src/orange/BoundingBox.hh
@@ -36,14 +36,6 @@ class BoundingBox
     using Real3 = Array<real_type, 3>;
     //!@}
 
-    // Closed, axis-aligned halfspace.
-    struct Halfspace
-    {
-        Sense sense;
-        Axis axis;
-        T position;
-    };
-
   public:
     // Construct from infinite extents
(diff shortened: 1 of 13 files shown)


src/orange/BoundingBox.hh
```diff
@@ -7,65 +7,81 @@
...
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
  * Axis-aligned bounding box.
  *
- * The bounding box can be constructed in an unassigned state, in which lower
- * and upper cannot be called.
+ * The default bounding box is "degenerate", which encloses no space and
```
elliottbiondo: I don't think "degenerate" is the correct term here. Planes, lines, and points are degenerate cuboids. I would stick to using "null" for describing ((inf, inf, inf), (-inf, -inf, -inf)) as you do in the text of this PR.

src/orange/BoundingBox.hh
```diff
@@ -7,65 +7,81 @@
...
-    CELER_FORCEINLINE_FUNCTION explicit operator bool() const;
+    // Intersect in place with a half-space
+    CELER_CONSTEXPR_FUNCTION void
+    clip(Sense sense, Axis axis, real_type position);
 
   private:
     Real3 lower_;
     Real3 upper_;
+
+    // Implementation of 'from_unchecked' (true type 'tag')
```
elliottbiondo, line 84: true type 'tag'?

src/orange/BoundingBox.hh
```diff
@@ -101,40 +142,55 @@ template<class T>
...
- * Upper bbox coordinate (must be valid).
+ * Whether the bbox is in a nondegenerate state.
  */
 template<class T>
-CELER_FUNCTION auto BoundingBox<T>::upper() const -> Real3 const&
+CELER_CONSTEXPR_FUNCTION BoundingBox<T>::operator bool() const
 {
-    CELER_EXPECT(*this);
-    return upper_;
+    return lower_[to_int(Axis::x)] <= upper_[to_int(Axis::x)]
```
elliottbiondo, line 183: These need to be `<` not `<=` to differentiate between null and degenerate bboxes. Consider the bbox: `{{1, 1, 1}, {2, 2, 2}}`. If you intersect it a null bbox, it should be a no-op. However, if you intersect it with `{{0,0,0}, {0, 0, 0}}`, you should get `{{0, 0, 0}, {2, 2, 2}}`. I could see problems arise downstream from not differentiating between these two cases due to bool returning false in both cases.

src/orange/BoundingBox.hh
```diff
@@ -7,65 +7,81 @@
...
 
-    // Lower bbox coordinate
-    CELER_FORCEINLINE_FUNCTION Real3 const& lower() const;
+    //! Lower bbox coordinate
+    CELER_CONSTEXPR_FUNCTION Real3 const& lower() const { return lower_; }
+
+    //! Upper bbox coordinate
+    CELER_CONSTEXPR_FUNCTION Real3 const& upper() const { return upper_; }
+
+    // Whether the bbox is nondegenerate
```
elliottbiondo: Non-null, change throughout

test/orange/BoundingBoxUtils.test.cc
```diff
@@ -22,43 +26,86 @@ TEST(BoundingBoxUtilsTest, is_inside)
...
+    }
+    {
+        SCOPED_TRACE("double degenerate");
+        auto ddbox = calc_union(BBox{}, BBox{});
+        EXPECT_FALSE(ddbox);
+    }
 }
 
-TEST(BoundingBoxUtilsTest, bbox_union)
+TEST_F(BoundingBoxUtilsTest, bbox_intersection)
```
elliottbiondo, line 128: Explicitly test the null vs degenerate cases here.

elliottbiondo (changes requested):

elliottbiondo:
@sethrj please differentiate between null and degenerate bboxes throughout the code and tests.

src/orange/BoundingBox.hh
sethrj: Good point, I agree completely.

src/orange/BoundingBox.hh
sethrj: If you *intersect* you should get a degenerate bbox, because no point is in both 1 and 2. It can't be simply degenerate because those bounding boxes do contain points that are coincident on the faces.

If you *union* you get `{{0, 0, 0}, {2, 2, 2}}`, which is perhaps what you mean?

I'm adding more tests, comments, and examples to demonstrate all this.

src/orange/BoundingBox.hh
sethrj: A "tag" struct is a type that's only used to assist in type-dependent overloading, or altering the behavior of a function based on some condition. Here, `std::true_type` is a "tag" that means "do not check".

sethrj:
Thanks @elliottbiondo for the feedback! I think I've addressed everythign.

src/orange/BoundingBox.hh
elliottbiondo: Yeah I meant to say union

elliottbiondo (approved):

PR #890, opened 2023-08-14
Add JSON input and more control options to celer-g4

amandalund:
This adds an optional JSON input plus some more physics and field options to the celer-g4 app.


whokion:
@amandalund This is a very nice extension.  It would be good to add a default json something like, app/celer-g4/simple-cms-test.json.in and generate an example json when building celer-g4.

amandalund:
The celer-g4 test failure with `CELERITAS_CORE_GEO="Geant4"`is a timeout where we're seeing tons of the `src/accel/detail/HitProcessor.cc:352: warning: Failed to bump navigation state up to a distance of 1 [mm]` errors (and tracks getting stuck and killed taking zero steps). This started failing when I added a field to the celer-g4 test, but we also get a ton of these errors (~50k) on develop with no field using the geant4 geo.

sethrj:
Thanks @amandalund , maybe we want to just disable that high-level test when using GEANT_CORE_GEO. I've always seen issues with the full-transport navigation (not worth tracking down, since that wasn't the main point), but it sounds out of hand now.

amandalund:
Agreed, probably best to just disable it in that case.. I'll go ahead and do that here.

app/celer-g4/RunInputIO.json.cc
```diff
@@ -0,0 +1,145 @@
...
+    RI_LOAD_OPTION(step_diagnostic_bins);
+
+#undef RI_LOAD_OPTION
+#undef RI_LOAD_REQUIRED
+
+    CELER_VALIDATE(v.physics_list == PhysicsListSelection::geant_physics_list
+                       || !j.contains("physics_options"),
+                   << "'physics_options' can only be specified for "
+                      "'geant_physics_list'");
+    CELER_VALIDATE(v.field != RunInput::no_field()
```
whokion:
```suggestion
    CELER_VALIDATE((v.field != RunInput::no_field() || v.field_type == "rzmap" )
```
which may be better (?) when field_type="rzmap" which does not need a constant field vector.

app/celer-g4/RunInput.cc
```diff
@@ -0,0 +1,31 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2022-2023 UT-Battelle, LLC, and other Celeritas developers.
```
whokion:
```suggestion
// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
```

app/celer-g4/RunInput.hh
```diff
@@ -0,0 +1,100 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2022-2023 UT-Battelle, LLC, and other Celeritas developers.
```
whokion:
```suggestion
// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
```

whokion (approved):
@amandalund All looks good! Thanks.

app/celer-g4/RunInputIO.json.cc
amandalund: Good point, thanks!

app/celer-g4/celer-g4.cc
```diff
@@ -83,20 +84,28 @@ void run(int argc, char** argv)
...
         G4UIExecutive exec(argc, argv);
         exec.SessionStart();
         return;
     }
-
-    CELER_LOG(status) << "Executing macro commands from '" << macro_filename
-                      << "'";
-    ui->ApplyCommand(std::string("/control/execute ")
-                     + std::string(macro_filename));
+    if (std::regex_search(std::string(filename), std::regex("\\.mac$")))
```
sethrj: Or `ends_with` from `corecel/io/StringUtils.hh` 🙃

src/accel/AlongStepFactory.hh
```diff
@@ -98,14 +99,14 @@ class AlongStepFactoryInterface
  * Create an along-step method for a uniform (or zero) field.
  *
  * The constructor is a lazily evaluated function that must return the field
- * vector in native Geant4 units.  If unspecified, the field is zero.
+ * vector. If unspecified, the field is zero.
```
sethrj:
```suggestion
 * definition and driver configuration. If unspecified, the field is zero.
```

src/accel/AlongStepFactory.cc
```diff
@@ -61,18 +59,16 @@ auto UniformAlongStepFactory::operator()(AlongStepFactoryInput const& input) con
         }
 
         // Convert field units from tesla to native celeritas units
-        for (real_type& v : field)
+        for (real_type& v : get_field_().field)
```
sethrj: We should call `get_field_` only once in this function: what if it returns a copy?

Also, if we're really accepting the field *parameters* as the result, I don't think we want to scale those in place. The `celeritas/field` stuff should be unchanged once it's created, and in native celeritas units. I guess the purpose of this is to expose the field driver options through `accel`? You asked me about units at some point; was that about  this use case?

Maybe we should have a separate `DriverOptionsFunction` as a "power user" option for specifying the (celeritas native) options when needed. I definitely think we shouldn't be changing the units of `UniformFieldParams` on the fly.

sethrj (commented):
The field units are a little hinky, but everything else looks great!

src/accel/AlongStepFactory.cc
amandalund: Right, the goal was to be able to specify the field driver options based on user input. I asked about units partly because I wasn't sure at what point it made the most sense to do the conversion -- but you're right that it's not here.

sethrj (approved):
Thanks Amanda!

PR #893, opened 2023-08-16
Implement surface promotion in ORANGE

sethrj:
This adds the ability to convert "special case" quadrics into more general quadrics. This capability is needed for transformations and translations, as well as testing surface simplification.

### DIFF
--- a/src/orange/surf/ConeAligned.hh
+++ b/src/orange/surf/ConeAligned.hh
@@ -132,23 +132,25 @@ CELER_CONSTEXPR_FUNCTION SurfaceType ConeAligned<T>::surface_type()
 /*!
  * Construct from origin and tangent of the angle of its opening.
  *
- * Given a finite cone, the tangent is ratio of the radius at its base to the
- * its height.
+ * Given a finite cone, the tangent is the ratio of its base radius to its
+ * height.
  *
- * \pre
+ * In the cone, below, the tangent of the inner angle
+ * \f$ \tan(\theta) = r/h \f$ is the second argument.
+ * \verbatim
      r
    +-------*
    |   _--^
  h |_--
    O
-   \endpre
+   \endverbatim
  */
 template<Axis T>
 CELER_FUNCTION
 ConeAligned<T>::ConeAligned(Real3 const& origin, real_type tangent)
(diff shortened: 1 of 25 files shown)


src/orange/surf/ConeAligned.hh
```diff
@@ -126,11 +129,11 @@ CELER_CONSTEXPR_FUNCTION SurfaceType ConeAligned<T>::surface_type()
 /*!
  * Construct from origin and tangent of the angle of its opening.
  *
- * Given the triangular cross section of one octant of a finite cone (i.e. a
- * right triangle), the tangent is the slope of its hypotenuse (height / base).
+ * Given a finite cone, the tangent is ratio of the radius at its base to the
+ * its height.
```
mrguilima: The english here is broken.
(I can't find how to suggest changes myself, that button does not show up for me)

mrguilima (dismissed):

src/orange/surf/ConeAligned.hh
sethrj: Misplaced `the` 😄 I'll fix it.

src/orange/surf/Plane.cc
```diff
@@ -0,0 +1,31 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Promote from an axis-aligned plane.
+ */
+template<Axis T>
+Plane::Plane(PlaneAligned<T> const& other) noexcept
+    : normal_{other.calc_normal({0, 0, 0})}, d_{other.position()}
```
elliottbiondo: Could you do `calc_normal(Real3{})` since the argument isn't actually used in PlaneAligned::calc_normal? Or maybe just have a `calc_normal()` that calls `calc_normal(Real3)`? At first glance this looked like a bug since `{0, 0, 0}` is only on the plane when `other.position == 0`.

elliottbiondo (approved):
Looks good @sethrj, just the one minor comment

src/orange/surf/Plane.cc
sethrj: Good point, thanks. I'll add a `calc_normal()` method to the PlaneAligned that returns the normal, then add a `calc_normal(Real3)` to fulfill the necessary interface.

PR #894, opened 2023-08-16
Implement surface simplification

sethrj:
This implements type-safe conversion of more complex surfaces (either user-input or as a result of transformations) into simpler ones using a function-based interface that's compatible with both `Surfaces` (for compressed surface storage) and `std::variant`.

The simplifications here fix several bugs in the simplifications from SCALE ORANGE (see https://code-int.ornl.gov/rnsd/scale/-/issues/1988 ).


sethrj:
@elliottbiondo This one is a doozy but is ready for review. It's pretty heavy on `std::variant` which may be new to you, so I'm happy to walk through it with you in person.

src/orange/surf/SurfaceSimplifier.cc
```diff
@@ -0,0 +1,294 @@
...
+
+ORANGE_INSTANTIATE_OP(CylCentered, CylAligned);
+
+//---------------------------------------------------------------------------//
+/*!
+ * Plane may be flipped, adjusted, or become axis-aligned.
+ *
+ * If a plane has a normal of {-1, 0 + eps, 0}, it will first be truncated to
+ * {-1, 0, 0}, then flipped to {1, 0, 0}, and a new Plane will be return. That
+ * plane can *then* be simplified to
```
sethrj:
```suggestion
 * {-1, 0, 0}, then flipped to {1, 0, 0}, and a new Plane will be returned. That
 * plane can *then* be simplified to an axis-aligned one.
```

src/orange/surf/SurfaceSimplifier.cc
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * TODO: guard against different-signed GQs?
+ */
+auto SurfaceSimplifier::operator()(GeneralQuadric const& gq)
+    -> Optional<SimpleQuadric>
+{
+    SoftZero const soft_zero{tol_};
+
+    auto cross = gq.cross();
+    if (std::all_of(cross.begin(), cross.end(), SoftZero{tol_}))
```
sethrj:
```suggestion
    auto cross = gq.cross();
    if (std::all_of(cross.begin(), cross.end(), SoftZero{tol_}))
```

src/orange/surf/RecursiveSimplifier.hh
```diff
@@ -0,0 +1,98 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply to a surface with an unknown type.
+ */
+template<class F>
+void RecursiveSimplifier<F>::operator()(Sense sense, VariantSurface const& surf)
+{
+    return std::visit(detail::RecursiveSimplifierImpl<F>{func_, sense, tol_},
```
sethrj, line 95:
```suggestion
    CELER_ASSUME(!surf.valueless_by_exception());
    return std::visit(detail::RecursiveSimplifierImpl<F>{func_, sense, tol_},
```

src/orange/surf/SurfaceSimplifier.cc
```diff
@@ -0,0 +1,294 @@
...
+            return *s;
+    }
+    else if (num_pos == 2)
+    {
+        // One element is zero or negative: could be a cylinder or cone
+        if (num_neg == 1)
+        {
+            // Cone: one second-order term less than zero, others equal
+            detail::QuadricConeConverter to_cone{tol_};
+            if (auto c = to_cone(AxisTag<Axis::x>{}, sq))
```
elliottbiondo: for loop over Axis?

src/orange/surf/SurfaceSimplifier.cc
elliottbiondo: ...

src/orange/surf/SurfaceSimplifier.cc
```diff
@@ -0,0 +1,294 @@
...
+        else if (n[ax] < 0)
+        {
+            // Flip the sign of any remaining nonzero axes
+            // (previous axes are zero so just skip them)
+            for (auto ax2 : range(ax, to_int(Axis::size_)))
+            {
+                n[ax2] = 0 - n[ax2];
+            }
+            // Flip sign of d (without introducing -0)
+            d = 0 - d;
```
elliottbiondo: Make into a helper function since you do this a bunch of times

src/orange/surf/SurfaceSimplifier.cc
```diff
@@ -0,0 +1,294 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Plane may be flipped, adjusted, or become axis-aligned.
+ *
+ * If a plane has a normal of {-1, 0 + eps, 0}, it will first be truncated to
+ * {-1, 0, 0}, then flipped to {1, 0, 0}, and a new Plane will be return. That
+ * plane can *then* be simplified to
+ */
+auto SurfaceSimplifier::operator()(Plane const& p)
```
elliottbiondo, line 117: This is a long function with two distinct parts: 1) attempting to convert to axis-aligned, 2) flip/adjust. Consider splitting into two functions (i.e., both called by operator())

src/orange/surf/detail/QuadricConeConverter.hh
```diff
@@ -0,0 +1,127 @@
...
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Try to convert a simple quadric to a cone.
+ *
+ * The simple quadric must have already undergone normalization (zero or one
```
elliottbiondo: Check with DBRC, here and in other other `detail` files

test/orange/surf/RecursiveSimplifier.test.cc
```diff
@@ -0,0 +1,91 @@
...
+class RecursiveSimplifierTest : public ::celeritas::test::Test
+{
+  protected:
+};
+
+TEST_F(RecursiveSimplifierTest, single)
+{
+    std::vector<std::string> result;
+
+    RecursiveSimplifier simplify{[&](Sense sense, auto&& surf) {
```
elliottbiondo: I'd recommend putting the functor on it's own line, here and throughout. It's hard to read as an argument.

src/orange/surf/detail/RecursiveSimplifierImpl.hh
```diff
@@ -0,0 +1,67 @@
...
+
+#include <variant>
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Recursively simplify and call the given function on the final surface.
```
elliottbiondo: Recursively simplify, then call the given function on the final surface.

src/orange/surf/RecursiveSimplifier.hh
```diff
@@ -0,0 +1,98 @@
...
+#include "SurfaceSimplifier.hh"
+#include "VariantSurface.hh"
+#include "detail/AllSurfaces.hh"
+#include "detail/RecursiveSimplifierImpl.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Recursively simplify and call the given function on the final surface.
```
elliottbiondo: This would be a little more clear as:

Recursively simplify, then call the given function on the final surface.

src/orange/surf/RecursiveSimplifier.hh
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Example:
+ * \code
+  RecursiveSimplifier print_simplified([](Sense s, auto&& surf) {
+      cout << to_char(s) << surf << endl;
+  }, 1e-10);
+  print_simplified(Sense::inside, Plane{{1,0,0}, 4});
```
elliottbiondo, line 30: Delete this line

elliottbiondo (approved):
Looks good @sethrj

src/orange/surf/RecursiveSimplifier.hh
sethrj: This shows the templated `operator()` usage; the one below shows the variant usage. I'll add documentation to point out the difference.

test/orange/surf/RecursiveSimplifier.test.cc
sethrj: In general this is bad practice, as it changes the type from `RecursiveSimplifier<{functor}>` to `RecursiveSimplifier<{functor}&>` (potentially more indirection and extending the lifetime+scope of the functor). Since this is a unit test, readability is a plus so I'll implement that change in this file.

src/orange/surf/SurfaceSimplifier.cc
sethrj: Good suggestion. Defined and tested `negate` in `corecel/math/Algorithms.hh`.

src/orange/surf/SurfaceSimplifier.cc
sethrj: It would have to be template metaprogramming since these arguments are types; I figured this was clearer?

src/orange/surf/SurfaceSimplifier.cc
sethrj: I've broken out a couple little pieces to hopefully simplify it further.

src/orange/surf/detail/QuadricConeConverter.hh
sethrj: This would end up repeating the logic for counting zeros... since it's just a detail class used by the one surface simplifier can we skip it?

PR #895, opened 2023-08-16
Add surface "variant" class for host processing

sethrj:
This is a set of helper traits to use `std::variant` with all possible surface types, allowing efficient type-safe construction on CPU without having to use the compressed `Surfaces` class. In other words, we should be able to make a vector of surfaces and simplify them in place, then finally load them into the surface data.


sethrj:
@elliottbiondo This small PR is ready for review.

elliottbiondo (approved):
Looks good @sethrj

PR #896, opened 2023-08-16
Add surface clipping helper class

sethrj:
This reduces the size of bounding boxes based on the infinite and/or finite extent of the "inside" or "outside" space of a quadric.

A separate "clip" class may be needed to reduce a bounding box even further by intersecting surfaces with bounding boxes. (For example, we could construct a bounding box by intersecting a non-aligned plane with a bounding box for a hexagonal prism cell.)


sethrj:
Test failure is a random MPI glitch.

src/orange/surf/SurfaceClipper.hh
```diff
@@ -0,0 +1,95 @@
...
+{
+    CELER_EXPECT(bbox);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply to a surface with a known type.
+ */
+template<class S>
+void SurfaceClipper::operator()(Sense sense, S const& surf)
```
elliottbiondo, line 63: Can't you just `return detail::SurfaceClipperImpl<sense>{bbox_}(surf);`"

src/orange/surf/SurfaceClipper.hh
```diff
@@ -0,0 +1,95 @@
...
+        return detail::SurfaceClipperImpl<Sense::outside>{bbox_}(surf);
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply to a surface with an unknown type.
+ */
+void SurfaceClipper::operator()(Sense sense, VariantSurface const& surf)
+{
```
elliottbiondo, line 80: Same here

elliottbiondo (approved):

src/orange/surf/SurfaceClipper.hh
sethrj: No, because `sense` is runtime, not compile time. If I implement a truly generic visitor like in #911, or added a helper struct, then this could be rewritten as:
```c++
visit_sense_traits([&](auto csense) { return detail::SurfaceClipperImpl<csense>{bbox}; }, sense);
```
which is less error-prone but definitely harder to unpack.

PR #897, opened 2023-08-16
Implement chained transformations/translations

sethrj:
This creates a new transform/translation from an existing one. It uses `std::variant` to reduce the combinatorics of identity/translation/transformation, since presumably all transform-transform operations are being done on the host.

It defines two generic `apply_transform` functions that take a transform variant for the first argument, and a transform/surface variant for the second argument, returning the correctly transformed object.


sethrj:
@elliottbiondo This is now ready for review 🎉

src/orange/surf/VariantSurface.hh
```diff
@@ -18,5 +20,10 @@ namespace celeritas
 //! std::variant class of all surfaces.
 using VariantSurface = detail::VariantSurface_t;
 
+//---------------------------------------------------------------------------//
+// Apply a variant "daughter-to-parent" transform to a surface
+[[nodiscard]] VariantSurface
+apply_transform(VariantTransform const& left, VariantSurface const& right);
```
elliottbiondo: Why `left` and `right` instead of `transform` and `surface`? Consistency with other `apply_transform` or copy-paste error?

src/orange/transform/VariantTransform.cc
```diff
@@ -0,0 +1,98 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Apply the left "daughter-to-parent" transform to the right.
+ *
+ * The resulting variant may be a monostate (identity), translation (no
+ * rotation), or full transformation.
+ *
+ * The resulting transform has rotation
+ * \f[
+   \mathbf{R}' = \mathbf{R}_R
```
elliottbiondo: This is confusing. Are you using `R` to mean both "rotation" and "right"?

src/orange/transform/detail/TransformTransformer.hh
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * An instance of this class applies a daughter-to-parent translation to
+ * another transformation: the translation needed to relocate a "lower"
+ * universe to a new coordinate system in the "higher" universe.
+ *
+ * The operation returns a new transform defined
+ * \f[
+   \mathbf{T}' = \mathbf{T}_L \mathbf{T}_R
+ * \f]
+ * where T is the argument, \f$\mathbf{T}_L\f$ is the constructor argument (the
```
elliottbiondo: `\mathbf{T}_R` is the argument?

src/orange/transform/detail/TransformTransformer.hh
```diff
@@ -0,0 +1,114 @@
...
+ * operator itself), and the result \f$\mathbf{T}'\f$ is the return value.
+ * The resulting transform is:
+ * \f[
+   \mathbf{x}' = \mathbf{T}_L\mathbf{x}
```
elliottbiondo: What is `x` here? Could you double check this section? Lines 47-53 are easy to follow, but 38-45 seem like there might be an error somewhere.

src/orange/transform/detail/TransformTranslator.hh
```diff
@@ -0,0 +1,102 @@
...
+/*!
+ * Apply a translation to a transform to get another transform.
+ *
+ * An instance of this class applies a daughter-to-parent translation to
+ * another transformation: the translation needed to relocate a "lower"
+ * universe to a new coordinate system in the "higher" universe.
+ *
+ * The operation returns a new transform defined
+ * \f[
+   \mathbf{T}' = \mathbf{T}_L \mathbf{T}_R
```
elliottbiondo: I wouldn't use `T` to mean both "transform" and "translation".

src/orange/transform/detail/TransformTranslator.hh
```diff
@@ -0,0 +1,102 @@
...
+ * \f]
+ * where T is the argument, \f$\mathbf{T}_L\f$ is the constructor argument (the
```
elliottbiondo: Again, isn't `\mathbf{T}_R` the argument?

elliottbiondo (changes requested):
Code and tests look good, just a few doc string questions

src/orange/surf/VariantSurface.hh
sethrj: Consistency, but since it doesn't make sense to apply a surface to a transform I could rename them if you think it's clearer.

src/orange/transform/VariantTransform.cc
sethrj: Oh yeah, good point. I copied and modified this documentation but clearly didn't check it well.

src/orange/surf/VariantSurface.hh
elliottbiondo: I think it would be clearer to rename them, but I don't have a strong opinion

src/orange/transform/VariantTransform.cc
elliottbiondo: Oops, I think you still need to fix this one

elliottbiondo (changes requested):
The doc strings now look correct to me, but I think you missed one. Otherwise good to go.

elliottbiondo (approved):

PR #899, opened 2023-08-17
Add e-/e+ single Coulomb scattering process

hhollenb:
This integrates the `CoulombScatteringProcess` into Celeritas, which builds the model `WentzelModel`. It is enabled in `ProcessBuilder` and the corresponding test data has been updated to include Coulomb data from Geant. Proton id's were disabled because they're generally not supported at the moment, and I placed `TODO`s where I made changes.

Some attention is needed for `CoulombScatteringProcess::Options`. I needed data from `ImportEmParameters` as well as passing options for the form factor, which the other process don't really have outside multiple scattering. I implemented it as close as I could to other classes, but I'd like some feedback if I'm missing something.


src/celeritas/em/data/WentzelData.hh
```diff
@@ -24,11 +24,12 @@ struct WentzelIds
     ActionId action;
     ParticleId electron;
     ParticleId positron;
-    ParticleId proton;
+    // ParticleId proton; // TODO: enable when protons are supported
```
sethrj: I think you can just leave this as-is, since emitting ions is "optional" behavior

src/celeritas/ext/detail/GeantPhysicsList.cc
```diff
@@ -349,3 +349,4 @@ void GeantPhysicsList::add_e_processes(G4ParticleDefinition* p)
 //---------------------------------------------------------------------------//
 }  // namespace detail
 }  // namespace celeritas
+
```
sethrj: Stray committed line

test/celeritas/phys/ProcessBuilder.test.cc
```diff
@@ -446,6 +447,54 @@ TEST_F(ProcessBuilderTest, rayleigh)
...
+                EXPECT_TRUE(builders[elcomp_idx]);
+            }
+        }
+    }
+}
+
 //---------------------------------------------------------------------------//
 }  // namespace test
 }  // namespace celeritas
+
```
sethrj: What editor are you using? 😂

src/celeritas/em/process/CoulombScatteringProcess.hh
```diff
@@ -0,0 +1,80 @@
...
+{
+    //! Nuclear form factor model
+    NuclearFormFactorType form_factor_model{NuclearFormFactorType::exponential};
+
+    //! User defined screening factor
+    real_type screening_factor{1};
+
+    //! Check if the options are valid
+    explicit operator bool() const { return screening_factor > 0; }
+};
```
sethrj: Make these options a nested class inside CoulombScatteringProcess, rather than a detail. (The only reason to avoid nested classes is when the class is templated.)

src/celeritas/em/model/WentzelModel.cc
```diff
@@ -49,17 +50,18 @@ WentzelModel::WentzelModel(ActionId id,
     host_data.ids.action = id;
     host_data.ids.electron = particles.find(pdg::electron());
     host_data.ids.positron = particles.find(pdg::positron());
-    host_data.ids.proton = particles.find(pdg::proton());
+    // TODO: Enable protons when supported
+    // host_data.ids.proton = particles.find(pdg::proton());
```
sethrj: You can revert this change too, since proton will just be "false" if it's not there.

src/celeritas/em/process/CoulombScatteringProcess.cc
```diff
@@ -0,0 +1,70 @@
...
+    return imported_.step_limits(std::move(applic));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Name of the process.
+ */
+std::string CoulombScatteringProcess::label() const
+{
+    return "Coulomb Scattering";
```
sethrj:
```suggestion
    return "Coulomb scattering";
```

sethrj (commented):
Looking great! Just a couple of small comments.

test/celeritas/phys/ProcessBuilder.test.cc
hhollenb: the cluster I work on has incompatible versions of clang and cuda modules, so I regularly accidentally make commits without having clang-format installed. I wrote a script that de-loads cuda, loads clang-format, appends a newline to the end of each file in the most recent commit, then adds them and does a `commit --amend` to force re-run the formatting. I didn't think I was having issues with the last PR but I guess now I need to do something less hacky haha

src/celeritas/em/process/CoulombScatteringProcess.hh
hhollenb: I started off doing this, but I'd like to also pass the options to the model since that's what actually needs them. Including the process header was giving me a lot of cuda compiler errors (I'm purely guessing but I think it's because the process class is only host code while the model will be both host and device and they'll generate conflicting definitions in includes like `MaterialView`). Being able to forward declare the options class in the model file fixes this issue as the process only needs to be included in the .cc file.

I can fully unpack the options and pass them to the model, or I can also create option classes for the model and just copy the data over, but I don't think my attempt at nested classes will work.

test/celeritas/phys/ProcessBuilder.test.cc
hhollenb: Actually it only adds a space at the end of the `} // namespace celeritas` which clang-format normally gets, so I'm not sure where the newlines are coming from

sethrj: If you're doing `echo " " >> blah.hh` that would explain the extra newline. Also if your clang-format module is incompatible, you could try making a `function git-clang-format() { exec /full/path/to/gcf }` and setting the git `clangformat.binary` config option to the full path to clang-format?

src/celeritas/em/process/CoulombScatteringProcess.hh
sethrj: OK forward declarations are another  good reason not to nest the class 😄. I think you're right about `MaterialView`: I've seen that as well. I think it'd be better to put the options at the model level, since these are really adjusting the model implementation rather than the behavior of the process. Or are they reusable across multiple models in practice? If they're specific to the Wenzel model, call them `WenzelModelOptions` and put them in the `WenzelModel` header (and in the main namespace, not `detail`).

src/celeritas/em/process/CoulombScatteringProcess.hh
```diff
@@ -0,0 +1,80 @@
...
+                             Options const& options);
+
+    //! Construct the models associated with this process
+    VecModel build_models(ActionIdIter start_id) const final;
+
+    //! Get the interaction cross sections fro the given energy range
+    StepLimitBuilders step_limits(Applicability range) const final;
+
+    //! Whether to use the integral method to sample interaction length
+    bool use_integral_xs() const final { return false; }
```
amandalund: I think since electrons can lose energy over the step we should be using the integral approach to sample the interaction length -- in that case we'd want to add `use_integral_xs` to the process `Options` (true by default) and return that value here.

src/celeritas/em/process/CoulombScatteringProcess.hh
```diff
@@ -0,0 +1,80 @@
...
+    //! Construct from Wentzel scattering data
+    CoulombScatteringProcess(SPConstParticles particles,
+                             SPConstMaterials materials,
+                             SPConstImported process_data,
+                             Options const& options);
+
+    //! Construct the models associated with this process
+    VecModel build_models(ActionIdIter start_id) const final;
+
+    //! Get the interaction cross sections fro the given energy range
```
amandalund:
```suggestion
    //! Get the interaction cross sections for the given energy range
```

amandalund (commented):
Looks great @hhollenb!

test/celeritas/phys/ProcessBuilder.test.cc
hhollenb: Even though I have the path to the binary, the shared object libraries it depends on are incompatible with the other loaded modules and get force de-loaded unfortunately. I have modified the script to use the same `git-clang-format HEAD^` that the git post-commit hook uses so I shouldn't have to do any modifying file hacks anymore.

sethrj (approved):
Great work!

amandalund (approved):

PR #900, opened 2023-08-17
Add ROOT-based event exporter

stognini:
This PR adds the option to export Celeritas' primaries to a ROOT file and read those primaries back into Celeritas.

- RootOffloadWriter writes data to a ROOT file using basic types to avoid the use of dictionaries.
- RootOffloadReader reads back the data in an event-by-event basis.

The reader can be used by a future `class RootPrimaryGenerator final : public G4VPrimaryGenerator { };`.

The use of a ROOT file considerably decreases the output file size, as expected. For the standard `celer-g4` test using `celeritas/app/data/ttbarbits-12evt-118prim.hepmc3`:

```
1.9M  simple-cms.offloaded.hepmc3
357K  simple-cms.offloaded.root
```


src/accel/detail/RootOffloadReader.hh
```diff
@@ -0,0 +1,111 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file accel/detail/RootOffloadReader.hh
```
sethrj: This should be `celeritas/io/EventReader` since we will need to use it from the celer-sim app.

src/accel/detail/RootOffloadWriter.hh
```diff
@@ -0,0 +1,91 @@
...
+
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Export primary data to ROOT.
+ *
+ * One TTree entry represents one primary.
+ */
+class RootOffloadWriter
```
sethrj: Likewise since this class doesn't depend on anything in `accel` and we can use it for writing out whatever primaries, put this in celeritas/io as `RootEventWriter`. The only thing that `OffloadWriter` did was to wrap a class for thread safety.

src/accel/detail/RootOffloadWriter.cc
```diff
@@ -0,0 +1,79 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Export primaries to ROOT.
+ */
+void RootOffloadWriter::operator()(Primaries const& primaries)
+{
+    CELER_EXPECT(!primaries.empty());
+
+    std::scoped_lock{write_mutex_};
```
sethrj: You should remove this. The `OffloadWriter` is the only thing that needs to be thread safe. If we need thread safety outside what ROOT provides we can use the existing OffloadWriter and make its `EventWriter` a generic `std::function<void(vector<Primary>)>`. Also note that the way the reader is structured, you **MUST** lock out the file when writing an event, otherwise primaries from different events could be interleaved.

src/celeritas/ext/RootUniquePtr.root.cc
```diff
@@ -29,6 +30,21 @@ void RootWritableDeleter<T>::operator()(T* ptr)
...
+ * Call `TTree->AutoSave()` before deletion. Used by TTree when multiple tfiles
+ * are open at the same time.
+ */
+template<class T>
+void RootAutoSaveDeleter<T>::operator()(T* ptr)
+{
+    CELER_EXPECT(ptr);
+    CELER_LOG(debug) << "Writing " << ptr->ClassName() << " '"
+                     << ptr->GetName() << "'";
+    ptr->AutoSave();
```
sethrj, line 40: Does any other interface have `AutoSave`  or just TTree? If the latter you should just make it a template specialization. Or just make separate deleters for ttree and tfile.

src/celeritas/ext/RootUniquePtr.root.cc
```diff
@@ -48,6 +64,7 @@ void ExternDeleter<T>::operator()(T* ptr)
 //---------------------------------------------------------------------------//
 template struct RootWritableDeleter<TFile>;
 template struct RootWritableDeleter<TTree>;
+template struct RootAutoSaveDeleter<TTree>;
```
sethrj: Delete the WritableDeleter<TTree> since it doesn't do the expected thing.

src/accel/SharedParams.hh
```diff
@@ -87,13 +88,20 @@ class SharedParams
...
 
     //! Optional offload writer, only for use by LocalTransporter
     SPOffloadWriter const& offload_writer() const { return offload_writer_; }
 
+    //! Optional ROOT offload writer, only for use by LocalTransporter
+    SPRootOffloadWriter const& root_offload_writer() const
+    {
+        return root_offload_writer_;
+    }
+
```
sethrj: Delete all this code; we should be using the same offload writer interface: either HepMC3 or ROOT, we don't need separate copies of both. I just wanted you to add the `RootEvent{Reader,Writer}` class and I was going to add in a separate PR the extra bit of `std::functional` and "make a different writer based on the output extension" code needed to hook it in.

sethrj (changes requested):

src/accel/detail/RootOffloadWriter.cc
stognini: Just to be sure I got your point: so you're going to lock the `operator()` operation  itself, whether that's a HepMC3 or ROOT writer?

sethrj: Right now the `OffloadWriter` in accel has a single purpose: to wrap an `EventWriter` with a thread safe lock. I think we'll use that same mechanism to wrap the RootWriter, since for the reading to work, we have to write all tracks consecutively for a single event.

src/celeritas/ext/RootUniquePtr.root.cc
pcanal: `AutoSave` is only seen in `TTree`.  `TObject->Write` should probably not be used in general except for `TFile`.   So it might actually indeed make sense to have 2 concrete deletor.  

Side note: Instead `TObject->Write`, especially in a threaded environment one would want to use `dir->WriteObject(ptr)`. i.e. require the writer to specify exactly where to write. (Only `TTree` and Histogram know which file they are attached to/intended for)

src/accel/CMakeLists.txt
```diff
@@ -35,6 +35,10 @@ endif()
 if(CELERITAS_USE_HepMC3)
   list(APPEND PRIVATE_DEPS HepMC3::HepMC3)
   list(APPEND SOURCES HepMC3PrimaryGenerator.cc)
+
+  if (CELERITAS_USE_ROOT)
+    list(APPEND PRIVATE_DEPS ROOT::Tree)
+  endif()
```
sethrj: This can be reverted since the ROOT code is no longer in `accel`.

src/accel/SetupOptions.hh
```diff
@@ -112,6 +112,8 @@ struct SetupOptions
     std::string physics_output_file;
     //! Filename to dump a HepMC3 copy of offloaded tracks as events
     std::string offload_output_file;
+    //! Filename to dump a ROOT copy of offloaded tracks as events
+    std::string offload_root_output_file;
```
sethrj: Remove this; we'll reuse `offload_output_file` and switch format based on the file extension.

src/celeritas/ext/RootUniquePtr.root.cc
```diff
@@ -16,8 +16,7 @@ namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Call `TObject->Write()` before deletion. Used by TFile and TTree writer
- * classes.
+ * Call `TObject->Write()` before deletion.
```
sethrj: This would be a good place to mention that `Write` doesn't do what one would expect 😒

src/celeritas/ext/RootUniquePtr.root.cc
sethrj: OK I agree we should have separate non-templated `RootTreeDeleter` and `RootFileDeleter` then.

src/celeritas/io/EventReader.cc
```diff
@@ -86,21 +98,23 @@ auto EventReader::operator()() -> result_type
...
+            }
+            // Set the registered ID of the particle
+            primary.particle_id = particle_id;
         }
 
-        Primary primary;
-
-        // Set the registered ID of the particle
-        primary.particle_id = particle_id;
-
```
sethrj: I don't understand the use case for having an event reader that can't read the particle type. What can we do with primaries that are unknown particles? In all of the applications I can imagine, we'll be loading events into a simulation (or a test) that defines particle types already. I think the changes to EventReader.* should be reverted.

src/celeritas/io/RootEventReader.cc
```diff
@@ -0,0 +1,101 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with ROOT input filename.
+ */
+RootEventReader::RootEventReader(std::string const& filename,
+                                 SPConstParticles params)
+    : params_(std::move(params))
+{
+    CELER_EXPECT(!filename.empty());
```
sethrj, line 28: Add a `ScopedRootErrorHandler` here

src/celeritas/io/RootEventReader.cc
```diff
@@ -0,0 +1,101 @@
...
+    CELER_ASSERT(num_entries_ > 0);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Read single event from the primaries tree.
+ */
+auto RootEventReader::operator()() -> result_type
+{
+    CELER_EXPECT(entry_count_ <= num_entries_);
```
sethrj, line 47: Add another `ScopedRootErrorHandler` here too.

src/celeritas/io/RootEventReader.cc
```diff
@@ -0,0 +1,101 @@
...
+            = params_->find(PDGNumber{this->from_leaf<int>("particle")});
+        primary.energy
+            = units::MevEnergy{this->from_leaf<real_type>("energy")};
+        primary.time = this->from_leaf<real_type>("time");
+        primary.position = this->from_array_leaf("pos");
+        primary.direction = this->from_array_leaf("dir");
+        primaries.push_back(std::move(primary));
+    }
+
+    CELER_LOG_LOCAL(info) << "Read event " << this_evt_id << " with "
```
sethrj:
```suggestion
    CELER_LOG_LOCAL(debug) << "Read event " << this_evt_id << " with "
```

src/celeritas/io/RootEventReader.cc
```diff
@@ -0,0 +1,101 @@
...
+ * Helper function to fetch leaves containing `std::array<double, 3>`.
+ */
+Real3 RootEventReader::from_array_leaf(char const* leaf_name)
+{
+    CELER_EXPECT(ttree_);
+    auto const leaf = ttree_->GetLeaf(leaf_name);
+    CELER_ASSERT(leaf);
+    CELER_ASSERT(leaf->GetLen() == 3);
+    return {leaf->GetValue(0), leaf->GetValue(1), leaf->GetValue(2)};
+}
```
sethrj, line 115: Depending on how often we write a non-dictionary ROOT file it might be h
```c++
class RootLeafReader
{
explicit RootLeafReader(TTree* tree);

template<class T>
void operator()(char const* leaf_name, T* value) { /* default implementation */ }

template<class T, class N>
void from_leaf(char const* leaf_name, Array<T, N>* value) { /* call GetValue on each item */ }

template<class T>
void from_leaf(char const* leaf_name, OpaqueId<T>* value) {...}
```

so that you can make a local `RootLeafReader read_leaf{ttree_.get()}; read_leaf("pos", p.pos);` etc.

Probably not worth it for this small file but something to keep in mind.

sethrj (changes requested):

sethrj:
@stognini I refactored the event IO test (originally used for HepMC3 only) so that it can be reused for testing the ROOT reader and writer. Please take a look at the output: I think we're missing (1) the check to ensure that all primaries being written at once have the same event ID and (2) resetting the track IDs when reading back in. In fact I think we can entirely delete the track IDs from the ROOT output since we have to reset them sequentially anyway.

stognini:
> I refactored the event IO test (originally used for HepMC3 only) so that it can be reused for testing the ROOT reader and write

Thanks! 

>  In fact I think we can entirely delete the track IDs from the ROOT output since we have to reset them sequentially anyway.

Agreed. removed `track_id` branch from ROOT and manually set them during reading.

Now there is one detail that I am not sure how to proceed with the test: The `event_id` in `EventWriter` is set manually and sequentially, whereas the `RootEventWriter` initially simply assigned whatever `event_id` is stored in `celeritas::Primary` at the time, leading to the different results shown in the test. I have added an event id counter to the `RootEventWriter` to _manually match_ the exact behavior of `EventWriter`. Let me know if that is the way you prefer, or if I should amend the test and keep the `Root{reader/writer}` as previously (i.e. simply accepting that whatever information in `vector<Primary>` is correct to begin with).

sethrj:
> The event_id in EventWriter is set manually and sequentially, whereas the RootEventWriter initially simply assigned whatever event_id is stored in celeritas::Primary at the time, leading to the different results shown in the test. I have added an event id counter to the RootEventWriter to manually match the exact behavior of EventWriter. Let me know if that is the way you prefer, or if I should amend the test and keep the Root{reader/writer} as previously (i.e. simply accepting that whatever information in vector<Primary> is correct to begin with).

One big problem with the previous method is that interleaving the event IDs breaks the reader. Keeping the exact same behavior as the HepMC3 writer limits us, but it ensures no surprises when switching between `offload_output_file = "foo.hepmc3";` and `offload_output_file = "foo.root";`

src/accel/CMakeLists.txt
sethrj: @stognini You might have missed this comment earlier?

src/accel/SetupOptions.hh
sethrj: Also this one?

src/celeritas/io/EventReader.hh
```diff
@@ -46,13 +46,17 @@ class EventReader
...
+
+    // Construct from a filename and particle params
     EventReader(std::string const& filename, SPConstParticles params);
 
     //! Prevent copying and moving
     CELER_DELETE_COPY_MOVE(EventReader);
 
     // Read a single event from the event record
     result_type operator()();
+    // Use a template and return T as HepMC3RootEvent or celeritas Primary?
```
sethrj: This file should be reverted: event reader should always require particle params.

src/celeritas/io/EventReader.cc
sethrj: @stognini I think you missed this comment too.

src/celeritas/io/RootEventReader.cc
```diff
@@ -0,0 +1,118 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Read single event from the primaries tree.
+ */
+auto RootEventReader::operator()() -> result_type
+{
+    ScopedRootErrorHandler scoped_root_error;
+
+    CELER_EXPECT(entry_count_ <= num_entries_);
```
sethrj: Put preconditions at the very very top. Same with the constructor above.

```suggestion
    CELER_EXPECT(entry_count_ <= num_entries_);

    ScopedRootErrorHandler scoped_root_error;
```

src/celeritas/io/RootEventReader.hh
```diff
@@ -0,0 +1,104 @@
...
+template<class T>
+auto RootEventReader::from_leaf(char const*) -> T
+{
+    CELER_ASSERT_UNREACHABLE();
+}
+
+inline Real3 RootEventReader::from_array_leaf(char const*)
+{
+    CELER_ASSERT_UNREACHABLE();
+}
```
sethrj: I don't think you need to provide definitions for these since they're never called if ROOT is disabled. Did you see a compiler warning-turned-error if they are gone?

sethrj (changes requested):

src/accel/SetupOptions.hh
stognini: Ahh crap. Thanks for noticing this. I checked out back a few files from a previous commit just to test a `celer-g4` run and undid my fixes. I'll clean this up.

src/celeritas/io/RootEventReader.hh
stognini: Interesting, they can indeed be removed. I haven't tested without them initially as I assumed that a missing definition would throw a Werror.

sethrj: It's only a missing definition if it's called from somewhere.

sethrj (approved):
Looks good to me, sorry this took so long. I'll hook it up to `accel` next.

PR #901, opened 2023-08-17
Refactor StepLimit to SoA

esseivaju:
Convert the StepLimit collection in SimStateData to a SoA in preparation for using radix sort as having contiguous ActionId will help with access pattern and is even necessary with thrust/cub. 

The `StepLimit` struct is kept to group return values / functions arguments.

### DIFF
--- a/src/celeritas/global/alongstep/detail/PropagationApplier.hh
+++ b/src/celeritas/global/alongstep/detail/PropagationApplier.hh
@@ -156,7 +156,8 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView const& track)
     {
         // Stopped at a geometry boundary: this is the new step action.
         CELER_ASSERT(p.distance <= sim.step_length());
-        sim.force_step_limit({p.distance, track.boundary_action()});
+        sim.step_length(p.distance);
+        sim.post_step_action(track.boundary_action());
     }
     else if (p.distance < sim.step_length())
     {
@@ -164,7 +165,8 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView const& track)
         // all in the field propagator, and will get bumped a small
         // distance. This primarily occurs with reentrant tracks on a
         // boundary with VecGeom.
-        sim.force_step_limit({p.distance, track.propagation_limit_action()});
+        sim.step_length(p.distance);
+        sim.post_step_action(track.boundary_action());
     }
 }
 
--- a/src/celeritas/global/alongstep/detail/PropagationApplier.hh
+++ b/src/celeritas/global/alongstep/detail/PropagationApplier.hh
@@ -119,12 +119,12 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView const& track)
         // The track is looping, i.e. progressing little over many
         // integration steps in the field propagator (likely a low energy
         // particle in a low density material/strong magnetic field).
-        sim.step_length() = p.distance;
+        sim.step_length(p.distance);
 
         // Kill the track if it's stable and below the threshold energy or
         // above the threshold number of steps allowed while looping.
         auto particle = track.make_particle_view();
-        sim.force_step_limit([&track, &particle, &sim] {
+        sim.post_step_action([&track, &particle, &sim] {
             if (particle.is_stable()
                 && sim.is_looping(particle.particle_id(), particle.energy()))
             {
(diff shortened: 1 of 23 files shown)


src/celeritas/track/SimTrackView.hh
```diff
@@ -95,11 +95,14 @@ class SimTrackView
...
 
-    // Limiting step and action to take
-    CELER_FORCEINLINE_FUNCTION StepLimit const& step_limit() const;
+    // Limiting step
+    CELER_FORCEINLINE_FUNCTION real_type step_length() const;
 
-    // Mutable access to step limit (TODO: hack)
-    CELER_FORCEINLINE_FUNCTION StepLimit& step_limit();
+    // Mutable access to limiting step
+    CELER_FORCEINLINE_FUNCTION real_type& step_length();
```
esseivaju: we probably want to have a consistent way of updating the step length, along_step_action and post_step_action. Currently, we return a reference to step length and along_step_action and we have a setter force_step_limit for the post_step_action.

sethrj: Yeah. How about a `void step_limit(real_type length, ActionId post_step);` and a `void step_limit(ActionId post_step);` for the times where we directly set the post-step action? Maybe we should take a quick survey of when these are used?

esseivaju: There's one case where the step length is update alone in [UrbanMsc](https://github.com/celeritas-project/celeritas/blob/e16e9285edecaeb619c2f614c4deb16864050fce/src/celeritas/em/msc/UrbanMsc.hh#L182-L187). `step_limit(real_type length, ActionId post_step)` checks that the step length is limiting before updating the action and the step length, we sometimes need to update these regardless (e.g. looping track in PropagationApplier)

sethrj: OK. Let's just add raw accessors `void step_length(real_type length)` and `void post_step_action(ActionId post_step)`; and the first one can at least assert that the length is strictly positive.

src/celeritas/track/SimData.hh
```diff
@@ -131,8 +130,8 @@ struct SimStateData
                             //!< [s]
 
     Items<TrackStatus> status;
-    // TODO: separate into post-step action, distance
-    Items<StepLimit> step_limit;
+    Items<real_type> steps;
```
sethrj:
```suggestion
    Items<real_type> step_length;
```

src/celeritas/track/SimTrackView.hh
```diff
@@ -255,10 +260,10 @@ CELER_FUNCTION void SimTrackView::force_step_limit(ActionId action)
  */
 CELER_FUNCTION void SimTrackView::force_step_limit(StepLimit const& sl)
 {
-    CELER_ASSERT(sl.step >= 0
-                 && sl.step <= states_.step_limit[track_slot_].step);
+    CELER_ASSERT(sl.step >= 0 && sl.step <= states_.steps[track_slot_]);
 
-    states_.step_limit[track_slot_] = sl;
+    states_.steps[track_slot_] = sl.step;
+    states_.post_step_action[track_slot_] = sl.action;
```
sethrj: We might be able to eliminate this function now that we can manipulate the length and action separately. It seems to be used only once in the main branch (DiscreteSelectExecutor) which was even eliminated during this update.

sethrj (approved):
Looks much cleaner, thanks for doing this @esseivaju . Only two minor changes needed.

src/celeritas/global/alongstep/detail/PropagationApplier.hh
```diff
@@ -156,18 +155,18 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView const& track)
...
+    else if (p.distance < sim.step_length())
     {
         // Some tracks may get stuck on a boundary and fail to move at
         // all in the field propagator, and will get bumped a small
         // distance. This primarily occurs with reentrant tracks on a
         // boundary with VecGeom.
-        step_limit.step = p.distance;
-        step_limit.action = track.propagation_limit_action();
+        sim.step_length(p.distance);
+        sim.post_step_action(track.boundary_action());
```
sethrj: @esseivaju I think you made a copy-paste error here in your final change set?

esseivaju: oh yeah sorry! Good that you caught that.

PR #903, opened 2023-08-23
Use radix sort for ordering tracks

esseivaju:
Update the current implementation of sorting tracks to use radix sort. The goal of the first iteration is to adapt the collections to use Thrust's radix sort implementation which only works on arithmetic types. We need to launch an extra kernel that will follow the thread to track slot indirection and fill an array that can be used as a key to sort the track_slots indirection. This will still be useful if we decide to implement our own radix sort or use cub instead of thrust.

### DIFF
--- a/src/celeritas/track/detail/TrackSortUtils.cu
+++ b/src/celeritas/track/detail/TrackSortUtils.cu
@@ -19,6 +19,7 @@
 #include "corecel/Macros.hh"
 #include "corecel/data/Collection.hh"
 #include "corecel/data/Copier.hh"
+#include "corecel/data/DeviceVector.hh"
 #include "corecel/data/ObserverPtr.device.hh"
 #include "corecel/data/ObserverPtr.hh"
 #include "corecel/sys/Device.hh"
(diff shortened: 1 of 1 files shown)


src/celeritas/track/detail/TrackSortUtils.cu
```diff
@@ -59,14 +60,39 @@ void partition_impl(TrackSlots const& track_slots, F&& func, StreamId stream_id)
...
-    auto start = device_pointer_cast(track_slots.data());
-    thrust::sort(thrust::device.on(celeritas::device().stream(stream_id).get()),
-                 start,
-                 start + track_slots.size(),
-                 std::forward<F>(func));
+    if (ThreadId tid = celeritas::KernelParamCalculator::thread_id();
+        tid < size)
+    {
+        out_actions.get()[tid.get()]
+            = actions.get()[track_slots.get()[tid.get()]].get();
```
esseivaju: How many get can we get on one statement?

esseivaju:
Unfortunately I forgot my hardware key with OTP tokens so I can't compare the timing on Perlmutter right now but looking at the ptxas output during compilation thrust is using radix sort:
```
ptxas info    : Compiling entry function '_ZN3cub31DeviceRadixSortSingleTileKernelINS_21DeviceRadixSortPolicyIjjjE9Policy800ELb0EjjjEEvPKT1_PS4_PKT2_PS8_T3_ii' for 'sm_80'
ptxas info    : Function properties for _ZN3cub31DeviceRadixSortSingleTileKernelINS_21DeviceRadixSortPolicyIjjjE9Policy800ELb0EjjjEEvPKT1_PS4_PKT2_PS8_T3_ii
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 128 registers, 396 bytes cmem[0]
```

Also, regarding writing our own implementation of radix sort, one argument is that we know the number of items but it looks like [cub API](https://nvlabs.github.io/cub/structcub_1_1_device_radix_sort.html) allows us to specify that as well as the begin/end bit for keys comparison so do we really have an advantage over just using cub?

src/celeritas/track/detail/TrackSortUtils.cu
sethrj: There is an open issue about improving the nomenclature 😅

And maybe we need an array accessor too?

sethrj:
Nice! And platform portability is the only thing I can think of

esseivaju:
yeah platform portability is definitely one thing but if we implement our own radix sort wouldn't we want to make use of platform-specific instructions anyway, otherwise it kind of defeats the point of using radix sort in the first place.

sethrj (approved):
Nice find! Thanks.

PR #904, opened 2023-08-24
Connect ROOT event writer to accel and reader to celer-sim

sethrj:
- Add the ability to offload to ROOT from accel
- Fix the offload keyword and a Geant4 integer overflow problem in celer-g4
- Add the ability to generate events from a .root file in celer-sim

### DIFF
--- a/app/celer-g4/GlobalSetup.cc
+++ b/app/celer-g4/GlobalSetup.cc
@@ -153,7 +153,7 @@ void GlobalSetup::ReadInput(std::string const& filename)
     ui->ApplyCommand("/celer/cuda/defaultStream"
                      + to_string(input_.default_stream));
     ui->ApplyCommand("/celer/outputFile " + input_.output_file);
-    ui->ApplyCommand("/celer/offloadFile " + input_.offload_file);
+    ui->ApplyCommand("/celer/offloadOutputFile " + input_.offload_output_file);
 
     // Execute macro for Geant4 commands (e.g. to set verbosity)
     if (!input_.macro_file.empty())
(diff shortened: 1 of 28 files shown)


app/celer-sim/Runner.cc
```diff
@@ -361,23 +366,17 @@ void Runner::build_events(RunnerInput const& inp)
...
-        {
-            append(event);
-            event = generate_event(rng);
-        }
+        read_events([&generate_event, &rng] { return generate_event(rng); });
+    }
+    else if (ends_with(inp.event_filename, ".root"))
+    {
+        read_events(
+            RootEventReader(inp.event_filename, core_params_->particle()));
```
stognini, line 374: Since now we read from both HepMC3 and ROOT, we should update the doxygen comment of this function to something like
```
Read events from a HepMC3/ROOT input file or build using a primary generator.
```

app/celer-g4/GlobalSetup.cc
```diff
@@ -134,30 +134,26 @@ void GlobalSetup::ReadInput(std::string const& filename)
     nlohmann::json::parse(infile).get_to(input_);
     CELER_ASSERT(input_);
 
+    options_->output_file = input_.output_file;
+    options_->offload_output_file = input_.offload_output_file;
+    // XXX where are geometry and physics file set?
```
amandalund: I guess we only want to set the geometry filename in the accel `SetupOptions` if we want to load the geometry from the GDML rather than import it from Geant4... maybe we need another input option to specify which to do? The physics output file I never added to the celer-g4 run input, so we'd need to add that as well.

amandalund (approved):
Looks good @sethrj!

stognini (approved):

stognini:
Thanks @sethrj . Looks great.

app/celer-g4/GlobalSetup.cc
sethrj: I think we can move the geometry file from `RunAction::BeginOfRunAction`, and I might as well add the physics output option while i'm at it.

PR #905, opened 2023-08-25
Define CELER_ASSUME for simpler always-on compiler assumptions

sethrj:
This is a combination of an assertion and a compiler optimization (which remains enabled during optimized code). It's meant as an alternative to `CELER_ASSERT_UNREACHABLE()` that can help with bounds checking and other edge cases. The use case implemented in the variant dispatchers (previously with `CELER_ASSERT_UNREACHABLE`) reduces the object code size of `VariantTransform.o` (with `-O2` and no debug assertions) by 10% by avoiding the jump to exception handling code and the code itself:
```diff
$ diff -u VariantTransform.orig.s VariantTransform.s
--- VariantTransform.orig.s	2023-08-24 20:38:55
+++ VariantTransform.s	2023-08-24 20:38:29
@@ -12,14 +12,10 @@
 	.cfi_def_cfa w29, 16
 	.cfi_offset w30, -8
 	.cfi_offset w29, -16
-	str	x1, [sp]
-	ldr	w10, [x0, #96]
-	cmn	w10, #1
-	b.eq	LBB0_2
-; %bb.1:
 	mov	x9, x0
+	ldr	w10, [x0, #96]
 	mov	x11, sp
-	str	x11, [sp, #8]
+	stp	x1, x11, [sp]
 Lloh0:
 	adrp	x11, l___const._ZNSt3__116__variant_detail12__visitation6__base11__visit_altB6v15006INS1_9__variant15__value_visitorIN9celeritas12_GLOBAL__N_126VariantTransformDispatcherEEEJRKNS0_6__implIJNS_9monostateENS6_11TranslationENS6_14TransformationEEEEEEEDcOT_DpOT0_.__fmatrix@PAGE
 Lloh1:
@@ -31,55 +27,9 @@
 	ldp	x29, x30, [sp, #16]             ; 16-byte Folded Reload
 	add	sp, sp, #32
 	ret
-LBB0_2:
-	bl	std::__1::__throw_bad_variant_access[abi:v15006]()
 	.loh AdrpAdd	Lloh0, Lloh1
 	.cfi_endproc
                                         ; -- End function
-	.private_extern	std::__1::__throw_bad_variant_access[abi:v15006]() ; -- Begin function _ZNSt3__126__throw_bad_variant_accessB6v15006Ev
-	.globl	std::__1::__throw_bad_variant_access[abi:v15006]()
-	.weak_def_can_be_hidden	std::__1::__throw_bad_variant_access[abi:v15006]()
-	.p2align	2
-std::__1::__throw_bad_variant_access[abi:v15006](): ; @_ZNSt3__126__throw_bad_variant_accessB6v15006Ev
-	.cfi_startproc
-; %bb.0:
-	stp	x29, x30, [sp, #-16]!           ; 16-byte Folded Spill
-	.cfi_def_cfa_offset 16
-	mov	x29, sp
-	.cfi_def_cfa w29, 16
-	.cfi_offset w30, -8
-	.cfi_offset w29, -16
-	mov	w0, #8
-	bl	___cxa_allocate_exception
-Lloh2:
-	adrp	x8, vtable for std::bad_variant_access@GOTPAGE
-Lloh3:
-	ldr	x8, [x8, vtable for std::bad_variant_access@GOTPAGEOFF]
-	add	x8, x8, #16
-	str	x8, [x0]
-Lloh4:
-	adrp	x1, typeinfo for std::bad_variant_access@GOTPAGE
-Lloh5:
-	ldr	x1, [x1, typeinfo for std::bad_variant_access@GOTPAGEOFF]
-Lloh6:
-	adrp	x2, std::bad_variant_access::~bad_variant_access()@GOTPAGE
-Lloh7:
-	ldr	x2, [x2, std::bad_variant_access::~bad_variant_access()@GOTPAGEOFF]
-	bl	___cxa_throw
-	.loh AdrpLdrGot	Lloh6, Lloh7
-	.loh AdrpLdrGot	Lloh4, Lloh5
-	.loh AdrpLdrGot	Lloh2, Lloh3
-	.cfi_endproc
-                                        ; -- End function
-	.globl	std::bad_variant_access::~bad_variant_access()  ; -- Begin function _ZNSt18bad_variant_accessD1Ev
-	.weak_def_can_be_hidden	std::bad_variant_access::~bad_variant_access()
-	.p2align	2
-std::bad_variant_access::~bad_variant_access():         ; @_ZNSt18bad_variant_accessD1Ev
-	.cfi_startproc
-; %bb.0:
-	b	std::exception::~exception()
-	.cfi_endproc
-                                        ; -- End function
 	.p2align	2                               ; -- Begin function _ZNSt3__116__variant_detail12__visitation6__base12__dispatcherIJLm0EEE10__dispatchB6v15006IONS1_9__variant15__value_visitorIN9celeritas12_GLOBAL__N_126VariantTransformDispatcherEEEJRKNS0_6__baseILNS0_6_TraitE0EJNS_9monostateENS8_11TranslationENS8_14TransformationEEEEEEEDcT_DpT0_
 decltype(auto) std::__1::__variant_detail::__visitation::__base::__dispatcher<0ul>::__dispatch[abi:v15006]<std::__1::__variant_detail::__visitation::__variant::__value_visitor<celeritas::(anonymous namespace)::VariantTransformDispatcher>&&, std::__1::__variant_detail::__base<(std::__1::__variant_detail::_Trait)0, std::__1::monostate, celeritas::Translation, celeritas::Transformation> const&>(std::__1::__variant_detail::__visitation::__variant::__value_visitor<celeritas::(anonymous namespace)::VariantTransformDispatcher>&&, std::__1::__variant_detail::__base<(std::__1::__variant_detail::_Trait)0, std::__1::monostate, celeritas::Translation, celeritas::Transformation> const&): ; @_ZNSt3__116__variant_detail12__visitation6__base12__dispatcherIJLm0EEE10__dispatchB6v15006IONS1_9__variant15__value_visitorIN9celeritas12_GLOBAL__N_126VariantTransformDispatcherEEEJRKNS0_6__baseILNS0_6_TraitE0EJNS_9monostateENS8_11TranslationENS8_14TransformationEEEEEEEDcT_DpT0_
 	.cfi_startproc
```

### DIFF
--- a/src/corecel/Assert.cc
+++ b/src/corecel/Assert.cc
@@ -117,6 +117,8 @@ char const* to_cstring(DebugErrorType which)
             return "feature is not yet implemented";
         case DebugErrorType::postcondition:
             return "postcondition failed";
+        case DebugErrorType::assumption:
+            return "assumption failed";
     }
     return "";
 }
--- a/src/corecel/Assert.hh
+++ b/src/corecel/Assert.hh
@@ -50,6 +50,16 @@
  * side effects are as expected when leaving a function.
  */
 /*!
+ * \def CELER_ASSUME
+ *
+ * Always-on compiler assumption. This should be used very rarely: you should
+ * make sure the resulting assembly is simplified in optimize mode from using
+ * the assumption. For example, sometimes informing the compiler of an
+ * assumption can reduce code bloat by skipping standard library exception
+ * handling code (e.g. in \c std::visit by assuming \c
+ * !var_obj.valueless_by_exception() ).
+ */
+/*!
  * \def CELER_VALIDATE
  *
  * Always-on runtime assertion macro. This can check user input and input data
@@ -156,6 +166,14 @@
         CELER_DEBUG_THROW_(MSG, WHICH); \
         ::celeritas::unreachable();     \
     } while (0)
+#define CELER_NDEBUG_ASSUME_(COND)      \
+    do                                  \
+    {                                   \
+        if (CELER_UNLIKELY(!(COND)))    \
+        {                               \
+            ::celeritas::unreachable(); \
+        }                               \
+    } while (0)
 #define CELER_NOASSERT_(COND)   \
     do                          \
     {                           \
@@ -167,12 +185,14 @@
 #    define CELER_EXPECT(COND) CELER_DEBUG_ASSERT_(COND, precondition)
 #    define CELER_ASSERT(COND) CELER_DEBUG_ASSERT_(COND, internal)
 #    define CELER_ENSURE(COND) CELER_DEBUG_ASSERT_(COND, postcondition)
+#    define CELER_ASSUME(COND) CELER_DEBUG_ASSERT_(COND, assumption)
 #    define CELER_ASSERT_UNREACHABLE() \
         CELER_DEBUG_FAIL_("unreachable code point encountered", unreachable)
 #else
 #    define CELER_EXPECT(COND) CELER_NOASSERT_(COND)
 #    define CELER_ASSERT(COND) CELER_NOASSERT_(COND)
 #    define CELER_ENSURE(COND) CELER_NOASSERT_(COND)
+#    define CELER_ASSUME(COND) CELER_NDEBUG_ASSUME_(COND)
 #    define CELER_ASSERT_UNREACHABLE() ::celeritas::unreachable()
 #endif
 
@@ -367,6 +387,7 @@ enum class DebugErrorType
     unconfigured,  //!< Internal assertion: required feature not enabled
     unimplemented,  //!< Internal assertion: not yet implemented
     postcondition,  //!< Postcondition contract violation
+    assumption,  //!< "Assume" violation
 };
 
 enum class RuntimeErrorType
--- a/src/corecel/Macros.hh
+++ b/src/corecel/Macros.hh
@@ -15,6 +15,9 @@
 
 #include "celeritas_config.h"
 
+//---------------------------------------------------------------------------//
+// MACROS
+//---------------------------------------------------------------------------//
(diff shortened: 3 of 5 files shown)


esseivaju (approved):
Nice, LGTM

pcanal:
> by avoiding the jump to exception handling code 

I am a bit confused by the assembly code.  What is left does not seem to have any jump or process quit instructions? 

I.e. If I read the assembly code correctly, the assertion is completely gone (does not fulfill the 'Always-on compiler assumption')  (On the other hand the C/C++ code seem to say it should still be there).

What am I missing?  (Maybe in the example used the compiler can determine that the condition will indeed never be true so it can remove the code?)

sethrj:
> I am a bit confused by the assembly code. What is left does not seem to have any jump or process quit instructions?

Correct. The assumption the compiler is now *allowed* to make is "such and such invalid condition will now result in undefined behavior". So the compiler (in optimized mode with assertions disabled) is *told* that it doesn't need to check for a bad condition or jump to an exception.

> I.e. If I read the assembly code correctly, the assertion is completely gone (does not fulfill the 'Always-on compiler assumption') (On the other hand the C/C++ code seem to say it should still be there).
>
> What am I missing? (Maybe in the example used the compiler can determine that the condition will indeed never be true so it can remove the code?)

Exactly. If you write
```c++
int do_something(int val)
{
    CELER_ASSUME(val >= 0);
    if (val < 0) std::abort();
    return val;
}
```
the `if` statement will be completely removed by the optimizer. And if `CELER_DEBUG == 0`, the function will just `return val`. The cool thing about this for the `variant` case is that it's actually *optimizing out a runtime check in the standard library*.

PR #906, opened 2023-08-25
Add bounding box transformation

sethrj:
To support shape construction, add translate/transform functions for bounding boxes.


src/orange/BoundingBoxUtils.cc
```diff
@@ -0,0 +1,98 @@
...
+    CELER_EXPECT(a);
+    return {tr.transform_up(a.lower()), tr.transform_up(a.upper())};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the bounding box of a transformed box.
+ *
+ * This method has a custom implementation of GEMV for applying exact rotations
+ * to infinities if the matrix representation simply should switches vector
```
elliottbiondo: Switch

src/orange/BoundingBoxUtils.cc
```diff
@@ -0,0 +1,98 @@
...
+    CELER_EXPECT(a);
+    using real_type = BBox::real_type;
+    using Real3 = BBox::Real3;
+
+    constexpr real_type inf = numeric_limits<real_type>::infinity();
+
+    // Specialized GEMV that ignores zeros to correctly handle infinities
+    auto rotate = [&r = tr.rotation()](Real3 const& x) {
+        Real3 result;
+        for (size_type i = 0; i != 3; ++i)
```
elliottbiondo: Use `range` --- unless you are using this syntax because to be consistent to a GEMV implementation written in standard C

src/orange/BoundingBoxUtils.cc
```diff
@@ -0,0 +1,98 @@
...
+ */
+BBox calc_transform(Translation const& tr, BBox const& a)
+{
+    CELER_EXPECT(a);
+    return {tr.transform_up(a.lower()), tr.transform_up(a.upper())};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the bounding box of a transformed box.
```
elliottbiondo: I think you should emphasis this a bit more. This isn't transforming a bbox --- it's transforming a bbox and then finding an *axes-aligned* bbox around the translated bbox. In other words, this is doing something quite different from  other `calc_transform` functions in the code.

elliottbiondo (changes requested):
Looks good @sethrj! Presumably this will be made obsolete once Celeritas ORANGE has the ability to calculate cell bounding boxes? The bounding boxes created via this method are going to be larger than bounding boxes calculated from cells that have already been rotated. These larger bounding boxes will  impact BIH performance.

sethrj:
> Looks good @sethrj! Presumably this will be made obsolete once Celeritas ORANGE has the ability to calculate cell bounding boxes? The bounding boxes created via this method are going to be larger than bounding boxes calculated from cells that have already been rotated. These larger bounding boxes will impact BIH performance.

This functionality is actually necessary to have bounding boxes work at all. When building a shape, we can only calculate the precise bounding box in the local reference frame. This transformation is applied to calculate a bounding box at the universe level. The original bounding box is clipped (#896) first by in the local reference frame, and finally by the transformed surfaces in the global reference frame. The latter helps restore as much of the "tightness" of the bounding box as possible.

When you ask about "bounding boxes calculated from cells that have already been rotated" are you thinking we can generically calculate AABBs of an arbitrary rotated shape? I think that would require a lot of extra work.

elliottbiondo:
Sorry, I was mistakenly thinking the SurfaceSimplifier was handling transformations. Some surfaces are trivial to transform --- I would think there would be an opportunity to handle some cases a preprocessing step. For example if you had a universe comprised of only planes, you could rotate each plane as a  preprocessing step rather than transforming points at runtime. This would still be a lot of work though.

elliottbiondo (approved):

sethrj:
> Sorry, I was mistakenly thinking the SurfaceSimplifier was handling transformations. Some surfaces are trivial to transform --- I would think there would be an opportunity to handle some cases a preprocessing step. For example if you had a universe comprised of only planes, you could rotate each plane as a preprocessing step rather than transforming points at runtime. This would still be a lot of work though.

None of the bounding box transformation is happening at runtime; this is all for preprocessing. The plane rotations (i.e. aligned -> general -> rotated -> aligned) and resultant clipping happens "at the same time" as the bounding box rotation and clipping. So we really do get the best of both worlds.

PR #908, opened 2023-08-25
Use par_nosync execution policy to execute thrust algorithms

esseivaju:
The `par_nosync` execution policy hints thrust to remove stream synchronization points that are not necessary for correctness, making some algorithms such as sort asynchronous.

Thrust fill_n operation used in the `Filler<Device>` could also make use of it but as it's used in the Collection is not stream-aware this should be updated in a subsequent PR.


amandalund:
It looks like this feature may require a newer version of CUDA: `TrackInitAlgorithms.cu(38): error: namespace "thrust::cuda" has no member "par_nosync"` and is not implemented for HIP: `TrackInitAlgorithms.cu:38:17: error: no member named 'cuda' in namespace 'thrust'`?

esseivaju:
It requires [Thrust 1.16](https://nvidia.github.io/thrust/releases/changelog.html#thrust-1160), the CUDA version should be recent enough as I am using 11.8 locally as well. I suppose that the thrust version on the CI machines is too old. ~~It seems like HIP should [support it](https://github.com/ROCmSoftwarePlatform/rocThrust/blob/develop/examples/cuda/explicit_cuda_stream.cu) as well, maybe they'll ignore it and still synchronize but with a recent enough version of rocThrust it should compile.~~ For HIP, we need to change the [namespace](https://github.com/ROCmSoftwarePlatform/rocThrust/blob/d4c74b9d238dc79f588d9b8a02eb22c167701ef0/thrust/system/hip/detail/par.h#L196-L200).

sethrj:
Repeating what we said in the meeting today: CUDA 11.7 is the newest on summit and it only has thrust 1.15 😞

src/celeritas/ThrustConfig.hh
```diff
@@ -0,0 +1,31 @@
...
+//! \brief Platform and version-specific thrust setup
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <thrust/execution_policy.h>
+#include <thrust/version.h>
+
+namespace celeritas
+{
+#if CELERITAS_USE_CUDA
```
sethrj: Missing `"celeritas_config.h"`

src/celeritas/ThrustConfig.hh
```diff
@@ -0,0 +1,31 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/ThrustConfig.hh
```
sethrj: How about `celeritas/ext/ThrustNative.hh`? It's a wrapper around the external thrust library, and its main job (now) is to provide `thrust_native` wrappers?

(Actually I'm thinking `celeritas/ext/Thrust.device.hh` since we'll probably add more Celeritas↔︎Thrust utilities.)

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -34,11 +35,12 @@ size_type remove_if_alive(
...
 {
     auto start = device_pointer_cast(vacancies.data());
-    auto end = thrust::remove_if(
-        thrust::device.on(celeritas::device().stream(stream_id).get()),
-        start,
-        start + vacancies.size(),
-        IsEqual{occupied()});
+    auto end
+        = thrust::remove_if(thrust_execution_policy().on(
+                                celeritas::device().stream(stream_id).get()),
```
sethrj: This is going to be a pretty common idiom, so maybe `celeritas/ext/Thrust.device.hh` should provide a function that maps stream ID to execution space using the global device?

sethrj (changes requested):

src/celeritas/track/detail/TrackInitAlgorithms.cu
esseivaju: I ended up adding two functions to get an execution space from a stream id, with a version check for async execution.

sethrj (approved):
Nice!

sethrj:
@esseivaju Have you got any before-and-after profiler visualization to show how this affects the sort algorithm performance?

esseivaju:
I do but we end up not gaining anything, the average/median time for the NVTX sort range is the same. This is because thrust sort needs to allocate/deallocate device memory. So we have to synchronize the host thread to free that memory regardless. The top image is async and the bottom is sync, the synchronization between the two cudaFree is removed but we still wait for the 2nd free (DeviceVector in sort_impl) to finish before the host thread continues and schedules the next kernel. 

The time it takes for the GPU to execute all the kernels required to sort tracks is very short, on average 100 microsec/step, while the total time to execute the action is much slower, 1 ms/step.

image

image

sethrj:
I guess https://github.com/celeritas-project/celeritas/pull/910 could help with that? Is there a way to provide thrust with the necessary buffer space (or cache it for reuse) so it doesn't have to keep reallocating it?

esseivaju:
yeah NVIDIA/thrust#910 will help with the `DeviceVector` allocation. About thrust, I hope it is possible. We can always use CUB (HIP also has its implementation) where we specify the buffer that we allocate ourselves but this [example](https://github.com/NVIDIA/thrust/blob/main/examples/cuda/custom_temporary_allocation.cu) and the note suggest that it is possible, I'll have to dig a bit as I don't find anything in their documentation..

esseivaju:
I managed to have Thrust call the stream-ordered allocator API, we currently need to use a custom memory resource, Thrust might eventually implement its own (NVIDIA/cccl#768). Stream-ordered allocator uses a memory pool in the backend, however, all its memory gets released by default every time we synchronize with the stream, which is every simulation step in Celeritas. We can configure the pool to retain memory (more info [here](https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-1)) which we'll want to do as well for our `DeviceVector`.

src/celeritas/ext/Thrust.device.hh
```diff
@@ -0,0 +1,107 @@
...
+    {
+        return thrust_native::par.on(
+            celeritas::device().stream(stream_id).get());
+    }
+    else if constexpr (T == ThrustExecMode::Async)
+    {
+        using Alloc = thrust::mr::allocator<char, AsyncMemoryResource<void*>>;
+        Stream const& stream = celeritas::device().stream(stream_id);
+        // FIXME: one memory resource per stream
+        static AsyncMemoryResource async_memory_resource{stream.get()};
```
esseivaju: We need to store one instance per stream but I'm not sure where would be the best place to do that.

src/corecel/sys/Stream.hh
```diff
@@ -56,8 +103,13 @@ class Stream
     // Access the stream
     StreamT get() const { return stream_; }
 
+    // Access the thrust resource allocator associated with the stream
+    ResourceT& get_memory_resource() const { return memory_resource_; }
```
sethrj: In celeritas we avoid `get_` prefix for simple accessors: maybe just `memory_resource`? If the `Stream` is going to encapsulate more than just a CUDA stream (which I think is a good idea) perhaps we should rename `get()` above to `stream()`? I also think we should make this function non-const, remove `mutable` below, and let the device return a mutable stream, if we're going to store temporary device buffers as part of it.

src/corecel/sys/Stream.cc
```diff
@@ -11,15 +11,50 @@
...
+template<class Pointer>
+void AsyncMemoryResource<Pointer>::do_deallocate(pointer p,
+                                                 std::size_t,
+                                                 std::size_t)
+{
+#if !CELER_USE_DEVICE
+    (void)p;
+#endif
+
+    CELER_DEVICE_CALL_PREFIX(FreeAsync(p, stream_));
```
sethrj: Since this is presumably called during a destructor, we should probably wrap this in a try/catch that prints the error, as we do in the DeviceAllocation destructor.

src/corecel/sys/Device.cc
```diff
@@ -192,6 +192,13 @@ Device::Device(int id)
 
     // Save for possible block size initialization
     max_threads_per_block = props.maxThreadsPerBlock;
+
+    // TODO: Configurable threshold
```
sethrj: This would be a good candidate for an environment variable: see e.g. VecGeomParams:
```
        if (std::string var = celeritas::getenv("CUDA_STACK_SIZE");
            !var.empty())
        {
            int stack_size = std::stoi(var);
            CELER_VALIDATE(stack_size > 0,
                           << "invalid CUDA_STACK_SIZE=" << stack_size
                           << " (must be positive)");
            set_cuda_stack_size(stack_size);
        }
```

src/corecel/device_runtime_api.h
```diff
@@ -24,6 +24,10 @@
 #    include <hip/hip_runtime.h>
 #endif
 
+#if CELERITAS_USE_CUDA || CELERITAS_USE_HIP
+#    include <thrust/mr/memory_resource.h>
+#endif
```
sethrj, line 29: aw man, do we really have to propagate Thrust into all our cuda code? Can you measure the build time on develop before and after adding this include?

esseivaju: I don't like that either but a custom memory resource must derive from `thrust::mr::memory_resource ` so if we store it in `Stream` I'm not sure there is a way around it.

esseivaju: ndebug build with CUDA on Zeus (2 Intel Xeon Gold 5220):

53a86ed:

real    3m18.952s
user    22m21.945s
sys     3m6.637s

e7fe013:

real    3m18.567s
user    22m17.268s
sys     3m5.361s

sethrj: OK, that's not enough difference to worry about. Thanks for checking! (I think some of the thrust algorithms end up injecting a bunch of global data... which is probably why the timing can get bad. Presuably a low-level class like this doesn't affect it much.)

src/corecel/sys/Device.hh
```diff
@@ -76,6 +76,15 @@ class Device
...
     explicit Device(int id);
 
+    //// DESTRUCTORS ////
+
+    // Needed for StreamStorage pimpl
+    ~Device();
+
+    //// OPERATORS ////
+
+    Device& operator=(Device&&);
```
sethrj: You're fond of pointing out the rule of 5 😉

The better way to handle this, by the way, is to define a custom deleter for stream storage, declare it in this header, and define it in the Device.cc file. Then the default constructor and move constructors are handled natively by the unique pointer.

src/corecel/sys/Device.hh
```diff
@@ -118,10 +127,10 @@ class Device
     void create_streams(unsigned int num_streams) const;
 
     // Access a stream
-    Stream const& stream(StreamId) const;
+    Stream& stream(StreamId) const;
```
sethrj, line 121: I think it would be reasonable to for now make this non-const, or offer both versions:
```suggestion
    Stream& stream(StreamId);
```
since we virtually always access through `celeritas::device()`, which returns mutable, I think we can propagate the constness correctly.

esseivaju: `celeritas::device()` returns a const ref, so we have to make this const as well.

sethrj: Aha. Sorry, guess I was thinking of the global `set_device`.

src/corecel/sys/Device.hh
esseivaju: we end up having to use new/delete with the custom Deleter though instead of the compiler doing that for us, I'm not convinced I like that more.

sethrj (approved):
LGTM! Ready for merging? @amandalund did you have any feedback?

esseivaju:
yeah, good for me. I'll implement the rest in subsequent PR.

src/corecel/sys/Device.hh
sethrj: Yeah, not being able to use `make_unique` is the downside to this approach.

amandalund (approved):
Nope, looks good to me as well. Thanks @esseivaju!

PR #909, opened 2023-08-25
Accelerate ORANGE boundary crossing with BIH

elliottbiondo:
This PR modifies `SimpleUnitTracker::cross_boundary` to use the bounding interval hierarchy (BIH) when there are more than 2 neighbor cells to search (excluding the current cell). With this PR, bumped bounding boxes become strictly necessary for tracking. Likewise, a `calc_bumped` function has been added to `BoundingBoxUtils` to perform this bump prior to constructing the BIH.

### DIFF
--- a/src/orange/BoundingBoxUtils.hh
+++ b/src/orange/BoundingBoxUtils.hh
@@ -186,7 +186,7 @@ calc_intersection(BoundingBox<T> const& a, BoundingBox<T> const& b)
  * Infinite values are unchanged.
  */
 template<class T, class U>
-inline constexpr BoundingBox<T> calc_bumped(BoundingBox<U> const& bbox)
+inline BoundingBox<T> calc_bumped(BoundingBox<U> const& bbox)
 {
     CELER_EXPECT(bbox);
 
--- a/src/orange/BoundingBoxUtils.hh
+++ b/src/orange/BoundingBoxUtils.hh
@@ -178,4 +178,31 @@ calc_intersection(BoundingBox<T> const& a, BoundingBox<T> const& b)
 }
 
 //---------------------------------------------------------------------------//
+/*!
+ * Convert bbox with U type values to a bumped bbox with T type values.
+ *
+ * Each U lower value is bumped to the greatest T value less than the U value.
+ * Each U upper value is bumped to the lowest T value greater than the U value.
+ * Infinite values are unchanged.
+ */
+template<class T, class U>
+inline constexpr BoundingBox<T> calc_bumped(BoundingBox<U> const& bbox)
+{
+    CELER_EXPECT(bbox);
+
+    Array<T, 3> lower;
+    Array<T, 3> upper;
+
+    for (auto ax : range(to_int(Axis::size_)))
+    {
+        lower[ax] = std::nextafter(static_cast<T>(bbox.lower()[ax]),
+                                   -numeric_limits<T>::infinity());
+        upper[ax] = std::nextafter(static_cast<T>(bbox.upper()[ax]),
+                                   numeric_limits<T>::infinity());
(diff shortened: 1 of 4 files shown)


src/orange/BoundingBoxUtils.hh
```diff
@@ -177,5 +177,32 @@ calc_intersection(BoundingBox<T> const& a, BoundingBox<T> const& b)
...
+inline constexpr BoundingBox<T> calc_bumped(BoundingBox<U> const& bbox)
+{
+    CELER_EXPECT(bbox);
+
+    Array<T, 3> lower;
+    Array<T, 3> upper;
+
+    for (auto ax : range(to_int(Axis::size_)))
+    {
+        lower[ax] = std::nextafter(static_cast<T>(bbox.lower()[ax]),
```
sethrj, line 198: These are only constexpr since C++23 so I don't think this function can be constexpr.

src/orange/detail/UnitInserter.cc
```diff
@@ -191,7 +172,8 @@ SimpleUnitId UnitInserter::operator()(UnitInput const& inp)
         // Store the bbox or an infinite bbox placeholder
         if (inp.volumes[i].bbox)
         {
-            bboxes.push_back(make_fast_bbox(inp.volumes[i].bbox));
+            bboxes.push_back(
+                calc_bumped<fast_real_type, real_type>(inp.volumes[i].bbox));
```
sethrj: The way we set up the function signature, the second template argument can be inferred:
```suggestion
            bboxes.push_back(calc_bumped<fast_real_type>(inp.volumes[i].bbox));
```

test/orange/BoundingBoxUtils.test.cc
```diff
@@ -173,6 +175,42 @@ TEST_F(BoundingBoxUtilsTest, bbox_intersection)
     }
 }
 
+TEST_F(BoundingBoxUtilsTest, bumped)
+{
```
sethrj, line 177: I'd change all the code here to be specifically float/double rather than real/fast_real.

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -205,29 +205,46 @@ SimpleUnitTracker::cross_boundary(LocalState const& state) const
...
-        // Calculate the local senses and face
+        VolumeView vol = this->make_local_volume(id);
         auto logic_state
             = calc_senses(vol, detail::find_face(vol, state.surface));
 
-        // Evaluate whether the senses are "inside" the volume
-        if (!detail::LogicEvaluator(vol.logic())(logic_state.senses))
+        temp_vv = &vol;
+        temp_face = logic_state.face;
+        return detail::LogicEvaluator(vol.logic())(logic_state.senses);
```
sethrj: You're saving a pointer to a temporary here 😱

Instead, just make a local `OnLocalSurface on_surface`, and 
```suggestion
        if (detail::LogicEvaluator(vol.logic())(logic_state.senses))
        {
            // Inside: find and save the local surface ID, and end the search
            on_surface = surface(vol, logic_state.face);
            return true;
        }
        return false;
```

test/orange/BoundingBoxUtils.test.cc
```diff
@@ -1,3 +1,5 @@
+
+
```
sethrj:
```suggestion
```
(two lines accidentally inserted)

sethrj (changes requested):

elliottbiondo:
@sethrj pending CI this should be ready

sethrj (approved):
Looks great! Easy, clean, and hopefully much faster for pseudoarrays.

PR #910, opened 2023-08-26
Add pinned allocator and asynchronous memory operations

esseivaju:
CUDA and HIP both provide stream-ordered memory allocation API, it synchronizes allocation and deallocation with a stream instead of the entire device. In addition, it uses a memory pool and doesn't necessarily release the memory to the system, making it less expensive to call malloc/free ([see](https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-1/) for more details). This should improve performances when using multiple streams as we no longer need to synchronize the entire device and when temporary device storage is needed (e.g. sorting tracks)

This PR's goal is to add support for asynchronous memory operations. This is a follow-up to #908 which added support for asynchronous thrust. The following features need to be added:

- [x] Add support for pinned host memory allocation
- [x] Stream-aware `Copier`
- [ ] Stream-aware `DeviceAllocation` (i.e. Collection)
- [x] Stream-aware `Filler`


test/CMakeLists.txt
```diff
@@ -559,7 +559,7 @@ if(CELERITAS_CORE_GEO STREQUAL "Geant4")
   set(_fails_g4geo DISABLE)
 endif()
 
-celeritas_add_test(celeritas/user/DetectorSteps.test.cc GPU)
+celeritas_add_test(celeritas/user/DetectorSteps.test.cc LINK_LIBRARIES CUDA::cudart GPU)
```
esseivaju: Not sure what's the best way to handle the dependency to cudart since the allocator uses cuda to alloc host memory

src/corecel/data/detail/PinnedAllocator.hh
```diff
@@ -0,0 +1,49 @@
...
+#include "corecel/Assert.hh"
+#include "corecel/Macros.hh"
+#include "corecel/device_runtime_api.h"
+
+namespace celeritas
+{
+namespace detail
+{
+template<class T>
+struct PinnedAllocator
```
esseivaju: @sethrj Pimpl vs explicit instantiation to hide the CUDA/HIP dependency? I think I like the latter more, the only downside is that we'd have to specify instantiation for quite a few types.

esseivaju:
@sethrj @amandalund Still missing the update to `DeviceAllocation` but this is otherwise ready for review. Maybe implement that in another PR as it'll require quite a few changes?

src/celeritas/global/detail/PinnedAllocator.hh
```diff
@@ -0,0 +1,40 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/global/detail/PinnedAllocator.hh
```
sethrj: This probably would fit better in corecel/data, and we can have a separate template instantiation file in celeritas for the classes that depend on downstream types

src/celeritas/global/detail/PinnedAllocator.cc
```diff
@@ -0,0 +1,87 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/global/detail/PinnedAllocator.cc
```
sethrj, line 6: Let's move this file to `corecel/data/PinnedAllocator.t.hh` (out of the detail namespace since other classes will need to build vectors using this type), separate the `corecel` template instantiations into `.../PinnedAllocator.cc`, and move the `celeritas` instantiations to `celeritas/global/detail/PinnedAllocator.cc`: we do similar for `Filler`.

src/celeritas/global/detail/PinnedAllocator.hh
```diff
@@ -0,0 +1,40 @@
...
+#include <cstdlib>
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Custom allocator for pinned host memory when Celeritas is built with device
+ * support.
```
sethrj: Mention that this class satisfies the requirements for the std Allocator concept.

src/celeritas/global/detail/PinnedAllocator.hh
```diff
@@ -0,0 +1,40 @@
...
+ */
+template<class T>
+struct PinnedAllocator
+{
+    using value_type = T;
+
+    [[nodiscard]] T* allocate(std::size_t);
+
+    void deallocate(T*, std::size_t) noexcept;
+};
```
sethrj: I think you need to add free functions `operator==` (always true) and `operator!=` (always false) for this to fully satisfy the Allocator requirements: most of https://en.cppreference.com/w/cpp/named_req/Allocator is optional but equality looks required (maybe called through e.g. `std::vector::operator==` ?).

src/corecel/data/Copier.cc
```diff
@@ -11,18 +11,20 @@
...
-                std::size_t count)
+namespace
+{
+template<bool async>
+inline void copy_bytes_impl(MemSpace dstmem,
+                            void* dst,
+                            MemSpace srcmem,
+                            void const* src,
+                            std::size_t count,
+                            [[maybe_unused]] StreamId stream)
```
sethrj: This will probably need another GCC 8-9 workaround... rather than using the `template<bool async>` I think it'd be better to have a standalone function `decltype(auto) to_memcpy_kind(MemSpace src, MemSpace dst)` that you reuse in the separate sync/async functions.

src/corecel/data/Copier.hh
```diff
@@ -39,6 +40,8 @@ struct Copier
     static constexpr auto dstmem = M;
 
     inline void operator()(MemSpace srcmem, Span<T const> src) const;
+    inline void
+    operator()(MemSpace srcmem, Span<T const> src, StreamId stream) const;
```
sethrj: Since the stream will presumably be the same for all the copy operations made by a single Copier instance, it should go in the constructor rather than call operator.

src/corecel/data/detail/Filler.hh
```diff
@@ -42,6 +43,7 @@ struct Filler<T, MemSpace::device>
     T const& value;
 
     void operator()(Span<T>) const;
+    void operator()(Span<T>, StreamId) const;
```
sethrj: Same here: stream is probably a better fit for the constructor.

src/celeritas/global/detail/PinnedAllocator.cc
```diff
@@ -0,0 +1,87 @@
...
+template<class T>
+void PinnedAllocator<T>::deallocate(T* p, std::size_t) noexcept
+{
+#if CELER_USE_DEVICE
+    try
+    {
+#    if CELERITAS_USE_CUDA
+        CELER_DEVICE_CALL_PREFIX(FreeHost(p));
+#    elif CELERITAS_USE_HIP
+        CELER_DEVICE_CALL_PREFIX(HostFree(p));
```
sethrj: No point in using the "device prefix" when the whole functions are different....
```suggestion
        CELER_CUDA_CALL(cudaFreeHost(p));
#    elif CELERITAS_USE_HIP
        CELER_HIP_CALL(hipHostFree(p));
```

sethrj (commented):
Nice! Much cleaner than my idea of using a separate MemSpace. Looks great aside from the detail of where to define and instantiate it.

src/corecel/data/Copier.hh
esseivaju: Yeah, wasn't sure I ended up going for specifying it in the call operator because in the constructor we have to check at runtime which version to use and store an extra id. Pretty minimal overhead but nothing would technically prevent to reuse the same Copier on different streams.

sethrj: Yeah, it makes sense when you put it that way. But presumably since the destination is the same, if you're pinning it then it's async and on the same stream; if it's not, then you're doing synchronized anyway.

src/corecel/data/Copier.cc
```diff
@@ -11,9 +11,32 @@
...
+{
+    if (src == MemSpace::host && dst == MemSpace::device)
+        return CELER_DEVICE_PREFIX(MemcpyHostToDevice);
+    else if (src == MemSpace::device && dst == MemSpace::host)
+        return CELER_DEVICE_PREFIX(MemcpyDeviceToHost);
+    else if (src == MemSpace::device && dst == MemSpace::device)
+        return CELER_DEVICE_PREFIX(MemcpyDeviceToDevice);
+    else
+        CELER_ASSERT_UNREACHABLE();
+    return CELER_DEVICE_PREFIX(MemcpyDefault);
```
sethrj: Surely both of these aren't needed 😅 I think I originally had the default there just to avoid a potentially uninitialized variable. Maybe just
```suggestion
    CELER_ASSERT_UNREACHABLE();
```

esseivaju: yeah 😅 I remember having issues with unreachable and inline but it seems fine here.

sethrj (approved):

PR #911, opened 2023-08-26
Define variant helpers and refactor surface type dispatch

sethrj:
This refactors nasty macros as an elegant visitor. The runtime surface dispatch may be further transformed with template metaprogramming if we want.

The key here is that we can define a functor `[](auto surf_traits) { ... };` where `surf_traits` is an *instance* of the `SurfaceTypeTraits` class, which we can query using `decltype` to obtain the underlying surface type. An implicit `operator SurfaceType()` allows `surf_traits` to be implicitly cast to a constexpr `SurfaceType` enumeration so it can also be used in templates like `MySurfaceMagic<surf_traits>{}`...

I've also moved some of the common variant utilities (e.g. `Overload`) and abstracted `EnumVariant` so that it can be used in type-deleted transformations.

Also cool: with LLVM and debug assertions disabled, the following function *completely* optimizes out:
```c++
using SInt = std::underlying_type_t<SurfaceType>;

SInt surface_type_to_int(SurfaceType st)
{
    return visit_surface_type(
        [](auto traits) { return static_cast<SInt>(traits()); }, st);
}
```
showing that the switch and function dispatch can have zero overhead.

This replaces the awkward `make_surface_action` with a helper class that acts as `std::visit`.


sethrj:
Ping @elliottbiondo

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -0,0 +1,178 @@
...
+class LocalSurfaceVisitor
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using ParamsRef = NativeCRef<OrangeParamsData>;
+    //!@}
+
+  public:
+    // Construct from ORANGE params and simple unit record
```
elliottbiondo: and surfaces record

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -0,0 +1,178 @@
...
+    template<class T>
+    inline CELER_FUNCTION T make_surface(LocalSurfaceId data_offset) const;
+
+    // TODO: change surfaces record to use ItemMap, move this to Collection.hh
+    template<class T, class U>
+    static inline CELER_FUNCTION T get_item(Items<T> const& items,
+                                            ItemRange<T> const& range,
+                                            ItemId<U> item);
+
+    template<class T>
```
elliottbiondo, line 79: Add doc string

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -0,0 +1,178 @@
...
+    static inline CELER_FUNCTION T const*
+    get_ptr(Items<T> const& items, ItemId<T> item);
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from ORANGE data (inside simple unit tracker).
```
elliottbiondo: and surfaces record

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -0,0 +1,178 @@
...
+    constexpr size_type size{T::StorageSpan::extent};
+    CELER_ASSERT(offset + size <= params_.reals.size());
+
+    real_type const* data = this->get_ptr(params_.reals, offset);
+    return T{Span<real_type const, size>{data, size}};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a pointer to the item in the given range.
```
elliottbiondo: Get an item in the given range

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -0,0 +1,178 @@
...
+    CELER_EXPECT(*range.end() <= items.size());
+    CELER_EXPECT(item < range.size());
+
+    T const* ptr = LocalSurfaceVisitor::get_ptr(items, *range.begin());
+    return *(ptr + item.unchecked_get());
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a pointer to the item in the given range.
```
elliottbiondo: given collection

elliottbiondo (approved):

PR #912, opened 2023-08-28
Fix multi-level ORANGE direction change

tmdelellis:
This fixes an issue in `orange` in which the direction is not updated at all levels *above* the current track position.  It also hardcodes the maximum level from 3->4 to support common reactor geometry types.  This is a temporary fix until we can dynamically determine the maximum level during setup.


sethrj (changes requested):
This indicates a deficiency in the ORANGE testing; could you please add testing to `set_dir` in one of the `UniversesTest` blocks?

tmdelellis:
Test added

sethrj (approved):
Thanks!

PR #913, opened 2023-08-28
Hardcode `PrimaryGenerator` RNG engine and inherit from `EventReaderInterface`

amandalund:
Checks off one TODO.


sethrj (approved):
Nice refactor! Thanks.

PR #915, opened 2023-08-30
Automatically calculate ORANGE universe depth

elliottbiondo:
This PR recursively loops through the universes within a geometry to determine the maximum depth, where a single-universe geometry has a maximum depth of 1. In addition, the variable name `max_level` has been changed to `max_num_levels`, to avoid confusion with 0 vs 1 indexing.

### DIFF
--- a/src/orange/OrangeData.hh
+++ b/src/orange/OrangeData.hh
@@ -30,7 +30,7 @@ namespace celeritas
  */
 struct OrangeParamsScalars
 {
-    size_type max_level{};
+    size_type max_num_levels{};
     size_type max_faces{};
     size_type max_intersections{};
     size_type max_logic_depth{};
@@ -42,7 +42,7 @@ struct OrangeParamsScalars
     //! True if assigned
     explicit CELER_FUNCTION operator bool() const
     {
-        return max_level > 0 && max_faces > 0 && max_intersections > 0
+        return max_num_levels > 0 && max_faces > 0 && max_intersections > 0
                && bump_rel > 0 && bump_abs > 0;
     }
 };
@@ -390,21 +390,21 @@ struct OrangeStateData
     StateItems<LevelId> level;
     StateItems<LevelId> surface_level;
 
-    // Dimensions {num_tracks, max_level}
+    // Dimensions {num_tracks, max_num_levels}
     Items<Real3> pos;
     Items<Real3> dir;
     Items<LocalVolumeId> vol;
     Items<UniverseId> universe;
 
-    // Surface crossing, dimensions {num_tracks, max_level}
+    // Surface crossing, dimensions {num_tracks, max_num_levels}
     Items<LocalSurfaceId> surf;
     Items<Sense> sense;
     Items<BoundaryResult> boundary;
 
     // TODO: this is problem-dependent data and should eventually be removed
-    // max_level defines the stride into the preceding pseudo-2D Collections
-    // (pos, dir, ..., etc.)
-    size_type max_level{0};
+    // max_num_levels defines the stride into the preceding pseudo-2D
+    // Collections (pos, dir, ..., etc.)
+    size_type max_num_levels{0};
 
     // Scratch space
     Items<Sense> temp_sense;  // [track][max_faces]
@@ -427,7 +427,7 @@ struct OrangeStateData
             && surf.size() == pos.size()
             && sense.size() == pos.size()
             && boundary.size() == pos.size()
-            && max_level > 0
+            && max_num_levels > 0
             && !temp_sense.empty()
             && !temp_face.empty()
             && temp_distance.size() == temp_face.size()
@@ -452,7 +452,7 @@ struct OrangeStateData
         surf = other.surf;
         sense = other.sense;
         boundary = other.boundary;
-        max_level = other.max_level;
+        max_num_levels = other.max_num_levels;
 
         temp_sense = other.temp_sense;
         temp_face = other.temp_face;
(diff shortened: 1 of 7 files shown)


src/orange/OrangeData.hh
```diff
@@ -30,7 +30,7 @@ namespace celeritas
  */
 struct OrangeParamsScalars
 {
-    size_type max_level{};
+    size_type max_num_levels{};
```
sethrj: I'd add a comment about this (e.g., it's 1 for non-nested geometry, it's the depth of the universe tree DAG). It's also equivalent to the `max_depth` of the vecgeom implementation: maybe `max_depth` is a better name here? Especially because we have `max_{blah}` below this, so `max_num` is inconsistent.

src/orange/OrangeParams.cc
```diff
@@ -95,6 +95,65 @@ OrangeInput input_from_json(std::string filename)
...
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the maximum num of universe levels within the geometry
+ *
+ * A single-universe geometry will have a max_num_levels of 1.
+ * This is a recursive function, so for external callers, uid should be
+ * the root universe id.
+ */
+size_type calc_max_num_levels(HostVal<OrangeParamsData> const& data,
+                              UniverseId const& uid)
```
sethrj: I think this calculation would be easier to do on the input data, wouldn't it? Or is your thought that this can be reused even if we're building the shapes locally? Because it's using a lot of different low-level data structures in `OrangeParams`, it's going to have a relatively high maintenance burden compared to the simpler input abstractions.

src/orange/OrangeParams.cc
```diff
@@ -95,6 +95,65 @@ OrangeInput input_from_json(std::string filename)
...
+        for (auto vol_id : range(simple_unit_record.volumes.size()))
+        {
+            auto vol_record
+                = data.volume_records[simple_unit_record
+                                          .volumes[LocalVolumeId{vol_id}]];
+            if (vol_record.daughter_id)
+            {
+                max_sub_depth = std::max(
+                    max_sub_depth,
+                    calc_max_num_levels(
```
sethrj: This is probably fine for now, but in the future we'll want to keep a `map<UniverseId, size_type>` of the visited daughters so that we don't have to visit 1M pin cells: this would be refactored as
```
struct MaxNumLevelCalculator
{
    HostVal<OrangeParamsData> const& data;
    std::map<UniverseId, int> num_levels;
    int operator()(UniverseId);
};

File: test/orange/Orange.test.cc [hunk 4]
```diff
@@ -532,6 +532,7 @@ TEST_F(UniversesTest, params)
     OrangeParams const& geo = this->params();
     EXPECT_EQ(12, geo.num_volumes());
     EXPECT_EQ(25, geo.num_surfaces());
+    EXPECT_EQ(3, geo.host_ref().scalars.max_num_levels);
```
sethrj: Add a `max_depth` accessor to the host params, analogously to VecGeom.

src/orange/OrangeData.hh
elliottbiondo: I'll change to max_depth for consistency, but I do think `max_num_` more clearly conveys that something is indexed from 1.

src/orange/OrangeParams.cc
elliottbiondo: I figured it would better to operate on the final, already-built geometry since there may be different avenues for which the geometry is constructed. I see what you are saying about the data structures, but it doesn't seem to me that these would be any more likely to change than the ones on the input side.

src/orange/OrangeParams.cc
elliottbiondo: Ah that's a good point --- easy refactor

elliottbiondo:

src/orange/OrangeParams.cc
```diff
@@ -95,6 +95,82 @@ OrangeInput input_from_json(std::string filename)
...
+    // Calculate max depth
+    size_type operator()(UniverseId const& uid)
+    {
+        CELER_EXPECT(uid);
+
+        // Check for cached value
+        if (depths.find(uid) != depths.end())
+        {
+            return depths[uid];
+        }
```
sethrj:
```suggestion
Best-practices note: doing `find` + `[]` is repeating the same binary lookup; best practice is to do:
```c++
        auto&& [iter, inserted] = depths.insert({uid, {}});
        if (!inserted)
        {
            // Return cached value
            return iter->second;
        }
```

and then at the end of the function, set `iter->second` to the value you calculated. This way you only have one lookup per function instead of two.

And if you're not caching the value, you still shouldn't use `[]` after `find`: just save `if (auto iter = mymap.find(foo); iter != mymap.end()) { return iter->second; }`

src/orange/OrangeParams.cc
```diff
@@ -230,7 +306,8 @@ OrangeParams::OrangeParams(OrangeInput input)
     bbox_ = std::get<UnitInput>(input.universes.front()).bbox;
 
     // Update scalars *after* loading all units
-    host_data.scalars.max_level = input.max_level;
+    MaxDepthCalculator mdc(host_data);
+    host_data.scalars.max_depth = mdc(UniverseId{0});
```
sethrj:
```suggestion
    MaxDepthCalculator calc_depth(host_data);
    host_data.scalars.max_depth = calc_depth(UniverseId{0});
```
or 
```c++
host_data.scalars.max_depth = MaxDepthCalculator{host_data}(UniverseId{0});
```

sethrj (approved):
One and a half more comments.

elliottbiondo:
Pending CI this is ready

sethrj (approved):

sethrj:
Test failures are spurious.

PR #916, opened 2023-08-30
Fix unit conversion errors when reading from event file

amandalund:
I had been seeing some large discrepancies between the track step distributions from celer-g4 and celer-sim for some problems, which I tracked down to an issue with the HepMC3 unit conversion. Calling `GenEvent::set_units()` converts the length units for all particles and the momentum units for all vertices, _except_ for the root vertex. This means that particles without a vertex, which are assigned to the root vertex, do not have correctly converted position and time. Furthermore, vertices that don't have a position set will inherit their position from ancestor vertices or fall back on the event position. We might have caught this in the unit tests, but those expected values were also incorrect. This works around the problem by manually converting the units instead of using `GenEvent::set_units()`.


sethrj (approved):
Good lord. Nice catch @amandalund , sorry for not checking the gold values values closely enough when I input them!

amandalund:
I should have reviewed more carefully! ;)

sethrj:
CI failure is spurious.

PR #917, opened 2023-08-31
Implement generic transforms in multi-universe ORANGE

sethrj:
This adds under-the-hood support for transforms in Celeritas. I have:
- added a GPU-friendly "no transform" special case to replace the `std::monostate` in the existing `TransformVariant`
- implemented a `TransformVisitor` that will apply a functor based on a given transform ID
- added some helper functions to CollectionBuilder
- replaced TranslationId with TransformId
- added a `TransformInserter` that converts `std::variant` to a `TransformRecord` and adds it to orange params (currently the only signature used is to transform a `Real3` to `Translation` to keep the orange input interface the same to ease compatibility issues with SCALE)
- replaced all the explicit transforms in OrangeTrackView with a call to the visitor

@elliottbiondo : In the final commit I went the extra mile and implemented what I think should be full parent-to-daughter transformations (including rotating the direction up and down as needed). I can back that out and add it to a follow-up PR that adds full support for transformations in the SCALE front end, but I don't want to do that until the current queue of ORANGE SCALE MRs is empty.

### DIFF
--- a/src/corecel/data/CollectionBuilder.hh
+++ b/src/corecel/data/CollectionBuilder.hh
@@ -41,7 +41,7 @@ namespace celeritas
  * assertions enabled on host memory, it will assign garbage values to aid in
  * reproducible debugging.)
  */
-template<class T, MemSpace M, class I>
+template<class T, MemSpace M = MemSpace::host, class I = ItemId<T>>
 class CollectionBuilder
 {
   public:
@@ -56,13 +56,16 @@ class CollectionBuilder
 
   public:
     //! Construct from a collection
-    explicit CollectionBuilder(CollectionT* collection) : col_(*collection) {}
+    explicit CollectionBuilder(CollectionT* collection) : col_(*collection)
+    {
+        CELER_EXPECT(collection);
+    }
 
     // Increase size to this capacity
-    inline void resize(size_type count);
+    inline void resize(std::size_t count);
 
     // Reserve space
-    inline void reserve(size_type count);
+    inline void reserve(std::size_t count);
 
     // Extend with a series of elements, returning the range inserted
     template<class InputIterator>
@@ -78,6 +81,9 @@ class CollectionBuilder
     //! Number of elements in the collection
     size_type size() const { return col_.size(); }
 
+    //! Get the size as an ID type
+    ItemIdT size_id() const { return ItemIdT{size()}; }
+
   private:
     CollectionT& col_;
 
@@ -85,12 +91,20 @@ class CollectionBuilder
     StorageT& storage() { return col_.storage(); }
     StorageT const& storage() const { return col_.storage(); }
 
-    //! Maximum elements in a Collection, in underlying storage size
-    static constexpr size_type max_size()
+    //! Maximum elements in a Collection
+    static constexpr std::size_t max_size()
     {
         return std::numeric_limits<size_type>::max();
     }
 };
+
+//---------------------------------------------------------------------------//
+// DEDUCTION GUIDES
+//---------------------------------------------------------------------------//
+
+template<class T, MemSpace M, class I>
+CollectionBuilder(Collection<T, Ownership::value, M, I>*)
+    -> CollectionBuilder<T, M, I>;
 
 //---------------------------------------------------------------------------//
 // INLINE DEFINITIONS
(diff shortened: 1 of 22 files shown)


src/corecel/data/CollectionBuilder.hh
```diff
@@ -56,13 +56,16 @@ class CollectionBuilder
...
     //! Construct from a collection
-    explicit CollectionBuilder(CollectionT* collection) : col_(*collection) {}
+    explicit CollectionBuilder(CollectionT* collection) : col_(*collection)
+    {
+        CELER_EXPECT(collection);
+    }
 
     // Increase size to this capacity
-    inline void resize(size_type count);
+    inline void resize(std::size_t count);
```
elliottbiondo, line 65: What's the reason for changing all these?

src/orange/OrangeTrackView.hh
```diff
@@ -237,11 +246,17 @@ OrangeTrackView::operator=(Initializer_t const& init)
     local.surface = {};
     local.temp_sense = this->make_temp_sense();
 
+    // Helpers for applying mother-to-daughter transformations
```
elliottbiondo: parent-to-daughter for consistency

src/orange/detail/TransformInserter.hh
```diff
@@ -0,0 +1,124 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using Items = Collection<T, Ownership::value, MemSpace::host>;
+    //!@}
+
+  public:
+    // Construct with defaults
```
elliottbiondo: Construct with pointers to params data

src/orange/detail/TransformInserter.hh
```diff
@@ -0,0 +1,124 @@
...
+    TransformId null_transform_;
+    CollectionBuilder<TransformRecord> transforms_;
+    CollectionBuilder<real_type> reals_;
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with references to params data.
```
elliottbiondo: pointers

src/orange/detail/TransformInserter.hh
```diff
@@ -0,0 +1,124 @@
...
+TransformInserter::TransformInserter(Items<TransformRecord>* transforms,
+                                     Items<real_type>* reals)
+    : transforms_{transforms}, reals_{reals}
+{
+    CELER_EXPECT(transforms && reals);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from a translation
```
elliottbiondo: Missing period

src/orange/detail/TransformInserter.hh
```diff
@@ -0,0 +1,124 @@
...
+ */
+TransformId TransformInserter::operator()(VariantTransform const& tr)
+{
+    CELER_ASSUME(!tr.valueless_by_exception());
+    return std::visit(*this, tr);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from a type.
```
elliottbiondo: Construct from a transform with a known type

elliottbiondo (approved):

src/corecel/data/CollectionBuilder.hh
sethrj: Basically, `size_type` is usually less than `std::size_t`, so this is designed to catch the case where we accidentally allocate more than ~4B items. (Or if we're using a "tiny" opaque ID value type, it'll catch that case.)

src/orange/transform/TransformVisitor.hh
```diff
@@ -0,0 +1,133 @@
...
+/*!
+ * Apply a functor to a type-deleted transform.
+ *
+ * An instance of this class is like \c std::visit but accepting a TransformId
+ * rather than a \c std::variant .
+ *
+ * Example: \code
+ TransformVisitor visit_transform{params_};
+ auto new_pos = visit_transform(
+    [&pos](auto&& T) { return t.transform_up(pos); },
```
elliottbiondo: lowercase T

PR #918, opened 2023-08-31
Bikeshedding changes to BIH

sethrj:
While modifying the unit inserter and moving `Overload` and looking in more detail at the associated code, I tweaked a few things that I think improve code clarity. Please take a look and if you agree with the changes we can merge them.

### DIFF
--- a/src/orange/detail/BIHBuilder.cc
+++ b/src/orange/detail/BIHBuilder.cc
@@ -30,8 +30,9 @@ BIHBuilder::BIHBuilder(Storage* storage) : storage_(storage)
  */
 BIHTree BIHBuilder::operator()(VecBBox bboxes)
 {
-    // Store bounding boxes and their corresponding centers
     CELER_EXPECT(!bboxes.empty());
+
+    // Store bounding boxes and their corresponding centers
     bboxes_ = std::move(bboxes);
     centers_.resize(bboxes_.size());
     std::transform(bboxes_.begin(),
@@ -39,9 +40,9 @@ BIHTree BIHBuilder::operator()(VecBBox bboxes)
                    centers_.begin(),
                    &celeritas::calc_center<fast_real_type>);
 
+    // Partition bounding boxes into infinite and finite
     VecIndices indices;
     VecIndices inf_volids;
-
     for (auto i : range(bboxes_.size()))
     {
         LocalVolumeId id(i);
@@ -82,9 +83,8 @@ BIHTree BIHBuilder::operator()(VecBBox bboxes)
     else
     {
         // Degenerate case where all bounding boxes are infinite. Create a
-        // single empty leaf node, so that the existance of leaf nodes does not
+        // single empty leaf node, so that the existence of leaf nodes does not
         // need to be checked at runtime.
-
         VecLeafNodes leaf_nodes{1};
         tree.leaf_nodes
             = make_builder(&storage_->leaf_nodes)
@@ -125,7 +125,6 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
             = p.bboxes[Edge::right].lower()[ax];
 
         // Recursively construct the left and right branches
-
         for (auto edge : range(Edge::size_))
         {
             node.bounding_planes[edge].child = BIHNodeId(nodes->size());
@@ -150,7 +149,7 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
 
 //---------------------------------------------------------------------------//
 /*!
- * Seperate nodes into inner and leaf vectors and renumber accordingly.
+ * Partition nodes into inner and leaf vectors and renumber accordingly.
  */
 BIHBuilder::ArrangedNodes BIHBuilder::arrange_nodes(VecNodes nodes) const
 {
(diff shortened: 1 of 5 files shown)


src/orange/detail/BIHBuilder.cc
```diff
@@ -150,64 +149,67 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
...
     {
-        for (auto edge : range(BIHInnerNode::Edge::size_))
+        for (auto& bp : inner_node.bounding_planes)
         {
-            inner_node.bounding_planes[edge].child
-                = new_ids[inner_node.bounding_planes[edge].child.unchecked_get()];
+            bp.child = remapped_id(bp.child);
         }
-
-        // Handle root node
```
sethrj, line 198: It wasn't at all clear to me that the comment was describing the `if (false)` case: the remapping needed to be done in all cases *except* fro the root node. Putting the comment inside the `if` like `// Not the root node: update its parent ID`

src/orange/detail/BIHBuilder.cc
```diff
@@ -30,18 +30,19 @@ BIHBuilder::BIHBuilder(Storage* storage) : storage_(storage)
  */
 BIHTree BIHBuilder::operator()(VecBBox bboxes)
 {
-    // Store bounding boxes and their corresponding centers
     CELER_EXPECT(!bboxes.empty());
+
+    // Store bounding boxes and their corresponding centers
```
sethrj, line 35: Preconditions go before comments about the subsequent code

pcanal (approved):
No comment.   Don't look so shocked, it happens ... humm ... right one comment then ... Is the `BIHTraverser` constructor is trivial?

sethrj:
😆 No, the constructor isn't trivial: some nontrivial members are initialized, and a pointer to the given storage is saved:
```c++
    VecBBox bboxes_;
    VecReal3 centers_;
    Storage* storage_;
```

pcanal:
> No, the constructor isn't trivial:  ....

Fair enough.  It looks simple enough to not be worth the caching.

src/orange/detail/BIHBuilder.cc
```diff
@@ -150,64 +149,67 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
 
 //---------------------------------------------------------------------------//
 /*!
- * Seperate nodes into inner and leaf vectors and renumber accordingly.
+ * Partition nodes into inner and leaf vectors and renumber accordingly.
```
elliottbiondo: I intentionally did not use the word partition here, because that is not what is going on. Partitioning refers specifically to splitting a group of bounding boxes into two subgroups based on partition criteria. Here, we have a vector of nodes variants -- inner nodes and leaf nodes --- that we are separating into two vectors based only on type.

src/orange/detail/BIHBuilder.cc
```diff
@@ -150,64 +149,67 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
...
-                       inner_nodes.push_back(node);
-                       is_leaf.push_back(false);
-                   },
-                   [&](BIHLeafNode const& node) {
-                       new_ids.push_back(BIHNodeId(leaf_nodes.size()));
-                       leaf_nodes.push_back(node);
-                       is_leaf.push_back(true);
-                   }};
-
+    std::vector<std::size_t> new_idx;
```
elliottbiondo: `idx` is singular, whereas `ids` is plural and makes more sense for a vector. `ids` is used in many places in the code

elliottbiondo (changes requested):

src/orange/detail/BIHBuilder.cc
sethrj: The criteria is the type, isn't it? Or do you mean you're avoiding the term "partitioning" (a la [std::partition](https://en.cppreference.com/w/cpp/algorithm/partition)) because it's overloaded with the specific *spatial partitioning* done in BIH?

elliottbiondo: Yes, one of the principle operations performed in this class is spatial partitioning, so it is best not to reuse the word "partition" to mean something totally different.

sethrj: I don't think it's totally different at all. In one case, you're partitioning bounding box objects based on some criterion. In this case, we're partitioning node objects based on a different criterion applicable to "node objects". It's exactly the same generic partition "algorithm", so I'd argue it's more confusing to introduce a separate word. Here we've specified that the partitioning applies to nodes and the criterion is whether it's a leaf or internal; the `BIHPartitioner` specifies the object as bounding boxes and the criteria is the surface area heuristic.

elliottbiondo: Clarity is subjective, and we can agree to disagree, but FWIW:

Let's say in some MC code, you count the number of cross sections read from disk and print it out for diagnostic purposes. Calling this a "cross section tally" would be a grammatically appropriate use of the word "tally". However, "tally" has idiomatic significance within MC codes, so I would argue "cross section count" would be a less confusing choice.

That's what I am saying here: not that it isn't the same fundamental operation in each case, but rather that within a part of code pertaining to tree construction, for a reader who is not familiar with BIH construction (which will be most people, as BIH not as prevalent as BVH or k-d trees) I think seeing the word "partition" is going to make them think spatial partitioning is going on, when it's not.

elliottbiondo (approved):

PR #276, opened 2021-09-17
Explicitly declare Thread to avoid Geant4 conflict

sethrj:
This is an attempted workaround for a conflict with `PTL::Thread` when including both Celeritas and Geant4 (v11b) that looks like it results from an errant `using` bringing that typedef into the global scope.

```
from acceleritas/app/g4hpc/src/UserTaskManager.cc:10:
acceleritas/external/celeritas/src/base/OpaqueId.hh:121:34: error: using typedef-name ‘PTL::Thread’ after ‘struct’
 using ThreadId = OpaqueId<struct Thread>;
                                  ^~~~~~
In file included from geant4.11.beta/source/externals/ptl/include/PTL/AutoLock.hh:246,
                 from geant4.11.beta/source/externals/ptl/include/PTL/VTask.hh:32,
                 from geant4.11.beta/source/externals/ptl/include/PTL/Task.hh:34,
                 from geant4.11.beta/source/externals/ptl/include/PTL/TaskGroup.hh:35,
                 from acceleritas/app/g4hpc/include/UserTaskManager.hh:10,
                 from acceleritas/app/g4hpc/src/UserTaskManager.cc:8:
geant4.11.beta/source/externals/ptl/include/PTL/Threading.hh:128:41: note: ‘PTL::Thread’ has a previous declaration here
 typedef std::thread                     Thread;
                                         ^~~~~~
In file included from acceleritas/external/celeritas/src/physics/base/Applicability.hh:14,
                 from acceleritas/external/celeritas/src/physics/base/Model.hh:14,
                 from acceleritas/external/celeritas/src/physics/base/PhysicsParams.hh:15,
                 from acceleritas/app/g4hpc/src/UserTaskManager.cc:10:
acceleritas/external/celeritas/src/base/OpaqueId.hh:121:40: error: template argument 1 is invalid
 using ThreadId = OpaqueId<struct Thread>;
```

PR #919, opened 2023-09-01
Use trailing underscore for "sentinel" classes in OpaqueId

sethrj:
I ran into a problem on a branch that a new `namespace celeritas { namespace csg { struct Surface; }}` caused an ambiguity conflict with the implicitly defined struct in `namespace celeritas { using SurfaceId = OpaqueId<struct Surface>; }`. Since we have previously had similar problems (#276) I think it would make it clearer to mark with a trailing underscore the classes that *don't* correspond to a real class.

```
/Users/seth/.local/src/celeritas/test/orange/CsgTree.test.cc:107:20: error: reference to 'Surface' is ambiguous
    EXPECT_EQ(Node{Surface{S{1}}}, tree_[N{2}]);
                   ^
/Users/seth/.local/src/celeritas/src/orange/Types.hh:35:35: note: candidate found by name lookup is 'celeritas::Surface'
using SurfaceId = OpaqueId<struct Surface>;
                                  ^
/Users/seth/.local/src/celeritas/src/orange/construct/CsgTypes.hh:55:8: note: candidate found by name lookup is 'celeritas::csg::Surface'
struct Surface
       ^
1 error generated.
```

Also I've changed a couple places (`OnFaceT`, vecgeom `VolumeId` typedef) to reuse existing OpaqueId types rather than effectively redeclaring them.

### DIFF
--- a/src/orange/univ/detail/Types.hh
+++ b/src/orange/univ/detail/Types.hh
@@ -26,12 +26,9 @@ namespace detail
  * bit field to compress the ID and sense at the cost of reducing the max
  * number of surfaces/faces by a factor of two.
  */
-template<class ValueT>
+template<class IdT>
 class OnTface
 {
-  public:
-    using IdT = OpaqueId<ValueT>;
-
   public:
     //! Not on a surface
     constexpr OnTface() = default;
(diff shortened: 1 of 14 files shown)


test/CMakeLists.txt
```diff
@@ -201,6 +201,7 @@ celeritas_setup_tests(SERIAL PREFIX orange
 # Base
 celeritas_add_test(orange/BoundingBox.test.cc)
 celeritas_add_test(orange/BoundingBoxUtils.test.cc)
+celeritas_add_test(orange/CsgTree.test.cc)
```
amandalund:
```suggestion
```

amandalund (approved):
Looks good to me.

pcanal:
The commit message can be enhanced by adding a copy/paste of the error message (eg. the one listed in https://github.com/celeritas-project/celeritas/pull/276)

src/orange/univ/detail/Types.hh
```diff
@@ -26,12 +26,9 @@ namespace detail
  * bit field to compress the ID and sense at the cost of reducing the max
  * number of surfaces/faces by a factor of two.
  */
-template<class ValueT>
+template<class IdT>
```
pcanal, line 29: Isn't that a change in functionality?  (i.e. is it called out in the commit logs).   Namely to obtain the same effective type as before one needs to use `OnTface<OpaqueId< old_template_param > ` ?

sethrj: It's a change to a detail class, only used by the type aliases lower in that file.

pcanal: Exactly why a note in the commit log is appropriate.  The main change (` Open
Use trailing underscore for "sentinel" classes in OpaqueId`) is essential a non functional change.   If we come back later to look at the resulting commit and there is no note, the viewer will 'waste' time understanding how this part of the change is related (especially since it is a (internal but) functional change.

PR #920, opened 2023-09-02
Fix Geant4 step limiter when ionization is disabled

amandalund:
I believe the difference @mrguilima is seeing between Geant4 and Celeritas in the bremsstrahlung step length is because the step is *only* being limited by the distance to interaction (and not the range) in our Geant4 apps. When a process is registered to the physics list, the [default process ordering](https://github.com/Geant4/geant4/blob/master/source/physics_lists/builders/OrderingParameterTable) is used (this specifies in what order the processes' at rest, along-step, and post-step `GPIL`/`DoIt` methods are called by the stepping manager, and whether any of them are inactive). The along-step methods for ionization are active and for bremsstrahlung are inactive by default, since the ionization dE/dx and range tables will contain the contributions from all energy loss processes. **When ionization is missing, we need to activate the along-step for brems**—that is done here in our `GeantPhysicsList`. I also added an option to set the default cutoff value (mainly for debugging).

@mrguilima could you update the geant validation app with this change and see if the brems results improve?

### DIFF
--- a/src/celeritas/ext/detail/GeantPhysicsList.cc
+++ b/src/celeritas/ext/detail/GeantPhysicsList.cc
@@ -24,6 +24,7 @@
 #include <G4PhotoElectricEffect.hh>
 #include <G4PhysicsListHelper.hh>
 #include <G4Positron.hh>
+#include <G4ProcessManager.hh>
 #include <G4ProcessType.hh>
 #include <G4Proton.hh>
 #include <G4RayleighScattering.hh>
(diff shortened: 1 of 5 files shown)


sethrj (approved):
Great catch @amandalund ! I will wait on @mrguilima or @stognini to compare the results before merging.

src/celeritas/ext/detail/GeantPhysicsList.cc
```diff
@@ -85,6 +86,7 @@ GeantPhysicsList::GeantPhysicsList(Options const& options) : options_(options)
         em_parameters.SetMscEnergyLimit(100 * CLHEP::TeV);
     }
     em_parameters.SetApplyCuts(options.apply_cuts);
+    this->SetDefaultCutValue(options.default_cutoff * CLHEP::cm);
```
stognini, line 89: Wouldn't this forcefully override any cutoff value defined originally by the gdml?

amandalund: I'm not positive about that... but at least I don't think it will change the current behavior: `G4RunManagerKernel::InitializePhysics()` calls `physicsList->SetCuts()`. The default `SetCuts()` method in `G4VUserPhysicsList` calls `SetDefaultCutValue()`, so since we don't have our own `SetCuts()` method it will be called either way. So all this *should* do is possibly change the default from 1 mm to something else.

sethrj:
@stognini would this change have to be adapted to the separate validation code?

stognini:
Here is the summary after some testings today and discussing the results with @amandalund:
- Geant4 has a final step that streams the particle to the geometry boundary when it is below the interaction threshold. That led to results between G4 and Celeritas being in agreement for small steps, but with a discrepancy caused by the aforementioned final steps streaming out of the geometry.
- After updating the processing order in the `PhysicsList` of the `geant4-validation-app` to be identical to the Celeritas' `GeantPhysicsList`, the along-step makes the electrons lose energy until they reach `E = 0` and, thus, Celeritas and Geant4 results agree:

brems-g4-cel

stognini (approved):

stognini:
> @stognini would this change have to be adapted to the separate validation code?

Yes. I am pushing that update to the `benchmarks` repo in a bit.

src/celeritas/ext/detail/GeantPhysicsList.cc
stognini: So that means that we have to add the `SetCuts()` function in the near future, when we actually start using the secondary production thresholds correctly.

Issue #892, opened 2023-08-15
Investigate a rare crash from SBEnergyDistribution


whokion:
Investigate and resolve a rare assertion failure from [src/celeritas/em/distribution/SBEnergyDistribution.hh:168](https://github.com/celeritas-project/celeritas/blob/develop/src/celeritas/em/distribution/SBEnergyDistribution.hh#L168),
`CELER_ASSERT(xs >= 0 && xs <= 1 / inv_max_xs_);`
while running `app/celer-g4` with cms-run3 or cms-hllhc geometry configuration.  An example failure is the case for
xs = 3.36544, 1/inv_max_xs_ = 3.09831 with exit_energy = 0.000996311 MeV.

amandalund:
@whokion could you share the full input to reproduce this failure?

whokion:
@amandalund  I reran `celer-g4` of your branch, [celer-g4-input](https://github.com/amandalund/celeritas/tree/celer-g4-input) with `cms-hllhc.gdml` and `pythia_pp_14TeV_ttbar_10.hepmc3` which are available from https://github.com/celeritas-project/benchmarks/tree/main/full-cms-test/input and with one thread (i.e., export export G4FORCENUMBEROFTHREADS=1), and reproduced the crash from the 5th event on my system (V100 with gcc9.3.0, cuda 12.1.0) - with two threads, I could not reproduce the crash and the application ran successfully (10 events for this test case), so seems that the crash is sort of random.
```
/work1/celeritas/syjun/celeritas/src/celeritas/src/celeritas/em/distribution/SBEnergyDistribution.hh:171:
celeritas: internal assertion failed: xs >= 0 && xs <= 1 / inv_max_xs_
G4WT0 > 
-------- EEEE ------- G4Exception-START -------- EEEE -------
*** G4Exception : celer0003
      issued by : Thrust GPU library
copy_if failed to synchronize: cudaErrorLaunchFailure: unspecified launch failure
*** Fatal Exception *** core dump ***
G4WT0 > G4Track (0x7fb15d945320) - track ID = 487097, parent ID = 480560
G4WT0 >  Particle type : e- - creator process : hIoni, creator model : model_DeltaElectron
G4WT0 >  Kinetic energy : 1.85508 keV - Momentum direction : (-0.854488,0.33739,-0.394992)
G4WT0 >  Step length : 0 fm  - total energy deposit : 0 eV 
[...]

PR #921, opened 2023-09-03
Fix density correction calculation in Seltzer-Berger test

amandalund:
The density correction should be calculated from the total energy.


sethrj:
@amandalund Aha, so it should be total energy (rest + kinetic) rather than kinetic. Thanks! ~~Is this expected to fix Soon's issue #892 ?~~ Ok this is just a testing issue and not anything in the actual transport.

sethrj (approved):

amandalund:
Right, just a testing error I noticed while I was investigating that issue (but I do think I see the problem there, so should be a fix for that coming).

PR #401, opened 2022-04-05
Seltzer-Berger positron correction fix

amandalund:
I believe we were missing a factor of two pi in the SB positron cross section correction factor -- added here. I've also:
- Capped the number of iterations in the SB rejection loop for sampling the gamma energy at 100 (same as Geant4)
- Updated the geant exporter to use the relativistic pair production model instead of Bethe-Heitler and to use both the SB and relativistic brems models

PR #922, opened 2023-09-04
Fix Seltzer-Berger max xs for positrons

### LINKED ISSUE #892: Investigate a rare crash from SBEnergyDistribution

amandalund:
This reverts to using the max electron Seltzer-Berger DCS for positrons instead of the first DCS value (which I had mistakenly based on a zero production cut in #401). The xs scaling factor depends on the production cut, and will approach 1 as the exiting gamma energy approaches the cutoff value -- so in the extreme case the max positron xs will be the same as for electrons.

This should resolve the failure in #892, which occurs for low energy positrons when the sampled exiting gamma energy is very close to the production cut.

### DIFF
--- a/test/celeritas/em/SeltzerBerger.test.cc
+++ b/test/celeritas/em/SeltzerBerger.test.cc
@@ -207,9 +207,14 @@ TEST_F(SeltzerBergerTest, sb_energy_dist)
     };
 
     // Note: the first point has a very low cross section compared to
-    // ionization so won't be encountered in practice. The differential cross
-    // section distribution is much flatter there, so there should be lower
-    // rejection. The second point is where the maximum of the differential SB
+    // ionization so won't be encountered in practice. The electron
+    // differential cross section distribution is much flatter there, so there
+    // should be lower rejection. However, the scaled positron DCS drops
+    // quickly from its maximum (equal to the electron DCS) to near zero at
+    // this point, so the rejection rate will be very high for incident
+    // positrons (see
+    // https://github.com/celeritas-project/celeritas/pull/922#discussion_r1315295079).
+    // The second point is where the maximum of the electron differential SB
     // data switches between a high-exit-energy peak and a low-exit-energy
     // peak, which should result in a higher rejection rate. The remaining
     // points are arbitrary.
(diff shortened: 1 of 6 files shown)


test/celeritas/em/SeltzerBerger.test.cc
```diff
@@ -262,39 +279,23 @@ TEST_F(SeltzerBergerTest, sb_energy_dist)
...
-        0.077435723446377, 0.089209770419741, 0.064206311258969,
-        0.064814238042425};
+        0.90270157074556, 0.49736065674058, 0.27711716215819,
+        0.081515129333292, 0.068559142299853, 0.065803331441324,
+        0.064344514250384, 0.079512002547402, 0.077647502218254,
+        0.085615341879476, 0.086428313853775, 0.065321200129584};
     static double const expected_avg_engine_samples[] = {4.0791015625,
-        4.060546875, 5.13623046875, 4.6572265625, 4.43115234375, 4.35791015625,
-        9.3740234375, 4.65478515625};
+        137.044921875, 4.060546875, 15.74169921875, 5.103515625, 5.26953125,
```
sethrj, line 298: That's a lot of rejection, especially considering the comment above:

> Note: the first point has a very low cross section compared to ionization so won't be encountered in practice. The differential cross section distribution is much flatter there, so there should be lower rejection. 

Do you have any plots of the maximum cross section versus the actual distribution?

amandalund: Good point, I'll update the comment since it only applies to the electron DCS -- for low incident energies the positron DCS deviates a lot. These are cross sections for a 1 keV incident particle (same as that first point in the test) and a few different cutoff energies (0.9 keV is used in the test):
sb_dcs

So, the sampling efficiency for the positrons here will be quite poor. For higher incident energies though the positron DCS will be much closer to the electron DCS.

sethrj: Yuck, that rejection rate is high indeed, but it should be a very unusual case indeed. Thanks for the plot!

sethrj (approved):

PR #923, opened 2023-09-05
Implement CSG tree simplification in ORANGE

sethrj:
Constructing and simplifying a CSG is a critical step to building efficient volumes from shapes and surfaces and other more complex objects. This PR adds a `CsgTree` which simplifies nodes as they're added, and provides algorithms for simplifying the tree based on known quantities (e.g. the universe's bounding shape). The replacement and simplification is much more carefully implemented than in the SCALE version of ORANGE.

There's one major superficial issue we should discuss, but perhaps it should wait until we reorganize celeritas for v1.0: how do we break the code into subdirectories, and do we add sub-namespaces? I've been opposed to sub-namespaces since reading [Abseil: Namespace Naming](https://abseil.io/tips/130) but maybe should allow it for CSG and BIH? Or should I just prefix all the classes with `Csg` and be done with it? Or should I move the `CsgTypes` classes inside the `CsgTree` class definition? (As always, happy to hear @pcanal or @amandalund opinion on this.)


elliottbiondo:
Is there any natural place this could be split into multiple PRs?

sethrj:
@elliottbiondo I know it's a long one, sorry about that... I did look when I started the pull request. It makes more sense with them together, but I could definitely split into a "core CSG tree" and a "CSG algorithms" PR (maybe 2:1 code ratio?).

src/orange/construct/CsgTypes.hh
```diff
@@ -0,0 +1,223 @@
...
+// Write a variant node to a stream
+std::ostream& operator<<(std::ostream& os, Node const&);
+
+// Get a string representation of a variant node
+std::string to_string(Node const&);
+
+//---------------------------------------------------------------------------//
+// Helper functions
+//---------------------------------------------------------------------------//
+//! Whether a node is a boolean true/false
```
elliottbiondo: True/False

src/orange/construct/CsgTypes.cc
```diff
@@ -0,0 +1,87 @@
...
+#include <iostream>
+#include <sstream>
+
+#include "corecel/io/Join.hh"
+
+namespace celeritas
+{
+namespace csg
+{
+//---------------------------------------------------------------------------//
```
elliottbiondo, line 19: Doc strings are missing here, as in the .hh file. I think it's fine as is, but you could also add them for consistency.

src/orange/construct/CsgTypes.cc
```diff
@@ -0,0 +1,87 @@
...
+std::ostream& operator<<(std::ostream& os, False const&)
+{
+    os << "false";
+    return os;
+}
+
+//---------------------------------------------------------------------------//
+std::ostream& operator<<(std::ostream& os, Aliased const& n)
+{
+    os << "->{" << n.node.unchecked_get() << '}';
```
elliottbiondo, line 36: Is there value in having "->" be replaced with "=" to be consistent with the JSON representation? Same with "not" below.

src/orange/construct/CsgTree.hh
```diff
@@ -0,0 +1,112 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using Node = csg::Node;
+    using NodeId = csg::NodeId;
+    using size_type = NodeId::size_type;
+    //!@}
+
+  public:
+    // Construct with no nodes except "true" and "false"
```
elliottbiondo, line 54: Can you explain why there are true and false nodes created upon construction? Maybe this is best for the class level doc string.

sethrj:
Ping @elliottbiondo 😉

src/orange/construct/detail/NodeSimplifier.cc
```diff
@@ -0,0 +1,159 @@
...
+    // Replace any aliases in each daughter
+    for (NodeId& d : j.nodes)
+    {
+        if (auto repl = visit_node_(AliasSimplifier{}, d))
+        {
+            d = repl;
+        }
+        if (d == constant_node)
+        {
+            // Return early because
```
elliottbiondo: ...

src/orange/construct/CsgTree.hh
```diff
@@ -0,0 +1,112 @@
...
+  private:
+    // Tree structure: nodes may have been simplified
+    std::vector<Node> nodes_;
+
+    // Hashed nodes, both original and simplified
+    std::unordered_map<Node, NodeId> ids_;
+
+    //// HELPER FUNCTIONS ////
+
+    // Get a node (only this class can modify the node once added
```
elliottbiondo: added)

elliottbiondo (approved):

src/orange/construct/CsgTypes.cc
sethrj: I thought that when reading them, `->{3}` better conveys the meaning of "is a pointer to node 3" rather than "=3" which might mean "is surface 3" or something. I wanted to keep the JSON values a single character. Same with "not" etc. I could further expand this to "pointer to node 3" or something to further improve readability, but I also wanted to keep it compact.

sethrj:
Thanks @elliottbiondo . Hopefully the new CSG tree documentation sheds some light on the next review.

PR #924, opened 2023-09-06
Refactor surface input with vector of variants

sethrj:
This changes the ORANGE input to use surface variants rather than raw compressed surfaces. This should be a more stable method of passing data between SCALE and ORANGE in the short term. *Marking as draft until #923 is merged and the existing SCALE interface is updated.*


elliottbiondo:
As best as I can tell, this is going to break the `Celeritas_Converter` code within Geometria/orange, so maybe this should wait until @tmdelellis's branch is merged.

sethrj:
Yes, that's what I meant by
 > Marking as draft until https://github.com/celeritas-project/celeritas/pull/923 is merged and the existing SCALE interface is updated. 

Once the various SCALE-related branches have been merged then I'll activate this one.

sethrj:
Marking as draft until @elliottbiondo 's universe dispatch branch #973 is merged in.

sethrj:
@elliottbiondo Now that the last of the empire paper PRs are in, we can start moving forward with celeritas orange input development. The eventual goal I have in mind is for the `UnitInput` to be an intermediate object managed during construction: so we build up and deduplicate the vector of surfaces in this container.

src/orange/construct/OrangeInputIO.json.cc
```diff
@@ -192,8 +186,22 @@ void from_json(nlohmann::json const& j, RectArrayInput& value)
...
+            // Read and convert transform
+            Translation tr{slice<3>(make_span(translations), i)};
+            if (CELER_UNLIKELY(tr.translation() == (Real3{0, 0, 0})))
+            {
+                daughter.transform = NoTransformation{};
+            }
+            else
+            {
+                daughter.transform = tr;
+            }
```
amandalund: Could you use `make_transform()` here?
```suggestion
            daughter.transform
                = detail::make_transform(slice<3>(make_span(translations), i));
```

test/orange/OrangeGeoTestBase.cc
```diff
@@ -125,43 +122,32 @@ void OrangeGeoTestBase::build_geometry(TwoVolInput inp)
...
-        vi.bbox = bbox;
-        input.volumes.push_back(vi);
-    }
-
-    // Save bbox
-    input.bbox = {{-inp.radius, -inp.radius, -inp.radius},
-                  {inp.radius, inp.radius, inp.radius}};
-
-    input.label = "two volumes";
+        vi.bbox = BBox::from_infinite();
```
amandalund: Was the change to this bounding box intentional?

amandalund (approved):
Looks good to me @sethrj!

src/orange/construct/OrangeInputIO.json.cc
sethrj: Yes exactly, thank you. This file has been rebased too many times 😔

test/orange/OrangeGeoTestBase.cc
sethrj: Nope, great catch!

PR #861, opened 2023-07-25
Add Wentzel single Coulomb scattering model

hhollenb:
This PR incorporates just the Wentzel model used in single coulomb scattering, and its dependent files (e.g. data, interactor, distribution, etc.). It also includes regression testing for said files.

The code is still missing some final correctness checks and references for equations. I've tried to note where the code differs from Geant4's OK and VI implementation, specifically where I'm uncertain of the reason for their choices. One should note that the G4 implementation supports MSC, which is not included in this model. Before submitting a PR for the full process, I'll also include a write up justifying these differences. 

===== Geant4 implementation files:
G4eCoulombScatteringModel
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/src/G4eCoulombScatteringModel.cc
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/include/G4eCoulombScatteringModel.hh

G4WentzelOKandVIxSection
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/src/G4WentzelOKandVIxSection.cc
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/include/G4WentzelOKandVIxSection.hh

G4ScreeningMottCrossSeciton
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/src/G4ScreeningMottCrossSection.cc
https://github.com/Geant4/geant4/blob/master/source/processes/electromagnetic/standard/include/G4ScreeningMottCrossSection.hh

PR #926, opened 2023-09-06
Update documentation and precalculate Wenzel nuclear form factor constants

sethrj:
This fixes a few small issues I noticed while updating some doxygen documentation that we'd missed during review of #861:
- Private data didn't have `_` suffix, and since it is function-like it should be named like a function
- Move some reused values outside of expressions to improve readability
- Removed doxygen markers from *declarations*: they should only be on *definitions* (e.g. inline functions, class definitions, non-inline function definitions)
- Doxygen briefs should fit on a single line
- Interactor test should be named after the model, not the process
- Nuclear form prefactor can be precalculated (one real per isotope)
- Made a function that didn't depend on state data `static`

### DIFF
--- a/src/celeritas/em/data/WentzelData.hh
+++ b/src/celeritas/em/data/WentzelData.hh
@@ -82,9 +82,14 @@ struct WentzelData
 {
     template<class T>
     using ElementItems = celeritas::Collection<T, W, M, ElementId>;
+    template<class T>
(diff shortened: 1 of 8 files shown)


sethrj:
Ping @hhollenb : could you take a look at these minor changes to your code when you have a minute? Thanks!

hhollenb (approved):
Everything looks good!

PR #805, opened 2023-06-07
Create CUDA streams when integrating with Geant4

amandalund:
Previously we were only actually creating the streams in the celer-sim app.

PR #927, opened 2023-09-06
Fix resetting of CUDA streams when running through accel

sethrj:
@esseivaju 's recent speedup diagram of Celeritas with Tilecal shows very poor multithread scaling:
tilecal scaling

He tracked this down to the fact that all the kernels were using the default stream:
nsight screen grab

Further digging revealed that the stream IDs are being passed correctly through accel and to the kernels, but the CUDA streams themselves were being reset. It turns out that this is because we're calling `activate_device` on the Geant4 worker threads *after* calling `set_num_streams` on the main device. The subsequent calls were overwriting the device.

This additionally changes the behavior of `activate_device` so that it can be called *at most once* to activate one. There's too much code in Celeritas that uses the global `device` so it's likely that changing or resetting it could break something, as it did with the streams here. The `accel` setup code now uses `activate_device_local`.

After fixing the default stream, 16-core performance improved 20%:
updated tilecal scaling

### DIFF
--- a/src/celeritas/global/detail/ActionSequence.cc
+++ b/src/celeritas/global/detail/ActionSequence.cc
@@ -16,11 +16,14 @@
 #include "corecel/Types.hh"
 #include "corecel/cont/EnumArray.hh"
 #include "corecel/cont/Range.hh"
+#include "corecel/sys/Device.hh"
 #include "corecel/sys/ScopedProfiling.hh"
 #include "corecel/sys/Stopwatch.hh"
+#include "corecel/sys/Stream.hh"
 #include "celeritas/global/ActionInterface.hh"
 
 #include "../ActionRegistry.hh"
+#include "../CoreState.hh"
 
 namespace celeritas
 {
@@ -90,6 +93,12 @@ void ActionSequence::begin_run(CoreParams const& params, CoreState<M>& state)
 template<MemSpace M>
 void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
 {
+    Stream::StreamT stream = nullptr;
+    if (M == MemSpace::device && options_.sync)
+    {
+        stream = celeritas::device().stream(state.stream_id()).get();
+    }
+
     ScopedProfiling profile_this{"step"};
     if (M == MemSpace::host || options_.sync)
     {
@@ -101,7 +110,7 @@ void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
             actions_[i]->execute(params, state);
             if (M == MemSpace::device)
             {
-                CELER_DEVICE_CALL_PREFIX(DeviceSynchronize());
+                CELER_DEVICE_CALL_PREFIX(StreamSynchronize(stream));
             }
             accum_time_[i] += get_time();
         }
--- a/src/celeritas/global/detail/ActionSequence.cc
+++ b/src/celeritas/global/detail/ActionSequence.cc
@@ -93,7 +93,7 @@ void ActionSequence::begin_run(CoreParams const& params, CoreState<M>& state)
 template<MemSpace M>
 void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
 {
-    Stream::StreamT stream = nullptr;
+    [[maybe_unused]] Stream::StreamT stream = nullptr;
     if (M == MemSpace::device && options_.sync)
     {
         stream = celeritas::device().stream(state.stream_id()).get();
--- a/scripts/cmake-presets/wildstyle.json
+++ b/scripts/cmake-presets/wildstyle.json
@@ -18,7 +18,7 @@
         "CELERITAS_USE_SWIG":    {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_VecGeom": {"type": "BOOL", "value": "OFF"},
         "CMAKE_CXX_COMPILER": "/usr/bin/g++",
-        "CMAKE_CXX_FLAGS": "-Wall -Wextra -pedantic -Werror -Wno-error=deprecated-declarations -fdiagnostics-color=always",
+        "CMAKE_CXX_FLAGS": "-Wall -Wextra -pedantic -Werror -Wno-error=deprecated-declarations -Wno-attributes -fdiagnostics-color=always",
         "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O2 -g -DNDEBUG -fno-inline -fno-omit-frame-pointer",
         "CMAKE_CXX_FLAGS_RELEASE": "-O3 -DNDEBUG -march=skylake-avx512 -mtune=skylake-avx512",
         "CMAKE_CUDA_FLAGS": "-Werror all-warnings  -Wno-deprecated-gpu-targets",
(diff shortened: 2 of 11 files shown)


sethrj:
P.S. the unrelated changes are to get wildstyle builds to work and be quiet... i can move those to the release branch from here.

amandalund (commented):
Great catch! Alternatively, would calling `activate_device_local()` on the worker threads (like we do in celer-sim) instead of overwriting the device work?

sethrj:
Actually that's a good point... why aren't we doing that  🤔

amandalund:
I'm guessing we added support for streams/local device activation after the accel stuff and missed that we needed to update it... should have been done in #805.

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -101,7 +110,7 @@ void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
             actions_[i]->execute(params, state);
             if (M == MemSpace::device)
             {
-                CELER_DEVICE_CALL_PREFIX(DeviceSynchronize());
+                CELER_DEVICE_CALL_PREFIX(StreamSynchronize(stream));
```
amandalund, line 113: Have you looked at how this impacts performance? I recall having actually seen slightly _better_ performance _with_ a device sync than without when running with multiple streams, and doing a quick test just now (celer-sim with cms2018+field+msc, 8 events/threads) I see an even bigger improvement with the stream sync (like 1.2x faster). Curious if you see something similar, because it's not obvious to me why this might be.

amandalund (approved):
Thanks @sethrj and @esseivaju! 20% improvement sounds about right.. that's about the max speedup I saw using multiple streams vs. the default stream when I originally tested with celer-sim.

src/celeritas/global/detail/ActionSequence.cc
esseivaju: Interesting, I don't see a difference (tested with 32 threads and 32/64 events), sync doesn't affect the wall time.

amandalund: Weird... these are the timing results I'm getting with no sync:
```console
real	3m58.476s
user	7m30.713s
sys	0m2.825s
$ jq '.result["runner"]["time"]' no-sync.out.json 
{
  "actions": {},
  "setup": 31.052430465,
  "steps": [],
  "total": 206.873054614
}
```
and with sync:
```console
real	3m2.136s
user	14m19.799s
sys	0m3.653s
$ jq '.result["runner"]["time"]' sync.out.json 
{
  "actions": {},
  "setup": 30.891581001,
  "steps": [],
  "total": 150.659916742
}
```
and the input I'm using, with `OMP_NUM_THREADS=8`:
<details><summary>JSON input</summary>

```json
{
"geometry_filename": "/home/alund/celeritas_project/regression/input/cms2018.gdml",
"physics_filename": "/home/alund/celeritas_project/regression/input/cms2018.gdml",
"primary_gen_options": {
"seed": 0,
"direction": {"distribution": "isotropic", "params": []},
"energy": 10000,
"num_events": 8,
"pdg": 11, 
"position": [
0,
0,
0
],
"primaries_per_event": 1300
},
"geant_options": {
"eloss_fluctuation": false,
"em_bins_per_decade": 56, 
"msc": "urban_extended",
"physics": "em_basic"
},
"mag_field": [
0.0,
0.0,
1.0
],
"initializer_capacity": 8388608,
"max_events": 8,
"num_track_slots": 131072,
"max_steps": 1000000,
"secondary_stack_factor": 3.0,
"seed": 20220904,
"merge_events": false,
"sync": false,
"use_device": true
}
```
</details>

sethrj: @amandalund Interesting that the user time *doubled* even though the real time decreased. Is this on consumer hardware or HPC? It could be that the sync is causing the CPU threads to spinlock rather than context switch while waiting...

amandalund: Yeah I thought that was odd... this is on HPC (Intel Xeon Gold 6152 CPU 22c 2.10GHz + NVIDIA Tesla V100 SXM2 w/32GB HBM2).

sethrj: It could also be some weird interaction between OpenMP and CUDA? Even though CUDA nominally supports OpenMP, the latter prohibits any kind of thread interaction outside of OpenMP, and CUDA is definitely using pthread under the hood. Maybe that's why @esseivaju didn't see any difference: he's not using OpenMP? Still, I would expect the net effect to be a slowdown rather than speedup.

amandalund: Interesting, I also see it with celer-g4.
no sync:
```console
real	2m7.980s
user	3m43.651s
sys	0m1.806s
```
sync:
```console
real	1m46.656s
user	8m40.084s
sys	0m2.068s
```
with `G4FORCENUMBEROFTHREADS=8` and input:
<details><summary>JSON input</summary>

```json
{
"geometry_file": "/home/alund/celeritas_project/regression/input/cms2018.gdml",        
"output_file": "cms2018-celer-g4.out.json",                                            
"primary_options": {                                                                   
"seed": 0,                                                                             
"direction": {"distribution": "isotropic", "params": []},                              
"energy": 10000,                                                                       
"num_events": 8,
"pdg": 11,                                                                             
"position": [                                                                          
0,
0,                                                                                     
0                                                                                      
],                                                                                     
"primaries_per_event": 1300                                                            
},                                                                                     
"physics_list": "geant_physics_list",
"physics_options": {
"eloss_fluctuation": false,                                                            
"em_bins_per_decade": 56,
"msc": "urban_extended",
"physics": "em_basic"
},
"field_type": "uniform",
"field": [
0.0,
0.0,
1.0
],
"write_sd_hits": false,
"initializer_capacity": 8388608,                                                       
"max_events": 128,
"num_track_slots": 131072,                                                             
"max_steps": 1000000,                                                                  
"secondary_stack_factor": 3.0,                                                         
"seed": 20220904,                                                                      
"sync": true                                                                           
}
```

</details>

amandalund: Here’s a relevant question with an answer that I think gives a nice example/explanation as to why we might see better performance _with_ stream synchronization when using async copies with pageable memory: https://forums.developer.nvidia.com/t/performances-of-multi-thread-vs-multi-process-with-mps/64236/2

esseivaju: @amandalund #910 is still kinda broken (only compiles with CUDA) but I've done some profiling on it and it significantly reduces the number of memcpyasync to pageable memory. Before I had many ~25kb transfers, now the only memcpyasync to pageable memory left comes from thrust (exclusive_scan_counts, remove_if_alive in TrackInitAlgorithms, and copy_if in DetectorSteps) and are 4B (probably the return value).

If you can compile that branch, I'd be curious to know if it helps reduce the timing for the non-sync version.

amandalund: @esseivaju I tried out the pinned allocator branch, but looks like the time is still about the same as with no stream synchronization.

PR #929, opened 2023-09-07
Fix ORANGE multi-level surface crossing

tmdelellis:
This fixes the `onSurface` designation that used to be set on all geometry levels, but was not getting properly cleared on levels where the track was not on a surface.  This lead to situations (see attached figure) where a track on a lower level surface was being tested for `onSurface == true` at higher levels.  For surfaces like cylinders, this can have deleterius effects because the cylinder intersections algorithms tests `onSurface` as an optimization.  

The attached figure shows an example failure (added to `Orange.test.cc`.  In this example, when the particle was at the purple/green hex surface ($x=-2.0$) it was failing to calculate the correct distance to the cylindrical surface on level 0 because it thought the track was on a surface *in level 0*.  The new code now only stores one surface flag per track and it is only assigned true on the proper level.

mesh-track-on-lower-surface


tmdelellis:
Here is the complete notebook showing the figure above with additional data.

[models.pdf](https://github.com/celeritas-project/celeritas/files/12620628/models.pdf)

sethrj (approved):

PR #930, opened 2023-09-07
Add Windows support

sethrj:
This adds non-CI-tested support for building and running on Windows. (Compiled with MS Visual Studio 2022, 14.37.32822.) All applications and tests build and pass in a near-minimal (only JSON and Googletest enabled) configuration.

Key compatibility issues are:
- Fix many narrowing conversions, including one that would have introduced a serious bug due to integer overflow in the Dorman–Prince stepper constant
- Use `if constexpr` rather than triad to fix compiler crash from template recursion. Before this, `ipow` relied on an optimization trick.
- Fix some missing includes
- Add conditional support for `SIGUSR2` and `isatty`
- Fix  misuse of `string_view::const_iterator` as `char const*`
- Work around some MSVC lambda capture scope issues
- Normalize project source dir path written to celeritas test config file

Celeritas is about to become a dependency of [SCALE](https://www.ornl.gov/scale) to use ORANGE for CPU and GPU geometry. SCALE must compile on Windows.

### DIFF
--- a/src/corecel/data/Collection.hh
+++ b/src/corecel/data/Collection.hh
@@ -294,7 +294,7 @@ class Collection
     //! Direct accesors to underlying data
     CELER_FORCEINLINE_FUNCTION size_type size() const
     {
-        return this->storage().size();
+        return static_cast<size_type>(this->storage().size());
     }
     CELER_FORCEINLINE_FUNCTION bool empty() const
     {
--- a/src/corecel/io/ColorUtils.cc
+++ b/src/corecel/io/ColorUtils.cc
@@ -11,7 +11,7 @@
 #include <cstdlib>
 #include <string>
 #ifndef _WIN32
-#include <unistd.h>
+#    include <unistd.h>
 #endif
 
 #include "corecel/sys/Environment.hh"
--- a/src/corecel/io/ColorUtils.cc
+++ b/src/corecel/io/ColorUtils.cc
@@ -10,7 +10,9 @@
 #include <cstdio>
 #include <cstdlib>
 #include <string>
+#ifndef _WIN32
 #include <unistd.h>
+#endif
 
 #include "corecel/sys/Environment.hh"
 
@@ -23,7 +25,7 @@ namespace celeritas
 bool use_color()
 {
     const static bool result = [] {
-        FILE* stream = stderr;
+        [[maybe_unused]] FILE* stream = stderr;
         std::string color_str = celeritas::getenv("CELER_COLOR");
         if (color_str.empty())
         {
@@ -44,11 +46,13 @@ bool use_color()
             // Color is explicitly enabled
             return true;
         }
+#ifndef _WIN32
         if (!isatty(fileno(stream)))
         {
             // This stream is not a user-facing terminal
             return false;
         }
+#endif
         if (const char* term_str = std::getenv("TERM"))
         {
             if (std::string{term_str}.find("xterm") != std::string::npos)
--- a/src/corecel/math/detail/MathImpl.hh
+++ b/src/corecel/math/detail/MathImpl.hh
@@ -55,7 +55,7 @@ inline void sincospi(T x, T* sptr, T* cptr)
     }
     else
     {
-        x *= m_pi;
+        x *= static_cast<T>(m_pi);
         cval = std::cos(x);
         sval = std::sin(x);
     }
--- a/src/corecel/sys/MemRegistry.hh
+++ b/src/corecel/sys/MemRegistry.hh
@@ -20,12 +20,12 @@ namespace celeritas
 //! SI prefix for multiples of 1024
 struct Kibi
 {
-    static CELER_CONSTEXPR_FUNCTION int value() { return 1024; }
+    static CELER_CONSTEXPR_FUNCTION long int value() { return 1024; }
     static char const* label() { return "kibi"; }
 };
 
 //! 1024 bytes
-using KibiBytes = Quantity<Kibi, int>;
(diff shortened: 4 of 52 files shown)


src/corecel/data/Collection.hh
```diff
@@ -294,7 +294,7 @@ class Collection
     //! Direct accesors to underlying data
     CELER_FORCEINLINE_FUNCTION size_type size() const
     {
-        return this->storage().size();
+        return static_cast<size_type>(this->storage().size());
```
pcanal, line 297: why is the return type of `.size()` not a `size_type`?  Can it cause (actual) problems?

src/corecel/io/ColorUtils.cc
```diff
@@ -44,11 +46,13 @@ bool use_color()
             // Color is explicitly enabled
             return true;
         }
+#ifndef _WIN32
```
pcanal, line 49: Is there a `Windows` equivalent to tell if there a terminal attached or not?

src/corecel/data/Collection.hh
sethrj: The underlying data container can be a `Span` or `std::vector` (both `std::size_t`) or `DeviceVector` (`celeritas::size_type`), but more importantly, the `size_type` of the Collection is the `size_type` of the `OpaqueId` accessor, which could be much smaller. Thinking about it now, that may not be the best policy (we should prefer not to be allocating "small" arrays), but I *do* have logic in the collection builder and device copier to ensure that we don't exceed the size of the collection.

src/corecel/io/ColorUtils.cc
sethrj: I'm not sure at all... even if there is, the Windows terminals don't natively support ANSI color sequences (only when running through something like git-bash). Either way, for the use cases of concern to SCALE, we won't be running through the command line.

src/corecel/sys/Stream.cc
```diff
@@ -17,14 +17,19 @@
 
 namespace celeritas
 {
+#if CELER_USE_DEVICE
```
pcanal, line 20: Maybe add some information here about which gcc version made us make this change (so later somebody might dare remove it).

src/corecel/data/Collection.hh
pcanal: Fair enough.

src/corecel/io/ColorUtils.cc
pcanal: humm ... isn't the code then
```
#ifdef _WIN32
     return false;
#else 
... existing code...
#endif
```

src/corecel/math/detail/MathImpl.hh
```diff
@@ -55,7 +55,7 @@ inline void sincospi(T x, T* sptr, T* cptr)
     }
     else
     {
-        x *= m_pi;
+        x *= static_cast<T>(m_pi);
```
pcanal, line 58: Not related to this change but this template function could use a limiter (concept in C++) as it is valid only for floating point types.

src/corecel/sys/MemRegistry.hh
```diff
@@ -20,12 +20,12 @@ namespace celeritas
 //! SI prefix for multiples of 1024
 struct Kibi
 {
-    static CELER_CONSTEXPR_FUNCTION int value() { return 1024; }
+    static CELER_CONSTEXPR_FUNCTION long int value() { return 1024; }
```
pcanal: What the purpose of that change?  Rather than `long int`, do we mean `int64_t` ?  (And almost more important how to we remember to use the right type in similar future construct?)

src/corecel/sys/ScopedMem.cc
```diff
@@ -74,13 +76,13 @@ MemResult get_cpu_mem()
     return result;
 }
 
-std::ptrdiff_t get_gpu_mem()
+long int get_gpu_mem()
```
pcanal:
```suggestion
std::make_signed_t<std::size_t> get_gpu_mem()
```

src/corecel/math/detail/MathImpl.hh
sethrj: The public one that calls this has a limiter. The detail class won't cause any ambiguity.

src/corecel/sys/MemRegistry.hh
sethrj: Int is 32-bit on windows and I wanted something "long"er. I try to steer away from special integer types for JSON I/O though now that I think about it, I'm not sure why. I'll unify all this.

pcanal: > I wanted something "long"er

Why does it matter here?

sethrj: Because any memory change over 2GB overflows `int` on windows.

pcanal: right, I should have look up the usage :).

But then why using `size_t` isn't the right type? (or `make_signed<size_t>`)?

sethrj: it was the conversion between size_t and ptrdiff_t that was throwing me off. I've replaced it all with size_t.

src/corecel/io/ColorUtils.cc
sethrj: You could be running through git-bash but compiled against only regular windows. (This is how I was testing.) In that case the ANSI sequences are processed correctly because git-bash defines the `TERM` variable.

pcanal (approved):
LGTM.  Thanks.

PR #702, opened 2023-03-27
Add UniformAlongStepFactory for Geant4 integration

sethrj:
I've added a helper class `UniformAlongStepFactory` that translates a `G4ThreeVector` with the field magnitude in native CLHEP/G4 units to an along-step action. I also made a few minor quality of life improvements to the MT logger.

PR #931, opened 2023-09-08
Fix uniform field unit conversion

amandalund:
Well that's why our celer-g4 performance results with a uniform field looked suspiciously good...

### DIFF
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -145,7 +145,7 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
         // Convert field units from tesla to native celeritas units
         for (real_type& v : field)
         {
-            v /= units::tesla;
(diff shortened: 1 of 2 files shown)


sethrj (approved):
My bad for missing this in the last review... is it possible to add a unit test for the along-step factory? At some point we'll need to add something more than a "does it run" test for the front end apps too.

amandalund:
I don't _think_ I changed the field units when I refactored for the json input (except for messing up the `magnitude_tesla`), but I definitely should have caught this. Yeah, an along-step factory test is a good idea (and agreed about our app tests).

sethrj:
Is this bug present in v0.3?

amandalund:
Yep, looks like it is.

sethrj:
Damn, you're right. Thanks for the catch.

sethrj:
Yeah, the unit error was in the original #702 . It just got *moved* in #890 .

PR #932, opened 2023-09-10
Rewrite soft equality for simplicity

sethrj:
This rewrites the `SoftEqual` functor to be about 15× shorter. Because ARM and CUDA have intrinsics for fabs and fmax, this ends up being 8 nonbranching instructions on ARM versus ~30 instructions with branches. The only change in behavior is that infinities no longer compare "soft equal", but since `inf - inf = nan` I think this is reasonable behavior. A new functor `EqualOr` can be composed with `SoftEqual` to regain the original behavior for unit testing (where comparing infinities does make sense). The new soft equal expression is:
$$|a - b| < \max(\epsilon_r \max(|a|, |b|), \epsilon_a)$$

@tmdelellis If you're interested in seeing how much ARM vs x86 differ and for the old vs new implementation, see [godbolt compiler explorer](https://godbolt.org/z/d654anP4P).

| Implementation | LOC | ARM 64 (Clang 16) ASM-LOC | x86-64 (Clang 16) ASM-LOC  | NVCC PTX LOC |
| ---- | --- | --- | - | - |
| Old | 25 |  32 | 42 | 72 |
| New | 5 | 10 | 28 | 27 |

### DIFF
--- a/src/corecel/math/SoftEqual.hh
+++ b/src/corecel/math/SoftEqual.hh
@@ -19,10 +19,21 @@ namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Functor for floating point equality.
+ * Functor for noninfinite floating point equality.
  *
- * \note This functor is *not commutative*: eq(a,b) will not always give the
- * same as eq(b,a).
+ * This function-like class considers an \em absolute tolerance for values near
+ * zero, and a \em relative tolerance for valuse far from zero. It correctly
+ * returns "false" if either value being compared is NaN.  The call operator is
+ * \em commutative: \c eq(a,b) should always give the same as \c eq(b,a).
+ *
+ * The actual comparison is: \f[
+ |a - b| < \max(\epsilon_r \max(|a|, |b|), \epsilon_a)
+ \f]
+ *
+ * \note The edge case where both values are infinite (with the same sign)
+ * returns *false* for equality, which could be considered reasonable because
+ * relative error is meaningless. To explicitly allow infinities to compare
+ * equal, you must test separately, e.g., `a == b || soft_eq(a, b)`.
  */
 template<class RealType = ::celeritas::real_type>
 class SoftEqual
@@ -46,8 +57,7 @@ class SoftEqual
     //// COMPARISON ////
 
     // Compare two values (implicitly casting arguments)
-    bool CELER_FUNCTION operator()(value_type expected,
-                                   value_type actual) const;
+    bool CELER_FUNCTION operator()(value_type a, value_type b) const;
 
     //// ACCESSORS ////
 
@@ -61,7 +71,7 @@ class SoftEqual
     value_type rel_;
     value_type abs_;
 
-    using traits_t = detail::SoftEqualTraits<value_type>;
+    using SETraits = detail::SoftEqualTraits<value_type>;
 };
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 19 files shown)


test/celeritas/Constants.test.cc
```diff
@@ -86,7 +86,7 @@ TEST(ConstantsTest, derivative)
 {
     // Compared against definition of Dalton, table 8 of SI 2019
     EXPECT_SOFT_NEAR(1.66053906660e-27 * units::kilogram, atomic_mass, 1e-11);
-    EXPECT_SOFT_NEAR(1.602176634e-19, e_electron * units::volt, 1e-11);
+    EXPECT_SOFT_NEAR(1.602176634e-19, e_electron * units::volt, 1e-9);
```
amandalund: I think this is actually 1.602176634e-12.

amandalund (approved):
Looks good to me @sethrj.

test/celeritas/Constants.test.cc
sethrj: I'm missing the explicit use of `units::joule`; the numerical value is the same.
Screenshot 2023-09-11 at 16 43 26

test/celeritas/Constants.test.cc
```diff
@@ -86,7 +86,8 @@ TEST(ConstantsTest, derivative)
 {
     // Compared against definition of Dalton, table 8 of SI 2019
     EXPECT_SOFT_NEAR(1.66053906660e-27 * units::kilogram, atomic_mass, 1e-11);
-    EXPECT_SOFT_NEAR(1.602176634e-19, e_electron * units::volt, 1e-9);
+    EXPECT_SOFT_NEAR(
+        1.602176634e-19 * units::joule, e_electron * units::volt, 1e-9);
```
amandalund: Now we should be able to keep the original tolerance (but doesn't really matter).
```suggestion
        1.602176634e-19 * units::joule, e_electron * units::volt, 1e-11);
```

sethrj: Oh my god, I see what you mean now. I forgot that a unit energy in CGS is erg, not Joule. Jeez, when I read
```
/Users/seth/.local/src/celeritas-temp/test/celeritas/Constants.test.cc:89: Failure
Value of: e_electron * units::volt
  Actual: 1.6021766340000001e-12
Expected: 1.602176634e-19
Which is: 1.6021766339999999e-19
(Absolute error 1.6021764737823368e-12 exceeds tolerance 1e-14)
```
I missed the exponent actual and expected magnitudes were so different 😂: I was looking at the 9 decimal places of the expected value. I must have done this the first time through when I bumped the tolerance to 1e-11—deceptive for a number that small.

Thanks for catching that mistake, sorry I'm an idiot!

amandalund: lol it confused me too, don't worry!

src/corecel/math/SoftEqual.hh
```diff
@@ -19,10 +19,21 @@ namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Functor for floating point equality.
+ * Functor for noninfinite floating point equality.
+ *
+ * This function-like class considers an \em absolute tolerance for values near
+ * zero, and a \em relative tolerance for valuse far from zero. It correctly
```
mrguilima, line 25: Found this typo (valuse -> values) on a merged commit.

sethrj: Thanks! I'll fix in an upcoming PR.

sethrj: https://github.com/celeritas-project/celeritas/pull/933/commits/14a7b9fb4da9c652fd53bb1daae91e88066be434

PR #933, opened 2023-09-10
Add soft equivalence for quadric surfaces

sethrj:
This adds a functor for checking equivalence between surfaces that are approximately equal (with intersection points differing on the order of some epsilon). This is a more rigorous approach than the current SCALE surface deduplication, which relies on simply rounding and hashing the flattened floating point representation of the object. The new implementation accounts for the differences between first- and second-order terms, e.g. between the position of a plane and the radius of a sphere.


src/corecel/math/ArrayUtils.hh
```diff
@@ -124,6 +130,24 @@ CELER_FUNCTION T norm(Array<T, N> const& v)
     return std::sqrt(dot_product(v, v));
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a unit vector.
+ *
+ * Unit vectors have an Euclidian norm magnitude of 1.
+ */
+template<class T, size_type N>
```
elliottbiondo, line 139: Would this make more sense as an in-place operation, since there is already an existing array?

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -0,0 +1,173 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two planes for near equality.
+ */
+bool SoftSurfaceEqual::operator()(Plane const& a, Plane const& b) const
+{
+    // Guard against dot product being slightly greater than 1 due to fp
+    // arithmetic
+    real_type const ndot = dot_product(a.normal(), b.normal());
+    return (ndot >= 1 || std::sqrt(1 - ndot) < soft_eq_.rel() * 3)
```
elliottbiondo: I don't understand what is going on here with `std::sqrt(1 - ndot)`. I am thinking in terms of $a \cdot b = |a| |b| cos(\theta)$. With $a$ and $b$ as normalized vectors, we get  $\sqrt(1 - cos(\theta))$ which I don't think has any physical significance. `std::sqrt(1 - ndot*ndot)` would be $\sin(\theta)$, but I don't think that makes sense either. What about the factor of 3?

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -0,0 +1,173 @@
...
+ORANGE_INSTANTIATE_OP(CylAligned);
+//! \endcond
+
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two planes for near equality.
+ */
+bool SoftSurfaceEqual::operator()(Plane const& a, Plane const& b) const
+{
+    // Guard against dot product being slightly greater than 1 due to fp
```
elliottbiondo: Might it be slightly less than 1 as well?

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -0,0 +1,173 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Compare the distance between two points for being less than the tolerance.
+ */
+bool SoftSurfaceEqual::soft_eq_distance_(Real3 const& a, Real3 const& b) const
+{
+    // This is soft equal formula but using vector distance.
+    real_type rel = soft_eq_.rel() * std::fmax(norm(a), norm(b));
+    return distance(a, b) < 3 * std::fmax(soft_eq_.abs(), rel);
```
elliottbiondo: Factor of 3?

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -0,0 +1,173 @@
...
+//! \endcond
+
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two simple quadrics for near equality.
+ */
+bool SoftSurfaceEqual::operator()(SimpleQuadric const& a,
+                                  SimpleQuadric const& b) const
+{
+    return soft_eq_distance_(make_array(a.second()), make_array(b.second()))
```
elliottbiondo, line 166: Why are the squared terms being compared with a linear comparator?

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -0,0 +1,173 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two general quadrics for near equality.
+ */
+bool SoftSurfaceEqual::operator()(GeneralQuadric const& a,
+                                  GeneralQuadric const& b) const
+{
+    return soft_eq_distance_(make_array(a.second()), make_array(b.second()))
```
elliottbiondo, line 178: Same here

elliottbiondo (changes requested):

src/corecel/math/ArrayUtils.hh
sethrj: The `normalize_direction` already does that, but the reason for making this function is because it's much nicer to be able to construct something with an rvalue than to have to execute a second statement to modify an lvalue. For example:
```c++
EXPECT_FALSE(softeq_(ref, Plane{make_unit_vector(Real3{-1, 0, 1}), 2.0}));
```
rather than the existing
```c++
Real3 normal{-1, 0, 1};
normalize_direction(&normal);
EXPECT_FALSE(softeq_(ref, Plane{normal, 2.0}));
```
which also pollutes the rest of the scope with an unused mutable variable.

It looks like `y = make_unit_vector(y)` [generates identical assembly](https://godbolt.org/z/hdEaWe7P1) as `normalize_direction(&y)` so I actually think I'll replace the implementation of the latter with the former. In a separate PR I can delete `normalize_direction`.

src/orange/surf/SoftSurfaceEqual.cc
sethrj: You're right, that definition is ad hoc. (The 3  comes from the new `is_soft_unit_vector`; see https://github.com/celeritas-project/celeritas/pull/932/files#diff-aed6792f4adb3b016a019a1f65191218845f3e9aebdeb4e226b04c751c17f971 .) I've replaced it from a fresh derivation that ensures that the error δ in intercept to two slightly different planes is less than the tolerance ε:

normal-soft-eq

src/orange/surf/SoftSurfaceEqual.cc
sethrj: Less than 1 wouldn't cause `sqrt` to be out of range, but this has been removed.

src/orange/surf/SoftSurfaceEqual.cc
sethrj: Yeah see the justification for `soft_unit_vector`, but this has been removed as well.

sethrj: This was halfwittedly copied from the array soft equal. I've removed it.

src/orange/surf/SoftSurfaceEqual.cc
sethrj: They're being compared as vector terms: the distance comparison here does a dot product and a sqrt of that. It seems to pass the unit tests nicely.

src/orange/surf/SoftSurfaceEqual.cc
sethrj: Same

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -128,11 +141,23 @@ bool SoftSurfaceEqual::operator()(SimpleQuadric const& a,
...
 
 //---------------------------------------------------------------------------//
 // PRIVATE HELPER FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Compare the square of values for soft equality.
+ */
+bool SoftSurfaceEqual::soft_eq_sq_(real_type a, real_type b) const
+{
+    return soft_eq_(std::sqrt(a), std::sqrt(b));
```
mrguilima, line 192: Comparing a,b rather than sqrt(a),sqrt(b) is a lot cheaper in terms of computing cycles, and it is usually good enough.  
Also, the comment on L54 says 'Compare the square', not the square-root.

sethrj: I'll update the document. Of course it's cheaper to multiply a number against itself than to calculate the square root.  Unfortunately it's not the same thing to compare the values against epsilon^2 versus comparing the square roots against epsilon. I wanted to put the correct implementation in first, and then once all tests pass I planned to go back to look at microoptimization.

src/orange/surf/SoftSurfaceEqual.cc
elliottbiondo: This is a great diagram to accompany your description in the doc string, is there someway it could be included and referenced there?

elliottbiondo (approved):

src/orange/surf/SoftSurfaceEqual.cc
sethrj: Maybe it'll go in the ORANGE paper 😅

PR #935, opened 2023-09-12
Temporary PR: dynamic dispatch of Tracker objects

elliottbiondo:


src/orange/OrangeTrackView.hh
```diff
@@ -808,9 +869,20 @@ OrangeTrackView::find_next_step_impl(detail::Intersection isect)
...
+        // tracker.intersect(this->make_local_state(levelid),
+        //                                      isect.distance);
+        UniverseVisitor visit_universe{params_};
+        auto local_state = this->make_local_state(levelid);
+        detail::Intersection local_isect;
+        visit_universe(
+            [&local_state, &isect, &local_isect](auto&& t) {
+                local_isect = t.intersect(local_state, isect.distance);
+            },
+            lsa.universe());
```
sethrj, line 889: Just for future reference, but this should also work:
```suggestion
        auto local_isect = visit_universe(
            [&local_state, &isect, &local_isect](auto&& t) {
                return t.intersect(local_state, isect.distance);
            },
            lsa.universe());
```

PR #938, opened 2023-09-14
Correctly map HepMC3 events to Geant4 events

### LINKED ISSUE #643: HepMC3PrimaryGenerator generates out-of-order events
OK, I figured out what was going on. There's a mismatch between the HepMC3 GenEvent number (1-based) and the Geant4 Event ID (0-based).  Furthermore, because we read the HepMC3 events successively in time, rather than corresponding to the Geant4 event partitioning, the event IDs may not match the Geant4 ID.

Printing both the event ID and the HepMC3 event ID, I see this output from all the different worker threads:
```
[1/2] info: Read 100 primaries from HepMC event ID 1 for Geant4 event 2
[0/2] info: Read 100 primaries from HepMC event ID 2 for Geant4 event 0
[1/2] info: Read 100 primaries from HepMC event ID 3 for Geant4 event 3
[0/2] info: Read 100 primaries from HepMC event ID 4 for Geant4 event 1
[1/2] info: Read 100 primaries from HepMC event ID 5 for Geant4 event 4
[0/2] info: Read 100 primaries from HepMC event ID 6 for Geant4 event 6
[0/2] info: Read 100 primaries from HepMC event ID 7 for Geant4 event 7
[0/2] info: Read 100 primaries from HepMC event ID 8 for Geant4 event 8
[1/2] info: Read 100 primaries from HepMC event ID 9 for Geant4 event 5
[1/2] info: Read 100 primaries from HepMC event ID 10 for Geant4 event 10
[1/2] info: Read 100 primaries from HepMC event ID 11 for Geant4 event 11
[0/2] info: Read   8 primaries from HepMC event ID 12 for Geant4 event 9
```
The HepMC3 events are linear and 1-based, but they don't correspond to the Geant4 event IDs because of multithreading.

_Originally posted by @sethrj in https://github.com/celeritas-project/celeritas/pull/637#discussion_r1102846582_

sethrj:
This fixes #643 by reading consecutive events into a buffer (pushing them back onto a double-ended queue) until the one requested by the local thread is obtained. Subsequent threads that request earlier events remove them from the middle of the deque, popping any already-read events off the front. This minimizes the memory usage of the events to be read.

Even though the unit test is single-threaded, I've verified the code works in celer-g4:
```
[2/2] debug: Reading to event 2: buffer has [0, 0)
[1/2] debug: Reading to event 0: buffer has [0, 3)
[2/2] debug: Reading to event 3: buffer has [1, 3)
[1/2] debug: Reading to event 1: buffer has [1, 4)
[2/2] debug: Reading to event 4: buffer has [4, 4)
[1/2] debug: Reading to event 6: buffer has [5, 5)
...
```

### DIFF
--- a/src/accel/HepMC3PrimaryGenerator.cc
+++ b/src/accel/HepMC3PrimaryGenerator.cc
@@ -202,8 +202,17 @@ void HepMC3PrimaryGenerator::GeneratePrimaryVertex(G4Event* g4_event)
  * Read the given event from the file in a thread-safe manner.
  *
  * Each event can only be read once. Because reading across threads may be out
- * of order, the next event to read may not be the next event in the file. Read
- * all the events into
+ * of order, the next event to read may not be the next event in the file. To
+ * fix this with minimal performance and memory impact, we read all events up
+ * to the one requested into a buffer. Once the events are buffered, we release
+ * the shared pointer (marking its location in the buffer as empty) and return
+ * it to the calling thread. Before reading new events, empty elements at the
+ * front of the buffer are released. In the usual case, the buffer should only
+ * be size(num_threads), but in the worst case (the first event is very slow
+ * and the other threads keep processing new events) it can be arbitrarily
+ * large. However, since accessing an element in a deque is a constant-time
+ * operation, this function should be constant time at best and scale with the
+ * number of threads at worst.
  */
 auto HepMC3PrimaryGenerator::read_event(size_type event_id) -> SPHepEvt
 {
--- a/src/accel/HepMC3PrimaryGenerator.cc
+++ b/src/accel/HepMC3PrimaryGenerator.cc
@@ -114,7 +114,7 @@ HepMC3PrimaryGenerator::HepMC3PrimaryGenerator(std::string const& filename)
     num_events_ = [&filename] {
         SPReader temp_reader = open_hepmc3(filename);
         CELER_ASSERT(temp_reader);
-        int result = 0;
+        size_type result = 0;
         while (!temp_reader->failed())
         {
             temp_reader->skip(1);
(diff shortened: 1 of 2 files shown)


src/accel/HepMC3PrimaryGenerator.cc
```diff
@@ -212,10 +192,73 @@ void HepMC3PrimaryGenerator::GeneratePrimaryVertex(G4Event* g4_event)
...
+
+        auto read_evt_id = event_buffer_.back()->event_number();
+        CELER_VALIDATE(!reader_->failed(),
+                       << "event " << expected_id << " could not be read");
+
+        if (static_cast<size_type>(read_evt_id) != expected_id)
+        {
+            CELER_LOG(warning)
+                << "HepMC3 event IDs are not consecutive from zero: ID "
+                << read_evt_id << " is not expected ID " << expected_id;
```
amandalund, line 250: This seems like a reasonable assumption, but are we sure this is the case (and does it matter)? I haven't e.g. seen anything from HepMC3 suggesting the event IDs couldn't just be arbitrary (and some of the event files we use, like the one for our celer-g4 test, start at 1).

amandalund (approved):
Looks good @sethrj!

src/accel/HepMC3PrimaryGenerator.cc
sethrj: Geant4 starts at zero, so if we don't have that event do we just die immediately on the worker thread? Or read until the end of the file? What if the file has multiple events with the same ID? Bad planning on hepmc's part...

amandalund: Oh is the problem that we need to preserve the original IDs from the event record because they'll be needed at other stages in the simulation chain? I guess I'm missing why it matters if the events get assigned new IDs by Geant4. What do we do if some event records start at 0 and others at 1?

sethrj: That's why here were basically pretending the event ids on the file are sequential from zero. Before we pretended that *and* gave the event to the first core that asked for it.

whokion: In Geant4, the event ID is assigned by the number of events processed and it may not be necessarily same to one assigned by an external generator (or an event ID record, if any, in an input event file) - so it does not matter whether the EventID starts in 0 or 1, or nothing in input.  Only important thing may be the order of event-reading  or processing the sequence of events by the Geant4 event manager for `reproducibility`, especially for the MT mode - there are two different cases: one for the full reproducibility regardless of the number threads used (i.e, seeds are assigned in the run-level), another one for the reproducibility when the same number of threads is used (i.e., seeds are assigned in the thread-level).

amandalund: Thanks, yeah that makes sense. Since we're warning if the HepMC3 ID doesn't match the Geant4 ID I was wondering if there were any reason we might need to preserve the former, but it sounds like no (or at least like it would be trivial to create a mapping now that the ID sequence matches the order in the event file).

PR #939, opened 2023-09-14
Implement CSG tree simplification algorithms

sethrj:
This adds logic for replacing nodes in a CSG tree with a constant value, then propagating the implications of that down and up the tree. This is a substantial improvement of the simplification logic in SCALE, which still leaves unnecessary constants in the tree and cannot combine multiple nodes.

### DIFF
--- a/src/orange/construct/detail/NodeReplacementInserter.hh
+++ b/src/orange/construct/detail/NodeReplacementInserter.hh
@@ -33,20 +33,20 @@ class NodeReplacementInserter
     //!@}
 
   public:
-    // Construct with references to the stack and the replacement value
+    // Construct with pointers to the stack and replacement
     inline NodeReplacementInserter(VecNode* stack, Node const& repl);
-
-    // If replacing a queued node with a boolean, ours should match
-    inline void operator()(True const&);
-    inline void operator()(False const&);
-
-    // Surfaces cannot be simplified further
-    void operator()(Surface const&) {}
 
     // Simplify node types that reference other nodes
     inline void operator()(Aliased const& n);
     inline void operator()(Negated const& n);
     inline void operator()(Joined const& n);
+
+    // Check that replacement matches our stored boolean
+    inline void operator()(True const&);
+    inline void operator()(False const&);
+
+    // Surfaces cannot be simplified further
+    void operator()(Surface const&) {}
 
   private:
     VecNode* stack_;
@@ -59,7 +59,7 @@ class NodeReplacementInserter
 // INLINE DEFINITIONS
 //---------------------------------------------------------------------------//
 /*!
- * Construct with references to the stack and the replacement value.
+ * Construct with pointer to the stack and the replacement value.
  *
  * For now the replacement must be a boolean.
  */
@@ -85,7 +85,7 @@ NodeReplacementInserter::NodeReplacementInserter(VecNode* stack,
 
 //---------------------------------------------------------------------------//
 /*!
- * If replacing a queued node with a boolean, ours should match.
+ * Check that the replacement node matches this queued node.
  */
 void NodeReplacementInserter::operator()(True const&)
 {
@@ -94,7 +94,7 @@ void NodeReplacementInserter::operator()(True const&)
 
 //---------------------------------------------------------------------------//
 /*!
- * If replacing a queued node with a boolean, ours should match.
+ * Check that the replacement node matches this queued node.
  */
 void NodeReplacementInserter::operator()(False const&)
 {
@@ -103,6 +103,8 @@ void NodeReplacementInserter::operator()(False const&)
 
 //---------------------------------------------------------------------------//
 /*!
+ * Push the target of an aliased node onto the stack.
+ *
  * Aliasing a node implies the alias has the same value.
  */
 void NodeReplacementInserter::operator()(Aliased const& n)
@@ -112,6 +114,8 @@ void NodeReplacementInserter::operator()(Aliased const& n)
 
 //---------------------------------------------------------------------------//
 /*!
+ * Push a negated node onto the stack.
+ *
  * Negating a node implies its daughter has the opposite value.
  */
 void NodeReplacementInserter::operator()(Negated const& n)
--- a/src/orange/construct/detail/NodeReplacementInserter.hh
+++ b/src/orange/construct/detail/NodeReplacementInserter.hh
@@ -1,0 +1,142 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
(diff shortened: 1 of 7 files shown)


sethrj:
@elliottbiondo Don't forget about this one, please!

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+    //! \name Type aliases
+    using VecLogic = std::vector<logic_int>;
+    //!@}
+
+    static_assert(std::is_same_v<LocalSurfaceId::size_type, logic_int>,
+                  "unsupported: add enum logic conversion for different-sized "
+                  "face and surface ints");
+
+  public:
+    // Construct with reference to vector to append to
```
elliottbiondo: pointer

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+    explicit inline PostfixLogicBuilder(CsgTree const& tree, VecLogic* logic);
+
+    //! Build from a node ID
+    inline void operator()(NodeId const& n);
+
+    // Visit an actual surface
```
elliottbiondo: actual?

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+    inline void operator()(True const&);
```
elliottbiondo, line 49: Add doc strings

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+ */
+void PostfixLogicBuilder::operator()(Aliased const& n)
+{
+    // CELER_ASSERT_UNREACHABLE();
+    return (*this)(n.node);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * If replacing a queued node with a boolean, ours should match.
```
elliottbiondo: Add doc strings, current doc string as detail

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+ */
+void PostfixLogicBuilder::operator()(Negated const& n)
+{
+    (*this)(n.node);
+    logic_->push_back(logic::lnot);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * If replacing a queued node with a boolean, ours should match.
```
elliottbiondo: Same as above

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+
+#include "../CsgTree.hh"
+
+namespace celeritas
+{
+namespace csg
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Recursively construct a logic vector from a node with postfix operation.
```
elliottbiondo, line 21: This class could use some additional high-level documentation. While I that how the `ContainerVisitor` is calling the operator() functions, and that how `PostfixLogicBuilder::operator()(NodeId const& n)` provides recursion, this is pretty hard to follow.

src/orange/construct/detail/PostfixLogicBuilder.hh
```diff
@@ -0,0 +1,149 @@
...
+    CELER_EXPECT(logic_);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build from a node ID.
+ */
+void PostfixLogicBuilder::operator()(NodeId const& n)
+{
+    return this->visit_node_(*this, n);
```
elliottbiondo: Remove return here and in all other operator() functions

src/orange/construct/detail/NodeReplacementInserter.hh
```diff
@@ -0,0 +1,142 @@
...
+class NodeReplacementInserter
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using VecNode = std::vector<std::pair<NodeId, Node>>;
+    //!@}
+
+  public:
+    // Construct with references to the stack and the replacement value
```
elliottbiondo: pointer to the stack and reference to replacement value

src/orange/construct/detail/NodeReplacementInserter.hh
```diff
@@ -0,0 +1,142 @@
...
+    inline NodeReplacementInserter(VecNode* stack, Node const& repl);
+
+    // If replacing a queued node with a boolean, ours should match
```
elliottbiondo: This comment doesn't make sense without the context that the function itself has a DBC check. I recommend adding standard doc strings to these functions.

src/orange/construct/detail/NodeReplacementInserter.hh
```diff
@@ -0,0 +1,142 @@
...
+    else
+    {
+        negated_ = Node{True{}};
+        join_token_ = op_or;  // any{...} = false
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * If replacing a queued node with a boolean, ours should match.
```
elliottbiondo: Add standard doc string here and throughout

src/orange/construct/detail/NodeReplacementInserter.hh
```diff
@@ -0,0 +1,142 @@
...
+    Node const& repl_;
+    Node negated_;
+    OperatorToken join_token_;
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with references to the stack and the replacement value.
```
elliottbiondo: pointer

elliottbiondo (changes requested):

src/orange/construct/detail/PostfixLogicBuilder.hh
sethrj: Sometimes this is a good thing to do (it ensures that the function you're calling actually returns `void` so you're not discarding the result), but you're right it doesn't add anything here.

src/orange/construct/detail/NodeReplacementInserter.hh
sethrj: In the doc strings I think pointer and reference can be effectively interchangeable. The type itself is clear from the code, the point of the documentation is to emphasize that you're referencing memory outside the class itself.

src/orange/construct/detail/PostfixLogicBuilder.hh
elliottbiondo: The confusing part was really the inconsistency: some functions had returns and other didn't

src/orange/construct/detail/NodeReplacementInserter.hh
elliottbiondo: I agree that you shouldn't have to mention the type in a doc string, but if you are going to mention a type, it should be the correct type.

elliottbiondo (approved):

PR #941, opened 2023-09-17
Define unified tolerances class for ORANGE

sethrj:
This adds a single class responsible for setting relative and absolute error uniformly over ORANGE. The absolute error is set using a "length scale" (defaulting to 1, aka 1 cm when used by Shift and Celeritas) times the relative error. The default relative error is roughly the square root of machine precision: 1e-8 for double precision, 5e-4 for single precision.

I've also added a derivation for a reasonable rule for the approximating the second-order quadratic term as zero for near-parallel tracks: the previous default isn't quite consistent with the rest of the tolerances but I'm hesitant to change it at the moment.

I unified the bounding box, bumping, and quadric tolerances to all use this class.

### DIFF
--- a/src/orange/OrangeTypes.cc
+++ b/src/orange/OrangeTypes.cc
@@ -37,7 +37,7 @@ Tolerance<T> Tolerance<T>::from_default(real_type length)
         }
     }();
     static_assert(real_type{1} - ipow<2>(sqrt_emach) != real_type{1},
-                  "default tolerance is insufficient");
+                  "default tolerance is too low");
 
     return Tolerance<T>::from_relative(sqrt_emach, length);
 }
@@ -69,6 +69,10 @@ Tolerance<T> Tolerance<T>::from_relative(real_type rel, real_type length)
     Tolerance<T> result;
     result.rel = rel;
     result.abs = rel * length;
+
+    // Check that the tolerances aren't so low that they're denormalized
+    CELER_ENSURE(result.rel >= std::numeric_limits<T>::min());
+    CELER_ENSURE(result.abs >= std::numeric_limits<T>::min());
     CELER_ENSURE(result);
     return result;
 }
--- a/src/orange/OrangeTypes.cc
+++ b/src/orange/OrangeTypes.cc
@@ -7,10 +7,69 @@
 //---------------------------------------------------------------------------//
 #include "OrangeTypes.hh"
 
+#include <type_traits>
+
+#include "corecel/Assert.hh"
 #include "corecel/io/EnumStringMapper.hh"
+#include "corecel/math/Algorithms.hh"
+#include "corecel/math/SoftEqual.hh"
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Use a relative error of \f$ \sqrt(\epsilon_\textrm{machine}) \f$ .
+ *
+ * Technically we're rounding the machine epsilon to a nearby power of 10. We
+ * could use numeric_limits<real_type>::epsilon instead.
+ */
+Tolerances Tolerances::from_default(real_type length)
+{
+    constexpr real_type sqrt_emach = [] {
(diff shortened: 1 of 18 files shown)


src/orange/detail/UnitInserter.cc
```diff
@@ -150,11 +149,17 @@ SimpleUnitId UnitInserter::operator()(UnitInput const& inp)
...
+    // potential bump distance from a boundary so that the bbox will enclose
+    // the point even after a potential bump
+    BoundingBoxBumper<fast_real_type> calc_bumped{
+        [&tol = orange_data_->scalars.tol] {
+            Tolerance<real_type> bbox_tol;
+            bbox_tol.rel = 2 * tol.rel;
+            bbox_tol.abs = 2 * tol.abs;
+            CELER_ENSURE(bbox_tol);
+            return bbox_tol;
+        }()};
```
elliottbiondo, line 162: The `()` at the end is very subtle. Even though this creates an extra local variable, I think the basic way is much more readable

```
BoundingBoxBumper<fast_real_type> calc_bumped{bbox_tol};
CELER_ENSURE(bbox_tol);
Tolerance<real_type> bbox_tol{2*tot.rel, 2*tol.abs};
```

src/orange/OrangeTypes.cc
```diff
@@ -7,10 +7,72 @@
...
+        {
+            return 1.e-8;
+        }
+        else if constexpr (std::is_same_v<real_type, float>)
+        {
+            return 5.e-3f;
+        }
+    }();
+    static_assert(real_type{1} - ipow<2>(sqrt_emach) != real_type{1},
+                  "default tolerance is insufficient");
```
elliottbiondo: "too low"

src/orange/OrangeTypes.cc
```diff
@@ -7,10 +7,72 @@
...
+{
+    CELER_VALIDATE(rel > 0 && rel < 1,
+                   << "tolerance " << rel
+                   << " is out of range [must be in (0,1)]");
+    CELER_VALIDATE(length > 0,
+                   << "length scale " << length
+                   << " is invalid [must be positive]");
+    Tolerance<T> result;
+    result.rel = rel;
+    result.abs = rel * length;
```
elliottbiondo, line 71: Add a DBC  to ensure that the resulting absolute tolerance isn't below floating point epsilon

elliottbiondo (approved):

src/orange/OrangeTypes.cc
sethrj: The DBC ensures both are nonzero. "floating point epsilon" is a measure of *relative* accuracy since it's based on the change of magnitude by incrementing the mantissa by one bit, so it doesn't make sense to require any relationship between the absolute tolerance and epsilon. If we're doing transport on nanoscale materials, it's fine to have an absolute tolerance of 1e-17 since it'll be added to quantities that are `O(1e-7)`.

src/orange/detail/UnitInserter.cc
sethrj: I specifically don't want to do `Tolerance<real_type> bbox_tol{2*tot.rel, 2*tol.abs}` because that assumes the ordering of `rel` and `abs` in the struct, which could introduce a subtle bug. Once we move to C++20 and can do [designated initializers](https://www.cppstories.com/2021/designated-init-cpp20/), we can write this whole thing as
```c++
    BoundingBoxBumper<fast_real_type> calc_bumped{
        {.rel = 2 * tol.rel, .abs = 2 * tol.abs}};
```
Until then we've agreed as a team to our use of [IIFE](https://www.cppstories.com/2016/11/iife-for-complex-initialization/) even though it's not the prettiest thing in the world. (We talked about using `std::invoke`, but that isn't compatible with GPU.)

src/orange/OrangeTypes.cc
sethrj: I will put in a check that they're not denormalized.

PR #942, opened 2023-09-17
Fix Geant4 field construction in celer-g4

amandalund:
This fixes two issues with the field in the celer-g4 app:

- The field in Geant4 was not being constructed if no SDs were present.
- It looks like having a zero-strength field is not the same thing as having no field in Geant4 (I think the field propagator will still be used in the former case). Previously we were constructing the `G4MagneticField` even if the field magnitude was zero.

### DIFF
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -138,9 +138,9 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
         {
             CELER_LOG_LOCAL(info)
                 << "Using a uniform field (0, 0, " << field[2] << ") in tesla";
+            mag_field_ = std::make_shared<G4UniformMagField>(
+                convert_to_geant(field, CLHEP::tesla));
         }
-        mag_field_ = std::make_shared<G4UniformMagField>(
-            convert_to_geant(field, CLHEP::tesla));
 
         // Convert field units from tesla to native celeritas units
         for (real_type& v : field)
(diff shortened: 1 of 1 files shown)


sethrj (approved):
Nice catch @amandalund ! Is it correct then that our calculated speedup for "no magnetic field" will be worse (before, we were comparing celeritas linear propagator with geant4 field propagator), but our relative performance will be much better for the "no SD" case (because previously Geant4 had no magnetic field at all)?

amandalund:
Yep, that's what I'd expect.

sethrj:
I hope @whokion hasn't completed his rerun yet 😉

whokion:
@amandalund, @sethrj Sorry that I did not see/review this MR. 

> This fixes two issues with the field in the celer-g4 app:

I do not think that this change will affect anything as the zero-field equivalence in Geant4 is based on the filed manager ptr, not on a magnetic field ptr.
 
>     * The field in Geant4 was not being constructed if no SDs were present.

Field and SD are two independent things (even though they can be added separately inside the common interface), so I do not understand this statement.

>     * It looks like having a zero-strength field is not the same thing as having no field in Geant4 (I think the field propagator will still be used in the former case). Previously we were constructing the `G4MagneticField` even if the field magnitude was zero.

I thought that both celeritas and geant4 use the field propagator if the field type is uniform and the field strength is zero, so we are still comparing equivalent works for performance.  If we want to force not to use the field at all for the zero-field case, then we need to introduce field_type="none" or "no_field" explicitly.

amandalund:
> Field and SD are two independent things (even though they can be added separately inside the common interface), so I do not understand this statement.

@whokion the problem was we were returning immediately in `DetectorConstruction::ConstructSDandField()` if there were no SDs, and since the field was constructed at the end of the method it was skipped.

> I thought that both celeritas and geant4 use the field propagator if the field type is uniform and the field strength is zero, so we are still comparing equivalent works for performance.

[Celeritas](https://github.com/celeritas-project/celeritas/blob/2780ceaa7a68c0bb3729137f04f462ba55c54cfc/src/accel/AlongStepFactory.cc#L70) will use a linear propagator if the field strength is zero, while it looks to me like the [Geant4](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/processes/transportation/src/G4Transportation.cc#L260) transportation process checks if the `G4Field` pointer is null (rather than checking if the field strength is zero) and if not uses the field propagator.

whokion:
Okay,  it was a definitely bug (did not compare old/new codes).  For the second bullet, I did not know that celeritas uses the linear propagator when the field strength is zero - can you point where we do this in the along step action?

sethrj:
@whokion Yep it's done in the `accel` front end code: see https://github.com/celeritas-project/celeritas/blob/2780ceaa7a68c0bb3729137f04f462ba55c54cfc/src/accel/AlongStepFactory.cc#L70-L80

PR #943, opened 2023-09-17
Add option to disable SD hit collection in celer-g4

amandalund:


sethrj:
You're too modest with these tags 😉

sethrj (approved):

sethrj (approved):

PR #944, opened 2023-09-18
Add timers to celer-g4

amandalund:
This adds timers for setup, total, event, and accumulated action time in celer-g4 and outputs the results to JSON.

### DIFF
--- a/app/celer-g4/TimerOutput.cc
+++ b/app/celer-g4/TimerOutput.cc
@@ -46,7 +46,6 @@ void TimerOutput::output(JsonPimpl* j) const
         {"actions", action_time_},
         {"events", event_time_},
         {"total", total_time_},
-        //{"setup", result_.setup_time},
     };
 
     j->obj = std::move(obj);
@@ -57,7 +56,7 @@ void TimerOutput::output(JsonPimpl* j) const
 
 //---------------------------------------------------------------------------//
 /*!
- * Record the total time for the run.
+ * Record the total time spent in transport and hit I/O (excluding setup).
  *
  * This should be called once by the master thread.
  */
--- a/app/celer-g4/TimerOutput.cc
+++ b/app/celer-g4/TimerOutput.cc
@@ -46,23 +46,13 @@ void TimerOutput::output(JsonPimpl* j) const
         {"actions", action_time_},
         {"events", event_time_},
         {"total", total_time_},
+        {"setup", setup_time_},
     };
 
     j->obj = std::move(obj);
 #else
     (void)sizeof(j);
 #endif
-}
-
-//---------------------------------------------------------------------------//
-/*!
- * Record the total time spent in transport and hit I/O (excluding setup).
- *
- * This should be called once by the master thread.
- */
-void TimerOutput::RecordTotalTime(real_type time)
-{
-    total_time_ = time;
 }
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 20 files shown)


app/celer-g4/TimerOutput.hh
```diff
@@ -0,0 +1,70 @@
...
+#include "corecel/Types.hh"
+#include "corecel/io/OutputInterface.hh"
+
+namespace celeritas
+{
+namespace app
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Record the total, setup, event, and accumulated action time.
```
sethrj: I think the comment about multithreaded mode ignoring most of the timers should go here. Because most of our relevant GPU timing is going to be with multiple Geant4 threads sharing a single GPU, we really should think about how to add this capability: the easiest thing to do would be to have a `thread_local` output registry that gets written for each thread to a different JSON file.

src/accel/LocalTransporter.cc
```diff
@@ -200,6 +203,35 @@ void LocalTransporter::Finalize()
...
+ */
+auto LocalTransporter::GetActionTime() const -> MapStrReal
+{
+    CELER_EXPECT(*this);
+
+    MapStrReal result;
+    auto const& action_seq = step_->actions();
+    if (num_streams_ == 1
+        && (action_seq.sync()
+            || !celeritas::getenv("CELER_DISABLE_DEVICE").empty()))
```
sethrj: If I'm following the triple negative: can't this be
```suggestion
            || !celeritas::device()))
```
? The device should be set if and only if we're running the celeritas offload on device.

sethrj (approved):
Nice refactor and great capability! I think adding some form of MT timing output is important but can be done in a follow-on commit.

src/accel/LocalTransporter.cc
amandalund: Yep that should work, much easier to read.

app/celer-g4/TimerOutput.hh
amandalund: What kind of MT timing output did you have in mind? I wasn't sure if the wall time per event or for the actions would be very useful when we're running the events in parallel.

sethrj: I don't see why they wouldn't be? If we're scaling perfectly, the wall time per event will be the same whether we run 1 thread or 32. (And for the latter, we'll get our timing results faster.) If we're not, and one particular action is misbehaving (e.g. having a mutex in the detector conversion), then the time per actions will help us out in a way that we won't be able to see in single-threaded mode.

amandalund: Ok, in that case it shouldn't be too hard to get the event and action times in MT mode -- I can do it on this PR if you want.

sethrj: That sounds great! Thanks.

sethrj (approved):
Nice! Thanks.

app/celer-g4/TimerOutput.cc
```diff
@@ -0,0 +1,106 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2022-2023 UT-Battelle, LLC, and other Celeritas developers.
```
whokion, line 2:
```suggestion
// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
```

app/celer-g4/TimerOutput.hh
```diff
@@ -0,0 +1,73 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2022-2023 UT-Battelle, LLC, and other Celeritas developers.
```
whokion, line 2:
```suggestion
// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
```

whokion (approved):
This MR looks great!  One general question: Should I expect that the sum of event_time is close to the `user` time? - i.e., the sum of event_time is equal to the total time only for the single thread run.

amandalund:
Thanks @whokion, yes, the sum of the event times should be close to the user time (unless i guess if we're running with `CELER_DISABLE_DEVICE` and OpenMP enabled 😬), and equal to or a bit less than the total time with a single thread (the final merging/closing of the ROOT file for the hit I/O is only included in the total time).

app/celer-g4/TimerOutput.cc
amandalund: Oops, copy-paste error. We can tack these on to another PR.

sethrj:
Crap, I forgot to wait for @whokion 's review! Sorry 🤪

PR #945, opened 2023-09-19
Simplify nearly-zero surface displacements to zero

sethrj:
This fixes a couple of unit tests in the SCALE surface branch because previously we were snapping nearly-zero displacements to zero, and I neglected to add that to the Celeritas surface simplifier.


sethrj:
Ping @elliottbiondo . This ones's super short.

sethrj:
Actually I'm going to add snapping of cones to the origin.

elliottbiondo (approved):

PR #946, opened 2023-09-19
Use async allocation for temporary track sorting storage.

esseivaju:
Replace the temporary `DeviceVector` with a direct call to CUDA/HIP API until we implement a stream-aware Collection. Alternatively, could we allocate this storage once, hold a reference to it in the state and reuse it?


sethrj (approved):
Lgtm. Avoiding the reallocation can be done in a separate PR to better measure performance delta. How are we looking with the sort kernels these days? I really want to try this on AMD...

esseivaju:
The overhead should be minimal now that everything is asynchronous and stream-ordered with radix sort. I'll get some profiling again from celer-sim. I also got access to a system with AMD gpus (Instinct MI100) that I could try it on.

PR #947, opened 2023-09-20
Update roles to distinguish "core advisor"

sethrj:
The motivation is to make more clear to external folks that the advisors aren't typically working on celeritas development.


paulromano (approved):
Looks good to me!

pcanal (approved):

PR #856, opened 2023-07-14
Fix linking errors with CUDA+VecGeom caused by #847

pcanal:
Add explicit link dependency from `${target}_final` to `${target_static}`. This allows the use of `CMAKE_LINK_DEPENDS_NO_SHARED=OFF` and still get a new linking of the final library (i.e. executing `nvlink`) if any of the RDC cuda file have changed (and thus the internal CUDA RDC infrastructure symbol have change ; they are seeded with a 'random' number which change at each compilation).

PR #948, opened 2023-09-20
Fix link errors during repeated builds when CUDA RDC is enabled

pcanal:
Add explicit link dependency from ${target}_final to ${target_static}/

This allows the use of `CMAKE_LINK_DEPENDS_NO_SHARED=OFF` and still get a new linking of the final library (i.e. executing `nvlink`) if any of the RDC cuda file have changed (and thus the internal CUDA RDC infrastructure symbol have change ; they are seeded with a 'random' number which change at each compilation).

### DIFF
--- a/app/celer-g4/CMakeLists.txt
+++ b/app/celer-g4/CMakeLists.txt
@@ -54,11 +54,16 @@ if(CELERITAS_USE_ROOT AND CELERITAS_USE_Geant4)
     )
   endfunction(celeritas_local_root_generate_dictionary)
   celeritas_local_root_generate_dictionary(HitClassesRootInterfaces)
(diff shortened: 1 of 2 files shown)


sethrj:
@pcanal Will this fix https://github.com/celeritas-project/celeritas/pull/856#issuecomment-1665854068 ? Also was this intentionally closed?

pcanal:
Yes, it was intentionally closed as it was inadvertently opened with the wrong commits (The real PR is https://github.com/celeritas-project/celeritas/pull/949).  

But I need to double check the stray commits are still somewhere and check on https://github.com/celeritas-project/celeritas/pull/856#issuecomment-1665854068 which might have been overlooked over the summer breaks.

PR #950, opened 2023-09-20
Fix build errors on macOS

stognini:
This PR fixes two build errors on macOS:

```
test/testdetail/TestMacrosImpl.cc:19:14: fatal error: 'nlohmann/json.hpp' file not found
#    include <nlohmann/json.hpp>
             ^~~~~~~~~~~~~~~~~~~
```

```
app/celer-g4/TimerOutput.hh:34:29: error: no template named 'unordered_map' in namespace 'std'
    using MapStrReal = std::unordered_map<std::string, real_type>;
                       ~~~~~^
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -23,22 +23,13 @@ celeritas_add_library(testcel_harness
 target_compile_features(testcel_harness PUBLIC cxx_std_14)
 celeritas_target_link_libraries(testcel_harness
   PUBLIC Celeritas::corecel GTest::GTest
+  PRIVATE ${nlohmann_json_LIBRARIES}
 )
 celeritas_target_include_directories(testcel_harness
   PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
 )
-
-if (CELERITAS_USE_JSON)
-  celeritas_target_include_directories(testcel_harness
-    PUBLIC
-      $<BUILD_INTERFACE:${NLOHMANN_INCLUDE_DIR}>
-  )
-  celeritas_target_link_libraries(testcel_harness
-    PUBLIC nlohmann_json::nlohmann_json
-  )
-endif()
 
 #-----------------------------------------------------------------------------#
 # HIP SUPPORT
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,6 +30,16 @@ celeritas_target_include_directories(testcel_harness
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
 )
 
+if (CELERITAS_USE_JSON)
+  celeritas_target_include_directories(testcel_harness
+    PUBLIC
+      $<BUILD_INTERFACE:${NLOHMANN_INCLUDE_DIR}>
+  )
+  celeritas_target_link_libraries(testcel_harness
+    PUBLIC nlohmann_json::nlohmann_json
+  )
+endif()
+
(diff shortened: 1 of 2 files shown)


pcanal (approved):

test/CMakeLists.txt
```diff
@@ -30,6 +30,16 @@ celeritas_target_include_directories(testcel_harness
...
 
+if (CELERITAS_USE_JSON)
+  celeritas_target_include_directories(testcel_harness
+    PUBLIC
+      $<BUILD_INTERFACE:${NLOHMANN_INCLUDE_DIR}>
+  )
+  celeritas_target_link_libraries(testcel_harness
+    PUBLIC nlohmann_json::nlohmann_json
+  )
+endif()
```
sethrj: - The `include_directories` isn't necessary because the modern cmake targets propagate that information
- This should be PRIVATE not PUBLIC because it's only used in a .cc file
- `if (` -> `if(`
- Our top-level cmakelists defines `nlohmann_json_LIBRARIES` so that it's empty if json isn't available, so that makes it easier to add for the optional cases
- It turns out I'm not *directly* using JSON anyway for now, so we could've just deleted the include 😅

test/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ celeritas_add_library(testcel_harness
 target_compile_features(testcel_harness PUBLIC cxx_std_14)
 celeritas_target_link_libraries(testcel_harness
   PUBLIC Celeritas::corecel GTest::GTest
+  PRIVATE ${nlohmann_json_LIBRARIES}
```
pcanal, line 26: Is it really okay to have `PRIVATE` followed by nothing?
What if later on somebody add something after:
```suggestion
  PRIVATE ${nlohmann_json_LIBRARIES}
  OTHERTAG ${othervalues}
```
?

sethrj: I tested before pushing and it does work as expected.

test/CMakeLists.txt
stognini: Wait, so `target_link_libraries(...)` sets the include dirs as well?

sethrj: Exactly. Modern cmake targets propagate build flags, library flags, include directories, etc.

sethrj (approved):

test/CMakeLists.txt
sethrj: Also besides public and private what other tags go there?

pcanal: Beside the many legacy flags there is `INTERFACE` and then there is the new one that will be introduce in cmake v5.22 in a few years :).

> I tested before pushing and it does work as expected.

So you tested something like:
```
celeritas_target_link_libraries(testcel_harness
  PRIVATE ${nlohmann_json_LIBRARIES}
  PUBLIC Celeritas::corecel GTest::GTest
)  
```
or even
```
celeritas_target_link_libraries(testcel_harness
  PRIVATE 
  PUBLIC Celeritas::corecel GTest::GTest
)  
```
right?

sethrj: I didn't flip the lines, I tested the code that's in there. With JSON unavailable, the current code works.

pcanal: Yes, I expect so.  There is nothing to be confused with.  But if you flip the lines or Seth 2030 add a new tag afterwards, then it might (hopefully not) be a different story.

sethrj: Seth 2030 will probably see that there's a possibly-empty variable beforehand, and Philippe 2030 will give it a thorough code review 😉

pcanal: Fair enough ...

pcanal (approved):

PR #774, opened 2023-05-25
Support multithreaded CPU using single GPU in demo loop

amandalund:
This adds OpenMP multithreading to a loop over events in the demo loop, with each event processed by a separate thread and running on a single GPU (see #553). For multithreaded CPU-only runs, nested parallel regions are disabled until we can understand why the performance there is so poor. The GPU performance is definitely not as good as running all events simultaneously on a single thread (about a factor of two slower for cms2018+field+msc-vecgeom-gpu), but we should get some of that back when we're able to launch kernels on different CUDA streams.

PR #951, opened 2023-09-21
Fix accumulated action times in celer-sim

amandalund:
This fixes a problem with the accumulated action times in celer-sim (from #774) when running with a single stream and `merge_events = false`. Fortunately our regression timing results weren't affected by this since we merge the events.


sethrj (approved):
Great! Nice catch, thanks @amandalund .

PR #952, opened 2023-09-21
Avoid maybe_unused attribute because of old GCC

sethrj:
Several GCC versions that aren't used by our CI warn whenever a "maybe unused" variable is actually used, which is stupid because the point of the decorator is to disable warnings. This PR defines a new macro `CELER_NEVER_UNUSED(EXPRESSION)` that can be used to force the compiler to parse but not evaluate the given expression. It replaces all the `(void)sizeof(expr)` as well to improve readability.

### DIFF
--- a/src/corecel/Macros.hh
+++ b/src/corecel/Macros.hh
@@ -231,6 +231,6 @@
  * The argument is an unevaluated operand which will generate no code but force
  * the expression to be used.
  */
-#define CELER_NEVER_UNUSED(CODE) static_cast<void>(sizeof(CODE))
+#define CELER_NEVER_UNUSED(CODE) static_cast<void>(sizeof(CODE));
 
 //---------------------------------------------------------------------------//
--- a/src/corecel/Macros.hh
+++ b/src/corecel/Macros.hh
@@ -225,4 +225,12 @@
     CLS(CLS&&) = delete;                 \
     CLS& operator=(CLS&&) = delete
 
+/*!
+ * \def CELER_MARK_USED
+ *
+ * The argument is an unevaluated operand which will generate no code but force
+ * the expression to be used.
+ */
+#define CELER_NEVER_UNUSED(CODE) static_cast<void>(sizeof(CODE))
+
 //---------------------------------------------------------------------------//
--- a/src/corecel/Macros.hh
+++ b/src/corecel/Macros.hh
@@ -231,6 +231,6 @@
  * The argument is an unevaluated operand which will generate no code but force
  * the expression to be used.
  */
-#define CELER_NEVER_UNUSED(CODE) static_cast<void>(sizeof(CODE));
+#define CELER_DISCARD(CODE) static_cast<void>(sizeof(CODE));
 
 //---------------------------------------------------------------------------//
--- a/src/corecel/Macros.hh
+++ b/src/corecel/Macros.hh
@@ -225,16 +225,4 @@
     CLS(CLS&&) = delete;                 \
     CLS& operator=(CLS&&) = delete
 
-/*!
- * \def CELER_UNUSED_UNLESS_DEVICE
- *
- * GCC 8 and 9 are known to print a warning for "maybe unused" variables that
- * are actually used (after preprocessor logic).
- */
-#if CELER_USE_DEVICE
-#    define CELER_UNUSED_UNLESS_DEVICE
-#else
-#    define CELER_UNUSED_UNLESS_DEVICE [[maybe_unused]]
-#endif
-
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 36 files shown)


src/corecel/data/Copier.cc
```diff
@@ -62,13 +63,14 @@ void copy_bytes(MemSpace dstmem,
                 MemSpace srcmem,
                 void const* src,
                 std::size_t count,
-                CELER_UNUSED_UNLESS_DEVICE StreamId stream)
+                StreamId stream)
```
pcanal, line 66: In the case where the variable is actually never used in any configuration, should we keep the original (i.e. `[[maybe_unused]]`)?

sethrj: The problem is that GCC seems to warn when it's "never" instead of "maybe". I think it's a bug... basically it warns with or without the attribute unless it's actually used.

pcanal: I am confused .. I thought the bug was "only" that it complains if a `maybe not used` is actually used ... (and besides do we have to support that broken compiler version?)

sethrj: fe12ea2c948024693eb317635c091c060f1982e6 showed it when it wasn't used. This is on GCC 8.6 I think. When I have a moment to breathe I want to add some "oldest supported" compilers and dependencies to another CI toolchain so that we can be sure everything runs.

src/corecel/Macros.hh
```diff
@@ -226,15 +226,11 @@
     CLS& operator=(CLS&&) = delete
 
 /*!
- * \def CELER_UNUSED_UNLESS_DEVICE
+ * \def CELER_MARK_USED
```
esseivaju: `CELER_MARK_USED` or `CELER_NEVER_UNUSED`? I like the former (or ALWAYS_USED?) more than the double negation in "never_unused"

sethrj: I thought about "mark used" as well but I thought it would be more memorable if we have something closely parallel with the standard "maybe unused". Maybe just `CELER_MAYBE_USED` then? I thought about "always used" but it's not true... we are just marking it as "used". So maybe mark used is best but I worry that it'll be less memorable unless we have other `MARK_XX` macros

esseivaju: Well, "always used" and "never unused" are equivalents so both are not true. `CELER_MAYBE_[UN]USED`then as it's effectively a replacement for the `maybe_unused` attribute.

sethrj: Maybe we should stick with your `MARK_USED` though, since it's imperative (a command-like macro) as opposed to using the decorator-like `maybe_unused`.

esseivaju: Yeah, I don't have a strong opinion between these, I'd be fine with any. I agree with your point that `MARK_X` is not as memorable though. Since we have quite a few `CELER_VERB` macros, maybe `CELER_USE` or `CELER_DISCARD` since the macro is discarding the expression.

sethrj: CELER_IGNORE?

esseivaju: yeah, I like "discard" because that's the nomenclature [c++ uses](https://en.cppreference.com/w/cpp/language/expressions#Discarded-value_expressions) but CELER_IGNORE sounds good as well.

sethrj: Standard uses it, discard it shall be!

esseivaju (commented):
Do we know exactly what triggers this bug? There are more `[[maybe_unused]]`, e.g. in [DeviceAllocation](https://github.com/celeritas-project/celeritas/blob/1596c767cdadc620c258ea3dd85660881ecdc216/src/corecel/data/DeviceAllocation.cc#L60) that are either always or never used depending on the configuration but these are not producing warnings?

src/corecel/sys/Stream.cc
```diff
@@ -35,9 +36,11 @@ auto AsyncMemoryResource<Pointer>::do_allocate(
  * Deallocate device memory.
  */
 template<class Pointer>
-void AsyncMemoryResource<Pointer>::do_deallocate(
-    CELER_UNUSED_UNLESS_DEVICE pointer p, std::size_t, std::size_t)
+void AsyncMemoryResource<Pointer>::do_deallocate(pointer p,
```
pcanal, line 39: [pre-existing code question]  Why does the `do_deallocate` function takes 2 sizes as argument?

sethrj: Should be posted on #908 😉  This is to accommodate a thrust API. I'm guessing it's an alignment requirement, which isn't needed here because CUDA always aligns to something absurdly large (256 bytes IIRC).

pcanal (approved):

sethrj:
@esseivaju I'm not sure, but it seems to have something to do with preprocessing... I'm hoping at some point to add an "oldest compatible version" docker image that has Geant4 10.5 and gcc 8.3 or whatever.

PR #953, opened 2023-09-21
Implement asynchronous `DeviceAllocation`

esseivaju:
Add support for stream-ordered device allocation. The checks on whether we should use the sync or async device API are done at runtime to stay consistent with the `Copier` and `Filler` API though technically we could do everything at compile time with template specialization.

The next step is to update containers `Collection`, and `DeviceVector` to support stream-ordered semantics. Maybe for now we want to keep the implementation similar to the underlying components; add a constructor overload taking a stream.

The alternative would be to add a template argument (or Memspace / Ownership variant).

### DIFF
--- a/src/celeritas/track/detail/TrackSortUtils.cu
+++ b/src/celeritas/track/detail/TrackSortUtils.cu
@@ -19,6 +19,7 @@
 #include "corecel/Macros.hh"
 #include "corecel/data/Collection.hh"
 #include "corecel/data/Copier.hh"
+#include "corecel/data/DeviceVector.hh"
 #include "corecel/data/ObserverPtr.device.hh"
 #include "corecel/data/ObserverPtr.hh"
 #include "corecel/sys/Device.hh"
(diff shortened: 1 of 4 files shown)


sethrj:
How often are we going to have to allocate and reallocate memory? Would it be better to define a pinned-memory vector in the state (and only allocate it if `M == MemSpace::device`) and reuse that over multiple kernel launches?

esseivaju:
For now, I think the only times we need to allocate / deallocate device memory are during the sort track action. Since it's all done using the async API now and uses the cuda memory pool, the overhead is very small as we're already reusing that memory. 

Regarding the pinned memory, we never allocate it inside a step (and should not since it's expensive). So we definitely should allocate a buffer in the state and reuse it. We [have](https://github.com/celeritas-project/celeritas/blob/1596c767cdadc620c258ea3dd85660881ecdc216/src/celeritas/global/CoreState.hh#L146-L147C12) that already except it's not pinned memory.

sethrj (approved):

PR #954, opened 2023-09-22
Support energy loss fluctuations in uniform field along-step

amandalund:
This adds the option to enable fluctuations in the uniform field along-step kernel and disable fluctuations and/or msc in the rzmap field along-step kernel. I also added an option to forcibly disable energy loss fluctuations to preserve the original behavior so we can do comparisons with AdePT.

Using standalone celeritas on the GPU with a very simple problem (simple CMS and no msc or field) adding fluctuations increases the total time by ~6% and the along-step time by ~65%. For cms 2018 with msc and field the difference in both is pretty negligible.

### DIFF
--- a/app/celer-g4/RunInput.hh
+++ b/app/celer-g4/RunInput.hh
@@ -63,6 +63,7 @@ struct RunInput
     // Physics setup options
     PhysicsListSelection physics_list{PhysicsListSelection::ftfp_bert};
     GeantPhysicsOptions physics_options;
+    bool disable_eloss_fluct{false};  //!< Forcibly disable eloss fluctuations
 
     // Field setup options
     std::string field_type{"uniform"};
--- a/app/celer-g4/RunInput.hh
+++ b/app/celer-g4/RunInput.hh
@@ -63,7 +63,6 @@ struct RunInput
     // Physics setup options
     PhysicsListSelection physics_list{PhysicsListSelection::ftfp_bert};
     GeantPhysicsOptions physics_options;
-    bool disable_eloss_fluct{false};  //!< Forcibly disable eloss fluctuations
 
     // Field setup options
     std::string field_type{"uniform"};
--- a/src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cc
+++ b/src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cc
@@ -24,6 +24,7 @@
 
 #include "AlongStep.hh"
 #include "detail/FluctELoss.hh"
+#include "detail/MeanELoss.hh"
 #include "detail/RZMapFieldPropagatorFactory.hh"
 
 namespace celeritas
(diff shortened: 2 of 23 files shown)


app/celer-g4/RunInput.hh
```diff
@@ -63,6 +63,7 @@ struct RunInput
     // Physics setup options
     PhysicsListSelection physics_list{PhysicsListSelection::ftfp_bert};
     GeantPhysicsOptions physics_options;
+    bool disable_eloss_fluct{false};  //!< Forcibly disable eloss fluctuations
```
sethrj: How is this different from setting `GeantPhysicsOptions::eloss_fluctuation` to false? I thought we generally pulled this option through the EM parameters.

src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cc
```diff
@@ -75,15 +78,41 @@ AlongStepRZMapFieldMscAction::AlongStepRZMapFieldMscAction(
...
+    launch_impl(detail::TimeUpdater{});
+    if (this->has_fluct())
+    {
+        launch_impl(ElossApplier{FluctELoss{fluct_->ref<MemSpace::native>()}});
+    }
+    else
+    {
+        launch_impl(ElossApplier{MeanELoss{}});
+    }
+    launch_impl(TrackUpdater{});
```
sethrj, line 115: It would be interesting to see if there's any overhead from launching a bunch of separate CPU "kernels" versus a single one. The latter only has to check once for a track being disabled, for example. It might be better to put all the conditionals on the inside of the loop. There's going to be a tradeoff between improved locality of state data and locality of params data.

src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cu
```diff
@@ -43,11 +46,21 @@ void AlongStepRZMapFieldMscAction::execute(CoreParams const& params,
...
-        *this, fluct_->ref<MemSpace::native>(), params, state);
+    if (this->has_fluct())
+    {
+        detail::launch_apply_eloss(
+            *this, fluct_->ref<MemSpace::native>(), params, state);
+    }
+    else
+    {
+        detail::launch_apply_eloss(*this, params, state);
+    }
```
sethrj, line 63: I thought briefly about what it would take to make a general templated action (CUDA and C++) that would prevent this logic from being repeated across actions... we can revisit that later.

app/celer-g4/RunInput.hh
amandalund: That will only work with our geant physics list, not if we use FTFP BERT or something else... though you're right, for celer-sim it should be equivalent. Do you think we'd ever even need the original behavior where we don't allow fluctuations with a field, or should I just revert this? Now I'm not even sure why we were dong that in the first place instead of enabling/disabling them based on the imported EM parameters.

src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cc
amandalund: That's a good point, I can look into that.

app/celer-g4/RunInput.hh
sethrj: I think the capability predated the parameters. I don't think there's a reason to disable if the corresponding Geant option is enabled: we really only wanted to do this for comparison with adept and for performance testing, and for both those cases the celer sim option should be sufficient.

src/celeritas/global/alongstep/AlongStepRZMapFieldMscAction.cc
amandalund: Looking at a couple problems (one using general linear and one using uniform msc along-step) it looks like using a single CPU "kernel" results in ~5-10% improvement, so we might want to consider converting all of these.

sethrj (approved):
Looks great! Thanks @amandalund .

PR #955, opened 2023-09-23
Add collection builder that deduplicates on insertion

sethrj:
This is intended for use primarily with ORANGE to essentially "compress" parameter data while it's being assembled. It hashes every call to `insert_back` and returns a previously inserted range if available. There are a few subtleties:
1. The data cannot be modified after it's inserted. The CDF renormalization in physics violates this.
2. Each element must be "uniquely representable", i.e. it can't have padding nor contain floats. Specializations of `std::hash` must be created for each one.


src/corecel/data/CollectionBuilder.hh
```diff
@@ -76,7 +76,9 @@ class CollectionBuilder
 
     // Append a single element
     inline ItemIdT push_back(value_type const& element);
-    inline ItemIdT push_back(value_type&& element);
+
+    // Erase a number of elements from the end
+    inline void erase_back(size_type count = 1);
```
amandalund: Is this used anywhere (or tested)?

amandalund (approved):
Looks good @sethrj!

src/corecel/data/CollectionBuilder.hh
sethrj: Nope :) I originally put it in because I thought the dedupe collection builder was going to use it. I'll delete.

PR #956, opened 2023-09-23
Deduplicate ORANGE geometry on insertion

sethrj:
This adds deduplication for ORANGE parameter data. It uses the new deduplicating collection builder to reuse identical spans of data for surfaces, logic definitions, volume ID ranges, etc.

In the hex array test, the following simplifications occur:
```diff
- {"local_surface_ids":364,"local_volume_ids":364,"logic_ints":922,"reals":281}
+ {"local_surface_ids":191,"local_volume_ids":348,"logic_ints":585,"reals":272}
```
so mostly the logic and local surface IDs because so many hexes have the same definition.


elliottbiondo:
@sethrj did you intend to commit the regenerated json files to this branch? Seems like that should be a separate PR.

src/orange/OrangeParamsOutput.cc
```diff
@@ -0,0 +1,111 @@
...
+            OPO_SAVE_BIH_SIZE(leaf_nodes);
+#    undef OPO_SAVE_BIH_SIZE
+            sizes["bih"] = std::move(bih);
+        }
+
+        // Universe indexer data
+        {
+            sizes["universe_indexer"] = {
+                {"surfaces", data.universe_indexer_data.surfaces.size()},
+                {"volumes", data.universe_indexer_data.surfaces.size()},
```
amandalund:
```suggestion
                {"volumes", data.universe_indexer_data.volumes.size()},
```
And update the tests.

sethrj:
Good point @elliottbiondo . I'll open it separately. It shouldn't affect the deduplication...

amandalund (approved):
Looks good to me @sethrj!

src/orange/OrangeParamsOutput.cc
sethrj: Turns out these are both the same size of num_unverses + 1. I'll just omit writing it.

elliottbiondo (approved):

PR #914, opened 2023-08-29
Add option to generate primaries without an event file in celer-g4

amandalund:
This adds another `PrimaryGeneratorAction` to celer-g4 which lets us optionally generate primaries from a particle gun with parameters specified in the user input (similar to the `PrimaryGenerator` used in celer-sim).

PR #957, opened 2023-09-25
Add defaults and runtime checking to primary generator

sethrj:
Enables existing celer-sim (mainly the regression suite) inputs to work with #914 by providing (but warning about) a default seed. Also allow "isotropic" without specifying an empty "params" list. Finally, always check the number of parameters even if debug assertions are off.


amandalund (approved):
Thanks @sethrj!

sethrj:
Sorry @amandalund , one more thing 😄

PR #958, opened 2023-09-25
Fix ORANGE rectangular array construction from SCALE

elliottbiondo:
This PR updates Celeritas to expect rect array universes in .json geometry files created by SCALE to be provided in IJK iteration order. Two additional bug fixes are addressed:

1) The outside surfaces of the rect array are set to +/- infinity in each direction, essentially deduplicating them.
2) cross_boundary now expects a post-crossing state.


src/orange/detail/RectArrayInserter.cc
```diff
@@ -46,14 +46,16 @@ RectArrayId RectArrayInserter::operator()(RectArrayInput const& inp)
...
-        std::vector<double> const& grid = inp.grid[to_int(ax)];
+        std::vector<double> grid = inp.grid[to_int(ax)];
         CELER_VALIDATE(grid.size() >= 2,
                        << "grid for " << to_char(ax) << " axis in '"
                        << inp.label << "' is too small (size " << grid.size()
                        << ")");
         CELER_VALIDATE(std::is_sorted(grid.begin(), grid.end()),
                        << "grid for " << to_char(ax) << " axis in '"
                        << inp.label << "' is not monotonically increasing");
+        grid.front() = -std::numeric_limits<real_type>::infinity();
```
sethrj, line 60: Add a comment about what this is doing:
```suggestion
        // Suppress the outer grid boundaries to avoid coincident surfaces with other universes
        grid.front() = -std::numeric_limits<real_type>::infinity();
```
And technically I think we should be "bumping" these rather than making them infinite, but this is just as good if we know a priori the user didn't make an error in construction.

src/orange/univ/RectArrayTracker.hh
```diff
@@ -80,7 +80,7 @@ class RectArrayTracker
     inline CELER_FUNCTION Intersection intersect(LocalState const& state,
                                                  real_type max_dist) const;
 
-    // Find the new volume by crossing a surface
+    // Find the local volume, given a post-crossing state
```
sethrj: Comma splice
```suggestion
    // Find the local volume given a post-crossing state
```

src/orange/univ/RectArrayTracker.hh
```diff
@@ -181,23 +181,22 @@ RectArrayTracker::cross_boundary(LocalState const& state) const
...
+             && state.surface.sense() == Sense::outside));
+
     // Value for incrementing the axial coordinate upon crossing
-    int inc = (state.surface.sense() == Sense::outside) ? -1 : 1;
+    // NOTE: that surface sense is currently the POST crossing value
+    int inc = (state.surface.sense() == Sense::outside) ? 1 : -1;
 
     detail::OnLocalSurface new_surface(state.surface.id(),
-                                       flip_sense(state.surface.sense()));
+                                       state.surface.sense());
```
sethrj: This is just `state.surface` now, and you can replace the use of it below directly with that.

src/orange/univ/RectArrayTracker.hh
```diff
@@ -181,23 +181,22 @@ RectArrayTracker::cross_boundary(LocalState const& state) const
     auto coords = to_coords(state.volume.unchecked_get());
     auto ax_idx = this->find_surface_axis_idx(state.surface.id());
 
+    // Due to surface deduplication, crossing out of rect arrays is not
+    // possible
+    CELER_EXPECT(
```
sethrj:
```suggestion
    CELER_ASSERT(
```

sethrj (approved):
Looks good @elliottbiondo ! Just a couple of very minor suggestions.

elliottbiondo:
@sethrj I had to remove some tests that only work when the UniverseVisitor is "hooked up". This is good-to-go pending CI.

PR #959, opened 2023-09-25
Implement TrackerVisitor for multiply-dispatched tracker types

elliottbiondo:
This PR adds a `TrackerVisitor` class to facilitate multiply-dispatched tracker types (i.e., SimpleUnitTracker, RectArrayTracker). This functionality can be "hooked up" by replacing tracker calls like so:

```
auto tracker = this->make_tracker(uid);
auto tinit = tracker.initialize(local);
```

becomes:

```
TrackerVisitor visit_tracker{params_};
detail::Initialization tinit;
visit_tracker([&local, &tinit](auto&& t) { tinit = t.initialize(local); }, uid);
```

However, this is not done in the PR, as it appears this add 60% overhead to tracking with a single universe type.

### DIFF
--- a/src/orange/univ/UniverseTypeTraits.hh
+++ b/src/orange/univ/UniverseTypeTraits.hh
@@ -8,6 +8,7 @@
 #pragma once
 
 #include "orange/OrangeTypes.hh"
+#include "orange/univ/RectArrayTracker.hh"
 
 namespace celeritas
 {
@@ -31,8 +32,35 @@ struct UniverseTypeTraits;
     }
 
 ORANGE_UNIV_TRAITS(simple, SimpleUnit);
+ORANGE_UNIV_TRAITS(rect_array, RectArray);
 
 #undef ORANGE_UNIV_TRAITS
+
+//---------------------------------------------------------------------------//
+/*!
+ * Expand a macro to a switch statement over all possible universe types.
+ *
+ * The \c func argument should be a functor that takes a single argument which
+ * is a UniverseypeTraits instance.
+ */
+template<class F>
+CELER_CONSTEXPR_FUNCTION decltype(auto)
+visit_universe_type(F&& func, UniverseType ut)
+{
+#define ORANGE_UT_VISIT_CASE(TYPE)          \
+    case UniverseType::TYPE:                \
+        return celeritas::forward<F>(func)( \
+            UniverseTypeTraits<UniverseType::TYPE>{})
+
+    switch (ut)
+    {
+        ORANGE_UT_VISIT_CASE(simple);
+        ORANGE_UT_VISIT_CASE(rect_array);
+        default:
+            CELER_ASSERT_UNREACHABLE();
+    }
+#undef ORANGE_UT_VISIT_CASE
+}
 
 //---------------------------------------------------------------------------//
 }  // namespace celeritas
--- a/src/orange/univ/UniverseTypeTraits.hh
+++ b/src/orange/univ/UniverseTypeTraits.hh
@@ -8,13 +8,13 @@
 #pragma once
 
 #include "orange/OrangeTypes.hh"
-#include "orange/univ/RectArrayTracker.hh"
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 struct SimpleUnitRecord;
 class SimpleUnitTracker;
+class RectArrayTracker;
 
 //---------------------------------------------------------------------------//
 /*!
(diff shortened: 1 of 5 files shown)


src/orange/univ/UniverseTypeTraits.hh
```diff
@@ -8,6 +8,7 @@
 #pragma once
 
 #include "orange/OrangeTypes.hh"
+#include "orange/univ/RectArrayTracker.hh"
```
sethrj: Replace this with a forward declaration.

src/orange/univ/UniverseTypeTraits.hh
```diff
@@ -31,8 +32,35 @@ struct UniverseTypeTraits;
...
+ORANGE_UNIV_TRAITS(rect_array, RectArray);
 
 #undef ORANGE_UNIV_TRAITS
 
+//---------------------------------------------------------------------------//
+/*!
+ * Expand a macro to a switch statement over all possible universe types.
+ *
+ * The \c func argument should be a functor that takes a single argument which
+ * is a UniverseypeTraits instance.
```
sethrj:
```suggestion
 * is a UniverseTypeTraits instance.
```

src/orange/univ/UniverseVisitor.hh
```diff
@@ -0,0 +1,85 @@
...
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/univ/UniverseVisitor.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "corecel/math/Algorithms.hh"
+#include "orange/OrangeData.hh"
+
+#include "UniverseTypeTraits.hh"
```
sethrj: You'll need to include both `RectArrayTracker.hh` and `SimpleUnitTracker.hh` here.

src/orange/univ/UniverseVisitor.hh
```diff
@@ -0,0 +1,85 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Apply a functor to a simple or rect-array universe
```
sethrj:
```suggestion
 * Apply a functor to a universe tracker of unknown type.
```

src/orange/univ/UniverseVisitor.hh
```diff
@@ -0,0 +1,85 @@
...
+  public:
+    // Construct from ORANGE params
+    explicit inline CELER_FUNCTION UniverseVisitor(ParamsRef const& params);
+
+    // Apply the function to the universe specified by the given ID
+    template<class F>
+    CELER_FUNCTION decltype(auto) operator()(F&& func, UniverseId id);
+
+  private:
+    ParamsRef params_;
```
sethrj:
```suggestion
    ParamsRef const& params_;
```

src/orange/univ/UniverseVisitor.hh
```diff
@@ -0,0 +1,85 @@
...
+ * Construct from ORANGE params.
+ */
+CELER_FUNCTION UniverseVisitor::UniverseVisitor(ParamsRef const& params)
+    : params_(params)
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply the function to the transform specified by the given ID.
```
sethrj:
```suggestion
 * Apply the function to the universe specified by the given ID.
```

src/orange/univ/UniverseVisitor.hh
sethrj: I realize now there's an inconsistency between our "tracker" name and the "universe" here. I think universe is the abstraction of a unit, rectangular array; and a tracker is the helper class that does the actual navigation/tracking/state update. Just like we have `SurfaceTypeTraits` but `LocalSurfaceVisitor`, maybe rename this class `TrackerVisitor`. Also please add a small unit test.

sethrj (changes requested):

elliottbiondo:
@sethrj this should be ready pending CI

sethrj:
@elliottbiondo I was thinking [the unintentional params copy](https://github.com/celeritas-project/celeritas/pull/959#discussion_r1336451403) could be the source of the 60% slowdown you saw. (Did you look at the occpancy?) After merging, could you please try rerunning maybe a single case with pseudo old vs new on wildstyle to see if the 60% performance regression dissappears?

sethrj (approved):
Looks great, thanks!

elliottbiondo:
> @elliottbiondo I was thinking [the unintentional params copy](https://github.com/celeritas-project/celeritas/pull/959#discussion_r1336451403) could be the source of the 60% slowdown you saw. (Did you look at the occpancy?) After merging, could you please try rerunning maybe a single case with pseudo old vs new on wildstyle to see if the 60% performance regression dissappears?

Yep, and this might change the frontier results as well.

PR #960, opened 2023-09-26
Fix using ORANGE  geometry with pointer-appended GDML

sethrj:
I created a new testem3 GDML file with @stognini 's generator, and updated the equivalent ORANGE file, but it failed to run:
```
celeritas/src/celeritas/geo/GeoMaterialParams.cc:159: warning: Some geometry volumes do not have known material IDs: world@world, ....
celeritas/src/celeritas/geo/GeoMaterialParams.cc:187:
celeritas: runtime error: num_missing != geo.num_volumes() failed:
    no geometry volumes matched the available materials:
 materials: ..., {world@0x600001b3cd20,0}, ...
volumes: ..., world@world, ...
```

This is because the geo-material mapper only considered the options that either the name+extension match exactly (e.g. when loading with vecgeom gdml + vecgeom geometry), or when only the name matches (can happen when using geant with pointer stripping?), but it didn't allow for the case when loading a GDML through geant4 for physics but using the ORANGE geometry for tracking.

I've refactored the geomaterial code to do a better job of mapping volume names. Hopefully the code is clearer, and I've updated one of the tests to ensure it still works. Before merging I'd like to make sure that both the `celer-sim` and `accel` pathways work with geometries that have duplicate names (but uniquifying extensions) like in cms.


src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -28,6 +28,211 @@
...
+        // Hopefully user-provided and geo-provided volume labels match exactly
+        if (auto iter = materials_.find(vol_label); iter != materials_.end())
+        {
+            return iter->second;
+        }
+
+        if (mat_labels_.empty())
+        {
+            // Build user volume -> material mapping
+            this->build_mat_labels();
```
pcanal, line 106: This seems a somewhat expensive operation.  Is it done only once per process or more often?

src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -28,6 +28,211 @@
...
+        for (auto const& [mlabel, matid] : materials_)
+        {
+            mat_labels_.emplace(mlabel.name, PairExtMatid{mlabel.ext, matid});
+        }
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Whether a volume with a missing material needs to be reported to the user.
```
pcanal, line 159: Should the documentation describe and/or refers to the documentation of the heuristic/convention used.

src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -28,6 +28,211 @@
...
+std::vector<MaterialId>
+build_vol_to_mat(GeoParams const& geo, MapLabelMatId const& materials)
+{
+    std::vector<Label> missing_volumes;
+    std::vector<MaterialId> result(geo.num_volumes(), MaterialId{});
+
+    // Make sure at least one volume maps correctly
+    VolumeId::size_type num_missing{0};
+
+    // Map volume names to materiail names
```
pcanal:
```suggestion
    // Map volume names to material names
```

src/celeritas/geo/GeoMaterialParams.cc
sethrj: Once per process. It's cached here.

amandalund:
What was the reason for regenerating the GDML?

sethrj:
@amandalund I'm finally running a benchmark problem with simple CMS divided into r,z,theta to see how the scaling works. The geometry exporter by @stognini saves pointers (and it's reasonable to assume other user inputs will do the same). I regenerated the testem3 gdml only inside the test directory in order to check that leaving the pointers still works with GDML+ORANGE.

src/celeritas/geo/GeoMaterialParams.cc
pcanal: Ok.  So per convention this is okay but this code is not thread safe (and it does not claim to be since this is not a const function (but semantically it should be, shouldn't it).

Do we have a handle on the list of class/object that are shared amongst threads but are not thread safe? 

Actually the call to `build_mat_labels` is delayed until the first call to `operator ()`, do we really have a guarantee (in all problem setups) that there will be such a call before this is used in multi-thread mode?

sethrj: This is literally an implementation detail called locally by a constructor. It doesn't exceed the lifetime of a local function. It doesn't access any shared resources. It is therefore thread safe.

pcanal: Fair enough.  It may or may not be useful to mark a little more clearly (yes, I have finally noticed it is declared inside the source file) that this is an internal class (i.e "literally an implementation detail"), for example by putting into a 'detail' or 'unnamed' namespace.

test/celeritas/data/testem3-flat.gdml
```diff
@@ -4,1288 +4,1288 @@
...
-      <physvol name="absorber_lv_11">
-        <volumeref ref="absorber_lv_11"/>
-        <position name="absorber_lv_11_pos" unit="mm" x="-106.85" y="0" z="0"/>
+    <volume name="gap_00x600001b3cdc0">
+      <materialref ref="Pb0x14ce28a60"/>
+      <solidref ref="gap_shape0x60000013ce70"/>
+      <auxiliary auxtype="gammaECut" auxunit="MeV" auxvalue="0.0946952"/>
+      <auxiliary auxtype="electronECut" auxunit="MeV" auxvalue="1.00699"/>
+      <auxiliary auxtype="positronECut" auxunit="MeV" auxvalue="0.946936"/>
+      <auxiliary auxtype="protonECut" auxunit="MeV" auxvalue="0.07"/>
```
amandalund: More a question about the geometry exporter, but since the Geant4 GDML parser doesn't actually read/apply these cutoff values, is there any reason to export them? It seems like it might potentially cause confusion and bloat the GDML file.

sethrj: That is a good point. I guess we'd have to read these in the SD layer and later reuse those stored values in the physics? Sounds like a lot of work though, especially if no one uses them. @stognini this is your department: maybe we shouldn't save these?

src/celeritas/geo/GeoMaterialParams.cc
sethrj: @pcanal It *is* in an anonymous namespace.

test/celeritas/data/testem3-flat.gdml
stognini: We never fully implemented the import of these cutoff values, but I added them early in the exporter to have that ready. We could just set the export cutoff flag to false when exporting the gdmls until we're ready to use them.

src/celeritas/geo/GeoMaterialParams.cc
pcanal: Perfect then (and sorry; it is not obvious from the github output :) ).

src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -28,6 +28,211 @@
...
+{
+//---------------------------------------------------------------------------//
+using MapLabelMatId = std::unordered_map<Label, MaterialId>;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a label -> material map from the input.
+ *
+ * The input is effectively an "unzipped" unordered list of (volume label,
+ * material pairs).
```
amandalund:
```suggestion
 * material ID) pairs.
```
?

src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -28,6 +28,211 @@
...
+        }
+
+        // Multiple labels match
+        CELER_LOG(warning)
+            << "Multiple materials match the volume '" << vol_label << "': "
+            << join_stream(
+                   start, stop, ", ", [](std::ostream& os, auto&& mliter) {
+                       os << mliter.second.second.unchecked_get();
+                   });
+        return start->second.second;
```
stognini: Maybe I'm reading this wrong, but since from this point on we return `start->second.second` for the `MaterialId`, might be easier to read if we simply do:

```cpp
if (std::distance(start, stop) > 1)
{
    // Multiple labels match
    CELER_LOG(warning)
        << "Multiple materials match the volume '" << vol_label << "': "
        << join_stream(
                start, stop, ", ", [](std::ostream& os, auto&& mliter) {
                    os << mliter.second.second.unchecked_get();
                });
}
return start->second.second;
```

src/celeritas/geo/GeoMaterialParams.cc
stognini: Nitpick: I think here you meant `(volume label, material id) pairs.`

test/celeritas/data/testem3-flat.gdml
stognini: Just set `G4GDMLParser::SetEnergyCutsExport` flag to `false` in [utils/gdml-generator](https://github.com/celeritas-project/utils/tree/main/gdml-generator) for the time being.

amandalund (commented):
Thanks @sethrj, I think the logic is easier to follow now. I tried loading the cms2018, cms-run3, and cms-hllhc geometries through celer-sim and celer-g4. I didn't notice any changes with celer-g4; with celer-sim the warnings are slightly different, but as long as that's expected I think this looks good.

cms2018 has the additional warnings:
```console
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardHoldersConnector@0x7f4a8f51d600': 277, 277, 277, 277
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardHoldersConnector@0x7f4a8f4d5840': 277, 277, 277, 277
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardBladeBase@0x7f4a8f4cb7c0': 277, 277, 277, 277
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardModule@0x7f4a9a7ccb10': 277, 277, 277, 277
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardHoldersConnector@0x7f4a8f4d5a80': 277, 277, 277, 277
...
/home/alund/celeritas_project/celeritas/src/celeritas/geo/GeoMaterialParams.cc:125: warning: Multiple materials match the volume 'PixelForwardModule@0x7f4a9a7cfcf0': 277, 277, 277, 277
```
and all three geometries had fewer volumes without known material IDs.

src/celeritas/geo/GeoMaterialParams.cc
sethrj: Good point. Thanks! With the output @amandalund got though, it'll make even more sense to build a temporary 'set' and only print if more than one ID exists in the set.

amandalund (approved):

sethrj:
I think all the warnings @amandalund showed should be gone. It'll only warn if materials are different.

PR #962, opened 2023-09-26
Improve JSON conversion for bounding box edge cases

sethrj:
This simplifies ORANGE JSON files so that fully infinite bounding boxes (often the case for exterior volumes) are written as "null".


src/orange/BoundingBox.hh
```diff
@@ -93,6 +93,30 @@ class BoundingBox
...
+template<class T>
+CELER_CONSTEXPR_FUNCTION bool
+operator==(BoundingBox<T> const& lhs, BoundingBox<T> const& rhs)
+{
+    return lhs.lower() == rhs.lower() && lhs.upper() == rhs.upper();
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Test inequality of two arrays.
```
elliottbiondo: bounding boxes

src/orange/BoundingBoxIO.json.hh
```diff
@@ -36,36 +36,77 @@ inline void fix_inf(celeritas::Array<T, 3>* point)
         }
     }
 }
+
+//---------------------------------------------------------------------------//
+//! Replace "max" with "inf" since the latter can't be represented in JSON.
```
elliottbiondo: "inf" with "max"

src/orange/BoundingBoxIO.json.hh
```diff
@@ -36,36 +36,77 @@ inline void fix_inf(celeritas::Array<T, 3>* point)
...
+    if (bbox == BoundingBox<T>::from_infinite())
+    {
+        // Special case: save fully infinite is null
+        j = nullptr;
+        return;
+    }
+
+    auto lower = bbox.lower();
+    auto upper = bbox.upper();
+
```
elliottbiondo: Add comment about semi-infinite bounding boxes

elliottbiondo (changes requested):
If I understand correctly, if a bounding box is fully infinite, it is written as "null", but if it is semi-infinite, it is written to json as before, with +/- max double. I don't see how this is better: it seems like if you are already using max doubles in some cases, you should use them in all applicable cases. In addition, if someone opens up a json file and see `null`, i.e. how nlohmann json handles null pointers, they are probably more likely to assume the file is ill-formed, rather than handling a perfectly normal case, (i.e., an infinite bounding box).

sethrj:
@elliottbiondo The motivation was that the `[EXTERIOR]` volumes are always infinite, and every time I opened a diff with the new JSON files I was confused by the out-of-range bbox. You're totally right though that `null` isn't the right representation. I think:
- Missing bbox == infinite bbox (this is consistent with our old JSON files and current BIH implementation: if it's not there, we pretend it's infinite so that we always check)
- `null` bbox == null bbox (i.e. false/invalid)

I don't think we'll have semi-infinite bboxes very often, but if we did, I think it would be better to replace the infinite elements as strings `"-inf"` and `"+inf"`? Not worth the extra complexity now IMHO.

src/orange/BoundingBoxIO.json.cc
```diff
@@ -0,0 +1,120 @@
...
+        auto ax = to_int(axis);
+        if (std::fabs((*point)[ax]) == max_real)
+        {
+            (*point)[ax] = std::copysign(inf, (*point)[ax]);
+        }
+    }
+}
+
+//---------------------------------------------------------------------------//
+//! Replace "inf" with "max" since the latter can't be represented in JSON.
```
elliottbiondo: Since the former

src/orange/BoundingBoxIO.json.cc
```diff
@@ -0,0 +1,120 @@
...
+}
+
+//---------------------------------------------------------------------------//
+}  // namespace
+
+//---------------------------------------------------------------------------//
+/*!
+ * Read a bounding box from a JSON file.
+ *
+ * A bounding box can be either \c null , indicating an infinite or unknown
```
elliottbiondo: I would say that null doesn't _indicate_ an infinite bounding box, but rather it indicates an missing bbox that we treat as infinite.

elliottbiondo (approved):

src/celeritas/CMakeLists.txt
```diff
@@ -78,6 +78,9 @@ list(APPEND SOURCES
   user/StepCollector.cc
 )
 
+# Bounding box used for output is in "orange"
+list(APPEND PUBLIC_DEPS Celeritas::orange)
```
pcanal, line 82: This implies that `ORANGE` is always built, right?

pcanal: But it sounds like `bounding box` should be in a library that both `Orange` and the `output` depends on.

src/orange/BoundingBoxIO.json.cc
```diff
@@ -0,0 +1,120 @@
...
+        return;
+    }
+
+    // Replace unrepresentable infinities with large values
+    auto lower = bbox.lower();
+    auto upper = bbox.upper();
+    inf_to_max(&lower);
+    inf_to_max(&upper);
+
+    j = nlohmann::json::array({lower, upper});
```
pcanal, line 107: is the `nlohman` conventions to transform `max` into `"null"` ?

pcanal (dismissed):
A couple of clarifications.

src/orange/BoundingBoxIO.json.cc
sethrj: I think it transforms any unrepresentable number to null: ind, -inf, or nan. JSON itself doesn't allow inf or the like. Replacing with signed max is the only practical solution.

src/celeritas/CMakeLists.txt
sethrj: Yes, there are "core geometry " components in orange, such as types and interfaces. Splitting them out to another library is too far at this point. Better might be to move the other two wrappers into this one and rename the library "celgeo" or something

src/orange/BoundingBoxIO.json.cc
pcanal: > Replacing with signed max is the only practical solution.

Alright, so if I understand correctly the json file now contains literally the numerical value of `std::numeric_limits<T>::max()`, right?

If yes, then do we have a guaranteed that for all values we are using a size (number of bits) guaranteed to be stable on all platforms (eg. `int` or `long` are not).  If not, the produced file would become platform dependent.

src/celeritas/CMakeLists.txt
pcanal: Alright.  This is not strictly related to this PR so moving the discussion off-line.

src/orange/BoundingBoxIO.json.cc
sethrj: Correct. I think the potential issue isn't the number of representable bits (these are all IEEE, remember) but slight differences in the text representation of the same value. The produced file can be, but that's the case based on the NLJSON version as well. The important part is whether exporting on A and importing on B changes anything substantial. For ~ulp level differences the answer is that it might disrupt a random number stream or change an exact hash, but nothing should really be different. In this case, we're replacing a representable number `1e308` with an overflow value `inf`. In any place where we don't have an `isinf` these will be effectively equivalent.

The BIH construction *does* have a special case for +-inf, but maybe that threshold should be changed to huge as well? Or huge - epsilon? Or huge * .5 ? (Where does it end???)

pcanal: >  I think the potential issue isn't the number of representable bits

I am confused.  If on platform `A` the size of `T` is 32 bits and the value `inf` ... humm oh right ... this code actually in practice only applies to floating points which have stable number of bits, so my point is moot (the point was that `std::numeric_limits<T>::max()` will have completely different values on different platforms for `int` and `long`).

> The BIH construction does have a special case for +-inf, but maybe that threshold should be changed to huge as well? Or huge - epsilon? Or huge * .5 ? (Where does it end???)

The question is how the results varies if `BIH` sees `+inf` vs `huge - epsilon` (essentially the nature of the special casing).

sethrj: Performance will be the only difference; and it may not even be much. (Just having to check the "big" volume more than once potentially.)

PR #963, opened 2023-09-28
Fix build errors in SCALE CI

sethrj:
This fixes two downstream problems with SCALE:
- use of a host-device function by host code (because SCALE builds everything with CUDA)
- use of implicit `operator==` for the daughter struct (not sure why this worked in the current version)

```
celeritas/src/corecel/math/detail/FnvHasher.hh(131): error #20013-D: calling a constexpr __host__ function("to_integer") from a __host__ __device__ function("operator()") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
          detected during:
            instantiation of "void celeritas::detail::FnvHasher<T>::operator()(std::byte) const [with T=std::size_t]" 
(147): here
            instantiation of "void celeritas::detail::FnvHasher<T>::operator()(std::size_t) const [with T=std::size_t]" 
/home/s3j/.local/src/celeritas/src/corecel/math/HashUtils.hh(67): here
            instantiation of "std::size_t celeritas::hash_combine(const Args &...) [with Args=<std::__cxx11::string, std::__cxx11::string>]" 
/home/s3j/.local/src/celeritas/src/corecel/io/Label.hh(125): here

1 error detected in the compilation of "/home/s3j/.local/src/celeritas/test/orange/Orange.test.cu".
```

### DIFF
--- a/src/corecel/math/detail/FnvHasher.hh
+++ b/src/corecel/math/detail/FnvHasher.hh
@@ -88,16 +88,16 @@ class FnvHasher
 
   public:
     // Construct with a reference to the hashed value which we initialize
-    explicit inline CELER_FUNCTION FnvHasher(value_type* hash_result);
+    explicit inline FnvHasher(value_type* hash_result);
 
     // Hash a byte of data
-    CELER_FORCEINLINE_FUNCTION void operator()(std::byte b) const;
+    CELER_FORCEINLINE void operator()(std::byte b) const;
 
     // Hash a size_t (useful for std::hash integration)
-    inline CELER_FUNCTION void operator()(std::size_t value) const;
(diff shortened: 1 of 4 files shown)


pcanal:
< use of implicit operator== for the daughter struct (not sure why this worked in the current version)

Are the daughter deduplicatable? ie. Is it an option to implement the `operator==` instead of removing the deduplication?

sethrj:
@pcanal I don't think it will ever save much: daughter is a combination of transform + universe ID, so it's unique inside a universe. If you have a *bunch* of universes and they all reference a single universe with no translation then it'll save 8 bytes for each one. This can only be possible in a pathological case...

amandalund (approved):

PR #964, opened 2023-10-03
Add option to disable track count output in celer-sim

amandalund:
The celer-sim output can get quite large for some problems when writing the initializer, alive, and active tracks counts at each step (and for each event). This lets us optionally disable that output.

### DIFF
--- a/app/celer-sim/Runner.cc
+++ b/app/celer-sim/Runner.cc
@@ -350,6 +350,7 @@ void Runner::build_transporter_input(RunnerInput const& inp)
     transporter_input_->num_track_slots
         = ceil_div(inp.num_track_slots, core_params_->max_streams());
     transporter_input_->max_steps = inp.max_steps;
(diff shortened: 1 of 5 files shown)


app/celer-sim/Runner.cc
```diff
@@ -350,6 +350,7 @@ void Runner::build_transporter_input(RunnerInput const& inp)
     transporter_input_->num_track_slots
         = ceil_div(inp.num_track_slots, core_params_->max_streams());
     transporter_input_->max_steps = inp.max_steps;
+    transporter_input_->store_track_counts = inp.write_track_counts;
```
sethrj, line 353: Store in one place, write in another? 😅 I get it though. I think for 1.0 we'll revisit these options anyway.

sethrj (approved):

sethrj:
Thanks @amandalund , I've wanted this option for a while 😊

PR #965, opened 2023-10-03
Add unified memory support

esseivaju:
Add a `mapped` memspace that can be used on host and device. The device must support [unified addressing](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__UNIFIED.html#group__CUDART__UNIFIED).

### DIFF
--- a/src/celeritas/global/CoreState.hh
+++ b/src/celeritas/global/CoreState.hh
@@ -126,31 +126,11 @@ class CoreState final : public CoreStateInterface
 
     // Reference to the host ActionThread collection for holding result of
     // action counting
-    auto& action_thread_offsets()
-    {
-        if constexpr (M == MemSpace::device)
-        {
-            return host_thread_offsets_;
-        }
-        else
-        {
-            return thread_offsets_;
-        }
-    }
+    inline auto& action_thread_offsets();
 
     // Const reference to the host ActionThread collection for holding result
     // of action counting
-    auto const& action_thread_offsets() const
-    {
-        if constexpr (M == MemSpace::device)
-        {
-            return host_thread_offsets_;
-        }
-        else
-        {
-            return thread_offsets_;
-        }
-    }
+    inline auto const& action_thread_offsets() const;
 
     // Reference to the ActionThread collection matching the state memory
     // space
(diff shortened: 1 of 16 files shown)


src/celeritas/global/CoreState.hh
```diff
@@ -126,11 +126,31 @@ class CoreState final : public CoreStateInterface
 
     // Reference to the host ActionThread collection for holding result of
     // action counting
-    ActionThreads<MemSpace::host>& action_thread_offsets();
+    auto& action_thread_offsets()
```
sethrj: Move the definitions outside of the class definition (and mark this guy as `inline`)

src/corecel/data/detail/CollectionImpl.hh
```diff
@@ -198,6 +220,20 @@ struct CollectionAssigner<Ownership::value, MemSpace::device>
...
+//---------------------------------------------------------------------------//
+//! Assignment semantics for copying to mapped memory
+template<>
+struct CollectionAssigner<Ownership::value, MemSpace::mapped>
+{
+    template<class T, Ownership W2, MemSpace M2>
+    auto operator()(CollectionStorage<T, W2, M2> const& source)
+        -> CollectionStorage<T, Ownership::value, M2>
+    {
+        CELER_EXPECT(celeritas::device().support_mapped_memory());
```
sethrj: This should probably be a validate since it's dependent on the user system, and they may have only installed a nondebug build.

src/corecel/sys/Device.hh
```diff
@@ -108,6 +108,9 @@ class Device
     //! Default number of threads per block
     unsigned int default_block_size() const { return default_block_size_; }
 
+    //! Check that the device supports mapped pinned memory
+    bool support_mapped_memory() const { return support_mapped_memory_; }
```
sethrj: Can you add this to the DeviceIO.json.cc as well so we get the output as a diagnostic?

sethrj (commented):
I think this is great: despite a sort of false start a couple of years ago, I think it likely more systems will move toward a single memory space shared by ARM+GPU (Grace Hopper, Apple, ...). One missing piece from this PR, I think, is a careful review of the host/device specializations throughout the code. Some of the "host" specializations should probably be the default template code (with "device" being specialization). A couple places I'm seeing are:
- Filler
- Copier
- ObserverPtr (which probably needs some additional specializations)

src/corecel/data/detail/CollectionImpl.hh
esseivaju: I think semantically the best place to call `CELER_VALIDATE` would be in the CollectionStorage ctor but it would require quite a bit of refactoring without an actual difference at runtime.

esseivaju:
Good point! I'll have a look at these.

sethrj (approved):
Looks great. Hopefully this will help out with the sorting overhead? 🤞

PR #853, opened 2023-07-13
Add `__launch_bounds__` to kernels based on Executor properties

esseivaju:
Add template parameters to `ActionLauncher` to specify `__launch_bounds__` for each kernel launch. It probably make sense to add a compile-time option to enable/disable __launch_bounds__ but we'd ideally enable it independently for each kernel since it might improve performance only for a subset. Not sure what would be the most ergonomic way, maybe a cmake variable with a list of action labels?

PR #968, opened 2023-10-05
Fix performance regression on HIP

sethrj:
#853 seems to cause the local memory overflow to increase substantially, which in turn causes tracking performance to drop by a factor of 3 on crusher.

When compiling with CUDA the launch bound in the field propagator doesn't change the `max_blocks_per_cu` value in practice, in vecgeom or ORANGE. It *does* however force the occupancy in HIP to be way too high, reducing performance by a factor of 3. I've confirmed that the change in this PR resets the occupancy to a lower value with less spilling.

~~This doesn't explain why the non-field problems are so much slower on AMD than Nvidia though...~~
Somehow the last crusher run had absurdly long times for the first step, possibly indicative of some hardware fault. I'm rerunning before-and-after numbers.

### DIFF
--- a/src/celeritas/global/alongstep/detail/UniformFieldPropagatorFactory.hh
+++ b/src/celeritas/global/alongstep/detail/UniformFieldPropagatorFactory.hh
@@ -25,7 +25,6 @@ struct UniformFieldPropagatorFactory
 {
 #if CELER_USE_DEVICE
     inline static constexpr int max_block_size = 256;
-    inline static constexpr int min_warps_per_eu = 8;
 #endif
 
     CELER_FUNCTION decltype(auto) operator()(CoreTrackView const& track) const


esseivaju (approved):
I think the problem here is that the semantics of `min_warps_per_eu` in Celeritas are different for Cuda and HIP. For HIP, it is correct while for CUDA it is really `min_warps_per_cu`. We'd have the same problem the other way around if we switched to the CUDA `__launch_bounds__` semantics of `min_blocks_per_cu`.

esseivaju:
Regarding the absurdly long first step, I see that as well; the mempool backing `[cuda|hip]MallocAsync` doesn't have any memory allocated and needs to do syscall to do actual allocation so it is very slow.

image

sethrj:
Aha, that makes sense, thanks. But on Crusher it's taking almost 40s in some problems?!

I wonder if we could make the responsible kernels preallocate the necessary memory with the "about to run" kernel interface. Then at least the extra time would correctly show as setup rather than runtime.

sethrj:
Confirmed this fixes the crusher performance regression for the problems with field:
```diff
@@ -4149,13 +4149,13 @@ true
 {
 "const_mem": 0,
 "heap_size": 68702699520,
-"local_mem": 720,
-"max_blocks_per_cu": 8,
+"local_mem": 16,
+"max_blocks_per_cu": 2,
 "max_threads_per_block": 256,
-"max_warps_per_eu": 8,
+"max_warps_per_eu": 2,
 "name": "along-step-uniform-msc-propagate",
-"num_regs": 64,
-"occupancy": 1.0,
+"num_regs": 232,
+"occupancy": 0.25,
 "print_buffer_size": 0,
 "threads_per_block": 256
 },
```
and
```
-"total": 30.066421165
+"total": 17.710086015
```
where 7 seconds of that is the startup cost in the first step.

esseivaju:
Multiple seconds seem extremely slow, I'd like to see if it's only the memory operation taking this much time. What is the "about to run" interface?

sethrj:
Sorry, I wrote that without having the code in front of me. The `BeginRunActionInterface` defines `begin_run` functions that are called during the "Stepper" constructor right after the states are allocated. We currently use it for states that have data that depend on the number of actions (since that number isn't known at the construction of the action itself).

sethrj:
Also in the problems I'm running (the regression suite) we don't even *have* a step gather... so something else is taking the time. @tmdelellis and I will be looking at this today.

PR #970, opened 2023-10-09
Implement ScopedProfiling for HIP/ROCM

esseivaju:
rocTX has a very small [API](https://rocm.docs.amd.com/projects/roctracer/en/latest/roctracer_spec.html#roctx-application-code-annotation-api) compared to NVTX, it only supports marker and rangepush/pop, rangestart/end.


src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -0,0 +1,87 @@
...
+#include "corecel/io/Logger.hh"
+
+#include "Environment.hh"
+
+/**
+ * @file
+ *
+ * The roctx implementation of \c ScopedProfiling
+ */
+
```
sethrj: Just add this as a `//! \brief` below line 6

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -0,0 +1,87 @@
...
+namespace celeritas
+{
+namespace
+{
+
+//---------------------------------------------------------------------------//
+}  // namespace
+
```
sethrj:
```suggestion
```

sethrj (commented):
I'm curious to see if the HIP build passes: from https://rocm.docs.amd.com/projects/roctracer/en/latest/#roctx-api it looked like we might have to link with some extra library?

esseivaju:
> I'm curious to see if the HIP build passes: from https://rocm.docs.amd.com/projects/roctracer/en/latest/#roctx-api it looked like we might have to link with some extra library?

yeah, we'll have to link the library. It looks like we can use a cmake [target](https://github.com/ROCm-Developer-Tools/roctracer/blob/fb31233747c27202a04c4c6d9746ddee717baaf6/src/CMakeLists.txt#L187) for it but I can't import it from the default hip installation.

src/corecel/CMakeLists.txt
```diff
@@ -65,6 +65,14 @@ if(CELERITAS_USE_CUDA)
   )
 endif()
 
+if(CELERITAS_USE_HIP)
+  list(APPEND SOURCES
+    sys/ScopedProfiling.hip.cc
+  )
+  find_library(ROCTX_LIBRARY roctx64 REQUIRED HINTS ${ROCM_PATH}/lib64)
```
sethrj: This might be better as part of the `celeritas_device_toolkit` (see CMakeLists in the paernt directory). Also, and I'm assuming the answer is no, but is there a significant reason *not* to link this in when building HIP? I.e., does it inject any symbols that change the behavior of the HIP API functions, or could it be missing from some HIP installations? (In which case we need extra compile-time logic for the profile functions...)

sethrj (approved):
Looks great @esseivaju ! Your call if you want to consider the comment below.

src/corecel/CMakeLists.txt
esseivaju: Do you mean linking hip against roctx64? I don't think there's an option to do that. It's definitely possible that this library is missing from some installation as it has to be installed separately (at least from my experience installing hip from spack and from [here](https://rocm.docs.amd.com/en/latest/understand/cmake_packages.html#rocm-cmake-packages)) 

We probably should make it optional and have the ScopedProfiling print a warning if the library was not found?

sethrj: Yeah. OK if it has to be installed separately, we'll need to have some configuration options that describe the system setup. Let's do one of these two:
-  create a `celeritas_sys_config.h.in` using `celeritas_config.h.in` as a reference, and define `CELERITAS_HAVE_ROCTX` there (in the future we may have to add more special conditions for optional libraries)
- add a private `target_compile_definitions` with `-DCELERITAS_HAVE_ROCTX=1`? since we probably only need these configuration options inside corecel.

esseivaju: Should be good now, sorry for all the CI triggers 😅

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * This is true only if the \c CELER_ENABLE_PROFILING environment variable is
+ * set to a non-empty value.
+ */
+bool ScopedProfiling::enable_profiling()
+{
+    static bool const result = [] {
+        if (!celeritas::getenv("CELER_ENABLE_PROFILING").empty())
+        {
+            CELER_LOG(info) << "Enabling profiling support since the "
```
sethrj: wrap the info message with `if (CELERITAS_HAVE_ROCTX) { ... }`  and `else { CELER_LOG(warning) ...`

CMakeLists.txt
```diff
@@ -359,6 +359,13 @@ elseif(CELERITAS_USE_HIP)
   target_include_directories(celeritas_device_toolkit
     SYSTEM INTERFACE "${ROCM_PATH}/include"
   )
+  find_library(ROCTX_LIBRARY roctx64 HINTS ${ROCM_PATH}/lib64)
+  if(ROCTX_LIBRARY)
+    set(CELERITAS_HAVE_ROCTX ON)
+    target_link_libraries(celeritas_device_toolkit INTERFACE ${ROCTX_LIBRARY})
+  else()
+    message(STATUS "Roctx library not found. ScopedProfiling has no effect")
```
sethrj: Replace the configure-time warning (and *every reconfigure*-time warning) with the runtime warning that shows up only when the user requests profiling.

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -0,0 +1,85 @@
...
+//! \brief The roctx implementation of \c ScopedProfiling
+//---------------------------------------------------------------------------//
+
+#include "ScopedProfiling.hh"
+
+#include <roctracer/roctx.h>
+
+#include "corecel/io/Logger.hh"
+
+#include "celeritas_sys_config.h"
```
sethrj, line 16: Add this to the top-level `.clang-format` include sorting oder regex

esseivaju: This is already included in the regex no? There is the following rule:
```
  - Regex:           '^"celeritas_[a-z_]+\.h"'
 ```

File: CMakeLists.txt [hunk 7]
```diff
@@ -359,12 +359,10 @@ elseif(CELERITAS_USE_HIP)
   target_include_directories(celeritas_device_toolkit
     SYSTEM INTERFACE "${ROCM_PATH}/include"
   )
-  find_library(ROCTX_LIBRARY roctx64 HINTS ${ROCM_PATH}/lib64)
```
esseivaju: I removed the hint, I don't think it was necessary as CMAKE was already searching in that directory by default

src/corecel/sys/ScopedProfiling.hip.cc
sethrj: Oh ok, that's right. I thought from the output that this was out of order.

PR #975, opened 2023-10-11
Improve inlining and warning messages for scoped profiling

sethrj:
This reduces a bit of the code duplication by checking for "enabled" in the header. We also now disable profiling and warn if the device is unavailable or disabled.

### DIFF
--- a/src/corecel/sys/ScopedProfiling.cuda.cc
+++ b/src/corecel/sys/ScopedProfiling.cuda.cc
@@ -134,10 +134,16 @@ bool ScopedProfiling::enable_profiling()
 /*!
  * Activate nvtx profiling with options.
  */
-void ScopedProfiling::activate_(Input const& input)
+void ScopedProfiling::activate_(Input const& input) noexcept
 {
     nvtxEventAttributes_t attributes_ = make_attributes(input);
-    nvtxDomainRangePushEx(domain_handle(), &attributes_);
+    int result = nvtxDomainRangePushEx(domain_handle(), &attributes_);
+    if (result < 0)
+    {
+        activated_ = false;
+        CELER_LOG(warning) << "Failed to activate profiling domain '"
+                           << input.name << "'";
+    }
 }
 
 //---------------------------------------------------------------------------//
@@ -146,7 +152,11 @@ void ScopedProfiling::activate_(Input const& input)
  */
 void ScopedProfiling::deactivate_() noexcept
 {
-    nvtxDomainRangePop(domain_handle());
+    int result = nvtxDomainRangePop(domain_handle());
+    if (result < 0)
+    {
+        CELER_LOG(warning) << "Failed to deactivate profiling domain";
+    }
 }
 
 //---------------------------------------------------------------------------//
--- a/src/corecel/sys/ScopedProfiling.cuda.cc
+++ b/src/corecel/sys/ScopedProfiling.cuda.cc
@@ -4,8 +4,8 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file corecel/sys/ScopedProfiling.cuda.cc
+//! \brief The nvtx implementation of \c ScopedProfiling
 //---------------------------------------------------------------------------//
-
 #include "ScopedProfiling.hh"
 
 #include <mutex>
@@ -15,7 +15,6 @@
 
 #include "corecel/io/Logger.hh"
 
-#include "Device.hh"
 #include "Environment.hh"
 
 namespace celeritas
@@ -114,12 +113,6 @@ bool ScopedProfiling::enable_profiling()
     static bool const result = [] {
         if (!celeritas::getenv("CELER_ENABLE_PROFILING").empty())
         {
-            if (!celeritas::device())
-            {
-                CELER_LOG(warning) << "Disabling profiling support "
-                                      "since no device is available";
-                return false;
-            }
             CELER_LOG(info) << "Enabling profiling support since the "
                                "'CELER_ENABLE_PROFILING' "
                                "environment variable is present and non-empty";
--- a/src/corecel/sys/ScopedProfiling.cuda.cc
+++ b/src/corecel/sys/ScopedProfiling.cuda.cc
@@ -132,34 +132,19 @@ bool ScopedProfiling::enable_profiling()
 /*!
  * Activate nvtx profiling with options.
  */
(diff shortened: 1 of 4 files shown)


src/corecel/sys/ScopedProfiling.cuda.cc
```diff
@@ -132,33 +134,28 @@ bool ScopedProfiling::enable_profiling()
 /*!
  * Activate nvtx profiling with options.
  */
-ScopedProfiling::ScopedProfiling(Input input)
+void ScopedProfiling::activate_(Input const& input) noexcept
```
pcanal, line 130: Why the unusual naming convention (looks like data member rather than function member).

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -32,19 +34,21 @@ bool ScopedProfiling::enable_profiling()
     static bool const result = [] {
         if (!celeritas::getenv("CELER_ENABLE_PROFILING").empty())
         {
-            if (CELERITAS_HAVE_ROCTX)
+            if (!celeritas::device())
```
esseivaju: It could still be useful to allow the use of ScopedProfiling in a device build with the environment variable disabling the device for CPU profiling.

esseivaju (approved):
Looks good to me, just one suggestion.

src/corecel/sys/ScopedProfiling.hip.cc
sethrj: Oh, so the profilers will still gather useful CPU only data? That would indeed be useful. I'll revert that change.

src/corecel/sys/ScopedProfiling.cuda.cc
sethrj: I know we discussed this before (but maybe we didn't reach a conclusion or we ended up avoiding this)? The trailing underscore means "private" whether it's a data member or function member. You could technically have a std::function or other function-like data member that you'd call with `do_something_()`. I added another instance of this in #933, but since we didn't discuss there maybe we decided to avoid this construct and I forgot about that decision... can you dig up the past discussion? We can add the final decision to the development rst.

src/corecel/sys/ScopedProfiling.hip.cc
esseivaju: It doesn't collect that much data, for purely CPU profiling it'd be better to use another profiler, but it will still show the scope on the timeline, so it's possible to get the time spent in different functions.

src/corecel/sys/ScopedProfiling.cuda.cc
pcanal: I'll try to look it up.  In the meantime this line of code:
```
   this->activate_(input);
```
is ambiguous on whether the action is on the current object or (`operator()` called) on a data member named `activate_`.  i.e. the 'scope' of the action can not be determine without looking up the declaration (which naming convention are supposed to help avoid :) ).

pcanal: I can't find the conversion.  I remember it did only indirectly talk about public vs private.  Instead it focused on data object/struct (i.e. public) vs active object (i.e. private) and than indeed there was no ambiguity between:
```
struct ABC { int somedata; }; 
ABC value;
```
which then *always* (since there is no member function) as:
```
value.somedata;
```
so it can not be confused with a local variable, so it is okay in this case to not match the `data member` name convention.

As shown [above](https://github.com/celeritas-project/celeritas/pull/975/files#r1355589375), the same can not be said (and I don't recall we discussed it) for private data members vs private member function.

The resulting usage can be identical (if the data member's type has an `operator()` for example) but, of course, the implication on the scope of the action is completely different.

I.e. it is in my opinion worth to lexicographically differentiate them (private data member and private member function)

sethrj: I don't understand `the implication on the scope of the action is completely different`, and I think the `SoftSurfaceEq` makes the perfect counterexample for our recommendation:
```c++
    SoftEqual<> soft_eq_;

    bool soft_eq_sq_(real_type a, real_type b) const;
    bool soft_eq_distance_(Real3 const& a, Real3 const& b) const;
```

If we write a SoftEqSquare functor that does the same thing as `soft_eq_sq_`, we'll still be calling `this-> soft_eq_sq_(a, b);`. The "scope" of the action should be no more important than if a member function is const or non-const, or whether it has data that points to global data, etc. I think the convention of "this is private vs public" is the only important one.

sethrj: I think I understand by reading again. But I think the way you're using "scope" is imprecise. In both cases you know the thing being called is private and declared in the class. Anything further would have to be found by digging into the implementation in either case, so I don't see the point in differentiating between function-like member data versus member function.

pcanal: > Anything further would have to be found by digging into the implementation in either case,

Always.  The point of a naming convention is to reduce the digging necessary to understand and follow the code.

The distinction I see is upon seeing code like:
```
   this->activate_(input);
```
Do you know if any other data member of `this` might change or is it only the data member of `this->activate_` that might change?  Also to follow what code is being called do you look in the class of `this` or in the class of `this->activate_` ?   [ And of course my assumption will different from yours, reading this code I assume nothing but the content of the object `activate_` will change or depending on the time of the day, I might assume that this is a function pointer dispatch of some kind ]

sethrj: @pcanal  You could use the same argument to say we need to use different naming conventions between constant and mutable member functions, right?

pcanal: > between constant and mutable member functions, right?

You could and some naming conventions do.  The difference is that a data member and member function are of different nature (can't 'easily' change the code from to the other) while the other is an annotation (changed by "just" adding the `const` keyword).

sethrj: OK, that's fair. I think we could go either way on this but I'll concede here. I'll merge this as is but immediately add a new commit to change this instance, the existing one in SoftSurfaceEqual, and any other cases I might find. (I think we add `_impl` to some of the other "private" methods as a convention too, now that I htink about it.)

pcanal: > (I think we add _impl to some of the other "private" methods as a convention too, now that I think about it.)

That would work :)

PR #977, opened 2023-10-11
Refactor universe insertion and add universe accessors to orange params

sethrj:
This improves the construction of universe types (unit, rect array) from orange params by moving type-specific dispatch code from OrangeParams into the builders themselves.  Variants are super nice to work with, and this uses more of them.

### DIFF
--- a/src/orange/construct/DepthCalculator.hh
+++ b/src/orange/construct/DepthCalculator.hh
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <unordered_map>
+#include <vector>
 
 #include "corecel/cont/VariantUtils.hh"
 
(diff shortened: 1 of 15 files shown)


src/orange/construct/DepthCalculator.hh
```diff
@@ -0,0 +1,54 @@
...
+
+  public:
+    // Construct with a reference to all universe inputs
+    explicit DepthCalculator(VecVarUniv const&);
+
+    //! Calculate the depth of the global unit
+    size_type operator()() { return this->visit_univ_(*this, 0); }
+
+    // Calculate the depth of a unit
+    size_type operator()(UnitInput const& u);
```
elliottbiondo, line 39: The second two `operator()` functions can be made private

elliottbiondo (approved):

src/orange/construct/DepthCalculator.hh
sethrj: Actually, it can't, because they get called by `std::visit`'s low-level implementaiton. It's similar to how you can't make a private constructor if you use `std::make_shared`.

PR #978, opened 2023-10-11
Adapt to the new BVH navigator from VecGeom

mrguilima:
The AdePT BVH navigator was made available directly from VecGeom. This branch adapts Celeritas to use the BVHNavigator from VecGeom v2.x.

In case VecGeom v1.x  is used, based on the definition of VECGEOM_VERSION, the BVHNavigator from celeritas/ext/detail/BVHNavigator.hh will be used.

### DIFF
--- a/src/celeritas/ext/VecgeomTrackView.hh
+++ b/src/celeritas/ext/VecgeomTrackView.hh
@@ -211,16 +211,9 @@ VecgeomTrackView::operator=(Initializer_t const& init)
     vecgeom::VPlacedVolume const* worldvol = params_.world_volume;
     bool const contains_point = true;
 
-    // Note that LocateGlobalPoint sets `vgstate_`. If `vgstate_` is outside
-    // (including possibly on the outside volume edge), the volume pointer it
-    // returns would be null at this point.
-#if VECGEOM_VERSION >= 0x020000
+    // LocatePointIn sets `vgstate_`
     BVHNavigator::LocatePointIn(
         worldvol, detail::to_vector(pos_), vgstate_, contains_point);
-#else
-    BVHNavigator::LocatePointIn(
-        worldvol, detail::to_vector(pos_), vgstate_, contains_point);
-#endif
     return *this;
 }
 
--- a/src/celeritas/ext/VecgeomTrackView.hh
+++ b/src/celeritas/ext/VecgeomTrackView.hh
@@ -9,6 +9,7 @@
 
 #include <VecGeom/base/Config.h>
 #include <VecGeom/base/Cuda.h>
+#include <VecGeom/base/Version.h>
 #include <VecGeom/navigation/GlobalLocator.h>
 #include <VecGeom/navigation/NavStateFwd.h>
 #include <VecGeom/navigation/NavigationState.h>
@@ -22,9 +23,15 @@
 #include "orange/Types.hh"
 
 #include "VecgeomData.hh"
-#include "detail/BVHNavigator.hh"
 #include "detail/VecgeomCompatibility.hh"
 
+#if VECGEOM_VERSION >= 0x020000
+#    include <VecGeom/navigation/BVHNavigator.h>
+using BVHNavigator = vecgeom::BVHNavigator;
+#else
+#    include "detail/BVHNavigator.hh"
+using BVHNavigator = celeritas::detail::BVHNavigator;
+#endif
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
@@ -204,9 +211,13 @@ VecgeomTrackView::operator=(Initializer_t const& init)
     // Note that LocateGlobalPoint sets `vgstate_`. If `vgstate_` is outside
     // (including possibly on the outside volume edge), the volume pointer it
     // returns would be null at this point.
-    detail::BVHNavigator::LocatePointIn(
+#if VECGEOM_VERSION >= 0x020000
+    BVHNavigator::LocatePointIn(
         worldvol, detail::to_vector(pos_), vgstate_, contains_point);
-
+#else
+    BVHNavigator::LocatePointIn(
+        worldvol, detail::to_vector(pos_), vgstate_, contains_point);
+#endif
     return *this;
 }
 
@@ -316,12 +327,12 @@ CELER_FUNCTION Propagation VecgeomTrackView::find_next_step(real_type max_step)
     CELER_EXPECT(max_step > 0);
 
     // Use BVH navigator to find internal distance
-    next_step_ = detail::BVHNavigator::ComputeStepAndNextVolume(
-        detail::to_vector(pos_),
-        detail::to_vector(dir_),
-        max_step,
-        vgstate_,
-        vgnext_);
+    next_step_ = BVHNavigator::ComputeStepAndNextVolume(detail::to_vector(pos_),
+                                                        detail::to_vector(dir_),
+                                                        max_step,
+                                                        vgstate_,
+                                                        vgnext_);
+
     next_step_ = max(next_step_, this->extra_push());
     if (!this->is_next_boundary())
     {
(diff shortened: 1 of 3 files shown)


src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -365,12 +376,19 @@ CELER_FUNCTION real_type VecgeomTrackView::find_safety(real_type max_radius)
...
-    real_type safety = detail::BVHNavigator::ComputeSafety(
-        detail::to_vector(this->pos()), vgstate_, max_radius);
+
+    real_type safety = BVHNavigator::ComputeSafety(
+        detail::to_vector(this->pos()),
+        vgstate_
+#if VECGEOM_VERSION < 0x020000
+        ,
+        max_radius
+#endif
```
sethrj: I think we should backport this into their BVH, since it could accelerate safety calculations in field tracking, for example.

src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -22,9 +23,15 @@
 #include "orange/Types.hh"
 
 #include "VecgeomData.hh"
-#include "detail/BVHNavigator.hh"
 #include "detail/VecgeomCompatibility.hh"
 
+#if VECGEOM_VERSION >= 0x020000
+#    include <VecGeom/navigation/BVHNavigator.h>
+using BVHNavigator = vecgeom::BVHNavigator;
```
sethrj: Move these type aliases inside the class instead of the global namespace.

src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -9,6 +9,7 @@
 
 #include <VecGeom/base/Config.h>
 #include <VecGeom/base/Cuda.h>
+#include <VecGeom/base/Version.h>
```
sethrj, line 12: VecGeom 1.2.4 was the first one to include this header. I'm OK with this, but you'll need to update the version in the main CMakeLists.txt .

src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -365,12 +376,19 @@ CELER_FUNCTION real_type VecgeomTrackView::find_safety(real_type max_radius)
...
+#if VECGEOM_VERSION < 0x020000
+        ,
+        max_radius
+#endif
+    );
 
     // Since the reported "safety" is negative if we've moved slightly beyond
     // the boundary of a solid without crossing it, we must clamp to zero.
-    return max<real_type>(safety, 0);
+    return clamp(safety, 0., max_radius);
```
sethrj: This should be reverted. It's fine if we return a max radius that's larger. If this changes test results, those should be updated. (Also, you're assuming real_type is double here.)

sethrj (commented):
Looks pretty good, but I know that this isn't compatible with the current vecgeom master. What needs to be done to update it for that?

src/celeritas/ext/VecgeomTrackView.hh
mrguilima: I will do this and create a merge request in their repository.

mrguilima:
This MR *was meant* to be compatible with the master of VecGeom, and it works for my local tests  --  I used tag v2.0.0-dev.3.
The `celeritas_track_TrackSort` test is the only one which fails, but this is expected.

If it fails on your local tests, please let me know what the errors are.

src/celeritas/ext/VecgeomTrackView.hh
mrguilima: I've tentative set a minimum VecGeom version required, from 1.1.17 to 1.2.4.
Otherwise, I could not figure out a way to use VECGEOM_VERSION without #including that `<VecGeom/base/Version.h>` header.

sethrj:
Oh ok, that's good. I perhaps was misremembering the issue I had. Thanks Guilherme! I'll test it out locally.

test/celeritas/ext/Vecgeom.test.cc
```diff
@@ -139,11 +139,13 @@ class FourLevelsTest : public VecgeomTestBase
         return this->load_vgdml("four-levels.gdml");
     }
 
+#if VECGEOM_VERSION >= 0x020000
     SpanStringView expected_log_levels() const final
     {
         static std::string_view const levels[] = {"warning"};
         return make_span(levels);
     }
+#endif
```
sethrj: Just use `if (VECGEOM_VERSION >= 0x020000)` so we don't invoke the preprocessor; that's safer because it means the code inside the conditional has to be valid.

so:
```c++
{
if (VECGEOM_VERSION >= 0x020000)
{
    return {};
}
else
{
        static std::string_view const levels[] = {"warning"};
        return make_span(levels);
}
}
```

src/celeritas/ext/VecgeomTrackView.hh
sethrj: Yes, I think that's the right way. The only other way would be to add additional cmake logic that writes *another* vecgeom version, which is way too much work to support versions of vecgeom that are effectively broken anyway.

src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -204,9 +214,13 @@ VecgeomTrackView::operator=(Initializer_t const& init)
...
     // returns would be null at this point.
-    detail::BVHNavigator::LocatePointIn(
+#if VECGEOM_VERSION >= 0x020000
+    BVHNavigator::LocatePointIn(
         worldvol, detail::to_vector(pos_), vgstate_, contains_point);
-
+#else
+    BVHNavigator::LocatePointIn(
+        worldvol, detail::to_vector(pos_), vgstate_, contains_point);
+#endif
```
sethrj: These two calls look identical, shouldn't you just remove the `detail::` and it'll work for both cases?

src/celeritas/ext/VecgeomTrackView.hh
```diff
@@ -365,12 +379,19 @@ CELER_FUNCTION real_type VecgeomTrackView::find_safety(real_type max_radius)
...
+#if VECGEOM_VERSION < 0x020000
+        ,
+        max_radius
+#endif
+    );
 
     // Since the reported "safety" is negative if we've moved slightly beyond
     // the boundary of a solid without crossing it, we must clamp to zero.
-    return max<real_type>(safety, 0);
+    return max<real_type>(0., min<real_type>(safety, max_radius));
```
sethrj: Leave this as is:
```suggestion
    return max<real_type>(safety, 0);
```
And just change the `real_type safety = BVHNavigator::ComputeSafety` call to have a `min` inside the `VECGEOM_VERSION < 0x020000` conditional.

test/celeritas/ext/Vecgeom.test.cc
```diff
@@ -473,7 +481,12 @@ class SolidsTest : public VecgeomTestBase
 
     SpanStringView expected_log_levels() const final
     {
+#if VECGEOM_VERSION >= 0x020000
+        static std::string_view const levels[]
+            = {"warning", "warning", "warning"};
+#else
         static std::string_view const levels[] = {"warning"};
+#endif
```
sethrj: Use `if` here

test/celeritas/ext/Vecgeom.test.cc
```diff
@@ -782,6 +795,14 @@ class CmseTest : public VecgeomTestBase
...
   public:
     SPConstGeo build_geometry() final { return this->load_vgdml("cmse.gdml"); }
+
+#if VECGEOM_VERSION >= 0x020000
+    SpanStringView expected_log_levels() const final
+    {
+        static std::string_view const levels[] = {"warning"};
+        return make_span(levels);
+    }
+#endif
```
sethrj: Use `if` here too

src/celeritas/ext/VecgeomTrackView.hh
mrguilima: Your suggestion will make the `max_radius` ineffective for VecGeom v1.x.  Kept as is for now.

sethrj (approved):
Thanks @mrguilima !

PR #981, opened 2023-10-13
Support HIP 5.1.0 and add Frontier build scripts

sethrj:
This adds build scripts for the Frontier toolchain I built using ROCm 5.1. I'll next look at adding a 5.3/5.5 toolchain since that's what @tmdelellis has been using more recently.

### DIFF
--- a/src/corecel/sys/Stream.cc
+++ b/src/corecel/sys/Stream.cc
@@ -15,8 +15,51 @@
 #include "corecel/io/Logger.hh"
 #include "celeritas/Types.hh"
 
+#if CELERITAS_USE_CUDA
+#    define CELER_STREAM_SUPPORTS_ASYNC 1
+#elif CELERITAS_USE_HIP       \
+    && (HIP_VERSION_MAJOR > 5 \
+        || (HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR >= 2))
+#    define CELER_STREAM_SUPPORTS_ASYNC 1
+#else
+#    define CELER_STREAM_SUPPORTS_ASYNC 0
+#endif
+
 namespace celeritas
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+//! Malloc asynchronously for CUDA and newer HIP versions
+void* malloc_async_impl(std::size_t bytes, Stream::StreamT s)
+{
+    void* ptr{};
+#if CELER_STREAM_SUPPORTS_ASYNC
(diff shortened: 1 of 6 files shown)


src/corecel/sys/Stream.cc
```diff
@@ -15,8 +15,51 @@
...
 
+#if CELERITAS_USE_CUDA
+#    define CELER_STREAM_SUPPORTS_ASYNC 1
+#elif CELERITAS_USE_HIP       \
+    && (HIP_VERSION_MAJOR > 5 \
+        || (HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR >= 2))
+#    define CELER_STREAM_SUPPORTS_ASYNC 1
+#else
+#    define CELER_STREAM_SUPPORTS_ASYNC 0
+#endif
```
esseivaju, line 26: One suggestion would be to use more constexpr and if constexpr for conditional compilation when possible, e.g. 
```c++
#if CELERITAS_USE_CUDA
constexpr inline bool celer_stream_supports_async = true;
#elif CELERITAS_USE_HIP
constexpr inline bool celer_stream_supports_async = HIP_VERSION_MAJOR > 5 || (HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR >= 2);
#endif
```

esseivaju (approved):
LGTM

src/corecel/sys/Stream.cc
sethrj: I don't think that works here because the code inside the `if constexpr` brackets has to be valid: HIP 5.1 will still see a function call to a nonexistent `hipMallocAsync` and error, even though it's not evaluated. I think at this point in time, you still *have* to use macros for certain API changes.

esseivaju: ~~Right, but that's only in nontemplate code. I think this would work but macros aren't that bad 😁 so I'm not convinced this is better:~~ Nvm, it's just that statements in the discarded branch are not instantiated, it doesn't mean that we can use names that don't exists.
```c++
template<bool supports_async = celer_stream_supports_async>
void* malloc_async_impl(std::size_t bytes, Stream::StreamT s){
    void* ptr{};
    if constexpr (supports_async){
        CELER_DEVICE_CALL_PREFIX(MallocAsync(&ptr, bytes, s));
    } else {
        CELER_DISCARD(bytes);
        CELER_DISCARD(s);
        CELER_DEVICE_CALL_PREFIX(Malloc(&ptr, bytes));
    }
        return ptr;
}
```

PR #982, opened 2023-10-16
Add more profiling ranges and fix return code checks

sethrj:
I didn't think to check #975 in a real use case, but it seems that NVTX (at least in CUDA 11.5 on wildstyle) always returns -1 for the push/pop, even though they register correctly. This fixes that.

I've also extended the scope of the profiling ranges so we can see e.g. the slow spots in loading Geant4 and VecGeom, as well as the cost of initializing.

Finally, I added a "begin run" setup for `extend_from_secondaries` so that the initial `cudaMallocAsync` is moved outside the initial step.

zoom-full

zoom-construct

zoom-step

### DIFF
--- a/src/corecel/sys/ScopedProfiling.cuda.cc
+++ b/src/corecel/sys/ScopedProfiling.cuda.cc
@@ -21,21 +21,6 @@ namespace celeritas
 {
 namespace
 {
-
-//---------------------------------------------------------------------------//
-/*!
- * Global registry for strings used by NVTX.
- *
- * This is implemented as a free function instead of a class static member in
- * ScopedProfiling to hide the NVTX dependency from users of the
- * interface.
- */
-std::unordered_map<std::string, nvtxStringHandle_t>& message_registry()
-{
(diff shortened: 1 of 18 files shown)


src/corecel/sys/ScopedProfiling.cuda.cc
```diff
@@ -54,25 +54,27 @@ nvtxDomainHandle_t domain_handle()
...
     // We did not find the handle; try to insert it
-    auto [iter, inserted] = [&message] {
+    auto [iter, inserted] = [&message, &registry] {
         std::unique_lock lock(mutex);
-        return message_registry().insert({message, {}});
+        return registry.insert({message, {}});
     }();
     if (inserted)
     {
+        // Register the
```
esseivaju: domain

esseivaju (approved):
Looks good to me! On a side note, I've seen some crashes during the application teardown when running through a multi-threaded Geant4 app with profiling enabled, e.g. the atlas tilecal. Have you had this issue as well?

sethrj:
No, I haven't done much testing with the geant4 standalone apps...

PR #984, opened 2023-10-16
Improve performance with lower launch bounds

sethrj:
There should be no advantage to using the same environment-controlled "threads per block" but allowing the CUDA/ROCm compiler to "allow" a higher maximum launch count. This PR removes the environment variable and adds a CMake cache variable `CELERITAS_MAX_BLOCK_SIZE` that sets the default block size for kernels executed through `ActionLauncher`. The KernelAttributes collected during the first kernel launch are stored persistently as the block size.

## Occupancy

The occupancy increases for ORANGE boundary crossing and pre-step kernels and a couple others. Strangely the register usage decreases slightly in a few kernels despite no change in the occupancy...

Occupancy before and after show much better utilization on AMD:
occupancy-before

occpuancy-after

arch/geo/kernel/local spill (B)/register usage (B)/occupancy:
```diff
diff --git a/kernel-occupancy.md b/kernel-occupancy.md
index 857f705..f12508f 100644
--- a/kernel-occupancy.md
+++ b/kernel-occupancy.md
@@ -13,56 +13,56 @@
 | cuda | vecgeom | physics-discrete-select                     |     0 |      256 |     0.500 |
 | cuda | vecgeom | ioni-moller-bhabha                          |    48 |      236 |     0.500 |
 | cuda | vecgeom | brems-sb                                    |    48 |      304 |     0.375 |
-| cuda | vecgeom | brems-rel                                   |    48 |      296 |     0.375 |
+| cuda | vecgeom | brems-rel                                   |    48 |      440 |     0.250 |
 | cuda | vecgeom | photoel-livermore                           |    64 |      280 |     0.375 |
-| cuda | vecgeom | scat-klein-nishina                          |    48 |      232 |     0.500 |
+| cuda | vecgeom | scat-klein-nishina                          |    48 |      224 |     0.500 |
 | cuda | vecgeom | conv-bethe-heitler                          |    48 |      288 |     0.375 |
-| cuda | vecgeom | scat-rayleigh                               |   104 |      264 |     0.375 |
+| cuda | vecgeom | scat-rayleigh                               |   104 |      256 |     0.500 |
 | cuda | vecgeom | annihil-2-gamma                             |    48 |      224 |     0.500 |
 | cuda | vecgeom | geo-boundary                                |     0 |      880 |     0.125 |
-| cuda | vecgeom | extend-from-secondaries                     |     0 |      116 |     1.000 |
-| cuda | vecgeom | extend-from-secondaries                     |     0 |      176 |     0.625 |
+| cuda | vecgeom | extend-from-secondaries-locate-alive        |     0 |      116 |     1.000 |
+| cuda | vecgeom | extend-from-secondaries-process-secondaries |     0 |      176 |     0.625 |
 | cuda | orange  | extend-from-primaries                       |     0 |      112 |     1.000 |
 | cuda | orange  | initialize-tracks                           |   112 |      256 |     0.500 |
-| cuda | orange  | pre-step                                    |     0 |      260 |     0.375 |
-| cuda | orange  | along-step-uniform-msc-limit-step-msc-urban |    48 |      256 |     0.500 |
+| cuda | orange  | pre-step                                    |     0 |      256 |     0.500 |
+| cuda | orange  | along-step-uniform-msc-limit-step-msc-urban |    48 |      264 |     0.375 |
 | cuda | orange  | along-step-uniform-msc-propagate            |   240 |      512 |     0.250 |
-| cuda | orange  | along-step-uniform-msc-scatter-msc-urban    |    64 |      268 |     0.375 |
+| cuda | orange  | along-step-uniform-msc-scatter-msc-urban    |    64 |      296 |     0.375 |
 | cuda | orange  | along-step-uniform-msc-update-time          |     0 |       96 |     1.000 |
-| cuda | orange  | along-step-uniform-msc-apply-eloss-mean     |     0 |      224 |     0.500 |
+| cuda | orange  | along-step-uniform-msc-apply-eloss-mean     |     0 |      216 |     0.500 |
 | cuda | orange  | along-step-uniform-msc-update-track         |     0 |       64 |     1.000 |
-| cuda | orange  | along-step-neutral                          |     0 |      512 |     0.250 |
-| cuda | orange  | physics-discrete-select                     |     0 |      240 |     0.500 |
-| cuda | orange  | ioni-moller-bhabha                          |    64 |      240 |     0.500 |
-| cuda | orange  | brems-sb                                    |    64 |      320 |     0.375 |
-| cuda | orange  | brems-rel                                   |    64 |      308 |     0.375 |
+| cuda | orange  | along-step-neutral                          |     0 |      508 |     0.250 |
+| cuda | orange  | physics-discrete-select                     |     0 |      236 |     0.500 |
+| cuda | orange  | ioni-moller-bhabha                          |    64 |      236 |     0.500 |
+| cuda | orange  | brems-sb                                    |    64 |      304 |     0.375 |
+| cuda | orange  | brems-rel                                   |    64 |      292 |     0.375 |
 | cuda | orange  | photoel-livermore                           |    80 |      272 |     0.375 |
-| cuda | orange  | scat-klein-nishina                          |    64 |      236 |     0.500 |
-| cuda | orange  | conv-bethe-heitler                          |    64 |      296 |     0.375 |
-| cuda | orange  | scat-rayleigh                               |   104 |      264 |     0.375 |
-| cuda | orange  | annihil-2-gamma                             |    64 |      224 |     0.500 |
-| cuda | orange  | geo-boundary                                |   152 |      320 |     0.375 |
-| cuda | orange  | extend-from-secondaries                     |     0 |      116 |     1.000 |
-| cuda | orange  | extend-from-secondaries                     |     0 |      188 |     0.625 |
+| cuda | orange  | scat-klein-nishina                          |    64 |      232 |     0.500 |
+| cuda | orange  | conv-bethe-heitler                          |    64 |      272 |     0.375 |
+| cuda | orange  | scat-rayleigh                               |   104 |      252 |     0.500 |
+| cuda | orange  | annihil-2-gamma                             |    64 |      232 |     0.500 |
+| cuda | orange  | geo-boundary                                |   176 |      256 |     0.500 |
+| cuda | orange  | extend-from-secondaries-locate-alive        |     0 |      116 |     1.000 |
+| cuda | orange  | extend-from-secondaries-process-secondaries |     0 |      188 |     0.625 |
 | hip  | orange  | extend-from-primaries                       |     0 |       48 |     1.000 |
 | hip  | orange  | initialize-tracks                           |   120 |      360 |     0.625 |
 | hip  | orange  | pre-step                                    |     0 |      336 |     0.625 |
 | hip  | orange  | along-step-uniform-msc-limit-step-msc-urban |     0 |      452 |     0.500 |
 | hip  | orange  | along-step-uniform-msc-propagate            |    48 |      944 |     0.250 |
-| hip  | orange  | along-step-uniform-msc-scatter-msc-urban    |     0 |      452 |     0.250 |
+| hip  | orange  | along-step-uniform-msc-scatter-msc-urban    |     0 |      452 |     0.500 |
 | hip  | orange  | along-step-uniform-msc-update-time          |     0 |       80 |     1.000 |
 | hip  | orange  | along-step-uniform-msc-apply-eloss-mean     |     0 |      172 |     1.000 |
 | hip  | orange  | along-step-uniform-msc-update-track         |     0 |       44 |     1.000 |
-| hip  | orange  | along-step-neutral                          |    64 |      512 |     0.500 |
+| hip  | orange  | along-step-neutral                          |    16 |      640 |     0.375 |
 | hip  | orange  | physics-discrete-select                     |     0 |      248 |     1.000 |
-| hip  | orange  | ioni-moller-bhabha                          |    48 |      244 |     0.250 |
-| hip  | orange  | brems-sb                                    |    48 |      404 |     0.250 |
-| hip  | orange  | brems-rel                                   |    48 |      456 |     0.250 |
+| hip  | orange  | ioni-moller-bhabha                          |    48 |      244 |     1.000 |
+| hip  | orange  | brems-sb                                    |    48 |      404 |     0.500 |
+| hip  | orange  | brems-rel                                   |    48 |      496 |     0.500 |
 | hip  | orange  | photoel-livermore                           |    48 |      240 |     0.875 |
-| hip  | orange  | scat-klein-nishina                          |     0 |      272 |     0.250 |
+| hip  | orange  | scat-klein-nishina                          |     0 |      272 |     0.875 |
 | hip  | orange  | conv-bethe-heitler                          |     0 |      312 |     0.750 |
-| hip  | orange  | scat-rayleigh                               |    64 |      328 |     0.250 |
+| hip  | orange  | scat-rayleigh                               |    64 |      328 |     0.625 |
 | hip  | orange  | annihil-2-gamma                             |     0 |      312 |     0.750 |
 | hip  | orange  | geo-boundary                                |   304 |      368 |     0.625 |
-| hip  | orange  | extend-from-secondaries                     |     0 |      212 |     1.000 |
-| hip  | orange  | extend-from-secondaries                     |     0 |      240 |     0.250 |
+| hip  | orange  | extend-from-secondaries-locate-alive        |     0 |      212 |     1.000 |
+| hip  | orange  | extend-from-secondaries-process-secondaries |    35 |      256 |     1.000 |
```

## Performance

Unfortunately the Frontier performance statistics are borked by the first step, but the Summit results show ~10% improvement with ORANGE (but no change with VecGeom).

speedup-after

```diff
diff --git a/results/summit/speedup.md b/results/summit/speedup.md
index 786deef..057661e 100644
--- a/results/summit/speedup.md
+++ b/results/summit/speedup.md
@@ -1,19 +1,19 @@
 | Problem                      | Geometry |      Speedup |
 | ---------------------------- | -------- | ------------ |
-| cms2018 [Z]                  | vecgeom  |  9.6× (±0.1) |
+| cms2018 [Z]                  | vecgeom  |  9.7× (±0.1) |
 | cms2018+field+msc [ZFM]      | vecgeom  |  6.7× (±0.1) |
-| simple-cms+field [BF]        | orange   | 24.8× (±2.3) |
-| simple-cms+field+msc [BFM]   | orange   | 29.0× (±1.9) |
-|                              | vecgeom  | 27.8× (±1.6) |
-| simple-cms+msc [BM]          | orange   | 26.5× (±0.2) |
-| testem15 [A]                 | orange   | 22.4× (±2.4) |
-| testem15+field [AF]          | orange   | 25.5× (±2.6) |
+| simple-cms+field [BF]        | orange   | 30.0× (±1.7) |
+| simple-cms+field+msc [BFM]   | orange   | 33.3× (±1.5) |
+|                              | vecgeom  | 29.7× (±1.2) |
+| simple-cms+msc [BM]          | orange   | 29.8× (±0.5) |
+| testem15 [A]                 | orange   | 24.2× (±0.1) |
+| testem15+field [AF]          | orange   | 32.4× (±2.3) |
 |                              | vecgeom  |  nan× (±nan) |
-| testem15+field+msc [AFM]     | orange   | 32.1× (±3.2) |
-|                              | vecgeom  | 31.0× (±2.1) |
-| testem3-flat [C]             | orange   | 24.2× (±0.3) |
-|                              | vecgeom  | 25.6× (±2.0) |
-| testem3-flat+field [CF]      | orange   | 23.0× (±0.6) |
-| testem3-flat+field+msc [CFM] | orange   | 26.9× (±1.1) |
-|                              | vecgeom  | 15.5× (±0.3) |
-| testem3-flat+msc [CM]        | orange   | 33.7× (±1.4) |
+| testem15+field+msc [AFM]     | orange   | 34.2× (±3.1) |
+|                              | vecgeom  | 32.0× (±2.1) |
+| testem3-flat [C]             | orange   | 27.1× (±0.2) |
+|                              | vecgeom  | 24.8× (±1.5) |
+| testem3-flat+field [CF]      | orange   | 24.8× (±0.7) |
+| testem3-flat+field+msc [CFM] | orange   | 29.3× (±0.6) |
+|                              | vecgeom  | 15.3× (±0.2) |
+| testem3-flat+msc [CM]        | orange   | 36.1× (±0.9) |
```

### DIFF
--- a/app/demo-interactor/KNDemoKernel.cu
+++ b/app/demo-interactor/KNDemoKernel.cu
@@ -193,7 +193,7 @@ __global__ void cleanup_kernel(DeviceCRef<ParamsData> const params,
 // KERNEL INTERFACES
 //---------------------------------------------------------------------------//
 #define CDE_LAUNCH_KERNEL(NAME, BLOCK_SIZE, THREADS, ...) \
-    CELER_LAUNCH_KERNEL(NAME, BLOCK_SIZE, THREADS, 0, __VA_ARGS__)
+    CELER_LAUNCH_KERNEL(NAME, THREADS, 0, __VA_ARGS__)
 
 /*!
  * Initialize particle states.
--- a/app/demo-interactor/KNDemoKernel.cu
+++ b/app/demo-interactor/KNDemoKernel.cu
@@ -192,9 +192,6 @@ __global__ void cleanup_kernel(DeviceCRef<ParamsData> const params,
 //---------------------------------------------------------------------------//
 // KERNEL INTERFACES
 //---------------------------------------------------------------------------//
-#define CDE_LAUNCH_KERNEL(NAME, BLOCK_SIZE, THREADS, ...) \
-    CELER_LAUNCH_KERNEL(NAME, THREADS, 0, __VA_ARGS__)
-
 /*!
  * Initialize particle states.
  */
@@ -205,12 +202,7 @@ void initialize(DeviceGridParams const& opts,
 {
     CELER_EXPECT(states.alive.size() == states.size());
     CELER_EXPECT(states.rng.size() == states.size());
-    CDE_LAUNCH_KERNEL(initialize,
-                      opts.threads_per_block,
-                      states.size(),
-                      params,
-                      states,
-                      initial);
+    CELER_LAUNCH_KERNEL(initialize, states.size(), 0, params, states, initial);
 }
 
 //---------------------------------------------------------------------------//
@@ -222,12 +214,10 @@ void iterate(DeviceGridParams const& opts,
              DeviceRef<StateData> const& states)
 {
     // Move to the collision site
-    CDE_LAUNCH_KERNEL(
-        move, opts.threads_per_block, states.size(), params, states);
+    CELER_LAUNCH_KERNEL(move, states.size(), 0, params, states);
 
     // Perform the interaction
-    CDE_LAUNCH_KERNEL(
-        interact, opts.threads_per_block, states.size(), params, states);
+    CELER_LAUNCH_KERNEL(interact, states.size(), 0, params, states);
 
     if (opts.sync)
     {
@@ -245,14 +235,11 @@ void cleanup(DeviceGridParams const& opts,
              DeviceRef<StateData> const& states)
 {
     // Process hits from buffer to grid
-    CDE_LAUNCH_KERNEL(process_hits,
-                      opts.threads_per_block,
-                      states.detector.capacity(),
-                      params,
-                      states);
+    CELER_LAUNCH_KERNEL(
+        process_hits, states.detector.capacity(), 0, params, states);
 
     // Clear buffers
-    CDE_LAUNCH_KERNEL(cleanup, device().threads_per_warp(), 1, params, states);
+    CELER_LAUNCH_KERNEL(cleanup, 1, 0, params, states);
 
     if (opts.sync)
     {
--- a/src/corecel/sys/Device.cc
+++ b/src/corecel/sys/Device.cc
@@ -156,12 +156,14 @@ Device::Device(int id) : id_{id}, streams_{new detail::StreamStorage{}}
     unsigned int max_threads_per_block = 0;
 #if CELER_USE_DEVICE
 #    if CELERITAS_USE_CUDA
-    cudaDeviceProp props;
+    using DevicePropT = cudaDeviceProp;
 #    elif CELERITAS_USE_HIP
-    hipDeviceProp_t props;
+    using DevicePropT = hipDeviceProp_t;
 #    endif
 
+    DevicePropT props;
     CELER_DEVICE_CALL_PREFIX(GetDeviceProperties(&props, id));
+
     name_ = props.name;
     total_global_mem_ = props.totalGlobalMem;
     max_threads_per_block_ = props.maxThreadsDim[0];
--- a/src/corecel/sys/Device.cc
+++ b/src/corecel/sys/Device.cc
@@ -223,22 +223,31 @@ Device::Device(int id) : id_{id}, streams_{new detail::StreamStorage{}}
 
     // See device_runtime_api.h
     eu_per_cu_ = CELER_EU_PER_CU;
+    // Default to the configuration-time option
+    default_block_size_ = CELERITAS_MAX_BLOCK_SIZE;
 
     // Set default block size from environment
     std::string const& bsize_str = celeritas::getenv("CELER_BLOCK_SIZE");
     if (!bsize_str.empty())
     {
         default_block_size_ = std::stoi(bsize_str);
-        CELER_VALIDATE(default_block_size_ >= threads_per_warp_
-                           && default_block_size_ <= max_threads_per_block,
-                       << "Invalid block size: number of threads must be in ["
-                       << threads_per_warp_ << ", " << max_threads_per_block
-                       << "]");
-        CELER_VALIDATE(default_block_size_ % threads_per_warp_ == 0,
-                       << "Invalid block size: number of threads must be "
-                          "evenly divisible by "
-                       << threads_per_warp_);
+        CELER_VALIDATE(default_block_size_ <= CELERITAS_MAX_BLOCK_SIZE,
+                       << "invalid block size: default kernel block size "
+                          "CELER_BLOCK_SIZE="
+                       << default_block_size_
+                       << " (from environment) exceeds "
+                          "CELERITAS_MAX_BLOCK_SIZE="
+                       << CELERITAS_MAX_BLOCK_SIZE << " (from configuration)");
     }
+    CELER_VALIDATE(default_block_size_ >= threads_per_warp_
+                       && default_block_size_ <= max_threads_per_block,
+                   << "invalid block size: number of threads must be in ["
+                   << threads_per_warp_ << ", " << max_threads_per_block
+                   << "]");
+    CELER_VALIDATE(default_block_size_ % threads_per_warp_ == 0,
+                   << "invalid block size: number of threads must be "
+                      "evenly divisible by "
+                   << threads_per_warp_);
 
     CELER_ENSURE(*this);
     CELER_ENSURE(!name_.empty());
--- a/src/corecel/sys/Device.cc
+++ b/src/corecel/sys/Device.cc
@@ -153,7 +153,6 @@ Device::Device(int id) : id_{id}, streams_{new detail::StreamStorage{}}
 
     CELER_LOG_LOCAL(debug) << "Constructing device ID " << id;
 
-    unsigned int max_threads_per_block = 0;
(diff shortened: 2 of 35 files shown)


app/demo-interactor/KNDemoKernel.cu
```diff
@@ -193,7 +193,7 @@ __global__ void cleanup_kernel(DeviceCRef<ParamsData> const params,
 // KERNEL INTERFACES
 //---------------------------------------------------------------------------//
 #define CDE_LAUNCH_KERNEL(NAME, BLOCK_SIZE, THREADS, ...) \
```
esseivaju: I'd remove the block_size parameter since it's no longer used, making it clear that `DeviceGridParams::threads_per_block` has no effect
```suggestion
#define CDE_LAUNCH_KERNEL(NAME, THREADS, ...) \
```

esseivaju (approved):
LGTM! Just one minor comment. 

> Strangely the register usage decreases slightly in a few kernels despite no change in the occupancy...

Yeah, I'm not sure what heuristic would reduce the # of registers with a lower max_threads_per_block if occupancy doesn't increase, it doesn't make much sense. For the boundary kernel, the occupancy increase requires more local memory so we'd have to check if it's worth it.

src/corecel/sys/Device.cc
```diff
@@ -222,22 +223,6 @@ Device::Device(int id) : id_{id}, streams_{new detail::StreamStorage{}}
     // See device_runtime_api.h
     eu_per_cu_ = CELER_EU_PER_CU;
```
esseivaju, line 224: Since we're updating launch bounds and with the recent fix in #968, where do we get the fact that CUDA has one execution unit per compute unit? According to the [definition](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/kernel_language.html#cu-and-eu-definitions) `Compute Unit == Streaming multiprocessor` and `Execution Unit == Warp Scheduler`. Almost every single cuda architecture (except 6.0) have [4 warp scheduler per SM](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#architecture-8-x) and warps are statically bound to one scheduler until completion.

app/demo-interactor/KNDemoKernel.cu
sethrj: Yeah, I just did this to minimize the impact on the macro usage in the file. This whole macro can be removed.

src/corecel/sys/Device.cc
sethrj: Oh yeah, we'd talked about this. I think I based "EU=CU" from the HIP launch bounds definition and some AMD documentation. Would you mind fixing this definition in a new PR? (You have a better understanding about this than I do.) Then maybe our second-argument launch bound can have a meaning that more directly translates to "occupancy".

esseivaju: sure, I'll have a look at it.

esseivaju:
Nice, let me have a quick go at tuning the block size, for some kernel slightly adjusting it can increase the occupancy at no other cost.

esseivaju:
Nevermind, I'll do that in a different pull request as we should add some code to check the GPU architecture and also check that these launch bounds are actually a performance gain. I just updated the InteractionApplier in the same way that we set up the PropagationApplier.

PR #985, opened 2023-10-17
Run "warmup step" to hide once-off overhead from total run time

sethrj:
Our AMD run times for the very small performance benchmark problems have been terrible because there seems to be a 0.7s overhead for every call to `hipFuncGetAttributes`, as well as additional overhead that seems to come from the first kernel calls in the Thrust library. This adds a new option to `celer-sim` (on by default when building with device) that runs an initial stepping loop with zero active tracks.

Here's the beginning of TestEM3 with ORANGE in the CUDA systems profiler, with the first step doing no actual work (and not counting as part of the celeritas "total run time" counter):
cuda-systems-testem3

Besides initializing the static KernelParamsCalculator objects, it also hides the mysterious lag that in this case is showing up as part of `geo-boundary`.

Here's the visualization of Crusher running TestEM3 for 32 steps: the warmup "step" is highlighted in dark blue and takes up almost the entire simulation. Previously this overhead (due to `hipFuncGetAttributes`) was misattributed to the first step.
crusher-warmup

In the shorter celeritas regression runs, this ends up giving a nontrivial performance boost: we're now up to 300× GPU:CPU core equivalence for some problems.
speedup

### DIFF
--- a/app/celer-sim/Runner.cc
+++ b/app/celer-sim/Runner.cc
@@ -109,7 +109,7 @@ auto Runner::operator()(StreamId stream, EventId event) -> RunnerResult
     CELER_EXPECT(event < this->num_events());
 
     auto& transport = this->get_transporter(stream);
-    return (*transport)(make_span(events_[event.get()]));
+    return transport(make_span(events_[event.get()]));
 }
 
 //---------------------------------------------------------------------------//
@@ -121,7 +121,7 @@ auto Runner::operator()() -> RunnerResult
     CELER_EXPECT(events_.size() == 1);
 
     auto& transport = this->get_transporter(StreamId{0});
-    return (*transport)(make_span(events_.front()));
+    return transport(make_span(events_.front()));
 }
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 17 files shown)


esseivaju:
Other than the CI failure triggered by the ScopedMem, this looks good to me. Have you consistently seen latency in kernel launch (geo-boundary) during the first step? First time I've seen this behavior.

sethrj:
Just in recent looking at the system, but it wasn't tied specifically to that one kernel.

sethrj:
Ouch, I guess the scoped profiling mutex prevented this crash before. We can't do scoped memory from thread-local code.

esseivaju (approved):

sethrj:
CI failure unrelated:
```
178/191 Test  #37: corecel/sys/MpiCommunicator/2 ...................................................***Failed  Required regular expression not found. Regex=[tests PASSED
]  0.49 sec
--------------------------------------------------------------------------
A call to mkdir was unable to create the desired directory:

  Directory: /tmp/ompi.2b22a4550bcb.0/pid.4503
  Error:     No such file or directory

Please check to ensure you have adequate permissions to perform
the desired operation.
```

PR #986, opened 2023-10-17
Support device-specific launch bounds

esseivaju:
Optimize block size for occupancy based on nsight compute occupancy calculator.

TODO:

- [x] Check that it results in performance gain
- [x] Conditional compilation depending on GPU architecture

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,9 +353,7 @@ add_library(celeritas_device_toolkit INTERFACE)
 add_library(Celeritas::DeviceToolkit ALIAS celeritas_device_toolkit)
 
 if(CELERITAS_USE_CUDA)
-  target_include_directories(celeritas_device_toolkit
-    SYSTEM INTERFACE $<$<COMPILE_LANGUAGE:C,CXX>:${CUDAToolkit_INCLUDE_DIRS}>
-  )
+  target_link_libraries(celeritas_device_toolkit INTERFACE CUDA::toolkit)
 elseif(CELERITAS_USE_HIP)
   if(CMAKE_HIP_COMPILER_ROCM_ROOT)
     # Undocumented CMake variable
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -34,6 +34,17 @@ endif()
 # Geant4/ROOT data interface
 #-----------------------------------------------------------------------------#
 
+# Orange JSON updater script
+celeritas_add_executable(orange-update orange-update.cc)
+celeritas_target_link_libraries(orange-update
+  Celeritas::orange
+)
+if(CELERITAS_USE_JSON)
+  celeritas_target_link_libraries(orange-update
+    nlohmann_json::nlohmann_json
+  )
+endif()
+
 # Exporter
 celeritas_add_executable(celer-export-geant celer-export-geant.cc)
 celeritas_target_link_libraries(celer-export-geant
--- a/app/celer-dump-data.cc
+++ b/app/celer-dump-data.cc
@@ -621,7 +621,7 @@ void print_atomic_relaxation_data(
 
 //---------------------------------------------------------------------------//
 /*!
- * Dump the contents of a ROOT file writen by celer-export-geant.
+ * Execute and run.
  */
 int main(int argc, char* argv[])
 {
--- a/app/celer-export-geant.cc
+++ b/app/celer-export-geant.cc
@@ -95,11 +95,7 @@ GeantPhysicsOptions load_options(std::string const& option_filename)
 
 //---------------------------------------------------------------------------//
 /*!
- * This application exports particles, processes, models, XS physics
- * tables, material, and volume information constructed by the physics list and
- * loaded by the GDML geometry.
- *
- * The data is stored into a ROOT file as an \c ImportData struct.
+ * Execute and run.
  */
 int main(int argc, char* argv[])
 {
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -21,6 +21,7 @@
 #include <G4VPhysicalVolume.hh>
 
 #include "corecel/io/Logger.hh"
+#include "celeritas/Quantities.hh"
 #include "celeritas/field/RZMapFieldInput.hh"
 #include "celeritas/field/RZMapFieldParams.hh"
 #include "celeritas/field/UniformFieldData.hh"
@@ -133,11 +134,11 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
     }
     else if (field_type == "uniform")
     {
-        auto field = GlobalSetup::Instance()->GetMagFieldZTesla();
+        auto field = GlobalSetup::Instance()->GetMagFieldTesla();
         if (norm(field) > 0)
         {
             CELER_LOG_LOCAL(info)
-                << "Using a uniform field (0, 0, " << field[2] << ") in tesla";
+                << "Using a uniform field (0, 0, " << field[2] << ") [tesla]";
             mag_field_ = std::make_shared<G4UniformMagField>(
                 convert_to_geant(field, CLHEP::tesla));
         }
@@ -145,7 +146,7 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
         // Convert field units from tesla to native celeritas units
         for (real_type& v : field)
         {
-            v *= units::tesla;
+            v = native_value_from(units::FieldTesla{v});
         }
 
         UniformFieldParams input;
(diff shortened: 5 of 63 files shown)


esseivaju:
@sethrj I don't think there is a good way to have compile-time, architecture-specific launch bounds... `__CUDA_ARCH__` is only defined when compiling device code so there's no way to conditionally compile different launch bounds depending on the GPU architecture.

sethrj:
Ping me when you have the launch bounds set and I'll run the quick performance suite on summit!

esseivaju:
@sethrj bounds should set now, I'll check for performance gain/regression as well

sethrj:
@esseivaju Since these are all physics kernels I don't expect much of a performance change overall (except maybe in testem15). We should also be careful that if we modify kernels that use nontrivial geometry function calls, we'll need to also conditionally specify the core geometry.

esseivaju:
yeah, I don't expect a big difference if any. It might even be worse since these kernels were very short and it's gonna be harder for the runtime to schedule parallel kernels with the increased occupancy. Good point about the geometry.

sethrj:
It's looking uniformly a bit slower... first table is local [B] /register [B] / occupancy ; second is speedup.
```diff
diff --git a/kernel-occupancy.md b/kernel-occupancy.md
index f12508f..3b89dbc 100644
--- a/kernel-occupancy.md
+++ b/kernel-occupancy.md
@@ -2,7 +2,7 @@
 | ---- | ------- | ------------------------------------------- | ----- | -------- | --------- |
 | cuda | vecgeom | extend-from-primaries                       |     0 |      112 |     1.000 |
 | cuda | vecgeom | initialize-tracks                           |     0 |      880 |     0.125 |
-| cuda | vecgeom | pre-step                                    |     0 |      280 |     0.375 |
+| cuda | vecgeom | pre-step                                    |     0 |      280 |     0.438 |
 | cuda | vecgeom | along-step-uniform-msc-limit-step-msc-urban |     0 |      880 |     0.125 |
 | cuda | vecgeom | along-step-uniform-msc-propagate            |     0 |     1016 |     0.125 |
 | cuda | vecgeom | along-step-uniform-msc-scatter-msc-urban    |     0 |      880 |     0.125 |
@@ -12,19 +12,19 @@
 | cuda | vecgeom | along-step-neutral                          |     0 |     1016 |     0.125 |
 | cuda | vecgeom | physics-discrete-select                     |     0 |      256 |     0.500 |
 | cuda | vecgeom | ioni-moller-bhabha                          |    48 |      236 |     0.500 |
-| cuda | vecgeom | brems-sb                                    |    48 |      304 |     0.375 |
-| cuda | vecgeom | brems-rel                                   |    48 |      440 |     0.250 |
-| cuda | vecgeom | photoel-livermore                           |    64 |      280 |     0.375 |
+| cuda | vecgeom | brems-sb                                    |    48 |      304 |     0.328 |
+| cuda | vecgeom | brems-rel                                   |    48 |      440 |     0.156 |
+| cuda | vecgeom | photoel-livermore                           |    64 |      280 |     0.438 |
 | cuda | vecgeom | scat-klein-nishina                          |    48 |      224 |     0.500 |
-| cuda | vecgeom | conv-bethe-heitler                          |    48 |      288 |     0.375 |
-| cuda | vecgeom | scat-rayleigh                               |   104 |      256 |     0.500 |
-| cuda | vecgeom | annihil-2-gamma                             |    48 |      224 |     0.500 |
+| cuda | vecgeom | conv-bethe-heitler                          |    48 |      288 |     0.438 |
+| cuda | vecgeom | scat-rayleigh                               |   104 |      256 |     0.438 |
+| cuda | vecgeom | annihil-2-gamma                             |    48 |      224 |     0.562 |
 | cuda | vecgeom | geo-boundary                                |     0 |      880 |     0.125 |
 | cuda | vecgeom | extend-from-secondaries-locate-alive        |     0 |      116 |     1.000 |
 | cuda | vecgeom | extend-from-secondaries-process-secondaries |     0 |      176 |     0.625 |
 | cuda | orange  | extend-from-primaries                       |     0 |      112 |     1.000 |
 | cuda | orange  | initialize-tracks                           |   112 |      256 |     0.500 |
-| cuda | orange  | pre-step                                    |     0 |      256 |     0.500 |
+| cuda | orange  | pre-step                                    |    32 |      224 |     0.547 |
 | cuda | orange  | along-step-uniform-msc-limit-step-msc-urban |    48 |      264 |     0.375 |
 | cuda | orange  | along-step-uniform-msc-propagate            |   240 |      512 |     0.250 |
 | cuda | orange  | along-step-uniform-msc-scatter-msc-urban    |    64 |      296 |     0.375 |
@@ -34,13 +34,13 @@
 | cuda | orange  | along-step-neutral                          |     0 |      508 |     0.250 |
 | cuda | orange  | physics-discrete-select                     |     0 |      236 |     0.500 |
 | cuda | orange  | ioni-moller-bhabha                          |    64 |      236 |     0.500 |
-| cuda | orange  | brems-sb                                    |    64 |      304 |     0.375 |
-| cuda | orange  | brems-rel                                   |    64 |      292 |     0.375 |
-| cuda | orange  | photoel-livermore                           |    80 |      272 |     0.375 |
+| cuda | orange  | brems-sb                                    |    64 |      304 |     0.328 |
+| cuda | orange  | brems-rel                                   |    88 |      256 |     0.469 |
+| cuda | orange  | photoel-livermore                           |    80 |      264 |     0.438 |
 | cuda | orange  | scat-klein-nishina                          |    64 |      232 |     0.500 |
-| cuda | orange  | conv-bethe-heitler                          |    64 |      272 |     0.375 |
-| cuda | orange  | scat-rayleigh                               |   104 |      252 |     0.500 |
-| cuda | orange  | annihil-2-gamma                             |    64 |      232 |     0.500 |
+| cuda | orange  | conv-bethe-heitler                          |    64 |      272 |     0.438 |
+| cuda | orange  | scat-rayleigh                               |   104 |      256 |     0.438 |
+| cuda | orange  | annihil-2-gamma                             |    64 |      224 |     0.562 |
 | cuda | orange  | geo-boundary                                |   176 |      256 |     0.500 |
 | cuda | orange  | extend-from-secondaries-locate-alive        |     0 |      116 |     1.000 |
 | cuda | orange  | extend-from-secondaries-process-secondaries |     0 |      188 |     0.625 |
diff --git a/results/summit/speedup.md b/results/summit/speedup.md
index 4e39daf..45b83cc 100644
--- a/results/summit/speedup.md
+++ b/results/summit/speedup.md
@@ -2,18 +2,18 @@
 | ---------------------------- | -------- | ------------ |
 | cms2018 [Z]                  | vecgeom  |  9.9× (±0.1) |
 | cms2018+field+msc [ZFM]      | vecgeom  |  6.7× (±0.1) |
-| simple-cms+field [BF]        | orange   | 35.2× (±0.3) |
-| simple-cms+field+msc [BFM]   | orange   | 38.2× (±0.5) |
-|                              | vecgeom  | 37.0× (±0.4) |
-| simple-cms+msc [BM]          | orange   | 41.0× (±0.6) |
-| testem15 [A]                 | orange   | 39.1× (±0.6) |
-| testem15+field [AF]          | orange   | 40.3× (±0.3) |
+| simple-cms+field [BF]        | orange   | 34.4× (±0.4) |
+| simple-cms+field+msc [BFM]   | orange   | 37.4× (±0.5) |
+|                              | vecgeom  | 37.1× (±0.4) |
+| simple-cms+msc [BM]          | orange   | 40.1× (±0.6) |
+| testem15 [A]                 | orange   | 37.8× (±0.6) |
+| testem15+field [AF]          | orange   | 39.1× (±0.2) |
 |                              | vecgeom  |  nan× (±nan) |
-| testem15+field+msc [AFM]     | orange   | 42.7× (±0.4) |
-|                              | vecgeom  | 43.1× (±0.3) |
-| testem3-flat [C]             | orange   | 34.5× (±0.4) |
-|                              | vecgeom  | 30.9× (±0.4) |
-| testem3-flat+field [CF]      | orange   | 27.3× (±0.1) |
-| testem3-flat+field+msc [CFM] | orange   | 31.7× (±0.1) |
+| testem15+field+msc [AFM]     | orange   | 40.6× (±3.1) |
+|                              | vecgeom  | 43.1× (±0.2) |
+| testem3-flat [C]             | orange   | 33.6× (±0.4) |
+|                              | vecgeom  | 31.0× (±0.4) |
+| testem3-flat+field [CF]      | orange   | 26.7× (±0.3) |
+| testem3-flat+field+msc [CFM] | orange   | 31.3× (±0.1) |
 |                              | vecgeom  | 15.9× (±0.0) |
-| testem3-flat+msc [CM]        | orange   | 39.8× (±0.1) |
+| testem3-flat+msc [CM]        | orange   | 38.9× (±0.2) |
```

Physics is *barely* a bit slower. Maybe because `pre-step` is spilling now and it didn't previously?

esseivaju:
yeah, that might be the case. Which CUDA architecture are you using? I'm not sure why you're also seeing lower occupancy for a few physics kernels with vecgeom.

esseivaju:
Since we're not getting performance improvement by specifying bounds for block size, let's just merge the bugfix to ActionLauncher for now and investigate the occupancy issues later, performance gains are likely very minor if any.

sethrj (approved):
Thanks!

PR #987, opened 2023-10-18
Improve Quantity for field conversions and non-real_type use cases

sethrj:
This defines a `FieldTesla` quantity and uses it to convert field units because we keep confusing ourselves. I've also added support to Quantity so that it works with non-`real_type` values since that's related.

The RZ field map input now takes everything in *native* units, but the JSON I/O supports a field for specifying that the inputs are in tesla. For backward compatibility, if this `_units` field is empty, it is assumed that the data is in T (@whokion this should let it work with the CMS field profiles you already have).

### DIFF
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -133,7 +133,7 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
     }
     else if (field_type == "uniform")
     {
-        auto field = GlobalSetup::Instance()->GetMagFieldZTesla();
+        auto field = GlobalSetup::Instance()->GetMagFieldTesla();
         if (norm(field) > 0)
         {
             CELER_LOG_LOCAL(info)
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -21,6 +21,7 @@
 #include <G4VPhysicalVolume.hh>
 
 #include "corecel/io/Logger.hh"
+#include "celeritas/Quantities.hh"
 #include "celeritas/field/RZMapFieldInput.hh"
 #include "celeritas/field/RZMapFieldParams.hh"
 #include "celeritas/field/UniformFieldData.hh"
@@ -137,7 +138,7 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
         if (norm(field) > 0)
         {
             CELER_LOG_LOCAL(info)
-                << "Using a uniform field (0, 0, " << field[2] << ") in tesla";
+                << "Using a uniform field (0, 0, " << field[2] << ") [tesla]";
             mag_field_ = std::make_shared<G4UniformMagField>(
                 convert_to_geant(field, CLHEP::tesla));
         }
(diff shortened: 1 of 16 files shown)


src/accel/RZMapMagneticField.hh
```diff
@@ -65,18 +56,16 @@ RZMapMagneticField::RZMapMagneticField(SPConstFieldParams params)
...
 {
     // Calculate the magnetic field value in the native Celeritas unit system
-    Real3 result
-        = calc_field_(Real3{pos[0], pos[1], pos[2]} * this->to_celer_cm());
+    Real3 result = calc_field_(convert_from_geant(pos, CLHEP::cm));
     for (auto i = 0; i < 3; ++i)
     {
         // Return values of the field vector in CLHEP::tesla for Geant4
-        field[i] = result[i] * this->from_celer_tesla();
+        field[i] = convert_to_geant(result[i], CLHEP::tesla);
```
amandalund: Now I'm confused, wasn't this originally returning the field in mT?

test/accel/RZMapMagneticField.test.cc
```diff
@@ -0,0 +1,66 @@
...
+    RZMapMagneticField g4mag_field{params};
+
+    G4ThreeVector g4point;
+    G4ThreeVector g4field_value{0, 0, 0};
+
+    // Calculate inside the grid
+    g4point = convert_to_geant(Real3{5, 1, 2}, CLHEP::cm);
+    g4mag_field.GetFieldValue(&g4point[0], &g4field_value[0]);
+    EXPECT_VEC_SOFT_EQ((Real3{0, 0, 1 * units::tesla}),
+                       convert_from_geant(g4field_value, CLHEP::tesla));
```
amandalund, line 56: Isn't this comparing Gauss to tesla?

sethrj: Ugh you're right, I messed up the test. Hold on...

test/accel/RZMapMagneticField.test.cc
```diff
@@ -0,0 +1,67 @@
...
+class RZMapMagneticFieldTest : public ::celeritas::test::Test
+{
+  protected:
+    void SetUp() override {}
+};
+
+TEST_F(RZMapMagneticFieldTest, uniform_z)
+{
+    std::shared_ptr<RZMapFieldParams> params = [] {
+        // NOTE: RZ field map input is in tesla
```
amandalund:
```suggestion
        // NOTE: RZ field map input is in native units (gauss)
```

app/celer-g4/DetectorConstruction.cc
```diff
@@ -133,19 +134,19 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
...
+                << "Using a uniform field (0, 0, " << field[2] << ") [tesla]";
             mag_field_ = std::make_shared<G4UniformMagField>(
                 convert_to_geant(field, CLHEP::tesla));
         }
 
         // Convert field units from tesla to native celeritas units
         for (real_type& v : field)
         {
-            v *= units::tesla;
+            v = native_value_from(units::FieldTesla{v});
```
amandalund, line 149: You could also do the same for the conversion in celer-sim/Runner.cc while you're at it.

amandalund (approved):
Thanks @sethrj, definitely improves clarity with the unit conversions!

Issue #721, opened 2023-04-13
Standardize nomenclature


sethrj:
# Singular/plural names for variables/accessors 

Add to the documentation (after assuming the team agrees):
- Scalar member variables get variable names of the singular item (material, physics, geometry)
- Vector/collection/array-like variables get plural names (reals, materials, processes)

This means a couple of changes:
- Accessors to `SPMaterialParams` should be `material()` not `materials` even though those classes store vectors
- A single `FooState` object should be `state` even though it represents a vector of states.
- In a few places at least, I gave singular names to the `Collection` storage back ends (see https://github.com/celeritas-project/celeritas/pull/718#discussion_r1165426355 ).

Weird cases to consider:
- Scalar integers like `num_streams`, `max_tracks`?
- Params?

# Primary, generator, initializer

"Primary" particles:
- Direct production of particles from disk or gun
- Offload of Geant4 primaries
- Offload of Geant4 secondaries
- Optical photons generated by Celeritas EM loop

Groups of new particles:
- Primary vertices: `[(position, [(particle type, energy, direction), ...]), ...]`
- Generic event or offloaded tracks: `[(position, particle type, energy, direction), ...]`
- Separable user distribution : `(particle type, spatial distrib, angular distrib, energy distrib)`
- Celeritas EM Secondaries : `[(in-flight track, [(particle type, energy, direction), ...])]`
- Step distributions for continuous processes: `[(pre, post, de/dx, count, process)]`

Data structures where we have track-related information (see #1390):
- Track slots (running)
- Track initializers (queue for track slots)
- Secondary stack (requires existing track in order to generate a track initializer)
- `PrimaryStateData` (aux data in ExtendFromPrimariesAction)
- `optical::GeneratorStateData::distributions` for emitting from cherenkov/scint/primary

# Single-item data retrieval

- `get()` and `unchecked_get()` for OpaqueID, PDGNumber (`get` is like `std::` smart pointers)
- `value()` for Quantity, InitializedValue
- `obj` for JsonPimpl, ReprImpl

Let's make OpaqueId look like `std::optional`: `value` for checked access, `operator*` for unchecked. Same for Quantity and Initialized_value. I think `obj` is fine for JSON because it's a json object.

Note that some C++ facade classes have `base()` indicating the single value that the class wraps, so maybe that's a better word too.

# ID-based accessors

- `get()` for KernelRegistry, MemRegistry etc.
- `id_to_label` and `get` for views with MaterialParams etc.
[...]

Issue #491, opened 2022-08-18
Reorganize libraries and build tree


sethrj:
## Library changes

The goal is to start incorporating into Celeritas the reusable components needed for Geant4 offloading.

- New `src/acceleritas` directory: move all `ext/geant` code there (GeantSetup, GeantImporter)
- Separate libraries for corecel, orange, celeritas, acceleritas
- Separate module for ROOT I/O wrappers: *just* the module? or also include the RootImporter/Exporter and have apps that use ROOT link against it?

```mermaid
graph LR;

acceleritas --> celeritas
acceleritas --> corecel
acceleritas --> Geant4
celeritas_root --> celeritas
celeritas_root --> ROOT
celeritas --> orange
celeritas --> corecel
orange --> corecel
```

## Build tree changes

Goal is to make the build tree look structurally like the installation.

- Set `CMAKE_LIBRARY_OUTPUT_DIRECTORY` at `src/` scope to put libraries in `build/${CMAKE_INSTALL_LIBDIR}` (i.e. `lib` or `lib64`): note that `root_generate_dictionary` only respects the default cmake variable, not the target property
- Set `CMAKE_RUNTIME_OUTPUT_DIRECTORY` in `app/` to put binaries in `build/bin`
- Similar for config files
- Similar include for generated config file paths

@pcanal If we put the generated ROOT files into `build/lib` but don't link against them, what's the best/most elegant way to get the dynamic library lookup for ROOT to find the correct `celeritas_root.so` module automatically?

sethrj:
@pcanal I'm thinking first at build time: if the build files are in `build/lib`, would our tests at `test/blah` be able to find it? I guess not. So maybe it is best to keep the ROOT loading routines in the same library file, so that way tests and apps can be guaranteed to find it using the built-in `LC_LOAD_DYLIB` machO commands on macOS or `rpath` on Linux.

pcanal:
Actually, it will also search the `rpath` (and probably the `LC_LOAD_DYLIB` but I can't test that).

> So maybe it is best to keep the ROOT loading routines in the same library file,

Either way, I don't really see any advantage to separate them, is there?

sethrj:
> Either way, I don't really see any advantage to separate them, is there?

Only if it makes sense to treat the ROOT generated code as a "module" and the celeritas ROOT code as a "library"... but if the celeritas `Root{Ex,Im}porter` needs the generated module code then I guess we should keep them together.

pcanal:
[...]

PR #988, opened 2023-10-18
Add compile-time support for single-precision Celeritas

sethrj:
This adds a new configuration option `CELERITAS_REAL_TYPE` which can be float or double and allows the whole Celeritas+ORANGE system to work with single-precision arithmetic. To make the PR more reviewable I'm splitting it into two, of which this is the first half:
- Add CMake option and constraints for when it can work
- Fix type errors (narrowing conversions of non-constant expressions) in src/ and app/
- Always use double precision for IO and Geant4 interaction
- Add tolerances/fudge factors specific to single precision
- Add an LLDB formatter to make debugging Spans super easy! (see celeritas/scripts/dev/celerlldb.py)
- Fix a propagation failure from having a zero-length chord (for steps that are far too small)
- Add assertions to check for nonzero movement (which happens a lot in single precision but would also break double-precision)

This adds a single-precision build (with no testing) to the CI. A follow-on PR will fix (most of) the tests.


src/celeritas/io/EventReader.cc
```diff
@@ -115,21 +115,24 @@ auto EventReader::operator()() -> result_type
...
-        primary.position = {pos.x() * units::centimeter,
-                            pos.y() * units::centimeter,
-                            pos.z() * units::centimeter};
+        auto to_cm = [](double v) {
+            return static_cast<real_type>(v) * units::centimeter;
+        };
+        primary.position = {to_cm(pos.x()), to_cm(pos.y()), to_cm(pos.z())};
 
         // Get the lab-frame time [s]
         primary.time = pos.t() * units::centimeter / constants::c_light;
```
amandalund: Should we do this for time as well?
```suggestion
        primary.time = to_cm(pos.t()) / constants::c_light;
```

sethrj:
@pcanal @amandalund ready for review :)

CMakeLists.txt
```diff
@@ -271,9 +282,19 @@ if(CELERITAS_USE_ROOT)
...
+    find_package(SWIG 4.1 REQUIRED)
+  endif()
+
+  if(CELERITAS_REAL_TYPE STREQUAL "float")
+    celeritas_error_incompatible_option(
+      "Single-precision runtime is not compatible with SWIG"
+      CELERITAS_USE_SWIG
+      OFF
+    )
+  endif()
```
```suggestion
  if(CELERITAS_REAL_TYPE STREQUAL "float")
    celeritas_error_incompatible_option(
      "Single-precision runtime is not compatible with SWIG"
      CELERITAS_USE_SWIG
      OFF
    )
  else()
    if(NOT SWIG_FOUND)
      # Note: 4.1 is required for C++17 parsing
      find_package(SWIG 4.1 REQUIRED)
    endif()
  endif()
```
No point in complaining twice if using single precision and not finding SWIG.

app/demo-interactor/KNDemoRunner.cc
```diff
@@ -136,8 +136,9 @@ auto KNDemoRunner::operator()(KNDemoRunArgs args) -> result_type
     }
 
     // Copy integrated energy deposition
-    result.edep.resize(detector_params.tally_grid.size);
-    celeritas::app::finalize(params, state, make_span(result.edep));
+    std::vector<real_type> edep(detector_params.tally_grid.size);
```
pcanal, line 139: That pattern may (or may not) be worth a wrapper.

src/accel/LocalTransporter.cc
```diff
@@ -105,8 +105,8 @@ void LocalTransporter::Push(G4Track const& g4track)
 
     track.particle_id = particles_->find(
         PDGNumber{g4track.GetDefinition()->GetPDGEncoding()});
-    track.energy = units::MevEnergy{
-        convert_from_geant(g4track.GetKineticEnergy(), CLHEP::MeV)};
+    track.energy = units::MevEnergy(
```
pcanal, line 108: Why the switch from `{}` to `()` ?

src/orange/Constants.hh
```diff
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file orange/Constants.hh
+//! \todo move to corecel/
```
pcanal, line 7: +1

pcanal (commented):
See comments.

amandalund (approved):
Looks good to me!

CMakeLists.txt
sethrj: Yeah, the reason I had it this way was uniformity with the other packages that have to be loaded first to determine compatibility. But are you manually requesting SWIG? Or is it just finding too old a version on your system?

src/accel/LocalTransporter.cc
sethrj: C++11 brace initialization prohibits casting that could lose precision; so either I could manually insert `static_cast` (which would look like a static cast of double to double in most cases) or switch to parens. I'm not a huge fan of either option...

src/orange/Constants.hh
sethrj: Maybe we need another issue like https://github.com/celeritas-project/celeritas/issues/491 — we should reorganize before 1.0 as well as renaming from #721

app/demo-interactor/KNDemoRunner.cc
sethrj: I'd rather delete this demo app 😅  in fact that may be a good "cleanup" activity for 1.0.

CMakeLists.txt
pcanal: I dont have SWIG at all.

sethrj: Then it should have automatically set `CELERITAS_USE_SWIG=OFF` it it wasn't specified... do you have it on in your cmake user presets?

pcanal: No I did not request SWIG explicitly [independently, I don't use a preset].  Swig got turned on automatically when I switch my build from VecGeom to Orange ... but this is unrelated to the change and logic in this PR.

src/accel/LocalTransporter.cc
pcanal: On the other hand the warning is real ... there is a narrowing ... Using paranthesis to mark the narrowing as been checked it is 'okay' is quite 'subtle'.

CMakeLists.txt
sethrj: That shouldn't have happened. I can't think of why it would...

src/accel/LocalTransporter.cc
sethrj: It's like the difference between explicitly and implicitly constructing an object... C++11 tried to make things better with brace initialization but I think it made things worse. (Also consider construction  for `struct Foo`: `Foo a;` doesn't initialize it, `Foo a{};` does 🤪 )

CMakeLists.txt
pcanal: Me neither ... I'll mention it again if/when I see the issue.

src/accel/LocalTransporter.cc
pcanal: and `Foo a();` declares a function :)

sethrj: Most vexing!

PR #989, opened 2023-10-18
Fix linking to CUDA toolkit when using VecGeom

esseivaju:
Update how the main and final libraries are linked to cudart. Previously, the cuda rpath wasn't embedded in the main library unless a dependency required it. If the user environment wasn't configured to find cudart (ld.so.conf, ld_library_path, standard search path) and the application linking to the main/final library doesn't specify a rpath, we end up with an error. The final library needs to be explicitly linked to cudart again because Celeritas::DeviceToolkit is a private dependency of the main library.
```shell
# old
esseivaj@login13:/global/cfs/cdirs/atlas/esseivaj/devel/celeritas> readelf -d build-ndebug/lib64/libcorecel_final.so | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [/global/cfs/cdirs/atlas/esseivaj/devel/celeritas/build-ndebug/lib64::]

# new
esseivaj@login13:/global/cfs/cdirs/atlas/esseivaj/devel/celeritas> readelf -d build-ndebug/lib64/libcorecel_final.so | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [/pscratch/sd/e/esseivaj/spack/var/spack/environments/celeritas/.spack-env/view/lib64:/global/cfs/cdirs/atlas/esseivaj/devel/celeritas/build-ndebug/lib64::]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,9 +353,7 @@ add_library(celeritas_device_toolkit INTERFACE)
 add_library(Celeritas::DeviceToolkit ALIAS celeritas_device_toolkit)
 
 if(CELERITAS_USE_CUDA)
-  target_include_directories(celeritas_device_toolkit
-    SYSTEM INTERFACE $<$<COMPILE_LANGUAGE:C,CXX>:${CUDAToolkit_INCLUDE_DIRS}>
-  )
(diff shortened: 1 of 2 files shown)


sethrj (approved):
Nice!

PR #992, opened 2023-10-20
Implement concrete G4VFastSimulationModel for offload to Celeritas

### LINKED ISSUE #823: Add "fast simulation" interface to accel
If we want to support exact leptonuclear physics or Bertini cascades (#431 ) we'll need to return particles to Geant4, which will require us to expose Celeritas to Geant4 using the `G4VFastSimulationModel` interface, which needs the `flush` method introduced in Geant4@11.0 (see `param-V11-00-01`).

We also probably want to support offloading EM particles only for the inside of the detector, so we don't have to return hits to the muon calorimeter or any of the far detectors.

drbenmorgan:
A _very_ preliminary interface for offload of tracks to Celeritas via Geant4's fast simulation process/hooks as an alternative to the use of `G4UserTrackingAction::PreUserTrackingAction`.

Mostly copy-pastes implementations of `SimpleOffload::PreUserTrackingAction`, `SimpleOffload::EndOfEventAction` into implementations of abstract member functions of a concrete `G4VFastSimulationModel`. Requires Geant4 >= 11.1 for full support of `G4VFastSimulationModel::Flush` member function to complete tracking at end of event.

There are still a few todos to be discussed after any standard review items:

- Where best to test: `tests/` or through addition of support in `celer-g4`?
- If in `celer-g4` how best to switch between use of `PreUserTrackingAction` and `FastSimulationOffload`
  -  It's probably relatively easier to test here at high level, as by making the fast simulation apply to the `DefaultRegionForTheWorld` `G4Region`, the model _should_ produce identical scoring/diagnostic results.

Fixes: #823


src/accel/FastSimulationOffload.hh
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * This class must be constructed locally on each worker thread/task, typically
+ * within the application's concrete implementation of
+ * `G4VUserDetectorConstruction::ConstructSDandField()`.
+ */
+class FastSimulationOffload final : public G4VFastSimulationModel
+{
+  public:
+    //! Construct with name, shared (across threads) params, and thread-local
+    //! transporter
```
sethrj: The convention for Celeritas documentation is to put Doxygen comments with *definitions* not *declarations*: so the class doxygen goes in the header file (where the class itself is declared); inline functions in the header get short `//!` remarks; and non-inline functions get a `/*! */` definition in the `.cc` file. To make class definitions more readable we usually put a one-line non-doxygen comment above the member function declarations.

src/accel/FastSimulationOffload.hh
```diff
@@ -0,0 +1,60 @@
...
+    explicit FastSimulationOffload(G4String const& name,
```
sethrj: With more than one argument, this doesn't need the `explicit` annotation.

src/accel/FastSimulationOffload.cc
```diff
@@ -0,0 +1,88 @@
...
+                                             SharedParams const* params,
+                                             LocalTransporter* local)
+    : G4VFastSimulationModel(name), params_(params), transport_(local)
+{
+    CELER_VALIDATE(G4VERSION_NUMBER > 1110,
+                   << "the current version of Geant4 (" << G4VERSION_NUMBER
+                   << ") is too old to support the fast simulation Flush() "
+                      "interface");
+    CELER_EXPECT(params);
+    CELER_EXPECT(local);
```
sethrj, line 32: To reduce duplication, you could have this forward (assuming the base class takes a nullptr if the region is unspecified?)
```suggestion
    : FastSimulationOffload(name, nullptr, params, local)
{
```

src/accel/FastSimulationOffload.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "FastSimulationOffload.hh"
+
+#include "corecel/Assert.hh"
+
+#include "ExceptionConverter.hh"
+#include "LocalTransporter.hh"
+#include "SharedParams.hh"
+
+namespace celeritas
+{
```
sethrj, line 17:
```suggestion
{
//---------------------------------------------------------------------------//
/*!
 * Construct a model to be used in all volumes of the problem.
 */
```

src/accel/FastSimulationOffload.cc
```diff
@@ -0,0 +1,88 @@
...
+    CELER_VALIDATE(G4VERSION_NUMBER > 1110,
+                   << "the current version of Geant4 (" << G4VERSION_NUMBER
+                   << ") is too old to support the fast simulation Flush() "
+                      "interface");
+    CELER_EXPECT(params);
+    CELER_EXPECT(local);
+}
+
+FastSimulationOffload::FastSimulationOffload(G4String const& name,
```
sethrj, line 42:
```suggestion
//---------------------------------------------------------------------------//
/*!
 * Construct a model for a specific volume of space (an "envelope").
 *
 * If the envelope is \c nullptr, the model is applicable globally.
 * 
 */
 FastSimulationOffload::FastSimulationOffload(G4String const& name,
```

src/accel/FastSimulationOffload.cc
```diff
@@ -0,0 +1,88 @@
...
+    return true;
+}
+
+void FastSimulationOffload::DoIt(G4FastTrack const& track, G4FastStep& step)
+{
+    // Offload this track to Celeritas for transport
+    CELER_EXPECT(track.GetPrimaryTrack());
+    CELER_EXPECT(*transport_);
+ 
+    ExceptionConverter call_g4exception{"celer0001", params_};
```
sethrj, line 98: Should we reuse the same "code" as in the `./SimpleOffload` (since presumably we're not going to have both simultaneously); or do we want to make the codes unique over all `accel` classes?

sethrj (changes requested):
Nice work Ben! Regarding testing: since this is more oriented toward downstream users, perhaps a new executable in `example/accel` would be a good place to demonstrate this?

src/accel/FastSimulationOffload.hh
drbenmorgan: Done, and added some (hopefully useful) clarifications in the descriptions.

src/accel/FastSimulationOffload.hh
drbenmorgan: Done!

src/accel/FastSimulationOffload.cc
drbenmorgan: This isn't possible as the `G4VFastSimulationModel::G4VFastSimulationModel(const G4String&, G4Envelope*, G4bool)` constructor requires the `G4Envelope` pointer to be not `nullptr`. It doesn't check this though, and yes, these constructors are pretty old and in need of modernisation!

src/accel/FastSimulationOffload.cc
drbenmorgan: Done, but with clarification that the envelope pointer cannot be `nullptr` (Could we use/add `not_null` or something similar for cases like this?).

src/accel/FastSimulationOffload.cc
drbenmorgan: Hmm, I guess the problem with unique codes over all of `accel` would be tracking them and adding new ones with the correct incremented number? Or could we just require each code to be unique within a class, with some magic to turn `celer0001` into, say `accel_FastSimulationOffload_celer0001`.

I don't have a strong opinion either way...

drbenmorgan:
> Nice work Ben! Regarding testing: since this is more oriented toward downstream users, perhaps a new executable in `example/accel` would be a good place to demonstrate this?

Sounds good - do you want it added in this PR? Might be worth discussing inputs/outputs in todays stand up?

src/accel/FastSimulationOffload.cc
sethrj: If they're unique within a class then either the codes need unique prefixes or the same code needs to have the same meaning if it's used multiple times. Ultimately these aren't very important because they just show up in the user's G4Exception code, and even Geant4 doesn't consistently use codes. (Some have a prefix, some are debugging information, etc.) We can take a minute to discuss at the meeting but I think they would have the most utility if they're unique across the codebase: presumably the most common use case is a user saying "I encountered error celer1234"? Even then the associated exception information is the important part...

sethrj:
I think a basic executable alongside `~/Code/celeritas/example/accel/accel.cc` would be great for this PR. (Maybe rename `accel.cc` to `simple-offload.cc` and add `fastsim-model.cc`?) Be sure to update the documentation in `doc/examples/geant4.rst`.

drbenmorgan:
Example added, though very basic for now. The addition of `G4FastSimulationPhysics` to the main physics list with explicit e+/e-/gamma activation should really use Celeritas' supported offload particles, but that isn't available(?) until after G4 initialisation, by which time  I _think_ it's too late to fiddle with the physics list...

sethrj (approved):
Looks great. I'm glad you added the example... it wouldn't have been obvious that the fast simulation *also* needs the simple offload 😄

drbenmorgan:
Ha, indeed that's a bit strange naming wise! Don't know if it would make sense down the road to think about splitting up/refactoring `SimpleOffload` into smaller chunks if more offload "patterns" come along (there's at least one more, using `G4VTrackingManager`), but not critical for now.

PR #993, opened 2023-10-20
Unify names for celer-sim and celer-g4 input

amandalund:
This modifies the app input to use the same name for equivalent celer-sim and celer-g4 input options where we weren't already. I stuck with the more recent celer-g4 names, but @sethrj you probably have stronger opinions on the naming than I do, so let me know if you have different preferences. These are the changes:

| Old | New |
|----|-----|
| `*_filename` | `*_file` |
| `primary_gen_options` | `primary_options` |
| `step_diagnostic_maxsteps` | `step_diagnostic_bins` | 
| `geant_options` | `physics_options` |
| `mag_field` | `field` |

I should probably make sure the accel `SetupOptions` are consistent too...

### DIFF
--- a/app/celer-g4/RunInput.hh
+++ b/app/celer-g4/RunInput.hh
@@ -32,8 +32,6 @@ enum class PhysicsListSelection
 //---------------------------------------------------------------------------//
 /*!
  * Input for a single run.
- *
- * \todo Unify names with celer-sim and SetupOptions.
  */
 struct RunInput
 {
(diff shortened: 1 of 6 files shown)


sethrj (approved):
Looks great! Is there any documentation that needs updating or is that it? 😄

amandalund:
Yeah I didn't see any documentation that needed updating (maybe we need some).

PR #995, opened 2023-10-23
Use fused multiply-add to improve performance and numerical accuracy

sethrj:
FMA introduces smaller errors in floating point arithmetic compared to manual $`x * y + z`$ calculations. This might help with stability of single-precision celeritas, and there's even a chance it'll make CPU code faster: it should at least make sure CPU and GPU are doing the same thing without having to add weird math flags. (Of course, we're memory latency limited, so it's unlikely to help much.)

Some interesting results from this change:
- Register usage goes down by 8 bytes in a few kernels (no change in occupancy though)
- Summit CPU performance *decreases* slightly in along-step kernel
- Frontier CPU performance *doubles* in some cases? but I think this may be a very odd timing issue looking into the results more closely...

### DIFF
--- a/src/celeritas/random/distribution/NormalDistribution.hh
+++ b/src/celeritas/random/distribution/NormalDistribution.hh
@@ -85,14 +85,15 @@ CELER_FUNCTION auto NormalDistribution<RealType>::operator()(Generator& rng)
     if (has_spare_)
     {
         has_spare_ = false;
-        return spare_ * stddev_ + mean_;
+        return std::fma(spare_, stddev_, mean_);
     }
 
-    real_type theta = 2 * constants::pi * generate_canonical(rng);
-    real_type r = std::sqrt(-2 * std::log(generate_canonical(rng)));
+    constexpr auto twopi = static_cast<real_type>(2 * m_pi);
+    real_type theta = twopi * generate_canonical<RealType>(rng);
+    real_type r = std::sqrt(-2 * std::log(generate_canonical<RealType>(rng)));
     spare_ = r * std::cos(theta);
     has_spare_ = true;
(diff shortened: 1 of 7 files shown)


pcanal (approved):
LGTM

src/celeritas/random/distribution/NormalDistribution.hh
```diff
@@ -85,14 +85,15 @@ CELER_FUNCTION auto NormalDistribution<RealType>::operator()(Generator& rng)
     if (has_spare_)
     {
         has_spare_ = false;
-        return spare_ * stddev_ + mean_;
+        return std::fma(spare_, stddev_, mean_);
     }
 
-    real_type theta = 2 * constants::pi * generate_canonical(rng);
-    real_type r = std::sqrt(-2 * std::log(generate_canonical(rng)));
+    constexpr auto twopi = static_cast<real_type>(2 * m_pi);
```
pcanal, line 91: We have 9 (now 10) occurrence of `m_pi` and 30 (now 29) of `constants::pi`.  Is there a difference? Do we have guidance of when to use which?

Here is the `constexpr` or the `static_cast` the 'relevant' part?

Either way, if it is usefull, should it be promoted to a global constant?  (of the 30 usage of `constants::pi`, 20 are for `2*`

amandalund (approved):
Looks good!

src/celeritas/random/distribution/NormalDistribution.hh
sethrj: All the `constants` are `real_type`; whereas `m_pi` is `double`. I'll clarify this on the single-precision PR where that's more relevant.

PR #996, opened 2023-10-23
Implement LdgIterator abstraction

esseivaju:
Implement iterator for read-only device data using `__ldg` primitive


sethrj:
Do we want to do a single PR that includes both the new LDG implementation *and* its use in `Collection<T, const_reference>`?

esseivaju:
I'd do both in the same PR. The iterator implementation should be ready for review, I'll update the collection.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,210 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using difference_type = std::ptrdiff_t;
+    using value_type = T;
+    using pointer = std::add_pointer_t<T const>;
+    using reference = std::add_lvalue_reference_t<T const>;
+    using iterator_category = std::random_access_iterator_tag;
+    using self = LDGIterator<T>;
```
sethrj: Self or This, make it private, or just use LDGIterator because doesn't it default to using the same T?

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,210 @@
...
+    //!@}
+
+    // Iterator requirements
+    CELER_CONSTEXPR_FUNCTION reference operator*() const noexcept
+    {
+#if CELER_DEVICE_COMPILE
+        return __ldg(ptr_);
+#else
+        return *ptr_;
+#endif
```
sethrj: May want to put these in a helper class as static functions, along with the returned value and reference type: that'll let us return an object by reconstructing it from its underlying data

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,210 @@
...
+    constexpr LDGIterator() noexcept = default;
+    constexpr LDGIterator(self const&) noexcept = default;
+    CELER_CONSTEXPR_FUNCTION LDGIterator(std::nullptr_t) noexcept {}
+    CELER_CONSTEXPR_FUNCTION explicit LDGIterator(pointer ptr) noexcept
+        : ptr_{ptr}
+    {
+    }
+    //!@}
+
+    // Iterator requirements
```
sethrj: Put these in doxygen blocks too

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,210 @@
...
+    CELER_CONSTEXPR_FUNCTION self operator++(int) noexcept
+    {
+        self tmp{ptr_};
+        ++ptr_;
+        return tmp;
+    }
+    CELER_CONSTEXPR_FUNCTION friend bool
+    operator==(self const& lhs, self const& rhs)
+    {
+        return lhs.ptr_ == rhs.ptr_;
```
sethrj: This doesn't need to be a friend does it? And should we allow rhs to be a different type so it can do casting (but maybe not since T had to be pod and we can't realistically use it for mutable or volatile data)...

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,210 @@
...
+{
+    return lhs.swap(rhs);
+}
+
+// Helper
+template<class T>
+inline LDGIterator<T> make_ldgiterator(T const* ptr) noexcept
+{
+    return LDGIterator<T>{ptr};
+}
```
sethrj, line 245: I don't think we need this with c++17; make a deduction guide?

sethrj (changes requested):

src/corecel/data/LdgIterator.hh
esseivaju: yeah, it doesn't have to be a friend. We can do `static_cast<const T*>(lhs) == static_cast<const T*>(rhs)` instead or add a `get()`.

src/corecel/data/LdgIterator.hh
esseivaju: We probably don't even need the deduction guide since we already have a constructor taking the same arguments, this was mostly to be consistent with other classes such as `make_observer`or `make_span`.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,261 @@
...
+namespace detail
+{
+//---------------------------------------------------------------------------//
+
+template<class T, typename = void>
+struct LDGLoad
+{
+    using value_type = T;
+    using pointer = std::add_pointer_t<value_type const>;
+    using reference = std::add_lvalue_reference_t<value_type const>;
```
esseivaju: since `__ldg` should only ever return constant primitive type it should be cheaper to return a copy instead of const ref in most cases?

sethrj: I would think the compiler would optimize to the same thing if it returns an lvalue or not... but isn't the signature `T __ldg(T*)` i.e. it always returns an rvalue?

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,253 @@
...
+ * Reads a value T using __ldg builtin and return a copy of it
+ */
+template<class T, typename = void>
+struct LDGLoad
+{
+    using value_type = T;
+    using pointer = std::add_pointer_t<value_type const>;
+    using reference = value_type;
+
+    CELER_CONSTEXPR_FUNCTION static reference read(pointer p)
```
sethrj: Since this isn't trying to conform to any existing interface, perhaps `LdgLoader` with `operator()`? Accessing with an instance could make sense if we extend LDG loading to "read member data from a struct" (e.g., for struct `T`, the loader could hold `U T::*` as class data and read member data from it using ldg?) I think this and the traits class should be in `data/detail/LdgIteratorImpl.hh` since they're more than a couple of lines.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,253 @@
...
+#endif
+    }
+};
+
+/*!
+ * Specialization when T == OpaqueId.
+ * Wraps the underlying index in a OpaqueId when returning it.
+ */
+template<class T>
+struct LDGLoad<T, std::enable_if_t<is_opaqueid_v<T>>>
```
sethrj: Do you write it like this instead of
```c++
template<class I, class T>
struct LDGLoad<OpaqueId<I, T>, void>
```
because it's more similar to C++20 concepts? That's how I have `detail::RangeTypeTraits` set up. Speaking of, we'll probably want another specialization for enumerations... that can go in a follow-up commit of course.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,253 @@
...
+    }
+    CELER_CONSTEXPR_FUNCTION LDGIterator operator++(int) noexcept
+    {
+        LDGIterator tmp{ptr_};
+        ++ptr_;
+        return tmp;
+    }
+    CELER_CONSTEXPR_FUNCTION pointer operator->() const noexcept
+    {
+        return ptr_;
```
sethrj, line 81: Interesting to note that if we have structs we're not going to be able to `__ldg` their data unless we return some sort of DH dereferencing wrapper to the type.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,253 @@
...
+//---------------------------------------------------------------------------//
+}  // namespace detail
+
+//---------------------------------------------------------------------------//
+/*!
+ * Iterator for read-only device data. Use __ldg intrinsic to load data in
+ * read-only cache.
+ */
+template<class T>
+class LDGIterator
```
sethrj: Since `LDG` isn't an abbreviation for anything but `ldg` should this be `Ldg`?

src/corecel/data/LdgIterator.hh
sethrj: Oh yeah, I forget about the circumstances when CTAD works automatically... also because it's changed from C++17 to 20. I think one key one is that 17 can't deduce automatically for struct-like member initialization ...

src/corecel/data/LdgIterator.hh
esseivaju: that was the idea but using specialization only without resorting to SFINAE is better.

src/corecel/data/LdgIterator.hh
esseivaju: yeah, I added this to follow the Iterator requirement. In that case, it's not useful, we could just remove it.

src/corecel/data/LdgIterator.hh
esseivaju: I think it's easier to make this class stateless otherwise the LdgIterator has to make sure they both stay consistent.

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+#include "corecel/Types.hh"
+#include "corecel/data/detail/LdgIteratorImpl.hh"
+#include "corecel/math/Algorithms.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Iterator for read-only device data. Use __ldg intrinsic to load data in
+ * read-only cache.
```
sethrj:
```suggestion
 * Iterator for read-only device data in global memory.
 *
 * This wraps pointer accesses with the \c __ldg intrinsic to load
 * read-only data using texture cache.
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+    //! \name Type aliases
+  private:
+    using LDGLoadPolicy = detail::LdgLoader<T>;
+
+  public:
+    using difference_type = std::ptrdiff_t;
+    using value_type = typename LDGLoadPolicy::value_type;
+    using pointer = typename LDGLoadPolicy::pointer;
+    using reference = typename LDGLoadPolicy::reference;
+    using iterator_category = std::random_access_iterator_tag;
```
sethrj:
```suggestion
  private:
    using LoadPolicyT = detail::LdgLoader<T>;

  public:
    //!@{
    //! \name Type aliases
    using difference_type = std::ptrdiff_t;
    using value_type = typename LoadPolicyT::value_type;
    using pointer = typename LoadPolicyT::pointer;
    using reference = typename LoadPolicyT::reference;
    using iterator_category = std::random_access_iterator_tag;
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+        return *this;
+    }
+    CELER_CONSTEXPR_FUNCTION LdgIterator&
+    operator-=(const difference_type n) noexcept
+    {
+        ptr_ -= n;
+        return *this;
+    }
+    CELER_CONSTEXPR_FUNCTION reference
+    operator[](const difference_type n) const noexcept
```
sethrj: `const` in argument signatures is a little weird:
```suggestion
    operator+=(difference_type n) noexcept
    {
        ptr_ += n;
        return *this;
    }
    CELER_CONSTEXPR_FUNCTION LdgIterator&
    operator-=(difference_type n) noexcept
    {
        ptr_ -= n;
        return *this;
    }
    CELER_CONSTEXPR_FUNCTION reference
    operator[](difference_type n) const noexcept
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+        : ptr_{ptr}
+    {
+    }
+    //!@}
+
+    //!@{
+    //! \name RandomAccessIterator requirements
+    CELER_CONSTEXPR_FUNCTION reference operator*() const noexcept
+    {
+        return LDGLoadPolicy::read(ptr_);
```
sethrj:
```suggestion
        return LoadPolicyT::read(ptr_);
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+    CELER_CONSTEXPR_FUNCTION LdgIterator&
+    operator-=(const difference_type n) noexcept
+    {
+        ptr_ -= n;
+        return *this;
+    }
+    CELER_CONSTEXPR_FUNCTION reference
+    operator[](const difference_type n) const noexcept
+    {
+        return LDGLoadPolicy::read(ptr_ + n);
```
sethrj:
```suggestion
        return LoadPolicyT::read(ptr_ + n);
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+    }
+    //!@}
+
+    // Conversion operators
```
sethrj:
```suggestion
```

src/corecel/data/LdgIterator.hh
```diff
@@ -0,0 +1,218 @@
...
+swap(LdgIterator<T>& lhs, LdgIterator<T>& rhs) noexcept
+{
+    return lhs.swap(rhs);
+}
+//!@}
+
+//!@{
+//! Helper
+template<class T>
+inline LdgIterator<T> make_LdgIterator(T const* ptr) noexcept
```
sethrj:
```suggestion
inline LdgIterator<T> make_ldg_iterator(T const* ptr) noexcept
```

Also the `const` here makes me think that we *should* use CTAD to construct `LdgIterator<int>` from `int const*`, and maybe a static assert to require `!is_const_v<T>`?

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -0,0 +1,65 @@
...
+#include "corecel/Macros.hh"
+#include "corecel/OpaqueId.hh"
+#include "corecel/Types.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+
```
sethrj:
```suggestion
```

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -0,0 +1,65 @@
...
+/*!
+ * Reads a value T using __ldg builtin and return a copy of it
+ */
+template<class T>
+struct LdgLoader
+{
+    static_assert(std::is_integral_v<T>);
```
sethrj: This is too strict: we need to be able to read floating point values too!
```suggestion
    static_assert(std::is_arithmetic_v<T>);
```

test/corecel/data/LdgIterator.test.cc
```diff
@@ -0,0 +1,101 @@
...
+TEST_F(LdgIteratorTest, integral_t)
+{
+    using VecInt = std::vector<int>;
+    VecInt some_data = {1, 2, 3, 4};
+    auto n = some_data.size();
+    auto start = some_data.begin();
+    auto end = some_data.end();
+    auto ldg_start = make_LdgIterator(some_data.data());
+    auto ldg_end = make_LdgIterator(some_data.data() + n);
+    using ptr_type = typename decltype(ldg_start)::pointer;
```
sethrj, line 43:
```suggestion
    using ptr_type = typename decltype(ldg_start)::pointer;
    EXPECT_TRUE((std::is_same_v<ptr_type, int const*>));
```

test/corecel/data/LdgIterator.test.cc
```diff
@@ -0,0 +1,101 @@
...
+}
+
+TEST_F(LdgIteratorTest, opaqueid_t)
+{
+    using TestId = OpaqueId<struct Test>;
+    using VecId = std::vector<typename TestId::size_type>;
+    VecId some_data = {1, 2, 3, 4};
+    auto n = some_data.size();
+    auto ldg_start = LdgIterator<TestId>(some_data.data());
+    auto ldg_end = LdgIterator<TestId>(some_data.data() + n);
```
sethrj: Maybe use `make_ldg_iterator` in one place and CTAD in the other? And then add 
```c++
    EXPECT_TRUE((std::is_same_v<decltype(ldg_start), decltype(ldg_end)>));
```

esseivaju: can we use CTAD here? We have to somehow tell if the pointer points to regular builtin type or if it should reconstruct an OpaqueId.

sethrj: 🤔 but that's just about how it's loaded under the hood right? Both should be `LdgIterator<T>` and the under-the-hood loader handles the reconstruction implementation (and value type).

sethrj: Oh wait, I see: your underlying data is `size_type` not `OpaqueId`. We have some collections of `OpaqueId` itself, and I was thinking we `__ldg` the underlying data. (Do we need to add a `data()` accessor to OpaqueId, or is it permissible to use a cast?)

sethrj:
We could also postpone the OpaqueId stuff until later: being able to do `ldg` on native simple types would be a great start I think.

test/corecel/data/LdgIterator.test.cc
esseivaju: Oh I see, if we want to do that we need the address of `OpaqueId::value_` which we probably don't want to expose using an accessor, maybe declare the LdgLoader as friend.

sethrj: Good idea. As I said, I'm fine postponing that stuff to another PR if it increases the scope too much.

esseivaju: Looks like it's actually not that much change, done in f41f64a so it's probably fine to do it here. Maybe separate the collection update in another PR.

sethrj (approved):
LGTM now! Do you want to merge with the test as is (backing out the Collection header include) or implement an `LdgSpan` for `Collection<T, device, const_reference>` ?

Also, do we want to require `LdgIterator` work with LDG-compatible data (const and fundamental type) or fall back to not using LDG? I'm thinking the former.

esseivaju:
Let's do the `LdgSpan` and `Collection` in a separate PR. I agree, we should require LGD-compatible data. I've updated how constness is specified. Falling back to not using LDG  can lead to misusing the iterator.

esseivaju:
I'm done pushing to this branch unless you have more comments.

src/corecel/data/LdgIterator.hh
esseivaju: yeah, a function call that returns a non-reference is rvalue. In that case:
```c++
T const& f(T const* p)
{
    return __ldg(p);
}
```
I'd assume we bind a reference to temporary which results in UB?

esseivaju:
Realised that I was missing comparison operators while working on the LdgSpan

sethrj (approved):
Great! Even though the `const` requirement makes testing a little more awkward, it's the right thing to do from a device perspective: we never want to accidentally use LDG on non-const data.

PR #998, opened 2023-10-25
Fix test build with single precision and disable failing tests

sethrj:
- Fix comparison macros to work with different precision arguments
- Replace many `double` with `real_type`
- Add a "coarse tolerance" value to replace hardcoded `1-e8`–`1e-6` tolerances
- Disable some test portions (and some tests entirely!) that fail: some are marked as "fixme" because they look like they indicate actual errors, but others are disabled because they're mostly "gold" tests that would be trouble to maintain with single precision


amandalund (approved):
Looks good @sethrj! Just one more failing test.

test/CMakeLists.txt
```diff
@@ -601,7 +630,9 @@ if(CELERITAS_USE_Geant4)
     ENVIRONMENT "${_geant4_test_env}")
   celeritas_add_test(accel/detail/HitProcessor.test.cc
     ENVIRONMENT "${_geant4_test_env}")
-  celeritas_add_test(accel/detail/TouchableUpdater.test.cc)
+  if(CELERITAS_REAL_TYPE STREQUAL "double")
+    celeritas_add_test(accel/detail/TouchableUpdater.test.cc)
+  endif()
```
pcanal:
```suggestion
  celeritas_add_test(accel/detail/TouchableUpdater.test.cc ${_needs_double})
```

test/celeritas/field/Steppers.test.cc
```diff
@@ -75,14 +75,14 @@ class SteppersTest : public Test
...
-        real_type hstep = 2.0 * constants::pi * param.radius / param.nsteps;
+        real_type hstep = 2 * constants::pi * param.radius / param.nsteps;
 
         for (unsigned int i : range(param.nstates))
         {
             // Initial state and the epected state after revolutions
             OdeState y;
-            y.pos = {param.radius, 0.0, i * 1.0e-6};
-            y.mom = {0.0, param.momentum_y, param.momentum_z};
+            y.pos = {param.radius, 0, i * real_type{1e-6}};
```
pcanal, line 92: What's `1e-6` here?

test/CMakeLists.txt
sethrj: It's different: compiling versus not running. I'll check again to see if this one fails to compile and if so will update the comment

sethrj: Yeah I forgot that this test uses `GenericGeoTestBase<GeantGeoParams>` which won't compile with single precision.

test/celeritas/field/Steppers.test.cc
sethrj: A small bump? You'll have to ask @whokion .

pcanal: Well the related question is whether or not it needs to be adjusted for `double` vs `float`.

sethrj: The test passed without it so I guess not?

whokion: > A small bump? You'll have to ask @whokion .

This extra addition is just not to make the operation identical in the loop (i.e, may avoid that the compile is smart enough and execute the loop operation just once as the input/result is same - too smart?).  Anyway, it was introduced when we actually evaluated performance for different stepper cases, which is no longer the case now.

Issue #670, opened 2023-02-27
Incorporate geant4 validation app into Celeritas app


sethrj:
- [ ] Add more granular physics selection to GeantSetup
- [ ] Add ROOT Geant4 MC truth
- [ ] Replace sensitive detector with ROOT output one

PR #999, opened 2023-10-25
Allow `celer-g4` hit output to be opened in downstream ROOT

stognini:
This PR improves `celer-g4` IO as part of issue #670 :

- Decouple classes and data from `celer-g4`: Now `SensitiveHit` and `HitRootIO` (refactored as `RootIO`) only *use* a struct of data with basic types. This avoids adding `G4VHit` and `G4ThreeVector` to the ROOT dictionary, making the task of reading of the output files trivial.
- Move the output data used by `celer-g4` to `celeritas/io`: This consolidates both `ImportX` structs with the event output data used by `celer-g4` into a single dictionary. Thus, we only need to load `libceleritas` into ROOT to be able to read any ROOT file generated by Celeritas.
- ~~Expand the data to encompass both hits and steps: Beyond storing a collection of hits, as `celer-g4` already did, we now store step information.~~


src/celeritas/io/EventData.hh
```diff
@@ -0,0 +1,63 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Particle step data. Arrays are for pre- and post-steps.
+ *
+ * TODO: Map and add a unified process/action id
+ */
+struct StepData
```
sethrj: If we're using `Event` as an analog to `Import`, this should be `EventStepData`. `StepData` is just too generic.

src/celeritas/io/EventData.hh
```diff
@@ -0,0 +1,63 @@
...
+{
+    int volume{0};  //!< Logical volume ID
+    double energy_loss{0};  //!< Same as energy deposition in hit data
+    double length{0};  //!< Step length
+
+    // Pre- and post-step information
+    double energy[2]{0};  //!< Kinetic energy
+    std::array<double, 3> dir[2]{{0, 0, 0}};  //!< Unit vector
+    std::array<double, 3> pos[2]{{0, 0, 0}};  //!< [cm]
+    double time[2]{0};  //!< Global coordinate time
```
sethrj: Instead of this, use an `array<EventStepPointData, 2>` for pre- and post-step data.

src/celeritas/io/EventData.hh
```diff
@@ -0,0 +1,63 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Basic sensitive hit data.
+ */
+struct HitData
+{
+    unsigned int id{0};  //!< Volume's copy number
+    double edep{0};  //!< Energy deposition
+    double time{0};  //!< Global coordinate time
+};
```
sethrj, line 26: This is redundant with the step data; since you're always writing the step data then just delete this.

src/celeritas/user/RootStepWriter.cc
```diff
@@ -37,7 +37,7 @@ void copy_if_selected(const T1& src, T2& dst)
  */
 void copy_if_selected(Real3 const& src, std::array<real_type, 3>& dst)
 {
-    std::memcpy(&dst, &src, sizeof(src));
+    std::copy(src.begin(), src.end(), dst.begin());
```
sethrj, line 40: I think this was fixed in the single-precision PR :) should disappear with a merge with upstream/develop

src/celeritas/io/EventData.hh
```diff
@@ -0,0 +1,63 @@
...
+ * VolumeId is a contiguous Id available globally and mapped to its
+ * respective volume name.
+ */
+struct EventData
+{
+    using SensDetId = int;
+
+    int event_id{0};
+    std::vector<StepData> steps;
+    std::map<SensDetId, std::vector<HitData>> hits;
```
sethrj: Using consecutive `int` means you can just use a `vector` instead. Also isn't `event_id` redundant with ROOT's metadata (successive calls to `Fill`)?

app/celer-g4/GlobalSetup.cc
```diff
@@ -186,12 +186,14 @@ void GlobalSetup::ReadInput(std::string const& filename)
 #endif
     }
 
+#if CELERITAS_USE_JSON
```
sethrj: The code here will still compile if JSON is missing. Is it failing a validation somewhere? Just change the line below to:
```c++
if (CELERITAS_USE_JSON  && input_.output_file.empty())
```

app/celer-g4/SensitiveDetector.cc
```diff
@@ -25,7 +27,11 @@ namespace app
 SensitiveDetector::SensitiveDetector(std::string name)
     : G4VSensitiveDetector(name), hcid_{-1}, collection_{nullptr}
 {
-    this->collectionName.insert(name + "_HC");
+    this->collectionName.insert(name);
+
+#if CELERITAS_USE_ROOT
+    RootIO::Instance()->AddSensitiveDetector(name);
+#endif
```
sethrj: Shouldn't this use `GlobalSetup()::Instance()->GetWriteSDHits()`?

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +63,77 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
+        step.volume = log_vol->GetInstanceID();
+        step.energy_loss = edep;  // [MeV]
+        step.length = g4step->GetStepLength();  // [mm]
 
-    collection_->insert(new SensitiveHit(data));
+        // Pre- and post-step data
+        this->store_step_point(*pre_step, StepPoint::pre, step);
 
+        auto* post_step = g4step->GetPostStepPoint();  // Can be undefined
+        if (post_step)
```
sethrj:
```suggestion
        if (auto* post_step = g4step->GetPostStepPoint())
```

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +63,77 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
+    auto const* log_vol = phys_vol->GetLogicalVolume();
+    CELER_ASSERT(log_vol);
+    auto const p = static_cast<std::size_t>(point);
+
+    step.energy[p] = step_point.GetKineticEnergy();  // [MeV]
+    step.time[p] = step_point.GetGlobalTime();  // [ns]
+    auto const& pos = step_point.GetPosition();
+    step.pos[p][0] = pos.x();  // [mm]
+    step.pos[p][1] = pos.y();  // [mm]
+    step.pos[p][2] = pos.z();  // [mm]
```
sethrj: Add a helper function for `G4ThreeVector`->`std::array`. Also since we're writing to data structures in `celeritas/io` I think we should be using celeritas units (csg) or annotating the structs.

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +63,77 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
-    CELER_ASSERT(touchable);
+    // Insert hit
+    HitData hit;
+    {
+        auto* touchable = g4step->GetPreStepPoint()->GetTouchable();
+        CELER_ASSERT(touchable);
+        hit.id = touchable->GetVolume()->GetCopyNo();
+        hit.edep = edep;
+        hit.time = g4step->GetPreStepPoint()->GetGlobalTime();  // [ns]
+    }
```
sethrj: Redundant with the step data

app/celer-g4/RootIO.cc
```diff
@@ -140,9 +140,22 @@ void HitRootIO::WriteObject(HitRootEvent* hit_event)
...
+ * Map sensitive detectors to contiguous IDs.
  */
-void HitRootIO::Close()
+void RootIO::AddSensitiveDetector(std::string name)
+{
+    auto iter = detector_name_id_map_.find(name);
+    if (iter == detector_name_id_map_.end())
+    {
+        detector_name_id_map_.insert({name, ++detector_id_});
+    }
```
sethrj: You can (and it's good practice to) combine search and insertion:
```suggestion
    if (iter = detector_name_id_map_. insert({std::move(name), detector_id_});
        iter != detector_id_.end())
    {
        ++detector_id_;
    }
```

sethrj (changes requested):

whokion:
General comment:
Why do we store the StepData here?  Should it be part of diagnostic data or should it be stored in the format of histograms as the post analysis in the SteppingAction separately when an analysis flag is turned on, which we discussed earlier?
The current size of output is 5 times bigger than before:
```
ce-run3-cms-ttbar-rzmap-sd-t16.out.root   10573400861  bytes
g4-run3-cms-ttbar-rzmap-sd-t16.out.root   12240420289 bytes
```
(interestingly now, the output of Geant4 is 20% larger than that of Celeritas, which was opposite and the compression factor is much lower for vector<celeritas::HitData>, from ~5 to ~1.5).  Anyway, more 10GB for 32 events are simply too big, I guess.

src/celeritas/io/EventData.hh
mrguilima: Saving just post-step-point data for each step should be enough, as it becomes the pre-step-point for the next step.
Just to be sure, the pre-step point of first step should be saved as the production vertex in the track data.

app/celer-g4/RootIO.cc
```diff
@@ -62,76 +62,76 @@ HitRootIO::HitRootIO()
...
+    this->WriteObject(&event_data);
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Fill a HitRootEvent object.
+ * Fill event tree with event data.
  */
-void HitRootIO::WriteObject(HitRootEvent* hit_event)
+void RootIO::WriteObject(EventData* event_data)
```
pcanal, line 126: Just a reminder that the pattern in this function
```
SetAddress(...);
Fill();
ResetBranchAddresses();
```
is valid, if and only if there is NO other branches in the `TTree` as `TTree::Fill` will import the user data for ALL the branches.

app/celer-g4/EventAction.cc
```diff
@@ -75,7 +75,7 @@ void EventAction::EndOfEventAction(G4Event const* event)
     if (GlobalSetup::Instance()->GetWriteSDHits())
     {
         // Write sensitive hits
-        HitRootIO::Instance()->WriteHits(event);
+        RootIO::Instance()->Write(event);
```
whokion, line 78: Still prefer to use the explicit name, `HitRooIO` and `WriteHits` rather than too generic names.

app/celer-g4/RootIO.cc
```diff
@@ -62,76 +62,76 @@ HitRootIO::HitRootIO()
 
         file_.reset(TFile::Open(file_name_.c_str(), "recreate"));
         CELER_VALIDATE(file_->IsOpen(), << "failed to open " << file_name_);
-        tree_.reset(new TTree(
-            "Events", "Hit collections", this->SplitLevel(), file_.get()));
+        tree_.reset(new TTree(this->TreeName(),
+                              this->TreeName(),
```
whokion: This should be a tile - of course same as the name, but why is the point dropping/degrading the explicit title here?

src/celeritas/io/EventData.hh
stognini: Fair point. Let me see if I can incorporate this easily

src/celeritas/io/EventData.hh
stognini: Yes. `event_id` is redundant for now, since we force it to be contiguous and start at zero, but it could differ if we simply read a hepmc3 input with filtered events at some point. I can remove it if you prefer.

app/celer-g4/RootIO.cc
```diff
@@ -213,6 +231,31 @@ void HitRootIO::Merge()
...
+/*!
+ * Store TTree with sensitive detector names and their IDs (used by
+ * EventData).
+ */
+void RootIO::StoreSdMap(TFile* file)
+{
+    CELER_EXPECT(file && file->IsOpen());
+
+    auto tree = new TTree(
+        "sensitive_detectors", "sensitive_detectors", this->SplitLevel(), file);
```
whokion:
```suggestion
        "sensitive_detectors", "name_to_id", this->SplitLevel(), file);
```

app/celer-g4/GlobalSetup.cc
stognini: That was actually a "mistake" that I forgot to test. I saw that we needed a build option condition for this bit while reading the code, added the macro, and something got me sidetracked before testing turning off the json build option.

src/celeritas/io/EventData.hh
whokion: Actually I propose to remove StepData(~McTruth) completely from hit collections and keep this specific/minimal hit data (with position).

whokion (dismissed):
We need to decide/discuss what the role of this sensitive data - storing SD hits only or string an extended step data. I think that the latter should be separated from this hit collection. At least should have an option to decouple them.

app/celer-g4/EventAction.cc
stognini: I renamed mostly because the purpose expanded. So we were not writing hits exactly, even though we are using `ProcessHits` as the interface to collect the data.

app/celer-g4/RootIO.cc
stognini: Yes! Ideally we do only store one branch, which has the event struct with all the data. This way reading the root file is as simple as possible: load events tree; set the address of the event branch; Get entries.

But this is a good point and I'll add a comment to the code.

app/celer-g4/RootIO.cc
stognini: Just because we have to set up the same tree name in 2 different functions (`RootIO` and `Merge`). So I unified the name in a single place. Easier to maintain if we want to update it as well.

stognini:
> We need to decide/discuss what the role of this sensitive data - storing SD hits only or string an extended step data. I think that the latter should be separated from this hit collection. At least should have an option to decouple them.

We planned to simplify our code and use `ProcessHits` as the interface to collect all the data: either from sensitive detectors (defined by the gdml), or from all volumes by manually forcing them to be sensitive (this option will be addressed in a future PR). This way we only need to maintain the existing interface and be able to store all the data for analysis.

A histogram-only output is also going to be added as an option, but it is useful for our internal validation to have accumulated and binned quantities (histograms) for fast validation, and a thorough output for debugging and really knowing what is going on. Also, having step data level of granularity makes it trivial to integrate the output to our event display and visualize full 3D tracks, which is also a useful tool.

whokion:
This application serves as the base-line for celeritas performance (including IO) too.  Mixing the only necessary SD hits  with the rest of McTruth data should be avoided - need to introduce an additional switch to write out (even partial) McTruth which should be completely  decoupled from standard I/O.  I thought that this MR only address 1) removing the dependency of the ROOT dictionary and Geant4 data (G4VHit/G4ThreeVector) dependence, 2) adding a map to related sensitive name and the their position in the hit collections (so that we avoid of storing the SD name string for each event), 3) replacing position information from G4Step instead of from touchable.  For validation, the histogram analysis should be good enough (as there should be a set of standard observables (plots) for data monitoring (DMon) for each application) and does not require details raw data after the DMon is well defined (which also should be good enough for the purpose of most debugging).

sethrj:
@whokion One thing I'd like to do is to implement the `SimpleCalo` (in src/user) with a Geant4 SD and add an option to toggle between no SD, simple calo, and full mc truth. The histogram Stefano talks about should also be added. This will give us the ability to test the performance differences between:
- No data collection at all (SDs entirely off)
- Minimal GPU data collection (tallying occurs on GPU and only gets sent to the host at the end of the run)
- Lots of data collection but little I/O (Stefano's histograms)
- Lots of data collection and lots of I/O (full MC truth)

This will absolutely let us differentiate the performance degradation from copying to CPU and reconstructing hits, versus I/O overhead. This should be sufficient right?

whokion:
> This will absolutely let us differentiate the performance degradation from copying to CPU and reconstructing hits, versus I/O overhead. This should be sufficient right?

Yes, these four cases should cover different workflows that we'd like to have or test.  On the other hand, performance of the last two cases (histograming and  full MC truth) may not be interested or relevant for I/O studies (even though they give some extra information, but evaluating their overheads with respect to the first is not an essential matter or irrelevant).   Supporting these all four cases with a single I/O class/interface is fine, but at least we should cleanly decouple 2) and 3).

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +84,68 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
+    EventStepData step;
+    {
+        EventHitData hit;
+        {
+            hit.volume = log_vol->GetInstanceID();
+            hit.copy_num = phys_vol->GetCopyNo();
+            hit.energy_dep = edep / CLHEP::MeV;
+            hit.time = pre_step->GetGlobalTime() / CLHEP::s;
+        }
+        step.hit = std::move(hit);
```
sethrj:
```suggestion
        step.hit =  [&] {
            EventHitData hit;
            hit.volume = log_vol->GetInstanceID();
            hit.copy_num = phys_vol->GetCopyNo();
            hit.energy_dep = convert_from_geant(edep, CLHEP::MeV);
            hit.time = convert_from_geant(pre_step->GetGlobalTime(), CLHEP::s);
            return hit;
        }();
```

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +84,68 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
+        step.length = g4step->GetStepLength() / CLHEP::cm;
```
sethrj:
```suggestion
        step.length = convert_from_geant(g4step->GetStepLength(), CLHEP::cm);
```

app/celer-g4/SensitiveDetector.cc
```diff
@@ -57,30 +84,68 @@ void SensitiveDetector::Initialize(G4HCofThisEvent* hce)
...
+ */
+EventStepPointData SensitiveDetector::make_step_point(G4StepPoint& g4step_point)
+{
+    auto const* phys_vol = g4step_point.GetPhysicalVolume();
+    CELER_ASSERT(phys_vol);
+    auto const* log_vol = phys_vol->GetLogicalVolume();
+    CELER_ASSERT(log_vol);
+
+    EventStepPointData step_point;
+    step_point.energy = g4step_point.GetKineticEnergy() / CLHEP::MeV;
```
sethrj: Also `convert_from_geant` here (`celeritas/ext/Convert.geant.hh`)

app/celer-g4/RootIO.cc
```diff
@@ -62,76 +62,82 @@ HitRootIO::HitRootIO()
...
-    // Write the collection of sensitive hits into HitRootEvent
-    HitRootEvent hit_event;
-    hit_event.event_id = event->GetEventID();
-    for (int i = 0; i < hce->GetNumberOfCollections(); i++)
+    // Populate EventData using the collections of sensitive hits
+    EventData event_data;
+    event_data.steps.resize(detector_name_id_map_.size());
+
+    event_data.event_id = event->GetEventID();
+    for (int i = 0; i < hit_cols->GetNumberOfCollections(); i++)
```
sethrj: Use `range` here and for `j` below. Also consider renaming to `hc_id` and `hit_id`?

app/celer-g4/RootIO.cc
```diff
@@ -62,76 +62,82 @@ HitRootIO::HitRootIO()
...
- * Write sensitive hits to output in the form of HitRootEvent.
+ * Write sensitive hits to output in the form of EventData.
+ * See celeritas/io/EventData.hh
  */
-void HitRootIO::WriteHits(G4Event const* event)
+void RootIO::Write(G4Event const* event)
 {
-    G4HCofThisEvent* hce = event->GetHCofThisEvent();
-    if (hce == nullptr)
+    auto const hit_cols = event->GetHCofThisEvent();
```
sethrj, line 87:
```suggestion
    auto const* hit_cols = event->GetHCofThisEvent();
```

src/celeritas/io/EventData.hh
```diff
@@ -0,0 +1,78 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Example of a calorimeter hit.
+ */
+struct EventHitData
+{
+    int volume{0};  //!< Logical volume ID
+    int copy_num{0};  //!< Physical volume copy number
```
sethrj, line 23: Is the copy number something we need for validation? I'm guessing not...

sethrj (commented):
One last round of minor comments! It's in good shape.

src/celeritas/io/EventData.hh
stognini: We have copy numbers on `TestEm3` (slabs have the same name, same logical volume, but different copy numbers for placements). But the geometry is simple enough that we could get rid of them with some tweaking.

sethrj: Right. That's reasonable to keep then. I think for our regular one we use the "flat" one which doesn't have copies.

sethrj (approved):
I think this is good enough for merging. @whokion I know you want to have finer granularity on the level of IO but I think that should be postponed to a subsequent PR, and I think it's better to have "all I/O" and "no I/O" to fully bracket the possible time taken up by I/O and hit reconstruction. As we integrate with experiments, if any of their IO needs are different enough to need benchmarking we can add more specific output levels. Are you ok with this approach?

whokion:
@sethrj If we still write out the whole `EventStepData`, I am definitely against to merge this as it increase output size by adding none-essential SD information as well as discarding the more realistic existing (minimal) SD output mode.  We do not want to study I/O overhead with MC truth-like data, of which overhead may not linearly scalable by the output size and may be useless information.  Please keep the existing HitData (now `EventHitData`)  as the default I/O and add an option to write EventStepData with a separately mode, which is, I thought, the purpose of this MR.

sethrj (approved):
Awesome, thanks again and sorry again for the back and forth.

PR #1000, opened 2023-10-26
Use LdgIterator in Collection and Span

esseivaju:
Follow-up to #996 

- Use `__ldg` to read `Span` data when the template argument is `Span<LdgValue<T const>>` (`LdgSpan`)
- Automatically use  `LdgSpan` from `Collection` with const_reference ownership if `T` is supported
- Support `LdgIterator` of enums


src/corecel/OpaqueId.hh
```diff
@@ -30,7 +36,8 @@ namespace celeritas
 template<class ValueT, class SizeT = ::celeritas::size_type>
 class OpaqueId
 {
-    static_assert(static_cast<SizeT>(-1) > 0, "SizeT must be unsigned.");
+    static_assert(std::is_unsigned_v<SizeT> && !std::is_same_v<SizeT, bool>,
```
sethrj: Lol, what kind of evil genius would make an OpaqueId with `bool` 😂

src/corecel/cont/detail/SpanImpl.hh
```diff
@@ -8,14 +8,41 @@
...
 #include "corecel/Macros.hh"
+#include "corecel/OpaqueId.hh"
+#include "corecel/data/LdgIterator.hh"
 
 namespace celeritas
 {
 namespace detail
 {
+template<class T, typename = void>
+struct SpanTrait
```
sethrj:
```suggestion
struct SpanTraits
```

sethrj (commented):
Looking good! Nice work extending Span: everything will hopefully "just work"! There is the possibility that we need to make using `__ldg` a bit more explicit than `const`: for a couple of use cases (think cross section sampling, distance calculation, and other places where we fill a local state buffer in one place and pass it to be used in another) I think `__ldg` may be wrong. Also if we ever pass a span of constant or shared data by accident, ldg will break...

esseivaju:
I can see making the use of LdgIterator explicit. How would you specify that though? Use a separate span implementation that can be constructed from a `Span<T const>`?

sethrj:
What about some sort of tag class
```c++
template<class T>
struct LdgValue
{
};
```
that would trigger a specialization when doing `Span<LdgValue<real_type>>`? Our Collection would have
```c++
using SpanConstT = typename detail::CollectionTraits<T, W>:: SpanConstT;
```

It might take some trial and error for the types to get the `storage_` and the return types to play together...

sethrj:
@esseivaju Also, congratulations on opening our thousandth issue/PR! 🎉

esseivaju:
Hopefully, it's one that leads to performance speedup!

esseivaju:
@sethrj Before I start updating `Collection` and usage of `Span` in the codebase, do you agree with the API of `Span<T>` and `Span<LdgValue<T>>`? The significant difference is that `front`, `back`, and `operator[]` return by value when using `__ldg`. As the iterator returns a copy of the value read / we reconstruct an OpaqueId, we can't return a reference to the underlying data.

src/corecel/cont/detail/SpanImpl.hh
```diff
@@ -8,14 +8,39 @@
...
+    using iterator = pointer;
+    using const_iterator = const_pointer;
+    using reference = std::add_lvalue_reference_t<T>;
+    using const_reference = std::add_lvalue_reference_t<T const>;
+};
+template<class T>
+struct SpanTraits<LdgValue<T>>
+{
+    using element_type = typename LdgValue<T>::value_type;
+    using pointer = std::add_pointer_t<T const>;
```
sethrj: Isn't T always const here? We definitely need to document here what values of T are permissible and the fact that the `LdgValue<T>` is responsible for prohibiting (with static assert) invalid types. The `Span` class documentation also needs a description of when, where, and how `__ldg` is added to the dereferences: basically everything but `data()` calls through?

src/corecel/cont/detail/SpanImpl.hh
```diff
@@ -8,14 +8,39 @@
...
+    using const_pointer = pointer;
+    using iterator = LdgIterator<T const>;
+    using const_iterator = iterator;
+    using reference = std::remove_const_t<T>;
+    using const_reference = std::remove_const_t<T>;
```
sethrj: Since all LDG data is returned by value, and `s_.data[0]` (I think?) returns an ldg-fetched value, shouldn't these both return `element_type`?

src/corecel/cont/detail/SpanImpl.hh
```diff
@@ -47,7 +72,7 @@ struct SpanImpl
 {
     //// DATA ////
 
-    T* data = nullptr;
+    typename SpanTraits<T>::iterator data = nullptr;
```
sethrj: Add before the data section:
```c++
    //// TYPES ////
    using iterator = typename SpanTraits<T>::iterator;
    using pointer = typename SpanTraits<T>:: pointer;
```
and in the specialization below.

src/corecel/data/Collection.hh
```diff
@@ -132,8 +132,8 @@ using ItemRange = Range<OpaqueId<T, Size>>;
 template<class T1, class T2>
 class ItemMap
 {
-    static_assert(detail::IsOpaqueId<T1>::value, "T1 is not OpaqueID");
-    static_assert(detail::IsOpaqueId<T2>::value, "T2 is not OpaqueID");
+    static_assert(detail::is_opaque_id_v<T1>, "T1 is not OpaqueID");
+    static_assert(detail::is_opaque_id_v<T2>, "T2 is not OpaqueID");
```
sethrj, line 136: We should be more consistent on the naming scheme; I think sticking with our internal class conventions rather than making it look like `std` is better, since we have `Span`? (But looks like we might want to rename `celeritas::numeric_limits` 🤪 )

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -28,7 +28,7 @@ struct LdgLoader
                   "Only const arithmetic types are supported by __ldg");
     using value_type = T;
     using pointer = std::add_pointer_t<value_type>;
-    using reference = value_type;
+    using reference = std::remove_const_t<value_type>;
```
sethrj: Shouldn't `value_type` have the `remove_const_t` and only `pointer` should have `T const*`?

sethrj (commented):
@esseivaju I like this. `LdgValue` shouldn't be something most users see (this optimization should be *mostly* invisible to developers and users) so I think it's OK to be a little awkward, as long as we document when and how this takes effect. Returning values as "references" should be OK for the situations we care about, since they're all small const objects.

src/corecel/cont/detail/SpanImpl.hh
esseivaju: Yes, T is always const, I specify T const again so that it's more explicit but technically not needed. `data()` is the only way to bypass `__ldg`, I'll update the doc

src/corecel/cont/detail/SpanImpl.hh
esseivaju: `element_type` is const and we don't want to return a const value.

src/corecel/data/Collection.hh
esseivaju: I was looking through the codebase and most constexpr bool are using snake case (in ApplierTraits, ActionRegistryIMpl)

esseivaju:
Do you want to completely hide `LdgValue` from users? If we also want to make getting a `Span<LdgValue<T>>` from a `Collection` explicit, I was thinking of overloading operator[] with `operator[](AllItems<LdgValue<T>, M>)` and `operator[](Range<LdgValue<T>>)`

esseivaju:
Alright, I think all the features are implemented: two things potentially missing:

- Update Collection tests
- Go through the code, and find uses of `Span<T const>` potentially updating it to `Span<LdgValue<T>>`. I've seen `NonUniformGrid`,  which uses its own `Span<T const>` instead of reusing the one defined in `Collection` and there are probably other spots where this is the case.

src/corecel/data/Collection.hh
esseivaju: maybe we can update this in a separate PR?

sethrj: Works for me; added to #721.

src/orange/detail/UniverseIndexer.hh
```diff
@@ -83,7 +83,7 @@ class UniverseIndexer
         = Collection<size_type, Ownership::const_reference, MemSpace::native>;
     using AllVals = AllItems<size_type, MemSpace::native>;
     using SizeId = OpaqueId<size_type>;
-    using SpanIter = Span<size_type const>::const_iterator;
+    using SpanIter = typename DataRef::SpanConstT::const_iterator;
```
sethrj, line 86: This is sweet: it means with effectively no change to our code, the `upper_bound` implementation we borrowed from libcxx now automatically uses `__ldg` to converge on the data: @tmdelellis reported that a hand-rolled lower-bound written with `ldg` improved their performance substantially.

test/corecel/cont/Span.test.cc
```diff
@@ -253,6 +255,80 @@ TEST(SpanTest, io_manip)
         EXPECT_EQ("{123   ,456  ,789  }", os.str());
     }
 }
+TEST(SpanTest, ldg_span)
+{
+    {
```
sethrj: Just a minor thing, but how about breaking this test into two `TEST(LdgSpanTest, pod)` and `TEST(LdgSpanTest, opaque_id)` instead of two blocks in one test?

sethrj (approved):
This looks great @esseivaju ! Have you had a chance to run any performance numbers? I'll launch a test run on summit :)

src/orange/detail/UniverseIndexer.hh
esseivaju: yeah, since `LdgIterator` complies with all the requirements it can be used in most places where a standard iterator is expected, it should work with thrust as well.

src/corecel/cont/detail/SpanImpl.hh
```diff
@@ -8,14 +8,50 @@
...
+ * Type aliases when data is read using __ldg. \c LdgValue checks that T is a
+ * valid type (const arithmetic or OpaqueId). Since \c LdgIterator returns a
+ * copy of the data read, we can't return a reference to the original data, we
+ * need to return a copy as well.
+ */
+template<class T>
+struct SpanTraits<LdgValue<T>>
+{
+    // element_type is always const
+    using element_type = typename LdgValue<T>::value_type;
```
esseivaju, line 47: `element_type` should keep constness I think, it makes a difference in `Span(element_type (&arr)[N])` constructor. If we remove const here, then we can't construct a `LdgSpan` from a `T const[]`. Span can always remove const from the type if needed.

sethrj: Yeah I didn't realize that std array keeps the "const" in value type or that the double const didn't hurt. It gets away with it because it never returns value type.

esseivaju:
Didn't get to do any performance test yet. I've updated some use of Span in the codebase.  A few comments:

- We have some `Span<PDGNumber const>` which could be handled the same way we handle `OpaqueId`
- Support for enum
- `DeviceVector` and `StackAllocator` can return a constant view using `Span<T const>`, which could be specialized when T supports __ldg 

Should we do that later?

src/corecel/cont/detail/SpanImpl.hh
esseivaju: yes, Span has both element_type which keeps constness, and value_type which strips it, neither of them is ever returned so we just have to note that if we use them later. Maybe the naming of these aliases could be more clear.

esseivaju:
008c971 breaks orange so I'll just leave the `make_array` overload for now.

sethrj:
I reran the performance suite on summit, and alas that there appears to be zero performance difference in any of the cases :( I'd really hoped that ORANGE 

One minor but weird thing is that the memory spill for the livermore photoelectric kernel dropped from 80 to 64B...

esseivaju:
I tried running testem3 on Perlmutter with Orange and I got the same result, no performance difference and less local memory spill for Livermore photoelectric kernel. It would be interesting to compare GPU cache usage for the propagation kerne, I suppose nvcc was already able to optimize it.

Should we merge this as is or wait until we get some performance improvement?

sethrj:
Merge as is I think. It'll be easier to maintain. If we need to do comparisons later it's trivial to turn it off. Thanks and I'm sorry all this work didn't give us any immediate benefit!

src/celeritas/grid/ValueGridBuilder.hh
```diff
@@ -66,7 +66,7 @@ class ValueGridXsBuilder final : public ValueGridBuilder
   public:
     //!@{
     //! \name Type aliases
-    using SpanConstDbl = Span<double const>;
+    using SpanConstDbl = LdgSpan<double const>;
```
sethrj: These classes are all host-only (they're used in construction) so I don't think they should be modified.

src/celeritas/grid/ValueGridInserter.hh
```diff
@@ -46,7 +46,7 @@ class ValueGridInserter
         = Collection<real_type, Ownership::value, MemSpace::host>;
     using XsGridCollection
         = Collection<XsGridData, Ownership::value, MemSpace::host>;
-    using SpanConstDbl = Span<double const>;
+    using SpanConstDbl = LdgSpan<double const>;
```
sethrj: Same here.

src/celeritas/mat/MaterialParams.hh
```diff
@@ -42,9 +42,9 @@ class MaterialParams final : public ParamsDataInterface<MaterialParamsData>
   public:
     //!@{
     //! \name Type aliases
-    using SpanConstMaterialId = Span<MaterialId const>;
-    using SpanConstElementId = Span<ElementId const>;
-    using SpanConstIsotopeId = Span<IsotopeId const>;
+    using SpanConstMaterialId = LdgSpan<MaterialId const>;
+    using SpanConstElementId = LdgSpan<ElementId const>;
+    using SpanConstIsotopeId = LdgSpan<IsotopeId const>;
```
sethrj: Also a host-only class

src/celeritas/phys/PhysicsParams.hh
```diff
@@ -120,7 +120,7 @@ class PhysicsParams final : public ParamsDataInterface<PhysicsParamsData>
     using SPConstRelaxation = std::shared_ptr<AtomicRelaxationParams const>;
 
     using VecProcess = std::vector<SPConstProcess>;
-    using SpanConstProcessId = Span<ProcessId const>;
+    using SpanConstProcessId = LdgSpan<ProcessId const>;
```
sethrj: Also host-only

src/corecel/cont/LabelIdMultiMap.hh
```diff
@@ -47,7 +47,7 @@ class LabelIdMultiMap
     //!@{
     //! \name Type aliases
     using IdT = I;
-    using SpanConstIdT = Span<IdT const>;
+    using SpanConstIdT = LdgSpan<IdT const>;
```
sethrj: Also host only

src/corecel/cont/Span.hh
```diff
@@ -33,22 +35,30 @@ constexpr std::size_t dynamic_extent = detail::dynamic_extent;
...
 {
+    using SpanTraitsT = detail::SpanTraits<T>;
+
   public:
     //!@{
     //! \name Type aliases
-    using element_type = T;
-    using value_type = std::remove_cv_t<T>;
+    using element_type = typename SpanTraitsT::element_type;
+    using value_type = std::remove_cv_t<element_type>;
```
sethrj, line 54: I thought about this more after you pointed out that `const value_type` is ok even when `value_type` is const. The standard implementation of `std::array` and `std::span` both literally alias `value_type = T`, so maybe we should keep `value_type` const but let `element_type` be the non-const version?

src/corecel/grid/NonuniformGrid.hh
```diff
@@ -62,7 +62,7 @@ class NonuniformGrid
 
   private:
     // TODO: change backend for effiency if needeed
```
sethrj: This comment can now be deleted 🎉

src/corecel/cont/Span.hh
esseivaju: Looking at [std::span](https://en.cppreference.com/w/cpp/container/span), `element_type = T` and `value_type` removes cv qualifiers which is what we have now so I think it's fine. `std::array` is a bit different since it only has `value_type`

sethrj: Aha quite right. Thanks!

sethrj (approved):
Nice work @esseivaju ! I suppose we don't do enough bounds searches (reloading recently used data) to make a huge impact after all... that's a little disappointing.  But we can try turning off this capability once we have more complex problems and see if there's any speedup after all.

esseivaju:
yeah, it should be easy to turn off. Regarding my [previous comment](https://github.com/celeritas-project/celeritas/pull/1000#issuecomment-1786139735). Only enum support (`std::byte`) matters, I think. `PDGNumber` is host-only. `StackAllocator` uses more complex types and `DeviceVector` doesn't initialize data, `DeviceVector<T const>` doesn't make sense. It's trivial to create a `LdgSpan` from a `Span if needed so we should probably leave it that way. 

I can have a go at adding support for `LdgSpan<Byte>`, it should be fairly straightforward.

esseivaju:
One last review 😅

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -63,6 +63,31 @@ struct LdgLoader<OpaqueId<I, T> const>
...
+{
+    using value_type = T;
+    using pointer = std::add_pointer_t<value_type const>;
+    using reference = value_type;
+    using uderlying_type = std::underlying_type_t<T>;
+
+    CELER_CONSTEXPR_FUNCTION static reference read(pointer p)
+    {
+#if CELER_DEVICE_COMPILE
+        return value_type{__ldg(reinterpret_cast<uderlying_type const*>(p))};
```
esseivaju: I am not 100% sure this cast is legal. Looking at the reinterpet_cast rule: 

> Any object pointer type T1* can be converted to another object pointer type cv T2*. This is exactly equivalent to static_cast<cv T2*>(static_cast<cv void*>(expression)) (which implies that if T2's alignment requirement is not stricter than T1's, the value of the pointer does not change and conversion of the resulting pointer back to its original type yields the original value). In any case, the resulting pointer may only be dereferenced safely if allowed by the type aliasing rules (see below). 

The enum and its underlying type have the same size and alignment requirements so casting is fine but I think that dereferencing, which is probably done in `__ldg` implementation, breaks the aliasing rules.

esseivaju: In practice, it _most likely_ doesn't matter since we don't keep around the pointer obtained from `reinterpret_cast` there shouldn't be an opportunity for the compiler to take advantage of this rule. 

Still, if we want to be safe, [bit_cast](https://en.cppreference.com/w/cpp/numeric/bit_cast) would be useful, maybe we can use the suggested implementation using [std::memcpy](https://en.cppreference.com/w/cpp/string/byte/memcpy) to get around it?

sethrj: Yeah, I'd guess you're right. Oh well! Given the little we've gained yet, I suspect we won't see much by supporting enums. If we really wanted to though, we could store them on device as their underlying type, and static cast to return them? Kind of like how `range` works for enum iterators.

sethrj: Yeah, I guess we could use memcpy to avoid reinterpreting on the return type, but we have to reinterpret_cast the pointer to give it to `__ldg` right?

esseivaju: If you look at the PTX output of [this example](https://godbolt.org/z/qYezvhvhx), the first two blocks are exactly the same, so nvcc can optimize the memcpy to a cast. And yeah, that's pretty much for nothing supporting enum will most likely not make a difference...

sethrj: Not sure if that's the best example since it's doing everything locally? Here's an example of [loading enum directly versus using ldg](https://godbolt.org/z/jsbP793sW)... looks like casting or not ends up the same. (Using ldg or not changes the load command from `u64` to `s64`,  not sure if that matters...)

esseivaju: mmh yeah, I've added [one example](https://godbolt.org/z/bee5q7j8n) with the memcpy and static_cast, it is doing quite a bit more than just the reinterpret_cast. I suppose the main difference in the ldg load command is `.nc` for the non-coherent cache, not sure why it's switching between signed/unsigned. So, since `cast2` and `cast3` are the same, should we just go with the `reinterpret_cast`?

sethrj: Yes. Let's either reinterpret_cast (only needed for `__ldg(value)` on device) or elect not to try to accelerate enums.

esseivaju: Alright, so with that I think we're good!

esseivaju: Kinda surprising that nvcc does the actual copy whereas gcc/clang [optimize it](https://godbolt.org/z/voqh373E7) to a single move.

sethrj (approved):
LGTM! Thanks.

PR #1001, opened 2023-10-27
Add ORANGE surface construction components

sethrj:
This adds two classes necessary for constructing surfaces from shapes:
- `LocalSurfaceInserter` which deduplicates surfaces, *including* chaining nearly-adjacent surfaces (see test) which is something the SCALE ORANGE fails to do
- `FaceNamer` which constructs surface extensions (e.g. `px`, `p1`, `s`) to be combined with shape names to form surface names

Currently inserting N surfaces with deduplication is an `O(N^2)` operation because the initial implementation does *not* use hashing to reduce the number of surfaces needed to test. However, in opt mode (on my laptop) 32k surface insertions take under a second, and a debug build takes just over a second to do 8k surfaces.


src/orange/construct/detail/LocalSurfaceInserter.hh
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * There are three cases to consider:
+ * - The new surface is entirely unique: we insert and return the new ID.
+ * - The surface is soft equivalent but not exactly like an existing surface:
+ *   we insert but return an existing ID.
+ * - The surface is exactly the same: we do \em not insert, and return existing
+ *   id.
+ *
+ * The second case adds the surface so that multiple nearby surfaces can be
+ * \em chained together, even if the tolerance between the furthest apart is
```
elliottbiondo, line 37: I understand the purpose of the em-dashes above, but not in this paragraph.

src/orange/construct/detail/LocalSurfaceInserter.hh
```diff
@@ -0,0 +1,122 @@
...
+  public:
+    // Construct with tolerance and a pointer to the surfaces vector
+    LocalSurfaceInserter(VecSurface* v, Tolerance<> const& tol);
+
+    // Construct a surface with deduplication
+    template<class S>
+    inline LocalSurfaceId operator()(S const& surface);
+
+  private:
+    using MapSurfId = std::unordered_map<LocalSurfaceId, LocalSurfaceId>;
```
elliottbiondo, line 59: `>>> TYPES`, `>>> HELPER FUNCTIONS`, `>>> DATA` for consistency, but not necessary

src/orange/surf/FaceNamer.cc
```diff
@@ -0,0 +1,144 @@
...
+    return {'c', to_char(T)};
+}
+
+//! \cond
+ORANGE_INSTANTIATE_OP(CylCentered);
+//! \endcond
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a name for a sphere.
```
elliottbiondo: centered sphere

src/orange/surf/FaceNamer.cc
```diff
@@ -0,0 +1,144 @@
...
+std::string FaceNamer::Impl::operator()(Plane const&) const
+{
+    return "p" + std::to_string(state_->num_planes_++);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a name for a sphere.
+ */
+std::string FaceNamer::Impl::operator()(Sphere const&) const
```
elliottbiondo, line 106: Are sphere and centered sphere supposed to have the same name?

elliottbiondo (approved):

src/orange/construct/detail/LocalSurfaceInserter.hh
sethrj: The `em` is the `<em>` html tag aka "emphasis" (i.e., italicized).

src/orange/surf/FaceNamer.cc
sethrj: Yeah, that's how it is in the current SCALE builder. I can't think of any shape that has anything *but* a single sphere...

PR #1002, opened 2023-11-01
Make boundary distance persistent to improve Shift performance

sethrj:
This will allow Shift to eliminate the redundant "find next distance" call in its "move to and cross boundary" kernel (and its "move to point kernel").  This cacheing isn't needed in Celeritas because we know at the start of the "find distance" kernel (aka propagate) that we will move to the boundary if it's shorter than the next distance:
| Function  |  Celeritas action  |  Shift kernel |
| -------- | --- | --- |
| `find_next_step` | `AlongStep` | `d2b_kernel` |
| `move_to_boundary` | `AlongStep` | `move_across_surface_kernel` |
| `move_internal` | `AlongStep` | `move_to_point_kernel` |
| `cross_boundary` | `Boundary` | `move_across_surface_kernel` |

I'm going to see if the "always store in global" implementation has any effect on performance. If not, we can merge this. If so, I'll try keeping the distance local and making the next-distance caching *optional* by using a parameter.

### DIFF
--- a/src/orange/OrangeData.hh
+++ b/src/orange/OrangeData.hh
@@ -418,6 +418,12 @@ struct OrangeStateData
     StateItems<Sense> sense;
     StateItems<BoundaryResult> boundary;
 
+    // "Local" state, needed for Shift {num_tracks}
+    StateItems<real_type> next_step;
+    StateItems<SurfaceId> next_surface;
+    StateItems<Sense> next_sense;
+    StateItems<LevelId> next_level;
+
     // State with dimensions {num_tracks, max_depth}
     Items<Real3> pos;
     Items<Real3> dir;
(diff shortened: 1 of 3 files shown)


sethrj:
@tmdelellis This change increases occupancy for our along-step while not increasing local memory spill. However it results in a overall performance *decrease* of about 3% on summit for all cases... so I'm going to try the "caching as an option" implementation.

sethrj:
@tmdelellis I think a lot of the performance drop is due to some local "next distance" caching in the along-step function that's not actually useful. I've removed it in #1004 and will pull that in after this goes in. If the SCALE code works then that should be good enough for doing the EMPIRE runs.

elliottbiondo (approved):

tmdelellis (approved):
To be clear - at this point there is no toggle for turning of caching.  As I read this cacheing is always on.

PR #1003, opened 2023-11-01
Add Orange Shift tracking flow control test

tmdelellis:
This adds a simple test to `Orange.test.cc` that mirrors the Orange tracking sequence on the GPU, but in a CPU-only, single threaded test:

```cpp
while (bnd_state == BoundaryState::INSIDE)
{
    this->distance_to_boundary(dbnd);
    this->move_across_surface(bnd_state, cell);
    ++steps[n];
}


sethrj (approved):
Looks great! Thanks @tmdelellis .

PR #1004, opened 2023-11-02
Optimize ORANGE surface crossings and checks

sethrj:
This follow-on to #1002 does a few optimizations (avoid back-and-forth universe indexing) and code improvements (adding `clear_next` and `clear_surface` helper functions), and drops caching of `find_next(max_step)` calls, which will all hopefully mitigate the performance drop from #1002.

For TestEM3 on summit (with field plus msc) this results in a ~5% overall speedup compared to #1002 and a ~2% overall speedup compared to before that change.

### DIFF
--- a/src/orange/OrangeData.hh
+++ b/src/orange/OrangeData.hh
@@ -419,10 +419,10 @@ struct OrangeStateData
     StateItems<BoundaryResult> boundary;
 
     // "Local" state, needed for Shift {num_tracks}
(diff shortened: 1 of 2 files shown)


sethrj:
@elliottbiondo If you can review this first thing, we can get it into the upcoming SCALE PR (and might have a small performance boost from it).

elliottbiondo (approved):

Issue #850, opened 2023-07-10
Extend celer-g4 for detailed Geant4/Celeritas comparisons


sethrj:
The goal here is to allow us to do fully apples-to-apples comparisons between Geant4 and Celeritas for CMS.

Core celeritas enhancements:
- [x] Export HepMC3/event file with incoming secondaries (use single file plus mutex) (@sethrj , #854)
- [x] Add a Geant4 magnetic field subclass that can *wrap* a Celeritas field (@whokion, #874)

celer-g4 app enhancements:
- [x] Add none/uniform/RZ magnetic field map option to `DetectorConstruction::ConstructSDandField`
- [x] Add `CELER_DISABLE` option to run geant4-only mode (#860 )
- [x] Add track step counter via `G4UserSteppingAction` (@amandalund #862 )
- [x] ~~(possibly?) Add RootStepWriter to  `G4UserSteppingAction`~~

See also #670 for replacing the validation app.

sethrj:
I think 
> Add stepping counter (track energy vs number of substeps) for magnetic field (need @whokion help)

isn't realistic to add to the main code (likely too expensive as a runtime diagnostic) but it would be a good thing to hack in on a branch?

PR #1006, opened 2023-11-03
Add diagnostic to count number of field substeps vs. track energy

amandalund:
@sethrj this is the field diagnostic that I updated to tally the number of substeps per track as a function of track energy (see #850). The extra state data needed (number of substeps, pre-step energy) is only allocated if the diagnostic is enabled; it shouldn't be expensive, but the changes still might be too intrusive, so we can definitely just close this without merging and keep it in the history. As an example, here's a plot of the diagnostic output from running the cms2018 regression problem:
field-diagnostic1


sethrj:
beautiful

sethrj:
That has quite a bit of interesting structure...

amandalund:
Yeah, I'm not sure what causes it to oscillate like that...

sethrj:
@whokion Can you scope out what would be involved for creating a similar plot for the Geant4 substep count?

src/celeritas/field/FieldPropagator.hh
```diff
@@ -180,6 +180,7 @@ CELER_FUNCTION auto FieldPropagator<DriverT, GTV>::operator()(real_type step)
         // Advance up to (but probably less than) the remaining step length
         DriverResult substep = driver_.advance(remaining, state_);
         CELER_ASSERT(substep.step > 0 && substep.step <= remaining);
+        ++result.num_substeps;
```
sethrj, line 183: One other useful diagnostic would be to use the "diagnostic substepper" (or equivalent) to count the number of *integrations* as a function of energy.

PR #1007, opened 2023-11-04
Refactor detector construction and SharedParams output

sethrj:
This is a minor refactor of the `celer-g4` app's `DetectorConstruction` in preparation for adding a Celeritas-based simple calorimeter (which will still work with Celeritas disabled). Changes include:
- `SharedParams` can now "lazily" create several helper objects (Geant geometry wrapper, output registry, stream count) that are needed by Celeritas classes that can be used in standalone Geant4
- `SharedParams::Finalize()` should now be called even when Celeritas is disabled: this now writes diagnostic output, and defaults to a `celeritas.json` filename if none is available
- Geant diagnostics now outputs the build information if Celeritas is disabled
- A new `static bool SharedParams::CeleritasDisabled` hides and standardizes the `CELER_DISABLE` environment variable

As part of the simple calorimeter work:
- Simple calo now allows an optional label rather than a hardcoded one
- Detector construction now takes shared params
- The `DetectorContsruction::Construct()` method has been split up into functions for loading geometry and field

I also moved `GetThreadId` out of `SetupOptions` into a separate celeritas utility, consolidating it with `get_num_threads` formerly in `GeantSetup`, and moving both into a new `celeritas/ext/GeantUtils.hh` directory.

### DIFF
--- a/app/celer-g4/CMakeLists.txt
+++ b/app/celer-g4/CMakeLists.txt
@@ -91,10 +91,7 @@ configure_file(
   "simple-cms-test.inp.json.in"
   "simple-cms-test.inp.json" @ONLY
 )
-add_test(NAME "app/celer-g4"
-  COMMAND "$<TARGET_FILE:celer-g4>"
-  "simple-cms-test.inp.json"
-)
+
 set(_env
   ${CELER_G4ENV}
   "CELER_LOG_LOCAL=debug"
@@ -110,25 +107,60 @@ if(CELERITAS_USE_OpenMP)
 endif()
 
 if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
-  set(_props RESOURCE_LOCK gpu)
-  set(_labels app gpu)
+  set(_needs_device)
 else()
-  set(_props)
-  set(_labels app)
-endif()
-if(NOT CELERITAS_USE_HepMC3 OR NOT CELERITAS_USE_Geant4
-    OR CELERITAS_CORE_GEO STREQUAL "Geant4")
-  list(APPEND _props DISABLED true)
+  set(_needs_device DISABLED true)
 endif()
 
+set(_props)
+if(NOT CELERITAS_USE_HepMC3 OR NOT CELERITAS_USE_Geant4)
+  set(_props DISABLED true)
+endif()
+if(CELERITAS_CORE_GEO STREQUAL "Geant4")
+  # FIXME: geant4 navigator doesn't work at scale
+  set(_props DISABLED true)
+endif()
 if(CELERITAS_REAL_TYPE STREQUAL "float")
   # FIXME: propagation fails in single precision
-  list(APPEND _props DISABLED true)
+  set(_props DISABLED true)
 endif()
 
-set_tests_properties("app/celer-g4" PROPERTIES
+### CELERITAS OFFLOADING WITH GPU ###
+add_test(NAME "app/celer-g4:gpu"
+  COMMAND "$<TARGET_FILE:celer-g4>" "simple-cms-test.inp.json"
+)
+set_tests_properties("app/celer-g4:gpu" PROPERTIES
   ENVIRONMENT "${_env}"
-  LABELS "${_labels}"
+  RESOURCE_LOCK gpu
+  LABELS "app;nomemcheck;gpu"
+  ${_props}
+  ${_needs_device}
+)
+
+### CELERITAS OFFLOADING WITH CPU ###
+list(APPEND _env
+  "CELER_DISABLE_DEVICE=1"
+)
+add_test(NAME "app/celer-g4:cpu"
+  COMMAND "$<TARGET_FILE:celer-g4>" "simple-cms-test.inp.json"
+)
+set_tests_properties("app/celer-g4:cpu" PROPERTIES
+  ENVIRONMENT "${_env}"
+  RESOURCE_LOCK gpu
+  LABELS "app;nomemcheck"
+  ${_props}
+)
+
+### NO CELERITAS OFFLOADING ###
+list(APPEND _env
+  "CELER_DISABLE=1"
+)
+add_test(NAME "app/celer-g4:none"
+  COMMAND "$<TARGET_FILE:celer-g4>" "simple-cms-test.inp.json"
+)
+set_tests_properties("app/celer-g4:none" PROPERTIES
+  ENVIRONMENT "${_env}"
+  LABELS "app;nomemcheck"
   ${_props}
 )
 
--- a/app/celer-g4/CMakeLists.txt
+++ b/app/celer-g4/CMakeLists.txt
@@ -57,45 +57,25 @@ if(NOT CELERITAS_BUILD_TESTS)
   return()
 endif()
 
-if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
-  if(CMAKE_BUILD_TYPE STREQUAL "Release")
-    set(CELERG4_MAX_NUM_TRACKS 524288)
-    set(CELERG4_INIT_CAPACITY 4194304)
-  else()
-    # Use smaller number of tracks when running on CPU
-    set(CELERG4_MAX_NUM_TRACKS 2048)
-    set(CELERG4_INIT_CAPACITY 65536)
-  endif()
-else()
-  # Use smaller number of tracks when running on CPU
-  set(CELERG4_MAX_NUM_TRACKS 1024)
-  set(CELERG4_INIT_CAPACITY 32768)
-endif()
-
+set(_driver "${CMAKE_CURRENT_SOURCE_DIR}/test-harness.py")
+set(_exe "$<TARGET_FILE:celer-g4>")
+set(_model "${CELER_APP_DATA_DIR}/simple-cms.gdml")
 if(NOT CELERITAS_DEBUG)
-  set(CELERG4_EVENTS
+  set(_events
     "${CELER_APP_DATA_DIR}/ttbarsplit-12evt-1108prim.hepmc3")
 else()
   # Use fewer events for debug
-  set(CELERG4_EVENTS
+  set(_events
     "${CELER_APP_DATA_DIR}/ttbarbits-12evt-118prim.hepmc3")
 endif()
-if(CELERITAS_USE_ROOT)
-  set(CELERG4_WRITE_SD_HITS "true")
-  set(CELERG4_OFFLOAD_FILE "simple-cms.offloaded.root")
-else()
-  set(CELERG4_WRITE_SD_HITS "false")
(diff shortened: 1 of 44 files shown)


src/accel/SharedParams.cc
```diff
@@ -219,19 +249,24 @@ SharedParams::SharedParams(SetupOptions const& options)
...
  * Clear shared data after writing out diagnostics.
  *
- * This must be executed exactly *once* across all threads and at the end of
+ * This should be executed exactly *once* across all threads and at the end of
  * the run.
  */
 void SharedParams::Finalize()
 {
-    CELER_EXPECT(*this);
+    std::lock_guard scoped_lock{updating_mutex()};
```
amandalund, line 270: Why do we need a mutex here? Shouldn't this only be called by a single thread?

src/accel/SharedParams.cc
```diff
@@ -249,6 +284,75 @@ void SharedParams::Finalize()
     CELER_ENSURE(!*this);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Lazily created output registry.
```
amandalund:
```suggestion
 * Lazily obtained number of streams.
```

src/accel/SharedParams.cc
```diff
@@ -141,9 +143,33 @@ build_g4_particles(std::shared_ptr<ParticleParams const> const& particles,
...
 
+bool SharedParams::CeleritasDisabled()
+{
+    static bool const result = [] {
+        if (celeritas::getenv("CELER_DISABLE").empty())
+            return false;
+
+        CELER_LOG(info)
+            << "Disabling Celeritas offloading since the 'CELER_DISABLE' "
+               "environment variable is present and non-empty";
```
amandalund, line 168: We could remove the duplicate log message in celer-g4 now.

app/celer-g4/TrackingAction.cc
```diff
@@ -53,19 +45,15 @@ TrackingAction::TrackingAction(SPConstParams params,
...
 {
-    if (disable_offloading_)
-        return;
-
     CELER_EXPECT(track);
-    CELER_EXPECT(*params_);
-    CELER_EXPECT(*transport_);
+    CELER_EXPECT(static_cast<bool>(params_)
+                 == !SharedParams::CeleritasDisabled());
+    CELER_EXPECT(static_cast<bool>(params_) == static_cast<bool>(transport_));
```
amandalund:
```suggestion
    CELER_EXPECT(static_cast<bool>(*params_)
                 == !SharedParams::CeleritasDisabled());
    CELER_EXPECT(static_cast<bool>(*params_) == static_cast<bool>(*transport_));
```

app/celer-g4/TrackingAction.cc
```diff
@@ -53,19 +45,15 @@ TrackingAction::TrackingAction(SPConstParams params,
...
 
-    static G4ParticleDefinition const* const allowed_particles[] = {
-        G4Gamma::Gamma(),
-        G4Electron::Electron(),
-        G4Positron::Positron(),
-    };
+    if (!params_)
```
amandalund:
```suggestion
    if (SharedParams::CeleritasDisabled())
```

src/accel/SharedParams.cc
```diff
@@ -430,25 +547,44 @@ void SharedParams::initialize_core(SetupOptions const& options)
...
-        << "JSON support is not enabled, so no output will be written to \""
-        << output_filename_ << '"';
-#endif
+    std::string filename = output_filename_;
+    if (CELERITAS_USE_JSON && !params_ && filename.empty())
+    {
+        // Setup was not called but JSON is available: make a default filename
+        filename = "celeritas.json";
+        CELER_LOG(debug) << "Setting default Celeritas output filename";
+    }
```
amandalund: Since `output_filename_` is set in `initialize_core()`, if offloading is disabled this default name will be used instead of the filename in the `SetupOptions`.

amandalund (commented):
Looks good @sethrj, nice refactor!

src/accel/SharedParams.cc
sethrj: Should be, yes, but I figured someone is likely to call it from all threads... is the right move to allow it safely (basically no performance penalty since we only call it once per run) or to let it be undefined or to try to catch it?

src/accel/SharedParams.cc
sethrj: It's only the creation that's lazy 😉

src/accel/SharedParams.cc
sethrj: Thanks, missed that one!

app/celer-g4/TrackingAction.cc
sethrj: More descriptive and clearly less error prone 😅

src/accel/SharedParams.cc
sethrj: I think that's the only generic way to do this... we always load this through `celer-g4` but user apps won't usually provide a `SetupOptions`. Any ideas how to make both work better?

amandalund: Maybe we could use the G4UI, so if the output filename is set in the setup options we'll always use it? Something like:
```diff
diff --git a/src/accel/SetupOptionsMessenger.cc b/src/accel/SetupOptionsMessenger.cc
index a36f77bdd..0177d2aa5 100644
--- a/src/accel/SetupOptionsMessenger.cc
+++ b/src/accel/SetupOptionsMessenger.cc
@@ -89,6 +89,7 @@ class CelerCommand : public G4UIcommand
     using G4UIcommand::G4UIcommand;
 
     virtual void apply(G4String const& newValue) const = 0;
+    virtual G4String get() const = 0;
 };
 
 template<class T>
@@ -125,6 +126,11 @@ class CelerParamCommand final : public CelerCommand
         *this->dest_ = static_cast<T>(converted);
     }
 
+    G4String get() const final
+    {
+        return CmdTraits::to_string(*this->dest_);
+    }
+
   private:
     T* dest_;
 };
@@ -226,5 +232,13 @@ void SetupOptionsMessenger::SetNewValue(G4UIcommand* cmd, G4String val)
     celer_cmd->apply(val);
 }
 
+//---------------------------------------------------------------------------//
+//! Get the value of the given command
+G4String SetupOptionsMessenger::GetCurrentValue(G4UIcommand* cmd)
+{
+    auto* celer_cmd = dynamic_cast<CelerCommand*>(cmd);
+    return celer_cmd->get();
+}
+
 //---------------------------------------------------------------------------//
 }  // namespace celeritas
diff --git a/src/accel/SetupOptionsMessenger.hh b/src/accel/SetupOptionsMessenger.hh
index 68b5817a7..9e5f68e44 100644
--- a/src/accel/SetupOptionsMessenger.hh
+++ b/src/accel/SetupOptionsMessenger.hh
@@ -65,6 +65,7 @@ class SetupOptionsMessenger : public G4UImessenger
 
   protected:
     void SetNewValue(G4UIcommand* command, G4String newValue) override;
+    G4String GetCurrentValue(G4UIcommand* command) override;
 
   private:
     std::vector<std::unique_ptr<G4UIdirectory>> directories_;
diff --git a/src/accel/SharedParams.cc b/src/accel/SharedParams.cc
index a6fbd4e63..ccc11ab2d 100644
--- a/src/accel/SharedParams.cc
+++ b/src/accel/SharedParams.cc
@@ -18,6 +18,7 @@
 #include <G4ParticleTable.hh>
 #include <G4RunManager.hh>
 #include <G4Threading.hh>
+#include <G4UImanager.hh>
 
 #include "celeritas_config.h"
 #include "corecel/Assert.hh"
@@ -554,7 +555,8 @@ void SharedParams::try_output() const
         return;
     }
 
-    std::string filename = output_filename_;
+    G4UImanager* ui = G4UImanager::GetUIpointer();
+    std::string filename = ui->GetCurrentValues("/celer/outputFile");
     if (CELERITAS_USE_JSON && !params_ && filename.empty())
     {
         // Setup was not called but JSON is available: make a default filename
```

app/celer-g4/TrackingAction.cc
```diff
@@ -46,11 +46,11 @@ TrackingAction::TrackingAction(SPConstParams params,
 void TrackingAction::PreUserTrackingAction(G4Track const* track)
 {
     CELER_EXPECT(track);
-    CELER_EXPECT(static_cast<bool>(params_)
+    CELER_EXPECT(static_cast<bool>(*params_)
                  == !SharedParams::CeleritasDisabled());
-    CELER_EXPECT(static_cast<bool>(params_) == static_cast<bool>(transport_));
+    CELER_EXPECT(static_cast<bool>(*params_) == static_cast<bool>(transport_));
```
amandalund: One more:
```suggestion
    CELER_EXPECT(static_cast<bool>(*params_) == static_cast<bool>(*transport_));
```

src/accel/SharedParams.cc
amandalund: But it's returning the number of streams, not the registry :)

src/accel/SharedParams.cc
amandalund: Yeah, this way seems better.

src/accel/SharedParams.cc
sethrj: 🤦‍♂️

app/celer-g4/TrackingAction.cc
sethrj: `******transport_`

src/accel/SharedParams.cc
sethrj: Brilliant! Added.

sethrj:
OK @amandalund finally think I have everything! I added a no-celeritas test so I confirmed that the code was broken and your suggestion fixed it 😆

app/celer-g4/CMakeLists.txt
```diff
@@ -110,25 +107,60 @@ if(CELERITAS_USE_OpenMP)
...
+### CELERITAS OFFLOADING WITH CPU ###
+list(APPEND _env
+  "CELER_DISABLE_DEVICE=1"
+)
+add_test(NAME "app/celer-g4:cpu"
+  COMMAND "$<TARGET_FILE:celer-g4>" "simple-cms-test.inp.json"
+)
+set_tests_properties("app/celer-g4:cpu" PROPERTIES
+  ENVIRONMENT "${_env}"
+  RESOURCE_LOCK gpu
```
amandalund:
```suggestion
```

amandalund (approved):
Awesome, thanks @sethrj!

I'm not sure why some of the celer-g4 tests are failing on the CI...

app/celer-g4/CMakeLists.txt
sethrj: Yeah I've refactored all the celer-g4 test code: the output files were stepping on each other, and now there's a driver to make sure everyone gets their own filenames.

PR #1008, opened 2023-11-07
Explicitely set shared mem carveout

### LINKED ISSUE #1009: Analyze performance impact of shared memory carveout
#1008 explores the performance impact of explicitly allocating the entire shared mem/L1 data cache to L1 as Celeritas kernels currently don't use shared mem. An initial assessment using the [regression](https://github.com/celeritas-project/regression/tree/7d601668e5b72b78f8fabf6358a61408012571ad) problems shows no difference in performance so we're putting this PR on hold for now. 

We should reassess performance after we improve L1 cache utilization. We currently have an L1 cache throughput of ~4% with Vecgeom and ~18% with Orange for the field propagation kernel

esseivaju:
Since Celeritas kernels don't use shared memory, explicitly set the shared memory carveout to 0 to maximize L1 cache size for devices sharing L1/shmem data cache. 

Still need to check the performance impact.

### DIFF
--- a/src/corecel/sys/KernelParamCalculator.device.hh
+++ b/src/corecel/sys/KernelParamCalculator.device.hh
@@ -230,7 +230,7 @@ void KernelParamCalculator::set_carveout(F* kernel_func_ptr,
     if (!attrs.shared_mem)
     {
         CELER_DEVICE_CALL_PREFIX(FuncSetAttribute(
-            kernel_func_ptr,
+            reinterpret_cast<void const*>(kernel_func_ptr),
             CELER_DEVICE_PREFIX(FuncAttributePreferredSharedMemoryCarveout),
             0));
         attrs.shared_mem_carveout = 0;
--- a/src/corecel/sys/KernelParamCalculator.device.hh
+++ b/src/corecel/sys/KernelParamCalculator.device.hh
@@ -236,6 +236,9 @@ void KernelParamCalculator::set_carveout(F* kernel_func_ptr) const
             CELER_DEVICE_PREFIX(FuncAttributePreferredSharedMemoryCarveout),
             CELER_DEVICE_PREFIX(SharedmemCarveoutMaxL1)));
     }
+#else
+    CELER_DISCARD(kernel_func_ptr);
+    CELER_ASSERT_UNREACHABLE();
 #endif
 }
 
--- a/src/corecel/sys/KernelParamCalculator.device.hh
+++ b/src/corecel/sys/KernelParamCalculator.device.hh
@@ -237,6 +237,7 @@ void KernelParamCalculator::set_carveout(F* kernel_func_ptr,
     }
 #else
     CELER_DISCARD(kernel_func_ptr);
+    CELER_DISCARD(attrs);
     CELER_ASSERT_UNREACHABLE();
 #endif
 }
--- a/src/corecel/sys/KernelParamCalculator.device.hh
+++ b/src/corecel/sys/KernelParamCalculator.device.hh
@@ -130,7 +130,7 @@ class KernelParamCalculator
     void register_kernel(std::string_view name, KernelAttributes&& attributes);
     void log_launch(size_type min_num_threads) const;
     template<class F>
-    void set_carveout(F* kernel_func_ptr) const;
+    void set_carveout(F* kernel_func_ptr, KernelAttributes& attrs) const;
 };
 
 //---------------------------------------------------------------------------//
@@ -157,8 +157,8 @@ template<class F>
 KernelParamCalculator::KernelParamCalculator(std::string_view name,
                                              F* kernel_func_ptr)
 {
-    this->set_carveout(kernel_func_ptr);
     auto attrs = make_kernel_attributes(kernel_func_ptr);
+    this->set_carveout(kernel_func_ptr, attrs);
     CELER_ASSERT(attrs.threads_per_block > 0);
     block_size_ = attrs.threads_per_block;
     this->register_kernel(name, std::move(attrs));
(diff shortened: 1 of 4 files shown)


src/corecel/sys/KernelParamCalculator.device.hh
```diff
@@ -128,6 +129,8 @@ class KernelParamCalculator
 
     void register_kernel(std::string_view name, KernelAttributes&& attributes);
     void log_launch(size_type min_num_threads) const;
+    template<class F>
+    void set_carveout(F* kernel_func_ptr) const;
```
esseivaju: setting the carveout ratio here works and requires minimal changes but I'm not convinced it makes sense semantically since the class is called `KernelParamCalculator` 🤔

esseivaju:
@sethrj I reran the regression problems; the timing hasn't changed much.
```diff
 | Problem                      | Geometry |      Speedup |
 | ---------------------------- | -------- | ------------ |
-| cms2018 [Z]                  | vecgeom  |  9.4× (±0.1) |
+| cms2018 [Z]                  | vecgeom  |  9.4× (±0.3) |
 | cms2018+field+msc [ZFM]      | vecgeom  |  5.9× (±0.1) |
-| simple-cms+field [BF]        | orange   | 37.9× (±0.7) |
-| simple-cms+field+msc [BFM]   | orange   | 39.8× (±1.1) |
-|                              | vecgeom  | 34.8× (±0.7) |
-| simple-cms+msc [BM]          | orange   | 43.0× (±0.6) |
-| testem15 [A]                 | orange   | 42.3× (±0.8) |
-| testem15+field [AF]          | orange   | 48.8× (±0.9) |
-| testem15+field+msc [AFM]     | orange   | 50.3× (±1.0) |
-|                              | vecgeom  | 42.4× (±0.1) |
-| testem3-flat [C]             | orange   | 37.8× (±0.5) |
+| simple-cms+field [BF]        | orange   | 37.5× (±1.2) |
+| simple-cms+field+msc [BFM]   | orange   | 39.8× (±1.3) |
+|                              | vecgeom  | 34.6× (±0.3) |
+| simple-cms+msc [BM]          | orange   | 43.4× (±0.4) |
+| testem15 [A]                 | orange   | 42.1× (±1.1) |
+| testem15+field [AF]          | orange   | 48.6× (±0.5) |
+| testem15+field+msc [AFM]     | orange   | 50.8× (±0.6) |
+|                              | vecgeom  | 42.1× (±0.4) |
+| testem3-flat [C]             | orange   | 37.7× (±0.3) |
 |                              | vecgeom  | 30.4× (±0.2) |
-| testem3-flat+field [CF]      | orange   | 31.5× (±0.1) |
-| testem3-flat+field+msc [CFM] | orange   | 32.9× (±0.3) |
-|                              | vecgeom  | 13.1× (±0.1) |
-| testem3-flat+msc [CM]        | orange   | 44.1× (±0.4) |
+| testem3-flat+field [CF]      | orange   | 30.9× (±0.4) |
+| testem3-flat+field+msc [CFM] | orange   | 33.0× (±0.2) |
+|                              | vecgeom  | 13.2× (±0.1) |
+| testem3-flat+msc [CM]        | orange   | 43.7× (±0.6) |
```

sethrj:
Well at least this change was an easy one to implement and test?😔

esseivaju:
yeah, do we still want to merge this PR or should we close it? It might be worth to test it on AMD GPUs.

src/corecel/sys/KernelParamCalculator.device.hh
sethrj: Well, shared memory is a kernel launch parameter so maybe it makes sense?

src/corecel/sys/KernelParamCalculator.device.hh
```diff
@@ -214,5 +218,29 @@ auto KernelParamCalculator::operator()(size_type min_num_threads) const
...
+#ifdef CELER_DEVICE_SOURCE
+    CELER_DEVICE_PREFIX(FuncAttributes) attr;
+    CELER_DEVICE_CALL_PREFIX(FuncGetAttributes(
+        &attr, reinterpret_cast<void const*>(kernel_func_ptr)));
+    if (!attr.sharedSizeBytes)
+    {
+        CELER_DEVICE_CALL_PREFIX(FuncSetAttribute(
+            kernel_func_ptr,
+            CELER_DEVICE_PREFIX(FuncAttributePreferredSharedMemoryCarveout),
+            CELER_DEVICE_PREFIX(SharedmemCarveoutMaxL1)));
```
sethrj: Looks like this is CUDA-only: https://cloud.cees.ornl.gov/jenkins-ci/blue/rest/organizations/jenkins/pipelines/Celeritas/branches/PR-1008/runs/2/nodes/18/steps/274/log/?start=0

src/corecel/sys/KernelParamCalculator.device.hh
```diff
@@ -214,5 +218,29 @@ auto KernelParamCalculator::operator()(size_type min_num_threads) const
...
+/*!
+ * Maximize L1 cache size if the kernel doesn't use static shared mem.
+ */
+template<class F>
+void KernelParamCalculator::set_carveout(F* kernel_func_ptr) const
+{
+#ifdef CELER_DEVICE_SOURCE
+    CELER_DEVICE_PREFIX(FuncAttributes) attr;
+    CELER_DEVICE_CALL_PREFIX(FuncGetAttributes(
+        &attr, reinterpret_cast<void const*>(kernel_func_ptr)));
```
sethrj: Since the getattr call is so expensive on some platforms (~1s on some AMD machines) it would be better if we could reuse the attributes we already fetched (might need to refactor the kernel registry return value).

sethrj (commented):
Since we're not seeing any performance benefit maybe we should just shelve this for now? It might be a good time to start doing some hardware counter profiling again...

esseivaju:
Sounds good, so we can close this PR and create an issue to keep track of it. CI failures are spurious, it should pass now.  

> It might be a good time to start doing some hardware counter profiling again...

I did a nsight compute profiling yesterday of testem3+field+msc with orange and Vecgeom for the propagate kernel if you're interested in taking a look at the files.

PR #1011, opened 2023-11-09
Add environment variable to disable ROOT and update celer-g4 input

sethrj:
The ROOT enable/disable toggle in the celer-g4 input is a little confusing. This refactors the input and ROOT option so that there's an SD selection (for now, either "none" or "event_hit"; soon to include "simple_calo") and an environment variable for the "unusual" decision to disable ROOT output even when collecting hits.

The `RootFileManager::use_root` option should now be used globally to determine whether to enable/disable ROOT. For now the only thing that works with this option is the RootIO calls for writing sensitive detectors.

I'm also removing some deprecated options from the input (and adding adapters for the JSON input). Instead of the plethora of accessors in `GlobalSetup` we should move toward accessing the options directly.

I've removed the "root buffer size" for simplicity. If that's something we need to configure an environment variable might be better for that too.


app/celer-g4/RunInput.cc
```diff
@@ -26,6 +27,34 @@ char const* to_cstring(PhysicsListSelection value)
...
+ * Whether the run arguments are valid.
+ */
+RunInput::operator bool() const
+{
+    return !geometry_file.empty() && (primary_options || !event_file.empty())
+           && physics_list < PhysicsListSelection::size_
+           && (field == no_field() || field_options)
+           && ((num_track_slots > 0 && max_events > 0 && max_steps > 0
+                && initializer_capacity > 0 && secondary_stack_factor > 0)
+               || !SharedParams::CeleritasDisabled())
```
amandalund:
```suggestion
               || SharedParams::CeleritasDisabled())
```

amandalund (approved):
Nice, thanks @sethrj!

stognini:
Awesome. I assume you prepared `RootIO` to use `RootFileManager` but that will be done when the IO is overhauled, since for now we're using `unique_ptr<TFile>`.

stognini (approved):

PR #1012, opened 2023-11-10
Refactor ParticleState and ParticleParams to SoA

esseivaju:
Attempt at refactoring ParticleData to SoA, from initial performance tests, it doesn't make a difference. 

It also updates `CollectionStorage` to use `LdgSpan` as backing storage when possible and adds support for reading `Quantity` from `LdgIterator`


sethrj:
Regarding pos/dir, you mean unfolding the x/y/z components and returning by value rather than a pointer to the first component? Won't that potentially increase register pressure?

src/celeritas/phys/ParticleData.hh
```diff
@@ -43,6 +43,12 @@ struct ParticleRecord
     }
 };
 
+enum class particle_partner : bool
```
sethrj: Maybe `MatterType`?  `ParticleCategory`?

sethrj (commented):
I think we should definitely merge this even if it's not noticeably faster: if nothing else, it might use *slightly* less memory because there's no padding for the particle struct?

esseivaju:
> Regarding pos/dir, you mean unfolding the x/y/z components and returning by value rather than a pointer to the first component? Won't that potentially increase register pressure?

That was the idea, but I'm not sure it's going to make a difference since the  FieldPropagator copies these to register variable (OdeState).

> I think we should definitely merge this even if it's not noticeably faster: if nothing else, it might use slightly less memory because there's no padding for the particle struct?

Yeah, it helps a bit with padding. Maybe we could also add `Quantity` support in `LdgSpan`

src/celeritas/phys/ParticleData.hh
```diff
@@ -19,36 +19,24 @@ namespace celeritas
...
-        return 0;
-    }
+enum class MatterType : char
+{
+    particle,
+    antiparticle
 };
 
+//! Value of decay_constant for a stable particle
+constexpr real_type stable_decay_constant = 0;
```
sethrj: Maybe this should go in `celeritas/Constants.hh`?

test/celeritas/MockTestBase.cc
```diff
@@ -105,24 +105,29 @@ auto MockTestBase::build_particle() -> SPConstParticle
...
     ParticleParams::Input inp;
-    inp.push_back({"gamma", pdg::gamma(), zero, zero, stable});
-    inp.push_back(
-        {"celeriton", PDGNumber{1337}, MevMass{1}, ElementaryCharge{1}, stable});
+    inp.push_back({"gamma", pdg::gamma(), zero, zero, stable_decay_constant});
+    inp.push_back({"celeriton",
+                   PDGNumber{1337},
+                   MevMass{1},
+                   ElementaryCharge{1},
+                   stable_decay_constant});
```
sethrj, line 115: Would've been fewer LOC changed if you just used `constexpr auto stable = stable_decay_constant;` 😉

src/celeritas/phys/ParticleData.hh
sethrj: Also this is missing an `inline`—otherwise it won't work with anything that takes `real_type const&` (which often shows up in templated containers for example).

sethrj (approved):

esseivaju:
I ran the [regression tests](https://github.com/celeritas-project/regression/tree/particledata-soa). As mentioned before, performances are mostly equivalent. The occupancy for a few kernels has changed, most likely due to the change to `CollectionStorage` that we missed.

PR #1014, opened 2023-11-13
Add G4SD "simple calorimeter" that outputs to JSON

sethrj:
This adds a "simple calorimeter" (event and thread integrated, differential per logical volume) detector that saves its output to JSON in the same format as the on-GPU calorimeters. This will let us do an accurate comparison of the cost for converting hits through the Geant4 layer versus directly storing on GPU.

### DIFF
--- a/src/accel/GeantSimpleCalo.cc
+++ b/src/accel/GeantSimpleCalo.cc
@@ -7,8 +7,6 @@
 //---------------------------------------------------------------------------//
 #include "GeantSimpleCalo.hh"
 
-#include <mutex>
-
 #include "corecel/cont/Range.hh"
 #include "corecel/io/Logger.hh"
 #include "celeritas/ext/GeantGeoParams.hh"
@@ -83,6 +81,10 @@ GeantSimpleCalo::GeantSimpleCalo(std::string name,
                        << this->label() << "'");
     }
 
+    // Resize data
+    storage_->num_threads = params_->num_streams();
+    storage_->data.resize(storage_->num_threads);
+
     CELER_ENSURE(!storage_->name.empty());
     CELER_ENSURE(storage_->volume_to_index.size() == volumes_.size());
 }
@@ -93,8 +95,6 @@ GeantSimpleCalo::GeantSimpleCalo(std::string name,
  */
 auto GeantSimpleCalo::MakeSensitiveDetector() -> UPSensitiveDetector
 {
-    this->setup_storage();
-
     UPSensitiveDetector detector;
 
     // Get thread ID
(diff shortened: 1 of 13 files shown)


sethrj:
@amandalund This is finally ready for review.

src/accel/detail/GeantSimpleCaloStorage.hh
```diff
@@ -0,0 +1,48 @@
...
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Shared storage for a single GeantSimpleCalo.
+ *
+ * This is created by the \c SharedGeantSimpleCalo and passed to all the \c
```
amandalund:
```suggestion
 * This is created by the \c GeantSimpleCalo and passed to all the \c
```

src/accel/detail/GeantSimpleCaloSD.cc
```diff
@@ -0,0 +1,65 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add energy deposition from this step to the corresponding logical volume.
+ */
+bool GeantSimpleCaloSD::ProcessHits(G4Step* g4step, G4TouchableHistory*)
+{
+    CELER_ASSERT(g4step);
+    CELER_ASSERT(g4step->GetPreStepPoint());
```
amandalund:
```suggestion
    CELER_EXPECT(g4step);
    CELER_EXPECT(g4step->GetPreStepPoint());
```

src/accel/GeantSimpleCalo.cc
```diff
@@ -0,0 +1,248 @@
...
+    if (storage_->num_threads == 0)
+    {
+        // Lock and check again
+        static std::mutex setup_mutex;
+        std::lock_guard scoped_lock{setup_mutex};
+        if (storage_->num_threads == 0)
+        {
+            // Resize data
+            storage_->num_threads = params_->num_streams();
+            storage_->data.resize(storage_->num_threads);
```
amandalund: Any reason not to resize the storage in the constructor instead of using a mutex on a worker thread?

src/accel/GeantSimpleCalo.cc
```diff
@@ -0,0 +1,248 @@
...
+ * Return the key in the JSON output.
+ */
+std::string GeantSimpleCalo::label() const
+{
+    return storage_->name;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Allocate storage space for all threads.
```
amandalund:
```suggestion
 * Write output to the given JSON object.
```

amandalund (approved):
Nice, this look great @sethrj! I tried it (and the Celeritas simple calo) out with testem3 and the results look like we expect:

testem3-simple-calo

src/accel/GeantSimpleCalo.cc
sethrj: Well, I thought I needed it due to ordering constraints (shared params isn't initialized when the simple calo is constructed), but looking at it again I don't think the number of streams is set by celeritas at either point. I'll see if it works in the constructor :)

PR #1015, opened 2023-11-15
Use consistent energy intervals for model applicability and selection

amandalund:
- Change the definition of applicability from `(lower, upper]` to `[lower, upper)` to match the behavior of the `GridIdFinder` used for model selection
- Use assertions to check that the incident particle energy is above the secondary production cut in the interactors


sethrj (approved):
I think this is sensible. We might need to check more carefully for edge cases (emitting secondaries/exiting particle with zero or slightly negative energy) but that might've been necessary before too.

I'm going to talk to @dalg24 today to see if we can get the CI back online 😢

src/celeritas/em/interactor/SeltzerBergerInteractor.hh
```diff
@@ -128,6 +128,7 @@ CELER_FUNCTION SeltzerBergerInteractor::SeltzerBergerInteractor(
     CELER_EXPECT(particle.particle_id() == shared_.ids.electron
                  || particle.particle_id() == shared_.ids.positron);
     CELER_EXPECT(gamma_cutoff_ > zero_quantity());
+    CELER_EXPECT(inc_energy_ > gamma_cutoff_);
```
stognini: Shouldn't we also add the upper limit?
```cpp
CELER_EXPECT(inc_energy_ < detail::seltzer_berger_limit());
```

stognini:
Awesome. Thanks @amandalund !

stognini (approved):

PR #1016, opened 2023-11-15
Add support for VecGeom surface model

mrguilima:
For now, this is the list of GDML files supported by VecGeom's surface model:
```
testem15.gdml
two-boxes.gdml
four-steel-slabs.gdml
simple-cms.gdml
testem3-flat.gdml.
```

Test celeritas_geo_Geometry passes for both geometries separately, but not for both geometries in the same job as it does for volume-based model.

Also tested with both CUDA=ON and CUDA=OFF.

In my local area, this is the list of failing tests:
```
The following tests FAILED:
	101 - celeritas/ext/Vecgeom:FourLevelsTest.* (Failed)
	102 - celeritas/ext/Vecgeom:SolidsTest.* (Failed)
	103 - celeritas/ext/Vecgeom:CmseTest.* (Failed)
	104 - celeritas/ext/Vecgeom:FourLevelsGeantTest.* (Failed)
	105 - celeritas/ext/Vecgeom:SolidsGeantTest.* (Failed)
	113 - celeritas/ext/GeantVolumeMapper (Failed)
	118 - celeritas/field/FieldPropagator (Failed)
	121 - celeritas/geo/Geometry (SEGFAULT)
	122 - celeritas/geo/GeoMaterial (SEGFAULT)
	124 - celeritas/global/AlongStep:-Em3*:SimpleCms* (Failed)
	137 - celeritas/global/Stepper:TestEm3Msc.* (Failed)
	138 - celeritas/global/Stepper:TestEm3MscNofluct.* (Failed)
	140 - celeritas/global/Stepper:OneSteelSphere.* (Failed)
	182 - celeritas/user/StepCollector (SEGFAULT)
	198 - app/demo-geo-check (Failed)
	199 - app/demo-rasterizer (Failed)
```


src/celeritas/ext/VecgeomParams.cc
```diff
@@ -403,5 +468,94 @@ void VecgeomParams::build_metadata()
...
+    if (std::string var = celeritas::getenv("CUDA_HEAP_SIZE"); !var.empty())
+    {
+        int heap_size = std::stoi(var);
+        CELER_VALIDATE(heap_size > 0,
+                       << "invalid CUDA_HEAP_SIZE=" << heap_size
+                       << " (must be positive)");
+        set_cuda_heap_size(heap_size);
+    }
+#endif
+    return true;
```
sethrj: Why return a bool that's always true?

src/celeritas/ext/VecgeomParams.cc
```diff
@@ -403,5 +468,94 @@ void VecgeomParams::build_metadata()
     }();
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Function for setting and/or checking size limits for CUDA stack and heap.
+ */
+bool VecgeomParams::setup_cuda_mem_limits()
```
sethrj: This also sets up the vecgeom verbosity, *and* most critically the limit setting is done because the first call to `CudaManager::Instance` has side effects (i.e. changing the cuda memory). This should be something more like "instantiate cuda volume manager".

src/celeritas/ext/VecgeomParams.cc
```diff
@@ -195,93 +258,21 @@ void VecgeomParams::build_tracking()
...
-            = {cudaLimitStackSize, cudaLimitMallocHeapSize};
-        char const* const cuda_attr_labels[] = {"stack size", "heap size"};
-        Array<std::size_t, std::size(cuda_attrs)> orig_limits;
-        for (auto i : range(orig_limits.size()))
-        {
-            CELER_CUDA_CALL(cudaDeviceGetLimit(&orig_limits[i], cuda_attrs[i]));
-        }
-
-        auto& cuda_manager = vecgeom::cxx::CudaManager::Instance();
+        CELER_ASSERT(setup_cuda_mem_limits());
```
sethrj: You *cannot* run anything with side effects inside an assertion! This means the cuda mem limits will not be set in release mode.

sethrj (changes requested):
I'm going to move one or two of the changes I made into a separate PR, including the addition of the CMS geometry vecgeom test (which doesn't seem to have made it into this PR).

sethrj (approved):

Issue #979, opened 2023-10-12
Crash in TrackSort test with VecGeom: illegal memory access 


mrguilima:
```
[ RUN      ] TestTrackSortActionIdEm3Stepper.device_is_sorted
status: Celeritas core setup complete
status: Celeritas core state initialization complete
/work1/celeritas/lima/work/cele/src/corecel/data/DeviceAllocation.cc:131: debug: While freeing device memory: /work1/celeritas/lima/work/cele/src/corecel/data/DeviceAllocation.cc:120:
celeritas: CUDA error: cudaFree(ptr) failed:
    an illegal memory access was encountered
/work1/celeritas/lima/work/cele/src/corecel/data/DeviceAllocation.cc:131: debug: While freeing device memory: /work1/celeritas/lima/work/cele/src/corecel/data/DeviceAllocation.cc:120:
celeritas: CUDA error: cudaFree(ptr) failed:
    an illegal memory access was encountered
/work1/celeritas/lima/work/cele/src/corecel/data/DeviceAllocation.cc:135: debug: Suppressing further DeviceFreeDeleter warning messages
/work1/celeritas/lima/work/cele/src/corecel/data/PinnedAllocator.t.hh:64: debug: While freeing pinned host memory: /work1/celeritas/lima/work/cele/src/corecel/d
ata/PinnedAllocator.t.hh:57:
celeritas: CUDA error: cudaFreeHost(p) failed:
    an illegal memory access was encountered
unknown file: Failure
C++ exception with description "/work1/celeritas/lima/work/cele/src/celeritas/track/detail/TrackSortUtils.cu:252:
celeritas: CUDA error: cudaStreamSynchronize(stream) failed:
    an illegal memory access was encountered" thrown in the test body.
Writing diagnostic output to screen because test failed
{
[... JSON output removed ...]
}
[  FAILED  ] TestTrackSortActionIdEm3Stepper.device_is_sorted (84 ms)
[----------] 2 tests from TestTrackSortActionIdEm3Stepper (89 ms total)

[----------] 2 tests from TestActionCountEm3Stepper
[ RUN      ] TestActionCountEm3Stepper.host_count_actions
unknown file: Failure
C++ exception with description "/work1/celeritas/lima/work/cele/src/corecel/sys/ScopedMem.cc:83:
celeritas: CUDA error: cudaMemGetInfo(&free, &total) failed:
    an illegal memory access was encountered" thrown in the test body.
[...]

PR #1017, opened 2023-11-16
Refactor vecgeom setup in preparation for surface support

sethrj:
Prerequisite to #1016 . This adds a helper class to simplify the limit save/restore, adds a simple CMS test, moves the vecgeom verbosity function to a static free function for reuse, and adds a function to check for the vecgeom build type.


mrguilima (changes requested):
I've some suggested changes to fix the new Vecgeom:SimpleCms* test, see branch `vg-pre-surf-refact` in my fork, see https://github.com/mrguilima/celeritas/tree/vg-pre-surf-refact

sethrj:
@mrguilima, your last patch https://github.com/mrguilima/celeritas/commit/daf58e3cae3f149d10a35929c1ddc823930a95ed isn't a problem with this PR. It's an extra thing for vecgeom v2 and would probably belong better on your branch, which is actually about supporting the newer vecgeom. This PR is *just* about laying the groundwork for it and adding tests that are known to work on v1.

mrguilima:
Ok regarding v1.x and ignoring v2.x for now.
In my local tests, there are two failing tests, trackSort with VecGeom enabled (see issue #979) and celer-g4:cpu.
Do they pass in your local tests?
Otherwise, LGTM.
Thanks!

mrguilima (approved):
LGTM

PR #1018, opened 2023-11-17
Optimize data access through ORANGE visitors

esseivaju:
`LocalSurfaceVisitor` and `TransformVisitor` were using raw pointers to interact with `Collection`.  Refactor `get_ptr` to return a Span so that values are read using `__ldg` when possible.


src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -157,23 +152,7 @@ CELER_FUNCTION T LocalSurfaceVisitor::get_item(Items<T> const& items,
...
- */
-template<class T>
-CELER_FUNCTION T const*
-LocalSurfaceVisitor::get_ptr(Items<T> const& items, ItemId<T> item)
-{
-    CELER_EXPECT(item < items.size());
-
-    T const* ptr = items[AllItems<T>{}].data();
-    return ptr + item.unchecked_get();
+    return items[range][item.unchecked_get()];
```
sethrj: While you're at it, could you rename the local variable `range` to something else? (It's shadowing the `celeritas::range` function.)

src/orange/surf/LocalSurfaceVisitor.hh
```diff
@@ -136,13 +131,13 @@ LocalSurfaceVisitor::operator()(F&& func, LocalSurfaceId id)
...
-    OpaqueId<real_type> offset
+    using ItemIdT = typename decltype(params_.reals)::ItemIdT;
+    ItemIdT offset
         = this->get_item(params_.real_ids, surfaces_.data_offsets, id);
     constexpr size_type size{T::StorageSpan::extent};
     CELER_ASSERT(offset + size <= params_.reals.size());
 
-    real_type const* data = this->get_ptr(params_.reals, offset);
-    return T{Span<real_type const, size>{data, size}};
+    return T{params_.reals[{offset, offset + size}]};
```
sethrj: Have you tested this is using LDGIterator and works on CUDA? Currently the surface constructors are written as
```c++
CELER_FUNCTION Plane::Plane(StorageSpan data)
    : normal_{data[0], data[1], data[2]}, d_{data[3]}
{}
```
where `StorageSpan` is a `Span<real_type const>`. We might need to template these constructors (`template<T>` and `Span<T>`? or `T const` and we need to change where the const in LdgValue belongs)?

esseivaju: yes, I'm building with CUDA, and tests pass locally. You can copy construct a `Span<T const>` from a `LdgSpan<T>` and the other way around. I tried to change the `StorageSpan` of [surfaces](https://github.com/celeritas-project/celeritas/pull/1000#issuecomment-1786195481) to `LdgSpan`. I haven't investigated why it breaks. Are these getting written to and read during a kernel?

sethrj: Read only, and it should always be from global memory too... but maybe there's some place it's constructing from a different memory space? As long as both span and ldgspan are available it should be ok...

sethrj: I guess I should clarify: on device in the main code they're always LDG, but they might not be using immutable device memory in some of the tests.

src/corecel/cont/Span.hh
```diff
@@ -237,11 +237,10 @@ CELER_FUNCTION Span<const typename T::value_type> make_span(T const& cont)
 
 //---------------------------------------------------------------------------//
 //! Construct an array from a fixed-size span
-template<class T, std::size_t N>
-CELER_CONSTEXPR_FUNCTION Array<std::remove_cv_t<T>, N>
-make_array(Span<T, N> const& s)
+template<class T, std::size_t N, class T2 = std::remove_cv_t<T>>
+CELER_CONSTEXPR_FUNCTION Array<T2, N> make_array(Span<T, N> const& s)
```
sethrj: Wouldn't it also work without the second (somewhat confusing) template argument to do:
```suggestion
template<class T, std::size_t N>
CELER_CONSTEXPR_FUNCTION auto make_array(Span<T, N> const& s)
```
and leave the return type in place?

src/corecel/cont/Span.hh
```diff
@@ -251,10 +250,17 @@ make_array(Span<T, N> const& s)
...
+         class T2 = std::remove_cv_t<typename LdgValue<T>::value_type>>
+CELER_CONSTEXPR_FUNCTION Array<T2, N> make_array(LdgSpan<T, N> const& s)
 {
-    return make_array<typename LdgValue<T>::value_type, N>(s);
+    Array<T2, N> result{};
+    for (std::size_t i = 0; i < N; ++i)
+    {
+        result[i] = s[i];
+    }
+    return result;
```
sethrj, line 266: We need to document why the code is being duplicated. Also is it possible there's an issue in the `template<class U> Span(Span<U>)` constructor?

src/corecel/math/ArrayOperators.hh
```diff
@@ -53,9 +55,9 @@ namespace celeritas
         return (result TOKEN## = y);                                       \
     }                                                                      \
                                                                            \
-    template<class T, size_type N>                                         \
+    template<class T, size_type N, class T2 = std::remove_cv_t<T>>         \
     inline CELER_FUNCTION Array<T, N> operator TOKEN(Array<T, N> const& x, \
-                                                     T const& y)           \
+                                                     T2 const& y)          \
```
sethrj, line 60: This is just an implementation detail I think it's fine, but where did this show up? (What was the error that required this?)

src/orange/surf/GeneralQuadric.hh
```diff
@@ -41,9 +41,9 @@ class GeneralQuadric
     //@{
     //! Type aliases
     using Intersections = Array<real_type, 2>;
-    using StorageSpan = Span<const real_type, 10>;
+    using StorageSpan = LdgSpan<const real_type, 10>;
     using Storage = StorageSpan;  // DEPRECATED
-    using SpanConstReal3 = Span<const real_type, 3>;
+    using SpanConstReal3 = LdgSpan<const real_type, 3>;
```
sethrj: Ah, this could be a problem! This is the return value from *local* data (registers/stack) that's already been loaded into the `GeneralQuadratic` object. Only the `StorageSpan` should be changed here. (I don't think it's *used* on device but it would cause a runtime error I think.)

src/orange/surf/SimpleQuadric.hh
```diff
@@ -42,9 +42,9 @@ class SimpleQuadric
     //@{
     //! \name Type aliases
     using Intersections = Array<real_type, 2>;
-    using StorageSpan = Span<const real_type, 7>;
+    using StorageSpan = LdgSpan<const real_type, 7>;
     using Storage = StorageSpan;  // DEPRECATED
-    using SpanConstReal3 = Span<const real_type, 3>;
+    using SpanConstReal3 = LdgSpan<const real_type, 3>;
```
sethrj:
```suggestion
    using SpanConstReal3 = Span<const real_type, 3>;
```

sethrj (changes requested):

src/corecel/math/ArrayOperators.hh
esseivaju: it showed up [here](https://github.com/celeritas-project/celeritas/blob/88237d15dbd170a7cf4a7fcd0939bfd985d29e9e/src/orange/surf/detail/SurfaceTransformer.cc#L195-L196) when make_array was returning an `array<T const>` if passed a `LdgSpan<T const>`. Since we make a copy of each element and I changed make_array in this PR to remove the constness of T this is no longer needed. Keeping this in might still make sense, `x` and `y` don't require the same const qualifiers for `T`.

src/corecel/cont/Span.hh
esseivaju: that constructor looks correct to me. If we take the scenarios of constructing a `Span<T const>` from `LdgSpan<T const` which is what we were doing before this PR: `data()` casts a `LdgIterator<T consts>` to `T const*` which is used to initialize the pointer of the new span which is the same type as the cast expression.

sethrj: OK as long as it's not accidentally `reinterpret_cast`ing ;)

src/orange/surf/GeneralQuadric.hh
esseivaju: Right, I don't know if it would be a runtime error or if the compiler figures out that __ldg is applied to a register variable and ignores it but it doesn't make sense to use LdgSpan here (unless we ever allocate a `GeneralQuadratic` in GMem?).

esseivaju: Thinking about it more, the compiler can't always find out which memory space the pointer belongs to. I don't think we should even change `StorageSpan`; `GeneralQuadric::data` returns a `StorageSpan` which would also result in a runtime error since the data is already loaded in registers. 
Similarly, the constructor `GeneralQuadric(Storage);` would restrict constructing a surface from a Span pointing to global memory data only. We can change the constructor to accept either `Span` or `LdgSpan` and users should wrap the `Span` returned by `data`, `first`, ... if they know the object resides in GMem.

esseivaju:
@sethrj This significantly increases the number of nc load instructions. The testem3+field+msc propagate kernel went from 0.56% to 2.37% of the total number of instructions executed being `__ldg` load. However, the timings between develop and this branch are identical.

sethrj:
Interesting. Is the register/local usage smaller? Is it basically allowing more global fetches because it thinks they'll be cached and reduce register pressure?

esseivaju:
There's no difference in local memory/register usage in any kernel. I suppose the change is too small to have a measurable performance impact.

sethrj:
Weird. Thanks for trying though.

sethrj:
Confirmed no change in performance or kernel stats on frontier either :(

sethrj (approved):
Even though the performance improvement isn't measurable I can't imagine this doing any harm (other than slowing down the debug build, which is the main reason I was using the raw data accessors there)...

src/corecel/math/ArrayOperators.hh
sethrj: Shouldn't this return `Array<T2, N>`?  Should we prohibit `Array<T const>`?

sethrj (commented):
Also disappointingly, even though `LdgIterator` is `constexpr __attribute__(forceinline)` on clang, its construction and pointer copy doesn't get optimized out, leading to a 10% slowdown when optimizations are off (as measured by test/celeritas_geo_Geometry on a debug build on my laptop).

sethrj:
Also do you think the code is more readable by reverting the local surface visitor changes needed to propagate the `Span` type, and just add `LdgValue`/`LdgIterator` explicitly?

```diff
From dc79db24b43f69ab5437c003b8926e82c8513058 Mon Sep 17 00:00:00 2001
From: Seth R Johnson <johnsonsr@ornl.gov>
Date: Wed, 22 Nov 2023 12:20:43 -0500
Subject: [PATCH] Explicitly add LdgIterator

---
 src/orange/surf/LocalSurfaceVisitor.hh   | 40 ++++++++++++++++++------
 src/orange/transform/TransformVisitor.hh |  5 ++-
 2 files changed, 35 insertions(+), 10 deletions(-)

diff --git a/src/orange/surf/LocalSurfaceVisitor.hh b/src/orange/surf/LocalSurfaceVisitor.hh
index 088131f7d..9285f9057 100644
--- a/src/orange/surf/LocalSurfaceVisitor.hh
+++ b/src/orange/surf/LocalSurfaceVisitor.hh
@@ -28,6 +28,10 @@ namespace celeritas
     [&pos](auto const& s) { return s.calc_sense(pos); },
     surface_id);
  \endcode
+ *
+ * Since all the params data is stored as "const reference", we can use
+ * LdgIterator (i.e. a wrapper around \c __ldg ) to read indices and construct
+ * the surfaces.
  */
 class LocalSurfaceVisitor
 {
@@ -74,6 +78,11 @@ class LocalSurfaceVisitor
     static inline CELER_FUNCTION T get_item(Items<T> const& items,
                                             ItemRange<T> const& range,
                                             ItemId<U> item);
+
+    // Get a pointer to the item at the given item index.
+    template<class T>
+    static inline CELER_FUNCTION T const*
+    get_ptr(Items<T> const& items, ItemId<T> item);
 };
 
 //---------------------------------------------------------------------------//
@@ -131,21 +140,18 @@ LocalSurfaceVisitor::operator()(F&& func, LocalSurfaceId id)
 template<class T>
 CELER_FUNCTION T LocalSurfaceVisitor::make_surface(LocalSurfaceId id) const
 {
-    using ItemIdT = typename decltype(params_.reals)::ItemIdT;
-    ItemIdT offset
+    OpaqueId<real_type> offset
         = this->get_item(params_.real_ids, surfaces_.data_offsets, id);
     constexpr size_type size{T::StorageSpan::extent};
     CELER_ASSERT(offset + size <= params_.reals.size());
-    // Construct a range used to index into collection using the appropriate
-    // operator[] overload then convert to a statically sized span using the
-    // first() member template function as required by surface ctor
-    return T{params_.reals[{offset, offset + size}].template first<size>()};
+
+    real_type const* data = this->get_ptr(params_.reals, offset);
+    return T{Span<LdgValue<real_type const>, size>{data, size}};
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Get a pointer to the item offset using the given range in a given
- * collection.
+ * Get a pointer to the suboffset in the given range in a collection.
  */
 template<class T, class U>
 CELER_FUNCTION T LocalSurfaceVisitor::get_item(Items<T> const& items,
@@ -154,7 +160,23 @@ CELER_FUNCTION T LocalSurfaceVisitor::get_item(Items<T> const& items,
 {
     CELER_EXPECT(*offsets.end() <= items.size());
     CELER_EXPECT(item < offsets.size());
-    return items[offsets][item.unchecked_get()];
+
+    T const* ptr = LocalSurfaceVisitor::get_ptr(items, *offsets.begin());
+    return *LdgIterator{ptr + item.unchecked_get()};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a pointer to the item at the given item index.
+ */
+template<class T>
+CELER_FUNCTION T const*
+LocalSurfaceVisitor::get_ptr(Items<T> const& items, ItemId<T> item)
+{
+    CELER_EXPECT(item < items.size());
+
+    T const* ptr = items[AllItems<T>{}].data();
+    return ptr + item.unchecked_get();
 }
 
 //---------------------------------------------------------------------------//
diff --git a/src/orange/transform/TransformVisitor.hh b/src/orange/transform/TransformVisitor.hh
index f5f9f199e..70f4d1d5c 100644
--- a/src/orange/transform/TransformVisitor.hh
+++ b/src/orange/transform/TransformVisitor.hh
@@ -124,7 +124,10 @@ TransformVisitor::make_transform(OpaqueId<real_type> data_offset) const
     constexpr size_type size{T::StorageSpan::extent};
     CELER_ASSERT(data_offset + size <= reals_.size());
 
-    return T{reals_[{data_offset, data_offset + size}]};
+    real_type const* data = reals_[AllItems<real_type>{}].data();
+    return T{Span<LdgValue<real_type const>, size>{
+        data + data_offset.unchecked_get(),
+        data + data_offset.unchecked_get() + size}};
 }
 
 //---------------------------------------------------------------------------//
-- 
2.42.0
```

esseivaju:
I think I like using the `Collection::operator[]` more since all the indexing logic is already there and it knows if the data storage is in device global memory and `__ldg` can be used. Using `LdgValue`/`LdgIterator` explicitly would break if we ever change the collection. The only thing I didn't like before was having to use `.template first<size>()` to cast from a dynamic extent span to a static extent, maybe I missed a more ergonomic way.

src/corecel/math/ArrayOperators.hh
esseivaju: It is already indirectly prohibited because the operations defined by `CELER_DEFINE_ARRAY_ASSIGN` only support non-const

esseivaju:
> Also disappointingly, even though `LdgIterator` is `constexpr __attribute__(forceinline)` on clang, its construction and pointer copy doesn't get optimized out, leading to a 10% slowdown when optimizations are off (as measured by test/celeritas_geo_Geometry on a debug build on my laptop).

Mmh, yeah I wouldn't care too much about debug performance but If doesn't help optimized builds either then I'm not sure if it's worth merging.

sethrj:
@esseivaju Have you seen this? I'm getting it when building through SCALE using CUDA 12.3.52:
```
/home/vmuser/build/rnsd/scale/external/celeritas/src/orange/transform/TransformVisitor.hh(127): error: more than one operator "[]" matches these operands:
            function "celeritas::Collection<T, W, M, I>::operator[](celeritas::Collection<T, W, M, I>::ItemRangeT) const [with T=celeritas::real_type, W=celeritas::Ownership::const_reference, M=celeritas::MemSpace::device, I=celeritas::OpaqueId<celeritas::real_type, celeritas::size_type>]" (declared at line 301 of /home/vmuser/build/rnsd/scale/external/celeritas/src/corecel/data/Collection.hh)
            function "celeritas::Collection<T, W, M, I>::operator[](celeritas::Collection<T, W, M, I>::AllItemsT) const [with T=celeritas::real_type, W=celeritas::Ownership::const_reference, M=celeritas::MemSpace::device, I=celeritas::OpaqueId<celeritas::real_type, celeritas::size_type>]" (declared at line 305 of /home/vmuser/build/rnsd/scale/external/celeritas/src/corecel/data/Collection.hh)
            operand types are: const celeritas::TransformVisitor::Reals [ {...} ]
      return T{reals_[{data_offset, data_offset + size}]};
                     ^
```

esseivaju:
I haven't seen that error before. Why would the AllItemsT overload even be considered? It's just an empty struct.

sethrj:
No idea. Sounds like a compiler bug. I'm going to push a fix for it anyway, since it breaks the downstream CI ☹️

PR #1020, opened 2023-11-18
Add GitHub actions to CI

sethrj:
Since our CI is down after almost a week, we've got to get some testing capability back online. This restores all the original build types and builds them through GitHub's runners using the original CI images.

Because the runners don't have GPUs, this builds the code and runs tests with `CELER_DISABLE_DEVICE`.
-  I added support for *runtime* skipping of device tests in googletest: if the device is unavailable, any test with `[Dd]evice` in the name gets skipped.
- The `app` testing checks for the environment variable at configure time, and it disables the GPU tests by default if found.
- To support runtime device-disabling for the hit manager (which uses pinned memory for the hits even in host-only mode) I have the `PinnedAllocator` call `Device::num_devices` to select between native (non-pinned) and device API mallocs.

The build names have been changed:
| Image | Old | New |
| ----- | --- | --- |
| ci-centos-rocm | asan      | asan-orange |
| ci-centos-rocm | minitest  | debug-orange |
| ci-centos-rocm | float     | debug-orange-float |
| ci-centos-rocm | geant4    | reldeb-geant4 |
| ci-centos-rocm | ndebug    | ndebug-orange |
| ci-ubuntu-cuda | full-novg        | debug-orange |
| ci-ubuntu-cuda | full-novg-ndebug | ndebug-orange |
| ci-ubuntu-cuda | vecgeom-reldeb   | reldeb-vecgeom |
| ci-ubuntu-cuda | vecgeom-tests    | debug-vecgeom |
| ci-ubuntu-cuda | vecgeom-demos    | ndebug-vecgeom |

I've also adjusted the cmake presets so that "ndebug" prefers static libraries. This doesn't work with ROOT (PCM loading looks only in bin, not lib) so I added cmake code to check for that. Thinking it might've been an omission in the scoped error handler, I updated the exception initalizer to try to load the PCM file for static builds.

TO DO in subsequent PRs:
- Fix the issues with examples
  * asan build needs `-fsanitize=address` in downstream example
  * rocm+ndebug (static libs) fails to propagate the HIP library linkage
  * SimpleOffload::Build crashes because MTLogger is called before run manager completes initialization
  * VecGeom build on Ubuntu requires `-Wl,--no-as-needed` which we provide to Celeritas via `CMAKE_{SHARED|EXE}_LINKER_FLAGS`: maybe @pcanal 's RDC code needs to propagate these? Should this be vecgeom's responsibility?
- Use a shallow clone and `git describe` or some better way of fetching tags
- Improve docker images
  * Add ccache to the build image and use it
  * Avoid custom user ID; GitHub requires root
  * Install all supported geant4 versions for a small build matrix
  * Fix entry point so that we don't have to manually source `/etc/profile` and re-export variables
- Add separate build matrix for geant4 versions (to test minimum and maximum supported)
- Add nightly minimal build matrix for various compilers on raw hardware (earliest GCC through latest, multiple clang versions)
- Add automatic clang formatting and/or fix?
- Add documentation


cmake/CeleritasAddTest.cmake
```diff
@@ -159,13 +159,16 @@ include_guard()
...
-  set(CELERITASTEST_NP_DEFAULT "1" CACHE INTERNAL
-    "Default number of processes to use in CeleritasAddTest")
+  list(APPEND _procs 2)
+  if(MPIEXEC_MAX_NUMPROCS GREATER 2)
+    list(APPEND _procs ${MPIEXEC_MAX_NUMPROCS})
+  endif()
 endif()
+set(CELERITASTEST_NP_DEFAULT "${_procs}" CACHE INTERNAL
+  "Default number of processes to use in CeleritasAddTest")
+set(_procs)
```
sethrj, line 171: This change is necessary because the CI runners have only 2 cores.

app/CMakeLists.txt
```diff
@@ -28,6 +28,21 @@ if(CELERITAS_BUILD_TESTS)
     set(CELER_PROCESSORS)
   endif()
   set(CELER_APP_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/data")
+
+  # Disable device tests if no CUDA/HIP or if environment is set
+  if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
+    if("$ENV{CELER_DISABLE_DEVICE}")
+      message(WARNING
+        "Disabling GPU testing because CELER_DISABLE_DEVICE is set"
+      )
```
sethrj, line 37: This warns every time because the value is updated at configure time and could change from one configure to the next.

scripts/ci/run-ci.sh
```diff
@@ -30,24 +30,9 @@ cmake --build --preset=${CMAKE_PRESET}
...
-    CTEST_ARGS="-L app"
-    ;;
-  valgrind )
-    CTEST_TOOL="MemCheck"
-    CTEST_ARGS="-LE nomemcheck"
-    ;;
-  * )
-    ;;
-esac
-
```
sethrj, line 47: We haven't had a valgrind job on jenkins for a while.

CMakePresets.json
```diff
@@ -4,17 +4,14 @@
...
-      "displayName": "Automatic options (debug with tests)",
+      "displayName": "Automatic options",
       "description": "Dependencies are enabled based on environment probing",
       "binaryDir": "${sourceDir}/build-${presetName}",
       "generator": "Ninja",
       "cacheVariables": {
-        "BUILD_SHARED_LIBS":    {"type": "BOOL",   "value": "ON"},
         "Celeritas_GIT_DESCRIBE": "",
         "CELERITAS_BUILD_TESTS":{"type": "BOOL",   "value": "ON"},
-        "CELERITAS_DEBUG":      {"type": "BOOL",   "value": "ON"},
```
sethrj, line 15: No longer default to debug, because some people might want to use production by default. Now daughter configurations explicitly have to inherit from ndebug/reldeb/debug .

CMakePresets.json
```diff
@@ -85,6 +82,7 @@
       "hidden": true,
       "description": "Disable debug mode",
       "cacheVariables": {
+        "BUILD_SHARED_LIBS":{"type": "BOOL",   "value": "OFF"},
```
sethrj, line 85: Building without shared libraries improves app load times but is incompatible with ROOT.

scripts/cmake-presets/ci-centos-rocm.json
```diff
@@ -34,59 +33,44 @@
...
-      "description": "Build debug with HIP/ROCM for AMD",
-      "inherits": "base"
-    },
-    {
-      "name": "minitest",
-      "description": "Build without any dependencies but JSON/Googletest",
-      "inherits": "base",
+      "name": "debug-orange",
+      "description": "Build with minimal reasonable dependencies",
+      "inherits": [".debug", "base"],
```
sethrj, line 47: Since debug + HIP doesn't work (too many registers required) I'm making this one the 'minimal' build.

src/celeritas/ext/ScopedRootErrorHandler.cc
```diff
@@ -64,9 +66,12 @@ void RootErrorHandler(Int_t rootlevel,
         auto log_msg = ::celeritas::world_logger()({"ROOT", 0}, level);
         if (location)
         {
-            log_msg << color_code('x') << location << color_code(' ') << ": ";
+            log_msg << color_code('x') << location << color_code(' ');
+        }
+        if (msg && msg[0] != '\0')
+        {
+            log_msg << ": " << msg;
```
sethrj, line 73: The `LoadPCM` failure message is empty, making the trailing colon confusing.

src/celeritas/ext/ScopedRootErrorHandler.cc
```diff
@@ -86,10 +91,13 @@ void ScopedRootErrorHandler::disable_signal_handler()
  * Install the Celeritas ROOT error handler.
  */
 ScopedRootErrorHandler::ScopedRootErrorHandler()
-    : previous_(SetErrorHandler(RootErrorHandler))
-    , prev_errored_{g_has_root_errored_}
 {
-    ScopedRootErrorHandler::disable_signal_handler();
+    if (RootFileManager::use_root())
```
sethrj, line 95: The call to `use_root` also triggers the dictionary initialization if needed.

test/CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ configure_file(celeritas_test_config.h.in celeritas_test_config.h @ONLY)
 celeritas_add_library(testcel_harness
   Test.cc
   testdetail/NonMasterResultPrinter.cc
-  testdetail/ParallelHandler.cc
```
sethrj, line 19: I inlined this while adding the device disabler.

test/orange/univ/SimpleUnitTracker.test.cc
```diff
@@ -760,7 +761,7 @@ TEST_F(FieldLayersTest, TEST_IF_CELERITAS_DOUBLE(heuristic_init))
         EXPECT_SOFT_EQ(0, result.failed);
     }
 
-    if (CELER_USE_DEVICE)
+    if (celeritas::device())
```
sethrj, line 764: This is slightly dangerous because a low-level bug that keeps the device from initializing means that this may pass.

pcanal:
1. > and runs tests with CELER_DISABLE_DEVICE.
2. > I added support for runtime skipping of device tests in googletest: if the device is unavailable, any test with [Dd]evice in the name gets skipped.

Doesn't 2 make 1. unnecessary?

sethrj:
@pcanal We definitely need it to disable the tests. Also `CELER_DISABLE_DEVICE` completely prevents hip/CUDA API calls, and I *think* the nvidia docker image has a trampoline function for its `libcuda` that tries to load shared libraries from the machine running `nvidia-docker`, which we don't have—so even calling `cudaGetDeviceCount` might fail.

It'd be good to document/check though so I'll try it out.

src/celeritas/ext/ScopedRootErrorHandler.cc
pcanal: The change it self is fine and make sense.  That there an error message with no message is odd/surprising :(. (At least for `LoadPCM` I don't see how it happened)

scripts/cmake-presets/ci-centos-rocm.json
```diff
@@ -34,59 +33,44 @@
...
     {
-      "name": "debug",
-      "description": "Build debug with HIP/ROCM for AMD",
-      "inherits": "base"
-    },
-    {
-      "name": "minitest",
-      "description": "Build without any dependencies but JSON/Googletest",
-      "inherits": "base",
+      "name": "debug-orange",
```
pcanal:
```suggestion
      "_comment1": "Since debug + HIP doesn't work (too many registers required) I'm making this one the 'minimal' build.",
```

CMakePresets.json
pcanal: I am confused by this statement.  This section of settings does not seems to mention while seemingly using the feature that this said not to work with ROOT.

scripts/ci/run-ci.sh
pcanal: Wouldn't this be an occasion to reintroduce it or it is known not to work?

sethrj:
@pcanal I [confirmed the failure mode](https://github.com/celeritas-project/celeritas/actions/runs/6932317862/job/18855973178?pr=1020) when omitting `CELER_DISABLE_DEVICE`:
```
terminate called after throwing an instance of 'celeritas::RuntimeError'
  what():  /__w/celeritas/celeritas/src/corecel/sys/Device.cc:107:
celeritas: CUDA error: CUDA driver version is insufficient for CUDA runtime version
```

pcanal:
> We definitely need it to disable the tests ... I [confirmed the failure mode](https://github.com/celeritas-project/celeritas/actions/runs/6932317862/job/18855973178?pr=1020) when omitting CELER_DISABLE_DEVICE:

So, how does the ` runtime skipping of device tests in googletest` help?

scripts/cmake-presets/ci-centos-rocm.json
sethrj: I've renamed it and updated the description so it's more clear. I'm not sure if CMakeUserPresets allows arbitrary fields since it's validated against a schema.

sethrj:
The `CELER_DISABLE_DEVICE` ensures `celeritas::device()` is false. However, the device tests will *fail* if it's false. Therefore we have to pass the macro to disable the device during setup, *and* skip the test if the device is unavailable.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,121 @@
...
+    paths-ignore:
+      - '**.rst'
+      - '**.md'
+      - 'scripts/dev'
+  merge_group:
+    types: [checks_requested]
+
+concurrency:
+  group: ${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
+  cancel-in-progress: true
```
pcanal, line 21: This settings means that for all workflow group, a new items will cancel the previous on going or queue runs.  For pull-request this is clearly what we want.  For push in the repository it is more debatable; having the setting on for those means that technically not all pushes are tested/processed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,121 @@
...
+      # NOTE: depth must be enough to include the previous tag
+      - name: Check out Celeritas
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 255
+          fetch-tags: true
+      - name: Configure Celeritas
+        run: |
+          git config --global --add safe.directory ${PWD}
+          ln -fs scripts/cmake-presets/ci-${{matrix.image}}.json CMakeUserPresets.json
```
pcanal, line 95: This means that we don't (yet) support Windows build :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,121 @@
...
+        run: |
+          cmake --build --preset=${CMAKE_PRESET}
+      - name: Test Celeritas
+        working-directory: build
+        run: |
+          ctest --timeout 180 --output-on-failure \
+            --test-output-size-passed=65536 --test-output-size-failed=1048576
+      - name: Install Celeritas
+        run: |
+          . /etc/profile
```
pcanal: Why is this step setting (again?) the environment?

.github/workflows/ci.yml
sethrj: Good point; how can we set that up to apply only to pull requests?

.github/workflows/ci.yml
sethrj: Our CI has never supported windows builds and it's running inside a docker container that's linux only. 😝 

This PR is really just to get us back in business to most of the capability we had before. I'd like to extend testing (see comment in the PR description) and if you want to add a windows build I'm a ll for it.

.github/workflows/ci.yml
sethrj: `/etc/profile` doesn't get loaded by the github runner because it isn't using the entrypoint for some reason. This is necessary to set up the spack envrionment so we can get `/opt/view/bin` (and whatever else) into the path.

After the smoke clears from the R&D evaluation I'm planning to redo the CI images so this hopefully isn't necessary.

.github/workflows/ci.yml
pcanal: We  will need two (or 3) yml files/workflow and use '[reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)`

Something like:
```
name: ci_push

on:
  push:
    branches:
      - develop
      - backports/**

  jobs:
  call-workflow:
    uses: .github/workflows/ci.yml
```
and `ci.yml` will replace the push section with: ```workflow_call:``` section.

It might be clean to also extract the `pull_request` section is a similar manner,

.github/workflows/ci.yml
pcanal: The weird looking part is that it is call here but not in other runs (in particular `- name: Build Celeritas` and `Test Celeritas`)

sethrj: Ahhh. OK it might no longer be necessary, but the build/install hooks are using whatever CMake cached from its initial run, rather than depending on the environment. I'll try deleting it.

sethrj: Oh actually it definitely can be deleted. It's not being used for the install.

.github/workflows/ci.yml
sethrj: Actually are you sure it will cancel the runs on `develop`? I think it'll cancel only other jobs in the same group, and if it's not a pull request, then it'll use the unique run number.
```
  group: ${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
```

scripts/ci/run-ci.sh
sethrj: No this is not the occasion; that could be a separate PR if it's needed. I don't see any advantage since we have the clang address sanitizer.

.github/workflows/ci.yml
pcanal: I am not sure, this might indeed be sufficient if I read the doc for `github.run_number` correctly.

amandalund (approved):
Good on my end, thanks for getting this set up @sethrj!

.github/workflows/ci.yml
pcanal: On the other hand, using a different workflow name would make the display at: https://github.com/celeritas-project/celeritas/actions
clearer and easier to find the result of the `check-on-push` results by separating them from the `pull request`.

sethrj: Yeah, I've got a draft of this in #1021

PR #1021, opened 2023-11-20
Fix unchecked merge train wreck

sethrj:
I didn't set up the merge check correctly after #1020 so a broken merge (the first one I merged unfortunately, #1012) got into develop. 

This also moves the building into a separate workflow step and adds a catch-all so that we can more simply require

### DIFF
--- a/.github/workflows/build-full.yml
+++ b/.github/workflows/build-full.yml
@@ -1,7 +1,4 @@
-name: build-gpu
-# TODO: this currently includes non-GPU builds as well
-# and won't actually run on device
-
+name: full-build
 on:
   workflow_dispatch:
   workflow_call:
@@ -11,8 +8,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  containers:
-    name: GPU-enabled builds
+  # TODO: this currently includes non-GPU builds as well
+  # and won't actually run on device
+  gpu:
+    name: gpu
     strategy:
       matrix:
         special: [null]
--- a/.github/workflows/build-full.yml
+++ b/.github/workflows/build-full.yml
@@ -1,4 +1,4 @@
-name: full-build
+name: build-full
 on:
   workflow_dispatch:
   workflow_call:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,7 @@
+name: nightly
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  #schedule:
+  #  - cron:  '0 1 * * *'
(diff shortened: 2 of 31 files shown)


.github/workflows/build-full.yml
```diff
@@ -1,28 +1,17 @@
...
 concurrency:
   group: ${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: CI
+  # TODO: this currently includes non-GPU builds as well
+  # and won't actually run on device
+  gpu:
```
pcanal, line 13: Why name it gpu if it will also run on cpu?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,35 @@
+name: pr
+
+on:
+  push:
+    branches:
+      - develop
+      - backports/**
```
pcanal: Did you mean to keep that here rather than in its own file?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,7 @@
+name: nightly
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  #schedule:
+  #  - cron:  '0 1 * * *'
```
pcanal: You may want to remove this file for now.  It leads to:
```
Error: .github#L1
No jobs defined in `jobs`
```

pcanal (changes requested):
The file nightly.yml as is lead to "check failures"

.github/workflows/build-full.yml
sethrj: Once I have time I plan to only use these images for GPU builds, and have separate ones for cpu only. I think. Either way, they're both using images designed for nvidia-docker and rocm-docker.

pcanal: > Once I have time I plan to only use these images for GPU builds

Would it make sense to change the name when that happens?

sethrj: Right now they are GPU builds. We're just not running them on GPU.

pcanal (approved):
Assuming the test passes, this looks good to go.

sethrj:
@pcanal Have any suggestions (in follow-up) for [this error](https://github.com/celeritas-project/celeritas/actions/runs/6936697810)?
```
Canceling since a deadlock for concurrency group 'refs/heads/develop-1-push' was detected between 'top level workflow' and 'build'
```

pcanal:
not yet :(

pcanal:
I would remove the concurrency group in the inner workflow.  I suspect this is related to: https://github.com/github/vscode-github-actions/issues/135  even-though the group name ought to be different (different workflow name).

PR #1022, opened 2023-11-21
Fix CI for "push" events

sethrj:
This fixes the duplicate values for `concurrency` (thanks @pcanal ) and improves the titles that show up in the [actions tab](https://github.com/celeritas-project/celeritas/actions), although it's not possible to get the "title" of a commit message. See [the run push actions](https://github.com/sethrj/celeritas/actions/).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -10,14 +10,19 @@ on:
     branches:
       - develop
       - backports/**
+      - fix-push-ci
 
 concurrency:
   group: push-${{github.ref}}-${{github.run_number}}-${{github.workflow}}
   cancel-in-progress: true
 
 jobs:
-  build:
-    uses: ./.github/workflows/build-full.yml
-  # TODO: build and push documentation
+  all:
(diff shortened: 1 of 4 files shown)


pcanal (changes requested):
Some of the debugging change in pr.yml still need to be reverted.

sethrj:
Sorry should've marked it as draft once we started diddling around again. Try again?

.github/workflows/push.yml
```diff
@@ -1,4 +1,8 @@
 name: push
+# run-name: >-
+#   ${{github.ref_name}}:
+#   ${{github.event.head_commit.message}}
+#   (${{github.event.pusher.name}})
```
pcanal, line 6:
```suggestion
# run-name: >-
#   ${{github.ref_name}}:
#  Would like to use the first line of the message but the line below brings the whole commit message
#   ${{github.event.head_commit.message}}
#   (${{github.event.pusher.name}})
```

pcanal (approved):

PR #1024, opened 2023-11-21
Use default field driver options in celer-g4 test

amandalund:
I looks like when we added the celer-g4 test harness in #1007 the input field driver options were changed, and that's why the celer-g4 GPU test is failing (sometimes, since it's not yet reproducible ;)). We should probably make sure our field propagator works as expected for reasonable choices of these values, but for the test I think it's fine to just use the defaults.

### DIFF
--- a/app/celer-g4/test-harness.py
+++ b/app/celer-g4/test-harness.py
@@ -82,10 +82,10 @@ inp = {
      1.0
     ],
     "field_options": {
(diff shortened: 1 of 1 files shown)


sethrj:
Whoops, you're quite right, I forgot about that in my haste to try to make it quicker. Thanks!

sethrj (approved):

PR #1025, opened 2023-11-21
CI: Add sha to merge name

pcanal:


sethrj (commented):
We agreed to stop using merge queues, and I don't think this adds anything useful. The commit isn't going to be easily readable (one or two characters of the hash might show up), and it won't make an active link. And you can see the commit sha as a link if you just click on the merge train action anyway.

pcanal:
>  and it won't make an active link. 

This is the part I wanted to clarify/check.  If it does not, it is indeed useless.

PR #1013, opened 2023-11-12
Use Celeritas error and output handling in celer-g4 app

sethrj:
I got tired of trying to find error messages among the signal handler–added backtrace output. This removes all the `ScopedGeantException` code from `celer-g4`, adds log and exception wrapping, and adds diagnostic output of exceptions, and memory and environment diagnostics.

Checks off 3 of the 4 boxes in #670.

PR #1026, opened 2023-11-22
Add multithread exception support and logging to Geant4 app

sethrj:
In #1013 I didn't realize that allowing exceptions to propagate out of worker threads would simply call `std::terminate` due to C++ exception handling rules. I also neglected to add exception handlers and logging redirection to the worker thread.

This fixes both of those implementations:
- Restore wrapping of `CELER_TRY_HANDLE(..., call_g4exception)` around potentially-throwing Celeritas code
- Add a `MultiExceptionHandler` instance to the `GeantDiagnostics` class
- Construct a thread-local `ExceptionHandler` class (which implements the Geant4 `G4Exception` notification interface) given that instance
- Store the given `G4Exception` and call `G4RunManager::AbortRun` in that exception handler
- When finalizing `GeantDiagnostics` on the manager thread, check for exceptions, log all but one (depending on how many threads died), and throw the last one, which will be caught by `celer-g4.cc`.

### DIFF
--- a/app/celer-g4/GeantDiagnostics.cc
+++ b/app/celer-g4/GeantDiagnostics.cc
@@ -76,7 +76,7 @@ GeantDiagnostics::GeantDiagnostics(SharedParams const& params)
         output_reg->insert(std::make_shared<BuildOutput>());
     }
 
-    // Create shared exception handler (note: make_shared acts screwy here...)
+    // Create shared exception handler
     meh_ = std::make_shared<MultiExceptionHandler>();
 
     CELER_ENSURE(*this);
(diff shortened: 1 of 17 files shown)


app/celer-g4/GeantDiagnostics.cc
```diff
@@ -76,6 +76,9 @@ GeantDiagnostics::GeantDiagnostics(SharedParams const& params)
         output_reg->insert(std::make_shared<BuildOutput>());
     }
 
+    // Create shared exception handler (note: make_shared acts screwy here...)
```
sethrj:
```suggestion
```

This was because the accessor name was `MultiExceptionHandler` to follow geant4 conventions, so it shadowed the class name 🤪

amandalund (approved):
Thanks @sethrj, looks good to me!

PR #1029, opened 2023-11-22
Simplify data access for host collections

esseivaju:
In case we have a Collection in host memory space with a type supported by `__ldg`, use a regular `Span` to read data instead of relying on `LdgIterator` doing the correct thing.


sethrj (approved):
Looks good, but the one disadvantage is that this will potentially make it harder to catch programming errors when using CPU-only programming (e.g., `int const* dumb = &native_cref[IntId{0}]`). Since we have CI device builds I think that's not a big problem.

PR #1030, opened 2023-11-26
Update Celeritas CMake defaults and add documentation build

sethrj:
This changes the default CMake options for Celeritas:
- Tests and demos (interactor and raytrace) are off by default so that the build is more suitable for "end users" than developers.
- Debug is also off by default for the same reason; this results in the default build type being "Release".
- The SWIG auto-finding now only looks for 4.1 and higher, so that users with SWIG 4.0 or less won't see an error on their first Celeritas configuration.
- Frontier and Crusher now use the ROCm 5.6 toolchain.

I've also added a GHA workflow for building documentation. Later it will be extended to automatically upload to the github pages site on successful merges to `develop`.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,17 +21,19 @@ jobs:
   build:
     uses: ./.github/workflows/build-full.yml
   doc:
-    if: ${{ contains(github.ref_name, 'release-v')
-           || contains(github.ref_name, 'doc-') }}
+    if: ${{ contains(github.head_ref, 'release-v')
+           || contains(github.head_ref, 'doc-') }}
     uses: ./.github/workflows/doc.yml
 
   # Specifying a dependent job allows us to select a single "requires" check
   all:
+    # Complete if jobs were skipped (but not if any failed or were cancelled)
     if: ${{ always() && !failure() && !cancelled() }}
     needs: [build, doc]
     runs-on: ubuntu-latest
     steps:
     - name: Success
-      run: "true"
+      run: |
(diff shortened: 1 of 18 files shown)


.github/workflows/pr.yml
```diff
@@ -20,13 +20,19 @@ concurrency:
 jobs:
   build:
     uses: ./.github/workflows/build-full.yml
+  doc:
+    if: ${{contains(github.head_ref, 'release-v') || contains(github.head_ref, 'doc-')}}
+    uses: ./.github/workflows/doc.yml
```
pcanal, line 25: Is there anything that can break the doc build?  If so we ought to build it for each PR, it not this is fine.

pcanal (approved):
The PR contains changes that not directly related to the heading (improvement in the documentation) that slightly obscure the salient points (for example it "seems" that the only required change in the source is the one in `Types.hh` but it is not clear).

.github/workflows/pr.yml
sethrj: We could put it at every PR true; I only left it off to save compute power. It's possible to break the build by adding bad rst or really bad doxygen. The goal with the conditional above is "codify" what I've been doing all along: some doc stuff may break with PRs and then we make sure it all works at release time. I think a follow-up could enforce it at every pull request? (But we definitely don't want to upload at every pull.)

pcanal: > I think a follow-up could enforce it at every pull request? 

Sure

> (But we definitely don't want to upload at every pull.)

I completely agree.

PR #1032, opened 2023-11-26
Add preflight github action

sethrj:
This adds a preflight build for testing Celeritas cached on native GitHub hardware before proceeding to the full build. Because it uses ccache, it should only take a few seconds for each push, versus the 12 minutes or so for the cuda/hip builds (at least 2 minutes of which is downloading the docker images).

It uses a release build (to avoid debug symbols that increase cache size) and debug assertions (to catch simple issues during testing).

### DIFF
--- a/.github/workflows/build-local.yml
+++ b/.github/workflows/build-local.yml
@@ -64,6 +64,7 @@ jobs:
         run: |
           mkdir build && cd build
           cmake -GNinja \
+            -DCeleritas_GIT_DESCRIBE=";-pr.${{github.event.pull_request.number}};" \
             -DCELERITAS_BUILD_DEMOS:BOOL=ON \
             -DCELERITAS_BUILD_TESTS:BOOL=ON \
             -DCELERITAS_USE_SWIG=OFF \
--- a/.github/workflows/build-local.yml
+++ b/.github/workflows/build-local.yml
@@ -5,7 +5,7 @@ on:
   workflow_call:
 
 concurrency:
-  group: build-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
+  group: build-local-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
   cancel-in-progress: true
 
 jobs:
@@ -51,9 +51,6 @@ jobs:
           $CXX --version | head -1
       - name: Check out Celeritas
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 255
-          fetch-tags: true
       - name: Cache ccache
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-local.yml
```diff
@@ -0,0 +1,76 @@
+# Build directly on the GitHub runner
+name: build-local
+on:
+  workflow_dispatch:
+  workflow_call:
+
+concurrency:
+  group: build-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
```
pcanal: For the group, I would replace `build-` with `build-local-` or `prebuild-`, to differentiate it from the full build.

.github/workflows/build-local.yml
```diff
@@ -0,0 +1,76 @@
...
+          sudo apt-get -y update
+          sudo apt-get -y install \
+            ccache cmake ninja-build python3 \
+            nlohmann-json3-dev libgtest-dev libhepmc3-dev \
+            ${{matrix.compiler}}
+      - name: Check out Celeritas
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 255
+          fetch-tags: true
```
pcanal: Do we need the tag and the 255 depth for the quick build?  Wouldn't the already downloaded by github repo be sufficient?

.github/workflows/build-local.yml
```diff
@@ -0,0 +1,76 @@
...
+            -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic -Werror -Wno-error=deprecated-declarations" \
+            ..
+      - name: Build all
+        working-directory: build
+        run: |
+          ninja
+      - name: Run tests
+        working-directory: build
+        run: |
+          ctest --parallel 2 --timeout 15 --output-on-failure \
```
pcanal, line 84: Is the number of core given by github known?

pcanal (changes requested):
group name needs review.  Other comments are potential optimization.

.github/workflows/build-local.yml
sethrj: Actually good point, we don't . It was only truly necessary for the "install" step of the full thing.

.github/workflows/build-local.yml
sethrj: I don't think so: I looked it up, and it just had a hard-coded info page about the github runners.

pcanal: It is probably not enough of a gain to worry about but we could also determine the number of core at run-time (https://github.com/orgs/community/discussions/25057)

pcanal (approved):
Looks good (albeit maybe a future improve to increase the parallelism of ctest).  ccache is working as hope for (first build 5 minutes, second identical build a few seconds).

.github/workflows/build-local.yml
sethrj: Actually I forgot I added a script earlier (scripts/ci/host-system-info.cmake) that's meant to pull this number in. We'll just have to enable it.

PR #1033, opened 2023-11-27
Hotfix: CUDA 12.3.52 build error

sethrj:
I think there's a compiler bug because AllItemsT definitely can't be constructed from two values...
```
scale/external/celeritas/src/orange/transform/TransformVisitor.hh(127): error: more than one operator "[]" matches these operands:
            function "celeritas::Collection<T, W, M, I>::operator[](celeritas::Collection<T, W, M, I>::ItemRangeT) const [with T=celeritas::real_type, W=celeritas::Ownership::const_reference, M=celeritas::MemSpace::device, I=celeritas::OpaqueId<celeritas::real_type, celeritas::size_type>]" (declared at line 301 of /home/vmuser/build/rnsd/scale/external/celeritas/src/corecel/data/Collection.hh)
            function "celeritas::Collection<T, W, M, I>::operator[](celeritas::Collection<T, W, M, I>::AllItemsT) const [with T=celeritas::real_type, W=celeritas::Ownership::const_reference, M=celeritas::MemSpace::device, I=celeritas::OpaqueId<celeritas::real_type, celeritas::size_type>]" (declared at line 305 of /home/vmuser/build/rnsd/scale/external/celeritas/src/corecel/data/Collection.hh)
            operand types are: const celeritas::TransformVisitor::Reals [ {...} ]
      return T{reals_[{data_offset, data_offset + size}]};
                     ^
```

This showed up after merging #1018.

### DIFF
--- a/src/orange/surf/LocalSurfaceVisitor.hh
+++ b/src/orange/surf/LocalSurfaceVisitor.hh
@@ -131,15 +131,17 @@ LocalSurfaceVisitor::operator()(F&& func, LocalSurfaceId id)
 template<class T>
 CELER_FUNCTION T LocalSurfaceVisitor::make_surface(LocalSurfaceId id) const
 {
-    using ItemIdT = typename decltype(params_.reals)::ItemIdT;
+    using Reals = decltype(params_.reals);
+    using ItemIdT = typename Reals::ItemIdT;
+    using ItemRangeT = typename Reals::ItemRangeT;
     ItemIdT offset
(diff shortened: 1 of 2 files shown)


esseivaju (approved):
LGTM!Does that only happen with cuda>=12.3?

sethrj:
I'm not sure if a newer version fixes it. It would be nice to have a more full CUDA build matrix so I could tell for sure.

PR #1034, opened 2023-11-28
Write celer-g4 input to output file or screen

sethrj:
This implements a `--dump-default` command line argument for `celer-g4` that's useful for setting up inputs. It also dumps the user input (assuming JSON) to the output file.


amandalund (approved):
Looks good @sethrj!

PR #1036, opened 2023-11-28
Fix Jenkins CI build and ccache use

sethrj:
The Jenkins runner came back online this  morning, and the jenkins file is broken due to #1020 and subsequent github CI changes.

Also, [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) is apparently unable to use cache from another incoming PR or other branch; so I think we need to enable the "fast" builds on develop so that the cache will be ready for the branches.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,8 +2,7 @@
 
 # Title
 
-Should be an imperative statement (title-cased first word,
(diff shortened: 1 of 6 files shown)


pcanal (approved):
LGTM

PR #1040, opened 2023-11-29
Fix GeantSimpleCalo output when Celeritas is enabled

sethrj:
The simple calorimeter SD implementation in #1014 was only tested with Geant4 disabled, so the call to `SharedParams::Initialize` would reset the output and prevent the calo from being printed at the end of the program run.

I also updated the local logger to use a mutex for output to avoid stepping on each other, and I fixed a few other small bugs while trying to get this to work.


amandalund (approved):
Thanks @sethrj! Ugly but working :)

PR #1041, opened 2023-11-29
Update Perlmutter CMake profile and env script

esseivaju:
Fix ndebug configure profile on Perlmutter to build shared library. Also, somehow I'm having issues again with libraries not having the RPATH pointing to the spack lib dir (though it works fine on other machine like Zeus), I'm just setting LD_LIBRARY_PATH for now.

### DIFF
--- a/scripts/cmake-presets/perlmutter.json
+++ b/scripts/cmake-presets/perlmutter.json
@@ -9,7 +9,7 @@
       "binaryDir": "${sourceDir}/build-${presetName}",
       "generator": "Ninja",
       "cacheVariables": {
-        "BUILD_SHARED_LIBS":     {"type": "BOOL", "value": "OFF"},
+        "BUILD_SHARED_LIBS":     {"type": "BOOL", "value": "ON"},
         "CELERITAS_BUILD_DOCS": {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_OpenMP":  {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_Geant4":  {"type": "BOOL", "value": "ON"},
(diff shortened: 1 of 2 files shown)


scripts/cmake-presets/perlmutter.json
```diff
@@ -18,16 +18,15 @@
         "CELERITAS_USE_HIP":  {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_JSON":    {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_MPI":    {"type": "BOOL", "value": "OFF"},
-        "CELERITAS_USE_ROOT":    {"type": "BOOL", "value": "OFF"},
+        "CELERITAS_USE_ROOT":    {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_SWIG":    {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_VecGeom": {"type": "BOOL", "value": "ON"},
         "CMAKE_CXX_FLAGS": "-Wall -Wextra -Wno-psabi -pedantic -pedantic-errors",
-        "CMAKE_CUDA_FLAGS": "-Xcompiler -Wno-psabi",
+        "CMAKE_CUDA_FLAGS": "-lineinfo -Xptxas=-v -Xcompiler -Wno-psabi",
```
sethrj, line 25: I'm guessing `-lineinfo` increases the size of the device code but doesn't negatively affect performance?

sethrj (approved):

scripts/cmake-presets/perlmutter.json
esseivaju: I haven't compared but I wouldn't expect any difference. I can check whenever Perlmutter is back online.

sethrj: Nah not important. Certainly it looked like the GPU performance was fine in your last run 😄

PR #1043, opened 2023-12-01
Update cuda build instructions

esseivaju:
This update the instructions for building dependencies using cuda. They were out of date because the `packages:all:variants` within `spack.yaml` were overriding the user-scope config. 

Also update vecgeom minimum required version to match CMake and fix a few broken links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ $ make && ctest
 
 Celeritas guarantees full compatibility and correctness only on the
 combinations of compilers and dependencies tested under continuous integration:
-- Compiler:
+- Compilers:
     - GCC 8.4, 12.3
     - GCC 11.3 + NVCC 11.8
     - HIP-Clang 10.0, 15.0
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -81,14 +81,14 @@ The full set of configurations is viewable on [the CI web site][jenkins].
 Compatibility fixes that do not cause newer versions to fail are welcome.
 
 [spack]: https://github.com/spack/spack
-[install]: doc/installation.rst
-[jenkins]: https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Celeritas/activity?branch=master
+[install]: doc/main/installation.rst
+[jenkins]: https://cloud.cees.ornl.gov/jenkins-ci/job/celeritas/job/develop
```
sethrj, line 90: Could you also add a link to the github actions? We've now got CI added for several additional compilers.

sethrj (approved):
Thanks for that Julien! Even though 11.0.3 is still the only supported geant version, could you also add that we test several other compilers in the CI?

esseivaju:
Alright, should be good now assuming I didn't miss any tested compiler.

PR #1044, opened 2023-12-02
Support sorting tracks by particle types

esseivaju:
Add an option to sort tracks by particle types. Most of the existing code used to sort on `ActionId` can be reused, I just needed to had a macro to launch a kernel template.

### DIFF
--- a/src/celeritas/Types.cc
+++ b/src/celeritas/Types.cc
@@ -61,6 +61,7 @@ char const* to_cstring(TrackOrder value)
         "sort_along_step_action",
         "sort_step_limit_action",
         "sort_action",
+        "sort_particle_type",
     };
     return to_cstring_impl(value);
 }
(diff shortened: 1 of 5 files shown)


sethrj (approved):
LGTM! I know we've had to avoid using `CELER_LAUNCH_KERNEL` in a couple of places because of templates; is it reasonable to use the new macros with those in this PR to keep everything in sync?

Also, I think we've reached the point where we should start running some performance tests on a variety of scenarios (celer-sim, -g4; HIP, CUDA)...

esseivaju:
> I know we've had to avoid using CELER_LAUNCH_KERNEL in a couple of places because of templates; is it reasonable to use the new macros with those in this PR to keep everything in sync?

There were two places where it applies: 

- In the ActionLauncher, though, we can't directly reuse the macro because the `KernelParamCalculator` is a class data member instead of a local static.
- the `tracks_per_action_kernel`. Here it would be possible to do some refactoring; give a pointer to the correct `ActionId` collection instead of the switch statement inside the kernel. I can open a PR with some of these refactoring changes to simplify the implementation a bit.

> Also, I think we've reached the point where we should start running some performance tests on a variety of scenarios (celer-sim, -g4; HIP, CUDA)...

Yeah, that would be interesting. I still don't have access to AMD gpus.

sethrj:
Did we get you an account on the [UOregon OACISS](https://systems.nic.uoregon.edu/internal-wiki/index.php?title=Category:Servers) systems? They have some. If you have a problem matrix set up to run, I can also launch them on crusher (MI250x).

PR #1045, opened 2023-12-02
Improve JSON I/O for celer-g4/sim apps

sethrj:
- Write entries that are inapplicable as "null" placeholders
- Save version and format to output, check input
- Set default celer-g4 output name to `celeritas.out.json`

### DIFF
--- a/app/celer-sim/RunnerInputIO.json.cc
+++ b/app/celer-sim/RunnerInputIO.json.cc
@@ -52,11 +52,13 @@ namespace app
  */
 void from_json(nlohmann::json const& j, RunnerInput& v)
 {
-#define LDIO_LOAD_OPTION(NAME)          \
-    do                                  \
-    {                                   \
-        if (j.contains(#NAME))          \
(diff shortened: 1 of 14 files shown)


app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -148,70 +139,60 @@ void from_json(nlohmann::json const& j, RunnerInput& v)
  */
 void to_json(nlohmann::json& j, RunnerInput const& v)
 {
+#define LDIO_SAVE(NAME) CELER_JSON_SAVE(j, v, NAME)
+#define LDIO_SAVE_WHEN(NAME, COND) CELER_JSON_SAVE_WHEN(j, v, NAME, COND)
+#define LDIO_SAVE_OPTIONAL(NAME) \
```
amandalund: `OPTIONAL` here but `OPTION` everywhere else?

amandalund (approved):
Nice addition @sethrj!

PR #1046, opened 2023-12-03
Support using VecGeom without VGDML

sethrj:
The default CMS build does not enable VecGeom's GDML reader. When using celer-g4 or the `accel` library, we also don't need to use it since we rely on the in-memory Geant4-vecgeom converter. The main change is to update the ReflFactory to its newer path (requires VecGeom 1.2.3 or higher, but we have a new minimum of 1.2.4), then disable the built-from-gdml support in VecgeomParams.

Because most of the "core" celeritas unit tests require building through the standalone VecGeom ".gdml" input, I tried to support building through the G4GDMLParser. Unfortunately that crashes the code:
- Instantiating the parser instantiates G4GDMLMessenger
- Its constructor adds code to the G4UImanager (`G4UIcommand::G4UIcommandCommonConstructorCode`)
- `G4UImanager::GetUIpointer` has code to check if it's been destroyed and if so return a null pointer which crashes
It crashed in the relevant cases because GeantSetup had been called which destroys the run manager. I've added error checking code to the relevant places to keep this crash from happening again.  Digging into this also lets me remove a line of hacky code from the `load_geant_geometry` call.

So instead, since this is an unusual case, I've prevented building the unit tests when VGDML is unavailable. **However**, you can still check that celeritas is working correctly through the Geant4 converter by defining `BUILD_TESTING=ON` and running the `celer-g4` test.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,21 +332,21 @@ if(CELERITAS_USE_VecGeom)
 
   if(CELERITAS_USE_CUDA AND NOT VecGeom_CUDA_FOUND)
     celeritas_error_incompatible_option(
-      "VecGeom installation at ${VECGEOM_INSTALL_DIR} is not CUDA-enabled"
+      "VecGeom installation at \"${VecGeom_DIR}\" is not CUDA-enabled"
       CELERITAS_USE_CUDA
       "${VecGeom_CUDA_FOUND}"
     )
   endif()
   if(CELERITAS_REAL_TYPE STREQUAL "float" AND NOT VecGeom_single_precision_FOUND)
     celeritas_error_incompatible_option(
-      "VecGeom installation at ${VECGEOM_INSTALL_DIR} uses double precision"
+      "VecGeom installation at \"${VecGeom_DIR}\" uses double precision"
       CELERITAS_REAL_TYPE
       "double"
     )
   endif()
   if(CELERITAS_REAL_TYPE STREQUAL "double" AND VecGeom_single_precision_FOUND)
     celeritas_error_incompatible_option(
-      "VecGeom installation at ${VECGEOM_INSTALL_DIR} uses single precision"
+      "VecGeom installation at \"${VecGeom_DIR}\" uses single precision"
       CELERITAS_REAL_TYPE
       "float"
     )
@@ -354,7 +354,7 @@ if(CELERITAS_USE_VecGeom)
   if(CELERITAS_BUILD_TESTS AND NOT VecGeom_GDML_FOUND
       AND CELERITAS_CORE_GEO STREQUAL "VecGeom")
     celeritas_error_incompatible_option(
-      "VecGeom installation at ${VECGEOM_INSTALL_DIR} was not built with VGDML:
+      "VecGeom installation at \"${VecGeom_DIR}\" was not built with VGDML:
       celer-sim and many tests will fail"
       CELERITAS_BUILD_TESTS
       OFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,8 +318,16 @@ if(CELERITAS_USE_SWIG)
 endif()
 
 if(CELERITAS_USE_VecGeom)
+  set(_min_vecgeom_version 1.2.4)
   if(NOT VecGeom_FOUND)
-    find_package(VecGeom 1.2.4 REQUIRED)
+    find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
+  elseif(VecGeom_VERSION VERSION_LESS _min_vecgeom_version)
+    # Another package, probably Geant4, is already using vecgeom
+    celeritas_error_incompatible_option(
+      "VecGeom version \"${VecGeom_VERSION}\" at \"${VecGeom_DIR}\" is too old for Celeritas to use: you must update to ${_min_vecgeom_version} or higher"
+      CELERITAS_USE_VecGeom
+      OFF
+    )
   endif()
 
   if(CELERITAS_USE_CUDA AND NOT VecGeom_CUDA_FOUND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,9 +343,6 @@ if(CELERITAS_USE_VecGeom)
       "float"
     )
   endif()
-  if(NOT VecGeom_GDML_FOUND)
-    message(SEND_ERROR "VecGeom GDML capability is required for Celeritas")
-  endif()
 endif()
 
 if(CELERITAS_BUILD_DEMOS AND NOT CELERITAS_USE_JSON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,15 @@ if(CELERITAS_USE_VecGeom)
       "float"
     )
   endif()
+  if(CELERITAS_BUILD_TESTS AND NOT VecGeom_GDML_FOUND
+      AND CELERITAS_CORE_GEO STREQUAL "VecGeom")
+    celeritas_error_incompatible_option(
+      "VecGeom installation at ${VECGEOM_INSTALL_DIR} was not built with VGDML:
+      celer-sim and many tests will fail"
+      CELERITAS_BUILD_TESTS
+      OFF
+    )
+  endif()
 endif()
 
 if(CELERITAS_BUILD_DEMOS AND NOT CELERITAS_USE_JSON)
--- a/app/celer-sim/CMakeLists.txt
+++ b/app/celer-sim/CMakeLists.txt
@@ -63,7 +63,8 @@ set(_env
 )
 
 if(CELERITAS_USE_Geant4 AND CELERITAS_USE_HepMC3 AND CELERITAS_USE_Python
-    AND (NOT CELERITAS_CORE_GEO STREQUAL "Geant4"))
+    AND (NOT CELERITAS_CORE_GEO STREQUAL "Geant4")
+    AND (VecGeom_GDML_FOUND OR NOT (CELERITAS_CORE_GEO STREQUAL "VecGeom")))
   set(_needs_deps)
 else()
   set(_needs_deps DISABLED true)
--- a/src/celeritas/CMakeLists.txt
+++ b/src/celeritas/CMakeLists.txt
@@ -231,6 +231,7 @@ endif()
 if(CELERITAS_USE_VecGeom)
   list(APPEND SOURCES
     ext/VecgeomParams.cc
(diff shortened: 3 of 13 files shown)


whokion:
I keep getting the following include path error even though the file is under the right path:
```
/work1/g4gpu/syjun/cms/test/src/celeritas/src/celeritas/ext/g4vg/Converter.cc:19:10: fatal error: VecGeom/management/ReflFactory.h: No such file or directory
   19 | #include <VecGeom/management/ReflFactory.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
even though the installation area of VecGeom without GDML is added to CMAKE_PREFIX_PATH.  Is the included path changed or am I doing something wrong?

sethrj:
@whokion Can you post the build line for it? Also, can you confirm that you're using vecgeom 1.2.4 and higher? (I assume so otherwise it wouldn't have configured?)

whokion:
As an initial test before further simplification for cmsdist (i.e., celeritas.spec), I used the follow cmake commands for building VecGeom (v1.2.5) and celeritas (no-vgdml):
```
make ${SRC_DIR}/VecGeom.v1.2.5 \
      -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} \
      -DVECGEOM_BACKEND:STRING="scalar" \
      -DBUILD_SHARED_LIBS:BOOL="ON" -DBUILD_TESTING:BOOL="ON" \
      -DVECGEOM_BUILTIN_VECCORE:BOOL="ON" \
      -DCMAKE_BUILD_TYPE:STRING="Release" \
      -DCMAKE_CXX_STANDARD:STRING="17" \
      -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL="OFF" \
      -DCTEST:BOOL="OFF" \
      -DVECGEOM_GDML:BOOL="OFF" -DVECGEOM_GEANT4:BOOL="OFF" \
      -DVECGEOM_ROOT:BOOL="OFF" \
      -DUSE_NAVINDEX:BOOL="ON"
```
and
```
cmake ${SRC_DIR}/celeritas \
      -DCELERITAS_USE_CUDA=OFF \
      -DCELERITAS_USE_HIP=OFF \
      -DCELERITAS_USE_MPI=OFF -DCELERITAS_USE_VecGeom=ON \
      -DCMAKE_BUILD_TYPE="Release" \
      -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic -Werror" \
      -DCMAKE_PREFIX_PATH="${INSTALL_DIR};${GEANT4_ROOT}" \
      -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}/celeritas" \
      -DCELERITAS_USE_Geant4=ON \
      -DGeant4_DIR=${GEANT4_BUILD_DIR} \
      -DCELERITAS_DEBUG=OFF \
      -DCELERITAS_USE_ROOT=OFF \
      -DCELERITAS_USE_HepMC3=OFF \
      -DCELERITAS_BUILD_TESTS=OFF \
      -DCELERITAS_USE_Python=ON \
      -DCELERITAS_USE_JSON=OFF \
      -DCELERITAS_USE_SWIG=OFF \
      -DCELERITAS_BUILD_DEMOS=OFF

```

sethrj:
Looks reasonable. What's the compiler line that produces the failing error? Could it be somehow not propagating the vecgeom `-I` include?

whokion:
Nothing much. The full error message is
```
/work1/g4gpu/syjun/cms/test/src/celeritas/src/celeritas/ext/g4vg/Converter.cc:19:10: fatal error: VecGeom/management/ReflFactory.h: No such file or directory
   19 | #include <VecGeom/management/ReflFactory.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [src/celeritas/CMakeFiles/celeritas_g4vg.dir/build.make:76: src/celeritas/CMakeFiles/celeritas_g4vg.dir/ext/g4vg/Converter.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```
It seems that other included files (<VecGeom/volumes/LogicalVolume.h> ...) are found from the same file (changing the order).  So something for the directory, `management`.   The same VecGeom version works with CUDA and GDML `ON` with the branch, so I will try to test little bit more what I screwed up.

sethrj:
Weird. I did check on my mac and also verified there's not any case sensitivity issues...

whokion:
I find that a) works, but b) fails with the compilation reported earlier:
a) 
```
cmake ${SRC_DIR}/celeritas \
      -DCELERITAS_USE_CUDA=OFF \
      -DCELERITAS_USE_HIP=OFF \
      -DCELERITAS_USE_VecGeom=ON \
      -DCMAKE_BUILD_TYPE="Release" \
      -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic -Werror" \
      -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}/celeritas" \
      -DCMAKE_PREFIX_PATH="${VECGEOM_ROOT}" \
      -DCELERITAS_USE_Geant4=OFF
```
b)
```
cmake ${SRC_DIR}/celeritas \
      -DCELERITAS_USE_CUDA=OFF \
      -DCELERITAS_USE_HIP=OFF \
      -DCELERITAS_USE_VecGeom=ON \
      -DCMAKE_BUILD_TYPE="Release" \
      -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic -Werror" \
      -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}/celeritas" \
      -DCMAKE_PREFIX_PATH="${VECGEOM_ROOT};${GEANT4_ROOT}" \
      -DCELERITAS_USE_Geant4=ON
```
As I look the relevant section at lines 100-112, inside `src/celeritas/CMakeLists.txt`
```
    100 if(CELERITAS_USE_VecGeom AND CELERITAS_USE_Geant4)
    101   set( _cg4vg_sources
    102     ext/g4vg/Converter.cc
    103     ext/g4vg/LogicalVolumeConverter.cc
    104     ext/g4vg/SolidConverter.cc
    105   )
    106   set(_cg4vg_libs Celeritas::corecel VecGeom::vecgeom ${Geant4_LIBRARIES})
    107 
    108   celeritas_add_object_library(celeritas_g4vg ${_cg4vg_sources})
    109   celeritas_target_link_libraries(celeritas_g4vg PRIVATE ${_cg4vg_libs})
    110 
    111   list(APPEND SOURCES $<TARGET_OBJECTS:celeritas_g4vg>)
    112   list(APPEND PRIVATE_DEPS celeritas_g4vg)
    113 endif()
```
it seems that the include path to VecGeom and the library  VecGeom::vecgeom are not somehow pickup inside the if-block,
which causes the compilation (and the linking error later).

There is one more VecGeom compilation error with gcc12 (version used by the latest CMSSW, 14_0_0_pre1) at
```
VecGeom/base/VariableSizeObj.h:194:46: warning: unused parameter 'nvalues' [-Wunused-parameter]
  194 |   static constexpr size_t SizeOfExtra(size_t nvalues) { return 0; }
      |                                       ~~~~~~~^~~~~~~
```

whokion:
@sethrj As a second thought, I realized that CMS ${GEANT4_ROOT} here was not build with VecGeom..1.2.5, which may be the resason, why both VecGeom paths are not propagated to Geant4 routines.  Will try again, but seems that MR looks good. Will let you know how it goes.

sethrj:
Ahhh ok, since Geant4 is built against vecgeom,
```
    find_package(VecGeom 1.2.4 REQUIRED)
```
doesn't get called so the version requirement doesn't get enforced. I'll make sure the version gets checked and a diagnostic printed if that's the case.

sethrj:
@whokion Can you see if celeritas now correctly catches the too-old version?

sethrj:
Also @whokion can you please post the *compile line* (you may need to `make VERBOSE=1` or use ninja) for the errors? The compiler should be using `-isystem/path/to/vecgeom/include` which shouldn't propagate warnings to includes outside our control.

whokion:
With the local Geant4 built (with options close to the cmsdist/geant4.spec) with the local [VecGeom v1.2.5 without GDML], the branch is built successfully - and, do not see the compilation warning from VecGeom either.  Have not tested the latest commits yet, but all seem fine now.  Will double check shortly.
One related issue is that there is no tests with celeritas without GDML support (do we?), which may not be necessary as the build is to support a special case (like CMSSW), but may add later if needed.

sethrj:
@whokion the `celer-g4` tests (enabled by using `BUILD_TESTING=ON` which enables CTest but leaving `CELERITAS_BUILD_TESTS=OFF` to avoid the unit tests) will check the ability to set up and run vecgeom through GPU via the Geant4 converter. I'll add a note in the description.

sethrj:
If you think we *need* most of the tests to be on, I could add that in a follow-up PR?

whokion:
> If you think we _need_ most of the tests to be on, I could add that in a follow-up PR?

I don't think so as the geometry related tests should work switching from GDML input to an equivalent detector construction manually (which is an internal step/conversion anyway). Of course, no hurt to add something if it does not require much work, but  I was just double checking! Thanks.

I tested the latest updated, which looks good for what we are going to request to the cmsdist team.  I also got the expected cmake warning if I use both the default VecGeom (1.2.1 without GDML) and Geant4 (11.1.2 with GDML support) from the CMSSW_14_0_0_pre1:
```
 CMake Error at cmake/CeleritasUtils.cmake:339 (message):
  Invalid setting CELERITAS_USE_VecGeom=ON: VecGeom version "1.2.1" at
  "/cvmfs/cms.cern.ch/slc7_amd64_gcc12/external/vecgeom/v1.2.1-35adfe5c99700701c19371de9114c36a/lib64/cmake/VecGeom"
  is too old for Celeritas to use: you must update to 1.2.4 or higher

      Possible fix: cmake -DCELERITAS_USE_VecGeom=OFF /work1/g4gpu/syjun/cms/test/build/celeritas
Call Stack (most recent call first):
  CMakeLists.txt:326 (celeritas_error_incompatible_option)
```
So, the MR is good to go.  I will quickly go through all commits for the final look before the approval.
Thanks for the update!!!

sethrj:
Great! Sounds like the next step as far as CMSSW is concerned is to update to VecGeom 1.2.5.

whokion:
> Great! Sounds like the next step as far as CMSSW is concerned is to update to VecGeom 1.2.5.

Yup, it will be a part of out request to the cmsdist, which I will open sometime today.

whokion (approved):

PR #1047, opened 2023-12-04
Refactor TrackSortUtils

esseivaju:
Simplify the `TrackSortUtils` implementation:
- Extract common behavior to helper functions
- Remove unecessary template arguments of function template
- Remove unnecssary streamid in host-only helper functions
- Add track sorting by particle type on host (missed in #1044)
- Remove count_tracks_per_action impl

### DIFF
--- a/src/celeritas/track/detail/TrackSortUtils.cu
+++ b/src/celeritas/track/detail/TrackSortUtils.cu
@@ -46,7 +46,9 @@ using ThreadItems
 using TrackSlots = ThreadItems<TrackSlotId::size_type>;
 
 //---------------------------------------------------------------------------//
-
+/*!
+ * Partition track_slots based on predicate.
+ */
 template<class F>
 void partition_impl(TrackSlots const& track_slots, F&& func, StreamId stream_id)
 {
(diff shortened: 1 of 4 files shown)


src/celeritas/track/detail/TrackSortUtils.hh
```diff
@@ -89,13 +85,14 @@ struct alive_predicate
     }
 };
 
-struct action_comparator
+template<class Id>
+struct id_comparator
```
sethrj: While you're at it, could you fix the spelling of these structs `IdComparator`, `AlivePredicate`, ...?

src/celeritas/track/detail/TrackSortUtils.cu
```diff
@@ -96,13 +96,15 @@ void sort_impl(TrackSlots const& track_slots,
     CELER_DEVICE_CHECK_ERROR();
 }
 
-// PRE: get_action is sorted, i.e. i <= j ==> get_action(i) <=
-// get_action(j)
-template<class F>
-__device__ void
-tracks_per_action_impl(Span<ThreadId> offsets, size_type size, F&& get_action)
+// PRE: actions are sorted
```
sethrj: I guess I allowed the separators/comments to be skipped during the initial review, but 
 ```suggestion
//---------------------------------------------------------------------------//
/*!
 * Calculate thread boundaries based on action ID.
 * \pre actions are sorted
 */
```

sethrj (approved):
Nice simplification. Just a couple of style comments.

PR #1048, opened 2023-12-05
Support JSON via stdin for celer-g4

sethrj:
This fully supports the stdin/stdout use case for celer-g4: input can be passed via JSON with the `-` argument, and this fixes the `accel` output so that the "pre-run" JSON output is not written to stdout.


amandalund (approved):
Thanks @sethrj!

PR #1049, opened 2023-12-05
Implement jump ahead for XORWOW RNG

amandalund:
This lets us initialize the XORWOW RNG with a subsequence and an offset to skip ahead in the stream. The jump ahead method is implemented using the polynomial representation of the recurrence, which requires less memory than the jump matrices. The jump polynomials are precalculated and can be generated using a [script in the utils repo](https://github.com/celeritas-project/utils/blob/main/prng/xorwow-jump.py).


src/celeritas/random/XorwowRngEngine.hh
```diff
@@ -51,15 +67,46 @@ class XorwowRngEngine
...
+                                          TrackSlotId tid);
+
+    // Initialize state
+    inline CELER_FUNCTION XorwowRngEngine& operator=(Initializer_t const&);
 
     // Generate a 32-bit pseudorandom number
     inline CELER_FUNCTION result_type operator()();
 
+    // Jump ahead \c num_skip steps
+    inline CELER_FUNCTION void skipahead(ull_int num_skip);
```
sethrj: Rename to `discard` to conform to `std` PRNGs: i.e. [this LCG](https://en.cppreference.com/w/cpp/numeric/random/linear_congruential_engine/discard)

src/celeritas/random/XorwowRngEngine.hh
```diff
@@ -51,15 +67,46 @@ class XorwowRngEngine
...
+
+    // Jump ahead \c num_skip subsequences (\c num_skip * 2^67 steps)
+    inline CELER_FUNCTION void skipahead_subsequence(ull_int num_skip);
```
sethrj: It would be good to annotate that this function is public only for testing purposes (or maybe it would be better to just make it private?)

src/celeritas/random/XorwowRngData.hh
```diff
@@ -18,19 +18,31 @@ namespace celeritas
...
+    Array<uint_t, 1> seed;
+
+    // Jump polynomials
+    ArrayJumpPoly jump;
+    ArrayJumpPoly jump_subsequence;
 
     //// METHODS ////
 
+    static CELER_CONSTEXPR_FUNCTION size_type num_words() { return 5; }
+    static CELER_CONSTEXPR_FUNCTION size_type num_bits() { return 32; }
```
sethrj: I think this should work and be more expressive:
```suggestion
    static CELER_CONSTEXPR_FUNCTION size_type num_words() { return JumpPoly{}.size(); }
    static CELER_CONSTEXPR_FUNCTION size_type num_bits() { return 8 * sizeof(uint_t); }
```
or should the second one be `ArrayJumpPoly{}.size()`?

src/celeritas/random/CuHipRngEngine.hh
```diff
@@ -97,10 +100,13 @@ class GenerateCanonical<CuHipRngEngine, double>
  * Construct from state.
  */
 CELER_FUNCTION
-CuHipRngEngine::CuHipRngEngine(StateRef const& state, TrackSlotId const& id)
+CuHipRngEngine::CuHipRngEngine(ParamsRef const& params,
+                               StateRef const& state,
+                               TrackSlotId tid)
+    : params_(params)
```
sethrj: Is `params_` unused? You could just omit the local storage of it for clarity. The important thing of course is just that the interface is consistent across RNGs.

src/celeritas/random/XorwowRngEngine.hh
```diff
@@ -88,20 +135,95 @@ class GenerateCanonical<XorwowRngEngine, RealType>
...
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize the RNG engine.
+ *
+ * This moves the state ahead to the given subsequence (a subsequence has size
+ * 2^67) and skips \c offset random numbers.
+ *
+ * It is recommended to initialize the state using a very different generator
+ * from the one being initialized to avoid correlations. Here the 64-bit
+ * SplitMis64 generator is used for initialization (See Matsumoto, Wada,
```
sethrj:
```suggestion
 * SplitMix64 generator is used for initialization (See Matsumoto, Wada,
```

src/celeritas/random/XorwowRngEngine.hh
```diff
@@ -111,9 +233,98 @@ CELER_FUNCTION auto XorwowRngEngine::operator()() -> result_type
...
+/*!
+ * Generate a 64-bit pseudorandom number using the SplitMix64 engine.
+ *
+ * This is used to initialize the XORWOW state. See https://prng.di.unimi.it.
+ */
+CELER_FUNCTION uint64_t XorwowRngEngine::SplitMix64::operator()()
+{
+    uint64_t z = (state += 0x9e3779b97f4a7c15);
+    z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
+    z = (z ^ (z >> 27)) * 0x94d049bb133111eb;
```
sethrj:
```suggestion
    uint64_t z = (state += 0x9e3779b97f4a7c15ull);
    z = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9ull;
    z = (z ^ (z >> 27)) * 0x94d049bb133111ebull;
```

test/celeritas/random/XorwowRngEngine.test.cc
```diff
@@ -205,6 +205,62 @@ TEST_F(XorwowRngEngineTest, moments)
     tally.check(num_samples * num_seeds, 1e-3);
 }
 
+TEST_F(XorwowRngEngineTest, jump)
```
sethrj, line 208: It would be also good to show that discarding `N` produces the same value as evaluating `N` times? (For small N obviously; there's no need to discard 2^67...)

sethrj (commented):
Looks great! Just a couple of clarifications.

src/celeritas/random/XorwowRngEngine.hh
amandalund: Made them private.

test/celeritas/random/XorwowRngEngine.test.cc
amandalund: I show this for all `N` between 0 and 2^16, I think that's what you meant?

sethrj: Oh. Sorry I missed that!

sethrj: Ok I see, it's being done implicitly through initialization rather than through discard. 🤔

amandalund: Any other ideas for testing if discard is private (in general, not just for this test)?

sethrj: Oh sorry I just meant the "sequence" skipahead... the regular `discard(count)` is public for the standard-provided RNGs so that seems reasonable to keep.

amandalund: Ahh right that makes sense.

test/celeritas/random/XorwowRngEngine.test.cc
```diff
@@ -227,38 +227,36 @@ TEST_F(XorwowRngEngineTest, jump)
...
         skip_rng = init;
 
-        rng.skipahead_subsequence(1 << 18);
-        rng.skipahead_subsequence(1 << 18);
-        rng.skipahead(1023);
-        rng.skipahead(1);
+        // Skip 2**67 times to get to the next subsequence
+        for (size_type i = 0; i < 8; ++i)
+        {
+            rng.discard(-1ull);
```
sethrj: Could you do `numeric_limits<unsigned long long>::max()` for clarity? (Or `0xffffffffffffffffull` if you want to be exact?)

sethrj (approved):

PR #1050, opened 2023-12-06
Implement offload to Celeritas from Geant4 using G4VTrackingManager hook

### LINKED ISSUE #1035: Implement offload to Celeritas via G4VTrackingManager
Per the original mention in #992 and discussion in the AdePT/Celeritas catch up, the third(!) offload option to to try out is through the per-particle `G4VTrackingManager` interface introduced in Geant4 11.0. References:

- https://indico.cern.ch/event/1073798/contributions/4516033/attachments/2304502/3920446/specialized-tracking-manager.pdf
- https://indico.cern.ch/event/1052654/contributions/4524767/attachments/2309218/3929219/G4HepEm_SpecTracking_MNovak.pdf
- AdePT's example integration using this is https://github.com/apt-sim/AdePT/tree/master/examples/Example21

It's a little more involved due to it being an addition/override of the physics list. Might be a topic to cover during the review week, especially as related to the region/kinematics/etc filters on offload that have been mentioned.

drbenmorgan:
This implements a minimal concrete class of the `G4VTrackingManager` interface for offloading Geant4 tracks to Celeritas, complementing the existing user tracking action and fast simulation strategies. Introduced in Geant4 11.0, a `G4VTrackingManager` instance can be added to one or more `G4ParticleDefinition` instances to customise how tracks of that particle the will be handled. Overviews of this capability, and an example are documented at these links:

- [Specialized Tracking Managers in Geant4](https://indico.cern.ch/event/1073798/contributions/4516033/attachments/2304502/3920446/specialized-tracking-manager.pdf)
- [G4HepEm and Specialized Stepping/Tracking in Geant4](https://indico.cern.ch/event/1052654/contributions/4524767/attachments/2309218/3929219/G4HepEm_SpecTracking_MNovak.pdf)
- [Geant4 runAndEvent RE07 Example](https://indico.cern.ch/event/1156193/contributions/5048470/attachments/2514883/4326508/20220927_Hahnfeld_RE07.pdf)
  - [Current implementation in Geant4 11.1](https://gitlab.cern.ch/geant4/geant4/-/tree/v11.1.3/examples/extended/runAndEvent/RE07?ref_type=tags)

Because implementation requires inheritance from a class only available from Geant4 v11.0, the build/use has been restricted to that version or newer for now.

The implementation here for the offload/flush is largely the same as in `FastSimulationOffload`, using `SharedParams` and `LocalTransporter`. 

The main detail is the necessity to implement the `Prepare/BuildPhysicsTable` member functions as per-particle tracking managers completely overrides physic building for the particle in question (see [here](https://gitlab.cern.ch/geant4/geant4/-/blob/v11.1.3/source/run/src/G4VUserPhysicsList.cc?ref_type=tags#L793-799), and [here](https://gitlab.cern.ch/geant4/geant4/-/blob/v11.1.3/source/run/src/G4VUserPhysicsList.cc?ref_type=tags#L644-658) in `G4VUserPhysicsList`). If we don't do this, no Geant4 physics data is available for that particle, so Celeritas can't construct its physics properly. `Prepare/BuildPhysicsTable` are therefore implemented to follow the default in `G4VUserPhysicsList`, as is also shown in [the runAndEvent/RE07 example](https://gitlab.cern.ch/geant4/geant4/-/blob/v11.1.3/examples/extended/runAndEvent/RE07/src/SpecializedTrackingManager.cc?ref_type=tags). _Aside: this rebuild of the physics data would also allow filtering in `HandOffOneTrack` to pass back to general tracking if certain preconditions are met, much like in fast simulation as well._

Like `FastSimulationOffload`, the current implementation does rely on the setup and use of `SimpleOffload`s Run and Event actions to properly setup/teardown params and transporter(s). An example of the use of `TrackingManagerOffload` is shown in a new program in `example/accel` which is identical to the others in problem setup, just using the new offload mechanism. This has also been documented.

Possible TODO: the example shows that the custom tracking manager is constructed and added to the particle definitions in the `ConstructProcess` member function. AFAICT this is only called on the master thread and thus with the `LocalTransporter` for that thread. However, local tests showed (via logging the address) that the correct thread-local instances are all used on the worker threads. The `G4VTrackingManager` instance held by `G4ParticleDefinition` _is_ part of the split data for that class, so seems to work correctly, but I'd still like to understand this better!

Addresses: #1035

### DIFF
--- a/example/accel/CMakeLists.txt
+++ b/example/accel/CMakeLists.txt
@@ -52,38 +52,48 @@ endif()
 #-----------------------------------------------------------------------------#
 
 include(CTest)
-function(add_example target)
-  add_test(NAME "${target}" COMMAND "$<TARGET_FILE:${target}>")
-endfunction()
-
-set(_test_env)
+set(_g4data_env)
 foreach(_ds IN LISTS Geant4_DATASETS)
-  list(APPEND _test_env
+  list(APPEND _g4data_env
     "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}")
 endforeach()
+
+function(add_example target)
+  add_example_impl(${target} Serial)
+  if(Geant4_multithreaded_FOUND)
+    add_example_impl(${target} MT)
+    add_example_impl(${target} Tasking)
+  endif()
+endfunction()
+
+function(disable_example target)
+  set(_tests ${target}-Serial)
+  if(Geant4_multithreaded_FOUND)
+    list(APPEND _tests ${target}-MT ${target}-Tasking)
+  endif()
+  set_tests_properties(${_tests} PROPERTIES
+    DISABLED true
+  )
+endfunction()
+
+function(add_example_impl target rmtype)
+  set(_test_name "${target}-${rmtype}")
+  add_test(NAME "${_test_name}" COMMAND "$<TARGET_FILE:${target}>")
+  set_tests_properties(${_test_name} PROPERTIES
+    ENVIRONMENT "${_g4data_env};G4RUN_MANAGER_TYPE=${rmtype}"
+    LABELS "app"
+  )
+endfunction()
 
 #-----------------------------------------------------------------------------#
 
 add_example(simple-offload)
 add_example(fastsim-offload)
 
-set_tests_properties(simple-offload fastsim-offload PROPERTIES
-  ENVIRONMENT "${_test_env}"
-  LABELS "app"
-)
-
 if(Geant4_VERSION VERSION_LESS 11.1)
-  set_tests_properties(fastsim-offload PROPERTIES
-    DISABLED true
-  )
+  disable_example(fastsim-offload)
 endif()
 
 if(Geant4_VERSION VERSION_GREATER_EQUAL 11.0)
   add_example(trackingmanager-offload)
-  set_tests_properties(trackingmanager-offload PROPERTIES
-    ENVIRONMENT "${_test_env}"
-    LABELS "app"
-  )
 endif()
-
-
--- a/example/accel/CMakeLists.txt
+++ b/example/accel/CMakeLists.txt
@@ -35,6 +35,16 @@ celeritas_target_link_libraries(fastsim-offload
   ${Geant4_LIBRARIES}
 )
 
+if(Geant4_VERSION VERSION_LESS 11.0)
+  message(WARNING "G4VTrackingManager offload requires Geant4 11.0 or higher")
+else()
+  add_executable(trackingmanager-offload trackingmanager-offload.cc)
+  celeritas_target_link_libraries(trackingmanager-offload
+    Celeritas::accel
+    ${Geant4_LIBRARIES}
+  )
+endif()
+
 # END EXAMPLE CODE
 
 #-----------------------------------------------------------------------------#
@@ -67,4 +77,13 @@ if(Geant4_VERSION VERSION_LESS 11.1)
     DISABLED true
   )
 endif()
+
+if(Geant4_VERSION VERSION_GREATER_EQUAL 11.0)
+  add_example(trackingmanager-offload)
+  set_tests_properties(trackingmanager-offload PROPERTIES
+    ENVIRONMENT "${_test_env}"
+    LABELS "app"
+  )
+endif()
+
 
--- a/example/accel/trackingmanager-offload.cc
+++ b/example/accel/trackingmanager-offload.cc
@@ -91,6 +91,7 @@ class EMPhysicsConstructor final : public G4EmStandardPhysics
 
     void ConstructProcess() override
     {
+        CELER_LOG_LOCAL(status) << "Setting up tracking manager offload";
         G4EmStandardPhysics::ConstructProcess();
 
         // Add Celeritas tracking manager to electron, positron, gamma.
--- a/example/accel/trackingmanager-offload.cc
+++ b/example/accel/trackingmanager-offload.cc
@@ -1,0 +1,209 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file example/accel/trackingmanager-offload.cc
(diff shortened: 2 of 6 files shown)


example/accel/trackingmanager-offload.cc
```diff
@@ -0,0 +1,209 @@
...
+};
+
+//---------------------------------------------------------------------------//
+class EMPhysicsConstructor final : public G4EmStandardPhysics
+{
+  public:
+    using G4EmStandardPhysics::G4EmStandardPhysics;
+
+    void ConstructProcess() override
+    {
```
sethrj, line 93: The physics constructor class instance is "shared", but is `ConstructProcess` done thread-locally? For the sake of clarity for the example I'd add:

```suggestion
    {
        CELER_LOG_LOCAL(status) << "Setting up tracking manager offload";
```

example/accel/trackingmanager-offload.cc
```diff
@@ -0,0 +1,209 @@
...
+        G4EmStandardPhysics::ConstructProcess();
+
+        // Add Celeritas tracking manager to electron, positron, gamma.
+        auto* celer_tracking = new celeritas::TrackingManagerOffload(
+            &shared_params, &local_transporter);
+        G4Electron::Definition()->SetTrackingManager(celer_tracking);
+        G4Positron::Definition()->SetTrackingManager(celer_tracking);
+        G4Gamma::Definition()->SetTrackingManager(celer_tracking);
```
sethrj, line 102: Does anyone take ownership of this pointer or is it leaked?

src/accel/TrackingManagerOffload.cc
```diff
@@ -0,0 +1,141 @@
...
+
+#include "ExceptionConverter.hh"
+#include "LocalTransporter.hh"
+#include "SharedParams.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a tracking manager with data needed to offload to Celeritas
```
sethrj: For the "autobrief" doxygen to work, the first line here (and in all the rest of the docs) need a trailing full stop `.` .

example/accel/CMakeLists.txt
```diff
@@ -68,3 +78,12 @@ if(Geant4_VERSION VERSION_LESS 11.1)
   )
 endif()
 
+if(Geant4_VERSION VERSION_GREATER_EQUAL 11.0)
+  add_example(trackingmanager-offload)
+  set_tests_properties(trackingmanager-offload PROPERTIES
+    ENVIRONMENT "${_test_env}"
+    LABELS "app"
+  )
+endif()
```
sethrj, line 99: I wonder with the more complex threading going on, should we run these examples with serial/mt/task? (Task is the default in 11.)

src/accel/TrackingManagerOffload.cc
```diff
@@ -0,0 +1,141 @@
...
+ * The implementation follows that in \c G4VUserPhysicsList::BuildPhysicsTable
+ * , see also Geant4 Extended Example runAndEvent/RE07.
+ */
+void TrackingManagerOffload::BuildPhysicsTable(G4ParticleDefinition const& part)
+{
+    G4ProcessManager* pManager = part.GetProcessManager();
+    G4ProcessManager* pManagerShadow = part.GetMasterProcessManager();
+
+    G4ProcessVector* pVector = pManager->GetProcessList();
+    for (std::size_t j = 0; j < pVector->size(); ++j)
```
sethrj:
```suggestion
    CELER_ASSERT(pManager && pManager->GetProcessList());
    for (G4VProcess* process : *pManager->GetProcessList())
```
and below `(*pVector)[j]` -> `process`

sethrj (commented):
Looks great! So glad to have you helping us with this, as it looks like the thread abstractions are a real headache 😄

example/accel/trackingmanager-offload.cc
drbenmorgan: The deletion of the pointer is done in the destructor of `G4VUserPhysicsList` via the [`RemoveTrackingManager` member function](https://gitlab.cern.ch/geant4/geant4/-/blob/v11.1.3/source/run/src/G4VUserPhysicsList.cc?ref_type=tags#L326-356). That's analogous to what's done for process managers, albeit making the ownership non-definite.

example/accel/CMakeLists.txt
drbenmorgan: It certainly doesn't hurt - would just be a case of using the `G4FORCE_RUN_MANAGER_TYPE` env var with `Serial`, `MT`, `Tasking` to select the mode (so, three tests per app assuming an MT supporting install of Geant4). That's only going to be a smoke test though as there's otherwise no scoring or reference output check done.

src/accel/TrackingManagerOffload.cc
drbenmorgan: Despite its name, `G4ProcessVector` doesn't provide an interface compatible with, or trivially adaptable to, range-based for ☹️! We could write the loop as is done in `GeantImporter`:

```cpp
    CELER_ASSERT(pManager);
    G4ProcessVector* pVector = pManager->GetProcessList();
    CELER_ASSERT(pVector);
    for(auto j : range(pVector->size())
    {
        G4VProcess* p = *pVector[j];
        ....
```
?

example/accel/CMakeLists.txt
drbenmorgan: Sorry, realised my comment here was a bit vague - I'll go ahead and add these tests unless you say otherwise.

drbenmorgan: Now done - it should work all the way back to Geant4 10.6, worst case 10.7.

src/accel/TrackingManagerOffload.cc
drbenmorgan: Implemented as above, which I think is the best that can be done without adding an iterator adaptor for `G4ProcessVector` (will add a Geant4 issue to review collection interfaces for "proper" behaviour!).

example/accel/CMakeLists.txt
sethrj: Nice!! Great work, thanks @drbenmorgan !

sethrj (approved):
Excellent!

src/accel/TrackingManagerOffload.cc
sethrj: Ha, sounds good. Cheers!

PR #1051, opened 2023-12-06
Uniquify ROOT file name when output is directed to stdout

pcanal:
Use `stdout-${PID}.root` for the ROOT outputname.

### DIFF
--- a/app/celer-g4/RootIO.cc
+++ b/app/celer-g4/RootIO.cc
@@ -56,7 +56,7 @@ RootIO::RootIO()
 
     if (file_name_ == "-")
     {
-        file_name_ = "stdout-" + std::to_string(::getpid());
+        file_name_ = "stdout-" + std::to_string(::getpid()) + ".root";
     }
 
     if (G4Threading::IsWorkerThread())
--- a/app/celer-g4/RootIO.cc
+++ b/app/celer-g4/RootIO.cc
@@ -54,6 +54,11 @@ RootIO::RootIO()
         file_name_ = "celer-g4.root";
     }
 
+    if (file_name_ == "-")
+    {
+        file_name_ = "stdout-" + std::to_string(::getpid());
+    }
+
     if (G4Threading::IsWorkerThread())
     {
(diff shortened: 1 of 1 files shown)


app/celer-g4/RootIO.cc
```diff
@@ -54,6 +54,11 @@ RootIO::RootIO()
         file_name_ = "celer-g4.root";
     }
 
+    if (file_name_ == "-")
+    {
+        file_name_ = "stdout-" + std::to_string(::getpid()) + ".root";
```
sethrj, line 63: I'm sure ROOT leaks the include, but this requires `#include <unistd.h>` on POSIX machines... does it require anything on windows?

sethrj (changes requested):

app/celer-g4/RootIO.cc
pcanal: It does need `process.h`.  What is our Windows marker/ifdef?

pcanal: It is the (misnamed) `_WIN32`.

app/celer-g4/RootIO.cc
```diff
@@ -28,6 +28,10 @@
 #include "GlobalSetup.hh"
 #include "SensitiveHit.hh"
 
+#ifdef _WIN32
+#    include <process.h>
```
sethrj, line 32: Should this be
```suggestion
#ifndef _WIN32
#    include <unistd.h>
#else
#    include <process.h>
```
maybe? (or `defined(__APPLE__) || defined(__linux__)`?)

sethrj: We can IWYU later.

sethrj (approved):

app/celer-g4/RootIO.cc
pcanal: I prefer
```
#ifdef __WIN32
do something silly
#else
do something sensitive
#endif
```
(or `#ifndef `)

pcanal: You merged while the commit was moving towards github :)

PR #1052, opened 2023-12-06
Update tests to pass with cuRAND

amandalund:
This makes a few minor changes to our unit tests so they all pass with both Celeritas core RNGs.


sethrj (approved):
Looks great!

PR #1053, opened 2023-12-07
Include documentation as mandatory CI

sethrj:
This updates the build flags and workflow to:
- require documentation to be built as a precondition to the "full" CI jobds
- reduce the documentation build time for pull requests
- mark warnings as errors for PRs
- don't error when "push" to develop (in case something odd happens due to the change from disabling to enabling dot)
- fix a few warnings and suppress some for the doxygen 1.9.1 that's in ubunto


pcanal (approved):
LGTM

PR #1054, opened 2023-12-07
Fix ROOT dictionary paths

pcanal:
This allows to run code requiring the dictionary (eg. `celer-sim`) from the install directory (and when the source is not accessible)

In addition explicit include `<unistd.h>` for process id.


sethrj (approved):
Oh ROOT. Thanks @pcanal!

PR #1057, opened 2023-12-09
Fix celer-g4 primary generator MT reproducibility

amandalund:
This reseeds the celer-g4 primary generator RNG for each event so MT results _without_ Celeritas offloading are reproducible. It also fixes the check that `num_events` have been generated (which was based on the per-thread count rather than the total count).

### DIFF
--- a/app/celer-g4/PGPrimaryGeneratorAction.cc
+++ b/app/celer-g4/PGPrimaryGeneratorAction.cc
@@ -32,9 +32,7 @@ PGPrimaryGeneratorAction::PGPrimaryGeneratorAction(
     // Generate one particle at each call to \c GeneratePrimaryVertex()
     gun_.SetNumberOfParticles(1);
 
-    // Seed with an independent value for each thread
-    rng_.seed(options.seed + get_geant_thread_id());
-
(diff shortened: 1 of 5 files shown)


app/celer-g4/PGPrimaryGeneratorAction.cc
```diff
@@ -57,32 +55,38 @@ PGPrimaryGeneratorAction::PGPrimaryGeneratorAction(
...
     {
         return;
     }
 
+    // Seed with an independent value for each event. Since Geant4 schedules
+    // events dynamically, the same event ID may not be mapped to the same
+    // thread across multiple runs. For reproducibility, Geant4 reseeds each
+    // worker thread's RNG at the start of each event with a seed calculated
+    // from the event ID.
+    rng_.seed(seed_ + event_id);
```
sethrj, line 71: Currently the primary generator RNGs are all `std::mt1997`, but we should consider some sort of wrapper interface that lets us do seed+stride for the XORWOW generators (and any others we implement in the future)

sethrj (approved):
Nice! Good catch.

PR #1059, opened 2023-12-10
Insert SortTrackAction when sorting by particle type

esseivaju:
I didn't update the switch statement which is inserting the SortTracksAction in #1044. 

Also refactoring CoreState to only have the necessary thread_offsets_ collection, moving all the template code to the internal state struct `CoreStateThreadOffsets`.

### DIFF
--- a/src/celeritas/global/CoreParams.cc
+++ b/src/celeritas/global/CoreParams.cc
@@ -264,7 +264,9 @@ CoreParams::CoreParams(Input input) : input_(std::move(input))
             insert_sort_tracks_action(TrackOrder::sort_step_limit_action);
             insert_sort_tracks_action(TrackOrder::sort_along_step_action);
             break;
-        default:
+        case TrackOrder::unsorted:
+        case TrackOrder::shuffled:
+        case TrackOrder::size_:
             break;
     }
 
--- a/src/celeritas/global/CoreParams.cc
+++ b/src/celeritas/global/CoreParams.cc
@@ -255,6 +255,7 @@ CoreParams::CoreParams(Input input) : input_(std::move(input))
         case TrackOrder::partition_status:
         case TrackOrder::sort_step_limit_action:
         case TrackOrder::sort_along_step_action:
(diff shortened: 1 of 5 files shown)


sethrj:
Changing the "default" to spelling out the remaining options would be safer since clang will catch the omission.

esseivaju:
yes, I was considering doing that (or using pragma diagnostic switch-enum?) but given that we do a switch on that enum in many places and it's not likely to get updated often, I ended up leaving the default. Maybe just doing it in accel is fine since it's not closely related to SortTracksAction you're more likely to forget it here.

sethrj:
For some reason github isn't letting me code review. Two comments:

1. Can you add a super small test to TrackSort.test.cc that fails before the fix and passes after? (maybe we need something more "manual" than running a whole tracking loop? i.e. something to fill up some bogus states and just launch the action)
2. Does `is_sort_by_action` in SortTracksAction.cc also need to be updated? If so maybe it needs to be rephrased by negating the enums that *don't* need sorting.

esseivaju:
> 1. Can you add a super small test to TrackSort.test.cc that fails before the fix and passes after? (maybe we need something more "manual" than running a whole tracking loop? i.e. something to fill up some bogus states and just launch the action)

Mmh, how would you test that? You have to instantiate `CoreParams` and check that it is in the expected state given its input. Wouldn't that fit better in a dedicated `CoreParams` unit test?

> 2. Does `is_sort_by_action` in SortTracksAction.cc also need to be updated? If so maybe it needs to be rephrased by negating the enums that _don't_ need sorting.

This doesn't need to be updated. This is used when sorting on a key that would allow to launch a kernel with a reduced grid size. This is not the case for particle_type.

sethrj:
TrackSort.test.cc already does create core instances. You could manually call the initialize track action to set a bunch of primaries, then just manually call the sort action, then check that the order is correct?

sethrj (approved):
Nice improvement. Thanks for figuring out a way to add the tests!

PR #1060, opened 2023-12-11
Add environment variable to kill Celeritas-supported tracks in Geant4

amandalund:
This lets us estimate the maximum speedup for offloading tracks to Celeritas. For the latest run3/4 on A100 I measured:

cms-run3: 3.06x
cms-run3+rzmap: 3.45x
cms-run3+uniform: 5.99x
cms-hllhc: 3.23x
cms-hllhc+rzmap: 3.44x
cms-hllhc+uniform: 5.29x

### DIFF
--- a/app/celer-g4/TrackingAction.cc
+++ b/app/celer-g4/TrackingAction.cc
@@ -17,6 +17,8 @@
 #include "corecel/Macros.hh"
 #include "accel/ExceptionConverter.hh"
 
+#include "GlobalSetup.hh"
+
 namespace celeritas
 {
 namespace app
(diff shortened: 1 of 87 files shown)


sethrj:
Nice! Thanks for answering the question from the seminar 😄

sethrj (changes requested):
I think I need a little convincing that this is the right approach: if we disable celeritas via an environment variable, maybe we should do one here too; this could also be useful to expose downstream rather than in just Celeritas. I'm also hesitant to hardcode the particle list for the celeritas case (maybe we could just do that for geant4 only?).

amandalund:
I think an environment variable makes sense. Any preference on what to call it? Right now the hardcoded particle list should only be used when Celeritas is disabled. I'll add an assertion and a comment to clarify.

sethrj:
How about `CELER_KILL_KILL_KILL`

app/celer-g4/celer-g4.cc
```diff
@@ -88,6 +88,7 @@ void print_usage(std::string_view exec_name)
                  "  CELER_DISABLE: nonempty disables offloading\n"
                  "  CELER_DISABLE_DEVICE: nonempty disables CUDA\n"
                  "  CELER_DISABLE_ROOT: nonempty disables ROOT I/O\n"
+                 "  CELER_KILL_OFFLOAD: nonempty kills offload tracks\n"
```
sethrj, line 91: Could you also add this to doc/main/usage.rst?

app/celer-g4/TrackingAction.cc
```diff
@@ -50,19 +50,23 @@ void TrackingAction::PreUserTrackingAction(G4Track const* track)
...
+        {
+            // Celeritas is transporting this track
+            ExceptionConverter call_g4exception{"celer0003", params_.get()};
+            CELER_TRY_HANDLE(transport_->Push(*track), call_g4exception);
+            const_cast<G4Track*>(track)->SetTrackStatus(fStopAndKill);
+        }
+        else if (SharedParams::KillOffloadTracks())
+        {
+            const_cast<G4Track*>(track)->SetTrackStatus(fStopAndKill);
+        }
```
sethrj, line 67: The search through particles (in the no-celeritas, no-kill case, i.e. our geant benchmark case) probably isn't anything to worry about performance wise, but I wonder if it'd be better to have the function look like:
```c++
bool const use_celeritas = !SharedParams::CeleritasDisabled();
bool const kill_em = SharedParams::KillOffloadTracks();

if (!use_celeritas && !kill_em)
    return;

// Now search
if (is_em(track)) {
if (use_celeritas) {
  offload(track);
}
kill(track);
}
```
so avoid the search in the geant4-only cacse, and remove the duplicate stop-and-kill call

sethrj (commented):
Thanks! Just two more small comments 😉

app/celer-g4/TrackingAction.cc
amandalund: `kill 'em`

sethrj: 🤖 🔪

sethrj (approved):
Looks great and a nice feature for the many times this question will arise. Thanks!

PR #1061, opened 2023-12-11
Ensure reproducibility when using MT Geant4 with Celeritas offloading

### LINKED ISSUE #1023: Ensure reproducibility when using MT Geant4 with Celeritas offloading
In MT Geant4 applications, the scheduling of events is done dynamically. This means events are not guaranteed to be mapped to the same threads over multiple runs. For reproducibility, Geant4 reseeds each worker thread's RNG for each event. We will need to do something similar with our RNG states in Celeritas to make sure simulations with Celeritas offloading are reproducible.

The Geant4 `/run/eventModulo <N> <seedOnce>` UI command can be used to control the chunk size and seed frequency:
- The chunk size `N` defaults to `int(sqrt(num_events / num_threads))` when `N` is 0.
- `seedOnce` controls how often the RNG on each worker thread is seeded. If the default value (0) is used, each thread's RNG will be seeded for each event and the simulation will be reproducible. If `seedOnce` is set to 1, the RNG will only be seeded for the first event on each worker thread.

amandalund:
This reinitializes the Celeritas RNG states at the start of each event when integrating with MT Geant4 so that results are reproducible. Each state is initialized with the same seed and a unique subsequence so that sequences on different threads will
not be correlated. Using a simple problem (simple CMS with 1024 events and 128 primaries per event) I confirmed for both the cuRAND and XORWOW RNGs that the results were not reproducible before and are with this fix and that there was no change in performance.

Closes #1023.


src/accel/LocalTransporter.hh
```diff
@@ -66,6 +66,9 @@ class LocalTransporter
     // Set the event ID
     void SetEventId(int);
 
+    // Set the event ID and reseed the Celeritasa RNG at the start of an event
```
sethrj:
```suggestion
    // Set the event ID and reseed the Celeritas RNG at the start of an event
```

src/accel/LocalTransporter.cc
```diff
@@ -124,6 +128,38 @@ void LocalTransporter::SetEventId(int id)
...
+//---------------------------------------------------------------------------//
+/*!
+ * Set the event ID and reseed the Celeritas RNG at the start of an event.
+ */
+void LocalTransporter::InitializeEvent(int id)
+{
+    CELER_EXPECT(*this);
+    CELER_EXPECT(id >= 0);
+
+    this->SetEventId(id);
```
sethrj: Since ensuring reproducibility is a "bug fix", I think we can change the behavior so we *always* initialize at the start of an event. I like your name change that reflects the initialization cost, so what about having `SetEventId` be a `[[deprecated]]` inline alias to `InitializeEvent` (update the comment to say `to be removed in v1.0`), and replacing this call with the `SetEventId` code?

src/accel/LocalTransporter.cc
```diff
@@ -124,6 +128,38 @@ void LocalTransporter::SetEventId(int id)
...
+                    step_->state());
+            detail::reseed_rng(rng_->device_ref(), state.ref().rng, id);
+        }
+        else
+        {
+            auto const& state = dynamic_cast<CoreState<MemSpace::host> const&>(
+                step_->state());
+            detail::reseed_rng(rng_->host_ref(), state.ref().rng, id);
+        }
+    }
```
sethrj: Let's perhaps add a `reseed` method to the stepper? It'll make this a single line, remove the const casting, and hide the tempating.

src/accel/detail/RngReseed.cu
```diff
@@ -0,0 +1,73 @@
...
+        celeritas::KernelParamCalculator::thread_id().unchecked_get()};
+    if (tid.get() < state.size())
+    {
+        TrackSlotId tsid{tid.unchecked_get()};
+        RngEngine::Initializer_t init;
+#if CELERITAS_CORE_RNG == CELERITAS_CORE_RNG_XORWOW
+        init.seed = params.seed[0];
+#else
+        init.seed = params.seed;
+#endif
```
sethrj: Maybe the RNG engine initializer should have a `seed` that's the same type as `params.seed`?

sethrj (changes requested):
I think we should move the CUDA code to the main celeritas library to live with the RNG or global core state/params. That will also allow us to add a test 😉

src/accel/detail/RngReseed.cu
amandalund: Yeah good point, I can change the initializer seed to an array. Is this still a plan for xorwow?
```
// TODO: 256-bit seed used to generate initial states for the RNGs
// For now, just 4 bytes (same as our existing cuda/hip interface)
Array<uint_t, 1> seed;
```

sethrj: It could be? 😅 it would make sense to do, but yeah I don't see any practical use until we start feeding hashed time stamps from a python front end

src/accel/LocalTransporter.cc
amandalund: Good call.

amandalund:
@sethrj I pushed the changes, but I still need to add a test.

src/accel/LocalTransporter.cc
```diff
@@ -114,14 +115,23 @@ LocalTransporter::LocalTransporter(SetupOptions const& options,
...
     event_id_ = EventId(id);
     track_counter_ = 0;
+
+    if (G4Threading::IsMultithreadedApplication()
+        && !G4MTRunManager::SeedOncePerCommunication())
+    {
+        // Since Geant4 schedules events dynamically, reseed the Celeritas RNGs
+        // using the Geant4 event ID for reproducibility.
+        step_->reseed(id);
+    }
```
sethrj, line 135: Oh, I learned that "reproducibility" for them doesn't mean "running it the same twice gives the same results" like it does for normal people. 😉  It means they want an event-level *restart* capability so that they can run a single event out of a stream of them. So I think we should remove the `if` condition here and *always* reseed the RNG based on the event.

src/celeritas/global/Stepper.cc
```diff
@@ -103,6 +105,27 @@ auto Stepper<M>::operator()(SpanConstPrimary primaries) -> result_type
...
+void Stepper<M>::reseed(size_type event_id)
+{
+    if constexpr (M == MemSpace::device)
+    {
+        reseed_rng(params_->rng()->device_ref(), state_.ref().rng, event_id);
+    }
+    else
+    {
+        reseed_rng(params_->rng()->host_ref(), state_.ref().rng, event_id);
+    }
```
sethrj: From `corecel/data/Ref.hh`:
```suggestion
    reseed_rng(get_ref<M>(*params_->rng()), state_.ref().rng, event_id);
```

src/celeritas/global/Stepper.hh
```diff
@@ -90,6 +90,9 @@ class StepperInterface
     // Transport existing states and these new primaries
     virtual StepperResult operator()(SpanConstPrimary primaries) = 0;
 
+    // Reseed the RNGs at the start of an event for reproducibility
+    virtual void reseed(size_type event_id) = 0;
```
sethrj:
```suggestion
    virtual void reseed(EventId event_id) = 0;
```

sethrj (commented):
Looking good! Just the test now :)

src/accel/LocalTransporter.cc
amandalund: Geant4 has an MT option ("seed once") to turn off this event reproducibility and only seed once per worker thread at the beginning (which they claim may give better performance in certain cases), and that's what this condition checks. That's a good point though, probably in serial mode we always should.

src/celeritas/global/Stepper.cc
amandalund: thanks :)

sethrj (approved):
Nice! Good work @amandalund , this definitely addresses a capability gap that we needed.

PR #1062, opened 2023-12-12
Remove extra constructor templates in device ActionLauncher

esseivaju:
Since #986 `KernelParamCalculator` reads the kernel block size from function attributes instead of passing it as a constructor argument. The overloaded constructors are no longer needed, making the code easier to read.


sethrj (approved):
Thanks!

Issue #253, opened 2021-06-11
Refactor physics to have a single combined energy loss/range "process"


amandalund:
In Geant4, the dE/dx table is created by summing the contribution from each process that provides energy loss for a given particle. The range table is then calculated from this total energy loss. Our current implementation assumes per-process dE/dx and range tables, though what we are actually importing is a combination of process and total dE/dx and range.

sethrj:
It's been a while since we had our discussion, but what I recall is that the only per-process "value grid" we should be storing is cross section; and we should have one "range table" and one "energy loss table" in the `ProcessGroup` (corresponding to the process-integrated values for that single particle type).

sethrj:
This is related to the discussion @amandalund, @whokion, @paulromano and I had last week. As we start implementing MSC, we should consider refactoring the Physics/Process with these considerations:
- A single (optional) runtime "energy loss" process; so eliminate the generic eloss/range tables for discrete processes. Possibly rename the device data to reflect that "discrete processes" are exclusively for end-of-step interactions where exactly one is selected if the step is limited by the (exponential-sampled) physics interaction length.
- Generalize and separate the components of `calc_step_limit` to step limiters: discrete cross section calculation, range (from the energy loss process), and MSC step limiter. As we add hadronics, "decay" will probably be a step limiter as well.
- The `Process` host class is responsible for setting up the on-device data but doesn't necessarily have to be mirrored in on-device data. So it might be OK to have `energy_loss` and `msc` accessors with a setup check that no more than one process  is MSC and no more than one is energy loss.

sethrj:
Sketching out some ideas:

- Make an XsGridType enum for the various grid types (EM grid, generic grid, ...)
- XsGridId for indexing into the various grids

New XsParamsData:
- ids and types to index into them
- reals for dynamic storage

XsView maybe to create XS calculators?

- Delete ValueGridInserter: instead just build stuff in XsParamsData
- Delete ValueGridType, ValueGridArray
- ValueGridBuilder could just be helper functions for constructing in XsParamsData
- Model::micro_xs returns a range of XsGridId?
- Process::step_limits should become macro_xs (and should we have an option for calculating by summing the elemental cross sections?) and build a new xs/XsGridId

sethrj:
[...]

PR #1063, opened 2023-12-12
Remove deprecated code from v0.4

sethrj:
These should have been removed as part of the 0.4 release. I added documentation for future releases to prevent this from happening again, and I've postponed the only external-facing deprecation (mt_logger) till v1.0 . 

Deprecated code removed are:
- `normalize_direction`
- `host` and `device` accessor names from CollectionMirror
- import tables that we'll never use (some of which were removed from geant4@11)
- `Storage` type alias from ORANGE
- `Bytes` (replaced with `std::byte`)

### DIFF
--- a/.github/pull_request_template.release.md
+++ b/.github/pull_request_template.release.md
@@ -11,6 +11,7 @@ Target: upstream/develop (always!)
 - [ ] Tag the develop branch with ``vX.Y.Z-rc.N`` where N starts with 1, and increment for every time you return to this step due to new pull requests.
 - [ ] Run performance regression tests on Summit, Crusher/Frontier, and an additional machine with debug assertions enabled (e.g., Wildstyle).
 - [ ] Update documentation with release notes from all pull requests newly included in the release.
+- [ ] Check for (and delete if found) code marked as "deprecated: to be removed in vX.Y".
 - [ ] Ensure the code documentation builds with as few warnings as possible in the `doc` workflow on the CI.
 
 ## Post-merge checklist
--- a/src/celeritas/ext/detail/GeantProcessImporter.cc
+++ b/src/celeritas/ext/detail/GeantProcessImporter.cc
@@ -224,25 +224,14 @@ void append_table(G4PhysicsTable const* g4table,
     switch (table_type)
     {
         case ImportTableType::dedx:
-        case ImportTableType::dedx_process:
-        case ImportTableType::dedx_subsec:
-        case ImportTableType::dedx_unrestricted:
-        case ImportTableType::ionization_subsec:
             table.x_units = ImportUnits::mev;
             table.y_units = ImportUnits::mev_per_cm;
             break;
-        case ImportTableType::csda_range:
         case ImportTableType::range:
-        case ImportTableType::secondary_range:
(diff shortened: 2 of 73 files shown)


.github/pull_request_template.release.md
```diff
@@ -11,6 +11,7 @@ Target: upstream/develop (always!)
 - [ ] Tag the develop branch with ``vX.Y.Z-rc.N`` where N starts with 1, and increment for every time you return to this step due to new pull requests.
 - [ ] Run performance regression tests on Summit, Crusher/Frontier, and an additional machine with debug assertions enabled (e.g., Wildstyle).
 - [ ] Update documentation with release notes from all pull requests newly included in the release.
+- [ ] Check for (and delete if found) code marked as "deprecated: to be removed in vX.Y".
```
pcanal, line 14: This can be helped by marking the deprecated features by a macro that turns into a Warning/Reminder/Error when the version number reaches the target version number.

pcanal (approved):
LGTM.  Please consider to NOT squash the commits in this special case.  They are all independent and and if they need to be reversed (feature becomes needed again or bisecting) it will be clearer if the removals are still independent commits.

src/celeritas/ext/detail/GeantProcessImporter.cc
```diff
@@ -224,25 +224,14 @@ void append_table(G4PhysicsTable const* g4table,
     switch (table_type)
     {
         case ImportTableType::dedx:
-        case ImportTableType::dedx_process:
```
amandalund, line 227: It's possible we may use the per-process energy loss tables in the future (#253), but we can add it back in if needed.

amandalund (approved):
Looks good, thanks!

src/celeritas/ext/detail/GeantProcessImporter.cc
sethrj: If we do that, I think it'd make more sense to still just call it `dedx` (and omit the `range` at that point) since the data is associated with a single process.

.github/pull_request_template.release.md
sethrj: Good idea... the only potential issue is that we would be unable to tag the develop branch upstream as `next.rc-1` because then the tag would cause the deprecations to trigger.

sethrj:
Remember @pcanal that the PR branches (individual commits) are still accessible after merge:
```console
$ git fetch upstream pull/1063/head:pr-1063
$ git checkout pr-1063
```

pcanal:
< Remember @pcanal that the PR branches (individual commits) are still accessible after merge:

Besides that I might indeed forget, it is also impractical for bisection and weird for reversion (reverting a commit that is not actually on the stream).  As I was hinting, this is a representative case where it would have made more sense to not squash. I.e. this PR is not a feature PR but a mash-up PR doing of several independent things but together solely for the benefit of reducing the number of PR.

sethrj:
As we discussed in person, pooling all deprecations into a single pull request is a matter of choice of granularity. If, in the future, Celeritas removes a feature that could be used by others downstream, it will be worth putting in an individual pull request for that. I think minor/internal deprecations can still be lumped together.

PR #1065, opened 2023-12-14
Save action times on multi-stream CPU celer-sim

sethrj:
When we switched the regression problems to `"merge_events": false` on CPU, we lost the action-granularity timing results. This restores them. To be consistent with current output, the timing is *averaged* over the streams.


app/celer-sim/Runner.cc
```diff
@@ -159,22 +159,24 @@ size_type Runner::num_events() const
 /*!
  * Get the accumulated action times.
  *
- * Action times are only collected by the transporter when running with a
- * single stream.
+ * This is a *mean* value over all streams.
  */
-auto Runner::get_action_times() -> MapStrReal
+auto Runner::get_action_times() -> MapStrDouble
```
pcanal:
```suggestion
auto Runner::get_action_times()  const -> MapStrDouble
```

app/celer-sim/Transporter.cc
```diff
@@ -105,9 +105,6 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries)
     if (store_track_counts_)
     {
         append_track_counts(track_counts);
-    }
-    if (num_streams_ == 1)
-    {
         result.step_times.push_back(get_step_time());
```
amandalund, line 112: Do you think it's useful to have the step times for each thread? Maybe storing the event times would be more useful?

sethrj: Good point... I was thinking `store_track_counts` would be the switch between "minimal output" and "all the output". Event times would be more useful indeed...

app/celer-sim/Runner.cc
sethrj: Ah actually because this access the transporters, which were non-const because they did lazy construction. I've amended with a const-correct accessor.

app/celer-sim/Transporter.cc
sethrj: I've just added a new option to switch off step times (usually we'll want to be looking at these on GPU anyway).

amandalund: There's one more place we'll have to update this below.

sethrj:
I can't reproduce the celer-sim crash locally :(

sethrj:
@amandalund Do you have any idea why this could be segfaulting now? I'm at a loss.

amandalund:
I'll see if I can reproduce it.

sethrj:
Also failed to reproduce on wildstyle/openmp/cuda for debug, release, and reldeb :(

amandalund:
No luck either so far...

amandalund:
Ok, I was able to reproduce it with debug assertions on. In the new transporter accessor in the `Runner`, looks like the transporter isn't allocated:
```console
status: Creating states
status: Celeritas core state initialization complete
status: Warming up
status: Transporting 5 primaries
status: Creating states
status: Creating states
status: Celeritas core state initialization complete
status: Celeritas core state initialization complete
status: Transporting 5 primaries
status: Transporting 5 primaries
/home/alund/celeritas_project/celeritas/app/celer-sim/Transporter.cc:119: error: Exceeded step count of 256: aborting transport loop
/home/alund/celeritas_project/celeritas/app/celer-sim/Transporter.cc:119: error: Exceeded step count of 256: aborting transport loop
/home/alund/celeritas_project/celeritas/app/celer-sim/Transporter.cc:119: error: Exceeded step count of 256: aborting transport loop
/home/alund/celeritas_project/celeritas/app/celer-sim/celer-sim.cc:268: critical: While running input at <stdin>: /home/alund/celeritas_project/celeritas/app/celer-sim/Runner.cc:543:
celeritas: postcondition failed: result
status: Saving output
Failure written to simple-cms-cpu.out.failed.json
fatal: run failed with error 1
```

amandalund:
Are there more threads than events in this test?

amandalund (approved):
Thanks @sethrj!

Issue #1068, opened 2023-12-19
Add G4VG support for ALICE Geant4 geometry components


sethrj:
From @sawenzel:

> For the moment, I seem to be blocked by 2 error messages:
> 
> a) unsupported solid type G4DisplacedSolid
> 
>  b) unsupported type 'G4PVDivision' for physical volume 'ZNTX_PV' (corresponding LV: "ZNTX"@0xaaaabefaf6d0 (ID=106))
> 
> - G4DisplacedSolids are sometimes used inside booleans. Not sure why you don't see this with CMS but
> it could be a speciality of the GDML constructed with ROOT.  We would need to implement the solid in VecGeom first and ensure it works on CUDA.
> 
> - The other issue about G4PVDivision can probably be solved in the geometry converter by simply creating distinct VecGeom physical volumes or maybe there is an option in the ROOT geometry exporter not to use divisions in the output... I'll take a look.

From discussions with the VecGeom team it sounds like we might be able to reduce the displaced solids to regular solids by propagating the displacement into the "placement" of the physical volumes.

For the physical volumes, we should be able to replicate them manually.

sawenzel:
> From discussions with the VecGeom team it sounds like we might be able to reduce the displaced solids to regular solids by propagating the displacement into the "placement" of the physical volumes.

Yes, that's a good idea. In Geant4 a boolean has 2 logical volumes as members whereas in VecGeom there are 2 placed volumes. A logical displaced volume should be straightforwardly convertible.

sawenzel:
@sethrj : After inspection of the conversion code, I believe you should already be doing the right thing. But there might be a wrong assumption that only the second volume could be a displaced volume. In fact, after switching on assertations, I see:

```
.../app/celer-g4/celer-g4.cc:297: critical: While running input_gun.json: celeritas/ext/g4vg/SolidConverter.cc:696:
celeritas: Geant4 error: celer0001 failed:
    precondition failed: !dynamic_cast<G4DisplacedSolid const*>(solids[0])
```
So simply extending the treatment to the first volume might be enough. Shall I provide a PR?

sethrj:
@sawenzel Yes that was an assumption inherited from an earlier version of the vecgeom->Geant4 converter. I actually just found the same thing and am about to submit the PR! Thanks though. If you'd like to add the PV expansion we'd love to include your contribution 😄

sawenzel:
Since you have the code already I would say just go ahead -:)

[...]

PR #1069, opened 2023-12-19
Fix conversion of boolean solids when the "left" solid is displaced

sethrj:
The first iteration of the vecgeom converter assumed (presumably based on G4GDML output?) that only the "right" boolean solid was displaced. This isn't the case for solids exported from ROOT:
```
LogicalVolume [459] "QCLint0x6000029e8b40":
   UnplacedBooleanVol<Union>{fLeft=[102]<[TEMP]@intTrousersC/left*/displaced_QCLint>, fRight=[103]<[TEMP]@intTrousersC/right*/placedB>}
fLeft:  [102] UnplacedCone {rmin1 0.00, rmax1 10.60, rmin2 0.00, rmax2 2.70, dz 44.44, phistart 0.00, deltaphi 6.28}
fRight: [103] UnplacedCone {rmin1 0.00, rmax1 10.60, rmin2 0.00, rmax2 2.70, dz 44.44, phistart 0.00, deltaphi 6.28}
```

See #1068

### DIFF
--- a/src/celeritas/ext/g4vg/SolidConverter.cc
+++ b/src/celeritas/ext/g4vg/SolidConverter.cc
@@ -713,8 +713,12 @@ auto SolidConverter::convert_bool_impl(G4BooleanSolid const& bs)
 
         // Construct name
         std::ostringstream label;
-        label << "[TEMP]@" << bs.GetName() << '/' << lr[i] << '/'
-              << solids[i]->GetName();
+        label << "[TEMP]@" << bs.GetName() << '/' << lr[i];
+        if (trans)
+        {
+            label << '*';
+        }
+        label << '/' << solids[i]->GetName();
 
         // Create temporary LV from converted solid
         auto* temp_lv = new LogicalVolume(label.str().c_str(), converted);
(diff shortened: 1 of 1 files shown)


src/celeritas/ext/g4vg/SolidConverter.cc
```diff
@@ -689,36 +689,40 @@ auto SolidConverter::convert_bool_impl(G4BooleanSolid const& bs)
...
     static Array<char const*, 2> const lr = {{"left", "right"}};
-    Array<Transformation3D const*, 2> transforms
-        = {{&Transformation3D::kIdentity, &inverse}};
     PlacedBoolVolumes result;
 
     for (auto i : range(solids.size()))
     {
-        // Convert solid
-        VUnplacedVolume const* converted = (*this)(*solids[i]);
+        G4VSolid const* solid = bs.GetConstituentSolid(i);
```
amandalund, line 694:
```suggestion
        G4VSolid const* solid = solids[i];
```

amandalund (approved):
Nice, thanks @sethrj!

src/celeritas/ext/g4vg/SolidConverter.cc
sethrj: Whoops, meant to take that array out entirely, thanks! The only reason we preconstructed that is because the "displacement" seemed to be an asymmetric case which apparently it isn't.

PR #1071, opened 2023-12-20
Support Geant4-VecGeom conversion of parameterized volumes

### LINKED ISSUE #1068: Add G4VG support for ALICE Geant4 geometry components

sethrj:
This adds support for parameterized Geant4 volumes. Closes #1068 .

### DIFF
--- a/src/celeritas/ext/g4vg/Converter.cc
+++ b/src/celeritas/ext/g4vg/Converter.cc
@@ -46,6 +46,16 @@ G4LogicalVolume const* get_constituent_lv(G4LogicalVolume const& lv)
 }
 
 //---------------------------------------------------------------------------//
+/*!
+ * Build a VecGeom transform from a Geant4 physical volume.
+ */
+vecgeom::Transformation3D
+build_transform(Transformer const& convert, G4VPhysicalVolume const& g4pv)
+{
+    return convert(g4pv.GetTranslation(), g4pv.GetRotation());
+}
+
+//---------------------------------------------------------------------------//
 //! Add all visited logical volumes to a set.
 struct LVMapVisitor
 {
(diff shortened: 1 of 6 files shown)


src/celeritas/ext/g4vg/Converter.cc
```diff
@@ -159,45 +230,56 @@ auto Converter::build_with_daughters(G4LogicalVolume const* mother_g4lv)
...
-        vecgeom::ReflFactory::Instance().Place(
-            (*this->convert_transform_)(g4pv->GetTranslation(),
-                                        g4pv->GetRotation()),
-            vecgeom::Vector3D<double>{1.0, 1.0, flip_z ? -1.0 : 1.0},
-            g4pv->GetName(),
-            daughter_lv,
-            mother_lv,
-            g4pv->GetCopyNo());
+            // Loop over number of replicas
+            for (auto j : range(div->GetMultiplicity()))
```
sethrj: @sawenzel You used `g4pv->GetReplicationData` but while digging into the replication stuff this morning, it looked like this was the method more frequently used. Is there a subtle difference?

test/celeritas/ext/Vecgeom.test.cc
```diff
@@ -1460,6 +1472,33 @@ TEST_F(SolidsGeantTest, reflected_vol)
...
+    // an ALICE component
+    static char const* const expected_mid_volumes[]
+        = {"World", "ZNENV", "ZNST", "ZNST",  "ZNST", "ZNST", "ZNST",
+           "ZNST",  "ZNST",  "ZNST", "ZNST",  "ZNST", "ZNST", "ZNST",
+           "ZNST",  "ZNST",  "ZNST", "ZNST",  "ZNST", "ZNST", "ZNST",
+           "ZNST",  "ZNST",  "ZNST", "ZNENV", "World"};
+    static real_type const expected_mid_distances[]
+        = {6.38, 0.1,  0.32, 0.32, 0.32, 0.32, 0.32, 0.32, 0.32,
+           0.32, 0.32, 0.32, 0.32, 0.32, 0.32, 0.32, 0.32, 0.32,
+           0.32, 0.32, 0.32, 0.32, 0.32, 0.32, 0.1,  46.38};
```
sethrj, line 1487: These results are from the geant4 navigator; the geometry is a subset of the ZDC input that has the division volumes.

src/celeritas/ext/g4vg/Converter.cc
```diff
@@ -159,45 +230,56 @@ auto Converter::build_with_daughters(G4LogicalVolume const* mother_g4lv)
...
-            flip_z = true;
+            DaughterPlacer place_daughter(convert_daughter,
+                                          *convert_transform_,
+                                          g4pv->GetLogicalVolume(),
+                                          mother_lv);
+
+            // Place daughter, accounting for reflection
+            place_daughter(placed);
         }
+        else if (auto* div = dynamic_cast<G4PVDivision*>(g4pv))
```
sawenzel: looking good to me overall. Maybe one could generalize from G4PVDivision to G4PVReplica ?(although I am no real expert on this).

sawenzel (approved):

src/celeritas/ext/g4vg/Converter.cc
sawenzel: No. I was simply not aware if this leaner API.

src/celeritas/ext/g4vg/Converter.cc
sethrj: Looks like "replica" doesn't necessarily mean "parameterized", but that's a good idea so I've generalized it a bit.

amandalund (approved):
Looks good, thanks @sethrj!

PR #1072, opened 2023-12-22
Automatically calculate number of events in apps

amandalund:
The removes the previously required `max_events` input option from `celer-sim` and `celer-g4`.

### DIFF
--- a/app/celer-g4/GlobalSetup.cc
+++ b/app/celer-g4/GlobalSetup.cc
@@ -153,10 +153,14 @@ void GlobalSetup::ReadInput(std::string const& filename)
         // Apply Celeritas \c SetupOptions commands
         options_->max_num_tracks = input_.num_track_slots;
         options_->max_num_events = [this] {
+            CELER_VALIDATE(input_.primary_options || !input_.event_file.empty(),
+                           << "no event input file nor primary options were "
+                              "specified");
             if (!input_.event_file.empty())
             {
-                return static_cast<size_type>(
-                    HepMC3PrimaryGenerator(input_.event_file).NumEvents());
+                hepmc_gen_ = std::make_shared<HepMC3PrimaryGenerator>(
+                    input_.event_file);
+                return static_cast<size_type>(hepmc_gen_->NumEvents());
             }
             else
             {
--- a/app/celer-g4/GlobalSetup.cc
+++ b/app/celer-g4/GlobalSetup.cc
@@ -19,6 +19,7 @@
 #include "celeritas/ext/RootFileManager.hh"
 #include "celeritas/field/RZMapFieldInput.hh"
 #include "accel/ExceptionConverter.hh"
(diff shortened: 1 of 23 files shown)


app/celer-g4/GlobalSetup.cc
```diff
@@ -151,7 +152,17 @@ void GlobalSetup::ReadInput(std::string const& filename)
 
         // Apply Celeritas \c SetupOptions commands
         options_->max_num_tracks = input_.num_track_slots;
-        options_->max_num_events = input_.max_events;
+        options_->max_num_events = [this] {
+            if (!input_.event_file.empty())
+            {
+                return static_cast<size_type>(
+                    HepMC3PrimaryGenerator(input_.event_file).NumEvents());
```
sethrj: Each time we open the primary generator it reads the whole file (then reopens it)... is there a reasonable way to keep the generator around?

src/celeritas/io/EventReader.hh
```diff
@@ -50,23 +50,31 @@ class EventReader : public EventReaderInterface
...
+    EventReader(std::string const& filename, SPConstParticles particles);
 
     //! Prevent copying and moving
     CELER_DELETE_COPY_MOVE(EventReader);
 
     // Read a single event from the event record
     result_type operator()() final;
 
+    //! Get total number of events
+    size_type num_events() const { return num_events_; }
```
sethrj: Since this function is being added to `RootEventReader` too, I think you can go ahead and make it part of the `EventReaderInterface`?

app/celer-g4/GlobalSetup.cc
amandalund: Maybe we can store it in the `GlobalSetup`?

amandalund:
@sethrj do you know what I'm missing?
```
/home/runner/work/celeritas/celeritas/src/celeritas/io/EventReader.cc:46:26: error: no member named 'skip' in 'HepMC3::Reader'
            temp_reader->skip(1);
```

sethrj:
@amandalund I think the problem is that the HepMC3 available from debian (looks like [3.1.2](https://packages.ubuntu.com/focal/libhepmc3-dev), but I'm not sure why the cmake config isn't echoing the version in the CI) is old, and previously only the `accel` library used the `skip` function (which looks like it was first released in [3.2.0]), and it isn't compiled because the "fast" VM doesn't have Geant4. So I guess we could:
1. Require HepMC3 3.2.0 and newer, and disable it on the CI (not great because then we can't run the `app/` unit tests)
2. Add include guards using `#define HEPMC3_VERSION_CODE 3001002` and replace skip with `readevent` where needed (?)
3. Install a newer hepmc3 as part of the CI?

amandalund:
Ahh, thank you @sethrj! Maybe I'll add the include guards for now, and if we want to require a newer HepMC3 version we can revert that later.

sethrj (approved):
Looks great. Thanks @amandalund !

PR #1073, opened 2023-12-28
Use environment variables to control OpenMP behavior in `celer-sim`

amandalund:
This removes the hardcoded 'num_threads()" clause and `omp_set_max_active_levels()` call from celer-sim. It also changes the calculated number of streams when `OMP_NUM_THREADS` is unset from 1 to the (implementation defined) default (typically the maximum available).

I noticed with MT celer-sim on the CPU with `merge_events` off the performance was much worse when `OMP_NUM_THREADS` was not set than when it was explicitly set to 1. It looks like a parallel region with one thread is _not_ an active parallel level (even though it spawns a thread), so when `OMP_NUM_THREADS` was not set the inner loops over tracks were the first active parallel levels and were using the default (maximum available) number of threads -- i.e., we were actually parallelizing the track loops rather than the event loop. This change should give more reasonable/expected behavior and give us more control through `OMP_NUM_THREADS` and `OMP_MAX_ACTIVE_LEVELS` (though I'm still not entirely sure why the performance was _so_ much worse previously when the environment variable was not set).

### DIFF
--- a/app/celer-sim/Runner.cc
+++ b/app/celer-sim/Runner.cc
@@ -14,6 +14,10 @@
 #include <utility>
 #include <vector>
 
+#ifdef _OPENMP
+#    include <omp.h>
+#endif
+
 #include "corecel/cont/Span.hh"
 #include "corecel/io/Logger.hh"
 #include "corecel/io/OutputRegistry.hh"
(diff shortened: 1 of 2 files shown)


sethrj (approved):
Sounds like a nice improvement! The one thing that might be missing is that it would be really good to save the number of threads to the output somewhere... is that a possibility?

amandalund:
We do save the number of streams to the `Runner` output... do you think we want something in addition, like the value of the OpenMP environment variables?

sethrj:
I guess I'd just like to confirm that the number of streams that's *saved* is the value after updating here, rather than just the original input value.

amandalund:
Yep, it's the value from `calc_num_streams()` (so the smaller of `num_threads` and `num_events`).

PR #1076, opened 2024-01-02
Add option to use CLHEP and SI unit systems instead of Gaussian CGS

sethrj:
This was my winter break "for fun" project, following up on a quick discussion of unit system with J Apostolakis at CERN in December. It will let us change the internal unit system of Celeritas to comply with external packages without multiplicative factors everywhere. Many unit tests fail for the non-CGS option; fixes to the tests and core, as well as support for unit conversion in the import data will be added in a follow-on request.


src/celeritas/Quantities.hh
```diff
@@ -4,116 +4,50 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file celeritas/Quantities.hh
-//! \brief Unit struct definitions for use with the Quantity class
+//! \brief Derivative unit classes and annotated Quantity values
 //---------------------------------------------------------------------------//
 #pragma once
 
+#include "celeritas_config.h"
```
pcanal: This seems to not be 'directly' used in this header file. (on the other hand `"corecel/Macros.hh"` is).

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,265 @@
...
+    static char const* label() { return "c"; }
+};
+
+//! "Natural units" for energy
+struct Mev
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
+        return units::megaelectronvolt;
```
pcanal, line 40: Why the `#if` here and not in `Units.hh`?  I.e. why not define `units::megaelectronvolt` for the non-CLHEP cases?

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,265 @@
...
+    }
+    static char const* label() { return "MeV/c^2"; }
+};
+
+//! Magnitude of electron charge (positive sign)
+struct EElectron
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
```
pcanal: Why the #if here and not in `Units.hh`? I.e. why not define units::e_electron for the non-CLHEP cases?

src/celeritas/Quantities.hh
```diff
@@ -4,116 +4,50 @@
...
-        return constants::c_light;
-    }
-};
+//!@{
+//! \name Derivative units
 
-//! Unit for precise representation of particle charges
-struct EElectron
+//! "Natural units" for momentum
+struct MevPerC : UnitDivide<Mev, CLight>
```
pcanal: What is the advantage of having this 3 units types defined in `Quantities.hh` rather than with the other unit types (i.e. in `UnitTypes.hh`)?

src/celeritas/UnitTypes.hh
sethrj: In all unit systems that use "macro scale" quantities (CGS and SI), `megaelectronvolt` requires a physical constant (e_electron) that's defined in `Constants.hh` (which is dependent on `Units.hh`). CLHEP is the weird exception because it indirectly defines charge and mass based on the value of `e_electron`.

src/celeritas/UnitTypes.hh
sethrj: The comment on the next line was an attempt to explain it but I guess it could use more elaboration. The `units::e_electron` here should be *mathematically* equal to `constants::e_electron`, but might not be *numerically*.  Since we define `units::C` as `1/e` and then define `constants::e` as `e * units::C`, the compiler is doing `(x * (1/x))` which could be one or two ULP off from 1 (i.e. `(x / x)` which is an identity except with nan/zero/inf).

The "constants" should I think for clarity remain constant and not have exceptions based on the unit. The "unit types" here ensures that the unit values for CLHEP are exactly unity.

src/celeritas/Quantities.hh
sethrj: The only reason was that `UnitDivide` is defined in `Quantity.hh` which isn't included by `UnitTypes`, but maybe I should either remove those helpers or move them somewhere else...

src/celeritas/UnitTypes.hh
pcanal: I am really confused :( ... the else part really seems to read Mega (1e6) Electron Volt, so I don't see why it is not in unit as `megaelectronvolt` ...

src/celeritas/UnitTypes.hh
pcanal: > The "constants" should I think for clarity remain constant

Do you mean that their numerical value are the same independently of the configuration or simply that their are marked `constexpr`?

sethrj: The first :)

src/celeritas/Quantities.hh
pcanal: > move them somewhere else...

+1

src/celeritas/UnitTypes.hh
pcanal: Then we ought to find a different spelling for the namespace.  For me (and likely others), it says the second :)

src/celeritas/UnitTypes.hh
sethrj:
```c++
#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
        return units::megaelectronvolt;
#else
        return real_type(1e6) * constants::e_electron * units::volt;
#endif
```

CLHEP is the only system that defines MeV as a "unit" (value of 1). The other unit systems require a fundamental constant (which *can't* be included by Units.hh).

src/celeritas/UnitTypes.hh
pcanal: for example `invariants`

stognini:
This looks pretty great. All the physics model tests pass, so it seems correct. A lot of our data in those models is hardcoded, But as long as XS and material data is converted accordingly, I guess we should have consistent results...

So this seems like we kinda convert between CLHEP / CGS / SI, but we keep CGS in our stepping loop mostly. Am I thinking correctly?

sethrj:
We already do a little conversion between CLHEP/CGS/SI (SI is used for all our fundamental constant definitions) but the "over the break" goal (not a priority for our real work) was to be able to use CLHEP inside the stepping loop for potentially tighter Geant4 integration in the future.

src/celeritas/UnitTypes.hh
sethrj: in celeritas/Constants.hh in the constant namespace (wikipedia):
> A physical constant, sometimes fundamental physical constant or universal constant, is a [physical quantity](https://en.wikipedia.org/wiki/Physical_quantity) that is generally believed to be both universal in nature and have [constant](https://en.wikipedia.org/wiki/Constant_(mathematics)) value in time. It is distinct from a [mathematical constant](https://en.wikipedia.org/wiki/Mathematical_constant), which has a fixed numerical value, but does not directly involve any physical measurement.

in corecel/Constants.hh in the constant namespace (wikipedia):
>  A mathematical constant is a key [number](https://en.wikipedia.org/wiki/Number) whose value is fixed by an unambiguous definition, often referred to by a special symbol (e.g., an [alphabet letter](https://en.wikipedia.org/wiki/Letter_(alphabet))), or by mathematicians' names to facilitate using it across multiple [mathematical problems](https://en.wikipedia.org/wiki/Mathematical_problem).[[1]](https://en.wikipedia.org/wiki/Mathematical_constant#cite_note-1) Constants arise in many areas of [mathematics](https://en.wikipedia.org/wiki/Mathematics), with constants such as [e](https://en.wikipedia.org/wiki/E_(mathematical_constant)) and [π](https://en.wikipedia.org/wiki/Pi) occurring in such diverse contexts as [geometry](https://en.wikipedia.org/wiki/Geometry), [number theory](https://en.wikipedia.org/wiki/Number_theory), [statistics](https://en.wikipedia.org/wiki/Statistics), and [calculus](https://en.wikipedia.org/wiki/Calculus).

Am I missing something???

sethrj: And in `celeritas/Units.hh` in the `units` namespace:
> A system of units of measurement, also known as a system of units or system of measurement, is a collection of [units of measurement](https://en.wikipedia.org/wiki/Units_of_measurement) and rules relating them to each other. Systems of measurement have historically been important, regulated and defined for the purposes of [science](https://en.wikipedia.org/wiki/Science) and [commerce](https://en.wikipedia.org/wiki/Commerce). Instances in use include the [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units) or SI (the modern form of the [metric system](https://en.wikipedia.org/wiki/Metric_system)), the [British imperial system](https://en.wikipedia.org/wiki/British_imperial_system), and the [United States customary system](https://en.wikipedia.org/wiki/United_States_customary_system).

pcanal: It is unqualified so maybe it [refers to a collection of:](https://en.cppreference.com/w/cpp/language/cv)
```
Appear in any type specifier, including decl-specifier-seq of [declaration grammar](https://en.cppreference.com/w/cpp/language/declarations), to specify constness or volatility of the object being declared or of the type being named.

    const - defines that the type is constant.
```

Taking from the quotes above, maybe we mean `physical_constant` (or `math_constant`) :)

pcanal: More importantly it does not indicates (for better or worse) whether the `constant` numerical value are configuration dependent (here depending on the units selected) or not.  Or maybe it does not matter.

sethrj: They're always constant and in the native unit system.

src/celeritas/io/ImportPhysicsTable.hh
```diff
@@ -41,17 +41,25 @@ enum class ImportTableType
...
-    cm_inv,  //!< Macroscopic xs [1/cm]
-    cm_mev_inv,  //!< Macroscopic xs divided by energy [1/cm-MeV]
-    mev_2_per_cm,  //!< Macroscopic xs with energy^2 factored in [MeV^2/cm]
-    cm_2,  //!< Microscopic cross section [cm^2]
+    mev_per_len,  //!< Energy loss [MeV/length]
+    mev_per_cm = mev_per_len,  //!< Deprecated
+    len,  //!< Range [length]
+    cm = len,  //!< Deprecated
+    len_inv,  //!< Macroscopic xs [1/length]
+    cm_inv = len_inv,  //!< Macroscopic xs [1/cm]
```
amandalund:
```suggestion
    cm_inv = len_inv,  //!< Deprecated
```

src/celeritas/mat/MaterialView.hh
```diff
@@ -27,6 +27,9 @@ namespace celeritas
  * The \c get_element_density and \c element_view helper functions can be used
  * to calculate elemental densities and properties.
  *
+ * \note Units listed are for the native Celeritas [CGS] system. Quantities
+ * will differ if using CLHEP.
```
amandalund, line 31: Do you plan to update them (since you have elsewhere)?

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,299 @@
...
+    {
+        return units::nanosecond;
+    }
+    static char const* label() { return "ns"; }
+};
+
+//! Unit mass in CLHEP system
+struct ClhepUnitMass
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
```
amandalund, line 209: It looks this and `ClhepUnitBField::value()` evaluate to 1 in every unit system... is that right?

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,299 @@
...
+    {
+        // Quantities before 1e6 are unity in CLHEP system
+        constexpr auto kg_clhep = (Mev::value() / EElectron::value()
+                                   * (nanosecond * nanosecond)
+                                   / (millimeter * millimeter))
+                                  * (1e6 * coulomb);
+        return kg_clhep / kilogram;
+    }
+    static char const* label() { return "m_clhep"; }
```
amandalund: Maybe "mass_clhep" or something else so as not to confuse with meter?

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,299 @@
...
+//! Unit magnetic flux density in CLHEP system
+struct ClhepUnitBField
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+        constexpr auto tesla_clhep = Mev::value() / EElectron::value() * second
+                                     / (millimeter * millimeter) * 1e-12;
+        return tesla_clhep / tesla;
+    }
+    static char const* label() { return "m_clhep"; }
```
amandalund: Copy-paste error.

amandalund (commented):
Nice addition @sethrj! To be clear, we don't expect simulation results in the other unit systems to be correct currently, that will come in the future PR?

src/celeritas/UnitTypes.hh
pcanal: > as a "unit" (value of 1)....Units.hh

Humm .. another ambiguous name it seems ie. it is `UnitValues.hh` but it could have been `UnitSystem.hh`

src/celeritas/UnitTypes.hh
pcanal: Given this:
(a)
```
celeritas_setup_option(CELERITAS_UNITS CGS)
celeritas_setup_option(CELERITAS_UNITS SI)
celeritas_setup_option(CELERITAS_UNITS CLHEP)
celeritas_define_options(CELERITAS_UNITS
  "Native unit system for Celeritas")
 ```

(b)
 > They're always constant and in the native unit system.
 
(c)
> Do you mean that their numerical value are the same independently of the configuration or ...
> The first :)

Either the meaning of 'native' changes between (a) and (b) or (c) is incorrect (if the constant are expressed in native unit system and the native unit system change with the configuration ( (a) says so) then the numerical value changes with the configuration.

Comment (sethrj) [reply src/celeritas/UnitTypes.hh]: I see what you mean; the numerical value of `c_light` changes depending on the compile option for choosing the native unit system. They are always consistent with the "native" unit system, and they are fundamental physical constants.

Comment (pcanal) [reply src/celeritas/UnitTypes.hh]: Alright. So `celeritas::constants` currently contains universal constant both mathematical (and unit-less) (eg. `pi`) and physical (and with units) (eg. `c_light`).   A bit awkward but fine.  Another awkward-ness is that `orange/Constants.hh` declares things (somewhat unexpectedly) in the namespace `celeritas::constants`.

File: src/celeritas/UnitTypes.hh [hunk 10]
```diff
@@ -0,0 +1,291 @@
...
+};
+
+//! Unit mass in CLHEP system
+struct ClhepUnitMass
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+        // Note: order matters in CHLEP system to result in exactly 1 with
+        // floating point arithmetic
+        return constants::e_electron * (kilogram / coulomb) * 1e-6;
```
pcanal: if get 1.0 is important wouldn't the following be more resilient to change in the CPU ieee implementations?
```suggestion
        if constexpr (CELERITAS_UNITS == CELERITAS_UNITS_CLHEP)
           return 1.0 ...;
        else 
           return constants::e_electron * (kilogram / coulomb) * 1e-6;
```

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,291 @@
...
+    }
+    static char const* label() { return "c"; }
+};
+
+//! "Natural units" for energy
+struct Mev
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
```
pcanal, line 39: (Opening a new conversation to ask more clearly a question that was implicit in https://github.com/celeritas-project/celeritas/pull/1076/files#r1439655776:

Should the configuration depend code (i.e. `if constexpr` or `#if`) but only in `Units.hh` or also in `UnitsTypes.hh` (it feels that we could make it to be 'jiust' `Units.hh`).

src/celeritas/UnitTypes.hh
sethrj: Good catch, thanks. I'm dumb.

src/celeritas/UnitTypes.hh
sethrj: Yep, that's probably smarter.

src/celeritas/UnitTypes.hh
sethrj: It can't be *just* `Units.hh` because the non-CLHEP units require universal constants (electron charge) to compute this value.

amandalund (approved):

src/celeritas/UnitTypes.hh
```diff
@@ -0,0 +1,298 @@
...
+
+//! Magnitude of electron charge (positive sign)
+struct EElectron
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
+        return units::e_electron;  // Mathematically equal to constants
+#else
+        return constants::e_electron;
```
pcanal, line 65: (starting a new conversation to focus on a single point, see https://github.com/celeritas-project/celeritas/pull/1076/files#r1439657682)

> The units::e_electron here should be mathematically equal to constants::e_electron, but might not be numerically.

I think this is wrong and opens up to subtle inconsistencies.  In a unit system where `e_electron` is 1 there is 3 ways to get that value (and 2 other-wise)
* `units::e_electron`
* `constants::e_electron`
* `EElectron::value`
but if the above is true there is no way to guarantees the value are the same and unless we have an automated way to enforce only one is used (I guess given its existence it would be `EElectron::value`, we will eventually have code that use directly or indirectly all 3.

In other in a unit system that has decided that `e_electron` is 1, I would expect that `constants::e_electron` is that value exactly (since the physics constant in `constants` are expressed in the native unit system.

pcanal (changes requested):
See https://github.com/celeritas-project/celeritas/pull/1076/files#r1441172258

src/celeritas/UnitTypes.hh
sethrj: Fair enough. I've made a conditional expression in `Constants.hh` to guarantee both are exactly unity in CLHEP.

src/celeritas/UnitTypes.hh
sethrj: Yeah the ORANGE constants have a "todo: move to corecel" there. I think it's OK for both the math and physical constants to coexist. In SI or (a given derivative unit system) the main physical constants have an exactly specified numerical value, and the remaining (measured) physical constants have some fuzziness on them.

pcanal (approved):
Thanks,

PR #1077, opened 2024-01-02
Add specialized "permutation" transform

sethrj:
Geant4 defines its "solids" along the Z axis, and SCALE allows some shapes (e.g. cylinders) to be constructed around other axes. Ninety-degree rotations (and in the future for Geant4 geometry, reflections) are also common enough to make a special case for transformations, saving a lot of data and flops.

This PR provides basic functionality for the specialized rotation/reflection matrix, aka a signed permutation matrix.
- An integer-based "quarter turn" quantity allows construction of the permutation matrix the same way we do rotation matrices (using integer specializations of sin/cos)
- A simplification of the the rotation matrix construction while doing that
- A transformation-style class for signed permutation.

The signed permutation class is implemented by compressing the destination axes and their signs into a 9-bit unsigned integer, which can be represented exactly as a real type and therefore conveniently integrated into the existing transform storage.

I'm not sure what priority to put further development but I wanted to get at least this bit of the framework in conceptually as I continue with the shape/solid building. Full functionality would require:
- Adding it as a variant transform
- Combine with a translation (signed permutation for rotation/reflection, plus a translation)
- Transform simplification (demoting a near-permutation matrix to a sign swap)
- Transform transformation (applying to a second permutation, permuting a translation/transformation, etc)
- Surface transformation (could shortcut by promoting to a full transform and applying)


sethrj:
Ping @elliottbiondo

src/orange/transform/SignedPermutation.hh
```diff
@@ -0,0 +1,199 @@
...
+ * \endverbatim
+ * This allows the "rotate up" to simply copy one value at a time into a new
+ * position, and optionally flip the sign of the result.
+ */
+class SignedPermutation
+{
+  public:
+    enum Sign : char
+    {
+        pos = '+',
```
elliottbiondo: In some places in this class, pos means positive, whereas in other places it means position. This actually confused me when looking at the constructor in the .cc. Consider revising.

src/orange/transform/SignedPermutation.hh
```diff
@@ -0,0 +1,199 @@
...
+    [[nodiscard]] inline CELER_FUNCTION Real3 rotate_up(Real3 const& dir) const;
+
+    // Rotate from parent to daughter
+    [[nodiscard]] inline CELER_FUNCTION Real3
+    rotate_down(Real3 const& parent_dir) const;
+
+  private:
+    // At least 16 bits: more than enough to round trip through a float
+    using UIntT = short unsigned int;
+    static CELER_CONSTEXPR_FUNCTION UIntT max_value() { return (1 << 9) - 1; }
```
elliottbiondo: A comment to mention that this is used for bounds checking

test/orange/transform/SignedPermutation.test.cc
```diff
@@ -0,0 +1,106 @@
...
+//---------------------------------------------------------------------------//
+}  // namespace
+
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+/*!
+ * Test harness for signed permutation.
+ *
+ * Note that the rotation matrix is a quarter turn around Z, then a hal
```
elliottbiondo: Complete comment

elliottbiondo (changes requested):
In terms of the name of the class, I don't understand the "permutation" part. What about something like `OrthogonalTransform`?

src/orange/transform/SignedPermutation.hh
sethrj: Good point.

test/orange/transform/SignedPermutation.test.cc
sethrj: 🐿️

sethrj:
Thanks @elliottbiondo . Regarding the name, a "signed permutation matrix" is the specialized rotation matrix that this implements. I thought I put that in the documentation but apparently not. See https://en.wikipedia.org/wiki/Permutation_matrix .

src/orange/transform/SignedPermutation.hh
sethrj: I've renamed to psign/msign, but I think it might just be better to hard-code `'+'` and `'-'` everywhere instead... let me know if you have an opinion. I generally try to avoid hardcoded strings or magic characters, but this might be a good exception.

elliottbiondo: I agree that hardcoded is better in this case, as you decided.

elliottbiondo (approved):

PR #1080, opened 2024-01-04
Add Cerenkov distribution and generator

amandalund:
This adds the initial functionality for generating Cerenkov photons:
- Cerenkov "generator" to sample optical photons
- Cerenkov distribution input data for the generator
- Calculator to determine the mean number of Cerenkov photons produced per unit length
- Calculation of Cerenkov angle integrals
- `OpticalPrimary` data used to initialize the photon track
- Placeholder struct for the optical property data (currently only index of refraction)


src/celeritas/optical/PhotonPrimary.hh
```diff
@@ -0,0 +1,30 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Optical photon data used to initialize a photon track state.
+ */
+struct PhotonPrimary
+{
+    units::MevEnergy energy;
+    Real3 position{0, 0, 0};
+    Real3 direction{0, 0, 0};
```
whokion:
```suggestion
    Real3 direction{0, 0, 0};
    Real3 polarization{0, 0, 0};
```
@stognini was right that the polarization is needed (for optical photon propagation/interaction).

src/celeritas/optical/PhotonPrimary.hh
```diff
@@ -0,0 +1,30 @@
...
+    real_type time{};
+    VolumeId volume{};
```
whokion: Not sure whether volumeId is need here.  Instead, may add PhotonId (i.e., TrackId) and StatusId (alive, absorbed, ... ) or TypeId (i.e., cerenkov, remission, ...) later as need.

src/celeritas/optical/PhotonPrimary.hh
```diff
@@ -0,0 +1,30 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2020-2023 UT-Battelle, LLC, and other Celeritas developers.
```
whokion:
```suggestion
// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
```

src/celeritas/optical/PhotonPrimary.hh
```diff
@@ -0,0 +1,30 @@
...
+#include "orange/Types.hh"
+#include "celeritas/Types.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Optical photon data used to initialize a photon track state.
+ */
+struct PhotonPrimary
```
whokion: Just for a comment:  The direct use of `Photon` may somewhat ambiguous now. We may need to clearly distinguish from optical photon from particle photon.

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+            = Collection<real_type, Ownership::const_reference, MemSpace::native>;
+
+        GenericGridData const& data;
+        Values const& reals;
+    };
+
+  public:
+    // Construct from optical properties and distribution parameters
+    inline CELER_FUNCTION
+    CerenkovGenerator(RefractiveIndex const& refractive_index,
```
whokion: Not sure whether the index of refraction should be an input here.  Instead, should it be a part of `CerenkovDistribution` or be accessed through MaterialId (or index of MaterialPropertyId) so that the same generator can be used for different indices of refraction, which may be a case that the distribution is collected and stored in a buffer which will be looped over to generate optical photons all together.

src/celeritas/optical/CerenkovDistribution.hh
```diff
@@ -0,0 +1,50 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Pre- and post-step data for sampling Cerenkov photons.
+ */
+struct CerenkovStepData
+{
+    real_type mean_num_photons;  //!< Average photons produced per unit length
```
whokion: can this be calculate later in the generator locally as we pass velocity assuming that we the index of refraction is available inside the generator?

whokion (commented):
Looks very good!  One general comment: Can a simple unit test be added?

src/celeritas/optical/PhotonPrimary.hh
amandalund: I'll remove this and add polarization, and we can add more later as needed.

src/celeritas/optical/PhotonPrimary.hh
amandalund: Maybe we want to call it `OpticalPhotonPrimary`?

src/celeritas/optical/CerenkovDistribution.hh
amandalund: Good point, since this is not used to sample the number of photons there's no reason we can't calculate it in the generator. I'll add a helper implementing `G4Cerenkov::GetAverageNumberOfPhotons()`, which will also be used by the "pre-generator".

src/celeritas/optical/CerenkovGenerator.hh
amandalund: Yeah I agree, this is just a placeholder since we haven't implemented the optical properties data structures yet.  I imagine we would pass in the optical properties as an argument and use the material ID in the Cerenkov distribution to get the correct refractive index data.

src/celeritas/optical/PhotonPrimary.hh
whokion: Either `OpticalPhotonPrimary` or just `OpticalPhoton` should be good.  OpticalPhotonPrimary may be better as `OpticalPhoton` may stand for the particle definition for optical photons.

src/celeritas/optical/CerenkovGenerator.hh
stognini: Are we going to have one long-lived object that will be re-used for any material, or will we construct one generator every time we need to sample the photons within that material? If the latter, a refractive index in the constructor is already fine, and the `RefractiveIndex` data will just come from an `OpticalParamsView` or directly from `MaterialView`.

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+    GenericXsCalculator calc_refractive_index(refractive_index_.data,
+                                              refractive_index_.reals);
+    real_type inv_beta = 2 * constants::c_light
+                         / (dist_.pre.velocity + dist_.post.velocity);
+    real_type cos_max = inv_beta / calc_refractive_index(e_max);
+    real_type sin_max_sq = (1 - cos_max) * (1 + cos_max);
+
+    // Calculate changes over the step
+    Real3 delta_pos = dist_.post.pos;
+    for (int j = 0; j < 3; ++j)
```
stognini:
```suggestion
    for (auto j : celeritas::range(3))
```

src/celeritas/optical/CerenkovGenerator.hh
amandalund: Right now it's set up to be one generator per distribution, so the latter (but we can easily change that later if needed). So yes, either retrieving the optical data for the material inside the generator or passing it in in the constructor will work.

src/celeritas/optical/PhotonPrimary.hh
sethrj: Ideally the volume ID can be used to accelerate the geometry initialization for the track, and it could resolve ambiguities if emitting tracks near a boundary.

src/celeritas/optical/PhotonPrimary.hh
sethrj: Oh yeah, time to push an update :)

src/celeritas/optical/PhotonPrimary.hh
sethrj: I was thinking something similar after our discussions but worry that `OpticalPhotonPrimary` is too unwieldy. Perhaps we can abbreviate `OptPhoPrimary` or is that ugly? Maybe `OpticalPrimary`? We're going to have a lot of stuff in `celeritas/optical` that otherwise will get prefixes.

src/celeritas/optical/CerenkovDistribution.hh
```diff
@@ -0,0 +1,50 @@
...
+    real_type velocity;
+    Real3 pos;
```
sethrj:
```suggestion
    real_type mean_num_photons{};  //!< Average photons produced per unit length
    real_type velocity{};
    Real3 pos{};
```
for safety

src/celeritas/optical/CerenkovDistribution.hh
```diff
@@ -0,0 +1,50 @@
...
+ * Input data for sampling Cerenkov photons.
+ */
+struct CerenkovDistribution
+{
+    size_type num_photons;  //!< Sampled number of photons to generate
+    real_type time;  //!< Pre-step time
+    real_type step_length;
+    MaterialId material{};
+    CerenkovStepData pre;
+    CerenkovStepData post;
```
sethrj:
```suggestion
    size_type num_photons{};  //!< Sampled number of photons to generate
    real_type time{};  //!< Pre-step time
    real_type step_length{};
    MaterialId material{};
    CerenkovStepData pre;
    CerenkovStepData post;
```

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+    {
+        delta_pos[j] -= dist_.pre.pos[j];
+    }
+    real_type delta_num_photons = dist_.post.mean_num_photons
+                                  - dist_.pre.mean_num_photons;
+    real_type delta_velocity = dist_.post.velocity - dist_.pre.velocity;
+
```
sethrj: I think everything above this line should be in the constructor since it's independent of the RNG up to here

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+        // Sample energy and direction
+        real_type energy;
+        real_type cos_theta;
+        UniformRealDistribution<real_type> sample_energy(e_min, e_max);
+        do
+        {
+            energy = sample_energy(rng);
+            cos_theta = inv_beta / calc_refractive_index(energy);
+        } while (!BernoulliDistribution(
+            sin_max_sq / ((1 - cos_theta) * (1 + cos_theta)))(rng));
```
sethrj: I've seen the `(1 - x) * (1 + x)` in a couple places in geant4 code (which has been propagated into celeritas); maybe we should add an algorithm for that (presumably it's to reduce roundoff compared to `1 - x^2`).

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+            = u * dist_.step_length
+              / (dist_.pre.velocity + u * real_type(0.5) * delta_velocity);
+        photons_[i].time = dist_.time + delta_time;
+
+        // Sample position
+        photons_[i].position = dist_.pre.pos;
+        for (int j = 0; j < 3; ++j)
+        {
+            photons_[i].position[j] += u * delta_pos[j];
+        }
```
sethrj: You should be able to use `axpy` here

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,158 @@
...
+                         / (dist_.pre.velocity + dist_.post.velocity);
+    real_type cos_max = inv_beta / calc_refractive_index(e_max);
+    real_type sin_max_sq = (1 - cos_max) * (1 + cos_max);
+
+    // Calculate changes over the step
+    Real3 delta_pos = dist_.post.pos;
+    for (int j = 0; j < 3; ++j)
+    {
+        delta_pos[j] -= dist_.pre.pos[j];
+    }
```
sethrj: I think you can include `ArrayOperators` and just do `delta_pos -= dist_.pre.pos`

sethrj (commented):
Looking great!

src/celeritas/optical/PhotonPrimary.hh
amandalund: I like `OpticalPrimary`.

src/celeritas/optical/CerenkovGenerator.hh
amandalund: Ah, I was looking for that... thanks!

whokion:
@amandalund As I am prototyping scintillation distribution and generator, I find that there are common data between Cerenkov and Scintillation, so I propose something like
```
/*!                                                                             
 * Pre- and post-step data for sampling optical photons.                        
 */
struct OpticalStepData
{
  real_type velocity{};
  Real3 pos{};
};

/*!                                                                             
 * Input data for sampling optical photons.                                     
 */
struct OpticalDistribution
{
  size_type num_photons{};  //!< Sampled number of photons to generate          
  real_type time{};  //!< Pre-step time                                         
  real_type step_length{}; //!< Step length                                     
  OpticalStepData pre;
  OptocalStepData post;

  {
        return num_photons > 0 && step_length > 0
          && max(pre.mean_num_photons, post.mean_num_photons) > 0;
  }
};
```
and then add Cerenkov- or Scintillation-specific data in `CerenkovDistribution(.hh)` and `ScintillationDistribution(.hh)`, something like,
```
/*!
 * Input data for sampling Cerenkov photons.
 */
struct CerenkovDistribution
{
    OpticalDistribution dist; //<! Common data for sampling optical photons
    CerenkovData data; //<! Data (material properties) for sampling Cerenkov photons
   
    {
        return dist && data;
    }
};
```

amandalund:
Thanks @whokion! Good to know the input data is very similar between the two: I'll rename the distribution.

whokion:
Thanks! One down side is that there are too many layers of structs now, cerenkov->optical->step (sorry about that^^), which may be refactored/restructured later.

sethrj:
I think we also want to avoid calling them "distributions" since we reserve that for random number distributions (which return a value using a PRNG engine).

whokion:
Yes, "Distribution" is a bit miss-representing these data, how about (OpticalStepData+OpticalDistribtuion) -> OpticalStepData and CerenkovDistribution -> CenrenkovStepData (just two levels)?

amandalund:
Yeah I was thinking the same, `Distribution` isn't the right name... other options: `*DistributionData`, `*GeneratorInput`?

@whokion it might be easier to first sketch out the "distributions" independently, even though there will be a lot of overlap, and then as a follow up we can refactor them to reduce the duplicate code without adding too many indirections.

amandalund:
I pushed some updates, but it's still very much a WIP at this point, and I still have more to add (including a test).

src/celeritas/Types.hh
```diff
@@ -66,6 +66,9 @@ using ElementComponentId = OpaqueId<struct MatElementComponent>;
 //! Opaque index to one isotopic component datum in a particular element
 using IsotopeComponentId = OpaqueId<struct ElIsotopeComponent>;
 
+//! Opaque index to a material with optical properties
+using OpticalMaterialId = OpaqueId<struct OpticalMaterial_>;
```
sethrj, line 70: How often do different materials share optical properties? (Our current "material IDs" really should be "physics material IDs" since it incorporates the range cutoff info like "material cut couple" in Geant4. If this is really just the pure material IDs maybe it's time to rename our material ID and use this instead.)

src/celeritas/optical/CerenkovData.hh
```diff
@@ -0,0 +1,59 @@
...
+        CELER_EXPECT(other);
+        reals = other.reals;
+        angle_integral = other.angle_integral;
+        return *this;
+    }
+};
+
+using CerenkovDeviceRef = DeviceCRef<CerenkovData>;
+using CerenkovHostRef = HostCRef<CerenkovData>;
+using CerenkovRef = NativeCRef<CerenkovData>;
```
sethrj: The typedefs in the other interaction classes predate the `DeviceCRef<T>` typedefs: since those are pretty readable and easy to write, I think we can omit these.

src/celeritas/optical/CerenkovDndxCalculator.hh
```diff
@@ -0,0 +1,152 @@
...
+ */
+class CerenkovDndxCalculator
+{
+  public:
+    // Construct from optical properties and Cerenkov angle integrals
+    inline CELER_FUNCTION
+    CerenkovDndxCalculator(OpticalPropertyRef const& properties,
+                           CerenkovRef const& shared,
+                           OpticalMaterialId material,
+                           real_type charge);
```
sethrj: This should be a `units::ElementaryCharge` quantity I think?

src/celeritas/optical/CerenkovDndxCalculator.hh
```diff
@@ -0,0 +1,152 @@
...
+    OpticalPropertyRef const& properties,
+    CerenkovRef const& shared,
+    OpticalMaterialId material,
+    real_type charge)
+    : properties_(properties)
+    , shared_(shared)
+    , material_(material)
+    , alpha_zsq_over_hbar_c_(1e14 * constants::alpha_fine_structure
+                             * constants::e_electron * ipow<2>(charge)
+                             / (constants::hbar_planck * constants::c_light))
```
sethrj: What is 1e14?? If charge is indeed an elementary unity then this is equivalent to:

```suggestion
    , alpha_zsq_over_hbar_c_(1e14 * constants::alpha_fine_structure
                             * ipow<2>(native_value_from(charge))
                             / (constants::hbar_planck * constants::c_light * constants::e_electron))
```
but I don't see where the charge shows up in the earlier equation either...

src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -0,0 +1,61 @@
...
+        reals = other.reals;
+        refractive_index = other.refractive_index;
+        return *this;
+    }
+};
+
+using OpticalPropertyDeviceRef = DeviceCRef<OpticalPropertyData>;
+using OpticalPropertyHostRef = HostCRef<OpticalPropertyData>;
+using OpticalPropertyRef = NativeCRef<OpticalPropertyData>;
+
```
sethrj:
```suggestion
```

test/celeritas/grid/VectorUtils.test.cc
```diff
@@ -80,6 +82,25 @@ TEST(VectorUtils, logspace)
...
+    {
+        std::vector<double> v{10, 100, 1000, 1000};
+        EXPECT_FALSE(is_monotonic_increasing(make_span(v)));
+    }
+    {
+        std::vector<double> v{1e-16, 0, 1, 2};
+        EXPECT_FALSE(is_monotonic_increasing(make_span(v)));
+    }
+}
+
```
sethrj: This and the `delta_squared` change could probably go into a separate small PR.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * M. Daimon and A. Masumura. Measurement of the refractive index of distilled
+ * water from the near-infrared region to the ultraviolet region, Appl. Opt.
+ * 46, 3811-3820 (2007) via refractiveindex.info
+ *
+ * See G4OpticalMaterialProperties.hh.
+ */
+std::vector<double> const& get_wavelength()
+{
+    static std::vector<double> const wavelength = {
```
sethrj: This still ends up copying to the heap at runtime: instead, use `double const wavelength[]` and return a `Span`. (Assuming the data will eventually end up in the main source, that is: if it's really only for the unit test, then don't worry about it.)

src/celeritas/optical/CerenkovDndxCalculator.hh
amandalund: Sorry yeah, one of the parts I still need to fix up :)

test/celeritas/optical/Cerenkov.test.cc
amandalund: Ah thanks. It's possible this will end up in the main source... I'll update it in any case.

src/celeritas/Types.hh
amandalund: I was under the impression that only a subset of materials might have optical properties attached, so we would only want to store the grids for those materials. This isn't really necessary for this PR though, so I could just use `MaterialId` and we could think about this later (when we actually start adding the optical property data)...

whokion: In general, the number of optical materials used in practical detector simulation will be just a few, so introducing
OpticalMaterialId is a good way to go.

whokion (approved):
Looks great! - may be good enough to merge to the develop so that @whokion, @stognini can add other codes top on this MR.

src/celeritas/optical/CerenkovDistributionData.hh
```diff
@@ -0,0 +1,49 @@
...
+ */
+struct CerenkovDistributionData
+{
+    size_type num_photons{};  //!< Sampled number of photons to generate
+    real_type time{};  //!< Pre-step time
+    real_type step_length{};
+    units::ElementaryCharge charge;
+    OpticalMaterialId material{};
+    CerenkovStepData pre;
+    CerenkovStepData post;
```
sethrj: Maybe use `StepPoint` (`celeritas/Types.hh`) and `EnumArray<StepPoint, StepPointSelection> points;` like we do for `StepSelection` etc?

src/celeritas/optical/CerenkovDndxCalculator.hh
```diff
@@ -0,0 +1,153 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the mean number of Cerenkov photons produced per unit length.
+ *
+ * The average number of photons produced is given by
+ * \f[
+   dN = \frac{\alpha z^2}{\hbar c}\sin^2\theta d\epsilon dx = \frac{\alpha
+ z^2}{\hbar c}\left(1 - \frac{1}{n^2\beta^2}\right) d\epsilon dx,
```
sethrj: There's a macro in `doc/_static/macros.tex` that distinguishes the upright differential symbol "d"  from an italicized variable "*d*" (e.g., distance):
```suggestion
   \dif N = \frac{\alpha z^2}{\hbar c}\sin^2\theta \dif\epsilon \dif x = \frac{\alpha
 z^2}{\hbar c}\left(1 - \frac{1}{n^2\beta^2}\right) \dif\epsilon \dif x,
```

src/celeritas/optical/CerenkovDndxCalculator.hh
```diff
@@ -0,0 +1,153 @@
...
+    real_type energy_max = calc_refractive_index.grid().back();
+    if (inv_beta > calc_refractive_index(energy_max))
+    {
+        // Incident particle energy is below the threshold for Cerenkov
+        // emission (i.e., beta < 1 / n_max)
+        return 0;
+    }
+
+    GenericCalculator calc_angle_integral(shared_.angle_integral[material_],
+                                          shared_.reals);
```
sethrj: What's the units returned by the angular integral? MeV?

src/celeritas/optical/CerenkovDndxCalculator.hh
```diff
@@ -0,0 +1,153 @@
...
+        delta_energy = energy_max - energy;
+        angle_integral -= calc_angle_integral(energy);
+    }
+
+    // Calculate number of photons. This may be negative if the incident
+    // particle energy is very close to (just above) the Cerenkov production
+    // threshold
+    return clamp_to_nonneg(
+        alpha_zsq_over_hbar_c_
+        * (delta_energy - angle_integral * ipow<2>(inv_beta)));
```
sethrj: I think it would be a little clearer to just store `q / q_e`, and then use `delta_energy - ...})
```suggestion
        (constants::alpha_fine_structure / (constants::hbar_planck * constants::c_light))
        * zsq_
        * native_units_from(MevEnergy{delta_energy - angle_integral * ipow<2>(inv_beta)});
```

src/celeritas/optical/CerenkovDistributionData.hh
```diff
@@ -0,0 +1,49 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Pre- and post-step data for sampling Cerenkov photons.
+ */
+struct CerenkovStepData
+{
+    units::LightSpeed velocity;
```
sethrj: Isn't this technically speed?

src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -0,0 +1,203 @@
...
+    // Construct from optical properties and distribution parameters
+    inline CELER_FUNCTION
+    CerenkovGenerator(NativeCRef<OpticalPropertyData> const& properties,
+                      NativeCRef<CerenkovData> const& shared,
+                      CerenkovDistributionData const& dist,
+                      Span<OpticalPrimary> photons);
+
+    // Sample Cerenkov photons from the distribution
+    template<class Generator>
+    inline CELER_FUNCTION void operator()(Generator& rng);
```
sethrj: To be more consistent with how the `Interactor` works, and to make clear what is being sampled, I think this should return the `photons_` span. If needed, we can later change the input to be a stack allocator or a larger span that's used as a buffer.

src/celeritas/optical/CerenkovParams.cc
```diff
@@ -0,0 +1,62 @@
...
+    HostVal<CerenkovData> data;
+    for (auto mat_id :
+         range(OpticalMaterialId(properties.refractive_index.size())))
+    {
+        GenericGridData ai_grid;
+        auto const& ri_grid = properties.refractive_index[mat_id];
+        if (!ri_grid)
+        {
+            // No refractive index data stored for this material
+            make_builder(&data.angle_integral).push_back(ai_grid);
```
sethrj: Create a persistent builder for this and `reals` (outside the scope of the `for` loop)...

src/celeritas/optical/CerenkovParams.cc
```diff
@@ -0,0 +1,62 @@
...
+        auto const energy = properties.reals[ri_grid.grid];
+        std::vector<real_type> integral(energy.size());
+        for (size_type i = 1; i < energy.size(); ++i)
+        {
+            integral[i] = integral[i - 1]
+                          + real_type(0.5) * (energy[i] - energy[i - 1])
+                                * (1 / ipow<2>(refractive_index[i - 1])
+                                   + 1 / ipow<2>(refractive_index[i]));
+        }
+        auto reals = make_builder(&data.reals);
```
sethrj: ... and then using a `DedupeCollectionBuilder reals` at the parent scope will allow the same data to be automatically reused if the grids or refraction indices are identical.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -0,0 +1,372 @@
...
+    // Photons are emitted on the surface of a cone, with the cone angle
+    // measured with respect to the incident particle direction. As the
+    // incident energy decreases, the cone angle and the number of photons
+    // produced decreases, and the energy of the emitted photons increases.
+
+    // 10 GeV e-
+    {
+        dist.pre.velocity = units::LightSpeed(0.99999999869453382);
+        dist.post.velocity = units::LightSpeed(0.9999999986942727);
+        dist.step_length = 1;
```
sethrj:
```suggestion
        dist.step_length = 1 * units::centimeter;
```

sethrj (changes requested):
Looking great @amandalund ! Excellent work. I've got a few comments.

src/celeritas/optical/CerenkovDndxCalculator.hh
amandalund: Sounds good!

src/celeritas/optical/CerenkovDistributionData.hh
amandalund: Technically yes :)

sethrj (approved):
Excellent work!

PR #1042, opened 2023-12-01
Fix link errors in edge cases and restore HIP testing

sethrj:
#1038 had to disable HIP to get jenkins working because of problems with the runners. This:
- restores the HIP runtime execution
- adds a warning about the inability to link to *static* celeritas built against HIP (I think due to a downstream cmake bug)
- fixes linking a non-CUDA Celeritas against a CUDA-enabled VecGeom (@hhollenb )

PR #1085, opened 2024-01-18
Support native CLHEP unit system

sethrj:
This allows users to configure Celeritas to process "native" values in a different unit system. Having multiple unit systems supported by the code in CI provides a valuable check for correctness of units in physics code, unit tests, and translation layers.

Following on to #1076 as a holiday/vacation project, this fixes several unit system scaling issues in Celeritas and fixes (or disables) issues with missing system of units in the unit tests.
- Add missing scaling units to event writer
- Replace hardcoded `CLHEP::cm` with a variable `clhep_length` which translates from CLHEP to the native unit system
- Add units to ImportData so that we can convert in place after importing
- Add runtime enumeration for doing unit conversions on import quantities
- Activate VecGeom unit system scaling routines

Things to note:
- All data is converted to native unit system as soon as possible, and it's always exported in the native unit system
- VecGeom hardcodes "bump" values as fixed-precision values (rather than specific to a length scale), so we *cannot* get exactly the same results when using different scale factors.
- CGS doesn't work for JSON ORANGE problems at present because the scales are embedded into the preprocessed data (this will obviously change when we create geometries on the fly)
- Single-precision runtime arithmetic messes with the imported values because the unit system, like the constants, are stored as "real_type" rather than double. (This is to avoid accidentally introducing floating point arithmetic at runtime).

Changes split into separate PRs:
- vecgeom link fix is in #1042 
- unit traits are in #1088 
- ROOT include fix in #1087


test/celeritas/ext/Vecgeom.test.cc
```diff
@@ -156,39 +156,40 @@ TEST_F(SimpleCmsTest, track)
     auto geo = this->make_geo_track_view({0, 0, 0}, {0, 0, 1});
     EXPECT_EQ("vacuum_tube", this->volume_name(geo));
 
-    auto next = geo.find_next_step(100);
-    EXPECT_SOFT_EQ(100, next.distance);
+    auto next = geo.find_next_step(100 * units::centimeter);
```
mrguilima: Can we use here:
`    using cm = units::centimeter;`
and then use `cm` instead of `units::centimeter` in all the following lines?

sethrj: You must be looking at an old version? Those are all `to_cm(100)` (for example) now.

stognini:
Thanks Seth! Looks great.

stognini (approved):

src/celeritas/io/ImportUnits.cc
```diff
@@ -0,0 +1,125 @@
...
+        case ImportUnits::len_inv:
+            return 1 / len;
+        case ImportUnits::len_mev_inv:
+            return 1 / (len * mev);
+        case ImportUnits::mev_sq_per_len:
+            return ipow<2>(mev) / len;
+        case ImportUnits::len_sq:
+            return ipow<2>(len);
+        case ImportUnits::time:
+            return 1 / ipow<2>(time_scale(sys));
```
amandalund:
```suggestion
            return time_scale(sys);
```
Maybe this needs a test?

src/celeritas/ext/detail/GeantMicroXsCalculator.cc
```diff
@@ -54,17 +55,31 @@ void GeantMicroXsCalculator::operator()(VecDouble const& energy_grid,
...
+    auto calc_element_xs
+        = [this, &elements](std::size_t elcomp_idx, double energy) {
+              // Calculate microscopic cross section
+              double xs = model_.ComputeCrossSectionPerAtom(
+                  &particle_,
+                  elements[elcomp_idx],
+                  energy,
+                  secondary_cut_,
+                  /* max_energy = */ std::numeric_limits<double>::max());
+              return ::celeritas::max(xs, 0.0);
```
amandalund: Or:
```suggestion
              return clamp_to_nonneg(xs);
```

test/celeritas/em/RelativisticBrem.test.cc
```diff
@@ -35,10 +35,12 @@ class RelativisticBremTest : public InteractorHostTestBase
...
     void SetUp() override
     {
+        using namespace units;
+
         // Set up shared material data
         MaterialParams::Input mi;
-        mi.elements = {{AtomicNumber{82}, units::AmuMass{207.2}, {}, "Pb"}};
-        mi.materials = {{0.05477 * constants::na_avogadro,
+        mi.elements = {{AtomicNumber{82}, AmuMass{207.2}, {}, "Pb"}};
+        mi.materials = {{native_value_from(MolCcDensity{0.05477}),
```
amandalund, line 43: Any reason you removed the `* constants::na_avogadro` in some places but not others?

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -171,18 +182,23 @@ TEST_F(CerenkovTest, angle_integral)
...
-                     constants::alpha_fine_structure * units::Mev::value()
-                         / (constants::hbar_planck * constants::c_light),
-                     1e-6);
+    if (CELERITAS_UNITS == CELERITAS_UNITS_CGS)
+    {
+        EXPECT_SOFT_NEAR(369.81e6,
+                         constants::alpha_fine_structure * units::Mev::value()
+                             / (constants::hbar_planck * constants::c_light),
+                         1e-6);
+    }
```
amandalund: Or some variation of (units are 1 / MeV-cm):
```suggestion
EXPECT_SOFT_NEAR(369.81e6,
                 constants::alpha_fine_structure * units::Mev::value()
                     / (constants::hbar_planck * constants::c_light)
                     * units::centimeter,
                 1e-6);
```

test/celeritas/mat/Material.test.cc
```diff
@@ -157,6 +164,11 @@ TEST_F(MaterialTest, params)
 
 TEST_F(MaterialTest, material_view)
 {
+    if (CELERITAS_UNITS != CELERITAS_UNITS_CGS)
+    {
+        GTEST_SKIP() << "Test needs updating to include units";
```
amandalund, line 169: It looks like you already updated part of this test... were you planning on updating the rest?

amandalund (commented):
Cool, this is great @sethrj!

Are the tests expected to pass in all unit systems now? I'm still seeing a handful failing with SI and CLHEP (looks like device tests).

test/celeritas/em/RelativisticBrem.test.cc
sethrj: Find-replace gone wrong, I think. Thanks!

test/celeritas/mat/Material.test.cc
sethrj: I got lazy at some point... I just wanted enough to show that the code essentially works, and later (next Christmas?) polish everything to a shine.

amandalund:
Tests are passing now with CLHEP, but I'm still seeing the following failures with SI:
```console
	 88 - celeritas/Constants (Failed)
	 96 - celeritas/em/LivermorePE (Failed)
	103 - celeritas/em/UrbanMsc (Failed)
	111 - celeritas/ext/GeantGeo (Failed)
	113 - celeritas/ext/GeantImporter:FourSteelSlabs* (Failed)
	115 - celeritas/ext/GeantImporter:OneSteelSphere.* (Failed)
	124 - celeritas/field/LinearPropagator (Failed)
	136 - celeritas/global/AlongStep:SimpleCmsRZFieldAlongStepTest.msc_rzfield (Failed)
	166 - celeritas/phys/Physics (Failed)
	186 - celeritas/track/TrackSort (Failed)
	208 - app/demo-rasterizer (Failed)
```

sethrj:
Ah I never even tested the whole thing with SI (only tested the units/constants). How about I change the title and fix SI in a subsequent PR? :)

amandalund (approved):
Haha, sure that works.  Thanks!

Issue #886, opened 2023-08-10
Add optical photon simulation


whokion:
# Requirements discussion

This is an initial draft for supporting optical photon simulation and a placeholder to collect and identify sub-tasks
for implementing related codes and testing.  Your additional input, refinement and contributions are welcome.

### Major categories and sub-tasks:

**Optical materials and properties** 
- [X] Determine common optical properties
- [ ] ~~Automatically populate common optical materials such as Water, Liquid Xeon and Argon~~
- [x] Add `OpticalPropertyParams` (constant properties, vectors of properties such as momentum dependent optical properties) 
- [x] Construct `OpticalPropertyView` from material ID

**Optical photon physics**  
- **Add processes for generating optical photons in optical materials**
 - [x] Design consideration for interfaces to other physics actions (energy loss or deposition) and `Photino` (an object returned from processes of optical photon generation, which will be passed to the loop of optical photon propagation) — `OpticalPhotonDistribution` and related
 - [x] Scintillation
 - [x] Cerenkov
 - [ ] ~~Transition Radiation~~

- **Add processes for propagating optical photons in fiducial volumes/materials**
 - [x] Design consideration for `step limiter` (mfp) and `interactor` (final state sampling) and executing them on HPC systems
 - [x] Absorption
 - [x] Boundary and surface processes and associated models (GLISUR, UNIFIED)
 - [x] Scattering (Rayleigh and Mie)
 - [x] Wavelength shift or re-emission

**I/O**
- [x] Add input data structure, `OpticalTrackView` or `OpticalPhotonInput` for an optical action
- [ ] ~~Add default sensitive hits (detector Id, hit{position, energy, time}) or interfaces for user defined sensitive detectors~~

**Testing and physics validation/verification**
- [ ] ~~Verify physics results in the process level with respect to Geant4 - use the exiting physics validation suite~~
- [ ] Verify the hit position, energy, timing in sensitive detectors (PMT or Calorimeters) with respect to Geant4

**Performance evaluation**
- [ ] Evaluate performance (event throughput) with respect to the Geant4 optical photon simulation
- [ ] ~~Evaluate performance (event throughput) with respect to the Geant4+Opticks+OptiX with a test example~~
- [ ] Study performance and identify areas for improvement or optimization  
- [ ] ~~Design consideration for efficient ray tracing on accelerators~~

**Offloading interfaces and integration on experiment frameworks**
[...]

PR #1086, opened 2024-01-19
Add ScintillationGenerator and related data structures

whokion:
# Add an initial set of codes for generating scintillation photons and related data structures.

This MR adds an initial/minimal set of codes for generating scintillation photons and related data structure - following the earlier work by @amandalund for `CerenvkovGenerator` as closely as possible.  Nevertheless, a significant refactoring may be foreseen to consolidate the common code-basis and accommodate future developments for high-level `Cerenkov and Scintillation processes.`:
 - Implement ScintillationGenerator *[enhancement, physics]*
 - Add ScintillationData and ScintillationParams to manage related optical property data
 - Add a rudimentary test, ScintillationGenerator.test.


src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -15,6 +15,21 @@
...
+/*!
+ * Material dependent scintillation spectra.
+ *
+ * Support upto 3-components: [fast, medium, slow]
+ */
+struct ScintillationSpectra
+{
+    Real3 yield_prob{};  //!< Probability of the yield
+    Real3 lambda_mean{};  //!< Mean wavelength [meter]
+    Real3 lambda_sigma{};  //!< Standard dev. of wavelength [meter]
```
sethrj: Why are these in meters? This should be in the native unit system unless there's a really good reason not to... and if they are meters then we should define a unit type (helper class)

src/celeritas/optical/ScintillationData.hh
```diff
@@ -0,0 +1,48 @@
...
+#include "OpticalPropertyData.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ *  Scintillation data tabulated with the optical material id.
+ */
+template<Ownership W, MemSpace M>
+struct ScintillationData
```
sethrj, line 51: Is this class redundant with optical data?

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+    , sample_cost_(-1, 1)
+    , sample_phi_(0, 2 * constants::pi)
+    , sample_time_(real_type{1} / spectra_.fall_time[0])
+    , is_neutral_{native_value_from(input_.charge) == 0}
+{
+    CELER_EXPECT(input_);
+    CELER_EXPECT(photons_.size() == input_.num_photons);
+
+    delta_pos_ = input_.post_pos - input_.pre_pos;
+    delta_velocity_ = input_.post_velocity - input_.pre_velocity;
```
sethrj: Speed 😉 right?

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+        size_type num_photons = input_.num_photons * spectra_.yield_prob[sid];
+
+        CELER_EXPECT(num_generated + num_photons <= input_.num_photons);
+
+        for (size_type i : range(num_generated, num_generated + num_photons))
+        {
+            // Sample wavelength and convert to energy
+            real_type wave_length = sample_lambda_(rng);
+            CELER_EXPECT(wave_length > 0);
+            photons_[i].energy = Energy(this->hc() / wave_length);
```
sethrj: Instead of dividing by MeV in your `hc` (error prone) this would be better written as
```suggestion
            photons_[i].energy = native_units_to<Energy>(constants::c_light * constants::h_planck / wave_length);
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+template<class Generator>
+CELER_FUNCTION Span<OpticalPrimary>
+ScintillationGenerator::operator()(Generator& rng)
+{
+    // Loop for generating scintillation photons
+    size_type num_generated{0};
+
+    for (auto sid : range(3))
+    {
+        NormalDistribution<real_type> sample_lambda_(
```
sethrj:
```suggestion
        NormalDistribution<real_type> sample_lambda(
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+
+            // Sample polarization
+            photons_[i].polarization
+                = from_spherical(-std::sqrt(1 - ipow<2>(cost)), phi);
+            Real3 perp = cross_product(photons_[i].polarization,
+                                       photons_[i].direction);
+            phi = sample_phi_(rng);
+            photons_[i].polarization = std::cos(phi) * photons_[i].polarization
+                                       + std::sin(phi) * perp;
+
```
sethrj: Are you just using the first polarization assignment as a temporary?

src/celeritas/optical/ScintillationInput.hh
```diff
@@ -0,0 +1,40 @@
...
+#include "corecel/Types.hh"
+#include "celeritas/Types.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Input data for sampling optical photons for a given scintillation step.
+ */
+struct ScintillationInput
```
amandalund: If this really is identical to `CerenkovDistributionData`, I think we could just rename that one something like `OpticalDistributionData` and use it for both generators?

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+    ScintillationInput const& input,
+    NativeCRef<ScintillationData> const& shared,
+    Span<OpticalPrimary> photons)
+    : input_(input)
+    , spectra_(shared.spectra[input.matId])
+    , photons_(photons)
+    , sample_cost_(-1, 1)
+    , sample_phi_(0, 2 * constants::pi)
+    , sample_time_(real_type{1} / spectra_.fall_time[0])
+    , is_neutral_{native_value_from(input_.charge) == 0}
```
sethrj:
```suggestion
    , is_neutral_{input_.charge == zero_quantity()}
```

src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -15,6 +15,21 @@
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Material dependent scintillation spectra.
+ *
+ * Support upto 3-components: [fast, medium, slow]
+ */
+struct ScintillationSpectra
```
sethrj: An instance of this class is a spectrum (singular) right?
```suggestion
struct ScintillationSpectrum
```

src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -15,6 +15,21 @@
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Material dependent scintillation spectra.
+ *
+ * Support upto 3-components: [fast, medium, slow]
```
sethrj:
```suggestion
 * Support up to 3 components: [fast, medium, slow]
```

src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -15,6 +15,21 @@
...
+ */
+struct ScintillationSpectra
+{
```
sethrj: Should we have an enumeration for this? Should it be a runtime selectable value? Since the 3 components aren't spatial x/y/z, perhaps we should do:
```suggestion
{
    static inline constexpr size_type size = 3;
    using Real3 = Array<real_type, size>;
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+ * scintillation time structure with one or double exponentials.
+ */
+template<class Generator>
+CELER_FUNCTION Span<OpticalPrimary>
+ScintillationGenerator::operator()(Generator& rng)
+{
+    // Loop for generating scintillation photons
+    size_type num_generated{0};
+
+    for (auto sid : range(3))
```
sethrj:
```suggestion
    for (auto sid : range(spectrum_.size))
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+/*!
+ * Construct from shared scintillation data and input.
+ */
+CELER_FUNCTION
+ScintillationGenerator::ScintillationGenerator(
+    ScintillationInput const& input,
+    NativeCRef<ScintillationData> const& shared,
+    Span<OpticalPrimary> photons)
+    : input_(input)
+    , spectra_(shared.spectra[input.matId])
```
sethrj:
```suggestion
    , spectrum_(shared.spectrum[input.matId])
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,196 @@
...
+                real_type scint_time{};
+                real_type envelop{};
+                do
+                {
+                    scint_time = sample_time_(rng);
+                    envelop
+                        = 1 - std::exp(-scint_time / spectra_.rise_time[sid]);
+                } while (!BernoulliDistribution(envelop)(rng));
+                delta_time += scint_time;
+            }
```
sethrj, line 198:
```suggestion
            }
            CELER_ASSERT(delta_time >= 0);
```

src/celeritas/optical/ScintillationInput.hh
sethrj: These data look identical to the cerenkov one that @amandalund  added: perhaps we can just rename it to `OpticalDistributionData` and reuse it here?

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -0,0 +1,32 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with optical property data.
+ */
+ScintillationParams::ScintillationParams(OpticalPropertyCRef const& properties)
+{
+    HostVal<ScintillationData> data;
+    for (auto mid : range(OpticalMaterialId(properties.scint_spectra.size())))
+    {
+        make_builder(&data.spectra).push_back(properties.scint_spectra[mid]);
```
sethrj:
```suggestion
    CollectionBuilder spectra{&data.spectra};
    for (auto mid : range(OpticalMaterialId(properties.scint_spectra.size())))
    {
        spectra.push_back(properties.scint_spectra[mid]);
```

sethrj (commented):
Looking good, @whokion !

src/celeritas/optical/OpticalPropertyData.hh
```diff
@@ -33,6 +48,7 @@ struct OpticalPropertyData
 
     Items<real_type> reals;
     OpticalMaterialItems<GenericGridData> refractive_index;
+    OpticalMaterialItems<ScintillationSpectra> scint_spectra;
```
amandalund: I'm not sure I understand why this is stored both in the optical property data and in the scintillation data. I think this struct should contain the optical properties from Geant4. I assume an application will set the properties needed for scintillation (like `SCINTILLATIONCOMPONENT[1,2,3]` and we will import those from Geant4 -- how do we plan to map those to the `ScintillationSpectra`?

test/celeritas/optical/ScintillationGenerator.test.cc
```diff
@@ -0,0 +1,184 @@
...
+//---------------------------------------------------------------------------//
+
+TEST_F(ScintillationGeneratorTest, stress_test)
+{
+    // Generate a large number of optical photons
+    input_.num_photons = 1e+6;
+
+    // Output data
+    std::vector<OpticalPrimary> storage;
+    storage.reserve(input_.num_photons);
```
amandalund: Not sure if you tried running with debug assertions on, but I'm seeing some failures in the tests, e.g. from:
```suggestion
    std::vector<OpticalPrimary> storage(input_.num_photons);
```

amandalund (commented):
Looks good @whokion! I have a few higher-level questions/comments before I look at the details of the generator.

src/celeritas/optical/ScintillationInput.hh
sethrj: Exactly. Can you do that as part of your pull request?

src/celeritas/optical/ScintillationParams.cc
amandalund: Or in this case I think just `data.spectra = properties.scint_spectra` would work (although again, not sure why they are stored twice).

src/celeritas/optical/OpticalPropertyData.hh
whokion: You are right! In principle, `ScintillationProcess` will/should prepare `ScintillationParams/Data` directly  from` MaterialView/Params`(?) if there is any scintillation material) so that data can be decoupled from other processes.  We should revisit a general approach how to handle optical materials and properties when optical processes are implemented along with optical material and property data.  I will remove `ScintillationSpectra` data from OpticalPropertyData and directly construct `ScintillationParams` for an initial test.

test/celeritas/optical/ScintillationGenerator.test.cc
whokion: Sorry that I did not turn on DEBUG! I think that this error is somewhat related to the use of OpticalPropertyData for ScintillationGenerator.test (in which refraction_index data is not built for the scintillation test.  I will address this with the rest of comments by @sethrj this afternoon.  Many thanks!!

src/celeritas/optical/OpticalPropertyData.hh
whokion: Okay, the inline comment was miss-leading. Dropped explicit units.

src/celeritas/optical/ScintillationData.hh
whokion: Make it separate from OpticalPropertyData.

src/celeritas/optical/ScintillationGenerator.hh
whokion: Yes.

src/celeritas/optical/OpticalPropertyData.hh
whokion: Well, the actual emission spectrum is a convolution of one or more individual spectrum (so, I use the plural), but I agree that spectrum is more generic.  Also changed spectra to spectrum accordingly.

src/celeritas/optical/ScintillationInput.hh
whokion: Renamed CerenkovDistributionData to `OpticalDistributionData` and used it for ScintillationInput (which is now removed).

src/celeritas/optical/ScintillationParams.cc
whokion: Build it directly from ScintillationData.

src/celeritas/optical/ScintillationInput.hh
whokion: Done!

test/celeritas/optical/ScintillationGenerator.test.cc
amandalund: True, probably we should not require that refractive index (or any optical properties) be assigned in `OpticalPropertyData`.

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+template<class Generator>
+CELER_FUNCTION Span<OpticalPrimary>
+ScintillationGenerator::operator()(Generator& rng)
+{
+    // Loop for generating scintillation photons
+    size_type num_generated{0};
+
+    for (auto sid : range(spectrum_.size))
+    {
+        // Skip if there is no emission warelength
```
sethrj:
```suggestion
        // Skip if there is no emission wavelength
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+            real_type wave_length = sample_lambda(rng);
+            CELER_EXPECT(wave_length > 0);
+            photons_[i].energy 
+                = native_value_to<Energy>(this->hc() / wave_length);
+
+            // Sample direction
+            real_type cost = sample_cost_(rng);
+            real_type phi = sample_phi_(rng);
+            photons_[i].direction = rotate(from_spherical(cost, phi), 
+                                           dist_.points[StepPoint::post].pos);
```
sethrj: Shouldn't this be rotating using the *direction* of the step (normalized value of `delta_pos_`)? This is using the spatial position, I think.

src/celeritas/optical/ScintillationGenerator.hh
sethrj: I think it would be good to write that in the `sample polarization` comment, or just create a temporary variable (and hopefully the compiler will optimize it out).

Also since this expression is using *vector* calculations, it's making a bunch of temporaries (three if I count correctly) so it would be better to calculate sincos phi (you could even use the `sincospi` functions from `math/Algorithms.hh`) and loop over the vector elements:
```c++
real_type sinphi, cosphi;
sincospi(UniformRealDist{0, 1}(rng), &sinphi, &cosphi);
for (int j = 0; j < 3; ++j)
{
    photons_[i].polarization[j] = cosphi * temp[j] + sinphi * perp[j];
}
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+            }
+            else
+            {
+                real_type scint_time{};
+                real_type envelop{};
+                do
+                {
+                    scint_time = sample_time_(rng);
+                    envelop
+                        = 1 - std::exp(-scint_time / spectrum_.rise_time[sid]);
```
sethrj: I'm guessing that since this is only used for acceptance, we don't really care about numerical accuracy? Otherwise this would be better written as
```suggestion
                    envelop = -std::expm1(-scint_time / spectrum_.rise_time[sid]);
```

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+CELER_FUNCTION Span<OpticalPrimary>
+ScintillationGenerator::operator()(Generator& rng)
+{
+    // Loop for generating scintillation photons
+    size_type num_generated{0};
+
+    for (auto sid : range(spectrum_.size))
+    {
+        // Skip if there is no emission warelength
+        if(spectrum_.lambda_mean[sid] <= 0)
```
sethrj: When is this the case? Are we sure it corresponds to a zero yield probability? And if so, we shouldn't really need a special "exit early" here since line 145 will be an empty loop because `num_photons = 0`?

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+    for (auto sid : range(spectrum_.size))
+    {
+        // Skip if there is no emission warelength
+        if(spectrum_.lambda_mean[sid] <= 0) 
+           continue;
+
+        NormalDistribution<real_type> sample_lambda(
+            spectrum_.lambda_mean[sid], spectrum_.lambda_sigma[sid]);
+
+        size_type num_photons = dist_.num_photons * spectrum_.yield_prob[sid];
```
sethrj: With this construction, isn't it possible for the total number of photons to be rounded down so that the number of photons is less than `dist_.num_photons`?

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -0,0 +1,31 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with scintillation data.
+ */
+ScintillationParams::ScintillationParams(ScintillationDataCRef const& data)
+{
+    CELER_EXPECT(data);
+    HostVal<ScintillationData> host_data;
```
sethrj, line 26: Where should we put error checking (positive mean/sigma for the scintillation components, yields add up to 1, etc)?

src/celeritas/optical/OpticalPropertyData.hh
amandalund: Thanks @whokion, so if I understand correctly,  the Geant4 optical properties used for scintillation will more or less map directly to the `ScintillationSpectrum` fields. The only exception is the vector property `SCINTILLATIONCOMPONENT[1,2,3]`, the (Gaussian-like) scintillation emission spectrum tabulated for different photon energies, which in our data is replaced by the scalar properties `lambda_mean` and `lambda_sigma`, the parameters for sampling the wavelength from a Gaussian distribution. Regarding that:
 - Will the emission spectrum always be a Gaussian?
 - Since `SCINTILLATIONCOMPONENT` has energy (rather than wavelength) grid values, and ultimately we want to sample the energy, should we store `energy_mean` and `energy_sigma` instead?
 - Can we calculate the mean and stdev from the `SCINTILLATIONCOMPONENT`? If so, maybe this should be done in `ScintillationParams`, or at least as part of the test?
 - Or if not,  how do we get these parameters in an integrated Geant4 application? Do we require the user add new optical properties (which I guess can be done?) for the Gaussian distribution parameters and provide those instead of the tabulated data?

src/celeritas/optical/ScintillationGenerator.hh
amandalund: I think since the directions are sampled randomly (uniformly distributed points on the unit sphere) it can just be `photons_[i].direction = from_spherical(cost, phi)`.

src/celeritas/optical/ScintillationGenerator.hh
amandalund: Might need a `make_unit_vector()` in there too.

src/celeritas/optical/OpticalPropertyData.hh
whokion: - Gaussian should a good approximation as the scintillation wavelength is in reality a delta function (transition wavelength in the atom spectroscopy) with a (measurement) spread.
- As you know, we can always exchange energy vs wavelength, but the energy is a reciprocal Gaussian distribution as E~1/lambda if lambda is a Gaussian distribution which is much easier to sample randomly (than the reciprocal Gaussian). So, wavelength is a better choice for an input.
- Celeritias may provides an internal DB for constant properties (in this case, lambda_mean[] and lambda_sigma[] for known materials, but they should be in principle user inputs (at least, should support user defined constant properties).  So, this is a good item to consider how Celeritas supports optical properties (both constant and vector-like data) in a long run even there is no systematic solution in Geant4 either - one of our colleagues from MIT has proposed to support the better optical material and properties in Geant4 (like Geant4 defined optical materials and a complete set of associated optical properties) as he has a large collection of optical material properties, for which I suppose to be a technical assistant to integrate his DB into Geant4 (hope that this happens this year and also clear out all related licensing issues).

src/celeritas/optical/ScintillationGenerator.hh
whokion: Thanks, Fixed!

src/celeritas/optical/ScintillationGenerator.hh
whokion: Right, it should be just a random isotropic and no need a reference direction(i.e., rotational invariant). Fixed.

src/celeritas/optical/ScintillationGenerator.hh
whokion: Updated as suggested.

src/celeritas/optical/ScintillationGenerator.hh
whokion: A good point! I also thought about the better way how to deal with a potential mis-match due to double->int conversion.  A sure solution is to pass an extra num_to_generate[3] as a part of ScintillationParams of which sum should match to dist_.num_photons.  Here, the assumption is that the sum of yield_prob*dist_.num_photons is already adjusted to dist_.num_photons when input dist_ is prepared in the process level.  We can revisit this when ScintillationProcess is implemented.

src/celeritas/optical/ScintillationGenerator.hh
sethrj: Actually @whokion if the original sampling is isotropic, this whole series of operations simplifies immensely. If I did the math right, the cross product is `{sinphi, cosphi, 0}`, and you get a direction of `{sintheta * cosphi, sintheta * sinphi, costheta}` with a polarization of `{sinphi^2 + cosphi^2 * costheta, sinphi * cosphi * (1 + costheta), -cosphi * sintheta}`.

src/celeritas/optical/ScintillationGenerator.hh
whokion: Okay, this was a temporary solution to protect the zero standard inside `NormalDistribution` (at  [CELER_EXPECT(stddev > 0)](https://github.com/celeritas-project/celeritas/blob/develop/src/celeritas/random/distribution/NormalDistribution.hh#L73); ).  An alternative way is to support an arbitrary number of scintillation components (i.e, not as Real3, but a variable size of array along with num_components.  Any good solution or suggestion?

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,206 @@
...
+
+        CELER_EXPECT(num_generated + num_photons <= dist_.num_photons);
+
+        for (size_type i : range(num_generated, num_generated + num_photons))
+        {
+            // Sample wavelength and convert to energy
+            real_type wave_length = sample_lambda(rng);
+            CELER_EXPECT(wave_length > 0);
+            photons_[i].energy
```
stognini, line 150: Here the energy is being set using `native_value_to` but in `OpticalPrimary` the energy is defined as `MevEnergy`.

src/celeritas/optical/ScintillationGenerator.hh
sethrj: I think we can loosen the assertion to `stddev >= 0`. It may not be "smart" to construct (or sample from) a zero-width gaussian but I don't think it should be prohibited. (It's good to have a "use case" of that edge case, though.)

src/celeritas/optical/ScintillationGenerator.hh
sethrj: Besides the summation problem, I think truncating like here is actually biased. We should be using stochastic rounding:
```c++
real_type num_photons_real = dist_.num_photons * spectrum_.yield_prob[sid];
size_type num_photons = static_cast<size_type>(num_photons_real);
num_photons += BernoulliDistribution{num_photons_real - num_photons}(rng);
```

amandalund: True, we should. I guess then we'd need to make sure enough storage is allocated for a few extra photons, or later use a stack allocator to allocate it inside the generator, unless we move this part to the pre-generator.

sethrj: There's got to be a smart and efficient way to guarantee that we divide N photons fairly into M groups based on the yield probabilities... (the hard part might be to do it without allocating any extra space, and to make sure that all photons with the same spectra are grouped together so we can reuse the gaussian sampler...)

whokion:
@sethrj, @amandalund, @stognini I do not have a computer node to work on today and will address the rest of review comments later.  Sorry about being delayed.

src/celeritas/optical/ScintillationGenerator.hh
sethrj: And actually if we want to preserve the distribution of photons rather than just the mean, shouldn't we use a binomial distribution to sample the number from each group? (I.e. sometimes we might sample all short-wavelength photons, sometimes all long-wavelength.)

whokion: It is a good argument whether each component of scintillation spectra is independent of (which should be the case, but hard to related what amount of energy (i.e. from ionization) actually contributes to each scintillation component - in the Geant4 approach, the total number of photons is sampled by the Poisson distribution with the sum of mean number of photons for each component, which again is used to estimate the number of photons to generate for each component based on provided yields (so in some sense the stochastic treatment is already done before the generator.  As I shortly mentioned above,  a clean solution is to estimate all numbers of photons to generate  before the generator and pass them via the distribution (i.e, input data), which prevents the the rounding error from the loop (at the cost of additional data entry in the OpticalDistributionData, but some redundant information (i.e., sum of all components  == num_photons) or make the OpticalDistributionData different between Cerenkov (just need the total number of photons) and Scintillation (need numbers for all components separately). An alternative tweak with keeping current data structures as they are  is something  like (looks ugly) 
```
     // Calculate the number of photons for each scintillation component         
    Array<size_type, ScintillationSpectrum::size> num_photons{};
    size_type remaining = dist_.num_photons;
    for (auto sid : range(spectrum_.size))
    {
        num_photons[sid] = static_cast<size_type>(dist_.num_photons
                                                  * spectrum_.yield_prob[sid]);
        remaining -= num_photons[sid];
    }

    // Add remaining due to the rounding off, if any, to the first component (or can add to the last
    // none-empty yield component with extra codes    
    if (remaining != 0)
    {
        num_photons[0] += remaining;
    }
```
and loop over with all num_photons[3], at then ensure
`  CELER_ASSERT(num_generated == dist_.num_photons);`
after the loop. How do you think?

src/celeritas/optical/ScintillationGenerator.hh
whokion: In principle, the requirement is fully legitimated as stddev should be non-zero to define a Gaussian even though the normal distribution will just return the mean for the case.  Not sure whether the relaxing the assertion is a good practice even though it may be not harmful.  The above code looks ugly, but should be removed when we improve the code with an arbitrary number of components (which I do not know how, but should doable if the required memory for the size of SintillationSpectrum (which has all non-zero entries) is pre-allocated on the fly - I need your help for this).

src/celeritas/optical/ScintillationGenerator.hh
amandalund: I'm sure there are better ways, but one possibility to divide the number of photons fairly (similar to what @whokion suggested, but I think in an unbiased way) is to precompute the number of photons outside the loop, first calculating the integer part and then using the normalized fractional parts as a PMF to sample which spectra the remaining photon(s) will go to.  Something like:
```c++
Array<size_type, 3> num_photons;
Real3 frac;
real_type frac_sum{0};
size_type count{0};
for (int i = 0; i < 3; ++i) 
{
    real_type num_photons_real = dist_.num_photons * spectrum_.yield_prob[i];
    num_photons[i] = num_photons_real;
    frac[i] = num_photons_real - num_photons[i];
    frac_sum += frac[i];
    count += num_photons[i];
}
for (size_type i = count; i < dist_.num_photons; ++i)
{
    real_type cutoff = frac_sum * generate_canonical(rng);
    real_type frac_cumsum{0};
    for (int j = 0; j < 3; ++j)
    {
        frac_cumsum += frac[j];
        if (frac_cumsum > cutoff) 
        {
            ++num_photons[j];
            break;  
        }
    }
}
```
But maybe this is overkill.

whokion: As the number of photons is large (~40000/MeV energy loss for LAr, for an example, without a field), and the user provided yields is a rough estimation, the bias (where to assign extra 1, if happens, most of cases) may be really a negligible effect without loss of generality. Thank you for the suggestion anyway!

amandalund: Yeah I agree, there are enough approximations and it's such a small bias that it shouldn't matter :)

src/celeritas/optical/ScintillationParams.cc
whokion: This may be done when we build Params from the process level - and we can revisit later as necessary.

whokion:
Not sure whether I addressed all review comments (seems that some of earlier comments are disappeared), but please let me know if there are additional comments.  Thanks!

src/celeritas/optical/ScintillationGenerator.hh
sethrj: lol ok i didn't realize we were talking 40000 samples 😅

src/celeritas/optical/ScintillationGenerator.hh
amandalund: I think I also prefer the option of supporting an arbitrary number of components. One way to do this would be:
 - Change `ScintillationSpectrum` to something like `ScintillationComponent` and change the `Real3`s to `real_type`
 - Add a struct:
```c++
struct ScintillationSpectrum
{
    ItemRange<ScintillationComponent> components;
 };
```
 - And add a collection `Items<ScintillationComponent> components` to `ScintillationData`

Then in `ScintillationParams`:
 - Instead of using `HostCRef<ScintillationData>` as input, define a helper struct in the class like:
```c++
struct Input
{
   std::vector<std::vector<ScintillationComponent>> materials; // [num_optical_materials][num_components]
};
```
 - And use collection builders to build the `ScintillationData` from the input:
```c++
CollectionBuilder spectra(&host_data.spectra);
CollectionBuilder components(&host_data.components);
for (auto const& spec : input.materials)
{
    ScintillationSpectrum spectrum;
    spectrum.components = components.insert_back(spec.begin(), spec.end());
    spectra.push_back(spectrum);
}
```

whokion: Thank you for the tip.  Will try to implement as suggested.

whokion: @amandalund I modified the ScintillationParams/Data to support an arbitrary number of scintillation components.  Thanks again for the tip!

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,218 @@
...
+        size_type num_photons
+            = static_cast<size_type>(dist_.num_photons * component.yield_prob);
+        remaining -= num_photons;
+
+        // Add remaining due to the rounding off, if any, to the last component
+        if (sid.unchecked_get() + 1
+            == spectrum.components.end()->unchecked_get())
+        {
+            num_photons += remaining;
+        }
```
amandalund: I think this could be simplified slightly:
```suggestion
    size_type num_photons
        = sid.get() + 1 == spectrum.components.size()
              ? dist_.num_photons - num_generated
              : static_cast<size_type>(dist_.num_photons
                                       * component.yield_prob);
```

src/celeritas/optical/ScintillationParams.cc
amandalund: It still might be helpful to add some error checking of the spectra here for now, and we can move it around later if needed.

test/celeritas/optical/ScintillationGenerator.test.cc
```diff
@@ -0,0 +1,196 @@
...
+        dist_, params->host_ref(), make_span(storage));
+    RandomEngine& rng_engine = this->rng();
+
+    // Generate optical photons for a given input
+    auto photons = generate_photons(rng_engine);
+
+    // Check results
+    std::vector<real_type> energy;
+    std::vector<real_type> time;
+    std::vector<real_type> cos_theta;
```
amandalund, line 124: Should we also check the polarization?

test/celeritas/optical/ScintillationGenerator.test.cc
```diff
@@ -0,0 +1,196 @@
...
+                                       2.68409417173788e-06};
+
+    const real_type expected_cos_theta[] = {0.937735542248463,
+                                            -0.775070967887639,
+                                            0.744857640134601,
+                                            -0.748206733055997};
+
+    EXPECT_VEC_SOFT_EQ(expected_energy, energy);
+    EXPECT_VEC_SOFT_EQ(expected_time, time);
+    EXPECT_VEC_SOFT_EQ(expected_cos_theta, cos_theta);
```
amandalund: We'll need to update this for the tests to pass with real_type=float, e.g. put all this inside an `if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)`.

amandalund (commented):
Looks great @whokion! Just a few more small comments from me.

src/celeritas/optical/ScintillationGenerator.hh
whokion: Nice! Refactored.

test/celeritas/optical/ScintillationGenerator.test.cc
whokion: Okay, I will add a check for polarization.  Regarding to some error checking in ScintillationParams, I will add at least the size check for the input vector.  Do you also want to add something to check the validity of ScintillationComponent, such as the sum of yield and so on, here?

amandalund: Yeah I think it would be good to have those checks, and that seems like the best place for them for now.

src/celeritas/optical/ScintillationGenerator.hh
sethrj: This by the way should impose the constraint that the last spectrum component has a nonzero yield.

whokion: Yes, the yield should be always (assumed to be) non-zero for any scintillation component and the sum of yields should be one, which will be checked when each scintillation component is created and pushed data - related to the other comment by @amandalund.  On the other hand, this last component of the yield is not actually needed here due to the constraint though.

test/celeritas/optical/ScintillationGenerator.test.cc
whokion: Added a check for the validity of scintillation component data and the result of photon polarization.

test/celeritas/optical/ScintillationGenerator.test.cc
whokion: Added.

whokion:
@sethrj Not sure what the 1 pending check is, but I guess that all review comments are more or less addressed and this MR may be good enough to be merged now.  Thanks you all for your review comments and good suggestions!

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -27,6 +29,16 @@ ScintillationParams::ScintillationParams(ScintillationInput const& input)
...
         CELER_EXPECT(spec.size() > 0);
+
+        // Validity of scintillation component data
+        real_type total_yield{0};
+        for (auto const& comp : spec)
+        {
+            CELER_EXPECT(comp);
+            total_yield += comp.yield_prob;
+        }
+        CELER_ENSURE(total_yield == 1);
```
amandalund:
```suggestion
        CELER_ASSERT(spec.size() > 0);

        // Validity of scintillation component data
        real_type total_yield{0};
        for (auto const& comp : spec)
        {
            CELER_ASSERT(comp);
            total_yield += comp.yield_prob;
        }
        CELER_ASSERT(total_yield == 1);
```
We may need to do something like `CELER_ASSERT(soft_equal(real_type(1), total_yield))`, or renormalize the yields to make sure it's as close to 1 as possible.

amandalund (approved):
Very nice, thanks @whokion! Looks good on my end.

src/celeritas/optical/ScintillationParams.cc
whokion: Thanks! Fixed.

src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -0,0 +1,216 @@
...
+
+    // Sample Scintillation photons from the distribution
+    template<class Generator>
+    inline CELER_FUNCTION Span<OpticalPrimary> operator()(Generator& rng);
+
+    // TODO: Move it to detail/OpticalUtils.
+    static CELER_CONSTEXPR_FUNCTION real_type hc()
+    {
+        return constants::h_planck * constants::c_light;
+    }
```
sethrj: This seems too simple to move to optical utils :) I think it'd be better inlined into the code (a single operation doesn't seem worth creating a static function to help eliminate).

test/celeritas/optical/ScintillationGenerator.test.cc
```diff
@@ -0,0 +1,213 @@
...
+        EXPECT_VEC_SOFT_EQ(expected_polarization_x, polarization_x);
+        EXPECT_VEC_SOFT_EQ(expected_cos_polar, cos_polar);
+    }
+}
+
+//---------------------------------------------------------------------------//
+TEST_F(ScintillationGeneratorTest, stress_test)
+{
+    // Generate a large number of optical photons
+    dist_.num_photons = 1234567;
```
sethrj: That is a large number indeed!!

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -0,0 +1,52 @@
...
+        CELER_ASSERT(spec.size() > 0);
+
+        // Validity of scintillation component data
+        real_type total_yield{0};
+        for (auto const& comp : spec)
+        {
+            CELER_ASSERT(comp);
+            total_yield += comp.yield_prob;
+        }
+        CELER_ASSERT(soft_equal(real_type(1), total_yield));
```
sethrj, line 41: Missing `corecel/math/SoftEqual.hh`

sethrj (approved):
Nice work @whokion !

src/celeritas/optical/ScintillationGenerator.hh
whokion: Okay, moved it to the code.

test/celeritas/optical/ScintillationGenerator.test.cc
whokion: Reduce it to 123456 (~430ms --> 43ms for the stress test on my mac).

src/celeritas/optical/ScintillationParams.cc
whokion: Added.

whokion:
@sethrj All done.  Thanks!

sethrj:
Nice work! Cross referencing with #886 .

PR #1088, opened 2024-01-22
Define unit traits and support arbitrary RZ field map units

sethrj:
This improves the unit traits by adding an enum `UnitSystem` in the celeritas Types that corresponds to the configuration options. A traits visitor much like the one in ORANGE allows conversion from an enum to class data. The modifications to RZ field map input shows how the visitor is used to extract the value of the magnetic field or unit length from *another* unit system in the *native* unit system.

### DIFF
--- a/src/celeritas/Types.hh
+++ b/src/celeritas/Types.hh
@@ -10,6 +10,7 @@
 
 #include <cstdint>
 
+#include "celeritas_config.h"
 // IWYU pragma: begin_exports
 #include "corecel/Assert.hh"
 #include "corecel/Macros.hh"
@@ -80,6 +81,18 @@ using SubshellId = OpaqueId<struct Subshell_>;
 
 //---------------------------------------------------------------------------//
 // ENUMERATIONS
+//---------------------------------------------------------------------------//
+//! Unit system used by Celeritas
+enum class UnitSystem
+{
+    none,  //!< Invalid unit system
+    cgs,  //!< Gaussian CGS
+    si,  //!< International System
+    clhep,  //!< Geant4 native
+    size_,
+    native = CELERITAS_UNITS,  //!< Compile time selected system
+};
+
 //---------------------------------------------------------------------------//
 //! Interpolation type
 enum class Interp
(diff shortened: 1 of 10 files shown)


src/celeritas/Types.hh
```diff
@@ -172,6 +185,12 @@ struct StepLimit
 // HELPER FUNCTIONS (HOST)
 //---------------------------------------------------------------------------//
 
+// Get a string corresponding to a material state
```
amandalund:
```suggestion
// Get a string corresponding to a unit system
```

src/celeritas/field/RZMapFieldInputIO.json.cc
```diff
@@ -40,38 +43,95 @@ void from_json(nlohmann::json const& j, RZMapFieldInput& inp)
...
         {
-            convert_from_tesla = true;
+            CELER_LOG(warning)
+                << "Deprecated RZ field input units '" << ustr
+                << "': use SI units for length (m) and field (T) "
+                   "and set units to 'si'";
+            field_units = UnitSystem::si;
         }
-        else if (units == "native" || units == "gauss")
+        else if (ustr == "gauss" || Gauss::label() || ustr == "native")
```
amandalund:
```suggestion
        else if (ustr == "gauss" || ustr == Gauss::label() || ustr == "native")
```

test/celeritas/field/Fields.test.cc
```diff
@@ -47,8 +49,8 @@ TEST(CMSParameterizedFieldTest, all)
     CMSParameterizedField calc_field;
 
     int const nsamples = 8;
-    real_type delta_z = 25.0;
-    real_type delta_r = 12.0;
+    real_type delta_z = native_value_from(units::CmLength{25.0});
+    real_type delta_r = native_value_from(units::CmLength{12.0});
```
amandalund:
```suggestion
    real_type delta_z = from_cm(25.0);
    real_type delta_r = from_cm(12.0);
```

src/celeritas/field/RZMapFieldInputIO.json.cc
```diff
@@ -40,38 +43,95 @@ void from_json(nlohmann::json const& j, RZMapFieldInput& inp)
...
+                               << "' for \"_units\" field: " << e.what());
+            }
         }
     }
-    if (convert_from_tesla)
+    else
     {
-        using FieldTesla = Quantity<units::Tesla, double>;
+        auto msg = CELER_LOG(warning);
+        msg << "No units given in RZ field input: assuming CGS (cm, G)";
```
amandalund: I think this actually assumes CGS for length and SI for field units?

amandalund (commented):
Looks good @sethrj!

amandalund (approved):

PR #1089, opened 2024-01-23
Remove unused data from ImportElement/Material

sethrj:
These data aren't used by Celeritas; we reconstruct them from the important data (number density, etc.). Also units for some of the more obscure element data may have been wrong: Tsai radiation length seems to be number density rather than mass density?


stognini (commented):
The only one that may end up being useful soon is the material state, in case a muCF model asks for that info. If that's the case I'll re-add it in a few months.

stognini (approved):

sethrj:
The material state is still there; I just removed things like electron density which we recalculate anyway.

PR #1090, opened 2024-01-23
Change propagation fatal error to be recoverable

sethrj:
In #1085 the simple CMS test would occasionally hit an error due to a track in a field getting stuck on a cylindrical boundary with VecGeom. I saw this behavior sometimes in CMS2018 when debug assertions are turned on.

To allow those tests to pass, since there's no real workaround and those tests don't seem to hang, this changes the error from a fatal assertion to a nonfatal CPU-only "error" message.


amandalund (approved):
Looks good to me.

PR #1092, opened 2024-01-25
Add bounding box utilities for output and upcoming ORANGE work

sethrj:
This is a small PR for the pending object construction that:
- Moves the only host/device function in bbox utils to the main BoundingBox header file (the other functions are about construction and host output so maybe this is sensible? I'm open to argument though)
- Adds an `encloses` method for one bounding box enclosing another (all points in box A are inside box B)
- Adds stream operators to boundingbox for testing

### DIFF
--- a/src/orange/BoundingBoxUtils.hh
+++ b/src/orange/BoundingBoxUtils.hh
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file orange/BoundingBoxUtils.hh
+//! \brief Host-only utilities for bounding boxes
 //---------------------------------------------------------------------------//
 #pragma once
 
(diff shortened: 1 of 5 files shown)


src/orange/BoundingBoxUtils.hh
```diff
@@ -180,6 +163,24 @@ calc_intersection(BoundingBox<T> const& a, BoundingBox<T> const& b)
...
+ *
+ * All bounding boxes should enclose a "null" bounding box (there are no points
+ * in the null box, so no points are outside the big box). The null bounding
+ * box will enclose nothing but a bounding box.
+ */
+template<class T>
+inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
+{
+    auto axes = range(to_int(Axis::size_));
+    return all_of(axes.begin(), axes.end(), [&big, &small](int ax) {
```
elliottbiondo, line 178: Just curious why you unrolled for is_inside, but not here.

elliottbiondo (approved):

src/orange/BoundingBoxUtils.hh
sethrj: It's a little arbitrary I admit... but:
- `is_inside` is easier to write unrolled
- it's the only one used at runtime (running without optimization means it'll be a bit faster, a dumb reason)
- unrolling means not having to include the algorithms header

PR #1093, opened 2024-01-30
Add bounding zone

sethrj:
My plan for object construction (and possibly to be used for "safety" calculations) is to construct each CSG node with a "bounding zone", which is a pair of bounding boxes. The exterior bounding box is what we have presently: nowhere outside this box is inside a volume (i.e. the box entirely encloses the volume). For the interior bounding box, nowhere *inside* the box is *outside* that volume (i.e., the box is entirely enclosed in the volume). In the following cross section of a black wiggly shape, the green box is the interior and red is exterior:
bzone

The bounding zone can be "negated" just like a CSG object: in that case, everywhere *outside* the exterior box is known to be "inside", and everywhere *inside* the interior is known to be "outside". A double negation returns the original zone.


src/orange/BoundingBox.hh
```diff
@@ -222,27 +238,53 @@ BoundingBox<T>::BoundingBox(std::true_type, Real3 const& lo, Real3 const& hi)
...
+    return this->lower()[0] <= this->upper()[0]
+           && this->lower()[1] <= this->upper()[1]
+           && this->lower()[2] <= this->upper()[2];
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Reduce the bounding box's extent along an axis.
+ *
+ * This clips the bounding box along the given axis.
```
elliottbiondo: Detail line here doesn't provide any additional info

src/orange/BoundingBox.hh
```diff
@@ -222,27 +238,53 @@ BoundingBox<T>::BoundingBox(std::true_type, Real3 const& lo, Real3 const& hi)
...
-           && lower_[to_int(Axis::y)] <= upper_[to_int(Axis::y)]
-           && lower_[to_int(Axis::z)] <= upper_[to_int(Axis::z)];
+    return this->lower()[0] <= this->upper()[0]
+           && this->lower()[1] <= this->upper()[1]
+           && this->lower()[2] <= this->upper()[2];
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Reduce the bounding box's extent along an axis.
```
elliottbiondo: This is _conditionally_ shrinking the lower/upper bound along an axis, provided that the new position is results in a small bbox compared to the current position. I think this should be made clear in the doc string.

src/orange/BoundingBox.hh
```diff
@@ -222,27 +238,53 @@ BoundingBox<T>::BoundingBox(std::true_type, Real3 const& lo, Real3 const& hi)
...
+    {
+        p = ::celeritas::min(p, position);
+    }
+    points_[to_int(bnd)][to_int(axis)] = p;
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Intersect in place with a half-space.
+ * Increase the bounding box's extent along an axis.
```
elliottbiondo: Same comment as doc string above

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * This class partition space into "known inside" a volume, "maybe inside,
+ * maybe outside", and "known outside".
+ *
+ * - Known outside is the typical bounding box case: nowhere outside the
+ * bounding box is inside the volume; i.e., the box completely encloses the
+ * volume (but may also enclose parts that *aren't* volume).
+ * - Known inside can be used for safety distance calculation.
+ * - Flipping the two cases when applying a "not" operator to a CSG node: so
+ *   "known inside" becomes "known outside" and could potentially be flipped
```
elliottbiondo, line 47: Shouldn't `!(known inside) = maybe outside`?

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+/*!
+ * Exterior and interior CSG node bounding boxes.
+ *
+ * This class partition space into "known inside" a volume, "maybe inside,
+ * maybe outside", and "known outside".
+ *
+ * - Known outside is the typical bounding box case: nowhere outside the
+ * bounding box is inside the volume; i.e., the box completely encloses the
+ * volume (but may also enclose parts that *aren't* volume).
+ * - Known inside can be used for safety distance calculation.
```
elliottbiondo, line 45: You haven't actually explained what known inside is; it's clear to me due to your very nice diagram in the description of this MR, but I am not sure it would be clear without out. Maybe you could add an ASCII art image, or just say "the largest axes-aligned cuboid that fits entirely within the volume."

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+ * - Flipping the two cases when applying a "not" operator to a CSG node: so
+ *   "known inside" becomes "known outside" and could potentially be flipped
+ *   back in a subsequent simplification.
+ *
+ * The \c exterior box always encloses (or is identical to) \c interior in the
+ * bounding zone. If \c negated is \c false, then the exterior box is what we
+ * typically think of as a bounding box. If \c true, then even though the
+ * exterior bounding box is larger than the interior box, *all points inside
```
elliottbiondo: I don't understand what the size of the exterior bounding box has to do with this sentence. I think you can remove this clause.

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+ * | `~A | ~B` | `~A_x | ~B_x`  | `~(~A_i | ~B_i)`|
+ *
+ * The following set algebra can be used:
+ * - Involution: \verbatim ~~A <=> A \endverbatim
+ * - De Morgan's law 1: \verbatim ~A | ~B <=> ~(A & B) \endverbatim
+ * - De Morgan's law 1:\verbatim ~A & ~B <=> ~(A | B) \endverbatim
+ * - Set difference: \verbatim A & ~B <=> A - B \endverbatim
+ * - Negated set difference: \verbatim A | ~B <=> ~(B - A) \endverbatim
+ *
+ * When \c negated is set, the *inside* of the interior bounding box is "always
```
elliottbiondo: Already stated above

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+ * outside" and the *outside* of the exterior is "always inside".
+ *
+ * The default bounding zone places all points in the \em indeterminate zone:
+ * the exterior is the "universe set" (infinite) and the interior is the "empty
+ * set" (null).
+ */
+struct BoundingZone
```
elliottbiondo, line 95: `BoundingZone` is a somewhat confusing name, since "zone" evokes a singular region, which here, zone really refers to the partitioning of all space into inside, indeterminate, and outside regions. Maybe "BoundingPartitions"?

src/orange/orangeinp/detail/BoundingZone.cc
```diff
@@ -0,0 +1,178 @@
...
+ * different negations:
+ *
+ * | Input     | Interior     | Exterior    | Negated  |
+ * | ------    | ------------ | ----------- | -------- |
+ * | `A & B`   | `A_i & B_i`  | `A_x & B_x` | false    |
+ * | `A & ~B`  | `A_i - B_x`  | `A_x - B_i` | false    |
+ * | `~A & B ` | `B_i - A_x`  | `B_x - A_i` | false    |
+ * | `~A & ~B` | `A_i | B_i`  | `A_x | B_x` | true     |
+ *
+ * The above algebra for unions and intersections does *not* necessarily
```
elliottbiondo, line 80: They may produce orthogonal polyhedra, correct? As in polyhedra consisting of only right angles, but with concave regions?

src/orange/orangeinp/detail/BoundingZone.cc
```diff
@@ -0,0 +1,178 @@
...
+ * produce boxes. The interior has to "shrink" and the exterior has to "grow"
```
elliottbiondo: To be clear, they shrink or grow to produce the bounding box of the orthongonal polyhedron, correct?

src/orange/surf/SurfaceClipper.hh
```diff
@@ -11,85 +11,66 @@
...
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Truncate a bounding box to its intersection with a surface interior.
+ * Truncate a bounding zone using a surface.
+ *
+ * This \em reduces the size of inner and outer bounding boxes to fit a
+ * surface. The \c interior bounding box is entirely \em inside the surface,
```
elliottbiondo: The interior bounding box is _adjusted_ to be entirely inside..

src/orange/surf/SurfaceClipper.hh
```diff
@@ -11,85 +11,66 @@
...
+ * Shrinking bounding boxes will accelerate "distance to in" and "distance
+ * to out" calculations.
  */
 class SurfaceClipper
 {
   public:
-    // Construct with a reference to the bbox being clipped
-    explicit inline SurfaceClipper(BBox* bbox);
+    // Construct with interior and exterior bounding boxes
+    explicit SurfaceClipper(BBox* interior, BBox* exterior);
```
elliottbiondo, line 40: Why not construct from a SurfaceZone, which stores these BBoxes?

src/orange/surf/SurfaceClipper.cc
```diff
@@ -0,0 +1,168 @@
...
+//!\endcond
+
+//---------------------------------------------------------------------------//
+/*!
+ * Clip the bounding boxes to an axis-aligned cylinder.
+ */
+template<Axis T>
+void SurfaceClipper::operator()(CylCentered<T> const& s) const
+{
+    return (*this)(CylAligned<T>{s});
```
elliottbiondo, line 63: Does this return within a void function serve any purpose?

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * The following table shows the semantics of set operations on the bounding
+ * zones.  Below, \c ~zone means a zone with the \c negated flag set; \c ~box
+ * means "complement of a bbox": all space *except*
+ * inside box; and the union operator implies the (possibly disconnected and
+ * probably not box-shaped!) union of the two boxes.
+ *
+ * | Zone      | known inside   | known outside   |
+ * | ------    | ------------   | -------------   |
+ * | `A`       | `A_i`          | `~A_x`          |
```
elliottbiondo, line 75: `x` for exterior might be confusing notation in downstream code, but fine within the table

elliottbiondo (changes requested):

src/orange/orangeinp/detail/BoundingZone.hh
sethrj: That's what's tricky about the `negate`; it flips the intents of the "exterior" and "interior" boxes. I'm trying to add a diagram. But basically the region around a shape is always "maybe outside"; if you negate the shape, then the interior bounding box becomes "always outside" and the exterior bounding box is "always ~~outside~~ inside".

src/orange/orangeinp/detail/BoundingZone.hh
sethrj: Yep, adding ascii art.

src/orange/orangeinp/detail/BoundingZone.hh
sethrj: It's trying to emphasize that the exterior box is always the "larger" one, it doesn't mean "outside the volume". The exterior always encloses the interior box.

src/orange/orangeinp/detail/BoundingZone.hh
sethrj: I was thinking it's a zone that partitions "inside" from "outside", rather than a *box* that partitions inside from outside. Wouldn't partitions be potentially confusing with BIH/BVH?

src/orange/orangeinp/detail/BoundingZone.cc
sethrj: Yeah exactly. That's a good way of putting it. And in edge cases it could produce two disconnected boxes.

src/orange/orangeinp/detail/BoundingZone.hh
elliottbiondo: > and the exterior bounding box is "always outside".

NOT always inside?? I can't tell if this is a typo or if I have a fundamental misunderstanding of what is going on here.

src/orange/surf/SurfaceClipper.hh
sethrj: Mostly because the bounding zone is a detail of orangeinp and I wasn't sure I wanted to spread it around... I think once the construction work is done we can look at reorganizing the files more sensibly (surface clipper should maybe go into an `orangeinp` construction detail?) and then simplify this.

src/orange/surf/SurfaceClipper.cc
sethrj: Not in this case, no, but it can be a good way of enforcing that you don't discard the result of another function. Even for the void case it can say "I'm just going to pass this off to someone else".

src/orange/orangeinp/detail/BoundingZone.cc
sethrj: Yep, documentation added.

src/orange/orangeinp/detail/BoundingZone.hh
sethrj: Exterior and interior are about the size of the boxes, whether one box encloses another. "inside" and "outside" are about the CSG region.

In the typical case, "inside" is within the interior bounding box. If you negate the zone, which is when you take the complement of the CSG region, then everything *not enclosed by the exterior bounding box* is *inside* the CSG region, and everything *enclosed by the interior bounding box* is known to be *not inside* the CSG region.

In the degenerate case where the CSG region is a box, the interior and exterior boxes are the same, so `negate` just means everything *not* enclosed is inside the CSG region.

src/orange/orangeinp/detail/BoundingZone.hh
elliottbiondo: Okay I see what you are saying. To me, it's clear enough that external means "external to the volume", and negation doesn't change this relationship, but I understand how that might not be obvious to others.

src/orange/orangeinp/detail/BoundingZone.hh
elliottbiondo: Okay, so the "zone" is technically region between the two bounding boxes then? I like that a lot actually --- I think you should state that explicitly in the doc string.

elliottbiondo (approved):

PR #1094, opened 2024-01-31
Fix position check after propagating

amandalund:
Sorry, missed this in the review of #1090.


sethrj (commented):
Whoops, I guess the testing wouldn't "fail" but we'd just get a billion error messages 😂 sorry.

Do you think you (or I) could add a test for the error message? We could add one to the along-step test, use the scoped logger, and make a giant (km) cube of lead and start the particle far enough away and low enough energy that `x + step == x`.

amandalund:
That sounds good, I can try to put together a test.

sethrj:
@amandalund I reproduced locally the CI failure you're seeing:
```
122: C++ exception with description "/Users/seth/Code/celeritas-temp/src/celeritas/ext/GeantGeoParams.cc:250:
122: celeritas: internal assertion failed: lv" thrown in the test body.
```

Pretty easy fix; this is just to geant4 not doing cleanup correctly. We just have to add the correct `FILTER` arguments in the testing framework so geant4 internals are all reset.

sethrj (approved):
Thanks and nice test!

amandalund:
Ahh, thanks for reproducing and fixing that failure @sethrj!

amandalund:
Oops I think I needed to update the test to work with other unit systems...

sethrj:
WTF, why did it merge with the checks unpassed...

EDIT: grrr... https://emmer.dev/blog/skippable-github-status-checks-aren-t-really-required/

sethrj:
Also can you add
```c++
else
{
    EXPECT_TRUE(scoped_log_.empty()) << scoped_log_;
}
```
to the `if CELERITAS_DEBUG`? And should we also have a similar "scoped long" on a simpler case that's known to be correct? 😬

amandalund:
Will do!

PR #1095, opened 2024-02-01
Refactor Geant4 utilities in preparation for g4vg

sethrj:
This removes the `celeritas/io` dependency from the Geant4 volume visitor, and it breaks up the `GenericGeoTestBase` instantiations into separate files, combining them with the checked geo instantiation.


amandalund (approved):
Looks good @sethrj!

PR #1097, opened 2024-02-02
Expand position check test and fix for other unit systems

amandalund:
See #1094.


sethrj (approved):
Thanks!

sethrj:
I think the CI is borking the ROCm jobs that are on hardware:
```
unknown file: Failure
C++ exception with description "/var/jenkins/workspace/celeritas_PR-1097/src/corecel/sys/KernelAttributes.hh:75:
celeritas: device error: hipFuncGetAttributes(&attr, reinterpret_cast<void const*>(func)) failed:
    shared object initialization failed" thrown in the test body.
```

PR #1098, opened 2024-02-02
Fix non-sticky CMake options

sethrj:
When an upstream project sets a local default value for any of the Celeritas options with a predetermined list (RNG, geometry, units, ...) it creates a cache entry with the user-selected value. Previously the default cache entry would be blank, so that the first configure respects the default value but subsequent configures (if the upstream uses a "set default" mechanism) would reset to blank. This fixes the issue by writing the current value as the default cache value.


amandalund (approved):
Thanks @sethrj!

PR #1099, opened 2024-02-02
Reorganize geometry wrappers into a new "geocel" directory

sethrj:
This allows downstream codes like AdePT to use the Celeritas geometry conversion facilities without having to build any physics code whatsoever.

This creates a new geometry-only library that sits above corecel but below celeritas and ORANGE. It fixes a couple of long-standing minor organizational problems (having both "orange types" and "types" in the same dir, having the geo interface sit in ORANGE, etc.).

- Split the test directory into several subdirectories
- Moved geant utilities and geometry into `geocel/g4`
- Moved vecgeom utilties and geometry into `geocel/vecgeom`
- Moved g4vg into `geocel/g4vg`
- Moved params interface, output interface into `geocel`
- Added a "length units" header inside `geocel` to do conversions


drbenmorgan:
All looks good to me @sethrj, modulo fixing the build/link issues, but structure/layout makes sense!

drbenmorgan (approved):

pcanal:
> modulo fixing the build/link issues, 

Working on it.

PR #1100, opened 2024-02-02
Fix unit tests when using SI

sethrj:
This fixes the last couple of broken tests on my machine. Since we don't have a use case for SI (do any other codes we could couple against use it?) I don't think we need it under CI, but it's good to fix what we can.
- Units in vacuum permittivity test were wrong
- Missing units in track sort test
- Lowered tolerance in along-step test
- Disabled a Geant4 geometry wrapper test that's bizarrely skipping over a sphere when reinitializing


amandalund (approved):
Thanks @sethrj! demo-rasterizer is failing for me, but I confirmed that all other tests pass.
```console
1/1 Test #209: app/demo-rasterizer ..............***Failed    0.54 sec
status: Loading VecGeom geometry from GDML at /home/alund/celeritas_project/celeritas/app/data/two-boxes.gdml
info: CUDA stack size was changed from 1024 to 8192; restoring to original values
status: Tracing geometry
/home/alund/celeritas_project/celeritas/app/demo-rasterizer/demo-rasterizer.cc:173: critical: caught exception: /home/alund/celeritas_project/celeritas/app/demo-rasterizer/RDemoKernel.cu:125:
celeritas: CUDA error: cudaDeviceSynchronize() failed:
    unspecified launch failure
```

PR #1101, opened 2024-02-02
Optimize log message code generation

sethrj:
While testing the contents of `g4vg.dylib` with `nm` to  make sure no Celeritas physics was included, I found that template instantiations for the logger message on `char const[N]` were being saved as actual function calls. This is pretty silly because those functions unroll effectively to `put_char(ptr, N)`, so there's not going to be any code size gain by *not* inlining those.

The second optimization is that I've marked as "unlikely" that the log level is high enough to print, moving the printing code to a "cold" section. This improves instruction cache locality in the typical case where a logger message is ignored.

I've also replaced `std::string` with `std::string_view` and made better use of move semantics to further reduce the amount of code needed to create and destroy a logging message, especially when it's not printed.

----

For reference, given the code:
```c++
void write_message(int i, char const* s)
{
    CELER_LOG(debug) << "Error: expected " << i << " to be " << s
                     << "(but double is " << 1.234 << ')';
}
```

With the new code, there are 15 instructions between "get the world logger" and the function exit. With the old code, there's ~120 (but the "not logging" case skips some of those). The old code also has several calls to `delete` due to using `string` rather than `string_view`.


src/corecel/io/detail/LoggerMessage.cc
```diff
@@ -21,44 +21,37 @@ namespace detail
...
-LoggerMessage::~LoggerMessage()
+void LoggerMessage::destroy_impl() noexcept
 {
-    if (os_)
+    try
     {
-        try
-        {
-            auto& os = dynamic_cast<std::ostringstream&>(*os_);
+        auto& os = dynamic_cast<std::ostringstream&>(*os_);
```
pcanal: (comment on pre-existing code).
Given line 32, os is guaranteed to be either nullptr or of type std::ostringstream or am I missing something?

If I am not missing something, why not simply declare the data member:
```
std::unique_ptr<std::ostringstream> os_;
```
?

sethrj: No I think you're right... I'm guessing this was to "hide" the `<sstream>` include? Seems a bit excessive to do that...

pcanal: Furthermore ... it is already included indirectly by `<iostream>` :)

sethrj: Not in libcxx

pcanal: Humm .. Compiler Explorer lead me astray there: https://godbolt.org/z/4W5qo6j3M

pcanal (approved):
LGTM

PR #1102, opened 2024-02-06
Fix CMake for linking CUDA RDC for executables

pcanal:
* Properly trigger rdc linking in executable.

    Also deal with the new to explicit list libraries on link line

Performance improvement:

* cmake: Cache dependencies list
* cmake: Avoid spurrious foreach loop (foreach when giving a not-found item will still goes through one iteration)
* cmake: Remove redundant calls


sethrj (approved):
LGTM!

sethrj:
Confirmed the VecGeom-enabled configuration time went down from 15 seconds to 4 seconds on wildstyle (there may be ORNL spyware slowing down memory reallocation or something FWIW). Confirmed that the code builds and passes when merged into #1099 . Thanks!!

PR #1103, opened 2024-02-08
Support importing optical properties from Geant4

amandalund:
This adds functionality for importing optical material properties from Geant4. Currently only the properties used for scintillation and Cerenkov are imported, but we can add more as needed. Even though all the properties are stored together in the `G4MaterialPropertiesTable` in Geant4, from what I can tell most are only used by a single model (refractive index is an exception). We may end up refactoring or replacing a lot of this as we get a clearer understanding of how we ultimately want to handle the optical property data (e.g. @whokion's idea of having an internal database), but at least this should give us a starting point.


src/celeritas/ext/GeantImporter.cc
```diff
@@ -327,6 +366,79 @@ std::vector<ImportElement> import_elements()
...
+        auto const* mpt = material->GetMaterialPropertiesTable();
+        if (!mpt)
+        {
+            continue;
+        }
+        ImportOpticalMaterial optical;
+
+        // Save common properties
+        optical.properties.refractive_index
+            = get_vector_property(*mpt, "RINDEX");
```
sethrj: This code and below might be cleaner if you have a tiny helper struct `MatPropGetter` that has `mpt` as member data and `scalar(string)`, `vector(string)`, `scalar(string, int)`?

whokion:
Looks great!  A couple of general comments:
- The `ext_GeantImporter` test is broken (not only for `LarSphere`, but also for other test cases too).  It may be good to fix these first, especially for `LarSphere` as we are adding material properties that are not defined in Geant4 (i.e., LAMBDAMEAN[1-3] and LAMBDASIGMA[1-3]) -in principle, they should be properly added with `createNewKey=true` through the method, [AddConsProperty](https://geant4.kek.jp/lxr/source/materials/src/G4MaterialPropertiesTable.cc#L434), but need to be checked whether the gdml support automatically takes care of the importing under the hood.
- It may be better to use `OpticalPropertyParams::from_import(ImportData const& data)` in `Cerenkov.test.cc`  and `ScintillationGenerator.test.cc` instead of the manual input.
- It would be good to add, at least, `ABSLENGTH` as @hhollenb plans to implement an absorption process/interactor soon.  Even better to add  `RS_SCALE_FACTOR` and `ISOTHERMAL_COMPRESSIBILITY` (for Rayleigh) so that we may cover most of material properties for the earlier test cases of potential optical applications. 

I will give more detail comments, if any, later.  Thanks again for this important ground work!

amandalund:
Thanks @whokion! I'll add the properties you mentioned that we'll need in the near term.
> The ext_GeantImporter test is broken (not only for LarSphere, but also for other test cases too)

What failures are you seeing? It's passing on the CI and working locally for me as well.

> they should be properly added with createNewKey=true through the method, [AddConsProperty](https://geant4.kek.jp/lxr/source/materials/src/G4MaterialPropertiesTable.cc#L434), but need to be checked whether the gdml support automatically takes care of the importing under the hood

Yes, the GDML reader does add the properties [this way](https://geant4.kek.jp/lxr/source/persistency/gdml/src/G4GDMLReadMaterials.cc#L868).

> It may be better to use OpticalPropertyParams::from_import(ImportData const& data) in Cerenkov.test.cc and ScintillationGenerator.test.cc instead of the manual input.

I think that would require adding either a Geant4 or ROOT dependency to the optical tests, right? I do think it would be good to test the `Params::from_import()` somewhere though, and maybe the optical material ID as well.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -29,19 +57,28 @@ ScintillationParams::ScintillationParams(ScintillationInput const& input)
...
+            comp[i].lambda_mean = comp_inp[i].lambda_mean;
+            comp[i].lambda_sigma = comp_inp[i].lambda_sigma;
+            comp[i].rise_time = comp_inp[i].rise_time;
+            comp[i].fall_time = comp_inp[i].fall_time;
+            norm += comp_inp[i].yield;
+        }
+        for (auto i : range(comp.size()))
+        {
+            comp[i].yield_prob = comp_inp[i].yield / norm;
+            CELER_ASSERT(comp[i]);
```
sethrj: Since this depends on user-provided data, I think it should be a `CELER_VALIDATE` and possibly include the data that could be invalid (or worry about details later; for now just validate).

test/celeritas/ext/GeantImporter.test.cc
```diff
@@ -1293,6 +1303,91 @@ TEST_F(OneSteelSphereGG, physics)
...
+                                           1.527,
+                                           1.5545,
+                                           1.793,
+                                           1.7826,
+                                           1.6642,
+                                           1.5545,
+                                           1.4536,
+                                           1.4536};
+    EXPECT_VEC_EQ(expected_ri_x, properties.refractive_index.x);
+    EXPECT_VEC_EQ(expected_ri_y, properties.refractive_index.y);
```
sethrj: If all these properties have four digits of precision maybe we should store them as `float` 😆 (just kidding, that would be premature optimization... we can worry about that later).

src/celeritas/optical/OpticalPropertyParams.cc
```diff
@@ -0,0 +1,87 @@
...
+        {
+            // No refractive index data for this material
+            refractive_index.push_back(grid);
+            continue;
+        }
+
+        // In a dispersive medium the index of refraction is an increasing
+        // function of photon energy
+        CELER_ASSERT(is_monotonic_increasing(make_span(ri_vec.x)));
+        CELER_ASSERT(is_monotonic_increasing(make_span(ri_vec.y)));
```
sethrj: Should be `CERER_VALIDATE` since it's fed straight from user data

whokion:
@amandalund Okay, I forgot to filter the test while running ext_GeantImporter and did not pay attention to the error message.  Sorry about the false alarm.  Regarding to optical tests, you are right that we should not have Geant4 and root dependency there and `LarSphere.optical` should be good enough to check input material properties.  So, let's keep optical tests as they are. Thanks for clarifying!

sethrj (commented):
Looks great!

src/celeritas/optical/OpticalPropertyParams.cc
```diff
@@ -0,0 +1,87 @@
...
+
+        grid.grid = reals.insert_back(ri_vec.x.begin(), ri_vec.x.end());
```
whokion, line 81: It is just a convention, but isn't the (imported) input vector of a refractive index as a function of photon wavelength (tabulated [here](https://geant4.kek.jp/lxr/source/materials/include/G4OpticalMaterialProperties.hh), for an example)?  If so,  should wavelength be converted to energy here or is it converted another place?  Again for the later discussion, it would be nice that celeritas supports the functional input for the index of refraction as a function of wavelength (i.e., the Sellmeier equation) so that it can be calculated on-the-fly instead of a grid calculation.

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -0,0 +1,78 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Store material-dependent properties for scintillation spectrum.
+ */
+struct ImportScintComponent
+{
+    double yield{};  //!< Yield for this component
```
whokion, line 22: Just for a comment for further discussions later if we support or migrate to a DB approach in the future (so, no need to change anything for now):  Providing `yield` for each   component is somewhat awkward (i.e., following the Geant4 convention) as it will be summed and divided by the total later (so it should be just the `fraction` (that is calculated in `ScintillationParams`) to the actual total yield which will be determined later).  In the scintillation process (or whatever @stognini calls it), the total number of photons will be estimated with the global yield and resolution based on either Poisson or Gaussian, then the number of photons for each component will be equal to [the number of total photons x yield_fraction] (logically the end result is same, but the fixed input yield for each component is somewhat physically meaningless).  We may change the convention later.

test/celeritas/ext/GeantImporter.test.cc
```diff
@@ -1293,6 +1303,91 @@ TEST_F(OneSteelSphereGG, physics)
...
+                                           1.4955,
+                                           1.4988,
+                                           1.5264,
+                                           1.6185,
+                                           1.6176,
+                                           1.527,
+                                           1.5545,
+                                           1.793,
+                                           1.7826,
+                                           1.6642,
```
whokion: In the earlier, there was a check that the index of refraction is monotonically increasing.  Is this intentionally try to make the check failure?

whokion (commented):
@amandalund Left a couple of minor comments and questions.  Other than that, it is a very nice addition for supporting material properties for optical photon simulation.  Excellent!

src/celeritas/ext/GeantImporter.cc
amandalund: Good idea, thanks! I'd been trying and failing to make this cleaner.

src/celeritas/optical/OpticalPropertyParams.cc
amandalund: I think when the vectors are added to the material properties table or imported from the GDML the grid is already in energy units (e.g. those tabulated values are converted to energy before being returned). Adding support in the future for the functional input is a good idea and would be a nice improvement here.

test/celeritas/ext/GeantImporter.test.cc
amandalund: Good catch! No, this was not intentional :)

stognini:
Looks pretty great. One extra bit that could be useful is to add a `print_optical_materials(...)` to `celer-dump-data.cc` for convenience.

amandalund:
Good point @stognini, I'll add that.

sethrj (approved):
Nice work @amandalund !

app/celer-dump-data.cc
```diff
@@ -614,6 +614,71 @@ void print_atomic_relaxation_data(
...
+)gfm";
+
+    for (auto const& [key, val] : iom)
+    {
+        auto const& scint = val.scintillation;
+        cout << POM_STREAM_SCALAR(key, scint.yield);
+        cout << POM_STREAM_SCALAR(key, scint.resolution_scale);
+        for (auto const& scint_comp : scint.components)
+        {
+            cout << POM_STREAM_SCALAR(key, scint_comp.yield);
```
whokion: Use the same key [name] for all components?

app/celer-dump-data.cc
```diff
@@ -614,6 +614,71 @@ void print_atomic_relaxation_data(
...
+    "| " << setw(11) << ID << " | " << setw(24) << #NAME << " | " << setw(7) \
+         << NAME.x.size() << " | " << setw(9) << ""                          \
+         << " | (" << setprecision(3) << setw(10) << NAME.x.front() << ", "  \
+         << setprecision(3) << setw(10) << NAME.y.front() << ") -> ("        \
+         << setprecision(3) << setw(10) << NAME.x.back() << ", "             \
+         << setprecision(3) << setw(10) << NAME.y.back() << ") |\n";
+    cout << R"gfm(
+# Material optical property data
+
+| Material ID | Property                 | Size    | Scalar    | Vector endpoints (MeV, value)                        |
```
whokion: Print unit of Property?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -327,6 +368,102 @@ std::vector<ImportElement> import_elements()
...
+                                "SCINTILLATIONTIMECONSTANT",
+                                comp_idx,
+                                ImportUnits::time);
+            if (comp)
+            {
+                optical.scintillation.components.push_back(comp);
+            }
+        }
+
+        // Save Rayleigh properties
```
whokion, line 443: Very good! Thanks for adding these extra properties.

test/celeritas/ext/GeantImporter.test.cc
```diff
@@ -1293,6 +1309,89 @@ TEST_F(OneSteelSphereGG, physics)
...
+    EXPECT_EQ(OpticalMaterialId{0},
+              materials->get(lar_id).optical_material_id());
+
+    real_type const tol = this->comparison_tolerance();
+
+    // Check scintillation optical properties
+    auto const& scint = optical.scintillation;
+    EXPECT_TRUE(scint);
+    EXPECT_EQ(1, scint.resolution_scale);
+    EXPECT_EQ(5, scint.yield);
```
whokion: Is this yield/MeV?  If so, may this be changed 50000 to be realistic for LAr (i.e, from input gdml)?

test/celeritas/ext/GeantImporter.test.cc
```diff
@@ -1293,6 +1309,89 @@ TEST_F(OneSteelSphereGG, physics)
...
+    auto const& rayleigh = optical.rayleigh;
+    EXPECT_TRUE(rayleigh);
+    EXPECT_EQ(1, rayleigh.scale_factor);
+    EXPECT_REAL_EQ(0.024673059861887867,
+                   rayleigh.compressibility * units::gram
+                       / (units::centimeter * units::second * units::second));
+    EXPECT_EQ(11, rayleigh.mfp.x.size());
+    EXPECT_DOUBLE_EQ(1.55e-06, rayleigh.mfp.x.front());
+    EXPECT_DOUBLE_EQ(1.55e-05, rayleigh.mfp.x.back());
+    EXPECT_REAL_EQ(32142.9, to_cm(rayleigh.mfp.y.front()));
```
whokion, line 1376: Where does this value come from? (Just curious as it seems to be too long?)

test/celeritas/ext/GeantImporter.test.cc
```diff
@@ -1293,6 +1309,89 @@ TEST_F(OneSteelSphereGG, physics)
...
+    EXPECT_REAL_EQ(54.6429, to_cm(rayleigh.mfp.y.back()));
+
+    // Check absorption optical properties
+    auto const& abs = optical.absorption;
+    EXPECT_TRUE(abs);
+    EXPECT_EQ(2, abs.absorption_length.x.size());
+    EXPECT_DOUBLE_EQ(1.3778e-06, abs.absorption_length.x.front());
+    EXPECT_DOUBLE_EQ(1.55e-05, abs.absorption_length.x.back());
+    EXPECT_REAL_EQ(86.4473, to_cm(abs.absorption_length.y.front()));
+    EXPECT_REAL_EQ(0.000296154, to_cm(abs.absorption_length.y.back()));
```
whokion, line 1386: Same here. Seems too short?

test/geocel/data/lar-sphere.gdml
```diff
@@ -0,0 +1,112 @@
...
...359753784 3.5374e-06 1.2300572815744 3.5927e-06 1.2303953408143 3.6509e-06 1.2307581975201 3.7099e-06 1.2311340756410 3.7708e-06 1.2315309395240 3.8350e-06 1.2319584795948 3.9001e-06 1.2324030531751 3.9675e-06 1.2328742967333 4.0373e-06 1.2333745182601 4.1109e-06 1.2339164497558 4.1858e-06 1.2344833000133 4.2636e-06 1.2350878388706 4.3457e-06 1.2357460538231 4.4296e-06 1.2364381478993 4.5167e-06 1.2371803058492 4.6091e-06 1.2379930862779 4.7035e-06 1.2388529939487 4.8019e-06 1.2397811112689 4.9044e-06 1.2407852902585 5.0135e-06 1.2418956339948 5.1254e-06 1.2430824604497 5.2425e-06 1.2443774409398 5.3673e-06 1.2458225245477 5.4958e-06 1.2473825079216 5.6305e-06 1.2491028277676 5.7748e-06 1.2510450704786 5.9238e-06 1.2531685960320 6.0806e-06 1.2555430061400 6.2461e-06 1.2582127677221 6.4241e-06 1.2612929591155 6.6090e-06 1.2647427641559 6.8048e-06 1.2687046921432 7.0166e-06 1.2733876575744 7.2378e-06 1.2787793729478 7.4734e-06 1.2851700143090 7.7297e-06 1.2930098842738 7.9990e-06 1.3024417515349 8.2877e-06 1.3142220339702 8.5981e-06 1.3293224516058 8.9390e-06 1.3497665943227 9.3011e-06 1.3776746680090 9.6938e-06 1.4188784984198 1.0129e-05 1.4873838729402 1.0597e-05 1.6167515615703"/>
```
whokion, line 5: Where does this data come from? (looks very reasonable values, but just curious about the reference) - probably may add the reference in the test.

whokion (approved):
Thanks for addressing all review comments - left a couple of minor comments and questions.  Very nice and good to merge!

test/celeritas/ext/GeantImporter.test.cc
amandalund: Most of these values (including this Rayleigh mfp) I pulled from [this Geant4 example](https://geant4.kek.jp/lxr/source/examples/advanced/CaTS/gdml/LArTPC.gdml). I'll include that reference somwhere.

test/celeritas/ext/GeantImporter.test.cc
amandalund: From the same example, except I only used the endpoints in the test gdml instead of copying the whole absorption length vector. Since we're only using the input to test that the values are imported correctly right now I wasn't too worried about the accuracy, but it would be good to have realistic values if possible.

test/geocel/data/lar-sphere.gdml
amandalund: These I actually got from [this refractive index database](https://refractiveindex.info/?shelf=main&book=Ar&page=Grace-liquid-90K), with these particular values calculated using this reference:

E. Grace, A. Butcher, J. Monroe, J. A. Nikkel. Index of refraction, Rayleigh scattering length, and Sellmeier coefficients in solid and liquid argon and xenon, [Nucl. Instr. Meth. Phys. Res. A 867, 204-208 (2017)](https://doi.org/10.1016/j.nima.2017.06.031)

I'll make a note in the test.

app/celer-dump-data.cc
amandalund: Not sure what you mean? Currently they all have the same `key` (material ID) and the names are the same for each component (i.e., there is no `1`, `2`, `3` suffix added).

test/geocel/data/lar-sphere.gdml
whokion: Great!  There are even Sellmeier coefficients from the link which can be used for a functional input later.

app/celer-dump-data.cc
whokion: I meant the same name - can you add the suffix easily or is it intentionally dropped?

amandalund: I can add it.

test/celeritas/ext/GeantImporter.test.cc
whokion: Humm... It seems that I am one of authors of the example^^.  So, should blame myself.

amandalund:
Thanks for the helpful feedback! @whokion I think I've addressed all your comments now.

whokion:
@amandalund Thanks for taking into account all comments - looks great!  @sethrj You can merge this MR if there is no additional comment. Thanks.

PR #545, opened 2022-11-11
Sync the development of the CMake support for multi library Relocatable Device [Cuda] Code

pcanal:
This merges into the Celeritas repository the local development and those made in

https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/822?commit_id=5e465ab2f42569b395691a4a19d4ddb47f96579d

which became and was extended in:

https://github.com/drbenmorgan/CudaRDCExercise

In particular this fixes:

 - build of recently added tests
 - Ability to build project using the Celeritas libraries.

Note: Please do not squash those commits if possible as they help with the synchronization.

PR #1104, opened 2024-02-12
Split CMake RDC utilities from Celeritas macros

pcanal:


cmake/CeleritasUtils.cmake
```diff
@@ -266,45 +304,56 @@ endfunction()
...
+    set_source_files_properties(
+      ${_cuda_sources}
+      PROPERTIES LANGUAGE HIP
+    )
+  endif()
+# Potential speed-up (if it is a significant gain should it be applied to the
+# other calls to?)
+#  if(NOT CELERITAS_USE_VecGeom OR NOT CMAKE_CUDA_COMPILER OR NOT _cuda_sources)
+#    add_library(${target} ${ARGN})
+# else()
```
sethrj: Yes, I think we should fall through in non-vecgeom cases, here and the other macros, because it will produce fewer edge cases if we're trying to build on some obscure system or with an unexpected cmake change, etc.

cmake/CeleritasUtils.cmake
```diff
@@ -266,45 +304,56 @@ endfunction()
...
-      celeritas_strip_alias(_tgt ${_tgt})
-      set_target_properties(${_tgt} PROPERTIES POSITION_INDEPENDENT_CODE ON)
-    endif()
-  endif()
+  cuda_rdc_add_library(Celeritas::${target} ALIAS ${target})
 
   # Build all targets in lib/
-  set_target_properties(${_targets} PROPERTIES
+  cuda_rdc_set_target_properties(${target} PROPERTIES
+    POSITION_INDEPENDENT_CODE ON   # Technically we probably only need it on the OBJECT library
```
sethrj: Why was this added? PIC should only be added when it's asked for (or if, as you say, we're building object code to be included in a shared library). Nothing in Celeritas itself requires it; it's only needed for ROOT, etc.

sethrj (changes requested):

cmake/CeleritasUtils.cmake
pcanal: It was already there.

pcanal: It was introduced by 7ea96499838f8a006fa91f6374fcc79d4e60d4e3 but thanks to the squashing it is unclear why it is there.

sethrj: I know why it's required for object library, but it looks like this refector removed it from an "if (object library)" condition.

sethrj: PS you can look at the unsquash commits: https://github.com/celeritas-project/celeritas/pull/545/commits
PPS some aspects of this PR would've been easier to review with a separate granular commit for "find replace CELERITAS_RDC"

pcanal: Yes, that was a thinko.  The thinking was that if there is an object library then it is used to build the rest and so it does not hurt to have it applied on all ... but the `if` got left out :(

pcanal: > PPS some aspects of this PR would've been easier to review with a separate granular commit for "find replace CELERITAS_RDC"

Sorry.  My side-view of things (through the source repository) actually has this commit in per se so I did not notice.  I can easily rewrite the history if you prefer.

cmake/CeleritasUtils.cmake
pcanal: One option for this is 007ac62c4c0473c95ecc0ae256d9c3b97f41e374 or we can add more 5 liners wrappers ... neither is very attractive.

sethrj: I think 5 liners is fine.

cmake/CeleritasUtils.cmake
sethrj: Source repository? Nah, don't bother about it. If you really want you could
```console
$ git reset $(git commit-tree -m "Merge other history" -p HEAD -p other_sha HEAD^{tree})
```

pcanal: Source repository: https://github.com/pcanal/CudaRDCExercise :)

sethrj: Oh gotcha. Definitely don't worry about the history then.

cmake/CeleritasUtils.cmake
```diff
@@ -266,43 +318,82 @@ endfunction()
...
-  install(TARGETS ${_targets}
-    EXPORT celeritas-targets
-    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
-    COMPONENT runtime
-  )
+function(celeritas_target_link_libraries)
+  if(NOT CELERITAS_USE_VecGeom)
+    cuda_rdc_target_link_libraries(${ARGV})
+  endif()
```
sethrj:
```suggestion
  if(NOT CELERITAS_USE_VecGeom)
    target_link_libraries(${ARGV})
  else()
    cuda_rdc_target_link_libraries(${ARGV})
  endif()
```

cmake/CeleritasUtils.cmake
```diff
@@ -266,43 +318,82 @@ endfunction()
...
+endfunction()
+
+#-----------------------------------------------------------------------------#
+
+function(celeritas_target_include_directories)
+  if(NOT CELERITAS_USE_VecGeom)
+    cuda_rdc_target_include_directories(${ARGV})
+  endif()
```
sethrj:
```suggestion
  if(NOT CELERITAS_USE_VecGeom)
    target_include_directories(${ARGV})
  else()
    cuda_rdc_target_include_directories(${ARGV})
  endif()
```

sethrj (changes requested):

cmake/CeleritasUtils.cmake
pcanal: yep, I managed to mis-typed those :(

sethrj (approved):
LGTM, hope the CI passes :) Thanks @pcanal !

cmake/CeleritasLibrary.cmake
```diff
@@ -7,869 +7,10 @@
...
-    endif()
-  endif()
-  list(REMOVE_DUPLICATES ${outlist})
-  set_target_properties(${target} PROPERTIES CUDA_RDC_CACHED_LIB_DEPENDENCIES "${${outlist}}")
-  set(${outlist} ${${outlist}} PARENT_SCOPE)
-endfunction()
+include(CeleritasUtils.cmake)
 
-#-----------------------------------------------------------------------------#
+message(AUTHOR_WARNING "Deprecated file. Please use CeleritasUtils.cmake")
```
sethrj: Actually, don't we want people to only include `CudaRdcUtils` and use `cuda_rdc_add_library`? We *don't* want downstream code to add celeritas-specific aliases.

sethrj:
@pcanal for backward compatibility I've completed https://github.com/celeritas-project/celeritas/pull/1104/commits/ab1834e60bc74daac8911057e71856b316e70bdb ... hopefully it works!

cmake/CeleritasLibrary.cmake
pcanal: Yes, that would be better.  In this case we don't need to distribute `CeleritasUtils.mk` but need to update the examples.

sethrj:
@pcanal looks like there's one final issue; not sure why, but the wrapper functions aren't propagating "target includes" for the installed target:
```
/__w/celeritas/celeritas/example/minimal/minimal.cc:9:10: fatal error: celeritas/Constants.hh: No such file or directory
    9 | #include <celeritas/Constants.hh>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
```

pcanal:
I am looking into it.

cmake/CeleritasUtils.cmake
```diff
@@ -61,6 +61,21 @@ CMake configuration utility functions for Celeritas.
   the ``Celeritas::`` aliases, and is generated into the ``lib/`` build
   directory.
 
+.. command:: celeritas_install
+
+  Install library that correctly deal with CUDA relocatable device code.
+
+.. command:: celeritas_set_properties
```
sethrj:
```suggestion
.. command:: celeritas_set_target_properties
```

sethrj:
This is where the investment we've spent into CI really pays off... we've had to make like 10 changes to get build, install, test, downstream build working!

sethrj (approved):
Finally it all works!

PR #1105, opened 2024-02-13
Fix CTest names

sethrj:
In #1099 I accidentally find-replaced away the leading `celeritas/`, `corecel/`, and `orange/`. This restores them.

### DIFF
--- a/test/celeritas/CMakeLists.txt
+++ b/test/celeritas/CMakeLists.txt
@@ -115,7 +115,7 @@ celeritas_add_test(Units.test.cc)
 
 #-----------------------------------------------------------------------------#
 # EM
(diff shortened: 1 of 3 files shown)


amandalund (approved):
Thanks @sethrj!

sethrj:
Maybe it wasn't an accident... originally, the test for `test/A/B/C.test.cc` was built at `test/A_B_C`  and had the CTest name `test/A/B/C`. After #1099 the name was built at `test/A/B_C` but was incorrectly named `test/B/C`. After this PR, it's now correctly named `test/A/B/C` but now is built at `test/A/A_B_C` 🤪

PR #1106, opened 2024-02-13
Rename ORANGE construction files

sethrj:
This rearranges the ORANGE construction directory, moving the components used for "shape construction" (e.g. the CSG tree) into `orangeinp/`, and moving the `OrangeInput` class to the main directory. The CSG renaming solves all the weird namespace issues of `using namespace orange::csg`.


app/orange-update.cc
```diff
@@ -21,7 +21,7 @@
 #    include <fstream>
 #    include <nlohmann/json.hpp>
 
-#    include "orange/construct/OrangeInputIO.json.hh"
+#    include "orange/OrangeInputIO.json.hh"
```
elliottbiondo, line 25: This .cc file doesn't have any info about what version of ORANGE geometries are being updated. Though unrelated to this MR, consider adding a short description.

src/orange/detail/BIHBuilder.hh
```diff
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file orange/detail/BIHBuilder.hh
+//! \todo move to orangeinp/BihBuilder
```
elliottbiondo: Remove

src/orange/detail/BIHPartitioner.hh
```diff
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
 //! \file orange/detail/BIHPartitioner.hh
+//! \todo Move to orangeinp/detail/BihPartitioner
```
elliottbiondo: Remove

elliottbiondo (approved):

app/orange-update.cc
sethrj: Yeah, this just reads in a JSON file (which will translate any "old" parameter names, etc.) and writes it back out again. I'll add a brief in this file, but the app *is* actually documented in `doc/main/usage.rst`.

PR #1108, opened 2024-02-14
Add Cerenkov pre-generator

stognini:
This PR adds a capability needed for #886 :
- Add a new Cerenkov pre-generator class that uses the collected data step data to sample the number of photons per step and returns a populated `OpticalDistributionData`, which is an input of the Cerenkov generator class.

The `Cerenkov.test` has been refactored to use the pre-generator in all sampling tests.


src/celeritas/optical/OpticalStepCollector.hh
```diff
@@ -0,0 +1,103 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Callback class derived from `StepInterface` that stores
+ * \c OpticalStepCollectorData to be used by optical physics pre-generator
+ * classes.
+ *
+ */
+class OpticalStepCollector final : public StepInterface
```
sethrj: Didn't we decide to postpone the implementation of this class (since we might end up with something that's like the step collector but different)? This class isn't used or tested anywhere in the PR...

test/celeritas/optical/OpticalTestBase.hh
```diff
@@ -0,0 +1,65 @@
...
+#pragma once
+
+#include <memory>
+
+#include "celeritas/Quantities.hh"
+#include "celeritas/Types.hh"
+#include "celeritas/phys/ParticleData.hh"
+
+#include "DiagnosticRngEngine.hh"
+#include "celeritas_test.hh"
```
sethrj: I think you'll have to include `Test.hh` instead of this file, since `celeritas_test.hh` defines `main()`.

test/celeritas/optical/OpticalTestBase.hh
```diff
@@ -0,0 +1,65 @@
...
+    OpticalTestBase();
+    ~OpticalTestBase();
+    //!@}
+
+    //! Initialize particle state data with given energy
+    ParticleTrackView
+    make_particle_track_view(units::MevEnergy energy, PDGNumber pdg);
+
+    //! Get particle params data
+    std::shared_ptr<ParticleParams> particle_params()
```
sethrj: Return by const reference, and make this function const.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -307,12 +316,7 @@ TEST_F(CerenkovTest, TEST_IF_CELERITAS_DOUBLE(generator))
     };
 
     size_type num_samples = 64;
-
-    OpticalDistributionData dist;
-    dist.time = 0;
-    dist.charge = charge;
-    dist.material = material;
-    dist.points[StepPoint::pre].pos = {0, 0, 0};
+    OpticalStepCollectorData step_collector;
```
sethrj: I'd rename this `step` because it's the collected data over a step, it doesn't do any collecting. Same with above.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -378,6 +392,53 @@ TEST_F(CerenkovTest, TEST_IF_CELERITAS_DOUBLE(generator))
         EXPECT_SOFT_EQ(401.5, avg_engine_samples);
     }
 }
+
+//---------------------------------------------------------------------------//
+
+TEST_F(CerenkovTest, TEST_IF_CELERITAS_DOUBLE(pre_generator))
```
sethrj: Since the generator test above relies on the pregenerator working correctly, I'd move this test above the `generator` test.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -378,6 +392,53 @@ TEST_F(CerenkovTest, TEST_IF_CELERITAS_DOUBLE(generator))
...
+    for (auto p : range(StepPoint::size_))
+    {
+        EXPECT_SOFT_EQ(step_collector.points[p].speed.value(),
+                       result.points[p].speed.value());
+        for (auto i : range(3))
+        {
+            EXPECT_SOFT_EQ(step_collector.points[p].pos[i],
+                           result.points[p].pos[i]);
+        }
+    }
```
sethrj: Put this in a loop and generate a couple of samples (push back `num_photons` onto a vector and use `EXPECT_VEC_EQ`).

src/celeritas/optical/OpticalStepCollector.hh
stognini: Sorry, maybe I misunderstood the conclusion of our chat. We discussed that the existing `StepInterface` is not ideal for the optical physics, but we wouldn't make a new one as of now so I kept this in to have a starting point. I can remove it from this PR if you prefer.

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,131 @@
...
+ * where \f$ \ell_\text{step} \f$ is the step length.
+ */
+template<class Generator>
+CELER_FUNCTION OpticalDistributionData
+CerenkovPreGenerator::operator()(Generator& rng)
+{
+    auto const& pre = step_collector_data_.points[StepPoint::pre];
+    auto const& post = step_collector_data_.points[StepPoint::post];
+    units::LightSpeed beta(real_type{0.5}
+                           * (pre.speed.value() + post.speed.value()));
```
amandalund: This could be moved to the constructor since it doesn't depend on the rng.

src/celeritas/optical/OpticalStepCollector.hh
sethrj: I could've misunderstood or forgotten myself :)

Since it's not necessary to set up the step data in the tests, and it's not tested or used anywhere, let's just remove it for now. We can `git revert` the removal in a follow-on PR where we actually test the integration into the tracking loop.

stognini: Great. Sounds good to me.

src/celeritas/optical/OpticalDistributionData.hh
```diff
@@ -45,5 +48,22 @@ struct OpticalDistributionData
...
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Data used by \c OpticalStepCollector class implementation.
+ */
+struct OpticalStepCollectorData
+{
+    real_type time{};  //!< Pre-step time
+    real_type step_length{};  //!< Step length
```
amandalund: I think <s>time and </s> `step_length` could be retrieved directly from `SimTrackView` in the `PreGenerator`: that way we don't need to store this extra state data.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -378,6 +392,53 @@ TEST_F(CerenkovTest, TEST_IF_CELERITAS_DOUBLE(generator))
...
+    step_collector.points[StepPoint::pre].pos = {0, 0, 0};
+    step_collector.points[StepPoint::pre].speed
+        = units::LightSpeed{0.86286196322132458};
+    step_collector.points[StepPoint::post].pos
+        = {step_collector.step_length, 0, 0};
+    step_collector.points[StepPoint::post].speed
+        = units::LightSpeed{0.63431981443206786};
+    CELER_ASSERT(step_collector);
+
+    // 500 keV e-
```
amandalund: You could also test a case where the incident particle speed is below the threshold for Cerenkov production.

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,131 @@
...
+
+    // Sample number of photons from a Poisson distribution
+    real_type sampled_num_photons = PoissonDistribution<real_type>(
+        dndx_calculator_(beta) * step_collector_data_.step_length)(rng);
```
amandalund: Since dN/dx could be zero, I think we'll need to calculate it before sampling from the Poisson distribution and return early in that case.

amandalund (commented):
Thanks @stognini!

src/celeritas/optical/OpticalDistributionData.hh
stognini: You got me thinking. The `OpticalStepCollectorData` only existed because we'd need a `StepInterface` callback class. I guess if we have `SimTrackView`, we'd also have `GeoTrackView` to provide the position, since we'll probably have access to `CoreState` data. Should I make this independent from a callback class and assume we have global core params/state directly?

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,139 @@
...
+    CELER_EXPECT(step_collector_data_);
+
+    auto const& pre = step_collector_data_.points[StepPoint::pre];
+    auto const& post = step_collector_data_.points[StepPoint::post];
+    units::LightSpeed beta(real_type{0.5}
+                           * (pre.speed.value() + post.speed.value()));
+
+    CerenkovDndxCalculator dndx_calculator(
+        properties, shared, mat_id_, charge_);
+    num_photons_per_len_ = dndx_calculator(beta);
```
sethrj: Try to phrase functor instances as verbs, just like you would name a function: an instance of `FooCalculator` should be `calculate_foo`.

src/celeritas/optical/OpticalDistributionData.hh
amandalund: Well we'll still need the pre-step data which we'll have to cache somewhere, so I think having this for the time/step points makes sense.

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,139 @@
...
+ */
+template<class Generator>
+CELER_FUNCTION OpticalDistributionData
+CerenkovPreGenerator::operator()(Generator& rng)
+{
+    if (num_photons_per_len_ == 0)
+    {
+        return {};
+    }
+
```
sethrj, line 137: Do we have an idea of how often the number of sampled photons is zero? I wonder if we should throw out this edge case by changing the `PoissonDistribution` requirement to `lambda_ >= 0` from `lambda_ > 0`. (Then this would always result in a zero sample, so we'd throw away a random number but reduce complexity here.)

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,139 @@
...
+    // Sample number of photons from a Poisson distribution
+    auto sampled_num_photons = PoissonDistribution<real_type>(
+        num_photons_per_len_ * step_collector_data_.step_length)(rng);
+
+    if (sampled_num_photons == 0)
+    {
+        // Not an optical material or this step is below production threshold
+        return {};
+    }
```
sethrj: Is this special case necessary? I'd just put the `data.{charge ... material}` inside an `if (data.num_photons > 0)` block.

src/celeritas/optical/OpticalDistributionData.hh
```diff
@@ -23,6 +23,9 @@ struct OpticalStepData
 {
     units::LightSpeed speed;
     Real3 pos{};
+
+    //! Check whether the data are assigned
+    explicit CELER_FUNCTION operator bool() const { return speed.value() > 0; }
```
sethrj:
```suggestion
    explicit CELER_FUNCTION operator bool() const { return speed > zero_quantity(); }
```

src/celeritas/optical/OpticalDistributionData.hh
stognini: Correct. But since we are now down to just the points, I could just label it as a `Result = EnumArray<StepPoint, OpticalStepData>` of the class, which removes an indirection and uses an existing data struct created for the `OpticalDistributionData`.

sethrj: Yes, good idea.

amandalund: We'll need the pre-step time too, so we'll need somewhere to store that (sorry, my original comment was edited).

src/celeritas/optical/CerenkovPreGenerator.hh
amandalund: The Poisson distribution uses a normal distribution, so we'd have to do the same there... not sure if that makes sense or not. Right now the calculated dN/dx can be zero if:
- No angle integrals (refractive index data) was provided for this optical material
- The incident particle speed is below the threshold for Cerenkov emission
- More rarely, the incident particle energy is above but very close to the threshold

stognini:
Okay, assuming no callback class, assuming core state/params data is available whenever the pre-generator is called, and to improve organization (`OpticalDistributionData.hh` is only used as the base data for the generators), I've simplified the input of the PreGenerator with its own input struct.

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,139 @@
...
+class CerenkovPreGenerator
+{
+  public:
+    struct InputStepStateData
+    {
+        real_type time{};  //!< Pre-step time
+        real_type step_length{};  //!< Step length
+        units::ElementaryCharge charge;
+        OpticalMaterialId material;
+        EnumArray<StepPoint, OpticalStepData> points;  //!< Pre- and post-steps
```
amandalund: This may be a reasonable placeholder; however, I think what you had originally is closer to what this will ultimately look like. Since this input is essentially the same as `OpticalDistributionData`, now we would be creating the distribution outside the pre-generator, passing it in, copying it into a new struct and getting it back (plus the sampled number of photons). If we were to do things this way it would make more sense to have this class just sample and return the number of photons.

stognini: Fair. If we go for constructing it directly with views (so `ParticleTrack`, `SimTrack`, `GeoTrack` views), we'd still be missing pre-step time, right?

amandalund: Right, and pre-step position and speed. I think your extra struct with the data we can't get directly from the track views was a good solution for now (maybe `OpticalPreStepData` instead of `OpticalStepCollectorData`?). We can assume we'll have that data and all the track views when we construct the pre-generator.

stognini: Okay. Refactored so all the data that can come from views is pulled directly from them. Unavailable data is cached in a temporary placeholder, as you suggested.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -203,11 +211,96 @@ TEST_F(CerenkovTest, dndx)
...
+        EXPECT_SOFT_EQ(sim_view.step_length(), result.step_length);
+
+        for (auto p : range(StepPoint::size_))
+        {
+            EXPECT_SOFT_EQ(step.points[p].speed.value(),
+                           result.points[p].speed.value());
+            for (auto i : range(3))
+            {
+                EXPECT_SOFT_EQ(step.points[p].pos[i], result.points[p].pos[i]);
+            }
```
amandalund:
```suggestion
            EXPECT_VEC_EQ(step.points[p].pos, result.points[p].pos);
```

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -203,11 +211,96 @@ TEST_F(CerenkovTest, dndx)
...
+
+        // Remaining values are assigned to result from input data
+        EXPECT_SOFT_EQ(step.time, result.time);
+        EXPECT_EQ(particle_view.charge().value(), result.charge.value());
+        EXPECT_EQ(material, result.material);
+        EXPECT_SOFT_EQ(sim_view.step_length(), result.step_length);
+
+        for (auto p : range(StepPoint::size_))
+        {
+            EXPECT_SOFT_EQ(step.points[p].speed.value(),
```
amandalund: These could go inside the sampling loop.

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -203,11 +211,96 @@ TEST_F(CerenkovTest, dndx)
...
+            {
+                EXPECT_SOFT_EQ(step.points[p].pos[i], result.points[p].pos[i]);
+            }
+        }
+    }
+
+    // Below Cerenkov threshold
+    {
+        auto particle_view = this->make_particle_track_view(
+            units::MevEnergy{0.3}, pdg::electron());
```
amandalund:
```suggestion
            units::MevEnergy{0.1}, pdg::electron());
```
To match the speed.

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,154 @@
...
+        return {};
+    }
+
+    // Sample number of photons from a Poisson distribution
+    auto sampled_num_photons = PoissonDistribution<real_type>(
+        num_photons_per_len_ * step_len_)(rng);
+
+    OpticalDistributionData data;
+    data.num_photons = sampled_num_photons;
+    data.time = step_data_.time;
```
amandalund: I think @sethrj suggested something like:
```
OpticalDistributionData data;
data.num_photons = PoissonDistribution<real_type>(...)
if (data.num_photons > 0)
{
    data.time = step_data_.time;
    ...
}
```

src/celeritas/optical/CerenkovPreGenerator.hh
whokion: The first case is a bit odd to me as the refractive index should be a minimal requirement if the material has a optical property (necessary condition of the optical transport/physics) and therefore the angle integrals should be built if Chrenkov is one of optical processes. Okay, the extra check does not harm though.

src/celeritas/optical/CerenkovPreGenerator.hh
whokion: Similar logical comment for the comment line - Cerenkov pre-generator should be only called if the material where the step is, is an optical material.

stognini: Fair assumption. Guess I could update the comment to state it will return a false `OpticalDistributionData` object if step is below threshold, as we may not check the threshold before calling the pre-generator (just let it tell us that the number of sampled photons on that step is zero).

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,154 @@
...
+ * \endcode
+ */
+class CerenkovPreGenerator
+{
+  public:
+    // Placeholder for data that is not available through Views
+    struct OpticalPreGenStepData
+    {
+        real_type time{};  //!< Pre-step time
+        EnumArray<StepPoint, OpticalStepData> points;  //!< Pre- and post-steps
```
amandalund, line 54: We may end up moving this outside this class since the scintillation pre-generator will use it too. We also might be able to get away with just storing the pre-step speed and position and getting the post-step values from the particle/geo track views (don't have to do that now, just something to consider in the future).

amandalund (commented):
Looks good @stognini! Just a few more minor comments.

test/celeritas/optical/Cerenkov.test.cc
sethrj: Yes, since we're not doing any operations on the position (just a straight copy!) they should be exactly equal.

src/celeritas/optical/CerenkovPreGenerator.hh
amandalund: Maybe a naive question, but what can we assume about a material with optical properties? Should an optical material have _all_ optical properties assigned? Should it at least have refractive index (or some other minimal subset of data)? Could it just have an arbitrary subset of properties?

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,154 @@
...
+    {
+        CerenkovGenerator cerenkov_generate(... , optical_dist_data, ...);
+        cerenkov_generate(rng);
+    }
+ * \endcode
+ */
+class CerenkovPreGenerator
+{
+  public:
+    // Placeholder for data that is not available through Views
```
whokion, line 50: What is this extra placeholder necessary?  Can you just directly fill OpticalDistributionData from step input (whatever views) information?

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -0,0 +1,154 @@
...
+    auto sampled_num_photons = PoissonDistribution<real_type>(
+        num_photons_per_len_ * step_len_)(rng);
+
+    OpticalDistributionData data;
+    data.num_photons = sampled_num_photons;
+    data.time = step_data_.time;
+    data.step_length = step_len_;
+    data.charge = charge_;
+    data.material = mat_id_;
+    data.points = step_data_.points;
```
whokion: Can all these filled in the constructor if data is a private data?

whokion (commented):
Where is a buffer/collection to store OpticalDistributionData and how do trigger CerenkovGenerator with the collection when a number of OpticalDistributionData in the collection (or the total accumulated number of photons to generate) is above a certain threshold?  Shouldn't this high level interface support such a workflow or is it the responsibility of even a higher API?

src/celeritas/optical/CerenkovPreGenerator.hh
```diff
@@ -135,18 +135,17 @@ CerenkovPreGenerator::operator()(Generator& rng)
...
-    data.time = step_data_.time;
-    data.step_length = step_len_;
-    data.charge = charge_;
-    data.material = mat_id_;
-    data.points = step_data_.points;
-
+    if (size_type sampled_num_photons = PoissonDistribution<real_type>(
+            num_photons_per_len_ * step_len_)(rng))
+    {
+        data.num_photons = sampled_num_photons;
```
sethrj: What I actually wanted to suggest was:
```suggestion
    data.num_photons = PoissonDistribution<real_type>(
            num_photons_per_len_ * step_len_)(rng);
    if (data.num_photons > 0)
    {
```
This way you're not relying on the default initializer for `num_photons`, and I think the `if` statement is more readable.

stognini:
> Where is a buffer/collection to store OpticalDistributionData

This will be part of the integration with the main loop. For the pre-generators we were just assuming that we have access to core params/data. This includes why we removed the `OpticalStepCollector` class, which was a `StepInterface` derived class that collected time and pre/post steps to be passed to the pre-generator.

src/celeritas/optical/CerenkovPreGenerator.hh
stognini: I guess we could. But we'd need an `OpticalDistributionData result_;` private member. I don't think it makes much difference, since we always construct and operate once. Maybe it's bit better as is, since we only spend time copying data if `operator()` is successful.

src/celeritas/optical/CerenkovPreGenerator.hh
whokion: I think that supporting an arbitrary subset of properties is a good strategy/design choice as a user application can be an arbitrary combination of optical processes (even with only one specific process depending on their simulation need). Concerning to the refractive index, it may be a minimal requirement as I hardly imagine that there would be a meaning optical simulation without an optical media (i.e, which by definition should have the refractive index) even though one can do whatever they want - on the other hand, as a designer/developer, you may impose your own design style and a restriction of colors.

src/celeritas/optical/CerenkovPreGenerator.hh
stognini: We don't have time, nor pre-step information from existing views. So this has to be collected somewhere during the main loop. See my comment bellow about the `OpticalStepCollector` class, that was removed.

src/celeritas/optical/CerenkovPreGenerator.hh
sethrj: No we *definitely* want to return a sample result...

stognini: I mean, we still return a sample, and it is still false if num_photons is zero. So copying at construction and `return result_` in `operator()` seems like it would behave identically. We just moved the copy assignment to the construction (which when num_photons is zero, becomes unnecessary).

whokion:
> > Where is a buffer/collection to store OpticalDistributionData
> 
> This will be part of the integration with the main loop. For the pre-generators we were just assuming that we have access to core params/data. This includes why we removed the `OpticalStepCollector` class, which was a `StepInterface` derived class that collected time and pre/post steps to be passed to the pre-generator.

Then, we probably need to add another process-like-Id into OpticalDistributionData (photons to be generated are tagged or categorized by a specific process (i.e., distinguish them by Cerenkov photons, Scintillation photons and so on for propagation, at least for supporting MC-truth like information) -  we could trace them in a different way, but just for a quick easy solution for now.  Anyway, I do not see much use of this pre-generator class (just doing a Poisson sampling and preparing OpticalDistributionData which can be done anywhere else) even though it still may be a necessary layer or should be the layer.

sethrj (approved):
I think this is good to go. @whokion @amandalund if there are concerns about the scope/utility of this PR, maybe we discuss on zoom?

amandalund (approved):
Good from my end!

whokion:
Good from my side too.  If there is any issue, we may continue to discuss relevant items when we add an interface for Scintillation (which I assume the next).  Thanks @stognini for a good starting work!

PR #1109, opened 2024-02-14
Reduce number of Geant4 libraries linked in G4VG and elsewhere

sethrj:
This more carefully targets the libraries required in `geocel` (and therefore downstream) and `celeritas`.


drbenmorgan (approved):
Looks good @sethrj, LGTM!

PR #1112, opened 2024-02-16
Add deduplicating transform inserter

sethrj:
This is an `orangeinp` helper class for adding `VariantTransform` instances to a vector with deduplication. Same transforms will be combined to the same object. This differs from the `TransformRecordInserter` in that it actually deduplicates the object, not just the `real_type` data underneath, and it's going to be used locally to a single universe. Unlike the surface deduplication, this does *not* currently do any soft equivalence; it requires byte-for-byte identical results, which I think will be good enough for immediate use cases.

### DIFF
--- a/src/orange/orangeinp/detail/TransformInserter.hh
+++ b/src/orange/orangeinp/detail/TransformInserter.hh
@@ -43,6 +43,7 @@ class TransformInserter
 
   private:
     //// TYPES ////
+
     struct HashTransform
     {
         VecTransform* storage{nullptr};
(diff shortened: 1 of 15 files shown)


src/orange/orangeinp/detail/TransformInserter.hh
```diff
@@ -0,0 +1,69 @@
...
+        VecTransform* storage{nullptr};
+        bool operator()(TransformId, TransformId) const;
+    };
+
+    //// DATA ////
+
+    VecTransform* transform_;
+    std::unordered_set<TransformId, HashTransform, EqualTransform> cache_;
+
+    //! Get the ID of the next transform to be inserted
```
elliottbiondo, line 65: Should be in a `/// HELPER FUNCTIONS ///` section or at least not under DATA

elliottbiondo (approved):

PR #1113, opened 2024-02-16
Add convex surface builder

sethrj:
The "convex surface builder" is going to be the next building block above surfaces when constructing shapes, solids, etc. It's a single `intersection` operation on a list of convex surfaces, so it will be easy to reason about when formulating safety calculations, constructing building boxes, etc.

The expected use case will be to have a `ConvexRegion` class that can be reusable (for example doing rings and other partially negated regions). A `Shape` class, which "is-a" `Object`, i.e. something that can be turned into a Volume, will have a single convex region; a solid may have an inside and outside; etc.

```c++
// The building block of a shape or solid
class ConvexRegion
{
    virtual void build(ConvexSurfaceBuilder& add_surface) const = 0;
};

// A hemisphere as part of a shape or a solid
class HemisphereRegion : public ConvexRegion
{
    void build(ConvexSurfaceBuilder& add_surface) const final
    {
        add_surface(PlaneZ{0.0});
        add_surface(SphereCentered{radius_});
    }
};

// Primitive shape consisting of a single solid
class Shape : public Object
{
  public:
    virtual ConvexRegion const& interior() const;

    NodeId build(VolumeBuilder& vb) const final;
};

NodeId Shape::build(VolumeBuilder& vb) const final
{
    detail::ConvexSurfaceState css;
    css.transform = &sb.current_transform();
    css.object_name = this->label();
    ConvexSurfaceBuilder sb{vb.unit_builder(), &css};
    this->interior().build(sb);
    auto result = vb.add_csg_node(Join, std::move(css.nodes));
    // Save bounding zone and metadata for the shape
    return result;
}
```


src/orange/orangeinp/ConvexSurfaceBuilder.hh
```diff
@@ -0,0 +1,90 @@
...
+    void operator()(Sense sense, S const& surf);
+
+  public:
+    // "Private", to be used by testing and detail
+    using State = detail::ConvexSurfaceState;
+    using UnitBuilder = detail::CsgUnitBuilder;
+    using VecNode = std::vector<NodeId>;
+
+    // Construct with persistent unit builder and less persistent state
+    ConvexSurfaceBuilder(UnitBuilder* ub, State* state);
```
elliottbiondo, line 65: Should this be explicit? As I constructor it should be at the first function in the file.

src/orange/orangeinp/ConvexSurfaceBuilder.cc
```diff
@@ -0,0 +1,131 @@
...
+ConvexSurfaceBuilder::ConvexSurfaceBuilder(UnitBuilder* ub, State* state)
+    : ub_{ub}, state_{state}
+{
+    CELER_EXPECT(ub_ && state_);
+    CELER_EXPECT(*state_);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add a plane with a possibly *outside* sense.
```
elliottbiondo: Add a surface

src/orange/orangeinp/ConvexSurfaceBuilder.cc
```diff
@@ -0,0 +1,131 @@
...
+    RecursiveSimplifier clip_simplified_local(ClipImpl{&state_->local_bzone},
+                                              ub_->tol());
+    clip_simplified_local(sense, surf);
+
+    // Next, apply the transform and insert
+    return this->insert_transformed(state_->make_face_name(sense, surf),
+                                    sense,
+                                    apply_transform(*state_->transform, surf));
+}
+
```
elliottbiondo, line 81: Add `// HELPER FUNCTION` comment

src/orange/orangeinp/ConvexSurfaceBuilder.cc
```diff
@@ -0,0 +1,131 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Add a surface with an "inside" sense (true for negative quadric value).
+ */
+void ConvexSurfaceBuilder::insert_transformed(std::string&& ext,
```
elliottbiondo: It's not immediately clear why this variable is named `ext`

src/orange/orangeinp/ConvexSurfaceBuilder.cc
```diff
@@ -0,0 +1,131 @@
...
+        // "Inside" the surface (negative quadric evaluation) means we have to
+        // negate the CSG result
+        static_assert(Sense::inside == to_sense(false));
+        node_id = ub_->insert_csg(Negated{node_id});
+    }
+
+    // Add sense to "joined" region
+    state_->nodes.push_back(node_id);
+}
+
```
elliottbiondo, line 131: Add doc string

elliottbiondo: Differentiate between free functions and helper functions within this file as you do in the .cc

src/orange/orangeinp/detail/BoundingZone.hh
```diff
@@ -102,8 +102,14 @@ struct BoundingZone
 
     // Flip inside and outside
     inline void negate();
+
+    // Create an "everything is known inside" zone for intersecting
+    static BoundingZone from_infinite();
 };
 
+//---------------------------------------------------------------------------//
+// FREE FUNCTIONS
```
elliottbiondo, line 111: I suggest adding this to the .cc as well

src/orange/orangeinp/detail/NegatedSurfaceClipper.hh
```diff
@@ -0,0 +1,94 @@
...
+    explicit inline NegatedSurfaceClipper(BoundingZone* bz);
+
+    //! Clip axis-aligned planes.
+    template<Axis T>
+    CELER_FORCEINLINE void operator()(PlaneAligned<T> const& s)
+    {
+        return this->clip_impl(T, s.position());
+    }
+
+    // All other operations invalidate the "interior" box
```
elliottbiondo: //!

src/orange/surf/SurfaceClipper.cc
```diff
@@ -45,8 +45,8 @@ SurfaceClipper::SurfaceClipper(BBox* interior, BBox* exterior)
 template<Axis T>
 void SurfaceClipper::operator()(PlaneAligned<T> const& s) const
 {
+    int_->shrink(Bound::hi, T, s.position());
```
elliottbiondo, line 48: Do you intent to change this from Bound::lo to Bound::hi?

test/orange/orangeinp/ConvexSurfaceBuilder.test.cc
```diff
@@ -0,0 +1,256 @@
...
+        static int const expected_nodes[] = {3, 5};
+        EXPECT_VEC_EQ(expected_nodes, to_vec_int(css.nodes));
+    }
+    {
+        SCOPED_TRACE("slab");
+        auto css = this->make_state();
+        css.object_name = "sl";
+        ConvexSurfaceBuilder build{&unit_builder_, &css};
+        x_slab(build);
+        static real_type const expected_local_bz[] = {
```
elliottbiondo, line 121: Not super important, but being a bit less sparing with spacing would be helpful in this file for readability. For example, having chunks of code separated by newlines where each chunk initializes an expected vector then called the EXPECT().

elliottbiondo (approved):

src/orange/orangeinp/ConvexSurfaceBuilder.hh
sethrj: `explicit` is usually only needed for constructors that have a single argument, so if you have
```c++
struct Foo {
Foo(int a);
}
```
`explicit` will prevent a function `do_something(Foo const& f)` from matching a call to `do_something(123)`. The keyword also prevents brace initialization without the explicit use of the class name, so for example in a return-by-value:
```c++
ConvexSurfaceBuilder make_a_csb(...)
{
    return {ub, state}; // would be prohibited if the constructor were `explicit`
}
```

sethrj: I put the constructor below the "public" methods in a commented section because user classes are intended to *only* use an already-constructed CSB, and we put `private` sections below `public` since they're not part of the interface.

src/orange/surf/SurfaceClipper.cc
sethrj: Yeah, the previous value was a mistake. I updated the tests to match.

test/orange/orangeinp/ConvexSurfaceBuilder.test.cc
sethrj: Point taken... I should update the "expected output" code. Historically this was because putting the array assignments next to each other would destroy clang-formatting assignment, but now we don't align on `=` so I can do that. I think this is an improvement in readability:
```c++
    {
        SCOPED_TRACE("slab");
        auto css = this->make_state();
        css.object_name = "sl";
        ConvexSurfaceBuilder build{&unit_builder_, &css};
        x_slab(build);

        static real_type const expected_local_bz[] = {
            -0.5, -inf, -inf, 0.5, inf, inf, -0.5, -inf, -inf, 0.5, inf, inf, 1};
        static real_type const expected_global_bz[] = {
            -0.5, -inf, -inf, 0.5, inf, inf, -0.5, -inf, -inf, 0.5, inf, inf, 1};
        static int const expected_nodes[] = {6, 8};
        EXPECT_VEC_SOFT_EQ(expected_local_bz, flattened(css.local_bzone));
        EXPECT_VEC_SOFT_EQ(expected_global_bz, flattened(css.global_bzone));
        EXPECT_VEC_EQ(expected_nodes, to_vec_int(css.nodes));
    }
```

PR #1114, opened 2024-02-17
Detect and use all CPUs in GitHub CI

sethrj:
Recently the [default Ubuntu Github Runners now have 4 cores](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) instead of the previous 2. Add some code to detect and use the right number of cores during testing.


sethrj:
It doesn't improve the wall time at all (and seems to increase the total CPU time!) to do this, plus the node.js script doesn't work with the docker image.

PR #1115, opened 2024-02-18
Add JSON test comparator to improve testing reliability

sethrj:
A couple of tests (notably the Material test) seem to fail on some platforms due to differences in JSON output of floating point numbers. This implements a recursive comparator for JSON strings that includes soft equivalence.


amandalund (approved):
Nice, thanks @sethrj!

PR #1116, opened 2024-02-19
Add Wentzel OK&VI transport cross section calculator

amandalund:
This adds a Wentzel OK&VI transport cross section calculator (which will be needed for the Wentzel VI MSC model). It also renames the `WentzelRatioCalculator` so it's now a helper class.

### DIFF
--- a/src/celeritas/em/xs/WentzelHelper.hh
+++ b/src/celeritas/em/xs/WentzelHelper.hh
@@ -3,7 +3,7 @@
 // See the top-level COPYRIGHT file for details.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file celeritas/em/xs/WentzelRatioCalculator.hh
+//! \file celeritas/em/xs/WentzelHelper.hh
 //---------------------------------------------------------------------------//
 #pragma once
 
@@ -17,10 +17,13 @@ namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Calculate the ratio of the electron to total Wentzel cross sections for
- * elastic Coulomb scattering.
+ * Helper class for the Wentzel OK and VI Coulomb scattering model.
+ *
+ * This calculates the Moliere screening coefficient, the maximum scattering
+ * angle off of electrons, and the ratio of the electron to total Wentzel cross
+ * sections.
  */
-class WentzelRatioCalculator
+class WentzelHelper
 {
   public:
     //!@{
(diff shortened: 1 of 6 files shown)


src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -144,16 +144,17 @@ CELER_FUNCTION real_type WentzelRatioCalculator::calc_screening_coefficient(
...
-    return correction * screen_r_sq_elec() * sq_cbrt_z
+    return correction * this->screen_r_sq_elec() * sq_cbrt_z
            / value_as<MomentumSq>(particle.momentum_sq());
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Calculate the screening R^2 coefficient for incident electrons. This is
- * the constant prefactor of [PRM] eqn 8.51.
+ * Calculate the screening R^2 coefficient for incident electrons.
```
whokion: It is unclear what `R^2` is referring to here.  Also better to put the `version` of RPM always as the eqn number may be changed the later if RPM is updated.

src/celeritas/em/xs/WentzelTransportXsCalculator.hh
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * \note This performs the same calculation as Geant4's
+ * G4WentzelOKandVIxSection::ComputeTransportCrossSectionPerAtom.
+ */
+class WentzelTransportXsCalculator
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using XsUnits = units::Native;  // [len^2]
```
whokion, line 32: Will this be used elsewhere?

src/celeritas/em/xs/WentzelTransportXsCalculator.hh
```diff
@@ -0,0 +1,163 @@
...
+ * OK and VI model in \c WentzelHelper and depend on properties of the incident
+ * particle, the energy cutoff in the current material, and the target element.
+ */
+CELER_FUNCTION
+WentzelTransportXsCalculator::WentzelTransportXsCalculator(
+    ParticleTrackView const& particle, WentzelHelper const& helper)
+    : z_(helper.atomic_number())
+    , screening_coeff_(helper.screening_coefficient())
+    , costheta_max_elec_(helper.costheta_max_electron())
+    , spin_betasq_(real_type(0.5) * particle.beta_sq())
```
whokion: Is this related to the electron `spin` (0.5)? Okay, it seems that this is (`spin x beta^2`) even though the variable name is a somewhat artificial choice.

test/celeritas/em/Wentzel.test.cc
```diff
@@ -212,6 +212,71 @@ TEST_F(CoulombScatteringTest, mott_xs)
...
+    ElementId elm_id{0};
+
+    AtomicNumber const z = this->material_params()->get(elm_id).atomic_number();
+
+    // Incident particle energy cutoff
+    MevEnergy const cutoff = this->cutoff_params()->get(mat_id).energy(
+        this->particle_track().particle_id());
+
+    std::vector<real_type> xs;
+    for (real_type energy : {50, 100, 200, 1000, 13000})
```
whokion: Is the test below 100MeV (Wenzel low limit) meaningful? Okay for testing as it should be a soft-limit. Is 13GeV special energy point? May add at least 100GeV and 1TeV?

whokion (approved):
This MR looks good to me! Just for my curiosity, why does xs (probably macroscopic xs based on this) need to be calculated on the fly for the Wenzel msc? Is this the special case when the Wenzel msc is replaced by the Coulomb single scattering within the model?

src/celeritas/em/xs/WentzelHelper.hh
amandalund: I tried to clarify the comment and added a PRM reference in the class description.

src/celeritas/em/xs/WentzelTransportXsCalculator.hh
amandalund: This is just to annotate that the calculated cross section is in native Celeritas units.

src/celeritas/em/xs/WentzelTransportXsCalculator.hh
amandalund: Right, it is spin x beta^2. Updated this to try to make it clearer.

test/celeritas/em/Wentzel.test.cc
amandalund: I just used the same energies as the test for the Wentzel helper, so I'm not actually sure if there is a special meaning. I'll add some higher energy points.

amandalund:
@whokion right, it's used to recalculate the macro xs in the calculation of the true step length to account for the single scattering.

whokion:
@amandalund Thanks for all updates! @sethrj I think that this MR is good to be merged (unless @hhollenb has extra comments).

sethrj (approved):
Nice! Thanks.

PR #1118, opened 2024-02-20
Disable ROCm builds because of constant failures

sethrj:
The Jenkins CI has been failing intermittently, probably about half the time per ROCM GPU job. I've confirmed this is a problem with the CI hardware itself (Kokkos is seeing it too) and not something we've done.

This removes those jobs so that a red ❌  is actually meaningful. Even though we won't *run* ROCm code, our github workflow will still build it, so we won't accidentally introduce any non-HIP issues.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -10,40 +10,6 @@ pipeline {
   stages {
     stage('Build') {
       parallel {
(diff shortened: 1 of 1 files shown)


amandalund (approved):
Thanks @sethrj!

PR #1120, opened 2024-02-22
Add neutron elastic scattering physics 

whokion:
# Add neutron elastic scattering physics
 - Implement neutron-nucleus elastic scattering process, model, interactor and associated classes *[enhancement, physics]*
 - Add a test for neutron-nucleus elastic scattering model and interactor
 - Update import, process build, etc. for the newly added neutron elastic process and model

# Description
The MR includes the first implementation of the neutron physics (neutron elastic scattering) based on the Geant4's CHIPS model as a part of add neutron physics in the medium-high energy range (up to 20GeV).  The implementation is a minimal
without generating secondary ions, but features should be good enough for testing neutron transport in typical HEP user cases.  A minimal verification (cross section, scattered energy spectrum of neutrons and direction) has been done with the default Geant4 physics list, but more rigorous validations and verification are required once a test transport application is added.


whokion:
Sorry for a big set of commits! Please take your time as this MR should be totally decoupled from other tasks.  Thanks.

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Read the data for the given element.
+ */
+NeutronXsReader::result_type
+NeutronXsReader::operator()(AtomicNumber atomic_number) const
+{
+    CELER_EXPECT(atomic_number && atomic_number < AtomicNumber{92});
```
sethrj: There's no scattering data for uranium and transuranics? Or it just requires another model? Maybe it would be better to remove this hard-coded value and let it fail if there's no file at the requested element?

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,99 @@
...
+    CELER_LOG(debug) << "Reading neutron elastic xs data for Z=" << z_str;
+
+    result_type result;
+
+    // Read neutron elastic cross section data for the given atomic_number
+    {
+        std::string filename = path_ + "/el" + z_str;
+        std::ifstream infile;
+        infile.open(filename);
+        CELER_VALIDATE(infile.is_open(),
```
sethrj:
```suggestion
        std::ifstream infile(filename);
        CELER_VALIDATE(infile,
```
I think `operator bool` for istream catches more error conditions than `is_open`: https://stackoverflow.com/a/14920476/118160

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,99 @@
...
+        infile >> energy_min >> energy_max >> size >> size;
+        result.x.resize(size);
+        result.y.resize(size);
+        for (size_type i = 0; i < size; ++i)
+        {
+            CELER_ASSERT(infile);
+            infile >> result.x[i] >> result.y[i];
+            // Convert to the celeritas native length 1/[len^2] from
+	    // clhep::mm^2 as stored in G4PARTICLEXS/neutron/el data
+            result.y[i] *= ipow<2>(units::millimeter);
```
sethrj: Using quantities makes it clearer (less error prone):
```suggestion
            // G4PARTICLEXS/neutron/el stores data in clhep::mm^2
            using MmSqMicroXs = Quantity<UnitProduct<units::Millimeter, units::Millimeter>, double>;
            MmSqMicroXs input_xs;
            infile >> result.x[i] >> input_xs.value();
            result.y[i] = native_value_from(input_xs);
```

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,99 @@
...
+                       << "' (should contain cross section data)");
+
+        // Set the physics vector type
+        result.vector_type = ImportPhysicsVectorType::free;
+
+        // Read tabulated energies and cross sections
+        double energy_min = 0.;
+        double energy_max = 0.;
+        size_type size = 0;
+        infile >> energy_min >> energy_max >> size >> size;
```
sethrj, line 79:
```suggestion
        CELER_VALIDATE(size  > 0, << "incorrect neutron cross section size " << size);
```

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,189 @@
...
+    NeutronElasticIds ids;
+
+    //! Particle mass * c^2 [MeV]
+    units::MevMass neutron_mass;
+
+    //! Neutron elastic cross section (G4PARTICLEXS/neutron/elZ) data
+    NeutronElasticXsData<W, M> xs;
+
+    //! Neutron elastic differential cross section coefficients
+    NeutronElasticDiffXsData<W, M> Q2;
```
sethrj: Use words, not math symbols
```suggestion
    NeutronElasticDiffXsData<W, M> diff_xs;
```

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+    ParticleTrackView const& particle,
+    Real3 const& inc_direction,
+    IsotopeView const& target)
+    : shared_(shared)
+    , inc_energy_(particle.energy())
+    , inc_direction_(inc_direction)
+    , target_(target)
+    , neutron_mass_(value_as<Mass>(shared_.neutron_mass))
+    , neutron_p_(std::sqrt(inc_energy_.value()
+                           * (inc_energy_.value() + 2 * neutron_mass_)))
```
sethrj: Since the incoming particle is a neutron, shouldn't this just be:
```suggestion
    , neutron_p_(value_as<units::MevMomentum>(particle.momentum()))
```

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+/*!
+ * Sample the momentum transfer (\f$ Q^{2} = -t \f$) of the neutron-nucleus
+ * elastic scattering.
+ *
+ * \note This performs the same sampling routine as in Geant4's
+ *  G4ChipsElasticModel and G4ChipsNeutronElasticXS where the differential
+ *  cross section for quark-exchange and the amplitude of the scattering are
+ *  parameterized in terms of the neutron momentum (GeV/c) in the lab frame.
+ */
+class InvariantQ2Sampler
```
sethrj: Use `Sq` instead of `2` (which could mean second item, multiplied by 2, squared, ...) and maybe `MomSq`?

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+    //!@{
+    //! \name Type aliases
+    using Mass = units::MevMass;
+    //!@}
+
+  public:
+    // Construct with shared and target data, and the neutron momentum
+    inline CELER_FUNCTION InvariantQ2Sampler(NeutronElasticRef const& shared,
+                                             IsotopeView const& target,
+                                             real_type neutron_p);
```
sethrj: Input units should be `units::MevMomentum` right?

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+    //// DATA ////
+
+    ChipsDiffXsCoefficients::ChipsArray const& par_;
+
+    // Mass values in GeV
+    real_type neutron_mass_;
+    real_type target_mass_;
+
+    // Atomic mass number (A) of the target
+    real_type A_;
```
sethrj: this should be an `AtomicNumber` I think... why is it a real type since most comparisons against it are integers?

sethrj (commented):
A few initial comments... this looks great overall!

src/celeritas/neutron/executor/NeutronElasticExecutor.hh
```diff
@@ -0,0 +1,71 @@
...
+ */
+CELER_FUNCTION Interaction
+NeutronElasticExecutor::operator()(CoreTrackView const& track)
+{
+    auto particle = track.make_particle_view();
+    auto const& dir = track.make_geo_view().dir();
+    auto rng = track.make_rng_engine();
+
+    // Select a target element
+    auto material = track.make_material_view().make_material_view();
```
stognini: This funny line makes me think that should rename the first function call to `make_material_track_view()`.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+    bool heavy_target_{false};
+    // Momentum magnitude of the incident neutron (GeV/c)
+    real_type neutron_p_;
+    // Maximum momentum transfer for the elastic scattering
+    real_type max_Q2_;
+    // Parameters for the CHIPS differential cross section
+    ExchangeParameters par_Q2_;
```
stognini: Maybe rename `x_Q2` to `x_q_sq_`. I think we don't use capital letters for variables by standard, right?

src/celeritas/neutron/executor/NeutronElasticExecutor.hh
sethrj: Yeah I've also had a plan for refactoring some of the code underpinning views so that we can have just a single "material view". This was complicated by the fact that we needed a CPU-only material view that didn't have the full material state...

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
sethrj: yeah exactly. No capital letters or math symbols; use english or shortened english words (`mom_sq_`).

src/celeritas/io/NeutronXsReader.cc
whokion: Actually I mistakenly dropped the uranium when excluding transuranium elements as their neutron elastic cross sections do not exist (in G4PARTICLEXS data set) even though the CHIPS model itself uses a set of default values of parameters for them with a warning message (so, do not know whether it is meaningful or not).  Anyway, I remove the condition and make it fail when reading unavailable data file.

src/celeritas/io/NeutronXsReader.cc
whokion: Okay, I thought that using file open is more thread safe, but my recollection is hazy in these days. I take your suggestion.

src/celeritas/io/NeutronXsReader.cc
whokion: Very nice. I was thinking the similar thing, but was not good enough to think harder. Done.

src/celeritas/io/NeutronXsReader.cc
whokion: Added.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Okay, I have a big bias for Q2 (as HEP conventionally uses the capital Q^2 for the momentum transfer (in any virtual particle or quark/gluon exchanges.  I almost used q2 instead, but have failed to convince myself.  Not a big deal, so I changed it to diff_xs and  Q2->q2 in other places - I would like to keep q2 instead of q_square in most places if this is still acceptable for our coding style.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
whokion: Okay. I updated it accordingly and tried to use mass and momentum as much as possible instead of real_type in argument and declaration.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
whokion: Correct. Updated.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
whokion: Okay, using AtomicNumber.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
whokion: Again, Q^2 has a special (?) meaning for the momentum transfer square - so, `mom_sq_` is somewhat misleading too.  As it is equivalent to the Mandalstam variable `t`, I was kind of using `max_t_` or `par_t_`, but was curious that how many will catch the real meaning of `t`.  To me, I would like it keep it as x_q2_ at least for now unless you have a strong objection.

src/celeritas/io/NeutronXsReader.cc
sethrj: I've never heard that, but I've also never paid any attention to thread safety before working with Geant4 😅 I think the two should be essentially equivalent.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
sethrj: It's OK to use real_type inside the class; but the class interface (constructor, call function, although not necessarily the private functions) should use units/quantities if they're not native.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
```diff
@@ -0,0 +1,156 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Perform neutron elastic scattering based on the CHIPS (Chiral invariant
+ * phase space) model.
+ *
+ * \note This performs the sampling procedure as in G4HadronElastic,
+ * G4ChipsElasticModel and G4ChipsNeutronElasticXS, as partly documented
+ * in section 21.1.3 of the Geant4 Physics Reference (release 11.2).
+ */
+class NeutronElasticInteractor
```
sethrj: Should this be `ChipsInteractor` or `ChipsNeutronInteractor` or something? Will there be other neutron elastic scattering models? (In our neutron transport code, for elastic scattering above the thermal energy range and below 20 MeV, we use a super-simple point kinematic elastic scattering model...)

whokion: Actually, a good point for future extensions! The better choice may be `ChipsNeutronElasticModel` and `ChipsNeutronElasticInteractor` (as CHIPS supports not only neutron elastic, but other particles and inelastic interactions (In Geant4, Chips inelastic is still in use for Kaons at the high energy region).   One downside is that class names are getting longer for the combination of a specific particle and a process.  I will rename them as above (later today) since we may extend the elastic scattering for proton too.  Of course, we can always rename and restructure codes as more physics are added.  

The `neutron-HP` extension below 20MeV may be a challenging project for us as there are interference between the scattering  process and radiative decays (i.e., de/excitation) with different atomic levels, especially for heavy nuclei (which could have thousands or millions of combination for possible transitions) - may be a good project after retirement (even for you?).  Yes, we can enjoy the elastic scattering with billiard balls whenever physics need is classical.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
stognini: I agree that $Q^2$ has a special meaning. Even more so since this is a four-momentum and not the three-momentum $\vec{p}$, so I too prefer to keep it as Q. But since we lower case everything, and any private/public variable is well documented, I think `q_sq` looks pretty good (and maybe cleaner to read than something `four_mom_transf_sq`) and maintains consistency throughout the codebase.

I also think that moving from `Q2` to `q_sq` (or `QSq` for classes) is a smaller compromise than other examples in the past. We never use things like `m_eplus` or `E_eplus`. Even constants are lengthened for clarity: `hbar`---which would be universally agreed upon---we call it `hbar_planck`.

stognini: For the Mandelstam `t`, since it's also a four-momentum notation, how about something like `mandelstam_t` or `lorentz_inv_t`, since it is a lorentz invariant?

EDIT: FWIW, I am happy calling them `mandelstam_[s,t,u]` in the codebase.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
sethrj: So the interactor implementation for incident neutrons in the CHIPS model is substantially different from other hadrons?

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
hhollenb: My concern with capitalization is that in contexts like deep inelastic scattering, they're often defined as being opposite signs, e.g. if $p\_\mu$ is the incident 4-momenta of an electron and $p'_\mu$ is its final 4-momenta, then $q\_\mu = p\_\mu - p'\_\mu$ is the boson exchange momentum and $Q^2 = -q^2 \geq 0$ in the mostly minus metric. If we decide to use `q` or `Q` then we should definitely be explicit with denoting its sign. Mandelstam variables would also be a decent alternative.

I also strongly favor using `_sq` postfix instead of `2`. I found Geant's arrays of `A23` and `A17` confusing before learning they were just cached values of $A^{2/3}$ and $A^{1/7}$ resp.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
whokion: Actually not for the elastic scattering (the code structure is quite similar for different particles with different sets of XS data), but may be for inelastic (which I have not looked much as most of CHIPS inelastic channels are now obsolete in Geant4 and we are going to implement the Bertini cascade instead).  As you know, the proton scattering is not considered at this point (even though it may be easy to add once this MR is done) as it requires other processes like eloss, msc and etc.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
whokion: Okay, it is good that @hhollenb is taking QFT and bring us some new conventions!  Of course, we can rename anything as needed later - I prefer not to refactor codes too early until all physics verification with respect to Geant4 are done even though the right choice of naming convention is still a good practice and should be imposed no matter what from the beginning.

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,102 @@
...
+/*!
+ * Construct the reader using the G4PARTICLEXSDATA environment variable to get
+ * the path to the data.
+ */
+NeutronXsReader::NeutronXsReader()
+{
+    std::string const& dir = celeritas::getenv("G4PARTICLEXSDATA");
+    CELER_VALIDATE(!dir.empty(),
+                   << "environment variable G4PARTICLEXSDATA is not defined"
+                      " (needed to locate Livermore data)");
```
amandalund:
```suggestion
                      " (needed to locate neutron elastic cross section data)");
```

src/celeritas/neutron/xs/NeutronElasticMicroXsCalculator.hh
```diff
@@ -0,0 +1,82 @@
...
+#include "celeritas/Quantities.hh"
+#include "celeritas/Types.hh"
+#include "celeritas/grid/GenericCalculator.hh"
+#include "celeritas/neutron/data/NeutronElasticData.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate neutron elastic cross sections form NeutronElasticXsData
```
amandalund:
```suggestion
 * Calculate neutron elastic cross sections from NeutronElasticXsData
```

src/celeritas/neutron/xs/NeutronElasticMacroXsCalculator.hh
```diff
@@ -0,0 +1,86 @@
...
+
+#include "NeutronElasticMicroXsCalculator.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculates the macroscopic cross section.
+ */
+class NeutronElasticMacroXsCalculator
```
amandalund, line 25: Since this is essentially identical to the Livermore macro xs calculator, I wonder if there's a way we could use the same calculator for both and reduce the duplicate code...

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -416,6 +417,12 @@ CELER_FUNCTION real_type PhysicsTrackView::calc_xs(ParticleProcessId ppid,
                 params_.hardwired.eplusgg_data, material);
             result = calc_xs(energy);
         }
+        else if (model_id == params_.hardwired.chips)
+        {
+            auto calc_xs = EPlusGGMacroXsCalculator(
+                params_.hardwired.eplusgg_data, material);
```
amandalund:
```suggestion
            auto calc_xs = NeutronElasticMacroXsCalculator(
                params_.hardwired.chips_data, material);
```

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,189 @@
...
+    //// MEMBER DATA ////
+
+    IsotopeItems<ChipsDiffXsCoefficients> coeffs;
+
+    //// MEMBER FUNCTIONS ////
+
+    //! Whether all data are assigned and valid
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return !coeffs.empty() && !coeffs.empty();
```
amandalund:
```suggestion
        return !coeffs.empty();
```
Double checking? ;)

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,189 @@
...
+    //! Neutron elastic cross section (G4PARTICLEXS/neutron/elZ) data
+    NeutronElasticXsData<W, M> xs;
+
+    //! Neutron elastic differential cross section coefficients
+    NeutronElasticDiffXsData<W, M> diff_xs;
+
+    //// MEMBER FUNCTIONS ////
+
+    //! Model's minimum and maximum energy limit [MeV]
+    static CELER_CONSTEXPR_FUNCTION units::MevEnergy mim_valid_energy()
```
amandalund:
```suggestion
    static CELER_CONSTEXPR_FUNCTION units::MevEnergy min_valid_energy()
```

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,189 @@
...
+
+    template<class T>
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using ElementItems = Collection<T, W, M, ElementId>;
+
+    //// MEMBER DATA ////
+
+    Items<real_type> reals;
+    ElementItems<GenericGridData> elements;
```
amandalund: Any reason to have a separate struct for this, or could this go directly in `NeutronElasticData`?

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,189 @@
...
+ */
+template<Ownership W, MemSpace M>
+struct NeutronElasticDiffXsData
+{
+    template<class T>
+    using IsotopeItems = Collection<T, W, M, IsotopeId>;
+
+    //// MEMBER DATA ////
+
+    IsotopeItems<ChipsDiffXsCoefficients> coeffs;
```
amandalund: Same here?

amandalund (commented):
This looks great @whokion! Still working through it, but here are a few initial comments.

src/celeritas/phys/PhysicsTrackView.hh
sethrj: Maybe it's time to add kernels to replace these hard coded xs bits?

src/celeritas/neutron/executor/ChipsNeutronElasticExecutor.hh
```diff
@@ -0,0 +1,71 @@
...
+            track.make_material_view().element_scratch());
+        elcomp_id = select_el(rng);
+        CELER_ASSERT(elcomp_id);
+        track.make_physics_step_view().element(elcomp_id);
+    }
+    ElementView element = material.make_element_view(elcomp_id);
+
+    // Select a target nucleus
+    IsotopeSelector iso_select(element);
+    IsotopeView target = element.make_isotope_view(iso_select(rng));
```
amandalund, line 61: We cache the sampled element so we'll be able to ensure reproducibility if we're out of space for secondaries... will we need to do the same for the isotope (at some point)?

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,157 @@
...
+ * Perform neutron elastic scattering based on the CHIPS (Chiral invariant
+ * phase space) model.
+ *
+ * \note This performs the sampling procedure as in G4HadronElastic,
+ * G4ChipsElasticModel and G4ChipsNeutronElasticXS, as partly documented
+ * in section 21.1.3 of the Geant4 Physics Reference (release 11.2).
+ */
+class ChipsNeutronElasticInteractor
+{
+    //!@{
```
amandalund, line 38:
```suggestion
public:
    //!@{
```

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,157 @@
...
+template<class Engine>
+CELER_FUNCTION Interaction ChipsNeutronElasticInteractor::operator()(Engine& rng)
+{
+    // Scattered neutron with respect to the axis of incident direction
+    Interaction result;
+
+    // The momentum magnitude in the c.m. frame
+    real_type target_mass = value_as<Mass>(target_.nuclear_mass());
+    real_type neutron_mass = value_as<Mass>(neutron_mass_);
+    real_type neutron_energy = neutron_mass + inc_energy_.value();
```
amandalund: Maybe the member variables should be `real_type neutron_mass_` (instead of `Mass`) and `real_type neutron_energy_` (instead of `inc_energy_`, which isn't used anywhere else)?

src/celeritas/io/NeutronXsReader.cc
whokion: fixed.

src/celeritas/neutron/xs/NeutronElasticMicroXsCalculator.hh
whokion: Fixed.

src/celeritas/neutron/xs/NeutronElasticMacroXsCalculator.hh
whokion: Good idea as I will also add a similar set of calculators for `NeutronInelastic`.  I think that a templated class, something like,
```
template<class MicroXsT, typename MacroXsUnitsT>
class MacroXsCalculator
```
may be used as a common method.  As a quick look, this requires changes in the associated `MicroXsCalculator` (argument for constructor and operator) and a creation of a specific `MicroXsCalculator` before hand which imposes an explicit rule for `MacroXsCalculator`.  Please let me know if you have the better idea or if a potential consolidation may be a separated MR.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Fixed.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Fixed. Also inside `ChipsNeutronElasticModel::applicability()` where it is used.

src/celeritas/neutron/xs/NeutronElasticMacroXsCalculator.hh
sethrj: I like that idea. I think it would be better to create a new PR that refactors LivermorePEMacroXsCalculator to something like:
```c++
template<class F>
class MacroXsCalculator
{
    using ParamsRef = typename F::ParamsRef;
    using Energy = typename F::Energy;
    using XsUnits = units::Native;

    MacroXsCalculator(ParamsRef const&, MaterialView const& material);
    real_type operator()(Energy) const;
};
```

I think also `LivermorePEMicroXsCalculator` should be returning either a `units::Barn` quantity or a native (len^2) unit.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Okay, I put all data into NeutronElasticData.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Done.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Added.

src/celeritas/neutron/data/NeutronElasticData.hh
sethrj: Just gonna leave this star trek reference here...

image

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Changed them as suggested.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Well, it should be $Q^2$, not $Q2$. And Amanda `Lund`, not Rogers - sorry that I can not fully digest this joke(?) anyway.

amandalund: nor can I, star trek references are usually over my head ;)

sethrj: Not really a joke to be had except to share the bizarre associations my brain makes due to my high school nerdiness.

src/celeritas/io/NeutronXsReader.cc
```diff
@@ -0,0 +1,101 @@
...
+        MmSqMicroXs input_xs;
+        for (size_type i = 0; i < size; ++i)
+        {
+            CELER_ASSERT(infile);
+            // Convert to the celeritas native length 1/[len^2] from
+            // clhep::mm^2 as stored in G4PARTICLEXS/neutron/el data
+            infile >> result.x[i] >> input_xs.value();
+            result.y[i] = native_value_from(input_xs);
+        }
+        infile.close();
```
sethrj: Forgot to add that `close` is redundant with the destructor (we're at the end of the block so it's going out of scope. The only reason to call it independently is to be able to check `bool(infile)` afterward to make sure it closed correctly, which frankly doesn't matter.
```suggestion
```

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,129 @@
...
+    ChipsArray par{};  //!< Coefficients as a function of atomic mass numbers
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Model and particles IDs for neutron--nucleus elastic scattering.
+ */
+struct NeutronElasticIds
+{
+    ActionId action;
```
sethrj, line 54: The action IDs no longer need to be stored in device data like this (they could be instead stored as member data on the Model class itself). The only place it's used is in the `Model` where the action ID is returned (and indirectly where `make_action_track_executor` is used).

src/celeritas/phys/FourVector.hh
```diff
@@ -0,0 +1,69 @@
...
+/*!
+ * The momentum-energy four-vector (Lorentz vector).
+ */
+struct FourVector
+{
+    //!@{
+    //! \name Type aliases
+    using Real3 = Array<real_type, 3>;
+    //!@}
+
```
sethrj: You should be able to remove this and just use the typedef in `geocel/Types.hh`.

src/celeritas/phys/FourVector.hh
```diff
@@ -0,0 +1,69 @@
...
+ * Perform the Lorentz transformation (\f$ \Lambda^{\alpha}_{\beta} \f$) along
+ * the boost vector (\f$ \vec{v} \f$) for a four-vector \f$ p^{\beta} \f$:
+ *
+ * \f$ p^{\prime \beta} = \Lambda^{\alpha}_{\beta} (\vec{v}) p^{\beta} \f$.
+ *
+ */
+inline CELER_FUNCTION void boost(Real3 const& v, FourVector* p)
+{
+    const real_type v2 = dot_product(v, v);
+    CELER_ENSURE(v2 < real_type{1});
```
sethrj:
```suggestion
    CELER_EXPECT(v2 < real_type{1});
```
also `vsq` or `v_sq` ;)

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+    Real3 cm_mom = cm_p * from_spherical(cos_theta, sample_phi_(rng));
+    FourVector nlv1(
+        {cm_mom, std::sqrt(ipow<2>(cm_p) + ipow<2>(neutron_mass_))});
+
+    FourVector lv({{0, 0, value_as<units::MevMomentum>(neutron_p_)},
+                   neutron_energy_ + target_mass});
+    boost(boost_vector(lv), &nlv1);
+
+    result.direction
+        = make_unit_vector(rotate(make_unit_vector(nlv1.mom), inc_direction_));
```
sethrj: The outer call to make unit vector is unnecessary since rotating a unit vector will result in a unit vector

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
sethrj: @whokion so, is `neutron_p_(value_as<units::MevMomentum>(particle.momentum()))` correct? We've already got that accessor on the `Particle` type so it's better than duplicating the calculation here.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+    // Target nucleus
+    IsotopeView const& target_;
+
+    // Value of neutron MeVMass
+    real_type neutron_mass_;
+    real_type neutron_p_;
+
+    // Sampler
+    UniformRealDist sample_phi_;
+    detail::InvariantQ2Sampler sample_Q2_;
```
sethrj: Can you rename `2 -> _sq`  everwhere? And also please do not use capital letters in variable names.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+
+  public:
+    // Construct with shared and target data, and the neutron momentum
+    inline CELER_FUNCTION InvariantQ2Sampler(NeutronElasticRef const& shared,
+                                             IsotopeView const& target,
+                                             real_type neutron_p);
+
+    // Sample the momentum transfer
+    template<class Engine>
+    inline CELER_FUNCTION real_type operator()(Engine& rng);
```
sethrj: Is the result `MevMomentum`? It should definitely be a `Quantity` then. I think you can also use quantity to handle the transfer to and from GeV internally...

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+    real_type target_mass = value_as<Mass>(target_.nuclear_mass());
+    real_type neutron_energy = neutron_mass_ + inc_energy_.value();
+    real_type cm_p = neutron_p_
+                     / std::sqrt(1 + ipow<2>(neutron_mass_ / target_mass)
+                                 + 2 * neutron_energy / target_mass);
+
+    // Sample the scattered direction from the invariant momentum transfer
+    // squared (\f$ -t = Q^{2} \f$) in the c.m. frame
+    real_type cos_theta = 1 - real_type(0.5) * sample_Q2_(rng) / ipow<2>(cm_p);
+    clamp(cos_theta, real_type{-1}, real_type{1});
```
sethrj: The `clamp` operation is "pure" , i.e. it doesn't modify `cos_theta` but returns a clamped result, so this is a null-op. I think it should also be the burden of `sample_q2` to `CELER_ENSURE` that the result is in `[0, 2 * cm_p^2]`, and this should be an assertion here rather than a mandatory clamp.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Perform neutron elastic scattering based on the CHIPS (Chiral invariant
+ * phase space) model.
+ *
+ * \note This performs the sampling procedure as in G4HadronElastic,
+ * G4ChipsElasticModel and G4ChipsNeutronElasticXS, as partly documented
+ * in section 21.1.3 of the Geant4 Physics Reference (release 11.2).
```
sethrj: Can you add a description of the implementation here, especially if the documentation in the reference model is "partial"? Include:
- Discussion of translation between lab-frame and COM-frame using FourVector and boost
- Link to COM-frame scatter direction calculator (Q2)
- Resulting neutron and deposition energy are kinematically calculated from the scattering direction

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
sethrj: @whokion are there unpushed commits? This still looks like a real number. Can you also please avoid using upper case variable names: call it `target_a_` or something.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
```diff
@@ -0,0 +1,363 @@
...
+    real_type q2 = 0;
+
+    if (A_ == 1)
+    {
+        // Special case for the \f$ n + p \rightarrow n + p \f$ channel
+        real_type R1 = -std::expm1(-max_Q2_ * par_Q2_.slope[0]);
+        real_type R2 = -std::expm1(-max_Q2_ * par_Q2_.slope[1]);
+
+        real_type I1 = R1 * par_Q2_.expnt[0];
+        real_type I2 = R2 * par_Q2_.expnt[1] / par_Q2_.slope[1];
```
sethrj: No capital variable names. And you could even write this as
```c++
real_type const r[] = {..., ...};
```
and use `r[0]` and `r[1]`.

sethrj (changes requested):
@whokion I'm pausing my review because I think there's a lot of naming updates that need to be done, and it seems like you might have addressed some commits but not pushed.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the momentum transfer (\f$ Q^{2} = -t \f$) of the neutron-nucleus
+ * elastic scattering.
+ *
+ * \note This performs the same sampling routine as in Geant4's
+ *  G4ChipsElasticModel and G4ChipsNeutronElasticXS where the differential
+ *  cross section for quark-exchange and the amplitude of the scattering are
+ *  parameterized in terms of the neutron momentum (GeV/c) in the lab frame.
```
sethrj, line 33: How hard would it be to preprocess the data so that its units are consistent with the rest of the code? It seems a little odd to divide by 1000 when entering and then multiply by 1000 when leaving. If we don't have to do the extra division and multiplication, we can also make stronger guarantees about the resulting value and skip the extra `clamp`ing.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+
+    real_type cm_p = value_as<units::MevMomentum>(neutron_p_)
+                     / std::sqrt(1 + ipow<2>(neutron_mass_ / target_mass)
+                                 + 2 * neutron_energy_ / target_mass);
+
+    // Sample the scattered direction from the invariant momentum transfer
+    // squared (\f$ -t = Q^{2} \f$) in the c.m. frame
+    real_type cos_theta
+        = 1 - real_type(0.5) * sample_momentum_square_(rng) / ipow<2>(cm_p);
+    clamp(cos_theta, real_type{-1}, real_type{1});
```
sethrj: The `clamp` operation is "pure" , i.e. it doesn't modify `cos_theta` but returns a clamped result, so this is a null-op. I think it should also be the burden of `sample_q2` to `CELER_ENSURE` that the result is in `[0, 2 * cm_p^2]`, and this should be an assertion here rather than a mandatory clamp.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Perform neutron elastic scattering based on the CHIPS (Chiral invariant
+ * phase space) model.
+ *
+ * \note This performs the sampling procedure as in G4HadronElastic,
+ * G4ChipsElasticModel and G4ChipsNeutronElasticXS, as partly documented
+ * in section 21.1.3 of the Geant4 Physics Reference (release 11.2).
```
sethrj, line 33: Can you add a description of the implementation here, especially if the documentation in the reference model is "partial"? Include:
- Discussion of translation between lab-frame and COM-frame using FourVector and boost
- Link to `MomentumTransferSampler`
- Resulting neutron and deposition energy are kinematically calculated from the scattering direction

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+        factor = heavy_target_ ? ipow<7>(max_q_sq_) : max_q_sq_;
+        real_type r_3 = -std::expm1(-factor * par_q_sq_.slope[2]);
+        real_type r_4 = -std::expm1(-max_q_sq_ * par_q_sq_.slope[3]);
+
+        real_type i_1 = r_1 * par_q_sq_.expnt[0];
+        real_type i_2 = r_2 * par_q_sq_.expnt[1];
+        real_type i_3 = r_3 * par_q_sq_.expnt[2];
+        real_type i_4 = r_4 * par_q_sq_.expnt[3];
+        real_type i_5 = i_1 + i_2;
+        real_type i_6 = i_3 + i_5;
```
sethrj: On GPU it might be better to *not* precalculate these: we should do a test compile of this kernel and see if the register pressure here is high. (That can be a follow-on "optimization" check, just something to keep in mind.)

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+                        - par_q_sq_.slope[0])
+                       / tss;
+            }
+        }
+        else if (rand < i_5)
+        {
+            q_sq = clamp_to_nonneg(this->sample_q_sq(r_2, rng)
+                                   / par_q_sq_.slope[1]);
+            q_sq = (heavy_target_) ? std::pow(q_sq, one_fifth)
+                                   : std::pow(q_sq, one_third);
```
sethrj:
```suggestion
            q_sq = std::pow(q_sq, heavy_target_ ? one_fifth : one_third);
```

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+        real_type i_4 = r_4 * par_q_sq_.expnt[3];
+        real_type i_5 = i_1 + i_2;
+        real_type i_6 = i_3 + i_5;
+
+        real_type rand = (i_4 + i_6) * generate_canonical(rng);
+        if (rand < i_1)
+        {
+            real_type tss = 2 * par_q_sq_.ss;
+            q_sq = this->sample_q_sq(r_1, rng) / par_q_sq_.slope[0];
+            if (std::fabs(tss) > 1e-7)
```
sethrj: What does this tolerance signify? Move it to a `static constexpr real_type` function.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+                              + par_[27] / (1. + par_[28] / p2);
+            result.expnt[3]
+                = p2
+                  * (pah * par_[29] * std::exp(-pah * par_[30])
+                     + par_[31] / (1. + par_[32] * std::pow(p, par_[33])));
+            result.slope[3] = par_[34] * pa / p2 / (1. + pa * par_[35]);
+        }
+        else
+        {
+            result.expnt[0] = par_[0] / (1. + par_[1] / p4)
```
sethrj: These `1.` should all be `1` to avoid accidentally promoting to double

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+            result.expnt[2] = par_[23] / std::pow(p, par_[26])
+                                  / (1. + par_[27] / p12)
+                              + par_[24] / (1. + par_[25] / p6);
+            result.slope[2] = par_[28] / p8 + par_[29] / p2
+                              + par_[30] / (1. + par_[31] / p8);
+            result.expnt[3]
+                = (par_[32] / p4 + par_[37] / p) / (1. + par_[33] / p10)
+                  + (par_[34] + par_[35] * dl * dl) / (1. + par_[36] / p12);
+            result.slope[3] = par_[38] / (1. + par_[39] / p)
+                              + par_[40] * p4 / (1. + par_[41] * p5);
```
sethrj: These expressions are a little crazy and look error prone. Are these parameterizations documented anywhere? Can we make something like the `PolyEvaluator` class to encapsulate some of this?

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,378 @@
...
+    real_type p = value_as<units::MevMomentum>(neutron_p) * this->to_gev();
+    real_type lp = std::log(p);
+    real_type sp = std::sqrt(p);
+    real_type p2 = ipow<2>(p);
+    real_type p3 = p2 * p;
+    real_type p4 = p3 * p;
+
+    ExchangeParameters result;
+
+    if (a_ == AtomicMassNumber{1})
```
sethrj: I can't tell from looking whether the parameterizations between A=1, A<=6, and "heavy" A are different. Is it possible to preprocess all these data so that we just index into an array indexed by atomic weight, and pull out the parameters into a single evaluation?

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
```diff
@@ -0,0 +1,299 @@
...
+    real_type a9 = a8 * a;
+    real_type a10 = a9 * a;
+    real_type a12 = ipow<2>(a6);
+    real_type a14 = ipow<2>(a7);
+    real_type a16 = ipow<2>(a8);
+    real_type a32 = ipow<2>(a16);
+
+    // The differential cross-section is parameterized separately for
+    // A>6 & A<7
+    if (a < 6.5)
```
sethrj: I can tell where this statement came from 😂

Try again:
```suggestion
    if (amass <= AtomicMassNumber(6))
```

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
```diff
@@ -0,0 +1,299 @@
...
+ * in G4ChipsNeutronElasticXS::GetChipsCrossSection. The first 15 parameters
+ * of lastPAR are not used for sampling the momentum transfer between neutron
+ * and nucleus (Z > 0, N >= 1).
+ *
+ * M.V. Kossov, “Manual for the CHIPS event generator”, KEK internal report
+ * 2000-17, Feb. 2001 H/R; P.V. Degtyarenko, M.V. Kossov and H.P. Wellisch,
+ * Eur. Phys. J. A8, 217 (2000); P.V. Degtyarenko, M.V. Kossov and H.P.
+ * Wellisch, Eur. Phys. J. A9 (2001).
+ */
+void ChipsNeutronElasticModel::append_coeffs(AtomicMassNumber A,
```
sethrj:
```suggestion
void ChipsNeutronElasticModel::append_coeffs(AtomicMassNumber amass,
```

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
```diff
@@ -0,0 +1,299 @@
...
+        coeffs.par[34] = 0.8 * sa;
+        coeffs.par[35] = 0.02 * sa;
+        coeffs.par[36] = 1e8 / a3;
+        coeffs.par[37] = 3e32 / (a32 + 1e32);
+        // The effective slope (peh_ub)
+        coeffs.par[38] = 24;
+        coeffs.par[39] = 20 / sa;
+        coeffs.par[40] = 7e3 * a / (sa + 1);
+        coeffs.par[41] = 900 * sa / (1 + 500 / a3);
+    }
```
sethrj, line 293: It looks like the heavy and light nuclei have entirely different parameterizations? Maybe then it would be good to store them in two separate data types so that you can add semantics:
```c++
struct ChipsDiffXsLightCoefficients
{
     Array<real_type, 6> main_preex;
     Array<real_type, 6> main_slope;
     Array<real_type, 6> main_quadratic;
....
};
```

src/celeritas/neutron/xs/NeutronElasticMicroXsCalculator.hh
```diff
@@ -0,0 +1,82 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Compute microscopic (element) cross section
+ */
+CELER_FUNCTION
+real_type NeutronElasticMicroXsCalculator::operator()(ElementId el_id) const
+{
+    CELER_EXPECT(el_id);
```
sethrj:
```suggestion
    CELER_EXPECT(el_id < shared_.micro_xs.size());
```

test/celeritas/neutron/NeutronElastic.test.cc
```diff
@@ -0,0 +1,310 @@
...
+
+    // Set the target isotope: Cu63 (42 parameters for heavy nuclei, A > 6.5)
+    ChipsDiffXsCoefficients cu63_coeff = coeffs[IsotopeId{2}];
+    EXPECT_EQ(cu63_coeff.par.size(), 42);
+    EXPECT_EQ(cu63_coeff.par[0], 527.781478797624);
+    EXPECT_EQ(cu63_coeff.par[10], 9.842761904761872);
+    EXPECT_EQ(cu63_coeff.par[20], 4.5646562677427038);
+    EXPECT_EQ(cu63_coeff.par[30], 1984873.0860000001);
+    EXPECT_EQ(cu63_coeff.par[35], 0.15874507866387544);
+    EXPECT_EQ(cu63_coeff.par[41], 7129.2726746278049);
```
sethrj: These should all be `EXPECT_SOFT_EQ`

test/celeritas/neutron/NeutronElastic.test.cc
```diff
@@ -0,0 +1,310 @@
...
+            Interaction result = interact(rng_engine);
+            sum_energy += result.energy.value();
+            sum_angle += result.direction[2];
+            sum_count += rng_engine.count();
+        }
+        avg_rng_samples.push_back(sum_count / num_sample);
+        EXPECT_SOFT_EQ(weight * sum_energy, expected_energy[i]);
+        EXPECT_SOFT_EQ(weight * sum_angle, expected_angle[i]);
+    }
+    static int const expected_avg_rng_samples[] = {202, 602, 1103, 1703, 2303};
```
sethrj: That's a lot of RNG samples for a single collision... and I don't even see a rejection loop in the interactor. Am I missing something?

test/celeritas/neutron/NeutronElastic.test.cc
```diff
@@ -0,0 +1,310 @@
...
+    EXPECT_EQ(cu63_coeff.par[0], 527.781478797624);
+    EXPECT_EQ(cu63_coeff.par[10], 9.842761904761872);
+    EXPECT_EQ(cu63_coeff.par[20], 4.5646562677427038);
+    EXPECT_EQ(cu63_coeff.par[30], 1984873.0860000001);
+    EXPECT_EQ(cu63_coeff.par[35], 0.15874507866387544);
+    EXPECT_EQ(cu63_coeff.par[41], 7129.2726746278049);
+}
+
+TEST_F(NeutronElasticTest, basic)
+{
```
sethrj, line 164: Since there's lots of different logic for A=1 and A <=6 in the interactor itself, please add tests for those cases.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
```diff
@@ -0,0 +1,155 @@
...
+    real_type target_mass = value_as<Mass>(target_.nuclear_mass());
+
+    real_type cm_p = value_as<units::MevMomentum>(neutron_p_)
+                     / std::sqrt(1 + ipow<2>(neutron_mass_ / target_mass)
+                                 + 2 * neutron_energy_ / target_mass);
+
+    // Sample the scattered direction from the invariant momentum transfer
+    // squared (\f$ -t = Q^{2} \f$) in the c.m. frame
+    real_type cos_theta
+        = 1 - real_type(0.5) * sample_momentum_square_(rng) / ipow<2>(cm_p);
```
sethrj, line 136: It looks like this interactor is basically kinematics calculations wrapped around the momentum sampler, so in the future do you think we could use it for different models that have different $`Q^2`$ distributions?

sethrj (changes requested):
More comments now that I've had a chance to dig down into the guts of the momentum transfer sampler 😬

src/celeritas/io/NeutronXsReader.cc
whokion: Removed!

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: As all other models/data have this ActionId consistently,  It may be better to refactor all of them together with a separate MR (?).

src/celeritas/phys/FourVector.hh
whokion: Done.

src/celeritas/phys/FourVector.hh
whokion: Changed as suggested.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Okay. Thanks - forget to remove the outer one when I added the inner one.  Done.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
whokion: This was already done in the earlier commit.

src/celeritas/neutron/interactor/NeutronElasticInteractor.hh
whokion: This was also changed in the earlier commit - changed `InvariantQ2Sampler` to `MomentumTransferSampler` and `x2` to `x_sq` in relevant places.

src/celeritas/neutron/interactor/detail/InvariantQ2Sampler.hh
whokion: Actually the result is $({\rm MevMomentum})^{2}$.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Totally agree!  Not sure why the parameterization was done in GeV.  Nevertheless, I am not sure whether it is easy to re-parameterize those coefficients by MeV in order to remove this extra conversion.  For the moment, I would like to keep it as it is until all validations are done, then we may refactor it later.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Okay.  Good to know.  Remove this clamp and put an assert as cos_theta should not be the outside valid range.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Added a brief description for the operator.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Changed as suggested.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Have no idea! Added a static constexpr real_type function, `tolerance_slope_sq()`.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Okay. Fixed.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Failed to find a documentation so far.  Let's keep them as they are for this MR and refactor later after more validation with respect to Geant4 are done.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: I don't think so as the treatment for proton is quite different from the rest (as n+p channel is a simple without any nucleus structure).  Also, coefficients below A<=6 are also different forms, which may be hard to be consolidated with those with A > 6.

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
whokion: Changed as suggested.

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
whokion: Done.

src/celeritas/neutron/model/ChipsNeutronElasticModel.cc
whokion: Okay.  I will think about it, but again let's keep them as they are for now - may be more convenient for direct comparisons to Geant4 at this stage (which I actually verified all values with a couple a's between this code and Geant4) - just as a sanity check.

src/celeritas/neutron/xs/NeutronElasticMicroXsCalculator.hh
whokion: Done.

test/celeritas/neutron/NeutronElastic.test.cc
whokion: EXPECT_SOFT_EQ  gives an error, 
```
celeritas/src/celeritas/test/testdetail/TestMacrosImpl.hh:131:41: error: static assertion failed: Invalid types for soft equivalence
  131 |     static_assert(can_soft_equiv<VE, VA>(),
      |                   ~~~~~~~~~~~~~~~~~~~~~~^~
...
```

test/celeritas/neutron/NeutronElastic.test.cc
whokion: Okay, I added reset for each sampling and the average rng counts are 
expected_avg_rng_samples[] = {4, 4, 6, 6, 6};
which seems reasonable.

test/celeritas/neutron/NeutronElastic.test.cc
whokion: I added for A<=6.  The hydrogen (n+p) case is not added though - will add it later as a special case.

src/celeritas/neutron/interactor/ChipsNeutronElasticInteractor.hh
whokion: Yes.  The interactor should be same even for other projectiles (i.e. x + nucleus for any x).

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
sethrj: Yeah, it looks like there are a lot of odd exponents and factors...

test/celeritas/neutron/NeutronElastic.test.cc
sethrj: Sorry, I meant only the ones with floating point values, lines 151–156, and 141–146.

whokion: Thanks. I should have paid more attention.  Updated as suggested.

whokion:
@sethrj Trying to understand those failures 1) Are tests with float using different seeds or are those failures due to something else? 2) in the last test (clhep, ... ubuntu-cuda), values are off by a factor 10 which I do not understand.  Have any idea?  Thanks.

sethrj:
@whokion the single-precision failures are because the RNG stream changes and everything is super sensitive. if you look at the `test/celeritas/CMakeLists.txt` you'll see that most of the EM tests have ` ${_needs_double}` as part of the test call.. that'll disable the test for double precision.

For the failing `clhep` test, you need to do something like
```c++
        macro_xs.push_back(
            native_value_to<units::InvCmXs>(calc_macro_xs(MevEnergy{e})).value());
```

whokion:
@sethrj, @amandalund  Thank you for all your review comments! I hope that most of suggestions are reflected now and the MR may be good enough to be merged - I will continue to improve codes as I find more documentation related to the model and do more tests along the way.  Please let me know any major things that I missed to address.  Thanks again!

sethrj (approved):
I still feel icky about "here's some code that reproduces geant4 but we have no idea what it means, what it does, or where it's from", but I guess that's a good enough starting point for reference and comparison?

whokion:
> I still feel icky about "here's some code that reproduces geant4 but we have no idea what it means, what it does, or where it's from", but I guess that's a good enough starting point for reference and comparison?

I fully agree! On the other hand, we actually know what each component does and how things are related even though detail parameterizations (usually with respect to a certain data set, which sometime even have uncertainties or incompleteness) are not known or fully justified as the nature of NP would not reveal underlying physics cleanly.  Anyway, the best standard candle may be physics validation with respect to data besides Geant4 and other MCs results.  I will search more documentation on CHIPS or try to contact Kossov, if reachable, how the chips parameterizations were done and put more documentation into the code.  After all, details may not be sometimes digestible, but still good to have solid background information on them.  Otherwise, we should write a new simulation model based on the `celeriton` theory.

src/celeritas/neutron/data/NeutronElasticData.hh
```diff
@@ -0,0 +1,129 @@
...
+    NeutronElasticIds ids;
+
+    //! Particle mass * c^2 [MeV]
+    units::MevMass neutron_mass;
+
+    //! Microscopic (element) cross section data (G4PARTICLEXS/neutron/elZ)
+    Items<real_type> reals;
+    ElementItems<GenericGridData> micro_xs;
+
+    //! A-dependent coefficients for the momentum transfer) of the CHIPS model
```
amandalund:
```suggestion
    //! A-dependent coefficients for the momentum transfer of the CHIPS model
```

test/celeritas/neutron/NeutronTestBase.cc
```diff
@@ -0,0 +1,158 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Set particle parameters.
+ */
+void NeutronTestBase::set_particle_params(ParticleParams::Input inp)
+{
+    CELER_EXPECT(!inp.empty());
+    particle_params_ = std::make_shared<ParticleParams>(std::move(inp));
+    ps_ = StateStore<ParticleStateData>(particle_params_->host_ref(), 1);
+    //    cutoff_params_ = {};
```
amandalund:
```suggestion
```

test/celeritas/neutron/NeutronElastic.test.cc
```diff
@@ -0,0 +1,339 @@
...
+                                                      1.16444e-24,
+                                                      4.78256e-25,
+                                                      5.04635e-25};
+
+    real_type energy = 1e-5;
+    real_type const factor = 1e+1;
+    for (auto i : range(expected_micro_xs.size()))
+    {
+        XsCalculator calc_micro_xs(shared, MevEnergy{energy});
+        EXPECT_SOFT_EQ(calc_micro_xs(el_id), expected_micro_xs[i]);
```
amandalund, line 94: Since these cross sections are so small, I think even `EXPECT_SOFT_EQ(0, calc_micro_xs(el_id))` will pass. We could do something like convert these to barns before checking.

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,382 @@
...
+    //// COMMON PROPERTIES ////
+
+    // Covert from clhep::MeV value to clhep::GeV value
+    static CELER_CONSTEXPR_FUNCTION real_type to_gev() { return 1e-3; }
+
+    // S-wave for neutron p < 14 MeV/c (kinetic energy < 0.1 MeV)
+    static CELER_CONSTEXPR_FUNCTION Momentum s_wave_limit()
+    {
+        return Momentum{14};
+    }
```
amandalund: Not sure how much it will matter, but it looks like the limit in Geant4 is only approximately 14 MeV/c (actually 13.568559 MeV/c, or log of momentum in GeV/c = -4.3).

amandalund (approved):
Great work @whokion, thanks for our first neutron model! I agree there are still some parts that could be clarified/improved, but it makes sense to wait until after we've done some initial validation.

src/celeritas/neutron/data/NeutronElasticData.hh
whokion: Fixed.

test/celeritas/neutron/NeutronTestBase.cc
whokion: Fixed.

test/celeritas/neutron/NeutronElastic.test.cc
whokion: Okay.  I will consider this when I consolidate MacroXsCalculator where we probably use a common unit convention for all XsCalculator (next MR).

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
whokion: Okay. I changed it to 13.568559  even though it should be a rough value anyway.

whokion:
@amandalund Thanks for additional review comments!

src/celeritas/neutron/interactor/detail/MomentumTransferSampler.hh
```diff
@@ -0,0 +1,383 @@
...
+    //// COMMON PROPERTIES ////
+
+    // Covert from clhep::MeV value to clhep::GeV value
+    static CELER_CONSTEXPR_FUNCTION real_type to_gev() { return 1e-3; }
+
+    // S-wave limit for neutron, log(p) < -4.3 (GeV/c) (kinetic energy < 0.1
+    // MeV)
+    static CELER_CONSTEXPR_FUNCTION Momentum s_wave_limit()
+    {
+        return Momentum{13.568559};
```
sethrj: If you are going add 6 digits, might as well add them all, right?

```suggestion
        return Momentum{13.568559012200934}; // = exp(-4.3) * 1000
```

whokion: Updated as suggested.  Thanks!

PR #1119, opened 2024-02-21
Implement convex regions

sethrj:
This adds a few of the fundamental convex regions needed to build Geant4 (see #1078). Like Geant4 and unlike SCALE, these regions are all z-aligned and centered on the origin: it's expected that SCALE shapes will combine these primitives with a Transform (translation and/or permutation).
- Box
- Prism (extruded regular polygon)

DEFERRED to a follow-on commit:
- Cone
- Cylinder
- Ellipsoid
- Sphere

Each of these regions implements:
- Surface construction (automatically named and transformed using the convex surface builder #1113)
- Exterior bounding box construction
- Maximal-volume interior bounding box construction

Not implemented for now:
- Volume calculation
- Other convex regions (elliptical cylinder, half-hyperbola, elliptical cone, trapezoid, tetrahedron, parallelepiped, ...)

PR #1121, opened 2024-02-22
Add infix string builder

sethrj:
This is a utility intended for testing, broken off of #1119. It converts a CSG node to a string to easily read the logic expression.


elliottbiondo:
Approving after erroneously reviewing this code within #1119

elliottbiondo (approved):

PR #1122, opened 2024-02-22
Workaround missing G4persistency target bug in Geant4 11.2

drbenmorgan:
As reported on Slack, and the same report on [VecGeom JIRA](https://its.cern.ch/jira/browse/VECGEOM-621), Geant4 11.2 split the `G4persistency` target into subcomponents:

- `G4mctruth`
- `G4geomtext`
- `G4gdml` (optional, only if Geant4 built with GDML support)

and so projects use the direct CMake targets will fail to link with these versions. A bug fix will be made in Geant4 itself to address this, but we'll still need to support the affected versions of 11.2.

Workaround this issue in Celeritas by translating request for `persistency` target in `celeritas_get_g4libs` to appropriate underlying targets. Translation rather than creation of our own imported target was chosen to:

- Avoid possible (albeit unlikely) clashes with other projects that might create the same target (and likewise not mess with the `Geant4::` CMake namespace).
- Reduce dependence on use of Celeritas' FindGeant4 wrapper by downstream projects.

### DIFF
--- a/cmake/CeleritasUtils.cmake
+++ b/cmake/CeleritasUtils.cmake
@@ -644,31 +644,30 @@ function(celeritas_get_g4libs var)
       set(_ext "-static")
     endif()
 
-    foreach(_lib IN LISTS ARGN)
-      set(_lib Geant4::G4${_lib}${_ext})
-
-      # Tasking was merged into core management
-      if(_lib MATCHES "tasking" AND NOT TARGET "${_lib}")
-        continue()
-      endif()
-      # Workaround for differing target names in 11.2.0,1 for G4persistency
-      # We do this here and not in FindGeant4 because we have no
-      # guarantee projects using Celeritas and Geant4 won't mess with target
-      # names themselves (and if we had to create a "celeritas::" target,
-      # we'd still have to specialize here).
-      if(_lib MATCHES "persistency" AND NOT TARGET "${_lib}")
+    foreach(_shortlib IN LISTS ARGN)
+      set(_lib Geant4::G4${_shortlib}${_ext})
+      if(TARGET "${_lib}")
+        list(APPEND _result "${_lib}")
+      elseif(_shortlib STREQUAL "persistency")
+        # Workaround for differing target names in 11.2.0,1 for G4persistency
+        # We do this here and not in FindGeant4 because we have no
+        # guarantee projects using Celeritas and Geant4 won't mess with target
+        # names themselves (and if we had to create a "celeritas::" target,
+        # we'd still have to specialize here).
         list(APPEND _result Geant4::G4mctruth${_ext} Geant4::G4geomtext${_ext})
         if(TARGET "Geant4::G4gdml${_ext}")
           list(APPEND _result Geant4::G4gdml${_ext})
         endif()
+      elseif(_shortlib STREQUAL "tasking")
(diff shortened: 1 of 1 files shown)


cmake/CeleritasUtils.cmake
```diff
@@ -647,6 +647,10 @@ function(celeritas_get_g4libs var)
     foreach(_lib IN LISTS ARGN)
       set(_lib Geant4::G4${_lib}${_ext})
 
+      # Tasking was merged into core management
+      if(_lib MATCHES "tasking" AND NOT TARGET "${_lib}")
+        continue()
+      endif()
```
drbenmorgan: If we call

```
celeritas_get_g4libs(OUTPUT_VAR tasking)
```

then `OUTPUT_VAR` would be empty... The problem here is if `tasking` is requested, then do we want the stuff in `G4global`, or that in `G4run`? I'm guessing that in _most_ cases, it's going to be `G4run`, but I'd have to look at the specifics of each call.

sethrj: Maybe we just silently omit libraries that aren't there, and specify `tasking run` as arguments so that one or the other will get picked up?

drbenmorgan: Yes, that would work, we'll just have to be aware of that should those calls change.

sethrj: Well, how about we do it with the manual library checks and matching, and use the newer (more granular) libraries in the *calls* to this function and map them to the older (coarser) `persistency` library.

cmake/CeleritasUtils.cmake
```diff
@@ -646,10 +646,27 @@ function(celeritas_get_g4libs var)
...
+      # names themselves (and if we had to create a "celeritas::" target,
+      # we'd still have to specialize here).
+      if(_lib MATCHES "persistency" AND NOT TARGET "${_lib}")
+        list(APPEND _result Geant4::G4mctruth${_ext} Geant4::G4geomtext${_ext})
+        if(TARGET "Geant4::G4gdml${_ext}")
+          list(APPEND _result Geant4::G4gdml${_ext})
+        endif()
       else()
-        list(APPEND _result "${_lib}")
+        if(NOT TARGET "${_lib}")
```
sethrj: The cmake would probably be clearer if this were all inverted (and since we're reusing the "short" library name input by the user, it would make it easier not to rewrite it on the first line:
```cmake
foreach(_shortlib IN LISTS ARGN)
set(_lib Geant4::G4${_shortlib}${_ext})
if(TARGET "${_lib}")
  list(APPEND _result "${_lib}")
elseif(_short_lib STREQUAL "persistency")
  ....
elseif(...)
...
endif()
```

drbenmorgan: Yes, that's much nicer - implemented in the latest commit.

cmake/CeleritasUtils.cmake
drbenmorgan: I think for now it's probably better to stick with the existing names _for now_. We could review this depending on what is done in Geant4 11.2.2/11.3 with bug fixes in the former and what granularity is arrived at in 11.3.

sethrj (approved):
Nice! Thanks for that.

pcanal (approved):

PR #1124, opened 2024-02-23
Fix missing explicit instantiations in ORANGE CSB

sethrj:
It turns out that implicitly instantiating templates in a translation unit does *not* prevent those templates from being inlined, so the symbols won't necessarily be available when optimizations are turned on. (Using `std::visit` to easily instantiate function calls has been [tried](https://indii.org/blog/combinatorial-instantiation-of-templates-with-std-variant/) and [documented as failing](https://indii.org/blog/revisited-combinatorial-instantiation-of-templates-with-std-variant/), but the fix there uses non-standard extensions which I don't want to.)

This just adds explicit instantiations for the function calls in the convex surface builder.


drbenmorgan (approved):
Thanks @sethrj! This fixes the compile error on macOS(Sonoma) so good to go on that side. I'll approve modulo the errors in the doc build to fix!

PR #1125, opened 2024-02-23
Implement second-order convex regions

sethrj:
Follow-on to #1119.  This implements:
- Cone
- Cylinder
- Ellipsoid
- Sphere

There is a bugfix in the deduplication code (adding `+ 0.0` to clear any potential negative zero flag) to prevent simplified surfaces from having negative zeroes. This will make output more robust and hopefully do the same if `-ffast-math` is enabled.

PR #1126, opened 2024-02-23
Add shape and volume builder

sethrj:
Follow-on to #1125

This defines the beginning of the `Object` hierarchy, the `VolumeBuilder` helper class for constructing volumes from a shape, and the `Shape` class. There will also be `Transformed`, `Solid`, and other implementations of the `ObjectInterface`.

```mermaid
classDiagram
  Object <|-- Shape
  ShapeBase <|-- Shape
  class Object {
    +string_view label()*
    +NodeId build(VolumeBuilder&)*
  }
  <<Interface>> Object
  class ShapeBase {
    #void build_interior(ConvexSurfaceBuilder&)*
  }
  <<Abstract>> ShapeBase
  class Shape {
    -string label;
    -ConvexRegion region;
  }
  Shape *-- ConvexRegion
  class ConvexRegion {
    +void build(ConvexSurfaceBuilder&)*
  }
  <<Interface>> ConvexRegion
  ConvexRegion <|-- Box
  ConvexRegion <|-- Sphere
  class Box {
    -Real3 halfwidths
  }
  class Sphere {
    -real_type radius
  }

  Shape <|.. BoxShape
  Shape <|.. SphereShape

  BoxShape *-- Box
  SphereShape *-- Sphere
```


sethrj:
Thanks for the reviews @elliottbiondo ! This one is next.

src/orange/orangeinp/Shape.hh
```diff
@@ -0,0 +1,120 @@
...
+/*!
+ * A simple, convex region of space.
+ *
+ * This is an abstract class that implements \c build for constructing a volume
+ * by dispatching to a method \c build_interior that the daughters must
+ * override using a convex region.
+ *
+ * Use the implementation classes \c XShape where \c X is one of the convex
+ * region types in ConvexRegion.hh :
+ * - \c BoxShape
```
elliottbiondo, line 29: Since it's XShape you can just say Box, Cone, Cylinder, etc

elliottbiondo (approved):
There are quite a few levels of abstraction here, which I understand is necessary to build geometries from different sources. At some point it would be nice to have a brief, high-level design diagram/readme/document referenced within the orangeinp dir concisely summarizing what goes on in each step from convex surfaces -> region -> shape -> object -> (csg unit) -> volume. This would also be helpful because I think there are some minor naming clashes, such as "unit" within both CSGUnits, and universe units. Obviously these are very different things, but they could be easily mixed up if one does not know better.

src/orange/orangeinp/Shape.hh
sethrj: True, but I wanted to use the exact type alias name to elevate the type aliases which have to be at the bottom of the file. I'd hoped also that the doxygen would generate links from those exact names, but it seems not to ...

sethrj:
Thanks! I will definitely add some diagrams once this gets to the "build a unit/scene" phase.

PR #1129, opened 2024-02-23
Refactor MSC params and add Wentzel VI params and data

amandalund:
This add a data and params class for the Wentzel VI MSC model and a helper class for constructing the MSC params. I've tried to reduce the amount of duplicate code with some common structs for the data and helper functions for building the params, but would be happy for other suggestions on how to make this cleaner.


src/celeritas/em/MscParams.hh
```diff
@@ -0,0 +1,74 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+class ParticleParams;
+class MaterialParams;
+struct XsGridData;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Interface class for multiple scattering.
```
sethrj: It's not an interface class if has concrete methods... I suspect we want composition instead of inheritance. Maybe we should incorporate the common MSC options into the main physics params?

src/celeritas/em/MscParams.hh
```diff
@@ -0,0 +1,74 @@
...
+    //// HELPER FUNCTIONS ////
+
+    void build_parameters(MscParameters*, Options const&) const;
+    void build_ids(MscIds*, ParticleParams const&) const;
+    void build_xs(XsValues*,
+                  Values*,
+                  ParticleParams const&,
+                  MaterialParams const&,
+                  VecImportMscModel const&,
+                  ImportModelClass) const;
```
sethrj: The implementation of these definitely seem like they should be a helper class rather than a base class.

src/celeritas/io/ImportParameters.hh
```diff
@@ -43,6 +43,8 @@ struct ImportEmParameters
     double lowest_electron_energy{0.001};
     //! Whether auger emission should be enabled (valid only for relaxation)
     bool auger{false};
+    //! MSC geometry factor
```
sethrj: What does the factor do? I don't even see it being used...

src/celeritas/em/data/MscData.hh
```diff
@@ -0,0 +1,46 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Common settable parameters and default values for multiple scattering.
+ */
+struct MscParameters
+{
+    real_type lambda_limit{1 * units::millimeter};  //!< lambda limit
+    real_type geom_fact{2.5};  //!< geometry factor
+    real_type range_fact{0.04};  //!< range factor for e-/e+ (0.2 for muon/h)
+    real_type safety_fact{0.6};  //!< safety factor
```
sethrj: A little more commentary on these would be useful.

sethrj (commented):
I think it may be time to incorporate some of the core MSC functionality into the physics. I think it would also be not too hard to make a tiny PR where you do the UrbanMsc renaming (`mirror_`->`data_`, `import`->`data`, `params`->`msc_params`) and use `git reset $(git commit-tree ...)` to update this PR without having to deal with conflicts.

src/celeritas/em/MscParams.hh
amandalund: Yeah, I was debating this too.. I'll refactor into a helper class.

src/celeritas/em/MscParams.cc
```diff
@@ -0,0 +1,181 @@
...
+/*!
+ * Validate and save MSC parameters common to all models.
+ */
+void MscParams::build_parameters(MscParameters* params,
+                                 Options const& options) const
+{
+    CELER_VALIDATE(options.lambda_limit > 0,
+                   << "invalid lambda_limit=" << options.lambda_limit
+                   << " (should be positive)");
+    CELER_VALIDATE(options.safety_fact >= 0.1,
```
whokion: Probably a reasonable low limit for `safety_fact`.  Nevertheless, 0.1 is a somewhat arbitrary number. Could this be just > 0 as there exists the default value?

src/celeritas/em/MscParams.cc
```diff
@@ -0,0 +1,181 @@
...
+                   << "invalid safety_fact=" << options.safety_fact
+                   << " (should be >= 0.1)");
+    CELER_VALIDATE(options.range_fact > 0 && options.range_fact < 1,
+                   << "invalid range_fact=" << options.range_fact
+                   << " (should be within 0 < limit < 1)");
+    CELER_VALIDATE(options.geom_fact >= 1,
```
whokion: Again, is this magic number independent from a specific model-implementation?  What is the true meaning of geom_fact and where is it used in Urban and WenzelVI commonly.

src/celeritas/em/MscParams.cc
```diff
@@ -0,0 +1,181 @@
...
+{
+    ids->electron = particles.find(pdg::electron());
+    ids->positron = particles.find(pdg::positron());
+    CELER_VALIDATE(ids->electron && ids->positron,
+                   << "missing e-/e+ (required for MSC)");
+
+    // TODO: change IDs to a vector for all particles. This should apply
+    // to muons and charged hadrons as well
+    if (particles.find(pdg::mu_minus()) || particles.find(pdg::mu_plus())
+        || particles.find(pdg::proton()))
```
whokion: Is proton the only changed hardron applicable or are are others also applicable?  Probably there will be  separated implementations for hadrons and ions, we may just change the comment from `hadrons` to `protons` in the line 66 for the moment.

src/celeritas/em/MscParams.cc
```diff
@@ -0,0 +1,181 @@
...
+    {
+        CELER_LOG(warning) << "Multiple scattering is not implemented for for "
+                              "particles other than electron and positron";
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build the scaled cross section data.
```
whokion: To be specific, can you add what `scaled` means here? the scaled (macro) cross section by energy^2?

whokion (commented):
Good work!  I just left some minor comments (mostly at places from old codes).

src/celeritas/io/ImportParameters.hh
amandalund: The geometry factor is used in the step limitation algorithm (in Urban only in the "distance to boundary" algorithm, which we haven't implemented; in Wentzel VI also outside of that). I can hold off on adding this until we add the code where it is used.  It's to make sure a particle takes a minimum number of steps within a volume by imposing the limit `distance_to_boundary` / `geom_factor` on the step after entering a volume.

src/celeritas/em/MscParams.cc
amandalund: Replied above about the meaning of the geometry factor. Since `distance_to_boundary` / `geom_factor` is used to limit the step it should be >= 1.

src/celeritas/em/MscParams.cc
amandalund: This was chosen because it's the limit Geant4 places on the safety factor in [G4EmParameters](https://geant4.kek.jp/lxr/source/processes/electromagnetic/utils/src/G4EmParameters.cc#L855:~:text=853%20%7D%0A854-,855,-void%20G4EmParameters%3A%3A) (although I am also not sure what is special about 0.1).

src/celeritas/em/MscParams.cc
amandalund: Right, scaled by energy^2: I'll be more specific.

src/celeritas/io/ImportParameters.hh
sethrj: Yeah, could you delete it since it's not used in this PR, and `git revert` the deletion when you add the capability?

amandalund:
Thanks @sethrj and @whokion and sorry for letting this one sit for so long... I've finally updated this branch with your suggestions.

src/celeritas/em/WentzelVIMscParams.cc
```diff
@@ -0,0 +1,80 @@
...
+    helper.build_ids(&host_data.ids);
+    helper.build_xs(&host_data.xs, &host_data.reals);
+
+    // Save electron mass
+    host_data.electron_mass = particles.get(host_data.ids.electron).mass();
+
+    // Save high/low energy limits
+    auto const& grid = host_data.xs[ItemId<XsGridData>(0)].log_energy;
+    host_data.params.low_energy_limit = MevEnergy(std::exp(grid.front));
+    host_data.params.high_energy_limit = MevEnergy(std::exp(grid.back));
```
sethrj: Maybe we should add emin/emax to the XsCalculator clsas? This is done here and in UrbanMsc and it's a bit low level ...

src/celeritas/phys/PhysicsData.hh
```diff
@@ -239,6 +239,10 @@ struct PhysicsParamsScalars
     Energy lowest_electron_energy{};  //!< Lowest e-/e+ kinetic energy
     real_type linear_loss_limit{};  //!< For scaled range calculation
     real_type fixed_step_limiter{};  //!< Global charged step size limit [len]
+    real_type lambda_limit{};  //!< lambda limit
+    real_type range_fact{};  //!< range factor for e-/e+ (0.2 for muon/h)
+    real_type safety_fact{};  //!< safety factor
```
sethrj: Can you put these in a separate MSC section and note that the documentation for them is in `\c PhysicsParams`? Also is there a reason we abbreviate "factor" as "fact"? It doesn't save that much space...

sethrj (approved):
Looks good. Thanks @amandalund .

src/celeritas/phys/PhysicsData.hh
amandalund: Sure. The name would be a question for @whokion :) I'm happy to add the `or`.

whokion: Changing to `factor` is good!

whokion (approved):
Thanks for more touchups. Looks good!  Nothing to do with this MR, but the class name WentzelModel/Data/... seems to be too generic now as there is WentzelVIMsc.  Probably we may change WentzelModel/Data/... to something more specific later (something like WentzelCSModel/Data where CS (or Cs) stands for CoulombScattering).  Also, may have a sub-directory em/params to put all Params classes if there is no particular reason that they should be in em.

amandalund:
Thanks @whokion, totally agree about the naming (while working on this I kept opening `WentzelData` thinking it was `WentzeVI`). Maybe `CoulombScattering[/Data/Model/Interactor]`? I think adding a params subdirectory also makes sense... should I follow up on this in a separate PR?

whokion:
@amandalund @sethrj  Yes, the proposed renaming and directory arrangement can be done in a separate MR or later.  `CoulomScattering` sounds good to me. Thank again for this nice refactoring work!

sethrj (approved):
Thanks @amandalund !

PR #1128, opened 2024-02-23
Add transformed object construction

sethrj:
**CHAINED** follow-on to #1127. This adds support for constructing objects that have transformations applied to them, including transformations embedded inside other transforms.

```mermaid
classDiagram
  Object <|-- Transformed
  class Object {
    +string_view label()*
    +NodeId build(VolumeBuilder&)*
  }
  <<Interface>> Object
  class Transformed {
    -SPConstObject obj
    -VariantTransform transform
  }

  Transformed *-- Object
```

PR #1130, opened 2024-02-24
Add CSG operations on objects

sethrj:
**CHAINED** on #1128

This adds `NegatedObject`, `AnyObjects`, and `AllObjects` for defining CSG operations on other objects. It also defines two helper functions for common use cases: `make_subtraction`, which returns `A - B ==> all(A, !B)`; and `make_rdv`, which constructs the "region definition vectors" used by SCALE (which is an ALL-ed combination of objects and negated objects).


sethrj:
@elliottbiondo This one next 😄

elliottbiondo (approved):

PR #1131, opened 2024-02-26
Add transform simplifier

sethrj:
This adds a class to simplify nearly-equivalent transforms, dropping a transformation to a translation and/or a translation to no-transformation. The criteria for removing a rotation matrix is that it changes the position of an O(1) point no more than O(epsilon).


sethrj:
@elliottbiondo This one next, I think? 😄

src/orange/orangeinp/detail/CsgUnitBuilder.cc
```diff
@@ -49,6 +50,16 @@ BoundingZone const& CsgUnitBuilder::bounds(NodeId nid) const
     return iter->second.bounds;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Insert transform with simplification and  deduplication.
```
elliottbiondo: Extra space between "and" and "deduplication" --- not sure if Doxygen can strip that out

src/orange/transform/TransformSimplifier.hh
```diff
@@ -0,0 +1,77 @@
...
+    // Construct with tolerance
+    explicit inline TransformSimplifier(Tolerance<> const& tol);
+
+    //! No simplification can be applied to a null transformation
+    VariantTransform operator()(NoTransformation const& nt) { return nt; }
+
+    // Translation may simplify to no transformation
+    VariantTransform operator()(Translation const& nt);
+
+    // One transformation step can simplify to translation
```
elliottbiondo: Clarify here and in the .cc that it can simplify to a Translation *or* NoTranslation

elliottbiondo (approved):

PR #1136, opened 2024-03-01
Update CI images and workflow actions

sethrj:
Lots of our github actions are showing warnings [because github has deprecated Node 16](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), so we need to update our actions.

I've moved all of the no-HIP rocm builds to native github runners using spack and enabled ccache, so hopefully the pipeline time will decrease.

The new actions do *not* work on the versions of centos/ubuntu used by our GPU docker images, so I'm going to set up new docker containers soon (and finally expand geant4 testing to get the whole 10.5–11.2 range).


.github/workflows/build-docker.yml
```diff
@@ -11,7 +11,13 @@ jobs:
   # TODO: this currently includes non-GPU builds as well
   # and won't actually run on device
   gpu:
-    name: gpu
+    name: >-
+      ${{format('{0}-{1}-{1}{2}{3}',
```
pcanal: `{1}` is repeated and the last of the 5 arguments seems ignored, is that intentional?

.github/workflows/build-fast.yml
```diff
@@ -52,30 +54,22 @@ jobs:
...
-          restore-keys: ccache-${{matrix.runner}}-${{matrix.compiler}}-${{matrix.version}}
+          key: ccache-${{env.GHA_JOB_NAME}}-${{github.run_id}}
+          restore-keys: |
+            ccache-${{env.GHA_JOB_NAME}}
       - name: Zero ccache stats
         run: |
           ccache -z
       - name: Configure Celeritas
         run: |
-          mkdir build && cd build
```
pcanal, line 65: Why is the creation of the build directory no longer needed?

.github/workflows/build-docker.yml
```diff
@@ -25,38 +31,19 @@ jobs:
           - buildtype: 'debug'
             image: 'centos-rocm' # Debug builds don't work with HIP
         include:
-          - special: 'asan'
```
pcanal, line 28: I don't see the `asan` part of the matrix moved anywhere.  Is this for later or did I miss it?

.github/workflows/build-fast.yml
```diff
@@ -26,16 +26,18 @@ jobs:
...
+      GHA_JOB_NAME: "${{matrix.runner}}-${{matrix.compiler}}-${{matrix.version}}"
       CCACHE_DIR: "${{github.workspace}}/.ccache"
       CCACHE_MAXSIZE: "10G"
+      CMAKE_PRESET: fast
       CC: ${{matrix.compiler}}-${{matrix.version}}
       CXX: ${{matrix.compiler == 'gcc' && 'g++' || 'clang++'}}-${{matrix.version}}
+    runs-on: >-
+      ${{  matrix.runner == 'focal' && 'ubuntu-20.04'
+        || matrix.runner == 'jammy' && 'ubuntu-22.04'
+        || null
```
pcanal: (pre-existing).  Can you remind me what is the purpose of that `|| false`

.github/workflows/build-fast.yml
sethrj: It's handled by cmake presets now

.github/workflows/build-docker.yml
sethrj: Nope! Thanks for the eagle eyed catch.

.github/workflows/build-docker.yml
sethrj: Ooh also good catch, I meant to move this to `build-spack` but apparently forgot.

.github/workflows/build-fast.yml
sethrj: I just put it there to cause an understandable error if the `runner` isn't one of the preselected types?

.github/workflows/build-fast.yml
pcanal: So preset as a feature that create the directory requested when they are not there yet?

.github/workflows/build-fast.yml
pcanal: I am not sure how that works out to changing the error .... with or without the `|| null` the expression is logically the same, isn't it?

.github/workflows/build-fast.yml
sethrj: Correct:
```json
{
  "configurePresets": [
    {
```

.github/workflows/build-fast.yml
sethrj: Um. I guess you're right, the expression would evaulate to `false`? I'd been thinking of this more as a C-style ternary operator. I'll delete the unnecessary last line.

pcanal: >  the expression would evaulate to false?

Yes, it is explicitly send in the github action doc on expression that `null` turns to `false`.

pcanal (approved):
LGTM.  Thanks.

Issue #1078, opened 2024-01-03
Add ORANGE construction of single units


sethrj:
This is the next step for implementing GPU geometry in Celeritas for HEP applications.

# Construction object hierarchy

The planned hierarchy is:
 - Object
   - Negated[object] // Invert the embedded object
   - Joined[object...]  // Apply union/intersection
   - Transformed[object] (could be permuted, scaled, reflected?)
   - Solid // Joined-at-birth shapes, usually not convex
   - Shape // Convex intersections of surfaces

## Object

Abstract base class for constructing a CSG node.

## Solid

These objects are the building blocks of Geant4 geometries. Solids are usually concave when used (but most are convex in the "degenerate" case)

SCALE solids:
- Cylinder segment
- General quadric (?)
- Ring

## Shapes

Shapes are convex so the safety distance is the closest distance to any surface (when inside). When outside, we can probably construct a bbox for the shape and (if outside the bbox) return that as an approximate conservative distance to that or (if inside the bbox) return the closest distance to a surface.

SCALE shapes:
- Cone
- Cuboid
- Cylinder
- Elliptical cylinder
- Ellipsoid
- Extruded convex polygon
- Hopper
- Infinite (?)
- Parallelepiped
- Plane
- Prism
- Rhombdod
- Right tetrahedron
- Slab
- Sphere
- Triangular prism
- Wedge

SCALE "joined" shape
- Chorded shapes

Geant4 solids:
- Box (=cuboid)
- Parallelepiped/Para
- Orb (= sphere)
- Trd (= hopper)
- EllipticalTube/ELTU (= elliptical cylinder)
- Ellipsoid (= chorded ellipsoid)
- EllipticalCone
- GenTrap/GenericTrap
- Paraboloid (= chorded GQ)
- Tet
- Trapezoid/Trap

## Non-convex shapes

- Torus

sethrj:
Solids required for HGCAL:
- Box
- Polyhedra with 2 z planes (aka prism) with a rotated start angle
- Subtraction (box minus polyhedron)
- Tube
- Polycone with 2 planes and equal radii (aka a translated cylinder)

There are no reflections, and the maximum depth of logical volumes is 9.

[hgcal-volumes.pdf](https://github.com/celeritas-project/celeritas/files/14335376/hgcal-volumes.pdf)

sethrj:
[...]

PR #1137, opened 2024-03-01
Add CSG solids

sethrj:
**CHAINED** on #1142

This adds basic solids, which are shapes with interiors optionally subtracted, and azimuthal sections optionally removed.


sethrj:
@elliottbiondo This is the end of the queue!! 🎉

src/orange/orangeinp/Solid.hh
```diff
@@ -0,0 +1,185 @@
...
+
+namespace celeritas
+{
+namespace orangeinp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Define the angular region of a solid.
+ *
+ * This angle is a cross section perpendicular to the z axis, with a start
```
elliottbiondo, line 27: The angle is not a cross section perpendicular to the z axis, but rather it is defined on a cross section perpendicular to the z axis.

elliottbiondo: The angle is not a cross section perpendicular to the z axis, but rather it is defined on a cross section perpendicular to the z axis.

src/orange/orangeinp/Solid.hh
```diff
@@ -0,0 +1,185 @@
...
+ * angle of zero corresponding to the positive x axis. An interior angle of one
+ * results in no radial excluded in the resulting solid. A interior angle of
```
elliottbiondo, line 30: Clarify that  angle = 0.1 and angle = 0.9 would result in the same solid, just rotated.

src/orange/orangeinp/Solid.cc
```diff
@@ -0,0 +1,169 @@
...
+    if (auto* exclu = this->excluded())
+    {
+        // Construct the excluded region by building a convex solid, then
+        // negating it
+        NodeId smaller = build_convex_region(
+            vb, std::string{this->label()}, "excluded", *exclu);
+        nodes.push_back(vb.insert_region({}, Negated{smaller}));
+    }
+
+    if (auto const& sea = this->enclosed_angle())
```
elliottbiondo, line 84: I don't understand what is going on here. Won't this assignment operation always evaluate to true?

src/orange/orangeinp/Solid.hh
```diff
@@ -0,0 +1,185 @@
...
+    Turn interior() const { return interior_; }
+
+  private:
+    Turn start_{0};
+    Turn interior_{1};
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * A hollow shape with an optional start and end angle.
```
elliottbiondo, line 76: Can you say something like "excluded inner region" instead of hollow? Hollow made me think you were talking about a surface shell.

src/orange/orangeinp/Solid.hh
```diff
@@ -0,0 +1,185 @@
...
+    //! Allow construction and assignment only through daughter classes
+    SolidBase() = default;
+    virtual ~SolidBase() = default;
+    CELER_DEFAULT_COPY_MOVE(SolidBase);
+    //!@}
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * A shape that is hollow, is truncated azimuthally, or both.
```
elliottbiondo, line 110: I would say "interior excluded region" and/or a wedge subtracted.

I also don't understand why it couldn't be neither, though I see that that is how it is coded.

src/orange/orangeinp/Solid.hh
```diff
@@ -0,0 +1,185 @@
...
+    //! Get the user-provided label
+    std::string_view label() const final { return label_; }
+
+    //! Interior convex region interface for construction and access
+    ConvexRegionInterface const& interior() const final { return interior_; }
+
+    // Optional excluded
+    ConvexRegionInterface const* excluded() const final;
+
+    //! Angular restriction to add
```
elliottbiondo: Optional

elliottbiondo (approved):
Since this is the last PR in the queue, could you add a brief design document? Maybe you did in a previous MR after I already approved and didn't see it.

src/orange/orangeinp/Solid.cc
sethrj: This gets a reference to the enclosed angle, which is false if there's no azimuthal truncation.

src/orange/orangeinp/Solid.hh
sethrj: There's got to be a better way to describe it than either one... it's like a thick surface, which I think "hollow" fits pretty well. Excluded inner region sounds much more general than the Solid is capable of...

src/orange/orangeinp/Solid.hh
sethrj: I'm deliberately trying to prevent two `Object` types from being the same representation unless they compare equal... maybe I'll relax that eventually, but I think for now I don't think it's a problem to keep the restriction. Or do you think it's better to have a "Solid" that's the same as a "Shape", and later if we need to we can add a "simplify" method that'll let solids decay to shapes? That *would* remove some logic from the Geant4 translator...

src/orange/orangeinp/Solid.hh
elliottbiondo: I am not sure I understand --- the excluded region must be the same shape as the outer region? If so hollow is fine, but you should clarify that this is what you mean.

src/orange/orangeinp/Solid.cc
elliottbiondo: Could you add a comment for that?

src/orange/orangeinp/Solid.hh
elliottbiondo: So if a geometry consists of only simple cuboids for example (with no internal regions or wedges) then solids are never created? In a sense it would be nice to have geometry elements go through the same levels of abstraction, but it might not be worth it if that adds a bunch more code that is essentially a no-op. To weight in further I think need so study the forthcoming design diagram/document, because I am not clear on what exactly happens on each level of abstraction.

src/orange/orangeinp/Solid.cc
```diff
@@ -0,0 +1,169 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a wedge shape to intersect (inside) or subtract (outside).
+ */
+auto SolidEnclosedAngle::make_wedge() const -> SenseWedge
+{
+    CELER_EXPECT(*this);
+    // Get the start value between [0, 1)
+    real_type start = eumod(start_.value(), real_type{1});
```
elliottbiondo, line 49: Same comment as before on eumod

src/orange/orangeinp/Solid.hh
sethrj: just `Shape`s, not `Solid`. The abstraction is that they're both `Object` interfaces. So in Geant4 a "cone solid" could produce either a `ConeShape` in its simplest form or a `ConeSolid` if it's truncated.

src/orange/orangeinp/Solid.hh
sethrj: Yeah, exactly.

src/orange/orangeinp/Solid.hh
elliottbiondo: Okay I see. The nomenclature here is difficult especially since Geant4 nomenclature clashes with ORANGE nomenclature (recognizing that may be unavoidable).

src/orange/orangeinp/Solid.hh
sethrj: Rotated and negated?

elliottbiondo: Disregard this comment, I misunderstood

sethrj:
@elliottbiondo take a look at `doc/main/api/orange.rst`... that part describes the user-facing parts and some of the construction logic. I've got most of the basic "proto" construction written which is the final piece of the puzzle... perhaps we could do a detailed review then?

src/orange/orangeinp/Solid.hh
sethrj: Well, I've tried to delineate it a bit: "shape" is going to be for the simpler objects, "solids" are the weird Geant4 ones. See https://github.com/celeritas-project/celeritas/issues/1078#issue-2064118326

PR #888, opened 2023-08-11
Fix sampling of secondary gamma energies in EPlusGGInteractor

whokion:
This MR fixes a bug in sampling secondary gamma energies in `EPlusGGInteractor`, using the composition and rejection in which BernoulliDistribution(p) is used instead of BernoulliDistribution(1-p) where the rejection function, p is highly skewed, and addresses the difference seen in physics verification with respect to Geant4 reported by @mrguilima.

PR #1138, opened 2024-03-04
Fix Moller-Bhahba energy distribution

amandalund:
This fixes a bug in the rejection loop of the Moller and Bhabha energy distributions. The plots below show the energy deposition in the TestEM3 geometry before and after the fix with 250 MeV electron primaries and only ionization enabled:
testem3-simple-calo-ioni-250MeV
testem3-simple-calo-ioni-250MeV-fix
(This does not, however, eliminate the discrepancy in the energy deposition with all physics enabled.)

### DIFF
--- a/src/celeritas/em/distribution/BhabhaEnergyDistribution.hh
+++ b/src/celeritas/em/distribution/BhabhaEnergyDistribution.hh
@@ -98,7 +98,7 @@ CELER_FUNCTION real_type BhabhaEnergyDistribution::operator()(Engine& rng)
         epsilon = 1 / sample_inverse_epsilon(rng);
         g_numerator = this->calc_g_fraction(epsilon, epsilon);
 
-    } while (BernoulliDistribution(g_numerator / g_denominator)(rng));
+    } while (!BernoulliDistribution(g_numerator / g_denominator)(rng));
 
     return epsilon;
 }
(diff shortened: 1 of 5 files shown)


stognini:
Thanks @amandalund for fixing that. Looks great.

Got a list of errors (all the same, from lines `192` to `208`) on wildstyle while building it, which come from a recent PR:
```powershell
celeritas/src/orange/orangeinp/ConvexSurfaceBuilder.cc:192:39: error: extra ‘;’ [-Werror=pedantic]
 CSB_INSTANTIATE(PlaneAligned<Axis::x>);
```

Aside from that trivial fix, all the tests pass here.

whokion (approved):
Excellent! 

- This bug is the similar one that was fixed in EPlusGG model for using 1-p vs. p.  Probably it might be very difficult to catch as changing the rejection condition gives a little difference (just curious why? probably the reject envelop is very flat in p and close to the target probability distribution?)    
- As the longitudinal energy distribution (along z, layer number) is now an excellent agreement to that of Geant4, just curious how the lateral distribution (either x or y, or any r direction) looks like if it is easy to compare between celeritas and Geant4.  I expect that they should agree very well, but still give a good reference for the lateral energy distribution when MSC is added later - of course, they are decoupled, shoul be verified independently, but may give us a good base line when we interpret the lateral distribution with all processes are tuned on.

sethrj (approved):
Great catch @amandalund ! As Soon said , this is very similar to #888. I wonder if instead of `BernoulliDistribution`—which is the same signature as C++ `<random>` but not obvious—we should define `RejectionSampler`? Since a lot of times our `} while (...);` are formulated as rejections rather than acceptance.

The other thing that's striking is that the number of sampled RNGs in the broken rejection case was *way* higher before the fix. So maybe we should extend our sampling distribution tests and check the "expected samples" count to see if any others are unusually high.

sethrj:
See https://github.com/celeritas-project/celeritas/pull/1354

PR #1140, opened 2024-03-04
Support safety and reentrant boundaries for multi-universe ORANGE

elliottbiondo:
This MR adds support for reentrant cross_boundary() calls when a particle resides on boundary between geometric universes. In addition, safety distance calculations are updated to find the minimum safety distance over multiple levels. A TestEM3 unit test is added, using universes. This version of the this test problem supports simple safety. Finally, a bug in UnitInserter is fixed, in which volume flags were being erroneously unset.


src/orange/OrangeTrackView.hh
```diff
@@ -922,18 +915,28 @@ OrangeTrackView::find_next_step_impl(detail::Intersection isect)
...
+
+    real_type min_safety_dist = numeric_limits<real_type>::infinity();
+
+    for (auto lev : range(LevelId{this->level() + 1}))
+    {
+        auto lsa = this->make_lsa(lev);
+        auto sd = visit_tracker(
+            [&lsa](auto&& t) { return t.safety(lsa.pos(), lsa.vol()); },
+            lsa.universe());
+        min_safety_dist = std::min(min_safety_dist, sd);
```
sethrj: You can use `celeritas::min` (or omit the namespace), which uses `std::fmin` for floating point to implement the function as a single instruction rather than a branch + assignment.

src/orange/detail/UnitInserter.cc
```diff
@@ -285,8 +285,11 @@ UniverseId UnitInserter::operator()(UnitInput const& inp)
     {
         unit.background = LocalVolumeId(inp.volumes.size() - 1);
     }
+
+    // Simple safety if all volumes provide support, excluding the external
+    // volume, which appears first in the list
     unit.simple_safety = std::all_of(
-        vol_records.begin(), vol_records.end(), [](VolumeRecord const& v) {
+        vol_records.begin() + 1, vol_records.end(), [](VolumeRecord const& v) {
```
elliottbiondo, line 293: This is above-board if we guarantee that external is first, but I am not sure if that's a good idea.

elliottbiondo:
@sethrj you can remove draft and begin your review

src/orange/detail/UnitInserter.cc
sethrj: We do guarantee it by construction: see `OrangeTrackView::is_outside`. Maybe we should define: in OrangeData:
```c++
static inline constexpr LocalVolumeId local_orange_outside_volume{0};
```
and in your code here add:
```c++
static_assert(local_orange_outside_volume == LocalVolumeId{0});
```
and also use it in orange track data:
```c++
CELER_FUNCTION bool OrangeTrackView::is_outside() const
{
    // Zeroth volume in outermost universe is always the exterior by
    // construction in ORANGE
    auto lsa = this->make_lsa(LevelId{0});
    return lsa.vol() == local_orange_outside_volume;
}
```

test/orange/data/testem3.org.omn
```diff
@@ -0,0 +1,238 @@
...
+[UNIVERSE][CELL absorber]
+comp "Pb"
+shapes -abs_shape
+
+[UNIVERSE=general outer]
+interior "full"
+
+!#### LEVEL 0 UNIVERSE ####!
+
+[UNIVERSE][SHAPE=cuboid full]
```
sethrj: To be consistent with the gdml and "flat" geometry, could you rename this `calorimeter`, and add a cell called `world` that's outside it and inside a box with halfwidth 24?
```py
    {
        '_type': "box",
        'widths': [48, 48, 48],
        'name': "world",
    }
```

test/orange/Orange.test.cc
```diff
@@ -1081,6 +1122,18 @@ TEST_F(HexArrayTest, track_out)
     EXPECT_VEC_CLOSE(d2b, refd2b, real_type(1e-5), real_type(1e-5));
 }
 
+// Change direction on a universe boundary to reenter the cell
```
sethrj: Copy-paste error

test/orange/Orange.test.cc
```diff
@@ -1081,6 +1122,18 @@ TEST_F(HexArrayTest, track_out)
...
+TEST_F(TestEM3Test, safety)
+{
+    EXPECT_TRUE(this->params().supports_safety());
+
+    auto geo = this->make_track_view();
+
+    // Initialize in innermost universe
+    geo = Initializer_t{{19.99, 19.9, 19.9}, {0, 1, 0}};
+    EXPECT_SOFT_EQ(0.01, geo.find_safety());
```
sethrj, line 1134: Just for thoroughness, could you add an additional test point where the universe-1 boundary (`x=19.43`) is closer?

src/orange/OrangeTrackView.hh
```diff
@@ -705,17 +699,16 @@ CELER_FUNCTION void OrangeTrackView::set_dir(Real3 const& newdir)
...
+        // Changing direction on a boundary, which may result in not leaving
+        // current volume upon the cross_surface call
+        auto lsa = this->make_lsa(this->surface_level());
 
         TrackerVisitor visit_tracker{params_};
-        auto pos = this->pos();
         auto normal = visit_tracker(
-            [&pos, local_surface = this->surf()](auto&& t) {
+            [pos = this->pos(), local_surface = this->surf()](auto&& t) {
                 return t.normal(pos, local_surface);
```
sethrj, line 709: I think this should be the *local* position not global.

```suggestion
            [&pos = lsa.pos(), local_surface = lsa.surf()](auto&& t) {
```

src/orange/OrangeTrackView.hh
```diff
@@ -705,17 +699,16 @@ CELER_FUNCTION void OrangeTrackView::set_dir(Real3 const& newdir)
...
-        detail::UniverseIndexer ui(params_.universe_indexer_data);
+        // Changing direction on a boundary, which may result in not leaving
+        // current volume upon the cross_surface call
+        auto lsa = this->make_lsa(this->surface_level());
 
         TrackerVisitor visit_tracker{params_};
-        auto pos = this->pos();
         auto normal = visit_tracker(
-            [&pos, local_surface = this->surf()](auto&& t) {
+            [pos = this->pos(), local_surface = this->surf()](auto&& t) {
```
sethrj: I think this should be the *local* position not global.

```suggestion
            [&pos = lsa.pos(), local_surface = this->surf()](auto&& t) {
```

test/orange/data/testem3.org.omn
elliottbiondo: Sure, noting that the world cell will have an internal region, meaning `supports_safety_` will now be false

elliottbiondo: I don't know if its worth keeping the original geom as well in order to test that this returns true for it:
```
    supports_safety_
        = std::all_of(
              host_data.simple_units[AllItems<SimpleUnitRecord>()].begin(),
              host_data.simple_units[AllItems<SimpleUnitRecord>()].end(),
              [](SimpleUnitRecord const& su) { return su.simple_safety; })
          && host_data.rect_arrays.empty();
```

elliottbiondo: Model updated to include world

elliottbiondo:

sethrj (approved):
Nice!

PR #1142, opened 2024-03-06
Add CSG wedge and helper functions

sethrj:
This is in preparation for adding Geant4 "solid"s. The infinite wedge is used to cut out an azimuthal section of a Z-aligned shape: e.g., a cylindrical segment.

### DIFF
--- a/src/orange/orangeinp/ConvexRegion.cc
+++ b/src/orange/orangeinp/ConvexRegion.cc
@@ -297,11 +297,14 @@ void Ellipsoid::output(JsonPimpl* j) const
  * Construct from a starting angle and interior angle.
  */
 InfWedge::InfWedge(Turn start, Turn interior)
-    : start_{Turn{std::fmod(start.value(), real_type{1})}}, interior_{interior}
+    : start_{start}, interior_{interior}
 {
+    CELER_VALIDATE(start_ >= zero_quantity() && start_ < Turn{1},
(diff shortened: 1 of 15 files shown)


src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -290,6 +290,48 @@ void Ellipsoid::output(JsonPimpl* j) const
...
 }
 
+//---------------------------------------------------------------------------//
+// INFWEDGE
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from a starting angle and interior angle.
+ */
+InfWedge::InfWedge(Turn start, Turn interior)
+    : start_{Turn{std::fmod(start.value(), real_type{1})}}, interior_{interior}
```
elliottbiondo: Is there a good reason to properly handle start values greater than 1? It seems like this would indicate a issue upstream.

elliottbiondo (approved):

src/orange/orangeinp/ConvexRegion.cc
sethrj: Good point, the rest of the shapes "validate" that the inputs are correct rather than modifying them. I was thinking of the case where you specify the start angle in `(-1, 0]` but even then `fmod` would let the value remain negative (should've been `eumod`, which will be used in the next PR). Fixed.

PR #1146, opened 2024-03-09
Add "minimal" and "safety plus" Urban MSC step limit algorithms

amandalund:
This add the "minimal" and "safety plus" step limit algorithms for Urban MSC and imports the step limit algorithm selection from Geant4. I haven't implemented the final algorithm ("distance to boundary") since it doesn't seem like we have any need for it currently and it might require storing some extra state data.


src/celeritas/em/msc/detail/UrbanMscSafetyStepLimit.hh
```diff
@@ -143,21 +146,40 @@ UrbanMscStepLimit::UrbanMscStepLimit(UrbanMscRef const& shared,
                                 shared_.params.safety_fact * safety);
     }
     limit_ = max<real_type>(limit_, limit_min_);
+
+    if (use_safety_plus)
+    {
+        // Minimum range
+        real_type rho = 1e-3 * units::centimeter;
```
sethrj: I think this (and alpha below as well?) should be constexpr parameters in the data

src/celeritas/em/msc/detail/UrbanMscMinimalStepLimit.hh
```diff
@@ -0,0 +1,136 @@
...
+
+#include <cmath>
+
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+#include "corecel/math/Algorithms.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/Types.hh"
+#include "celeritas/em/data/UrbanMscData.hh"
+#include "celeritas/phys/Interaction.hh"
```
whokion:
```suggestion
```

src/celeritas/em/msc/detail/UrbanMscSafetyStepLimit.hh
```diff
@@ -143,21 +146,40 @@ UrbanMscStepLimit::UrbanMscStepLimit(UrbanMscRef const& shared,
                                 shared_.params.safety_fact * safety);
     }
     limit_ = max<real_type>(limit_, limit_min_);
+
+    if (use_safety_plus)
+    {
+        // Minimum range
```
whokion: Just for a note: "Minimum range" is somewhat misleading as it is the minimum to approximate this step function approach and the step limit is equal to `range`.  Probably "Minimum range for an empirical step-function approach".  In some sense, these `rho` and `alpha` should be configurable too (as similar in the energy loss).  As they are fixed default values here, make them static constexpr(?).

src/celeritas/em/msc/detail/UrbanMscMinimalStepLimit.hh
```diff
@@ -0,0 +1,136 @@
...
+    }
+    limit_ = msc_range.range_init;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the true path length using the Urban multiple scattering model.
+ */
+template<class Engine>
+CELER_FUNCTION real_type UrbanMscMinimalStepLimit::operator()(Engine& rng)
```
whokion, line 111: So, this operator is same as the one for the Safety/Plus (and may be same for the distance_to_boundary algorithm, if added).  Can add a common until function to share the code ?

test/celeritas/em/UrbanMsc.test.cc
```diff
@@ -383,6 +384,176 @@ TEST_F(UrbanMscTest, step_conversion)
...
+        EXPECT_VEC_SOFT_EQ(expected_range_init, result.range_init);
+        EXPECT_VEC_SOFT_EQ(expected_range_fact, result.range_fact);
+        EXPECT_VEC_SOFT_EQ(expected_limit_min, result.limit_min);
+    }
+    {
+        // "Use safety plus" algorithm
+        static double const expected_mean_step[] = {2.1788745016992e-06,
+                                                    0.00010590181504835,
+                                                    0.003070788811936,
+                                                    0.094489992160893,
```
whokion, line 534: Interesting results!  Just for a note: it would be great that we can verify these results with respect to Geant4 (something like TestEm15) (a sanity check in the model level comparison to build our confidence). Yes, it has been one of my to-do list too - or may we leave such a validation for a summer student^^.

whokion (approved):
Very nice addition/extension!  Thanks.

src/celeritas/em/msc/detail/UrbanMscMinimalStepLimit.hh
amandalund: Right, this is the same for safety(plus) and minimal. For the distance to boundary algorithm it may be slightly different as it looks like there are also checks against the "skin" parameter.

sethrj (approved):

src/celeritas/io/ImportParameters.hh
```diff
@@ -43,6 +44,8 @@ struct ImportEmParameters
     double lowest_electron_energy{0.001};
     //! Whether auger emission should be enabled (valid only for relaxation)
     bool auger{false};
+    //! MSC step limit algorithm
+    MscStepLimitAlgorithm msc_step_algorithm{MscStepLimitAlgorithm::safety};
```
sethrj, line 48: This ended up breaking the SWIG build (see #1159) in part because it's the only place where we use Celeritas types in the Import data structs. (Will this mess with ROOT? @pcanal ). Should we create a separate ImportMscAlgorithm, or ...?

pcanal: It should not break the ROOT/Cling parsing.

PR #1148, opened 2024-03-13
Add proto-universe and CSG Unit construction

sethrj:
This is the next-to-last step for programmatically building an ORANGE geometry from CSG objects. It defines a "proto-universe" interface for defining individual universes (aka "scenes") and building them. For testing and external integration with SCALE, the `UnitProto` class has a helper function for returning a CsgUnit which makes testing a cinch.

The following PR will take the constructed CSG unit, build a UniverseInput, and actually construct an OrangeParams object for tracking.


src/orange/orangeinp/ProtoInterface.hh
```diff
@@ -0,0 +1,71 @@
...
+namespace celeritas
+{
+namespace orangeinp
+{
+class ObjectInterface;
+class GlobalBuilder;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a universe for building ORANGE geometry.
```
elliottbiondo: an ORANGE geometry

src/orange/orangeinp/ProtoInterface.hh
```diff
@@ -0,0 +1,71 @@
...
+
+    //! Get a non-owning set of all daughters referenced by this proto
+    virtual VecProto daughters() const = 0;
+
+    //! Construct a universe input from this object
+    virtual void build(GlobalBuilder&) const = 0;
+
+  protected:
+    //!@{
+    //! Allow construction and assignment only through daughter classes
```
elliottbiondo: I would say "subclasses" to avoid confusion with the other use of the word daughter, i.e daughter universes

src/orange/orangeinp/UnitProto.hh
```diff
@@ -0,0 +1,202 @@
...
+class Transformed;
+
+namespace detail
+{
+struct CsgUnit;
+}  // namespace detail
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a "unit" universe.
```
elliottbiondo: Communicate that this *the* standard CSG universe, not some sort of special case

src/orange/orangeinp/UnitProto.hh
```diff
@@ -0,0 +1,202 @@
...
+ *   Global?    Description
+ *  ========== ==========================================================
+ *   No         Boundary is already truncated by higher-level unit
+ *   Yes        Boundary must explicitly be represented as a volume
+ *  ========== ==========================================================
+ *
+ * These result in different z ordering for the exterior:
+ *
+ *  === === ================== ========================================
+ *   G   P   Result             Description
```
elliottbiondo: Why does "P"  indicate boundary condition? Also "X" is not good for "explicit" both in general (in my opinion) but also explicitly in this case because "x" already means "exterior" in z-order

elliottbiondo: Also, does each row correspond to a z-order? If so, could you include the z-order as a column?

src/orange/orangeinp/UnitProto.cc
```diff
@@ -0,0 +1,180 @@
...
+auto UnitProto::interior() const -> SPConstObject
+{
+    return input_.boundary.interior;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a list of all daughter protos.
+ */
+auto UnitProto::daughters() const -> VecProto
```
elliottbiondo, line 72: Not sure if this is enough work to cross into "calc_daughters"

src/orange/orangeinp/UnitProto.cc
```diff
@@ -0,0 +1,180 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a standalone unit for testing or external interface.
+ *
+ * The universe ID in the resulting "fill" is the index into the input \c
+ * daughters list, which is the same ordering as the array of \c
+ * this->daughters().
+ *
+ * The exterior zorder
```
elliottbiondo: ...

elliottbiondo (approved):

src/orange/orangeinp/UnitProto.cc
sethrj: Maybe... maybe `get_daughters`? I'll revisit this.

PR #1150, opened 2024-03-13
Add neutron inelastic process

whokion:
# Add neutron inelastic process (minimal skeleton)
 - Add neutron inelastic process, model and related classes

# Description
This MR includes minimal skeletons of neutron inelastic processes for an initial development.


whokion:
This is a set of classes for neutron capture and inelastic processes and models (without any final state sampling).  It may require to refactor many codes later as there are somewhat redundant codes in the current structure, but it would be good to keep them separate until all sub-components are implemented and tested - later, we may consolidate codes into a neutron general process, which may be more compact and efficient. Anyway, these codes will serve as a ground for implementing and testing inreractor codes along the way.

sethrj (commented):
@whokion , even though 3500 lines of this is test data, could you please break this up into two pull requests, one for capture and one for inelastic?

whokion:
@sethrj Sorry for this big MR! Even though this MR can be separated by two, it actually follows a fairly parallel structure for capture and inelastic (as this set of classes are only loading cross section data and building the pipeline from process to interactor for capture and inelastic). Since there are no much variations between capture and inelastic without detail features for each interactor, I hope that they should be straightforward to review in parallel).  If it is too much load to review, please take your time.  I promise that this is the last big MRs for neutrons. Otherwise, I will break it by two.  If the size of cross section test data is too big to keep under the test area , I can prune them for minimal tests.

sethrj:
Please @whokion , you know we have a policy to keep pull requests as a small as possible to make them more reviewable. The fact that the two new processes are similar is another good reason to keep them separate. It's very hard to cross reference the multiple files within a single pull request interface.

I would recommend going through your pull request and simply *deleting* all the files/lines that aren't (e.g.) capture. Make sure that it builds, commit the changes, run `git checkout -b neutron-inelastic` and `git revert HEAD`. That gives you a new starting point for the follow-on pull request.

whokion:
No problem. Will break by two. Sorry about not knowing the policy.

sethrj:
Huh, I guess it wasn't officially codified, sorry for that omission. I'll update the contribution guide. It's definitely a best practice in general to have pull requests be as possible while still "substantive and self-contained" . See for example https://github.com/celeritas-project/celeritas/pull/1119 where I broke up a 1400 line PR into several different ones for Elliott.

whokion:
> I would recommend going through your pull request and simply _deleting_ all the files/lines that aren't (e.g.) capture. Make sure that it builds, commit the changes, run `git checkout -b neutron-inelastic` and `git revert HEAD`. That gives you a new starting point for the follow-on pull request.

@sethrj I removed all codes related to the capture and am ready to push the update to the remote (neutron-processes). Just curious, why do I need `git checkout -b neutron-inelastic` and `git revert HEAD`?  So, do I need to create another branch (neutron-inelastic) and push it to the remote? Thanks.

sethrj:
@whokion the recommendation I made was how to create a second "draft" branch that restores the removed code. So if you've deleted all the *capture* code in a single commit, you can create a new branch that restores all that code:
```console
$ git commit # after removing all the capture code
$ git push # makes this PR (#1150) only about inelastic
$ git checkout -b neutron-capture # create a new branch for a future merge request
$ git revert HEAD # restore all the neutron capture code
$ git push -u # push so that you can create a second "draft" PR
$ git checkout neutron-processes # so that you're ready to work on PR 1150 again
```

src/celeritas/io/ImportData.hh
```diff
@@ -58,7 +58,9 @@ struct ImportData
     using ImportSBMap = std::map<ZInt, ImportSBTable>;
     using ImportLivermorePEMap = std::map<ZInt, ImportLivermorePE>;
     using ImportAtomicRelaxationMap = std::map<ZInt, ImportAtomicRelaxation>;
+    using ImportNeutronCaptureMap = std::map<ZInt, ImportPhysicsVector>;
     using ImportNeutronElasticMap = std::map<ZInt, ImportPhysicsVector>;
+    using ImportNeutronInelasticMap = std::map<ZInt, ImportPhysicsVector>;
```
amandalund: We could maybe use a single type alias?
```suggestion
    using ImportPhysicsVectorMap = std::map<ZInt, ImportPhysicsVector>;
```

src/celeritas/io/NeutronXsReader.hh
```diff
@@ -35,18 +48,29 @@ class NeutronXsReader
...
-    explicit NeutronXsReader(char const* path);
+    // Construct the reader from the path to the data directory and the type
+    explicit NeutronXsReader(char const* path, NeutronXsType type);
 
     // Read the data for the given element
     result_type operator()(AtomicNumber atomic_number) const;
 
+  private:
+    // Get the string value for the cross section data type
+
```
amandalund:
```suggestion
```

src/celeritas/neutron/model/NeutronInelasticModel.cc
```diff
@@ -0,0 +1,144 @@
...
+#if !CELER_USE_DEVICE
+void NeutronInelasticModel::execute(CoreParams const&, CoreStateDevice&) const
+{
+    CELER_NOT_CONFIGURED("CUDA OR HIP");
+}
+#endif
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the model ID for this model.
```
amandalund:
```suggestion
 * Get the action ID for this model.
```

test/celeritas/neutron/NeutronInelastic.test.cc
```diff
@@ -0,0 +1,136 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/neutron/NeutronBertini.test.cc
```
amandalund:
```suggestion
//! \file celeritas/neutron/NeutronInelastic.test.cc
```

src/celeritas/neutron/interactor/NeutronInelasticInteractor.hh
```diff
@@ -0,0 +1,93 @@
...
+    CELER_EXPECT(particle.particle_id() == shared_.ids.neutron);
+}
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the final state of the neutron-nucleus inelastic interaction.
+ */
+template<class Engine>
+CELER_FUNCTION Interaction NeutronInelasticInteractor::operator()(Engine&)
+{
+    // Dummy for now
```
amandalund: Instead of having a placeholder dummy `Interactor`, maybe it would be better to postpone adding this and the `Executor` until the inelastic model is implemented and change the `NeutronInelasticModel::execute()` to:
```c++
void NeutronInelasticModel::execute(CoreParams const&, CoreStateDevice&) const
{
    CELER_NOT_IMPLEMENTED("Neutron inelastic interaction");
}
```

amandalund (commented):
Thanks @whokion! Just a few minor comments, this looks good to me!

src/celeritas/io/NeutronXsReader.hh
```diff
@@ -35,18 +48,29 @@ class NeutronXsReader
 
   public:
     // Construct the reader and locate the data using the environment variable
-    NeutronXsReader();
+    explicit NeutronXsReader(NeutronXsType type);
 
-    // Construct the reader from the path to the data directory
-    explicit NeutronXsReader(char const* path);
+    // Construct the reader from the path to the data directory and the type
+    explicit NeutronXsReader(char const* path, NeutronXsType type);
```
sethrj: Now that it has two arguments it shouldn't be explicit.
And since the reader is constructed as though the path is "optional", would it make more sense to have the type argument first?
```suggestion
    NeutronXsReader(NeutronXsType type , char const* path);
```

src/celeritas/neutron/data/NeutronInelasticData.hh
```diff
@@ -0,0 +1,95 @@
...
+    static CELER_CONSTEXPR_FUNCTION units::MevEnergy min_valid_energy()
+    {
+        return units::MevEnergy{1e-7};
+    }
+
+    static CELER_CONSTEXPR_FUNCTION units::MevEnergy max_valid_energy()
+    {
+        return units::MevEnergy{2e+4};
+    }
+
```
sethrj: It would be better to have these as part of a non-templated class. Perhaps the `NeutronInelasticIds` could become `NeutronInelasticScalars`, and these and `neutron_mass` could be moved into that?

src/celeritas/neutron/model/NeutronInelasticModel.cc
```diff
@@ -0,0 +1,144 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct neutron inelastic cross section data for a single element.
+ */
+void NeutronInelasticModel::append_xs(ImportPhysicsVector const& inp,
+                                      HostXsData* data) const
+{
+    auto reals = make_builder(&data->reals);
```
sethrj, line 123: We may want to refactor the construction so that the grids can be deduplicated (if multiple materials use the same energy grid, for example).  See `orange/detail/UnitInserter.hh` and other classes in that directory as an example. We could trivially add a class to be used here (and in other params classes that build grid data):
```c++
class GridInserter
{
  public:
    //!@{
    //! \name Type aliases
    template<class T>
    using Items = Collection<T, Ownership::value, MemSpace::host>;
    using GridId = ItemId<GenericGridData>;
    //!@}

  public:
    GridInserter(Items<GenericGridData>* grids, Items<real_type>* reals);

    GridId operator()(ImportPhysicsVector const&);

  private:
    CollectionBuilder<GenericGridData> grids_;
    DedupeCollectionBuilder<real_type> reals_;
};
```
but one of us could do this in a follow-on PR.

src/celeritas/neutron/xs/NeutronInelasticMicroXsCalculator.hh
```diff
@@ -0,0 +1,58 @@
...
+    using ParamsRef = NeutronInelasticRef;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate neutron inelastic cross sections from NeutronInelasticData.
+ */
+class NeutronInelasticMicroXsCalculator
+    : public NeutronMicroXsCalculator<NeutronInelasticMicroXsCalculator>
+{
```
sethrj, line 26: I don't understand this class design with the combination of inheritance, friends, traits, and CRTP. Wouldn't it be simpler to just template `NeutronMicroXsCalculator` on `ParamsRef`? Or since the structures of all the `ParamsRef` are essentially identical (shared_.micro_xs, shared_.reals), create a new class `NeutronMicroXsData<W,M>` that is used by both inelastic and elastic? Then you wouldn't have to template at all.

src/celeritas/phys/PhysicsData.hh
```diff
@@ -179,6 +180,11 @@ struct HardwiredModels
     ModelId chips;
     NeutronElasticData<W, M> chips_data;
 
+    // Neutron inelastic
+    ProcessId neutron_inelastic;
+    ModelId bertini;
+    NeutronInelasticData<W, M> bertini_data;
+
```
sethrj, line 187: It's definitely time to support models using independent kernels to generate their own cross sections... this is growing out of hand 😓

src/celeritas/neutron/model/NeutronInelasticModel.cc
sethrj: In fact I'm working on this right now.

src/celeritas/io/ImportData.hh
whokion: You are right. Changed as suggested.

src/celeritas/io/NeutronXsReader.hh
whokion: Removed.

src/celeritas/neutron/model/NeutronInelasticModel.cc
whokion: Changed as suggested.

test/celeritas/neutron/NeutronInelastic.test.cc
whokion: Fixed.

src/celeritas/neutron/interactor/NeutronInelasticInteractor.hh
whokion: Okay.  Changed as suggested.

src/celeritas/io/NeutronXsReader.hh
whokion: Changed as suggested.

src/celeritas/neutron/data/NeutronInelasticData.hh
whokion: Rename NeutronInelasticIds to NeutronInelasticScalars and move other scalar data to there.

src/celeritas/neutron/model/NeutronInelasticModel.cc
whokion: Thanks for another PR, [#1157](https://github.com/celeritas-project/celeritas/pull/1157)  on this!

src/celeritas/neutron/xs/NeutronInelasticMicroXsCalculator.hh
whokion: @sethrj You are right that `friend` is not necessary for this case (as the derives do not need to access any member of the mother in this unusual case).  The motivation was to not to use a templated (macro) of tempated (micro) parameter pass, but still to share a common method of the microscopic xs calculation (top on the already templated macroscopic  xs calculator) for different neutron processes.  I will think about the alternative and try to come up with a different solution if you do not like this.

sethrj: Is there ever going to be a difference between the calculation method for the different neutron micro XS? So far they're all "generic grid" interpolation... I think it would be much more sensible to reuse a single class and data type... maybe we can discuss in person tomorrow?

sethrj:
@whokion please merge and push, rather than force-pushing. It looks like you also have pushed some conflict markers.

whokion:
@sethrj Sorry about that.  Somehow I also messed up my local branch while I rebased it to the develop and tried to resolve conflicts.  As there were too many update since this was opened, I would like to close this MR and reopen it later.

sethrj:
OK. If you need a hand with `git reflog`, I can help restore your history (as long as it was `git commit`ted). It looks like d51d342 was the last commit you had before overwriting.

PR #1151, opened 2024-03-14
Import scintillation particle data

stognini:
This PR imports scintillation data for multiple particle types from Geant4 as a part of #886 .

Important caveat: `ScintillationParams` and `ScintillationData` are not updated to correctly use the imported data, this will be done in a follow-up PR.


stognini:
Interesting... the CI errors are all the same
```
ROOT: error: TList::Clear: A list is accessing an object (0xN) already deleted (list name = TList)
```
But the tests pass on my local build. I'll try to figure this out.

pcanal:
Which version of ROOT is being used for the failing run? i.e. are we hitting https://github.com/root-project/root/issues/14793 ?

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -34,17 +40,57 @@ struct ImportScintComponent
...
+ *
+ * The yield vector is the only necessary element, needed to calculate the
+ * yield based on the particle energy during the stepping loop.
+ * Components may not be assigned---they are the equivalent of
+ * \c k[Particle]ScintillationYield[i] in Geant4.
+ */
+struct ImportParticleScintSpectrum
+{
+    ImportPhysicsVector yield_vector;  //!< Particle yield vector
+    std::vector<ImportScintComponent> components;  //!< Fast/slow components
```
amandalund: Are particle-dependent `lambda_mean`, `lambda_sigma`, and `rise_time` used anywhere in scintillation? As far as I can tell only the `yield` and `fall_time` are needed.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -192,6 +192,72 @@ struct MatPropGetter
     }
 };
 
+//---------------------------------------------------------------------------//
+//! Map particles defined in \c G4MaterialConstPropertyIndex .
+std::map<std::string, PDGNumber> optical_particles_map()
+{
+    return {{"PROTON", pdg::proton()},
+            {"DEUTERON", pdg::deuteron()},
```
amandalund: Are these PDGs used anywhere?

amandalund (commented):
Thanks @stognini! Could you add the new properties to test/geocel/data/lar-sphere.gdml and test that they are imported properly in the `GeantImporter` test?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -192,6 +192,72 @@ struct MatPropGetter
     }
 };
 
+//---------------------------------------------------------------------------//
+//! Map particles defined in \c G4MaterialConstPropertyIndex .
+std::map<std::string, PDGNumber> optical_particles_map()
+{
+    return {{"PROTON", pdg::proton()},
```
sethrj: This should be a `static const` so it's created once and reused.

```suggestion
auto& optical_particles_map()
{
    static std::unordered_map<std::string, PDGNumber> const {{"PROTON", pdg::proton()},
```

src/celeritas/ext/GeantImporter.cc
```diff
@@ -192,6 +192,72 @@ struct MatPropGetter
...
+ */
+std::vector<ImportScintComponent>
+fill_vec_import_scint_comp(MatPropGetter& get_property,
+                           std::string particle_name = "")
+{
+    if (!particle_name.empty())
+    {
+        auto const map = optical_particles_map();
+        CELER_EXPECT(map.find(particle_name) != map.end());
+    }
```
sethrj: You're making a copy of the map here.
```suggestion
    CELER_EXPECT(particle_name.empty() || optical_particles_map().count(particle_name));
```

src/celeritas/io/ImportOpticalMaterial.hh
stognini: You are correct. But this ends up being a similar case of the material. The lambda mean and sigma are also not in the Geant4 material property index enums. This way I could use a single object type to describe any spectrum, be it from a specific particle, or a material spectrum, which just tries to be a catch-all spectrum.

src/celeritas/ext/GeantImporter.cc
stognini: Yes. Even though they are stored as `int`, we use those to construct a `map<pdg, spectrum>` for import. Further down the road (although not in this PR) these pdgs are then used in the `ScintillationParams` to store particle spectrum according to particle id in Celeritas.

EDIT: To clarify where they are used, see `ImportOpticalMaterial::ImportScintData`.

stognini:
> Could you add the new properties to test/geocel/data/lar-sphere.gdml and test that they are imported properly in the GeantImporter test

Good point! Will do.

src/celeritas/io/ImportOpticalMaterial.hh
amandalund: Not all the material-dependent components are in the Geant4 material properties table, but we do use them all in our scintillation generator. It seems like some of these particle-dependent components won't be used at all anywhere in the code.

stognini: @whokion 's implementation is a bit different than the G4 one, so he could confirm if my thought process here is correct.

In the same way we have a fully defined scintillation spectrum for a given material, and we use all those values to sample our optical photons in the `ScintillationGenerator`, we'd have a fully defined spectrum for each particle type. Then, if we are sampling using a per-particle spectrum, the generator would loop over the components for that specific particle, and perform the same sampling, generating all the photons produced by that particle.

So yeah, the `ScintillationGenerator` will be updated in a future PR to encompass this case. At that point, it will be able to sample the data based on a user-defined input, something like `use_material_spectrum{true/false};` as a part of our simulation input. And that will tell the generator if it should sample using material or particle data.

stognini:
> Which version of ROOT is being used for the failing run? i.e. are we hitting https://github.com/root-project/root/issues/14793 ?


Oh, that may be it. My local ROOT is `6.31/01`, which might be newer than the CI's one.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -404,7 +470,10 @@ ImportData::ImportOpticalMap import_optical()
     auto num_materials = pct.GetTableSize();
     CELER_ASSERT(num_materials > 0);
 
+    auto const particle_map = optical_particles_map();
```
sethrj:
```suggestion
    auto const& particle_map = optical_particles_map();
```
You're making a copy again :)

src/celeritas/ext/GeantImporter.cc
```diff
@@ -429,38 +498,28 @@ ImportData::ImportOpticalMap import_optical()
...
-                                ImportUnits::time);
-            if (comp)
+            optical.scintillation.material.components
+                = fill_vec_import_scint_comp(get_property);
+
+            // Particle scintillation properties
+            for (auto const& iter : particle_map)
             {
-                optical.scintillation.components.push_back(comp);
+                std::string particle_name = iter.first;
```
sethrj:
```suggestion
                std::string const& particle_name = iter.first;
```

src/celeritas/ext/GeantImporter.cc
```diff
@@ -192,6 +192,71 @@ struct MatPropGetter
     }
 };
 
+//---------------------------------------------------------------------------//
+//! Map particles defined in \c G4MaterialConstPropertyIndex .
+auto& optical_particles_map()
+{
+    static std::unordered_map<std::string, PDGNumber> map
```
sethrj: You forgot the const
```suggestion
    static std::unordered_map<std::string, PDGNumber> const  map
```

sethrj (approved):
Looks good! Going to wait for @whokion and @amandalund to weigh in.

stognini:
Thanks @sethrj , now working on @amandalund 's request for the `GeantImporter` test.

whokion:
@stognini @sethrj sorry for delayed the review.  Please allow me one more day for reviewing this MR.  Thanks.

sethrj:
Of course! Take your time, it's hardly a day old.

src/celeritas/ext/GeantImporter.cc
amandalund: Hmm I'm still not seeing where that's done... I see the map keys being used, but not the values anywhere. I'm also not seeing where the `particles` map in `ImportScintData` is saved... it looks like the particle scintillation spectra are created but never actually stored in the imported data?

src/celeritas/io/ImportOpticalMaterial.hh
amandalund: Oh ok, so is this because we want to be able to sample the optical photon energies from different distributions for different particle types (different from Geant4 where the energy is sampled from the same distribution for all particles)?. Right now the difference between our implementations is that Geant4 samples from a tabulated distribution while we use the distribution parameters (`lambda_mean`, `lambda_sigma`) to sample from a Gaussian.

whokion:
```suggestion
    std::vector<ImportScintComponent> components;  //!< scintillation components
```
Now, `Fast/slow` are somewhat misleading as we support an arbitrary number of components.  May be just scintillation components even though it helps nothing here.

Regarding to scintillation components for each particle type, we need, in principle, extended data with spectra[part][material_Id] compared to the current spectra[material] in `ScintillationData`- so what @stognini is structuring is a right approach.  However, I am really doubt whether there are noticeable differences in scintillation yields or models per particle type for the given same amount of energy loss  over a step - I know that the scintillation yield depends on presence and strength of an electric-field as a secondary factor besides the energy loss, but hardly imagine that the excitation/deexcitation (scintillation) pattern depends on the particle type - after all, those optical properties are need to be measured before setting up a meaning simulation.  Furthermore, as far as I know, it is not well-modeled what fractions of the energy loss actually contribute to the scintillation yield and the ionization separately.  Anyway, simulation features are bounded to user's needs/requirements and Geant4 should have a good practical example to support optical scintillation by the particle type (which I know none, and would like to know what/why HEP experiments require such a subtle optical photon simulation).

whokion (approved):
One general comment is that we may totally reconsider/skip this importing process from the Geant4 material property table, but support material properties with a celeritas-own-style instead, which requires a short cut to build `OpticalParams` of material properties directly from gdml/geometry or some kind of celeritas optical tag linked with input materials. This sounds like more complicated, but rather simpler than we may think as there will be just one optical radiator (i.e, material) in most application cases in HEP (for examples, for LZ/Xenon, LAr-TPC experiments/LAr, Calvision/crystal calorimeter, etc.), or at most two when there are core/outer optical volumes as in a typical detector concept for neutrino/dark matter search experiments.  Anyway, good progress!

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -34,17 +40,57 @@ struct ImportScintComponent
...
+    //! Whether all data are assigned and valid
+    explicit operator bool() const { return yield > 0 && !components.empty(); }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Store per-particle material scintillation spectrum information.
+ *
+ * The yield vector is the only necessary element, needed to calculate the
+ * yield based on the particle energy during the stepping loop.
```
whokion:
```suggestion
 * yield based on the particle energy `loss` during the stepping loop.
```

src/celeritas/io/ImportOpticalMaterial.hh
stognini: > Now, Fast/slow are somewhat misleading

Fair. I'll update the comment.

> extended data with spectra[part][material_Id] compared to the current spectra[material] in ScintillationData

Yes, this is pretty much ready and it will be the next PR. Just waiting for this one to be merged in.

> However, I am really doubt whether there are noticeable differences in scintillation yields or models per particle type for the given same amount of energy loss over a step

This is a point that I thought too, but I settled with: if Geant4 had different yields for a handful of particles, they must be non-negligible and useful to someone, otherwise it would have been a moot point implementing them. And I would guess a scintillation caused by a neutron via strong force is different from a scintillation caused by a muon, which is a MIP.

src/celeritas/ext/GeantImporter.cc
stognini: Ohh, thanks! Great catch. I mistakenly deleted the `insert()` call to  `ImportScintData::particles` while moving code around.

src/celeritas/io/ImportOpticalMaterial.hh
whokion: @stognini The question is whether one (who) really need to treat scintillation differently for different particles (major ones are likely electron vs. hadron) at the same application - i.e., the yield is not just a function of the energy loss, but also the particle type (for example, the energy loss by protons is treated differently from the same amount of energy loss by electrons). The neutron (neutral) should not be a part of the picture and photons produced by the deexcitation by the neutron-nucleus should be considered regular photons, not optical photons. Regarding to MIP, the real question is whether number of optical photons produced by the same amount of energy loss (per unit length) between muons and hadrons are different (does not matter whether MIP by muons or hadrons)?  Maybe yes, but how much, and who cares about their differences.  Why does one need such a fine detail simulation for what application?  As far as I know, Geant4 introduced this particle dependency recently, but the motivation is not clear, especially for HEP (in which the timing is more important than the number of photons collected in a photo-detector, i.e, intensity which can be always amplified by a photo-multiplier).

stognini:
Okay @amandalund , I fixed the missing `insert()` you found (again, great catch, thanks), added particle data to the `lar-sphere.gdml`, and expanded the `GeantImporter` test.

amandalund (approved):
Great, thanks @stognini!

sethrj:
@stognini It looks like the only remaining issue is with the ROOT compatibility. I think either of us could use the build on wildstyle to replace the ROOT data with a backward compatible version?

Also, can you resolve all the conversations that you've addressed? Thanks!

stognini:
>  I think either of us could use the build on wildstyle to replace the ROOT data with a backward compatible version

Oh, good point. I was building an older ROOT locally, but using wildstyle is a good idea. On it.

PR #1149, opened 2024-03-13
Add spack-based workflow to add all supported Geant4 versions to CI

sethrj:
This adds a new github action for testing arbitrary combinations of dependencies, starting with all Geant4 versions. It uses a spack buildcache hosted at https://github.com/celeritas-project/spack-gha-buildcache to speed up builds. Because of build/configure/example errors, basically everything but 11.0 is failing at the moment, so I am labeling the jobs as "experimental" (allowed to fail) and will fix the relevant code in a follow-on PR (#1152).

- [x] Needs https://github.com/spack/spack/pull/43212 to be merged so that the older Geant4 builds will compile.
- [x] Wait for https://github.com/celeritas-project/spack-gha-buildcache/actions/runs/8300993382/job/22720122879 to complete before pushing updated package

PR #1152, opened 2024-03-15
Fix Geant4 build from 10.5–10.7 and example build for 11.1 onward

sethrj:
- Const correctness issues caused some breakage with Geant4 10.5-10.7
- G4 run manager factory was required for examples but is only present in 11.0 and higher
- Issues with the new geant4 cmake variable cache break xercesc importing for 11.1–11.2

**TODO**: wait for #1149 to merge, then remove the 'experimental' status on the failing builds


drbenmorgan (approved):
All looks good @sethrj, LGTM!

PR #1153, opened 2024-03-17
Add particle scintillation data to `ScintillationParams`; Add `ScintillationPreGenerator`

stognini:
This PR is tied to issue #886 and is a follow-up from PR #1151 , focusing on 2 new features:

- Expand `ScintillationData` to include particle- and material-dependent scintillation data, updating `ScintillationParams` accordingly.
- Add a Scintillation pre-generator class to populate an `OpticalDistributionData` object based on step, params, and state information.

# Caveats
To avoid making this PR too large, we can only sample optical photons from scintillation based on material information. Therefore, follow up work will have to expand the current scintillation physics to:
- Add sampling of scintillation optical photons (in `ScintillationGenerator`) based on particle type.
- Sample the number of photons (in `ScintillationPreGenerator`) also based on the particle type.

Beyond the sampling per particle type, the importing of `G4OpticalParameters` will be necessary. For now, there is just a placeholder in `ImportParameters.hh`.

All these tasks marked with `// TODO:` comments.


src/celeritas/Types.hh
```diff
@@ -66,6 +66,12 @@ using IsotopeComponentId = OpaqueId<struct ElIsotopeComponent>;
 //! Opaque index to a material with optical properties
 using OpticalMaterialId = OpaqueId<struct OpticalMaterial_>;
 
+//! Opaque index to a scintillation particle id
+using ScintillationParticleId = OpaqueId<struct ScintillationParticle_>;
+
+//! Opaque index to a scintillation spectrum
+using ParticleScintSpectrumId = OpaqueId<struct ParticleScintillationSpectrum>;
+
```
sethrj: These are scintillation-specific, so let's move these to ScintillationData. Maybe we even need an `optical/Types.hh`.

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -92,7 +92,8 @@ struct ImportScintData
     //! Whether all data are assigned and valid
     explicit operator bool() const
     {
-        return static_cast<bool>(material) && resolution_scale >= 0;
+        return (static_cast<bool>(material) || !particles.empty())
```
sethrj, line 104: Add a note to the class docs discussing the "either particles or materials". Do we also want to prohibit having both particles *and* materials? Then it shoudl be
```suggestion
        return (static_cast<bool>(material) == particles.empty())
```

src/celeritas/optical/ScintillationData.hh
```diff
@@ -11,16 +11,21 @@
...
  * Material dependent scintillation property.
+ *
+ * Components represent different scintillation emissions, such as
+ * prompt/fast, intermediate, and slow. They can be material-only or depend on
+ * the incident particle type.
  */
 struct ScintillationComponent
 {
-    real_type yield_prob{};  //!< Probability of the yield
+    real_type yield_frac{};  //!< Fraction of total yield (yield/sum(yields))
```
sethrj, line 29: Isn't the probability equal to the fraction? Should we just call this `yield`?

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,67 +34,154 @@ struct ScintillationComponent
...
+
+    //! Material-only scintillation spectrum data
+    MaterialItems materials;  //!< [OpticalMaterialId]
+
+    //! Particle and material scintillation spectrum data
+    ParticleItems particles;  //!< [ParticleScintSpectrumId]
+
+    //! Backend storage for ParticleScintillationSpectrum::yield_vector
+    Items<real_type> grid_data;
+
```
sethrj: To be consistent with the other classes, maybe just name this `reals`? Also "group" it with `ScintillationSpectrum` under a comment `// Backend data` since these are the "common" blocks that are only meaningful when accessed by other classes.

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,67 +34,154 @@ struct ScintillationComponent
...
+    //! Components for either material or particle items
     Items<ScintillationComponent> components;
-    OpticalMaterialItems<ScintillationSpectrum> spectra;
+
+    size_type num_opt_materials{};
+    size_type num_opt_particles{};
```
sethrj: I think these are redundant with the size of `materials` and `particles`.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,225 @@ ScintillationParams::from_import(ImportData const& data)
...
+
+    // Fill with uninitialized data
+    // The final arrays will be mostly uninitialized, and that will determine
+    // which MaterialIds and ParticleIds are scintillation particles.
+    std::fill(input.matid_to_optmatid.begin(),
+              input.matid_to_optmatid.end(),
+              OpticalMaterialId{});
+    std::fill(input.pid_to_scintpid.begin(),
+              input.pid_to_scintpid.end(),
+              ScintillationParticleId{});
```
sethrj: `resize` default-initializes the values, so this isn't necessary.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,225 @@ ScintillationParams::from_import(ImportData const& data)
...
+        }
+
+        // Loop over particles and add all that are scintillation particles
+        auto const& iomsp = iom.scintillation.particles;
+        for (auto pid : range(data.particles.size()))
+        {
+            if (iomsp.find(data.particles[pid].pdg) != iomsp.end())
+            {
+                // Add new ScintillationParticleId
+                input.pid_to_scintpid[pid] = scintpid++;
```
sethrj: Do you need to save the number of scintillation particles?

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,225 @@ ScintillationParams::from_import(ImportData const& data)
...
+                auto iter = input.data[matid].particles.find(pdg);
+                CELER_ASSERT(iter != input.data[matid].particles.end()
+                             && pdg == iter->first);
+
+                auto const& ipss = iter->second;
+                ParticleScintillationSpectrum part_spec;
+                part_spec.yield_vector.grid = build_grid.insert_back(
+                    ipss.yield_vector.x.begin(), ipss.yield_vector.x.end());
+                part_spec.yield_vector.value = build_grid.insert_back(
+                    ipss.yield_vector.y.begin(), ipss.yield_vector.y.end());
```
sethrj: You should replace this with the new `GenericGridBuilder`.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,225 @@ ScintillationParams::from_import(ImportData const& data)
...
 
     mirror_ = CollectionMirror<ScintillationData>{std::move(host_data)};
     CELER_ENSURE(mirror_);
 }
 
+//---------------------------------------------------------------------------//
+/*
+ * Return a Celeritas ScintillationComponent from an imported one.
+ */
+std::vector<ScintillationComponent> ScintillationParams::copy_components(
```
sethrj: I'd rename this `build_components` or make it a helper class to insert them. It's not merely copying them, it's also normalizing and validating.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,225 @@ ScintillationParams::from_import(ImportData const& data)
...
+        norm += input_comp[i].yield;
+    }
+
+    // Store normalized yield
+    for (auto i : range(comp.size()))
+    {
+        comp[i].yield_frac = input_comp[i].yield / norm;
+    }
+
+    this->validate(comp);
```
sethrj: Why'd this move to a separate function?

src/celeritas/optical/ScintillationParams.hh
```diff
@@ -27,22 +28,37 @@ class ScintillationParams final : public ParamsDataInterface<ScintillationData>
...
   public:
-    // Construct with imported data
+    // Construct with imported data and particle params
     static std::shared_ptr<ScintillationParams>
-    from_import(ImportData const& data);
+    from_import(ImportData const& data, SPConstParticles particle_params);
 
     // Construct with scintillation components
-    explicit ScintillationParams(Input const& input);
+    explicit ScintillationParams(Input const& input,
```
sethrj: With two arguments it doesn't need to be explicit.

src/celeritas/optical/ScintillationParams.hh
```diff
@@ -27,22 +28,37 @@ class ScintillationParams final : public ParamsDataInterface<ScintillationData>
...
+        VecOptMatId matid_to_optmatid;  //!< MaterialId to OpticalMaterialId
+        VecSPId pid_to_scintpid;  //!< ParticleId to ScintillationParticleId
+        std::vector<ImportScintData> data;  //!< Indexed by OpticalMaterialId
+        bool scintillation_by_particle;  //!< Particle or material sampling
+
+        //! Whether all data are assigned and valid
+        explicit operator bool() const
+        {
+            return !data.empty()
+                   && (!matid_to_optmatid.empty() || !pid_to_scintpid.empty());
```
sethrj: Same thing about one or the other: it shouldn't be both?

src/celeritas/optical/ScintillationParams.hh
```diff
@@ -27,22 +28,37 @@ class ScintillationParams final : public ParamsDataInterface<ScintillationData>
...
     struct Input
     {
-        std::vector<ImportScintData> data;
+        using VecOptMatId = std::vector<OpticalMaterialId>;
+        using VecSPId = std::vector<ScintillationParticleId>;
+
+        VecOptMatId matid_to_optmatid;  //!< MaterialId to OpticalMaterialId
+        VecSPId pid_to_scintpid;  //!< ParticleId to ScintillationParticleId
+        std::vector<ImportScintData> data;  //!< Indexed by OpticalMaterialId
+        bool scintillation_by_particle;  //!< Particle or material sampling
```
sethrj: isn't this redundant with whether matid_to_optmatid or pid_to_scintpid is not empty? It's also missing a default initializer.

src/celeritas/optical/ScintillationPreGenerator.hh
```diff
@@ -0,0 +1,143 @@
...
+{
+    CELER_EXPECT(step_len_ > 0);
+    CELER_EXPECT(mat_id_);
+    CELER_EXPECT(shared_);
+    CELER_EXPECT(step_);
+
+    if (shared_.scintillation_by_particle())
+    {
+        // TODO: implement sampling for particles
+        CELER_NOT_IMPLEMENTED("scintillation by particle type");
```
sethrj: the "not implemented" should be already handled by the earlier input validation, and this function is meant mainly for host code. Maybe `CELER_ASSERT_UNREACHABLE()` with a comment...

src/celeritas/optical/ScintillationPreGenerator.hh
```diff
@@ -0,0 +1,143 @@
...
+ * Return an \c OpticalDistributionData object. If no photons are sampled, an
+ * empty object is returned and can be verified via its own operator bool.
+ */
+template<class Generator>
+inline CELER_FUNCTION OpticalDistributionData
+ScintillationPreGenerator::operator()(Generator& rng)
+{
+    // Material-only sampling
+    auto const& material = shared_.materials[mat_id_];
+    CELER_EXPECT(material);
```
sethrj: We should be asserting that the material is defined whenever we build and copy, not every time we sample. Maybe remove this, or at least it should be `CELER_ASSERT`.

src/celeritas/optical/ScintillationPreGenerator.hh
```diff
@@ -0,0 +1,143 @@
...
+
+    // TODO: Use visible energy deposition when Birks quenching is available
+    real_type mean_num_photons = material.yield * step_.energy_dep.value();
```
sethrj: Also maybe precalculate and save mean_num_photons in the input?

src/celeritas/optical/ScintillationData.hh
stognini: The `materials` yes. The `particles` is not, since `particles.size() == num_opt_materials * num_opt_particles`.
But your point is valid, since I guess we also won't really need those during the main loop, as we always convert material or particle id into their respective scintillation id and access the data directly from those ids.

src/celeritas/io/ImportOpticalMaterial.hh
stognini: The import bit I was on the fence on that. I left it as an `or` in case someone decides to store everything in their geometry description and decide to use either material or particle sampling at run time.

sethrj: Good point, I'm fine with that. (We'll have to remember to add such a switch or to validate that both aren't there.)

src/celeritas/optical/ScintillationParams.cc
stognini: Just readability. I can keep everything inside `build_components()` if you think it's better.

src/celeritas/optical/ScintillationParams.hh
stognini: Here in the params is where we break from the import: During import, we load whatever is in the geometry. Then, at startup time, our `ScintillationParams` is constructed based on the user-defined input and only loads one or another. This is decided via the `ScintillationParams::Input::scintillation_by_particle` boolean -- which could be also be an enum, now that I think I about it. Something like this would probably look better:
```cpp
enum class ScintillationType
{
    material,
    particle
};
```

src/celeritas/optical/ScintillationParams.cc
sethrj: I think it'd be better to just move the validations inside the final loop. Defining a separate class member function that takes a vector makes the purpose less clear. If you want to move the validates to a separate place, put them in an anonymous namespace and take a single component rather than a vector, since nothing about the validation requires information about multiple components simultaneously.

src/celeritas/optical/ScintillationParams.hh
sethrj: Yeah, either that or (better) a `std::variant` so the only data members that are truly available are the ones that correspond to the given option.

src/celeritas/optical/ScintillationParams.hh
stognini: Not really, since `matid_to_optmatid` and `pid_to_scintpid` are always the size of the total material list and total particle list. So even if we sample only by material, we still know which particles are being associated to the scintillation process. Of course, if the boolean is true and we do sample scintillation by particle, then we'd also use the `pid_to_scintpid` along with `ScintillationData::spectrum_index(...)` to get the index of `ScintillationData::particles` that is storing said specific particle+material spectrum.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -37,80 +40,212 @@ ScintillationParams::from_import(ImportData const& data)
...
     {
-        input.data.push_back(mat.second.scintillation);
+        input.matid_to_optmatid[matid] = optmatid++;
+        input.data.push_back(iom.scintillation);
+
+        if (!input.pid_to_scintpid.empty())
+        {
+            // Every material must have same number of scintillation particles
+            CELER_ASSERT(input.pid_to_scintpid.size()
+                         == iom.scintillation.particles.size());
```
sethrj: Should be `CELER_VALIDATE` since we're checking user data

src/celeritas/optical/ScintillationPreGenerator.hh
```diff
@@ -0,0 +1,149 @@
...
+{
+    // Material-only sampling
+    OpticalDistributionData result;
+    if (mean_num_photons_ > 10)
+    {
+        real_type sigma = shared_.resolution_scale[mat_id_]
+                          * std::sqrt(mean_num_photons_);
+        result.num_photons
+            = NormalDistribution<real_type>(mean_num_photons_, sigma)(rng)
+              + real_type{0.5};
```
sethrj:
```suggestion
        // Round to the nearest whole number and ensure we don't sample an invalid number of photons
        result.num_photons
            = clamp_to_nonneg(NormalDistribution<real_type>(mean_num_photons_, sigma)(rng)
              + real_type{0.5});
```

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,67 +35,155 @@ struct ScintillationComponent
...
+    //! Index between OpticalMaterialId and MaterialId
+    Collection<OpticalMaterialId, W, M, MaterialId> matid_to_optmatid;
+    //! Index between ScintillationParticleId and ParticleId
+    Collection<ScintillationParticleId, W, M, ParticleId> pid_to_scintpid;
+
+    //! Resolution scale for each material
+    Collection<real_type, W, M, OpticalMaterialId> resolution_scale;
+
+    //! Material-only scintillation spectrum data
+    MaterialItems materials;  //!< [OpticalMaterialId]
```
amandalund:
```suggestion
    //! Resolution scale for each material
    MaterialItems<real_type> resolution_scale;

    //! Material-only scintillation spectrum data
    MaterialItems<MaterialScintillationSpectrum> materials;
```
and
```c++
template<class T>
using MaterialItems = Collection<T, W, M, OpticalMaterialId>;
```
above might be slightly clearer.

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,67 +35,151 @@ struct ScintillationComponent
...
 struct ScintillationData
 {
     template<class T>
     using Items = Collection<T, W, M>;
-    template<class T>
-    using OpticalMaterialItems = Collection<T, W, M, OpticalMaterialId>;
+    using MaterialItems
+        = Collection<MaterialScintillationSpectrum, W, M, OpticalMaterialId>;
+    using ParticleItems
+        = Collection<ParticleScintillationSpectrum, W, M, ParticleScintSpectrumId>;
```
sethrj:
```suggestion
    template<class T>
        using OpticalMaterialItems = Collection<T, W, M, OpticalMaterialId>;
    template<class T>
    using ParticleSpectrumItems
        = Collection<T, W, M, ParticleScintSpectrumId>;
```
and then use `OpticalMaterialItems<MaterialScintillationSpectrum>` and `OpticalMaterialItems< real_type >`  and `ParticleSpectrumItems < ParticleScintillationSpectrum>` (or even omit the typedef no the last one since it's unique to that indexing.

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,67 +35,151 @@ struct ScintillationComponent
...
  * \c yield is the characteristic light yield of the material.
  * \c resolution_scale scales the standard deviation of the distribution of the
  * number of photons generated.
+ * \c components stores the fast/slow/etc scintillation components for this
+ * material.
  */
-struct ScintillationSpectrum
+struct MaterialScintillationSpectrum
 {
     real_type yield{};
```
sethrj: Maybe we should name this `yield_per_energy` and  definitely should note that it has units of 1/MeV. It looks like an average number of particles otherwise.

src/celeritas/optical/ScintillationPreGenerator.hh
```diff
@@ -0,0 +1,149 @@
...
+    real_type mean_num_photons_;
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with input parameters.
+ */
+inline CELER_FUNCTION ScintillationPreGenerator::ScintillationPreGenerator(
```
sethrj: The `inline` only needs to be in the declaration, not the definition.

sethrj (commented):
Last round of comments 😄

sethrj (approved):
Nice work @stognini !

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,35 +35,78 @@ struct ScintillationComponent
...
 /*!
- * Data characterizing the scintillation spectrum.
+ * Data characterizing material-only scintillation spectrum information.
  *
- * \c yield is the characteristic light yield of the material.
- * \c resolution_scale scales the standard deviation of the distribution of the
- * number of photons generated.
+ * - \c yield is the characteristic light yield of the material in [1/MeV]
+ *    units. The total light yield per step is then
+ *    `yield_per_energy * step_length` .
```
amandalund:
```suggestion
 * - \c yield_per_energy is the characteristic light yield of the material in [1/MeV]
 *    units. The total light yield per step is then
 *    `yield_per_energy * step_length` .
```
When is `yield * step_length` used? Should this say something like, "the mean number of photons produced in a step is `yield * energy_deposition`"?

src/celeritas/optical/ScintillationData.hh
```diff
@@ -29,35 +35,78 @@ struct ScintillationComponent
...
+ * Data characterizing the scintillation spectrum for all particles and
+ * materials.
+ *
+ * Sampling using material-only data or particle- and material-dependent data
+ * are mutually exclusive. Therefore, either \c materials or \c particles are
+ * loaded at the beginning of the simulation, but *never* both at the same
+ * time. The \c scintillation_by_particle() function can be used to check that.
+ *
+ * - \c matid_to_optmatid returns an \c OpticalMaterialId given a
+ *   \c MaterialId
```
amandalund:
```suggestion
```

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -36,11 +39,67 @@ ScintillationParams::from_import(ImportData const& data)
...
+    auto const& num_optmats = data.optical.size();
+
     Input input;
-    for (auto const& mat : data.optical)
+    input.resolution_scale.resize(num_optmats);
+    if (data.optical_params.scintillation_by_particle)
+    {
+        // Collect ScintillationParticleIds
+        input.pid_to_scintpid.resize(data.particles.size());
+        ScintillationParticleId scintpid;
```
amandalund:
```suggestion
        ScintillationParticleId scintpid{0};
```
By default this is initialized to `size_type(-1)`.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -36,11 +39,67 @@ ScintillationParams::from_import(ImportData const& data)
...
+        for (auto const& [matid, iom] : data.optical)
+        {
+            auto const& iomsp = iom.scintillation.particles;
+            for (auto const& [pdg, ipss] : iomsp)
+            {
+                auto const pid = particle_params->find(PDGNumber{pdg});
+                // Add new ScintillationParticleId
+                input.pid_to_scintpid[pid.get()] = scintpid++;
```
amandalund:
```suggestion
                if (auto const pid = particle_params->find(PDGNumber{pdg}))
                {
                    // Add new ScintillationParticleId
                    input.pid_to_scintpid[pid.get()] = scintpid++;
                }
```
Since electrons are the only particles of the optical particle types supported in Celeritas, only the electron particle-dependent spectrum data will ever be stored/used, right?

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -36,11 +39,67 @@ ScintillationParams::from_import(ImportData const& data)
...
+
+            for (auto const& [pdg, ipss] : iomsp)
+            {
+                ImportParticleScintSpectrum part_spec;
+                part_spec.yield_vector = ipss.yield_vector;
+                part_spec.components = ipss.components;
+                input.particles[num_optmats * scintpidx + optmatidx]
+                    = std::move(part_spec);
+                scintpidx++;
+            }
```
amandalund:
```suggestion
            for (auto const& [pdg, ipss] : iomsp)
            {
                if (auto const pid = particle_params->find(PDGNumber{pdg}))
                {   
                    auto scintpid = input.pid_to_scintpid[pid.get()];
                    CELER_ASSERT(scintpid);
                    ImportParticleScintSpectrum part_spec;
                    part_spec.yield_vector = ipss.yield_vector;
                    part_spec.components = ipss.components;
                    input.particles[num_optmats * scintpid.get() + optmatidx]
                        = std::move(part_spec);
                } 
            }
```
`input.particles` won't necessarily have the same size as `iom.scintillation.particles` (which can have up to the number of particles defined in `G4MaterialPropertiesIndex`, most of which aren't currently supported in Celeritas).

src/celeritas/optical/ScintillationParams.hh
```diff
@@ -16,30 +16,55 @@
...
+        using VecSPId = std::vector<ScintillationParticleId>;
+
+        std::vector<double> resolution_scale;
+
+        //! Material-only spectra
+        std::vector<ImportMaterialScintSpectrum> materials;
+
+        //!< ParticleId to ScintillationParticleId
+        VecSPId pid_to_scintpid;
+        //! Particle and material spectra [ParticleScintillationSpectrumId]
```
amandalund:
```suggestion
        //! Particle and material spectra [ParticleScintSpectrumId]
```

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -36,11 +39,67 @@ ScintillationParams::from_import(ImportData const& data)
...
+            auto const& iomsp = iom.scintillation.particles;
+            for (auto const& [pdg, ipss] : iomsp)
+            {
+                auto const pid = particle_params->find(PDGNumber{pdg});
+                // Add new ScintillationParticleId
+                input.pid_to_scintpid[pid.get()] = scintpid++;
+            }
+        }
+        // Resize particle- and material-dependent spectra
+        input.particles.resize(input.pid_to_scintpid.size() * num_optmats);
```
amandalund: Is the size of the particle- and material-dependent spectra meant to be `num_particles * num_optmats` or `num_scint_particles * num_optmats`?

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -50,67 +109,139 @@ ScintillationParams::from_import(ImportData const& data)
...
-        for (auto i : range(comp.size()))
+        CELER_ENSURE(host_data.num_scint_particles > 0);
+
+        // Store particle spectra
+        CELER_EXPECT(!input.particles.empty());
+        GenericGridBuilder build_grid(&host_data.reals);
+        CollectionBuilder build_particles(&host_data.particles);
+
+        for (auto spec : input.particles)
         {
```
amandalund, line 189: Maybe validate that `spec.yield_vector` is non-empty?

amandalund (changes requested):
Thanks @stognini! Not sure if there's a good place to test this, but we might want to make sure we can build the `ScintillationParams` from the imported data without any issue when `scintillation_by_particle` is enabled.

Since electrons are the only one of the scintillation particle types currently supported in Celeritas, how do we plan to use this to do the particle-dependent sampling? I guess with a standalone Celeritas app we won't be, at least anytime soon. If we're getting the scintillation distribution data in an integrated Geant4 app, would we need to also store the incident particle type in the distribution data, and import all of those scintillation particle types in our `ParticleParams`?

stognini:
> we might want to make sure we can build the ScintillationParams from the imported data without any issue when scintillation_by_particle is enabled.

This is tested in the `Scintillation` `particle_scint_params` test, which pulls data from the `lar-sphere.gdml`. I did not import that boolean from G4 in the `GeantImporter` yet, but the hard part, which is having the params being constructed accordingly, at least is (allegedly) working.

> Since electrons are the only one of the scintillation particle types currently supported in Celeritas, how do we plan to use this to do the particle-dependent sampling?

The particle sampling should happen both in the pre-generator and generator. The pre-generator will 1) use the `ParticleId` (from `ParticleTrackView`) to verify if the particle is a scintillation particle; 2) check if scintillation by particle is enabled (from `ScintillationData`); 3) verify if the data is correctly assigned through an operator bool (which will throw an assertion failure if it is not); and 4) sample the mean number of photons accordingly. The generator will have a similar behavior, depending on both particle and material id (the particle id will have to be added to `OpticalDistributionData` so we correctly fetch particle+material data during the sampling in the generator.

>  If we're getting the scintillation distribution data in an integrated Geant4 app, would we need to also store the incident particle type in the distribution data, and import all of those scintillation particle types in our ParticleParams?

The `ScintillationData::pid_to_scintpid` is taking care of that. So whatever scintillation particle data is imported by the `GeantImporter` (which is limited by Geant4's small list defined by their enums) ends up being constructed in the `ScintillationParams` and has a `ScintillationParticleId` assigned to it. Of course, if Celeritas does not transport that particle (e.g. an `ion` or a `proton`), we for now will have the data in-memory in `ScintillationData`, but the pre-generator and generator will never sample those, since we will never fetch said `ParticleId` from `ParticleTrackView` (since there is no `ParticleTrackView` for a proton or ion being constructed during the main loop). I thought about ignoring these extra particles if they did not exist in Celeritas, but seemed an extra complication that would probably make very little difference, given the small total size of the data.

src/celeritas/optical/ScintillationData.hh
stognini: Oops. That was a typo. It's `yield_per_energy * energy_dep`. I'll fix the comment.

src/celeritas/optical/ScintillationParams.cc
stognini: Oh, I did not initialize it so the very first element a few lines below has a `scintpid == 0`. Unless I am missing a better way of doing this (instead of starting at -1 and then using `scintpid++` to assign new values).

src/celeritas/optical/ScintillationParams.cc
stognini: That is a good point. Scratch my other comment about storing data from particles we do not have in our `ScintillationParams`, because it is wrong. This happens in the `GeantImporter`, but not in the `ScintillationParams`, since we depend on `ParticleParams` to begin with...

src/celeritas/optical/ScintillationParams.cc
stognini: Ohhh, good catch!

src/celeritas/optical/ScintillationParams.cc
amandalund: If you're starting at -1 you would need to pre-increment rather than post-increment, but also the `OpaqueId` `operator++` will check that the value is valid before incrementing.

stognini: Oh true. Fixing that.

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -47,19 +47,22 @@ ScintillationParams::from_import(ImportData const& data,
...
+                if (auto const pid = particle_params->find(PDGNumber{pdg}))
+                {
+                    // Add new ScintillationParticleId
+                    input.pid_to_scintpid[pid.get()] = scintpid++;
+                }
             }
         }
         // Resize particle- and material-dependent spectra
-        input.particles.resize(input.pid_to_scintpid.size() * num_optmats);
+        auto const num_scint_particles = scintpid.get() + 1;
```
amandalund:
```suggestion
        auto const num_scint_particles = scintpid.get();
```
Since `scintpid` is post-incremented.

amandalund (commented):
> This is tested in the Scintillation particle_scint_params test, which pulls data from the lar-sphere.gdml

This tests constructing the params from `ScintillationParams::Input`, not the `from_import()` method using the imported data, but I did confirm that with this last little fix that method is also working.

I guess my point about the particle-dependent sampling is that with standalone Celeritas we won't be using that capability since we only support one of those particle types (electron), and with a Geant4-integrated app we would need some additional changes to be able to use it:
  - Store the Geant4 particle type that produced the scintillation photons either in the distribution data or the optical primary (depending on the integration approach)
  - Change how we're storing the particle-dependent scintillation data in our params so that we save it for all scintillation particles types, not just the particles supported in Celeritas (electron only).

amandalund (approved):
Thanks again @stognini, unless @whokion has any comments I think this is good to merge.

whokion (approved):
@stognini @amandalund Thanks for nice work!

PR #1154, opened 2024-03-17
Rename `Wentzel` model to `CoulombScattering` and add em/params subdirectory

amandalund:
As discussed in #1129, this renames the `WentzelModel`/`Data`/`Interactor`/`Executor` to `CoulombScattering` and moves the em `Params` classes to a params subdirectory.


src/celeritas/em/params/WentzelVIMscParams.cc
```diff
@@ -12,14 +12,12 @@
 
 #include "corecel/io/Logger.hh"
 #include "corecel/sys/ScopedMem.hh"
-#include "celeritas/em/data/WentzelVIMscData.hh"
+#include "celeritas/em/detail/MscParamsHelper.hh"
```
sethrj: Shouldn't the "MSC params helper" be under `em/params/detail`?

sethrj (approved):

whokion (approved):
@amandalund All look good. Thanks!

PR #1155, opened 2024-03-18
Add utilities for building `UnitInput` from `CsgTree`

sethrj:
This adds several utility functions necessary to complete the construction of a nice clean happy UnitInput.
- Implemented `find_sorted` for treating a sorted vector like a map
- The postfix builder can remap surface IDs to remove references to unused surfaces
- The postfix builder now returns a {face, logic with face indices} pair
- Renamed `local_orange_outside_volume` to `orange_exterior_volume` and added `orange_global_universe` for UniverseId{0}
- Added a `ProtoMap` function for ordering proto-universes and mapping pointers to `UniverseId`s
- Added a simple function to return the external bounding box from a bounding zone
- Added a utility to determine whether a volume may have internal surface crossings


src/corecel/math/Algorithms.hh
```diff
@@ -253,6 +253,35 @@ CELER_FORCEINLINE_FUNCTION ForwardIt upper_bound(ForwardIt first,
...
+//---------------------------------------------------------------------------//
+/*!
+ * Find the given element in a sorted range.
+ */
+template<class ForwardIt, class T, class Compare>
+inline CELER_FUNCTION ForwardIt
+find_sorted(ForwardIt first, ForwardIt last, T const& value, Compare comp)
+{
+    auto iter = ::celeritas::lower_bound(first, last, value, comp);
+    if (iter == last || comp(*iter, value) || comp(value, *iter))
```
elliottbiondo, line 265: I believe `iter == last` isn't necessary here

src/orange/orangeinp/detail/InternalSurfaceFlagger.cc
```diff
@@ -0,0 +1,100 @@
...
+        CELER_ASSUME(!node.valueless_by_exception());
+        status = static_cast<Status>(std::visit(*this, node));
+    }
+    CELER_ENSURE(status >= 0);
+    return status;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Aliased nodes forward to the alias.
```
elliottbiondo: "Forward to aliased node" for sentence structure consistency

elliottbiondo (approved):

src/corecel/math/Algorithms.hh
sethrj: lower_bound returns `last` if the searched-for value is higher than the final value in the sequence. Not testing this means segfaulting on the following `comp` command.

elliottbiondo: When `iter == last`, you can enter the `if` statement and return last, or NOT enter the `if` statement and return `iter`, both returning equivalent results.

elliottbiondo: @sethrj  i.e. this is equivalent to what you have:
```
    if (comp(*iter, value) || comp(value, *iter))
    {
        // Insertion point is off the end, or value is not equal
        return last;
    }
    return iter;
```

elliottbiondo: Ah okay that makes sense

PR #1157, opened 2024-03-19
Add generic grid builder and refactor builder code that uses generic grids

sethrj:
This adds a helper class for building `GenericGridData` (which we can later rename to `GenericGridRecord` and probably consolidate some of the grid classes). It removes the unused "generic grid" value grid builder/inserter functions for now, and also removes the unsupported "interp" data for the grids, since those are also unused.

I also added some error handling code for the livermore PE reader and inserter, since hydrogen doesn't have low-energy cross sections.


src/celeritas/io/LivermorePEReader.cc
```diff
@@ -95,25 +97,32 @@ LivermorePEReader::operator()(AtomicNumber atomic_number) const
...
+            for (int i = 0; i < size; ++i)
             {
                 CELER_ASSERT(infile);
                 infile >> result.xs_lo.x[i] >> result.xs_lo.y[i];
             }
         }
+        else if (atomic_number != AtomicNumber{1})
+        {
+            // Hydrogen doesn't have low-energy cross sections
+            CELER_LOG(warning) << "No low-energy cross sections found in '"
```
amandalund: Total cross sections below the K-shell energy aren't present for elements with only one subshell, so this warning should be for Z > 2.

test/celeritas/grid/ValueGridBuilder.test.cc
```diff
@@ -123,30 +123,6 @@ TEST_F(ValueGridBuilderTest, log_grid)
     }
 }
 
-TEST_F(ValueGridBuilderTest, DISABLED_generic_grid)
```
amandalund, line 126: Maybe we should add a test for the new `GenericGridBuilder`?

whokion (approved):
Thanks for the work.  Probably @amandalund has more feedback.

src/celeritas/em/model/LivermorePEModel.cc
```diff
@@ -21,16 +21,18 @@
 #include "celeritas/global/CoreParams.hh"
 #include "celeritas/global/CoreState.hh"
 #include "celeritas/global/TrackExecutor.hh"
+#include "celeritas/grid/GenericGridBuilder.hh"
```
amandalund:
```suggestion
```

src/celeritas/em/model/detail/LivermoreXsInserter.hh
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+    using units::MevEnergy;
+
+    LivermoreElement el;
+
+    // Add tabulated total cross sections
+    if (inp.xs_lo)
+    {
+        // Hydrogen doesn't have low-energy cross sections
```
amandalund:
```suggestion
        // Z < 3 don't have low-energy cross sections
```

src/celeritas/phys/detail/ParticleInserter.hh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add a particle.
+ */
+auto ParticleInserter::operator()(Input const& inp) -> Id
+{
+    CELER_EXPECT(inp.mass >= zero_quantity());
+    CELER_EXPECT(inp.decay_constant >= 0);
```
amandalund: Should we be `CELER_VALIDATE`-ing these?

amandalund (approved):
Looks good, thanks @sethrj!

src/celeritas/phys/detail/ParticleInserter.hh
sethrj: Good point, I just moved these from `ParticleParams::ParticleParams` but yes we should.

PR #1160, opened 2024-03-21
Split ExplicitActionInterface to support optical data

pcanal:
ActionSequence::execute is now templated on the Params/State type and cast them to the concrete type.


src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -91,9 +92,15 @@ void ActionSequence::begin_run(CoreParams const& params, CoreState<M>& state)
...
-template<MemSpace M>
-void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
+template<typename Params, template<MemSpace M> class State, MemSpace M>
+void ActionSequence::execute(Params const& params, State<M>& state)
 {
+    static_assert(
+        std::is_same_v<
+            State<M>,
+            typename celeritas::detail::ParamsTraits<Params>::template State<M>>,
+        "The Params and State type are not matching.");
```
sethrj, line 102: Would it be better (and I'm not saying it will be) to replace the static assert with an `, enable_if_t<..., bool> = true`? Would the error message be clearer, and/or would it prevent an ambiguity?

src/celeritas/global/ActionInterface.hh
```diff
@@ -100,25 +100,52 @@ class BeginRunActionInterface : public virtual ActionInterface
...
     //! Execute the action with host data
     virtual void execute(CoreParams const&, CoreStateHost&) const = 0;
 
     //! Execute the action with device data
     virtual void execute(CoreParams const&, CoreStateDevice&) const = 0;
+};
 
-    //! Dependency ordering of the action
-    virtual ActionOrder order() const = 0;
+#ifdef HAVE_OPTICAL_PARAMS
```
sethrj: Instead of an invalid preprocessor, just forward-declare the not-yet-extant class and uncomment it.

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -107,7 +114,10 @@ void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
         {
             ScopedProfiling profile_this{actions_[i]->label()};
             Stopwatch get_time;
-            actions_[i]->execute(params, state);
+            auto const& concrete_action
+                = dynamic_cast<typename celeritas::detail::ParamsTraits<
+                    Params>::ExplicitAction const&>(*actions_[i]);
```
sethrj: This would be the place for a smart_cast right? Also for clarity move the `ExplicitAction` type alias to the top of the function. You also can omit the `celeritas::detail` since we're already in that namespace.

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -121,7 +131,10 @@ void ActionSequence::execute(CoreParams const& params, CoreState<M>& state)
         for (SPConstExplicit const& sp_action : actions_)
         {
             ScopedProfiling profile_this{sp_action->label()};
-            sp_action->execute(params, state);
+            auto const& concrete_action
+                = dynamic_cast<typename celeritas::detail::ParamsTraits<
+                    Params>::ExplicitAction const&>(*sp_action);
```
sethrj: And you can reuse your type alias here :)

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -140,6 +153,13 @@ ActionSequence::execute(CoreParams const&, CoreState<MemSpace::host>&);
 template void
 ActionSequence::execute(CoreParams const&, CoreState<MemSpace::device>&);
 
+#ifdef HAVE_OPTICAL_PARAMS
```
sethrj: Instead of dead code, write `// TODO: template on optical data`

src/celeritas/global/detail/ActionSequence.hh
```diff
@@ -20,6 +20,8 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 class ActionRegistry;
 class CoreParams;
+template<typename P>
+struct ParamsTraits;
```
sethrj: I think this was from a failed attempt.

src/celeritas/global/detail/ParamsTraits.hh
```diff
@@ -0,0 +1,51 @@
...
+#ifdef HAVE_OPTICAL_PARAMS
+template<>
+struct ParamsTraits<OpticalParams>
+{
+    template<MemSpace M>
+    using State = OpticalState<M>;
+
+    using ExplicitAction = ExplicitOpticalActionInterface;
+};
+#endif
```
sethrj: Same with the todos

src/celeritas/global/detail/ActionSequence.cc
pcanal: Which type alias?

src/celeritas/global/detail/ActionSequence.cc
sethrj: The answer is no.

src/celeritas/global/detail/ParamsTraits.hh
sethrj: @pcanal did you forget to push?

src/celeritas/global/detail/ActionSequence.cc
pcanal: `static_cast` does not work with virtual inheritance.

sethrj (approved):
Nice, thanks!

PR #1161, opened 2024-03-22
Add Parallelepiped shape

mrguilima:
Add support for the parallelepiped shape *[orange]*

Tracing unit tests to be implemented in another MR.

### DIFF
--- a/src/orange/orangeinp/ConvexRegion.cc
+++ b/src/orange/orangeinp/ConvexRegion.cc
@@ -355,6 +355,87 @@ void InfWedge::output(JsonPimpl* j) const
 }
 
 //---------------------------------------------------------------------------//
+// PARALLELEPIPED
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with a 3-vector of half-edges and three angles (in radians).
+ */
+Parallelepiped::Parallelepiped(Real3 const& half_projs,
+                               Turn alpha,
+                               Turn theta,
+                               Turn phi)
+    : hpr_{half_projs}, alpha_{alpha}, theta_{theta}, phi_{phi}
+{
+    for (auto ax : range(Axis::size_))
+    {
+        CELER_VALIDATE(hpr_[to_int(ax)] > 0,
+                       << "nonpositive half-edge - roughly along "
+                       << to_char(ax) << " axis: " << hpr_[to_int(ax)]);
+    }
+
+    CELER_VALIDATE(alpha_ >= zero_quantity() && alpha_ < Turn{0.25},
+                   << "invalid angle " << alpha_.value()
+                   << " [turns]: must be in the range [0, 0.25]");
+
+    CELER_VALIDATE(theta_ >= zero_quantity() && theta_ < Turn{0.25},
+                   << "invalid angle " << theta_.value()
+                   << " [turns]: must be in the range [0, 0.25]");
+
+    CELER_VALIDATE(phi_ >= zero_quantity() && phi_ < Turn{1.},
+                   << "invalid angle " << phi_.value()
+                   << " [turns]: must be in the range [0, 1)");
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build surfaces.
+ */
+void Parallelepiped::build(ConvexSurfaceBuilder& insert_surface) const
+{
+    constexpr auto X = to_int(Axis::x);
+    constexpr auto Y = to_int(Axis::y);
+    constexpr auto Z = to_int(Axis::z);
+
+    // cache trigonometric values
+    real_type sinth, costh, sinphi, cosphi, sinal, cosal;
+    sincos(theta_, &sinth, &costh);
+    sincos(phi_, &sinphi, &cosphi);
+    sincos(alpha_, &sinal, &cosal);
+
+    // base vectors
+    auto a = hpr_[X] * Real3{1, 0, 0};
+    auto b = hpr_[Y] * Real3{sinal, cosal, 0};
+    auto c = hpr_[Z] * Real3{sinth * cosphi, sinth * sinphi, costh};
+
+    // positioning the planes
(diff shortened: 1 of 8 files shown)


src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -230,6 +230,60 @@ class InfWedge final : public ConvexRegionInterface
...
+ * The shape has the center in the origin and it is defined by:
+ *
+ *   - `dX, dY, dZ:` half-lengths of the projections of the edges on X, Y
+ *     and Z. The lower Z face is positioned at `-dZ`, while the upper at
+ * `+dZ`.
+ *   - `alpha:` angle between the segment defined by the centers of the
+ *     X-parallel edges and Y axis `[-90,90]` in degrees
+ *   -   `theta:` theta angle of the segment defined by the centers of the Z
+ * faces;
+ *   - `phi:` phi angle of the same segment
```
sethrj: It looks like clang-format messed up the indentation here; could you fix?

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -354,6 +354,87 @@ void InfWedge::output(JsonPimpl* j) const
...
+    insert_surface(Sense::outside, PlaneZ{-hpr_[Z]});
+    insert_surface(Sense::inside, PlaneZ{hpr_[Z]});
+
+    // Build the side planes roughly perpendicular to y-axis
+    insert_surface(Sense::outside, Plane{ynorm, -yoffset});
+    insert_surface(Sense::inside, Plane{ynorm, yoffset});
+
+    // Build the side planes roughly perpendicular to x-axis
+    insert_surface(Sense::inside, Plane{-xnorm, xoffset});
+    insert_surface(Sense::inside, Plane{xnorm, xoffset});
```
sethrj, line 426: Could you also add the code to set the exterior bounding box when the faces are non-axis-aligned? I think this is correct:
```c++
auto half_diagonal = 0.5 * (a + b + c);
insert_surface(Sense::inside, BBox{-half_diagonal, half_diagonal});
```

sethrj (commented):
Looks great @mrguilima ! Nice work.

src/orange/orangeinp/ConvexRegion.hh
mrguilima: Fixed.

src/orange/orangeinp/ConvexRegion.cc
mrguilima: Done, except for the 0.5 factor which is not right, as `a,b,c` vectors themselves represent half-edges.

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -354,6 +354,91 @@ void InfWedge::output(JsonPimpl* j) const
     to_json_pimpl(j, *this);
 }
 
+//---------------------------------------------------------------------------//
+// PARALLELEPIPED
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with a 3-vector of half-edges and three angles (in radians).
```
```suggestion
 * Construct with a 3-vector of half-edges and three angles.
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -354,6 +354,91 @@ void InfWedge::output(JsonPimpl* j) const
...
+    for (auto ax : range(Axis::size_))
+    {
+        CELER_VALIDATE(hpr_[to_int(ax)] > 0,
+                       << "nonpositive half-edge - roughly along "
+                       << to_char(ax) << " axis: " << hpr_[to_int(ax)]);
+    }
+
+    CELER_VALIDATE(alpha_ >= zero_quantity() && alpha_ < Turn{0.25},
+                   << "invalid angle " << alpha_.value()
+                   << " [turns]: must be in the range [0, 0.25]");
```
sethrj: Since it can't be a full 90 degrees this is a half-open range
```suggestion
                   << " [turns]: must be in the range [0, 0.25)");
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -354,6 +354,91 @@ void InfWedge::output(JsonPimpl* j) const
...
+
+    CELER_VALIDATE(theta_ >= zero_quantity() && theta_ < Turn{0.25},
+                   << "invalid angle " << theta_.value()
+                   << " [turns]: must be in the range [0, 0.25]");
```
```suggestion
                   << " [turns]: must be in the range [0, 0.25)");
```

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -230,6 +230,60 @@ class InfWedge final : public ConvexRegionInterface
...
+ * parallelograms in the general case. The Z faces have 2 edges parallel
+ * with the X-axis.
+ *
+ * The shape has the center in the origin and it is defined by:
+ *
+ *   - `dX, dY, dZ:` half-lengths of the projections of the edges on X, Y
+ *     and Z. The lower Z face is positioned at `-dZ`, while the upper at
+ * `+dZ`.
+ *   - `alpha:` angle between the segment defined by the centers of the
+ *     X-parallel edges and Y axis `[-90,90]` in degrees;
```
sethrj:
```suggestion
 *     X-parallel edges and Y axis, up to a quarter turn (i.e. 90 degrees);
```
Is it supposed to be valid to have a -90 angle? The validation in the constructor prevents that.

sethrj (commented):
One last set of small comments before we merge this. Thanks again @mrguilima !

src/orange/orangeinp/ConvexRegion.hh
mrguilima: An alpha angle of +/-90deg would make Y become X, which would degenerate that face and invalidate the shape.  So I have updated this comment and kept the validation.
I have also included the negative angles, which I had originally excluded.

sethrj (approved):
Awesome! Great work @mrguilima .

PR #1163, opened 2024-03-25
Add optical track data classes

sethrj:
This adds the core classes we'll need to run a separate optical photon stepping loop and adds to-dos from https://github.com/celeritas-project/celeritas/issues/886#issuecomment-2073298919 .


src/celeritas/optical/OpticalTrackData.cc
```diff
@@ -0,0 +1,60 @@
...
+    CELER_VALIDATE(stream_id < params.scalars.max_streams,
+                   << "multitasking stream_id=" << stream_id.unchecked_get()
+                   << " exceeds max_streams=" << params.scalars.max_streams);
+#if CELERITAS_CORE_GEO != CELERITAS_CORE_GEO_GEANT4
+    resize(&state->geometry, params.geometry, size);
+#else
+    // Geant4 state is stream-local
+    resize(&state->geometry, params.geometry, stream_id, size);
+#endif
+    CollectionBuilder{&state->materials}.resize(size);
```
amandalund:
```suggestion
    resize(&state->materials, size);
```
But will the size actually be the number of volumes?

src/celeritas/optical/OpticalTrackData.hh
```diff
@@ -0,0 +1,162 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Memspace-independent core variables.
+ */
+struct OpticalScalars
+{
+    ActionId boundary_action;
+
+    StreamId::size_type max_streams{0};
+    OpticalMaterialId::size_type num_materials{0};
```
amandalund, line 53: What will `num_materials` be used for?

amandalund (approved):
Thanks @sethrj! I think we can merge this once the CI passes.

src/celeritas/optical/OpticalTrackData.cc
sethrj: If we need both the regular *and* optical material IDs when tracking photons, it would make more sense to have the matid -> optical ID indirection. Otherwise, I think it would be a little more efficient to directly store optical materials?

src/celeritas/optical/OpticalTrackData.hh
sethrj: Hmm not sure why we put that there

sethrj: hmm optical properties?

src/celeritas/optical/OpticalTrackData.hh
```diff
@@ -0,0 +1,162 @@
...
+struct OpticalStateData
+{
+    template<class T>
+    using Items = StateCollection<T, W, M>;
+
+    GeoStateData<W, M> geometry;
+    Items<OpticalMaterialId> materials;
+    OpticalPhysicsStateData<W, M> physics;
+    RngStateData<W, M> rng;
+    SimStateData<W, M> sim;  // TODO: has a few things we don't need
```
amandalund, line 116: Which sim data do we not need? I guess `num_steps`, `num_looping_steps`, `along_step_action`... do we still need the track/parent/event IDs?

PR #1165, opened 2024-03-25
Import WLS data for optical photons

stognini:
This PR is tied to issue #886 and imports optical photon wavelength shifting (WLS) data from Geant4.

### DIFF
--- a/app/celer-dump-data.cc
+++ b/app/celer-dump-data.cc
@@ -710,7 +710,7 @@ void print_optical_material_data(ImportData::ImportOpticalMap const& iom)
         cout << POM_STREAM_SCALAR(mid, wls, mean_num_photons, IU::unitless);
         cout << POM_STREAM_SCALAR(mid, wls, time_constant, IU::time);
         cout << POM_STREAM_VECTOR(mid, wls, absorption_length, IU::len);
-        cout << POM_STREAM_VECTOR(mid, wls, component, IU::mev);
+        cout << POM_STREAM_VECTOR(mid, wls, component, IU::unitless);
     }
     cout << endl;
 #undef PEP_STREAM_SCALAR
--- a/app/celer-dump-data.cc
+++ b/app/celer-dump-data.cc
@@ -702,6 +702,17 @@ void print_optical_material_data(ImportData::ImportOpticalMap const& iom)
         cout << POM_STREAM_VECTOR(mid, abs, absorption_length, IU::len);
     }
     cout << endl;
+    cout << "\n## WLS";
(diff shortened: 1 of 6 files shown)


app/celer-dump-data.cc
```diff
@@ -702,6 +702,17 @@ void print_optical_material_data(ImportData::ImportOpticalMap const& iom)
...
     cout << endl;
+    cout << "\n## WLS";
+    cout << header;
+    for (auto const& [mid, val] : iom)
+    {
+        auto const& wls = val.wls;
+        cout << POM_STREAM_SCALAR(mid, wls, mean_num_photons, IU::unitless);
+        cout << POM_STREAM_SCALAR(mid, wls, time_constant, IU::time);
+        cout << POM_STREAM_VECTOR(mid, wls, absorption_length, IU::len);
+        cout << POM_STREAM_VECTOR(mid, wls, component, IU::mev);
```
whokion: I find that the WLS component is unitless (it is a distribution of relative population along the re-emission energy axis and is used to build an inverse CDF to sample the emitted photon energy) - sorry about a wrong guess during our discussion at the hackaton.

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -147,6 +147,28 @@ struct ImportOpticalProperty
...
+    ImportPhysicsVector component;  //!< Emission spectrum [MeV, MeV]
+
+    //! Whether all data are assigned and valid
+    explicit operator bool() const
+    {
+        return mean_num_photons > 0 && time_constant > 0
+               && static_cast<bool>(absorption_length)
+               && static_cast<bool>(component)
+               && absorption_length.vector_type == ImportPhysicsVectorType::free
+               && absorption_length.vector_type == component.vector_type;
```
whokion:
```suggestion
               && component.vector_type == absorption_length.vector_type;
```
Not sure which is the better for readability.

whokion (approved):
Thanks @stognini for adding WLS material properties.  All look good except the unit for the WLS component which is unitless, which need to be changed.

app/celer-dump-data.cc
stognini: Ohh I see. Thanks, fixing that now.

src/celeritas/io/ImportOpticalMaterial.hh
stognini: I like your suggestion.

sethrj (approved):
LGTM! Thanks!

PR #1166, opened 2024-03-25
Complete ORANGE construction from CSG objects

### LINKED ISSUE #1078: Add ORANGE construction of single units

sethrj:
This completes the initial implementation of building surface-based multilevel ORANGE universes from shapes/solids/CSG operations. It transforms the constructed "CSG unit" into an `OrangeInput` classes. I've re-implemented the `universes.org.omn` test and ensured the constructed geometry is the same and that tracking works on it.


sethrj:
@elliottbiondo Don't worry,  most of the "changes" in the diff are new JSON files.

src/orange/orangeinp/CsgObject.hh
```diff
@@ -95,20 +95,24 @@ using AnyObjects = JoinObjects<op_or>;
 //! Intersection of the given objects
 using AllObjects = JoinObjects<op_and>;
 
+//! Shared pointer to an object
+using SPConstObject = std::shared_ptr<ObjectInterface const>;
+//! Type used for defining an RDV
```
elliottbiondo: Since this is an actual doc string, not a comment, write out region definition vector

elliottbiondo (approved):

PR #1168, opened 2024-03-28
Add sense evaluator for testing

sethrj:
This is a helpful utility for evaluating "construction-time" CSG volumes. The intended purpose is for testing, but this could also be used for some other sort of preprocessing (or a construction-time sanity check) in the future.


sethrj:
@elliottbiondo This one next please! 😄

elliottbiondo (approved):

Issue #1167, opened 2024-03-26
Convert Geant4 quadric geometry to ORANGE


sethrj:
# Objects and structure 
## Transforms

Geant4 CSG solids:
- [x] IntersectionSolid
- [x] UnionSolid
- [x] ReflectedSolid
- [x] SubtractionSolid

Booleans in Geant4 have some restrictions: they must be contiguous, subtractions can't delete a face, ...

## Solids

Without sections:
- [x] Paraboloid https://github.com/celeritas-project/celeritas/pull/1893
- [x] Tet

VecGeom/Geant4 solids with radial/azimuthal/polar sections:
- [x] Hype (hyperbola)
- [x] Cone/Cons
- [x] CutTube/CutTubs
- [x] Sphere
- ~~Torus2/Torus ~~
- [x] Tube/Tubs

VecGeom/Geant4 "csg" solids:
- [x] SExtruVolume/ExtrudedSolid[^1]
- [x] GenericPolycone
- [x] Polycone/PCON
- [x] Polyhedron/Polyhedra/PGON
- ~~Tessellated/TessellatedSolid[^2]~~

Geant4 solids (not implemented in vecgeom):
- ~~Twisted box~~
- ~~Twisted trap~~
- ~~Twisted trd~~
- ~~Twisted tubs~~
- [x] Multi-union (accelerated union)

[^1]: vecgeom supports extruded solid only with 2 sections, unit scale factors, and zero offsets
[^2]: tesselated solid isn't implemented in vecgeom GPU

## Logical volumes

Logical volumes start off as a solid with an extra name. They can be reflected using the `G4ReflectionFactory`, which effectively adjusts their transformation.

They also have *daughters*, which are physical volumes (see below).
As a first pass, logical volumes should be created as solids with a vector of daughters. We can then convert some of these into protos.
- If a logical volume has no daughters, it's just a "material". It's a leaf in the tree.
- If a logical volume only appears in one location, we *could* simply propagate its transform to its daughters.
- If a logical volume has daughters and is reused multiple times, it should be a "unit proto".

## Physical volumes

Physical volumes are some kind of modification of a logical volume:
- A `G4PVPlacement` is just a transformed logical volumes
- A `G4VPVParameterisation` is a repeated logical volumes (which could potentially be represented as an array, or just expanded out as a "pseudo-array")

The world is a physical volume.

# Use cases

## HGCAL
- [x] Box
[...]

PR #1169, opened 2024-03-28
Add conversion of Geant4 transforms and basic solids to ORANGE

sethrj:
This adds support for some of the basic Geant4 shapes and solids:
- Box
- Polyhedra with 2 z planes (aka prism) with a rotated start angle
- Subtraction (box minus polyhedron)
- Tube
- Polycone with 2 planes and equal radii (aka a translated cylinder)

It also adds:
- Cone shape
- Displaced solid
- Intersection solid
- Orb
- Para

See https://github.com/celeritas-project/celeritas/issues/1167.

To test the translation, a unit test converts the shapes, builds them in to surfaces, and evaluates a list of points for "is inside" and compares them to the "is inside" result from Geant4 shapes.

@drbenmorgan A few things that celeritas-project/g4vg#32 would help clear up so far:
- Deduplicating `#if G4VERSION_NUMBER >= ....` to get a consistent interface
- Using improved semantics like `std::optional<...> inner_shape;` instead of `rmin == 0`
- Remove inconsistent nomenclature like `GetStartPhi` in some places and `GetStartPhiAngle` in others
- Dangerous accessors (raw arrays) in GetOriginalParameters
- Clear up the bazillion redundant accessors for displaced solids (what reference frame, how the translation and rotation matrix are applied)


sethrj:
@elliottbiondo I see that this one is a bit large; I can break off a couple of pieces if you want.

elliottbiondo:
If there is a quick way of splitting it in half that would be great, but I can review it as-is as well

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,517 @@
...
+namespace
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Return theta, phi angles for a G4Para or G4Trap given their symmetry axis.
+ */
+[[maybe_unused]] auto calculate_theta_phi(G4ThreeVector const& axis)
+    -> std::pair<double, double>
+{
+    // The components of the symmetry axis for G4Para/Trap are alway encoded
```
mrguilima: alway -> always

PS: I still don't know why I can't suggest code changes directly in the github page.

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,517 @@
...
+}
+
+//---------------------------------------------------------------------------//
+//! Convert a parallelepiped
+auto SolidConverter::para(arg_type solid_base) -> result_type
+{
+    auto const& solid = dynamic_cast<G4Para const&>(solid_base);
+    CELER_DISCARD(solid);
+    CELER_NOT_IMPLEMENTED("para");
+}
```
mrguilima: Why are some of these functions still not using recently added shapes?

sethrj:
@elliottbiondo @mrguilima If this is too big I can break it further into [scaler+transformer, converter]. But the first two are important to understanding the second.

mrguilima:
Looks good to me.
Thanks, @sethrj!

mrguilima (approved):
LGTM.

src/orange/g4org/SolidConverter.cc
sethrj: I started this code before the parallelepiped was merged 😅 and didn't want to make this PR too big. Once this PR is merged, would you implement the conversion code for parallelepiped and gentrap in a new PR?

mrguilima: I can certainly give it a try!

sethrj: Actually I guess the draft version was missing Para but it's there now.

src/orange/g4org/Scaler.hh
```diff
@@ -0,0 +1,63 @@
...
+    Scaler() : scale_{celeritas::lengthunits::millimeter} {}
+
+    //! Scale with an explicit factor, probably for testing
+    explicit Scaler(double sc) : scale_{sc} { CELER_EXPECT(scale_ > 0); }
+
+    //! Convert a positional scalar
+    double operator()(double val) const { return val * scale_; }
+
+    //! Convert a two-vector
+    Array<double, 2> operator()(G4TwoVector const& vec) const
```
elliottbiondo, line 41: The two-vector and three-vector methods here don't involve any scaling, and are basically just type conversions, so the feel out of place here.

src/orange/g4org/SolidConverter.hh
```diff
@@ -0,0 +1,114 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using arg_type = G4VSolid const&;
+    using result_type = std::shared_ptr<orangeinp::ObjectInterface const>;
+    //!@}
+
+  public:
+    inline SolidConverter(Scaler const& convert_scale,
```
elliottbiondo: I get that `scale_` and `transform_` act as functors and are therefore named as verbs, not nouns, but that is an implementation detail isn't exposed to the caller. I would recommend changing `convert_scale` to just `scaler`and same with the second argument.

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,640 @@
...
+
+    // Calculation taken from GetTheta/Phi implementations in Geant4 11
+    double const theta
+        = std::atan(std::sqrt(tan_theta_cos_phi * tan_theta_cos_phi
+                              + tan_theta_sin_phi * tan_theta_sin_phi));
+    double const phi = std::atan2(tan_theta_sin_phi, tan_theta_cos_phi);
+    return {theta, phi};
+}
+
+//---------------------------------------------------------------------------//
```
elliottbiondo: Missing doc string

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,640 @@
...
+{
+    return std::make_shared<Shape<CR>>(std::string{solid.GetName()},
+                                       CR{std::forward<Args>(args)...});
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the enclosed azimuthal angle by a solid.
+ *
+ * This converts from radians (native, Geant4) to Turns.
```
elliottbiondo: "Turns" should not be capitalized here and elsewhere

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,640 @@
...
+ */
+template<class S>
+SolidEnclosedAngle get_polar_wedge(S const& solid)
+{
+    return SolidEnclosedAngle{
+        native_value_to<Turn>(solid.GetStartThetaAngle()),
+        native_value_to<Turn>(solid.GetDeltaThetaAngle())};
+}
+
+//---------------------------------------------------------------------------//
```
elliottbiondo, line 108: Doc string

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,640 @@
...
+        cache_iter->second = this->convert_impl(solid_base);
+    }
+
+    CELER_ENSURE(cache_iter->second);
+    return cache_iter->second;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a geant4 solid to a VecGeom sphere with equivalent capacity.
```
elliottbiondo: "Geant4" for consistency

src/orange/g4org/SolidConverter.cc
```diff
@@ -0,0 +1,640 @@
...
+    ConvertFuncPtr fp = func_iter->second;
+    result = (this->*fp)(solid_base);
+    CELER_ENSURE(result);
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+// CONVERTERS
+//---------------------------------------------------------------------------//
+//! Convert a box
```
elliottbiondo, line 219: Since this is an actual doxygen comment, shouldn't it end in a period?

src/orange/g4org/Transformer.hh
```diff
@@ -0,0 +1,155 @@
...
+    inline Translation operator()(G4ThreeVector const& t) const;
+
+    // Convert a pure rotation
+    inline Transformation operator()(G4RotationMatrix const& rot) const;
+
+    // Convert a translation + rotation
+    inline Transformation
+    operator()(G4ThreeVector const& t, G4RotationMatrix const& rot) const;
+
+    //! Convert a translation + optional rotation
```
elliottbiondo: No doxygen exclamation point here and below for declarations

elliottbiondo (approved):

src/orange/g4org/Scaler.hh
sethrj: They forward to the "scalar" scaling method. It's only for converting positions/lengths, not directions

src/orange/g4org/SolidConverter.cc
sethrj: Even as a class name quantity?

src/orange/g4org/SolidConverter.cc
sethrj: The one-line "briefs" like this one don't need it... I decided to use those instead of full doc blocks since the functions are generally short and obvious

src/orange/g4org/Scaler.hh
elliottbiondo: Oh, as a class name it's fine, but as a unit it would be lowercase of course.

PR #1170, opened 2024-03-29
Convert Geant4 logical and physical volumes

sethrj:
This builds temporary ORANGE input structures out of Geant4 logical and physical volumes. The next step will be to collapse "used-once" logical volume and build protos/materials out of the physical volumes.

### DIFF
--- a/src/orange/g4org/LogicalVolumeConverter.cc
+++ b/src/orange/g4org/LogicalVolumeConverter.cc
@@ -39,26 +39,51 @@ LogicalVolumeConverter::LogicalVolumeConverter(SolidConverter& convert_solid)
  */
 auto LogicalVolumeConverter::operator()(arg_type lv) -> result_type
 {
-    auto [cache_iter, inserted] = cache_.insert({&lv, nullptr});
-    if (inserted)
+    SPLV result;
+    auto [cache_iter, inserted] = cache_.insert({&lv, {}});
+    if (inserted || cache_iter->second.expired())
     {
-        // First time converting the volume
-        cache_iter->second = this->construct_impl(lv);
+        // First time converting the volume, or it has expired
+        result = this->construct_impl(lv);
+        cache_iter->second = result;
+    }
+    else
+    {
+        result = cache_iter->second.lock();
     }
 
-    CELER_ENSURE(cache_iter->second);
-    return cache_iter->second;
+    CELER_ENSURE(result);
+    return {result, inserted};
 }
 
 //---------------------------------------------------------------------------//
 /*!
  * Convert the raw logical volume without processing any daughters.
  */
-auto LogicalVolumeConverter::construct_impl(arg_type g4lv) -> result_type
+auto LogicalVolumeConverter::construct_impl(arg_type g4lv) -> SPLV
 {
     auto result = std::make_shared<LogicalVolume>();
+
+    // Save Geant4 volume pointer for later mappings
     result->g4lv = &g4lv;
 
+    // Save name
+    result->name = g4lv.GetName();
+    if (result->name.find("0x") == std::string::npos)
+    {
+        // No pointer address: add one
+        result->name = make_gdml_name(g4lv);
+    }
+
+    // Save material ID
+    if (auto* cuts = g4lv.GetMaterialCutsCouple())
+    {
+        auto idx = cuts->GetIndex();
+        CELER_ASSERT(idx >= 0);
+        result->material_id = MaterialId{static_cast<size_type>(idx)};
+    }
+
+    // Convert solid
     try
     {
         result->solid = convert_solid_(*g4lv.GetSolid());
@@ -74,22 +99,6 @@ auto LogicalVolumeConverter::construct_impl(arg_type g4lv) -> result_type
                            << to_string(*result->solid) << ")";
         CELER_LOG(info) << "Unsupported solid belongs to logical volume "
                         << PrintableLV{&g4lv};
-    }
-
-    // Save material ID
-    if (auto* cuts = g4lv.GetMaterialCutsCouple())
-    {
-        auto idx = cuts->GetIndex();
-        CELER_ASSERT(idx >= 0);
-        result->material_id = MaterialId{static_cast<size_type>(idx)};
-    }
-
-    // Save name
-    result->name = g4lv.GetName();
-    if (result->name.find("0x") == std::string::npos)
-    {
-        // No pointer address: add one
-        result->name = make_gdml_name(g4lv);
     }
 
     return result;
--- a/src/orange/g4org/LogicalVolumeConverter.cc
+++ b/src/orange/g4org/LogicalVolumeConverter.cc
@@ -1,0 +1,100 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2023-2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/g4org/LogicalVolumeConverter.cc
+//---------------------------------------------------------------------------//
+#include "LogicalVolumeConverter.hh"
+
+#include <G4LogicalVolume.hh>
+#include <G4MaterialCutsCouple.hh>
+#include <G4VSolid.hh>
+
+#include "corecel/Assert.hh"
+#include "corecel/io/Logger.hh"
+#include "geocel/GeantGeoUtils.hh"
+
+#include "SolidConverter.hh"
+#include "Volume.hh"
+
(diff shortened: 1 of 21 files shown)


src/orange/g4org/LogicalVolumeConverter.cc
```diff
@@ -0,0 +1,126 @@
...
+ * Construct with solid conversion helper.
+ */
+LogicalVolumeConverter::LogicalVolumeConverter(SolidConverter& convert_solid)
+    : convert_solid_(convert_solid)
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a Geant4 logical volume to a VecGeom LogicalVolume.
```
mrguilima: No VecGeom involved.

```suggestion
 * Convert a Geant4 logical volume to an ORANGE LogicalVolume.
```

sethrj:
Ping @elliottbiondo , sorry I forgot to call this one out explicitly earlier. It's the next in the series.

src/orange/g4org/PhysicalVolumeConverter.cc
```diff
@@ -0,0 +1,220 @@
...
+PhysicalVolumeConverter::~PhysicalVolumeConverter() = default;
+
+//---------------------------------------------------------------------------//
+auto PhysicalVolumeConverter::operator()(arg_type g4world) -> result_type
+{
+    CELER_EXPECT(!g4world.GetRotation());
+    CELER_EXPECT(g4world.GetTranslation() == G4ThreeVector(0, 0, 0));
+
+    CELER_LOG(status) << "Converting Geant4 geometry";
+    ScopedProfiling profile_this{"import-geant-geo"};
```
mrguilima, line 97: Consider renaming to something different than `geocel/g4vg/Converter`'s own scoped profiler.  
Maybe this is ok, as the use of orange or vecgeom are mutually exclusive?

src/orange/g4org/PhysicalVolumeConverter.cc
```diff
@@ -0,0 +1,220 @@
...
+    if (inserted)
+    {
+        // Queue up daughters for construction
+        auto num_daughters = g4lv->GetNoDaughters();
+        lv->daughters.reserve(num_daughters);
+        for (auto i : range(num_daughters))
+        {
+            G4VPhysicalVolume* g4pv = g4lv->GetDaughter(i);
+            CELER_ASSERT(g4pv);
+            daughter_queue.push_back({depth + 1, g4pv, lv});
```
mrguilima: Could there be any improved performance from data locality, by having the daughters immediately converted after its mother, rather than being queued for later conversion?
My thought is that for a point inside a mother volume, determining the next boundary will require mother's and daughter's surfaces data on memory (e.g. data locality).

src/orange/g4org/PhysicalVolumeConverter.cc
```diff
@@ -0,0 +1,238 @@
...
+        {
+            CELER_LOG(debug)
+                << "Processing parameterized volume " << g4pv.GetName()
+                << " with " << g4pv.GetMultiplicity() << " instances";
+        }
+
+        // Loop over number of replicas
+        for (auto j : range(g4pv.GetMultiplicity()))
+        {
+            // Use the paramterization to *change* the physical volume's
```
mrguilima:
```suggestion
            // Use the parameterization to *change* the physical volume's
```

src/orange/g4org/PhysicalVolumeConverter.hh
```diff
@@ -0,0 +1,57 @@
...
+#include <memory>
+
+#include "Volume.hh"
+
+class G4LogicalVolume;
+class G4VPhysicalVolume;
+
+namespace celeritas
+{
+namespace g4org
```
mrguilima, line 21: g4org always becomes g4.org in my mind...  Can it be renamed to g4orng?

src/orange/g4org/Transformer.hh
```diff
@@ -124,8 +133,7 @@ auto Transformer::operator()(G4ThreeVector const&,
 auto Transformer::operator()(G4AffineTransform const& affine) const
     -> Transformation
 {
-    return Transformation{transposed_from_geant(affine.NetRotation()),
-                          scale_.to<Real3>(affine.NetTranslation())};
+    return (*this)(affine.NetTranslation(), affine.NetRotation());
```
mrguilima, line 118: Why is the `scale` not needed here?

src/orange/transform/Transformation.cc
```diff
@@ -40,14 +40,16 @@ Transformation::Transformation() : Transformation{Translation{}} {}
...
+ * The input rotation matrix should be an orthonormal matrix. Its determinant
+ * is 1 if not reflecting (proper) or -1 if reflecting (improper).  It is the
+ * caller's job to ensure a user-provided low-precision matrix is
+ * orthonormal: see \c celeritas::orthonormalize .
  */
 Transformation::Transformation(Mat3 const& rot, Real3 const& trans)
     : rot_(rot), tra_(trans)
 {
-    CELER_EXPECT(soft_equal(determinant(rot_), real_type(1)));
+    // XXX: should this be a CELER_VALIDATE?
```
mrguilima: It is probably a good idea to provide clear warnings or errors even in non-debug mode, in case of interactive changes in GDML files, which is a common use case.

mrguilima (approved):
Awesome, @sethrj!
Just a few comments.
Thanks!

src/orange/g4org/Volume.hh
```diff
@@ -0,0 +1,66 @@
...
+{
+    std::string name;
+    size_type copy_number{};
+    VariantTransform transform;
+    std::shared_ptr<LogicalVolume const> lv;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * A reusable Object that can be turned into a UnitProto or a Material.
```
elliottbiondo, line 47: Here and throughout I think it would be helpful to be more specific with terminology since there is so much (often unavoidable) clash. In other words: "an ORANGE UnitProto or Geant4 Material".

test/orange/g4org/PhysicalVolumeConverter.test.cc
```diff
@@ -0,0 +1,260 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/g4org/PhysicalVolumeConverter.test.cc
```
elliottbiondo, line 6: Confirm that this test file sufficiently tests `LogicalVolumeConverter` as well

elliottbiondo (approved):

src/orange/g4org/PhysicalVolumeConverter.cc
sethrj: I think it's good to keep the same name now that they're two separate code paths that do the same function using the same profiling name—it'll make analysis easier.

src/orange/g4org/PhysicalVolumeConverter.cc
sethrj: Maybe, but remember this is all for preprocessing, and the Geant4 objects are not very memory efficient anyway. (Also we're just pushing back a pointer rather than dereferencing it and reusing it later). I think having a queue/stack gives us more flexibility for construction since we can switch from breadth-first to depth-first or (as I was originally wondering if we had to do ) reverse depth-first.

src/orange/g4org/PhysicalVolumeConverter.hh
sethrj: The file extension we've been using in SCALE for years is `.org` for ORANGE... this is an internal implementation detail that no one outside of us should have to see so I think it'll be ok.

src/orange/g4org/Transformer.hh
sethrj: This simply turns the affine transform into vector/matrix objects that are then passed to the Transformer function that *actually* does the conversion including scaling the position.

src/orange/transform/Transformation.cc
sethrj: My thought was that perhaps this validation should be done "closer" to the user input: for example, Geant4's GDML reader (or transform matrix constructor) executes `rectify` to construct an exact matrix. I think I'll remove the comment.

test/orange/g4org/PhysicalVolumeConverter.test.cc
sethrj: Confirmed. The two classes are pretty tightly coupled and LVC is very small. The test examines both the PVs and the LVs converted.

sethrj:
Thanks @mrguilima and @elliottbiondo for the reviews!

PR #1171, opened 2024-03-30
Add minimal version of GenTrap (Arb8) shape

mrguilima:
Add support for the GenTrap shape [orange]

Tracing unit tests to be implemented in another MR.

This is a simplified version of the GenTrap (or ROOT's Arb8) shape, which still does not support twisted side faces.

Also, for the case of triangular z-faces, it seems to work properly for well behaved cases (e.g. triangular prism), while failing for harder ones, like when the two z-face triangles are not directly aligned (e.g. mirror-reflected triangles).

Lastly, the ROOT docs suggest that a tetrahedron (1 + 3 vertices) should work, but this is not yet supported.

Still, the current implementation works for a lot of the intended cases, as far as the side-faces are planar.


src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file geocel/detail/LengthUnits.hh
+//! \brief NOTE: only use inside geocel; prefer celeritas/Units.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+//#define CELER_ICRT inline constexpr real_type
```
sethrj: Remove dead code, fix lines 6/7. You can use `celeritas-gen.py -n celeritas::orangeinp` to make a new version of this file in the correct namespace if you want.

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+#include <iostream>
+#include <vector>
+#include <corecel/cont/Array.hh>
+#include <corecel/math/ArrayUtils.hh>
+#include <orange/OrangeTypes.hh>
```
sethrj: Not system includes, use quotes

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+
+namespace celeritas
+{
+namespace geoutils
+{
```
sethrj: There's not yet a `geoutils` namespace, so just put this in `celeritas`. This file should not be in a `detail` directory if it lives in `geocel/` and is used by `orange/orangeinp`. I think `geocel` is a fine place for it since it's generic, but I can't see these being used outside of orange—so perhaps `orange/orangeinp/detail` in the `celeritas::orangeinp::detail` namespace is better?

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+
+template<typename T>
+using Real2 = Array<T, 2>;
+template<typename T>
+using VecReal2 = std::vector<Real2<T>>;
+template<typename T>
+using Real3 = Array<T, 3>;
+template<typename T>
+using VecReal3 = std::vector<Real3<T>>;
```
sethrj: For a template `T` these aren't `real`s: it would have to be `real_type`. Perhaps it would be best to remove thse to avoid conflicting with the definition of `Real3` elsewhere.

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+
+enum PgonOrient : int
```
sethrj:
```suggestion
enum class Orientation
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+{
+    clockwise = -1,
+    collinear = 0,
+    counterclockwise = 1,
+    kCW = clockwise,
+    kLine = collinear,
+    kCCW = counterclockwise
```
sethrj: Remove ’90s-style hungarian notation in favor of using the scoped enum above.
```suggestion
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+};
+
+//---------------------------------------------------------------------------//
+// Find orientation of ordered triplet (a, b, c).
+// The function returns following values
+//  0 --> input points are collinear
+// -1 --> Clockwise
+// +1 --> Counterclockwise
```
sethrj: Format as doxygen comment. Also the return value is an enum so the values don't need to be described.
```suggestion
/*!
 * Find orientation of ordered triplet (a, b, c).
 */
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+template<typename T>
+CELER_FUNCTION PgonOrient orientation_2D(Real2<T> const& a,
```
sethrj: There's no need to add the type, and the formatting is wrong anyway. I think this class can also explicitly use `real_type` rather than being templated. It also needs an `inline`.
```suggestion
CELER_FUNCTION Orientation orientation(Real2<T> const& a,
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+                                         Real2<T> const& b,
+                                         Real2<T> const& c)
+{
+    T crossp = (b[0] - a[0]) * (c[1] - b[1]) - (b[1] - a[1]) * (c[0] - b[0]);
+    if (crossp == 0)
+        return collinear;  // collinear
+    return (crossp < 0) ? kCW : kCCW;  // clockwise or counterclockwise
```
sethrj: With enum values, the code *is* the comment.
```suggestion
        return Orientation::collinear;
    return (crossp < 0) ? Orientation::clockwise : Orientation::counterclockwise;
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+//---------------------------------------------------------------------------//
+/*! Check if a 2D polygon is convex.
+ *
+ * /param corners the vertices of the polygon
+ * /param degen_ok if true, consecutive degenerate points are okay, still
+ *   returns true
+ */
+template<typename T>
+CELER_FUNCTION bool
+is_convex_2D(VecReal2<T> const& corners, bool degen_ok = false)
```
sethrj: - Instead of using a `vector<Real2>`, use a `Span<Real2 const>` so we could use other data types (e.g., if we have a fixed-size array)
- The flag here isn't obvious if you only see the calling function (`is_convex(..., true)` means what???) and it's never actually used in the calling code, so just remove it.

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+    auto vecsub = [&](Real2<T> const& v1, Real2<T> const& v2) {
+        return Real2<T>{v1[0] - v2[0], v1[1] - v2[1]};
+    };
+
+    // use the cross_product(last, first) as sign reference
+    auto N = corners.size();
+    auto vec1 = vecsub(corners[0], corners[N - 1]);
+    auto vec2 = vecsub(corners[1], corners[0]);
+    auto ref = crossp(vec1, vec2);
+    for (size_type i = 0; i < N - 1; ++i)
```
sethrj:
```suggestion
    for (auto i : range(N - 1))
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+    auto crossp = [&](Real2<T> const& v1, Real2<T> const& v2) {
+        return v1[0] * v2[1] - v1[1] * v2[0];
+    };
+
+    auto vecsub = [&](Real2<T> const& v1, Real2<T> const& v2) {
+        return Real2<T>{v1[0] - v2[0], v1[1] - v2[1]};
+    };
+
+    // use the cross_product(last, first) as sign reference
+    auto N = corners.size();
```
sethrj: No uppercase variables
```suggestion
    auto num_corners = corners.size();
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+    auto vec1 = vecsub(corners[0], corners[N - 1]);
+    auto vec2 = vecsub(corners[1], corners[0]);
+    auto ref = crossp(vec1, vec2);
+    for (size_type i = 0; i < N - 1; ++i)
+    {
+        vec1 = vec2;
+        vec2 = vecsub(corners[(i + 2) % N], corners[(i + 1) % N]);
+        T val = crossp(vec1, vec2);
+        // make sure the sign is the same as the reference
```
sethrj: Capitalize comments
```suggestion
        // Make sure the sign is the same as the reference
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+    return true;
+}
+
+//---------------------------------------------------------------------------//
+/*! Check if a 3D polygon is planar
+ *
+ * /param a,b,c - corners the vertices of the polygon
+ */
+template<typename T>
+CELER_FUNCTION bool is_planar_3D(Real3<T> const& a,
```
sethrj:
```suggestion
CELER_FUNCTION bool is_planar(Real3<T> const& a,
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+                                 Real3<T> const& b,
+                                 Real3<T> const& c,
+                                 Real3<T> const& d)
+{
+    using celeritas::axpy;
+    auto vecsub = [&](Real3<T> const& v1, Real3<T> const& v2) -> Real3<T> {
+        Real3<T> result{v1};
+        axpy(-1.0, v2, &result);
```
sethrj:
```suggestion
        axpy(T{-1}, v2, &result);
```

src/geocel/detail/PolygonUtils.hh
```diff
@@ -0,0 +1,126 @@
...
+        return result;
+    };
+
+    // use the cross_product(last, first) as sign reference
+    auto norm = make_unit_vector(cross_product(vecsub(b, a), vecsub(c, a)));
+    auto val = dot_product(norm, vecsub(d, a));
+    // return std::fabs(val) < Tolerance<T>::from_default(); // undef symbol
```
sethrj: This is because `Tolerance` is defined in ORANGE and you're using the class inside `geocel`. I think this shoudl probably be a utility in ORANGE then. And instead of using `from_default`, tolerance should be an argument.

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -196,6 +196,48 @@ class Ellipsoid final : public ConvexRegionInterface
...
+    // Build surfaces
+    void build(ConvexSurfaceBuilder&) const final;
+
+    // Output to JSON
+    void output(JsonPimpl*) const final;
+
+    //// ACCESSORS ////
+
+    //! Half-length along Z
+    real_type halfz() const { return hz_; }
```
sethrj: Rename to `halfheight` like the other equivalent convex regions have

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -196,6 +196,48 @@ class Ellipsoid final : public ConvexRegionInterface
...
+
+    //! GenTrap corners
+    VecReal2 const& low_corners() const { return lo_; }
+    VecReal2 const& high_corners() const { return hi_; }
```
sethrj:
```suggestion
    //! Polygon on -z face
    VecReal2 const& lower() const { return lo_; }
    //! Polygon on +z face
    VecReal2 const& upper() const { return hi_; }
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -309,6 +310,88 @@ void Ellipsoid::output(JsonPimpl* j) const
...
 }
 
+//---------------------------------------------------------------------------//
+// GENTRAP
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from half Z height and 1-4 vertices for top and bottom planes
+ */
+GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
+    : hz_{halfz}, lo_{lo}, hi_{hi}
```
sethrj: Why don't we capture and move these to avoid the extra allocation:
```suggestion
GenTrap::GenTrap(real_type halfz, VecReal2&& lo, VecReal2&& hi)
    : hz_{halfz}, lo_{std::move(lo)}, hi_{std::move(hi)}
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -309,6 +310,88 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+{
+    CELER_VALIDATE(hz_ > 0, << "nonpositive halfZ: " << hz_);
+    CELER_VALIDATE(lo_.size() <= 4, << "too many vertices for -Z polygon.");
+    CELER_VALIDATE(hi_.size() <= 4, << "too many vertices for +Z polygon.");
+
+    CELER_VALIDATE(lo_.size() >= 3 || hi_.size() >= 3,
+                   << "not enough vertices for both of the +Z/-Z polygons.");
```
```suggestion
    CELER_VALIDATE(hz_ > 0, << "nonpositive halfheight: " << hz_);
    CELER_VALIDATE(lo_.size() >= 3 && lo_.size() <= 4, << "invalid number of vertices (" << lo_.size() << ") for -Z polygon");
    CELER_VALIDATE(hi_.size() >= 3 && hi_.size() <= 4, << "invalid number of vertices (" << lo_.size() << ")  for +Z polygon");
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -309,6 +310,88 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+        c = Real3{hi_[j][0], hi_[j][1], hz_};
+        d = Real3{hi_[i][0], hi_[i][1], hz_};
+
+        // Calculate plane parameters
+        auto normal = make_unit_vector(cross_product(b - a, c - b));
+        auto offset = dot_product(d, normal);
+
+        // *Temporarily* throws if a side face is not planar
+        CELER_ENSURE(celeritas::geoutils::is_planar_3D(a, b, c, d));
+        CELER_ENSURE(std::fabs(dot_product(d - a, normal)) < 1.e-8);
```
sethrj: "ensure" is for postconditions, "assert" is for within-function sanity checks
```suggestion
        CELER_ASSERT(celeritas::geoutils::is_planar_3D(a, b, c, d));
        CELER_ASSERT(std::fabs(dot_product(d - a, normal)) < 1.e-8);
```

test/geocel/PolygonUtils.test.cc
```diff
@@ -0,0 +1,119 @@
...
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+
+using Real2 = Array<real_type, 2>;
+using VecReal2 = std::vector<Real2>;
+using Real3 = Array<real_type, 3>;
+using celeritas::axpy;
+using namespace celeritas::geoutils;
```
sethrj: Instead of `using namespace` like this, change the test namespace so that you're always in the `test` namespace underneath whatever functions you're testing. So for a file testing functions defined in `celeritas::orangeinp`, the test namespace should be `celeritas::orangeinp::test`.

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -371,6 +371,145 @@ TEST_F(EllipsoidTest, standard)
...
+                                                    "Plane: x=1",
+                                                    "Plane: y=-1"};
+
+    EXPECT_EQ(expected_node, result.node);
+    EXPECT_VEC_EQ(expected_surfaces, result.surfaces);
+    EXPECT_FALSE(result.interior) << result.interior;
+
+    // TODO: check why this is not lower=(-,-,-), upper=(+,+,+)
+    EXPECT_VEC_SOFT_EQ((Real3{1, 1, -3}), result.exterior.lower());
+    EXPECT_VEC_SOFT_EQ((Real3{-1, -1, 3}), result.exterior.upper());
```
sethrj: I think the orientation on your outward faces is flipped somehow: the `expected_node` shows that this is "below x=-1 and above x=1".

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -371,6 +371,145 @@ TEST_F(EllipsoidTest, standard)
...
+
+    EXPECT_EQ(expected_node, result.node);
+    EXPECT_VEC_EQ(expected_surfaces, result.surfaces);
+    EXPECT_FALSE(result.interior) << result.interior;
+    EXPECT_VEC_SOFT_EQ((Real3{-1, -inf, -3}), result.exterior.lower());
+    EXPECT_VEC_SOFT_EQ((Real3{inf, inf, 3}), result.exterior.upper());
+}
+
+// TODO: this should be valid
+/*TEST_F(GenTrapTest, pentahedron)
```
sethrj: You can use the `DISABLED_` prefix to make sure that the code compiles: and you can test it by running with special googletest flags.
```suggestion
// TODO: not yet implemented
TEST_F(GenTrapTest, DISABLED_pentahedron)
```

sethrj (changes requested):
Thanks @mrguilima ! There's still a bit of work to do but I this is solid.

src/geocel/detail/PolygonUtils.hh
mrguilima: I've kept the `degen_ok` flag for now, as it is going to be useful when degenerate points are properly treated.

src/geocel/detail/PolygonUtils.hh
mrguilima: Got rid of the whole template, as suggested.  `T{-1}` replaced wtih `real_type{-1}`.

src/geocel/detail/PolygonUtils.hh
mrguilima: Replaced `from_default()` with `from_softequal()`.

src/orange/orangeinp/ConvexRegion.cc
mrguilima: Sizes of 2 or even 1 may still be valid -- as long as `lo_.size() >= 3 || hi_.size() >= 3` still holds.
But I am taking your suggestions for now.

test/orange/orangeinp/ConvexRegion.test.cc
mrguilima: Indeed, I was using counterclockwise points for both lower and upper z-faces.  
Reverted both to using clockwise.

src/geocel/detail/PolygonUtils.hh
mrguilima: Ultimately using `SoftZero<real_type> is_zero(1.0e-12)` instead.

test/orange/orangeinp/ConvexRegion.test.cc
mrguilima: Added documentation to clarify expected ordering of vertices.

sethrj (approved):
Nice work @mrguilima !

PR #1173, opened 2024-04-01
Add optical collector for generating scintillation and Cerenkov distribution data

amandalund:
This adds an optical collector class to build the pre- and post-step actions for creating the optical distribution data. A test runs the stepping loop with these actions added and checks the generated distributions.


src/celeritas/io/LivermorePEReader.cc
```diff
@@ -117,7 +117,7 @@ LivermorePEReader::operator()(AtomicNumber atomic_number) const
                 infile >> result.xs_lo.x[i] >> result.xs_lo.y[i];
             }
         }
-        else if (atomic_number <= AtomicNumber{2})
+        else if (atomic_number > AtomicNumber{2})
```
sethrj: God damn it... maybe add a pull request to fix this bug I introduced (#1157) and if we really want to be sure then add check using a scoped logger.

src/celeritas/optical/OpticalCollector.hh
```diff
@@ -0,0 +1,73 @@
...
+                     size_type stack_capacity,
+                     size_type num_streams,
+                     ActionRegistry* action_registry);
+
+    // Default destructor and move and copy
+    ~OpticalCollector();
+    OpticalCollector(OpticalCollector const&);
+    OpticalCollector& operator=(OpticalCollector const&);
+    OpticalCollector(OpticalCollector&&);
+    OpticalCollector& operator=(OpticalCollector&&);
```
sethrj: Since all the data are `shared_ptr` you should be able to use the `CELER_DEFAULT_COPY` macros.

sethrj (commented):
I think this looks really good so far. I think it wouldn't be difficult to split the pregenerator refactoring into a separate PR?

amandalund:
Thanks @sethrj, yeah I can definitely move the pre-gen changes into a separate PR.

src/celeritas/io/LivermorePEReader.cc
amandalund: No prob, I'll split this fix out too.

src/celeritas/optical/OpticalGenData.hh
```diff
@@ -0,0 +1,146 @@
...
+        = StackAllocatorData<OpticalDistributionData, W, M>;
+
+    //// DATA ////
+
+    // Pre-step data for generating optical photon distributions
+    StateItems<OpticalPreStepData> step;
+
+    // Buffers of distribution data for generating optical primaries
+    DistributionStackData cerenkov;
+    DistributionStackData scintillation;
```
amandalund: I need to think a little more about how we're going to ensure reproducibility if we're using a stack allocator to store the distribution data over possibly multiple steps... maybe these should just be state-sized collections since there will be max one distribution per track per generating process at each step, and we can move the data to a separate larger buffer for long-term storage (like we do for the track initializers)? Or maybe even just store them directly in the larger buffer and then compact them with a `remove_if` at each step?

amandalund: Ok, I'm trying out the latter...

sethrj: Good point about reproducibility. The only other thing I can think would be to make a sparse version of that by also storing the originating thread/track ID... but `remove_if` sounds more efficient.

amandalund:
@sethrj I'm happy to split this up further if you'd like since it's still a little large (maybe into actions/executors and collector/test).

src/celeritas/optical/OpticalCollector.hh
```diff
@@ -0,0 +1,79 @@
...
+    using SPGenStorage = std::shared_ptr<detail::OpticalGenStorage>;
+    //!@}
+
+  public:
+    // Construct with optical params, number of streams, and action registry
+    OpticalCollector(SPConstProperties properties,
+                     SPConstCerenkov cerenkov,
+                     SPConstScintillation scintillation,
+                     size_type buffer_capacity,
+                     size_type num_streams,
```
sethrj: These are enough arguments it might be worth making an `Input` struct—the two adjacent `size_type`s may be dangerous.

src/celeritas/optical/OpticalCollector.hh
```diff
@@ -0,0 +1,79 @@
...
+                     ActionRegistry* action_registry);
+
+    // Default destructor and move and copy
+    ~OpticalCollector() = default;
+    CELER_DEFAULT_COPY_MOVE(OpticalCollector);
+
+    //! Get the distribution data
+    SPGenStorage storage() const { return storage_; };
```
sethrj:
```suggestion
    SPGenStorage const& storage() const { return storage_; };
```
to eliminate the atomic add

test/geocel/data/lar-sphere.org.json
```diff
@@ -0,0 +1,70 @@
+{
```
sethrj, line 1: I think with #1181 we don't need a separate ORANGE input file :) since the tests using it all require Geant4.

sethrj (approved):
Looks good!

test/geocel/data/lar-sphere.org.json
amandalund: Nice!

amandalund: Oh actually I do see test failures without it:
```console
[ RUN      ] LArSpherePreGenTest.host
unknown file: Failure
C++ exception with description "/home/runner/work/celeritas/celeritas/test/Test.cc:40:
celeritas: runtime error: std::ifstream(path).good() failed:
    Failed to open test data file at '/home/runner/work/celeritas/celeritas/test/geocel/data/lar-sphere.org.json'" thrown in the test body.
```

sethrj: Aha, I forgot single-precision mode doesn't yet support Geant4 conversion (mostly out of laziness). Your call whether to restore the file or put a `CELERITAS_TEST_IF_DOUBLE`/or `fixme_single` or whatever.

amandalund: Ah gotcha... I'll just revert and we can remove it again once that's supported.

PR #1175, opened 2024-04-03
Rename scale methods in g4vg

sethrj:
This is a very minor update to the g4vg conversion internals to be consistent with #1169.


mrguilima (approved):
LGTM. 
Thanks @sethrj !

PR #1176, opened 2024-04-04
Fix warning about missing low-energy cross sections

amandalund:
This fixes the warning for missing low-energy Livermore PE cross sections, which should be for Z > 2.


sethrj (approved):
Thanks @amandalund   Sorry about the sign mixup!

PR #1177, opened 2024-04-06
Implement G4Sphere conversion

sethrj:
This implements conversion of the Sphere shape, as long as it doesn't have any polar segmentation.

See #1167

@mrguilima This is an example of how to add the GenTrap conversion with new tests (except of course you don't need to create a new Solid, you should just be able to use `make_shape`), so please do something similar in a new PR!

### DIFF
--- a/src/orange/orangeinp/ConvexRegion.cc
+++ b/src/orange/orangeinp/ConvexRegion.cc
@@ -634,5 +634,14 @@ void Sphere::output(JsonPimpl* j) const
 }
 
 //---------------------------------------------------------------------------//
+/*!
+ * Whether this encloses another sphere.
+ */
+bool Sphere::encloses(Sphere const& other) const
+{
+    return radius_ >= other.radius();
+}
+
+//---------------------------------------------------------------------------//
 }  // namespace orangeinp
 }  // namespace celeritas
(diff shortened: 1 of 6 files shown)


src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -633,6 +633,15 @@ void Sphere::output(JsonPimpl* j) const
     to_json_pimpl(j, *this);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Whether this encloses another sphere.
+ */
+bool Sphere::encloses(Sphere const& other) const
```
mrguilima, line 640: Why this enclosing test does not account for the azimuthal wedge?
E.g. is a larger-radius sphere with azimuthal wedge supposed to enclose a concentric, smaller-radius sphere without an azimuthal wedge?

mrguilima (approved):
Looks good to me.

src/orange/orangeinp/ConvexRegion.cc
sethrj: This is just for the "convex region", the sphere itself. It's only use for verification when building a solid. It's not used for checking solid enclosures.

PR #1179, opened 2024-04-06
Construct proto-universe hierarchy from converted volumes

sethrj:
This adds the code to convert an ORANGE `LogicalVolume` hierarchy into a hierarchy of `UnitProto`s. Since Geant4 has no distinction between LogicalVolumes with and without children, we have a couple of heuristics to construct a universe hierarchy from these.
1. If a logical volume has no children, simply place it as a `volume` in its parent(s) as a material.
2. If a logical volume is used only once, it's probably meant for acceleration or convenience, and we don't have to make a universe from it; we can just make a volume. Construct a volume by subtracting its children and place it as a material.
3. Otherwise (or for the "world" volume), create a new UnitProto corresponding to the logical volume.
4. If a logical volume corresponding to a proto-universe has 2 or fewer children, construct an explicit "background" volume by subtracting out the daughters. Otherwise, use a "background" fill.

The choice of 2 is somewhat arbitrary but it provides a good value for testing; we can make it user-configurable if desired (right now it's a `static constexpr` value). The tradeoff is that transporting on a `background` volume *may* be more expensive than doing an explicit volume. That may change if we implement BVH/BIH tracking to accelerate tracking.

In the future we can look at further accelerating construction by merging volumes within the same region that have the same material. The ALICE ZDC components look like they have lots of ROOT-constructed "bounding boxes" that are inlined.


sethrj:
@elliottbiondo This one is also ready for review.

src/geocel/GeantGeoUtils.cc
```diff
@@ -191,6 +214,10 @@ void reset_geant_geometry()
 #if G4VERSION_NUMBER >= 1100
         G4ReflectionFactory::Instance()->Clean();
 #endif
+        free_and_clear(G4Material::GetMaterialTable());
+        free_and_clear(G4Element::GetElementTable());
+        free_and_clear(const_cast<std::vector<G4Isotope*>*>(
+            G4Isotope::GetIsotopeTable()));
```
sethrj, line 220: @drbenmorgan This is an inconsistency between the Geant4 isotope store and the material/element stores. Not sure if we should open a PR to fix it...

sethrj:
Ping @elliottbiondo

src/orange/g4org/ProtoConstructor.hh
```diff
@@ -0,0 +1,61 @@
...
+
+#include "Volume.hh"
+
+namespace celeritas
+{
+namespace g4org
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Build a proto-universe from a logical volume.
```
elliottbiondo: an ORANGE proto-universe from a GEANT4 logical volume

src/orange/g4org/ProtoConstructor.cc
```diff
@@ -0,0 +1,231 @@
...
+        input.background.label = Label::from_geant(lv.name);
+    }
+
+    CELER_ENSURE(input);
+    return std::make_shared<orangeinp::UnitProto>(std::move(input));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Place this PV into the proto.
```
elliottbiondo: Write out physical volume

elliottbiondo (approved):

PR #1180, opened 2024-04-07
Complete GDML-to-ORANGE geometry converter

sethrj:
This adds the capability to construct an ORANGE geometry from a GDML file. It test both the construction of the `OrangeInput` that's used as an intermediary, and of the tracking itself. A follow-on PR will change downstream tests to defaulting to directly converting the GDML.

It does *not* yet support direct Geant4 logical volume mapping so this won't yet work for Geant4 integration. However, the GDML inputs from more complex detectors can be used to test EM tracking performance in Celeritas.


sethrj:
@elliottbiondo This next, please!

src/orange/OrangeParams.cc
```diff
@@ -83,6 +73,40 @@ OrangeInput input_from_json(std::string filename)
...
+        {
+            // Load with Geant4: must *not* be using run manager
+            auto* world = ::celeritas::load_geant_geometry_native(filename);
+            auto result = g4org::Converter{}(world).input;
+            ::celeritas::reset_geant_geometry();
+            return result;
+        }
+        else
+        {
+            CELER_LOG(warning) << "Using ORANGE geometry with GDML suffix "
```
elliottbiondo, line 94: I am not sure I understand this. Is it worth assuming that if trying to run with x.gdml witout GEANT4 support, the user will happen to have an x.org.json file laying around instead?

elliottbiondo (approved):

src/orange/OrangeParams.cc
sethrj: This hack has been the only way we've tested ORANGE vs Geant4 up to now. I don't yet want to remove it because not all the CI (namely, the "fast" ones) could test *any* geometry-related code if we did. Most of the "test" GDML files in the code base, and a couple of the "regression" files in `celer-regression`, do indeed have `x.org.json` alongside them.

elliottbiondo: Okay no problem

PR #1181, opened 2024-04-07
Switch ORANGE unit tests to use GDML files

sethrj:
**CHAINED** on #1180 

When ORANGE is the "core" geometry and Geant4 is enabled, this activates Geant4 conversion for unit testing.


sethrj:
@elliottbiondo Tiny PR to activate the new GDML conversion functionality!

elliottbiondo (approved):

PR #1183, opened 2024-04-12
Improve HGCal build performance by a factor of 10 using surface hashing

sethrj:
On #1180, loading and converting the HGCal geometry takes ~12 seconds, most of which is spent in surface deduplication. As anticipated, it's *not* performant to search through all existing surfaces when inserting a new one, as that's a quadratic operation.

This PR generates a spatial "bin" for each surface type, such that all surfaces that *can* compare soft equal are in it (or, if the surface is at the edge of the bin, in an adjacent bin). On inserting a new surface, it adds its ID into an *unordered* multimap that can be searched for subsequent surface additions.

The spatial bins right now are hardcoded to 0.01 * the length scale (which is 1mm for converted Geant4 geometry). This doesn't mean that surfaces that close will be deduplicated; rather, it just means that all surfaces within 0.01mm of each other will be checked for soft equivalence.

## Performance testing

With current develop, the surface insertion performance test (which inserts planes and spheres) clearly demonstrates the quadratic performance:
```
Sampling 16...8.75e-06 s
Sampling 32...2.9792e-05 s
Sampling 64...0.000111834 s
Sampling 128...0.000434041 s
Sampling 256...0.00161142 s
Sampling 512...0.00625429 s
Sampling 1024...0.0245268 s
Sampling 2048...0.0968056 s
Sampling 4096...0.370775 s
Sampling 8192...1.42327 s
Sampling 16384...5.27299 s
Sampling 32768...18.7109 s
```

On this branch, performance is *much* closer to linear. The increase at higher surface counts is likely because the acceleration grid tolerance needs tweaking.
```
Sampling 16...1.0458e-05 s
Sampling 32...1.7834e-05 s
Sampling 64...3.325e-05 s
Sampling 128...7.0083e-05 s
Sampling 256...0.000154959 s
Sampling 512...0.00034425 s
Sampling 1024...0.00083325 s
Sampling 2048...0.00217404 s
Sampling 4096...0.00705275 s
Sampling 8192...0.0232007 s
Sampling 16384...0.0844078 s
Sampling 32768...0.324391 s
```


sethrj:
@elliottbiondo I'm confident now that this works, so you can review this as well. (When we do performance testing we can revert this PR pretty easily since it doesn't touch much code.)

src/orange/orangeinp/detail/LocalSurfaceInserter.cc
```diff
@@ -7,19 +7,38 @@
...
+ *
+ * This builds an acceleration grid using the length scale of the problem
+ * (deduced from the tolerance) and the relative tolerance.
  */
 LocalSurfaceInserter::LocalSurfaceInserter(VecSurface* v,
                                            Tolerance<> const& tol)
-    : surfaces_{v}, soft_surface_equal_{tol}
+    : surfaces_{v}
+    , soft_surface_equal_{tol}
+    , calc_hashes_{real_type{0.01} * calc_length_scale(tol), 2 * tol.rel}
```
sethrj: I will make this "magic" grid spacing factor into a static constexpr variable for now; we could make it user-configurable in the future (but it will only change construction-time performance so probably isn't worth that).

src/orange/orangeinp/detail/LocalSurfaceInserter.cc
```diff
@@ -33,52 +52,96 @@ template<class S>
...
+    // Add the non-exact surface even if it's a near match (since we need to
+    // chain near-duplicates)
     all_surf.emplace_back(std::in_place_type<S>, source);
 
-    // Store the equivalency relationship and potentially chain equivalent
-    // surfaces
-    return this->merge_impl(source_id, target_id);
+    if (near_match)
+    {
+        // Surface is eqivalent to an existing one but not identical: save and
```
sethrj:
```suggestion
        // Surface is equivalent to an existing one but not identical: save and
```

elliottbiondo:
(will do today)

src/orange/orangeinp/detail/SurfaceGridHash.hh
```diff
@@ -0,0 +1,69 @@
...
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Hash "similar" surfaces for faster lookups.
+ *
+ * This is meant to generate one or more "key" values for a hash
+ *
+ * This returns hashes to give local surface indices on an infinite grid.
```
elliottbiondo: This is kind of an awkward sentence and I had to read it several times. I don't think the sentence on line 27 should come first either. Perhaps you could lead with the purpose of the class, e.g.:

"This creates a hash map of local surfaces based on a characteristic dimension (i.e. the radius of a sphere), which is used to accelerate surface deduplication. For a given surface in bin N, possible duplicates may be found in bins N-1, N, N+1."

src/orange/orangeinp/detail/SurfaceGridHash.hh
```diff
@@ -0,0 +1,69 @@
...
+{
+namespace orangeinp
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Hash "similar" surfaces for faster lookups.
+ *
+ * This is meant to generate one or more "key" values for a hash
```
elliottbiondo: Missing period

src/orange/orangeinp/detail/SurfaceGridHash.hh
elliottbiondo: I see this is documented in `LocalSurfaceInserter.hh`, but it may be worth briefly mentioning here.

elliottbiondo (approved):

PR #1184, opened 2024-04-12
Add actions and executors for generating optical distribution data

amandalund:
This adds a gather action to cache the pre-step data needed for the scintillation and Cerenkov pre-generators and a post-step pre-generator action to create the optical distribution data. The distributions are stored in two large buffers indexed by the current size plus the track slot ID which is compacted at the end of each step by removing all the invalid distributions (this will ensure reproducibility). These actions will be created by the optical collector in #1173.


src/celeritas/optical/detail/OpticalGenStorage.hh
```diff
@@ -0,0 +1,31 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/user/detail/OpticalGenStorage.hh
```
sethrj: Fix path

src/celeritas/optical/detail/PreGenAction.cc
```diff
@@ -0,0 +1,166 @@
...
+ */
+size_type
+PreGenAction::remove_if_invalid(ItemsRef<MemSpace::host> const& buffer,
+                                size_type offset,
+                                size_type size,
+                                StreamId) const
+{
+    auto* start = static_cast<OpticalDistributionData*>(buffer.data());
+    auto* stop
+        = std::remove_if(start + offset, start + offset + size, IsInvalid{});
```
stognini, line 148: Since it looks like the `IsInvalid` is only used here, maybe it's easier to just make a local lambda function here to return if data is assigned or not?

src/celeritas/optical/detail/PreGenAction.hh
```diff
@@ -0,0 +1,111 @@
...
+    using SPConstCerenkov = std::shared_ptr<CerenkovParams const>;
+    using SPConstProperties = std::shared_ptr<OpticalPropertyParams const>;
+    using SPConstScintillation = std::shared_ptr<ScintillationParams const>;
+    using SPGenStorage = std::shared_ptr<detail::OpticalGenStorage>;
+    //!@}
+
+    //! Check if the distribution data is valid
+    struct IsInvalid
+    {
+        CELER_FUNCTION bool operator()(OpticalDistributionData data) const
```
sethrj:
```suggestion
        CELER_FUNCTION bool operator()(OpticalDistributionData const& data) const
```

src/celeritas/optical/detail/PreGenGatherExecutor.hh
```diff
@@ -0,0 +1,59 @@
...
+{
+    CELER_EXPECT(state);
+
+    auto particle = track.make_particle_view();
+    if (particle.is_stopped())
+    {
+        return;
+    }
+
+    OpticalPreStepData& step = state.step[track.track_slot_id()];
```
stognini, line 50: Do we need to worry about `track.track_slot_id() < state.step.size()` before accessing/writing step data? Or is this going to be always true?

amandalund: It should always be true (but can add an assertion).

src/celeritas/optical/detail/PreGenAction.cc
amandalund: It's actually also used in the device version (`thrust::remove_if`).

src/celeritas/optical/detail/OpticalGenStorage.hh
whokion:
```suggestion
//! \file celeritas/optical/detail/OpticalGenStorage.hh
```

src/celeritas/optical/detail/PreGenAction.cc
stognini: Ohh I missed that. Makes sense then.

src/celeritas/optical/detail/PreGenAction.cc
```diff
@@ -0,0 +1,166 @@
...
+                           SPGenStorage storage)
+    : id_(id)
+    , properties_(std::move(properties))
+    , cerenkov_(std::move(cerenkov))
+    , scintillation_(std::move(scintillation))
+    , storage_(std::move(storage))
+{
+    CELER_EXPECT(id_);
+    CELER_EXPECT(scintillation_ || (cerenkov_ && properties_));
+    CELER_EXPECT(!cerenkov == !properties);
```
whokion: Not sure whether this check is correct as optical properties (refraction_index?) may be still needed for optical interaction codes (for example, Rayleigh and boundary) even when cerenkov is not activated.

src/celeritas/optical/detail/PreGenExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+                                      pos,
+                                      optmat_id,
+                                      properties,
+                                      cerenkov,
+                                      state.step[tsid]);
+        cerenkov_dist = generate(rng);
+    }
+    if (scintillation)
+    {
+        auto edep = track.make_physics_step_view().energy_deposition();
```
whokion, line 94: Should we check whether edep > 0 here?

src/celeritas/optical/detail/PreGenAction.hh
```diff
@@ -0,0 +1,111 @@
...
+    void execute(CoreParams const&, CoreStateHost&) const final;
+
+    // Launch kernel with device data
+    void execute(CoreParams const&, CoreStateDevice&) const final;
+
+    //! ID of the model
+    ActionId action_id() const final { return id_; }
+
+    //! Short name for the action
+    std::string label() const final { return "optical-pre-generator-post"; }
```
whokion, line 69: What does the trailing "-post" stand for (just for my curiosity)?

whokion (approved):
Good work!  A couple pf general comments: 1) It is okay for now, but should actions for cerenkov and scintillation be separated as they may be independently added? 2) Do we need to add any unit test (okay if it is not necessary or difficult to add at this point though), especially for reproducibility, which I could not catch the requirement clearly from the code (so, please comment the place for the reproducibility requirement)? Thanks for the nice work!

src/celeritas/optical/detail/PreGenExecutor.hh
stognini: That, or similar to the previous statement:
```cpp
if (scintillation && (auto edep = track.make_physics_step_view().energy_deposition()))
{
    ScintillationPreGenerator ...
}
```
But even in the current case the pre-generator is fairly inexpensive to return a result with zero photons.

src/celeritas/optical/detail/PreGenAction.cu
```diff
@@ -0,0 +1,74 @@
...
+#include "OpticalGenStorage.hh"
+#include "PreGenExecutor.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Generate optical distribution data post-step.
```
stognini: Multiple files have this same doxygen description. Not sure if intentional, but we could be more clear on the specific function of each struct/class function.

src/celeritas/optical/detail/PreGenExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+        = state.scintillation[DistId(size.scintillation + tsid.get())];
+
+    // clear distribution data
+    cerenkov_dist = {};
+    scintillation_dist = {};
+
+    auto sim = track.make_sim_view();
+    auto optmat_id
+        = track.make_material_view().make_material_view().optical_material_id();
+    if (sim.status() == TrackStatus::inactive || !optmat_id
```
sethrj: I don't think the track status will ever be inactive if the `Executor` is called: the `TrackExecutor` should filter out inactive tracks. If a track *is* inactive, then you might have just accessed invalid memory by getting the optical material ID...

src/celeritas/optical/detail/PreGenAction.hh
sethrj: The `pre-` is that it generates the "before an optical photon"; the `-post` is it's doing the calculation after the step (post-step).

sethrj (approved):
Looking good!

src/celeritas/optical/detail/PreGenAction.cc
amandalund: Good point, you're right, I'll remove this assertion.

src/celeritas/optical/detail/PreGenExecutor.hh
amandalund: Right, the pre-generator will return zero photons if the energy deposition is zero.

src/celeritas/optical/detail/PreGenAction.cu
amandalund: Nah that was just laziness, I'll try to make these more descriptive.

src/celeritas/optical/detail/PreGenExecutor.hh
amandalund: Isn't that only true if it's a conditional executor, i.e. constructed with `make_active_track_executor()`? The memory should still be valid even if the track is inactive. I used a regular `TrackExecutor` here because we need to clear any old distribution data in the buffers before compacting with `remove_if`.

amandalund:
Thanks @whokion: yes, we could separate the Cerenkov and scintillation actions, but this should also work as is either if both of the processes are enabled or if only one or the other is. I've added a test in #1173 (which I'll open once this is merged) which runs the stepping loop with these actions and checks the buffered data. For reproducibility the order of the distribution data in the buffers should always be the same, which is done by making sure the distributions created at each step are ordered by thread ID (originally I had used a stack allocator to buffer the data, but this would have required some additional post-processing to sort the distributions since the atomic add leads to a non-deterministic order: see https://github.com/celeritas-project/celeritas/pull/1173#discussion_r1552483425).

src/celeritas/optical/detail/PreGenExecutor.hh
sethrj: Oh ok, you're right about the conditional executor. But it still could access uninitialized memory right? If a track has never been initialized, then:
```c++
track.make_material_view().make_material_view().optical_material_id()
```
could obtainan uninitialized/invalid material ID, and then crash when trying to read the optical material from the corresponding material view.

amandalund: Ahh ok I see what you mean: we'd be trying to construct a `MaterialView` with an invalid material ID. You're right, let me fix that...

sethrj: Yeah, I ran into a similar thing once with the cross section calculation.

sethrj (approved):
LGTM. Thanks!

src/celeritas/optical/detail/PreGenAction.cc
```diff
@@ -0,0 +1,170 @@
...
+ */
+void PreGenAction::pre_generate(CoreParams const& core_params,
+                                CoreStateHost& core_state) const
+{
+    TrackExecutor execute{
+        core_params.ptr<MemSpace::native>(),
+        core_state.ptr(),
+        detail::PreGenExecutor{properties_->host_ref(),
+                               cerenkov_->host_ref(),
+                               scintillation_->host_ref(),
```
amandalund, line 129: Actually this definitely won't work as is if only one of the processes is enabled... do you think it would be cleaner to just split this into two actions as @whokion suggested or try to fix it to work as a single action (@sethrj)?

sethrj: Two actions would probably be cleaner, but if you think it's less effort to fix the single action let's go for that.

src/celeritas/optical/detail/PreGenExecutor.hh
```diff
@@ -0,0 +1,103 @@
...
+    // Clear distribution data
+    cerenkov_dist = {};
+    scintillation_dist = {};
+
+    auto sim = track.make_sim_view();
+    bool inactive = sim.status() == TrackStatus::inactive;
+    auto optmat_id = inactive ? OpticalMaterialId{}
+                              : track.make_material_view()
+                                    .make_material_view()
+                                    .optical_material_id();
```
sethrj, line 65: @amandalund I know this has been moved around (moved to `Utils.hh` in #1217) but wanted to revisit this on the original implementation... Since the `PreGenExecutor` is a "post-post-step" action, the material here is the *post-boundary-crossing* material if a track is stopped by a boundary.  I don't think that's right... perhaps we should be collecting and storing the beginning-of-step optical material in `OpticalPreStepData`?

amandalund: Ah good catch, yeah we can store it with the rest of the pre-step data.

sethrj: @amandalund to be clear (and avoid unintentionally duplicating fixes as we've managed to do far too often 😅) would you like to address this or shall I?

amandalund: I'm happy to, if you haven't already done it 😉

sethrj: I haven't! I'd be grateful if you'd take handle it. Thanks!

Issue #778, opened 2023-05-28
Polish rasterizer geo-check into "celer-geo" verification app


sethrj:
- Add template on track view/params so the same app can be reused across geometry (not just "core")
- Add omnutils raytrace support
- Add a utility for reproducing line segments across all enabled geometries, CPU and GPU

PR #1185, opened 2024-04-13
Add small utilities for rasterization work

sethrj:
This does a little refactoring for upcoming work on #778 .
- Separated "soft unit vector" into a `ArraySoftUnit` utility class
- Added `CELER_DEFAULT_MOVE_DELETE_COPY`, and prohibted copying of `CollectionStateStore` since the ref would need to be updated on copying
- Add `CELER_JSON_PAIR` for JSON I/O


amandalund (approved):
Thanks @sethrj !

PR #1189, opened 2024-04-17
Parse log levels from Geant4 scoped logger

sethrj:
The default behavior of hiding Geant4 "diagnostic" messages meant I missed important warnings about the geometry. This PR adds some regular expressions to try to parse the Geant4 error/output message to determine its origin and severity.


sethrj:
@drbenmorgan or @pcanal , would one of you please take a look at this since you are the Geant4 infrastructure experts? Thanks!

drbenmorgan (approved):
Changes look fine to me - though not sure why some of the Jenkins builds aborted. May want to relaunch those just to be sure though suspect they'll be fine.

sethrj:
Yeah, the jenkin failures are because Kokkos takes so damn long to build 🙃 see #1186

PR #1190, opened 2024-04-17
Define geometry traits

sethrj:
This provides a uniform interface for instantiating and creating geometry classes. This will be used by the ray tracer to support multiple raytrace geometries with the same installation.

### DIFF
--- a/src/geocel/GeoTraits.hh
+++ b/src/geocel/GeoTraits.hh
@@ -21,7 +21,6 @@ class GeoParamsInterface;
  * Traits class for defining params and device data.
  * \tparam G Geometry params class, e.g. VecgeomParams
  *
- *
  * This traits class \em must be defined for all geometry types. The generic
  * instance here is provided as a synopsis and to improve error checking.
  */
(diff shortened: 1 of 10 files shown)


src/geocel/GeoTraits.hh
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This traits class \em must be defined for all geometry types. The generic
+ * instance here is provided as a synopsis and to improve error checking.
+ */
+template<class G>
+struct GeoTraits
+{
+    static_assert(std::is_base_of_v<GeoParamsInterface, G>,
+                  "G must be a geometry params, not params data");
+    static_assert(sizeof(G) == 0, "Geo traits must be specialized");
```
esseivaju: Just some C++ fun, `sizeof(G) == 0` is false regardless of `G` so the compiler could reject the template before instantiation, it's just not smart enough to do so (until C++23 where `static_assert(false)` should be accepted in non-instantiated templates). The correct way to do this would be: 
```c++
template<class T>
struct always_false : std::false_type {};

template<class G>
struct GeoTraits
{
static_assert(always_false<G>::value, "");
};
```
which could technically be true depending on  `always_false` specializations.

sethrj: Aha, I was confusing `sizeof(T) == 0` with the error message that comes from undefined types, `error: invalid application of 'sizeof' to incomplete type`.

amandalund (approved):
Thanks @sethrj!

PR #1193, opened 2024-04-18
Fix prism orientation in Geant4 ORANGE solid converter

sethrj:
Add a new test showing the failure, and fix the failure.

Before:
TBHGCal181Oct-unknown

After:
TBHGCal181Oct-unknown


mrguilima (approved):
Nice catch, @sethrj!

PR #1191, opened 2024-04-17
Add generic CPU/GPU raytracer for rasterizing

sethrj:
**CHAINED** on #1190 and #1194

`RaytraceImager` is the user-facing class that takes a geometry and traces an image

Unlike the original raytrace implementation, this one is actually unit tested. The highest level unit test generates ASCII art:
```
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJBBBBBBBBBBBBJJJJ|
JJJJBBBBBBBBBBBBJJJJ|
JJJJBBBBBBBBBBBBJJJJ|
JJJJBBBBBBBBBBBBJJJJ|
JJJJccccccccccccJJJJ|
JJJJccccccccccccJJJJ|
JJJJccaaaabbbbccJJJJ|
JJJJccaaaabbbbccJJJJ|
JJJJccaaaabbbbccJJJJ|
JJJJccaaaabbbbccJJJJ|
JJJJccccccccccccJJJJ|
JJJJPcccccccccccJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
JJJJJJJJJJJJJJJJJJJJ|
```

PR #1194, opened 2024-04-19
Add basic raytrace components

sethrj:
This restructures the data backend and core raytracing kernel from `demo-rasterizer`. That app is now *temporarily* disabled, to be restored in #1191.

This is working toward https://github.com/celeritas-project/celeritas/issues/778 . It refactors the GPU-only kernel as an executor and a couple of helper classes.  Unlike the demo rasterizer, the current imager and ray tracer should be compatible with any combination of geometry and execution space.
- `ImageParams` is the specification for a "window", a 2D rectangle in 3D space
- `Image` is an actual visualization of that window, possibly generated by different geometries, different methods, or by different memory/execution spaces
- `ImageLineView` is a helper class for updating a horizontal line of the image
- `Raytracer` calculates a single integer value for a pixel in the image


sethrj:
@elliottbiondo This next please. Even though I have others that are in the queue, I'd like to get the "capability" requests in before the performacne ones.

src/geocel/rasterize/ImageData.hh
```diff
@@ -7,28 +7,128 @@
...
-    Array<unsigned int, 2> dims;  //!< Image dimensions (j, i)
-    Span<int> image;  //!< Stored image [j][i]
+    using Size2 = Array<size_type, 2>;
+
+    Real3 origin{};  //!< Upper left corner
+    Real3 down{};  //!< Downward basis vector
+    Real3 right{};  //!< Rightward basis vector (increasing i, track movement)
+    real_type pixel_width{};  //!< Width of a pixel
+    Size2 dims{};  //!< Image dimensions (j, i)
+    real_type max_length{};  //!< Maximum distance along rightward to trace
```
mrguilima: Your Scalars structure contain several vectors!

sethrj: Ha good point, I thought about that... I named it like we do for the other "runtime constants" elsewhere in the code. I've added a note on https://github.com/celeritas-project/celeritas/issues/721 to rename `Scalars`→`Constants` everywhere.

test/geocel/MockGeoTrackView.hh
```diff
@@ -0,0 +1,208 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Mock track view for testing.
+ *
+ * The mock track view has boundaries at every integer z, with a volume ID
+ * that's equal to the floor of the current position. The geometry is outside
+ * for z < 0.
+ *
+ * \note This class is only used by the raytracer test at present, so it
+ * doesn't have full functionality. Consider using ORANGE geometery for more
```
mrguilima:
```suggestion
 * doesn't have full functionality. Consider using ORANGE geometry for more
```

src/geocel/rasterize/Raytracer.hh
```diff
@@ -0,0 +1,182 @@
...
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from geo track view, function to calculate ID, and image.
+ */
+template<class GTV, class F>
+CELER_FUNCTION
+Raytracer<GTV, F>::Raytracer(GTV&& geo, F&& calc_id, ImageLineView const& image)
+    : geo_{celeritas::forward<GTV>(geo)}
+    , calc_id_{celeritas::forward<F>(calc_id)}
```
mrguilima, line 85: For a real GTV (e.g. in the case of VecGeom or ORANGE), is this `calc_id` supposed to be the geometry engine's locate function, like VecGeom's `Locate()`?  Where is the similar ORANGE function?

mrguilima (approved):
LATM.  
A is for awesome.
Thanks, @sethrj!

src/geocel/rasterize/Raytracer.hh
sethrj: No, `calc_id` is meant to be primarily "get the volume ID", but we have the flexibility to make it "get the material ID" or (for vecgeom) "get the physical volume ID".

mrguilima: So this flexibility is helpful to define colors in different ways to paint components in rasterization pictures?
Nice!

sethrj: That's the end goal, yeah!

src/corecel/Types.hh
```diff
@@ -43,12 +51,12 @@ enum class MemSpace
     host,  //!< CPU memory
     device,  //!< GPU memory
     mapped,  //!< Unified virtual address space (both host and device)
+    size_,
```
mrguilima: Why not having `size_` as the last entry?

src/geocel/rasterize/ImageIO.json.cc
```diff
@@ -7,50 +7,65 @@
...
-
-//---------------------------------------------------------------------------//
-}  // namespace
-
 //---------------------------------------------------------------------------//
 //!@{
 //! I/O routines for JSON
-void to_json(nlohmann::json& j, ImageInput const& v)
+#define IM_LOAD_OPTION(NAME) CELER_JSON_LOAD_OPTION(j, v, NAME)
+#define IM_LOAD_REQUIRED(NAME) CELER_JSON_LOAD_REQUIRED(j, v, NAME)
```
mrguilima, line 26: Shouldn't these be `#undef`  after use?

mrguilima (approved):

src/corecel/Types.hh
sethrj: Because it adds 1 to the previous entry, which is `native` (which is aliased to a lower number!)  and so ends up being *not* the last one ... I originally did that and didn't even catch it immediately 😅

src/geocel/rasterize/ImageIO.json.cc
sethrj: Well, they're used until the end of the translation unit 😅 but sure, it would be clearer to scope them more locally.

src/geocel/rasterize/Image.cc
```diff
@@ -0,0 +1,118 @@
...
+
+    ImageParamsScalars scalars;
+
+    // Normalize rightward axis
+    scalars.right = make_unit_vector(inp.rightward);
+
+    // Vector pointing toward the upper right from the lower left corner
+    Real3 diagonal = inp.upper_right - inp.lower_left;
+
+    // Set downward axis to the diagonal with the rightward component
```
elliottbiondo: I think it would be helpful to document this method more. I don't understand why scalars.down should calculated that way that it is.

test/geocel/rasterize/Image.test.cc
```diff
@@ -0,0 +1,100 @@
...
+    EXPECT_EQ(123, result[0]);
+    EXPECT_EQ(-1, result[1]);
+    EXPECT_EQ(345, result[2]);
+    EXPECT_EQ(567, result[65]);
+}
+
+TEST_F(ImageTest, inexact)
+{
+    ImageInput inp;
+    inp.lower_left = {-1, 2, -1};
```
elliottbiondo, line 100: Point out how the input make this inexact

src/geocel/rasterize/Raytracer.hh
```diff
@@ -0,0 +1,182 @@
...
+    //! Sentinel value for needing the pixel to be reset
+    static CELER_CONSTEXPR_FUNCTION size_type invalid_pixel()
+    {
+        return static_cast<size_type>(-1);
+    }
+
+    //! Stop tracking the pixel if this many crossings occur
+    static CELER_CONSTEXPR_FUNCTION short int max_crossings_per_pixel()
+    {
+        return 32;
```
elliottbiondo, line 63: Should this be a constexpr somewhere more official, rather than a "magic number"? Also, I could see this being something that needs to be configurable for a suitably zoom-out ray trace of a geometry with fine features.

elliottbiondo (changes requested):

src/geocel/rasterize/Raytracer.hh
sethrj: I agree it should be configurable; previously it was a hard-coded constant in the implementation code itself. This way at least documents the capability at the class level; later we can extend the implementation to make it configurable.

sethrj:
@elliottbiondo Does af5b5d0bafa13d9d2f3d7ca5f8286f964abee007 address the documentation/explanation well enough?

src/geocel/rasterize/ImageIO.json.cc
```diff
@@ -0,0 +1,90 @@
...
+
+void from_json(nlohmann::json const& j, ImageInput& v)
+{
+    IM_LOAD_REQUIRED(lower_left);
+    IM_LOAD_REQUIRED(upper_right);
+    IM_LOAD_REQUIRED(rightward);
+    IM_LOAD_REQUIRED(vertical_pixels);
+    IM_LOAD_OPTION(horizontal_divisor);
+
+        real_type length{lengthunits::centimeter};
```
mrguilima:
```suggestion
    real_type length{lengthunits::centimeter};
```

mrguilima (commented):
One more typo.

src/geocel/rasterize/ImageIO.json.cc
sethrj: Oof rebase error, thanks

elliottbiondo:
@sethrj I am familiar with the the upper-left origin convention and think it's good to have here. My comment was w.r.t. why the downward basis vector is being set to a diagonal:

"Set downward axis to the diagonal with the rightward component"

sethrj:
@elliottbiondo You missed the second line of the comment:
```
    // Set downward axis to the diagonal with the rightward component
    // subtracted out; then normalize
```
i.e. it's turning the diagonal vector into an orthonormal vector by projecting the diagonal onto the rightward basis vector, subtracting out, and normalizing.

elliottbiondo:
@sethrj okay I see what you mean. I think something like "calculate the downward vector from the diagonal" would be more straightforward.

elliottbiondo (approved):

PR #1188, opened 2024-04-17
Pin sphinx version to fix documentation build

sethrj:
Sphinx 7.3 was just released (https://github.com/sphinx-doc/sphinx/issues/12242) and causes our CI to fail (https://github.com/sphinx-doc/sphinx/issues/12291). This update rejects sphinx 7.3.0 so we can hopefully keep on CI-ing...

PR #1196, opened 2024-04-21
Add Windows/Linux no-dependency builds

sethrj:
The "build with zero dependencies" use case is helpful for trying to get up and running on a completely new experimental system. Adding a Windows build with error checking improves our ability to catch subtle mistakes, and improves downstream compatibility with SCALE (@hartsw ) and other users.

A follow-on request will fix the errors from no-JSON and MSVC.


src/orange/MatrixUtils.cc
```diff
@@ -16,6 +16,26 @@ using Mat3 = celeritas::SquareMatrixReal3;
 
 namespace celeritas
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+//! Use fused multiply-add for floating point calculations
+template<class T, std::enable_if_t<std::is_floating_point<T>::value, bool> = true>
+T generic_fma(T a, T b, T y)
```
sethrj: I could probably rename this `fma`, add `CELER_FUNCTION` decorators, and put it in `Algorithms.hh`?

src/celeritas/optical/OpticalGenData.hh
```diff
@@ -42,7 +42,7 @@ struct OpticalGenParamsData
 
     bool cerenkov{false};  //!< Whether Cerenkov is enabled
     bool scintillation{false};  //!< Whether scintillation is enabled
-    real_type capacity{0};  //!< Distribution data buffer capacity
+    size_type capacity{0};  //!< Distribution data buffer capacity
```
sethrj: @amandalund I assume this was the intent, and that I'm not missing something?

.jenkins
```diff
@@ -2,7 +2,7 @@ pipeline {
   options {
     disableConcurrentBuilds(abortPrevious: true)
     newContainerPerStage()
-    timeout(time: 2, unit: 'HOURS')
+    timeout(time: 7, unit: 'HOURS')
```
sethrj, line 5: Follow-on to #1186

scripts/doc-requirements.txt
```diff
@@ -1,4 +1,4 @@
-Sphinx >= 4, < 7.3
+Sphinx >= 4, ~= 7.3.0
```
sethrj, line 1: Follow-on to #1188

src/celeritas/optical/OpticalGenData.hh
amandalund: Yes, thanks, that should be `size_type`.

sethrj:
Ping @pcanal

.github/workflows/build-ultralite.yml
```diff
@@ -0,0 +1,109 @@
...
+  workflow_dispatch:
+  workflow_call:
+
+concurrency:
+  group: build-ultralite-${{github.ref}}-${{github.event.pull_request.number || github.run_number}}-${{github.workflow}}
+  cancel-in-progress: true
+
+jobs:
+  linux:
+    name: ultralite-linux
```
pcanal: Should we be more specific as the recipe requires `apt` to run properly?
```suggestion
    name: ultralite-ubuntu
```

.github/workflows/build-ultralite.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Cache ccache
+        uses: actions/cache@v4
+        with:
+          path: ${{env.CCACHE_DIR}}
+          key: ccache-ultralite-${{github.run_id}}
+          restore-keys: |
+            ccache-ultralite
+      - name: Zero ccache stats
+        run: |
+          ccache -z
```
pcanal, line 37: ~If we reset the cache at each run, how does the `ccache` still help?~

pcanal: Nevermind, this reset only the statistics and keeps around the cached products.

pcanal:
Ideally commit 524fbd4dfcc36e5cb6b73e51f57e8ef10f29f712 (and any related addition) would become its own PR as it is distinct in nature (actual code fix) from the rest of the PR (infrastructure/CI extension).

sethrj:
@pcanal Agreed, I can do that.

src/orange/MatrixUtils.cc
pcanal: It does make sense.  Then for consistent we will have to consider replacing the other use of `std::fma` (maybe even in non-templated context with `generic_fma`)

src/corecel/sys/ScopedSignalHandler.cc
```diff
@@ -26,7 +26,9 @@ sig_atomic_t volatile g_celer_signal_bits_ = 0;
 //! Set the bit corresponding to a signal
 extern "C" void celer_set_signal(int signal)
 {
+#ifndef _WIN32
```
pcanal: We should add a comment why it okay to use in linux but not on windows.  Alternative should use
```
 /EHs
Enables standard C++ stack unwinding. Catches only standard C++ exceptions when you use catch(...) syntax. Unless /EHc is also specified, the compiler assumes that functions declared as extern "C" may throw a C++ exception.
```
If that is the behavior consistent with the linux build?

.github/workflows/build-ultralite.yml
```diff
@@ -0,0 +1,111 @@
...
+      CCACHE_MAXSIZE: "20Mi"
+      CMAKE_PRESET: ultralite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get -q -y update
+          sudo apt-get -q -y install \
+            ccache cmake ninja-build
+      - name: Check out Celeritas
```
pcanal, line 26: It feels like from there on, the code is not specific to `ultralite` (except for the cache name), if this is correct then in this PR and/or a subsequent PR we should consider factoring the code.

src/orange/MatrixUtils.cc
sethrj: Done in the follow-on PR.

.github/workflows/build-ultralite.yml
sethrj: It's effectively the same as the `fast` build yes, but I'm not sure how refactoring could improve much...

src/corecel/sys/ScopedSignalHandler.cc
sethrj: Done (for the follow-on PR)

pcanal: Thanks.  Additional question in the follow-on PR: https://github.com/celeritas-project/celeritas/pull/1200/files#r1578416785

.github/workflows/build-ultralite.yml
pcanal: Fair enough.

pcanal (approved):
LGTM

PR #1197, opened 2024-04-23
Move unit enumeration to corecel and improve assertion flexibility 

sethrj:
I'm extracting a few changes from #1194 that I added for the raytrace work (to improve feedback to the user). The Celeritas unit system enum now lives in `corecel/Types`, so that it can be used by `geocel`. I also refactored the runtime assertion errors for readability, and I've changed "not configured" and "not implemented" to runtime errors, since we might want to let users encounter those naturally rather than adding additional validation checks. (For example, it's easier to let the code hit a `NOT_CONFIGURED("Vecgeom")` then have that as a debug check *and* do an additional validation on the user input.)

### DIFF
--- a/src/corecel/Assert.cc
+++ b/src/corecel/Assert.cc
@@ -25,21 +25,6 @@ namespace
 {
 //---------------------------------------------------------------------------//
 /*!
- * Return whether to give an extra verbose message.
- */
-bool determine_verbose_message()
-{
-#if CELERITAS_DEBUG
-    // Always verbose if debug flags are enabled
-    return true;
-#else
-    // Verbose if the CELER_LOG environment variable is defined
-    return !celeritas::getenv("CELER_LOG").empty();
-#endif
-}
-
-//---------------------------------------------------------------------------//
-/*!
  * Construct a debug assertion message for printing.
  */
 std::string build_debug_error_msg(DebugErrorDetails const& d)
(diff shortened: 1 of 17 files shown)


sethrj:
@amandalund This should be pretty quick: and I think @elliottbiondo and @mrguilima gave most of it a look-through in the other PR.

amandalund (approved):
Looks good @sethrj!

src/corecel/Assert.cc
```diff
@@ -64,33 +49,31 @@ std::string build_debug_error_msg(DebugErrorDetails const& d)
...
 
-    if (d.which != RuntimeErrorType::validate || verbose_message)
+    msg << "celeritas: " << color_code('R') << to_cstring(d.which)
+        << " error: " << color_code(' ') << d.what;
+
+    if (verbose_message || d.which != RuntimeErrorType::validate
+        || d.what.empty())
     {
-        msg << color_code('W') << d.file;
+        msg << '\n' << color_code('W') << d.file;
```
pcanal: If I read the code correctly, wouldn't the output in case of `d.what.empty() == true` be something like:
```
celeritas: kind_of_error error: 
filename.cxx
```
?

sethrj: It says you've commented on an outdated code? And with the refinements of #1198 here's what the various combinations print as:
Screenshot 2024-04-24 at 12 37 37

pcanal: > It says you've commented on an outdated code?

I got to this already merged PR by way of #1198, so that might be why github complains.

pcanal: And thanks for the output, it indeed looks fine.

PR #1200, opened 2024-04-24
Fix no-JSON and Windows builds

sethrj:
This is a follow-on to #1196 . It fixes the new CI builds that were introduced.

## MSVC warnings that this fixes

Throwing from a C function with windows:
```
celeritas\src\corecel\sys\ScopedSignalHandler.cc(29): error C2220: the following warning is treated as an error
celeritas\src\corecel\sys\ScopedSignalHandler.cc(29): warning C4297: '`anonymous-namespace'::celer_set_signal': function assumed not to throw an exception but does
celeritas\src\corecel\sys\ScopedSignalHandler.cc(29): note: The function is extern "C" and /EHc was specified
```

Using `std::fma` in an integer template context:
```
celeritas\src\orange\MatrixUtils.cc(76): error C2220: the following warning is treated as an error
celeritas\src\orange\MatrixUtils.cc(76): warning C4244: '=': conversion from 'double' to 'int', possible loss of data
```

Missing include:
```
celeritas\src\orange\orangeinp\detail\PostfixLogicBuilder.cc(215): error C2039: 'sort': is not a member of 'std'
```

Not propagating compile flags to celer-g4 when Geant4 is disabled:
```
ostream(779): error C2220: the following warning is treated as an error
ostream(779): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
```

Accidental use of real_type instead of size_type:
```
celeritas\src\celeritas\optical\detail\../OpticalGenData.hh(150): warning C4244: 'argument': conversion from 'const celeritas::real_type' to 'size_t', possible loss of data
```

Implicit truncation from real_type to size_type:
```
2024-04-21T11:40:36.2485761Z D:\a\celeritas\celeritas\src\celeritas/optical/ScintillationPreGenerator.hh(120): error C2220: the following warning is treated as an error
2024-04-21T11:40:36.2488477Z D:\a\celeritas\celeritas\src\celeritas/optical/ScintillationPreGenerator.hh(120): warning C4244: '=': conversion from 'T' to 'celeritas::size_type', possible loss of data
```


src/corecel/sys/ScopedSignalHandler.cc
```diff
@@ -26,7 +26,12 @@ sig_atomic_t volatile g_celer_signal_bits_ = 0;
 //! Set the bit corresponding to a signal
 extern "C" void celer_set_signal(int signal)
 {
+#ifndef _WIN32
+    // Windows exception handling settings do not allow throwing from a C
+    // function. It's not safe to do anyway if called from the signal handler,
+    // so expect this to terminate the code if the signal is invalid.
```
pcanal: As a consequence, shouldn't we use a non-throwing assert here both for Linux and Windows?

sethrj: I think calling `CELER_ASSERT` and invoking UB is still cleaner than
```c++
if (CELERITAS_DEBUG && !(signal >= 0 && signal < static_cast<int>(sizeof(int) * 8 - 1)))
{
    std::cerr << "Got out-of-range signal at " << __FILE__ << ":" << __LINE__  << '\n';
    std::abort();
}
```
for this very edgey case...

pcanal: Since this is an edgey/unlikely case, wouldn't:
```
assert(signal >= 0 && signal < static_cast<int>(sizeof(int) * 8 - 1));
```
be sufficient?

> invoking UB is still cleaner than

Technically, that "could" be (since it is undefined :) ): not-throw the exception and ignore it completely.

sethrj: Yeah good call. If we're using C signals, it's probably better to use C assertions.

src/orange/MatrixUtils.cc
```diff
@@ -73,7 +74,7 @@ gemm(SquareMatrix<T, N> const& a, SquareMatrix<T, N> const& b)
             // Accumulate dot products
             for (size_type k = 0; k != N; ++k)
             {
-                result[i][j] = std::fma(b[k][j], a[i][k], result[i][j]);
+                result[i][j] = fma(b[k][j], a[i][k], result[i][j]);
```
pcanal, line 77: There are other usage that should be change:
```
src/orange/MatrixUtils.hh:            result[i] = std::fma(alpha, a[i][j] * x[j], result[i]);
src/orange/MatrixUtils.hh:            result[i] = std::fma(alpha, a[j][i] * x[j], result[i]);
src/corecel/math/ArrayUtils.hh:        (*y)[i] = std::fma(a, x[i], (*y)[i]);
src/corecel/math/ArrayUtils.hh:        result = std::fma(x[i], y[i], result);
```
and some that should be checked and/or just changed for consistency:
```
src/corecel/grid/Interpolator.hh:        std::fma(slope_, XTraits_t::add_transformed(offset_, x), intercept_));
src/celeritas/grid/PolyEvaluator.hh:        return std::fma(arg, calc_impl<M + 1>(arg), coeffs_[M]);
src/celeritas/random/distribution/UniformRealDistribution.hh:    return std::fma(delta_, generate_canonical<RealType>(rng), a_);
src/celeritas/random/distribution/NormalDistribution.hh:        return std::fma(spare_, stddev_, mean_);
src/celeritas/random/distribution/NormalDistribution.hh:    return std::fma(r * std::sin(theta), stddev_, mean_);
```

sethrj: Ah good point. I think the first ones make sense to change; maybe even `PolyEvaluator.hh`, but the floating-point-only classes we shouldn't have to. (Well, unless we some day want to try out differentiable programming...)

pcanal: > but the floating-point-only classes we shouldn't have to. 

I agree we don't have to and I am proposing those to be consistent (no need to 'think' which one to use) and future proofing (if the function becomes a template for whatever reason).

PR #1201, opened 2024-04-24
Add celer-geo app for visualizing geometry

### LINKED ISSUE #778: Polish rasterizer geo-check into "celer-geo" verification app

sethrj:
**CHAINED** on #1191 

This implements a new rasterization front end for visualizing geometry. Within one process execution, you can render the same geometry file in multiple geometries, with multiple slices, using host and device.

The interface uses the [JSON lines](https://jsonlines.org) format as a primitive client/server interface. The app's interface is documented in the documentation.

An example of using the server interactively with python, and improved visualization, will follow in the next pull request.

Closes #778 .


sethrj:
@elliottbiondo I've completed the front end application and tested it with the python layer, but to make the PR easier to review I've left the python tools to a subsequent merge. Please take a look at this one when you can!

sethrj:
Ping @elliottbiondo

doc/main/usage.rst
```diff
@@ -186,6 +186,63 @@ Output
...
+Usage::
+
+  celer-geo {input}.jsonl
+            -
+
+Input
+-----
+
+The input and output are both formatted as `JSON lines`_. A single input JSON
+on a single line executes a command in ``celer-geo``, and a single JSON line is
```
elliottbiondo: I had to reread this several times: "A single input JSON on a single line"

How about "Each line within the JSON lines files execute a command"?

doc/main/usage.rst
```diff
@@ -186,6 +186,63 @@ Output
...
+   {"geometry_file": "simple-cms.gdml"}
+
+Subsequent lines will each specify the imaging window, the geometry, the
+binary image output filename, and the execution space (device or host for GPU
+or CPU, respectively).::
+
+   {"image": {"_units": "cgs", "lower_left": [-800, 0, -1500], "upper_right": [800, 0, 1600], "rightward": [1, 0, 0], "vertical_pixels": 128}, "volumes": true, "bin_file": "simple-cms-cpu.orange.bin"}
+
+After the first image window is specified, it will be reused if the "image" key
+is omitted. A new geometry and/or memory space may be specified, useful for
```
elliottbiondo: `s/memory space/execution space/`

app/celer-geo/Runner.hh
```diff
@@ -0,0 +1,97 @@
...
+
+class G4VPhysicalVolume;
+
+namespace celeritas
+{
+namespace app
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Set up and run rasterization, caching geometry as needed..
```
elliottbiondo: remove extra period

app/celer-geo/Runner.hh
```diff
@@ -0,0 +1,97 @@
...
+    //!@}
+
+  public:
+    // Construct with model setup
+    explicit Runner(ModelSetup const& input);
+
+    // Perform a raytrace
+    SPImage operator()(TraceSetup const&, ImageInput const&);
+
+    // Perform a raytrace using the last image but a new geometry/
```
elliottbiondo: Trailing /

app/celer-geo/celer-geo.cc
```diff
@@ -0,0 +1,317 @@
...
+    {
+        CELER_LOG(error) << "Failed to parse JSON input: " << e.what();
+        CELER_LOG(info) << "Failed line: " << repr(jsonline);
+        return nullptr;
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Execute a single raytrace.
```
elliottbiondo: copy-paste error

doc/main/usage.rst
```diff
@@ -186,6 +186,63 @@ Output
...
+verifying different navigators behave identically::
+
+   {"bin_file": "simple-cms-cpu.geant4.bin", "geometry": "geant4"}
```
elliottbiondo, line 227: So this is just a binary dump of the raytrace data --- what will be your approach to converting this into a image?

app/celer-geo/Runner.cc
```diff
@@ -0,0 +1,228 @@
...
+{
+    // Create image params
+    last_image_ = std::make_shared<ImageParams>(image_inp);
+
+    return (*this)(trace);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Perform a raytrace using the last image but a new geometry/memspace.
```
elliottbiondo, line 71: I understand repeating in a different memspace (execution space?) but is there an actual use case for repeating the same trace with a new geometry?

app/celer-geo/Runner.cc
```diff
@@ -0,0 +1,228 @@
...
+        cached = geo;
+    }
+
+    CELER_ENSURE(cached);
+    CELER_ENSURE(geo);
+    return geo;
+}
+
+//---------------------------------------------------------------------------//
+#define CASE_RETURN_FUNC_T(T, FUNC, ...) \
```
elliottbiondo: Not necessary defining macros near why there are used, but confirm this placement was intention since it looks pretty strange where it is.

app/celer-geo/Runner.cc
```diff
@@ -0,0 +1,228 @@
...
+        CASE_RETURN_FUNC_T(Geometry::vecgeom, make_imager, );
+        CASE_RETURN_FUNC_T(Geometry::geant4, make_imager, );
+        default:
+            CELER_ASSERT_UNREACHABLE();
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Create a tracer.
```
elliottbiondo: Designate this as a helper function

elliottbiondo (changes requested):

doc/main/usage.rst
sethrj: Yeah, the verbiage needs work. The point is that each line *must* be a valid JSON input: maybe "Each line of input is a JSON representation that executes a command ..."?

doc/main/usage.rst
sethrj: The python visualization script (waiting for a subsequent PR, see https://github.com/sethrj/celeritas/commit/92ec52360838ffef2e03609252d0b51cffd3f80f#diff-6b30a61667ce57aee53738523a91f9f1bb5c87901132689130e4e4e883df7c7a ) reads in the binary file,  maps the integers to colors, and renders the image with matplotlib.

sethrj: I think at some point we'll want to take an approach closer to the openmc visualization tool: take an ID->color mapping from the user, generate an actual image on the GPU (or host), and send either the PNG or (if `libpng` isn't installed) the uncompressed image back to the host.

app/celer-geo/Runner.cc
sethrj: Oh absolutely, that's going to be the most common use case for the immediate future: it's the cleanest way to do geometry validation, so it lets us compare ORANGE rendering versus Geant4 versus VecGeom. It would've been really nice to have three years ago when VecGeom was riddled with bugs...

app/celer-geo/Runner.cc
sethrj: Yeah, this is definitely a weird place. I would've put it inside the function except I used it in two functions. I think it would look better with a namespace and put at the top of the file.

doc/main/usage.rst
elliottbiondo: Okay we will want to make sure that this the visualization script is mentioned here (in the next PR)

elliottbiondo (approved):

PR #1205, opened 2024-04-25
Improve gentrap documentation and fix typos

mrguilima:
Since the input lower points can be ordered in both CW or CCW orderings, the sense needs to take into account the sign of the plane offset with respect to the plane normal.

An extra test was added for opposite ordering of points (CCW ordering), and the documentation has been adjusted.

See [review process](https://github.com/celeritas-project/celeritas/blob/develop/doc/appendix/administration.rst#code-review) for descriptions of the labels and requirements.

### DIFF
--- a/src/orange/orangeinp/ConvexRegion.cc
+++ b/src/orange/orangeinp/ConvexRegion.cc
@@ -337,6 +337,7 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
     CELER_VALIDATE(detail::is_convex(make_span(hi_)),
                    << "+Z polygon is not convex");
 
+    // TODO: Temporarily ensure that all side faces are planar
     for (auto i : range(lo_.size()))
     {
         auto j = (i + 1) % lo_.size();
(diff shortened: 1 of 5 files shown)


src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -382,7 +381,8 @@ void GenTrap::build(ConvexSurfaceBuilder& insert_surface) const
         CELER_ASSERT(SoftZero<real_type>{}(dot_product(d - a, normal)));
 
         // Insert the plane
-        insert_surface(Sense::inside, Plane{normal, offset});
+        auto sense = (offset < 0 ? Sense::outside : Sense::inside);
```
sethrj: Isn't it possible to set up `lo`/`hi` values so that the trapezoid is offset (e.g.) so that it's entirely less than zero, in which case you could have an *inside* sense with a *negative* offset? For example, the face here would have a negative offset, a positive normal, and an *inside* sense:
Screenshot 2024-04-25 at 15 35 08


Can you add a test case like that to ensure it gives the expected behavior?

sethrj (changes requested):

sethrj:
@mrguilima Maybe some additional restrictions are needed on the points input to the gentrap. For example, are there restrictions about the top and bottom points not being rotated 180 degrees?
Screenshot 2024-04-25 at 18 49 23

sethrj:
@mrguilima Actually it looks like the problem is that the normal calculation assumes the points are CCW ... which is not the case.

sethrj (approved):

PR #1206, opened 2024-04-26
Fix gentrap orientation

sethrj:
This is a follow-on to #1205 and #1171. The calculation of the face normals assumed that the faces were oriented as CCW, whereas they were only asserted to be convex. I've fixed it by testing the orientation and reversing the order of the polygons if needed.


mrguilima (approved):
Looks good.
Thanks @sethrj

PR #1208, opened 2024-04-26
Update Frontier installation

sethrj:
This updates the build on Frontier to use the new `hep143` allocation and installation with ROCm 5.7.1. 

The only weird thing was that somehow thrust now assumes that it's building CUDA when we build from clang (and include it via `device_runtime_api.h`):
```
In file included from /ccs/home/s3j/Code/celeritas-frontier/src/corecel/sys/Device.cc:21:
In file included from /ccs/home/s3j/Code/celeritas-frontier/src/corecel/device_runtime_api.h:28:
In file included from /opt/rocm-5.7.1/include/thrust/mr/memory_resource.h:25:
In file included from /opt/rocm-5.7.1/include/thrust/detail/config/memory_resource.h:22:
In file included from /opt/rocm-5.7.1/include/thrust/detail/alignment.h:24:
/opt/rocm-5.7.1/include/thrust/detail/type_traits.h:31:10: fatal error: 'cuda/std/type_traits' file not found
#include <cuda/std/type_traits>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```


esseivaju:
Are you using clang directly or hipcc? Looking at rocThrust, [compiler.h](https://github.com/ROCm/rocThrust/blob/rocm-5.7.1/thrust/detail/config/compiler.h#L82-L90) and [device_system.h](https://github.com/ROCm/rocThrust/blob/rocm-5.7.1/thrust/detail/config/device_system.h#L27-L33), if `__hip__` isn't defined then it's picking cuda. Wouldn't you have to also define [__THRUST_DEVICE_SYSTEM_NAMESPACE](https://github.com/ROCm/rocThrust/blob/rocm-5.7.1/thrust/detail/config/device_system.h#L57-L67)

sethrj:
@esseivaju This was happening through the `.cc` files compiled by `clang++`. Thrust was setting `THRUST_DEVICE_COMPILER` to `THRUST_DEVICE_COMPILER_CLANG`, and then defaulting `THRUST_DEVICE_SYSTEM` to `THRUST_DEVICE_SYSTEM_CUDA`. By overriding `THRUST_DEVICE_SYSTEM` in `device_runtime_api.h` we give thrust the correct "device system" , and then it will automatically set `__THRUST_DEVICE_SYSTEM_NAMESPACE`.

The change is only to provide Thrust more information when going into `device_system.h`, not to replace that header.

esseivaju (approved):
On the rocmThrust readme they [recommend](https://github.com/ROCm/rocThrust?tab=readme-ov-file#build-and-install) using `hipcc` to compile cc files but I guess this work around works as long as we include `device_runtime_api.h` before any Thrust headers.

sethrj:
OLCF recommends using their wacky Cray compiler wrappers... and those guys forward to llvm directly apparently

PR #1209, opened 2024-04-26
Refactor `ActionSequence` to be templated on Params

esseivaju:
`ActionSequence` is now a class template on CoreParams / OpticalParams. This allows casting actions to the more specialized `ExplicitCoreActionInterface` or `ExplicitOpticalActionInterface` in the ctor instead of doing it in `ActionSequence::execute`.

This adds the constraint that a given `ActionSequence` instance operates only on either `ExplicitCoreActionInterface` or `ExplicitOpticalActionInterface` (or any other subtype of `ExplicitActionInterface`)

### DIFF
--- a/src/celeritas/global/Stepper.hh
+++ b/src/celeritas/global/Stepper.hh
@@ -138,22 +138,22 @@ class Stepper final : public StepperInterface
     ~Stepper();
 
     // Transport existing states
-    StepperResult operator()() override final;
+    StepperResult operator()() final;
 
     // Transport existing states and these new primaries
-    StepperResult operator()(SpanConstPrimary primaries) override final;
+    StepperResult operator()(SpanConstPrimary primaries) final;
 
     // Reseed the RNGs at the start of an event for reproducibility
-    void reseed(EventId event_id) override final;
+    void reseed(EventId event_id) final;
 
     //! Get action sequence for timing diagnostics
-    ActionSequence const& actions() const override final { return *actions_; }
+    ActionSequence const& actions() const final { return *actions_; }
 
     //! Access core data, primarily for debugging
     StateRef const& state_ref() const { return state_.ref(); }
 
     //! Get the core state interface for diagnostic output
-    CoreStateInterface const& state() const override final { return state_; }
+    CoreStateInterface const& state() const final { return state_; }
 
   private:
     // Params and call sequence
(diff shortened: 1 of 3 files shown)


src/celeritas/global/Stepper.hh
```diff
@@ -137,27 +138,27 @@ class Stepper final : public StepperInterface
...
     //! Get action sequence for timing diagnostics
-    ActionSequence const& actions() const final { return *actions_; }
+    ActionSequence const& actions() const override final { return *actions_; }
 
     //! Access core data, primarily for debugging
     StateRef const& state_ref() const { return state_.ref(); }
 
     //! Get the core state interface for diagnostic output
-    CoreStateInterface const& state() const final { return state_; }
+    CoreStateInterface const& state() const override final { return state_; }
```
sethrj: `override final` is redundant: leave it as `final`. I got this suggestion from 
[the google style guide](https://google.github.io/styleguide/cppguide.html#Inheritance) and it seems nicer to have one keyword instead of two :)

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -11,6 +11,7 @@
 #include <tuple>
 #include <type_traits>
 #include <utility>
+#include <celeritas/global/CoreParams.hh>
```
sethrj: should be quotes not `<>`

sethrj (approved):

src/celeritas/global/detail/ActionSequence.cc
esseivaju: my editor somehow decided to auto-include using angle brackets...

src/celeritas/global/Stepper.hh
esseivaju: The suggestion makes sense. I'd say `override` is less redundant since the class is already declared `final`.

sethrj: Just as redundant, but less consistent. Also less consistent with the code base and the style guides we use. Is this something we should change throughout celeritas? I would need a little more convincing...

esseivaju: probably not worth the time, they're pretty much equivalent in a final class.

sethrj (approved):
Thanks @esseivaju !

Issue #1172, opened 2024-04-01
Implement single-track CPU for performance and improve integration


sethrj:
One of the open questions for our CMS integration is how well the detectors will work if we invert the `[track, step]` loop to `[step, track]` as is necessary for GPU. I believe we can without too much effort add enhanced support for a single-trackslot mode that would give us enhanced CPU performance and better integration characteristics.
1. Add a new memspace for "compact host"
2. Value type for compact host stores `T` instead of `vector<T>`
3. `CoreStatePtr` becomes `CoreState<value>*` instead of `CoreState<reference>*` so that we don't have to do additional indirection on each state item.

esseivaju:
I did some CPU profiling using callgrind/cachegrind with the following setup:
* **System:** Perlmutter login node
* **Problem:** testem3-orange-field-msc, 1 event, 64 primaries/event
* single thread (OMP disabled)
* CMake `RelWithDebInfo` build, `CELERITAS_DEBUG=OFF`

The graph below shows the estimated cycles spent in each function, weighting instruction fetch, L1, and LL cache miss.

![testem3_fm_64p](https://github.com/celeritas-project/celeritas/assets/7202402/ab5b9e54-cdf2-41c8-b76c-2e47abc51a67)

I noticed that [axpy](https://github.com/celeritas-project/celeritas/blob/55dbb28e1a383277bf368414fdfc0bea453caea9/src/corecel/math/ArrayUtils.hh#L81-L84) leads to many instruction cache miss but it could be because I didn't pass `march/mtune` compiler options.

Looking at the L1 read miss, most of them come from [XsCalculator::get](https://github.com/celeritas-project/celeritas/blob/d60080882aed40ce5e1c4dc5b0a9ba4d0559f8a8/src/celeritas/grid/XsCalculator.hh#L166-L170) calls within `XsCalculator::operator()`

It'd be interesting to see the cache miss in a multithreaded scenario

sethrj:
@esseivaju Is this with one track slot or the usual number (65K)? I guess the reason I wondered about single-thread performance not being optimal is that we saw a substantial performance gap between single-slot and many-slot, and since the many-slot case is not really optimal (in terms of state cache locality *and* skipped loops due to masking) I wonder whether the call graph would look any different...

esseivaju:
This is with 4k track slots. 

> single-thread performance not being optimal is that we saw a substantial performance gap between single-slot and many-slot,

[...]

PR #1210, opened 2024-04-28
Use string view for interface labels and descriptions

sethrj:
This prevents repeated allocation/deallocation during the stepping loop as noticed by @esseivaju in #1172.

### DIFF
--- a/app/celer-sim/Transporter.cc
+++ b/app/celer-sim/Transporter.cc
@@ -88,6 +88,15 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries)
         result.alive.push_back(track_counts.alive);
     };
 
+    constexpr size_type min_alloc{65536};
+    result.initializers.reserve(std::min(min_alloc, max_steps_));
+    result.active.reserve(std::min(min_alloc, max_steps_));
+    result.alive.reserve(std::min(min_alloc, max_steps_));
+    if (store_step_times_)
+    {
+        result.step_times.reserve(std::min(min_alloc, max_steps_));
+    }
+
     // Abort cleanly for interrupt and user-defined signals
 #ifndef _WIN32
     ScopedSignalHandler interrupted{SIGINT, SIGUSR2};
--- a/app/celer-sim/Transporter.cc
+++ b/app/celer-sim/Transporter.cc
@@ -165,8 +165,7 @@ void Transporter<M>::accum_action_times(MapStrDouble* result) const
         CELER_ASSERT(action_ptrs.size() == times.size());
         for (auto i : range(action_ptrs.size()))
         {
-            auto&& label = action_ptrs[i]->label();
-            (*result)[label] += times[i];
+            (*result)[std::string{action_ptrs[i]->label()}] += times[i];
         }
     }
 }
--- a/src/accel/LocalTransporter.cc
+++ b/src/accel/LocalTransporter.cc
@@ -262,8 +262,7 @@ auto LocalTransporter::GetActionTime() const -> MapStrReal
         CELER_ASSERT(action_ptrs.size() == time.size());
         for (auto i : range(action_ptrs.size()))
         {
-            auto&& label = action_ptrs[i]->label();
-            result[label] = time[i];
+            result[std::string{action_ptrs[i]->label()}] = time[i];
         }
     }
     return result;
--- a/src/celeritas/global/ActionRegistry.cc
+++ b/src/celeritas/global/ActionRegistry.cc
@@ -49,7 +49,7 @@ void ActionRegistry::insert_const_impl(SPConstAction&& action)
  */
 void ActionRegistry::insert_impl(SPConstAction&& action)
 {
-    auto label = action->label();
+    auto label = std::string{action->label()};
     CELER_VALIDATE(!label.empty(), << "action label is empty");
 
     auto id = action->action_id();
--- a/src/corecel/io/OutputRegistry.cc
+++ b/src/corecel/io/OutputRegistry.cc
@@ -41,7 +41,7 @@ void OutputRegistry::insert(SPConstInterface interface)
 
     Category cat = interface->category();
     auto [prev, inserted]
-        = interfaces_[cat].insert({std::move(label), std::move(interface)});
+        = interfaces_[cat].insert({std::string{label}, std::move(interface)});
     CELER_VALIDATE(inserted,
                    << "duplicate output entry '" << prev->first
                    << "' for category '" << to_cstring(cat) << "'");
--- a/src/corecel/sys/ScopedProfiling.cuda.cc
+++ b/src/corecel/sys/ScopedProfiling.cuda.cc
@@ -37,7 +37,7 @@ nvtxDomainHandle_t domain_handle()
  *
  * Insert it if it doesn't already exist.
  */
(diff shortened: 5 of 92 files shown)


src/accel/LocalTransporter.cc
```diff
@@ -262,8 +262,7 @@ auto LocalTransporter::GetActionTime() const -> MapStrReal
         CELER_ASSERT(action_ptrs.size() == time.size());
         for (auto i : range(action_ptrs.size()))
         {
-            auto&& label = action_ptrs[i]->label();
-            result[label] = time[i];
+            result[std::string{action_ptrs[i]->label()}] = time[i];
```
esseivaju, line 265: We can change `MapStrReal` in `LocalTransporter` and `TimerOutput` to `std::unordered_map<std::string_view, real_type>`

src/celeritas/global/ActionRegistry.cc
```diff
@@ -49,7 +49,7 @@ void ActionRegistry::insert_const_impl(SPConstAction&& action)
  */
 void ActionRegistry::insert_impl(SPConstAction&& action)
 {
-    auto label = action->label();
+    auto label = std::string{action->label()};
```
esseivaju, line 52: same here,`string_view` as the key for the map, then `labels_.emplace_back(label)`

src/corecel/io/OutputRegistry.cc
```diff
@@ -41,7 +41,7 @@ void OutputRegistry::insert(SPConstInterface interface)
 
     Category cat = interface->category();
     auto [prev, inserted]
-        = interfaces_[cat].insert({std::move(label), std::move(interface)});
+        = interfaces_[cat].insert({std::string{label}, std::move(interface)});
```
esseivaju, line 44: we can also use a `string_view` as a key here

src/corecel/sys/ScopedProfiling.cuda.cc
```diff
@@ -37,31 +37,32 @@ nvtxDomainHandle_t domain_handle()
  *
  * Insert it if it doesn't already exist.
  */
-nvtxStringHandle_t message_handle_for(std::string const& message)
+nvtxStringHandle_t message_handle_for(std::string_view message)
 {
     static std::unordered_map<std::string, nvtxStringHandle_t> registry;
     static std::shared_mutex mutex;
 
+    std::string temp_message{message};
```
esseivaju, line 45: registry can also be `std::unordered_map<std::string_view, nvtxStringHandle_t>` and only instantiate the string when we register the domain.

test/celeritas/global/StepperTestBase.cc
```diff
@@ -72,16 +72,16 @@ auto StepperTestBase::check_setup() -> SetupCheckResult
 
     for (auto process_id : range(ProcessId{p.num_processes()}))
     {
-        result.processes.push_back(p.process(process_id)->label());
+        result.processes.push_back(std::string{p.process(process_id)->label()});
```
esseivaju: I think `emplace_back` looks better here (or vector of string_view)

test/celeritas/phys/Physics.test.cc
```diff
@@ -64,9 +64,9 @@ TEST_F(PhysicsParamsTest, accessors)
     std::vector<std::string> process_names;
     for (auto process_id : range(ProcessId{p.num_processes()}))
     {
-        process_names.push_back(p.process(process_id)->label());
+        process_names.push_back(std::string{p.process(process_id)->label()});
```
esseivaju: `emplace_back` here as well

esseivaju (commented):
Nice! Just a few comments.

src/accel/LocalTransporter.cc
sethrj: I don't think we can do it for `TimerOutput` because that might exceed the lifetime of the core params... I figured it was safer to do a few one-time string copies than to risk it...

src/corecel/sys/ScopedProfiling.cuda.cc
sethrj: I don't think it can, because some of the input strings are temporaries

src/accel/LocalTransporter.cc
esseivaju: So far we only use string literals that have static storage, but sure in general it's less error-prone for the map to own the key. It just doesn't look nice having to instantiate a string to look up the map when the string_view has the information needed. Gotta wait for C++20 support to fix that.

esseivaju (approved):
I reran the profiler for the one-slot scenario, we went from 93 million calls to free to 170k so it's pretty good.

app/celer-sim/Transporter.cc
```diff
@@ -165,8 +174,7 @@ void Transporter<M>::accum_action_times(MapStrDouble* result) const
         CELER_ASSERT(action_ptrs.size() == times.size());
         for (auto i : range(action_ptrs.size()))
         {
-            auto&& label = action_ptrs[i]->label();
-            (*result)[label] += times[i];
+            (*result)[std::string{action_ptrs[i]->label()}] += times[i];
```
pcanal, line 177: Doesn't that increase (keep the same) the number of allocation?

src/accel/LocalTransporter.cc
pcanal: Do you mean C++26? (which changes `operator[]`)?
Or should we use [`unordered_map::find`](https://en.cppreference.com/w/cpp/container/unordered_map/find) which is already (in C++14) templated on the parameter type?

src/corecel/io/OutputRegistry.cc
pcanal: Is the lifetime of `label` guaranteed to be longer than the lifetime of the`OutputRegistry`? (the `std::move` in the existing code suggest otherwise).

src/accel/LocalTransporter.cc
pcanal: I think you did mean C++20 as described here: https://www.cppstories.com/2021/heterogeneous-access-cpp20/.   I.e. the templated function/operator is not enough, one also has to customize the hash comparisons.

src/corecel/io/OutputRegistry.cc
sethrj: Actually that's a really good point. If you store a string view from an rvalue, you might be exceeding the lifetime of the object.

src/accel/LocalTransporter.cc
sethrj: Bummer, I saw that and hoped it would be automatic for the string/view case

app/celer-sim/Transporter.cc
sethrj: Yeah, but it should only happen once per stream per run — and usually it's going directly into a JSON object which has lots of allocations and can't accept `string_view` anyway.

sethrj:
Sorry, just noticed I didn't hit "send" on this comment yesterday...

---

@esseivaju I think it's much too dangerous to keep long-term string views. Even this is going to lead to an invalid read:
```c++
my_string_vec.emplace_back(label);
my_string_view_map.insert({my_string_vec.back(), value});
```
because once the vector changes capacity, the string views in the map are invalidated.

I think string views are meant primarily for interfaces and not for long-term storage.

PR #1212, opened 2024-04-29
Fix polar angle sampling and Mott factor in Wentzel distribution

amandalund:
This fixes what I think are a couple small errors in the Wentzel distribution:
- In the sampling of the scattering polar angle, looking through @hhollenb's (excellent!) [Coulomb scattering writeup](https://www.overleaf.com/read/pszrbsrkmfzc), I think the sampling function for mu (Eq. 2.10) should be
$$\mu = \mu_1 + \frac{(A + \mu_1) \xi (\mu_2 - \mu_1)}{A + \mu_2 - \xi (\mu_2 - \mu_1)}$$.
With $\mu = \frac{1}{2}(1 - \cos \theta)$ and $\mu_1 = \mu(0) = 0$, this gives
$$\cos \theta = 1 - \frac{2 A \xi \mu_2}{A + (1 - \xi) \mu_2}$$.
The sampled angle should now always be in [-1, 1), so we can remove the `clamp`.
- In both [Geant4](https://github.com/Geant4/geant4/blob/dda54bbdcfe0cc76177eaa03a657592aac0b1eb8/source/processes/electromagnetic/standard/src/G4WentzelOKandVIxSection.cc#L212) and the writeup (section 8) the Mott factor is given as $1 + 2\times 10^{-4} Z^2$. Updating this did get rid of an assertion failure in the rejection sampling where the `p_true` in the Bernoulli distribution was greater than one.

It would be good if @hhollenb could confirm that these changes are correct and I'm not missing something.


test/celeritas/em/CoulombScattering.test.cc
```diff
@@ -331,25 +310,38 @@ TEST_F(CoulombScatteringTest, simple_scattering)
...
+    static double const expected_energy[] = {200,
+                                             199.99999999847,
+                                             199.99999999994,
+                                             199.99999999953,
+                                             200,
+                                             199.99999999673,
+                                             199.99999999935,
+                                             199.99999999572,
+                                             200,
+                                             199.99999999943};
```
sethrj: Maybe we should be testing this at a lower energy (or a range thereof?) With this small a change in direction and energy would we be able to tell if there were a bug?

amandalund: Good point, a range of energies would be better.

sethrj (approved):
Nice find! Good to justify our general avoidance of `clamp` and its ilk...

sethrj:
Darn it, I guess we'll never replace multiple scattering with single coulomb at this rate...

I'm happy either to wait for @hhollenb 's double check before merging this or to let you go ahead and merge yourself.

hhollenb:
This looks great, and may fix some of the small discrepancies in the scattering angle.

I used clamp out of a concern for other parts of the code requiring $|\cos\theta| \leq 1$ and there being small rounding errors. More than happy to replace it with an ensure post-condition.

I'll approve it and merge it into the tests I have right now. Finding that source for @amandalund coulomb limit made me want to revisit the weird false scattering code and make sure its assumptions are correct.

hhollenb (approved):
I'll revisit the Coulomb scattering limit with Soon once he returns and I review the paper.

PR #830, opened 2023-06-27
Fix build errors from Geant4@10

drbenmorgan:
As reported in #828, the latest `develop` has a few compilation errors against Geant4 10.6 due to missing accessors. These are worked around here through use of the `G4VERSION_NUMBER` macro and direct/intermediate implementations:

- Missing `G4UIcommand::ConvertToLongInt()` is fixed by directly using `std::istringstream` as in the Geant4 implementation available from 10.7.
- Missing `GetTheta()`, `GetPhi()` accessors in `G4Trap` and `G4Para` calculate these parameters as in Geant4 11.0, using `GetSymAxis()` to obtain the needed inputs.
- Missing `GetAlpha...()` accessors in `G4Trap` simply use the `GetTanAlpha...()` accessors and `std::atan`.

The use of the direct Geant4 accessors is retained if the version supports them. Both implementations have been tested against Geant4 11, and produce the same results. Compilation against 10.6 works, but a few tests still fail:

```
The following tests FAILED:
	 81 - celeritas/ext/GeantImporter:FourSteelSlabs* (Failed)
	 83 - celeritas/ext/GeantImporter:OneSteelSphere.* (Failed)
	 98 - celeritas/global/AlongStep:Em3AlongStepTest.nofluct_nomsc (Failed)
	100 - celeritas/global/AlongStep:Em3AlongStepTest.msc_nofluct_finegrid (Failed)
	111 - celeritas/global/Stepper:TestEm3MscNofluct.* (Failed)
	151 - celeritas/user/Diagnostic:TestEm3* (Failed)
```

These need more investigation but on first look appear due to small (but above tolerance) differences in imported data, which is perhaps not surprising. I didn't look at trying to address these here until the compilation error fixes are reviewed and merged.

PR #1213, opened 2024-04-29
Add ORANGE converters for G4Trap and G4GenericTrap

mrguilima:
The new ORANGE converters for G4Trap and G4GenericTrap are based on the recently implemented generic trap shape.

G4Trap corners are needed, and not available from the public G4Trap interface, hence they are re-calculated as needed for the generic trap construction.


src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,55 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
...
+    }
+    real_type hh = scale_(solid.GetZHalfLength());
+
+    return make_shape<GenTrap>(solid, hh, lower, upper);
+}
+
+//---------------------------------------------------------------------------//
+
+//! Convert a trapezoid
+auto SolidConverter::trap(arg_type solid_base) -> result_type
```
sethrj: Could you please move this function back between `torus` and `trd` so it remains in alphabetical order?

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -195,6 +195,93 @@ TEST_F(SolidConverterTest, displaced)
...
+            R"json({"_type":"shape","interior":{"_type":"gentrap",
+            "halfheight":4.0,
+            "lower":[[1.1,-2.0],[1.1,2.0],[-0.9,2.0],[-0.9,-2.0]],
+            "upper":[[3.1,-4.0],[3.1,4.0],[-2.9,4.0],[-2.9,-4.0]]},
+            "label":"trap_GenTrap"})json",
+            {{-1, -2, -4 - 1.e-6}, {-1, -2, -3}, {0.5, 1, 3}, {1, 1, 3}});
+    }
+
+    // TODO: most generic gentrap with twisted side faces
+    /*
```
sethrj: For "todo" code chunks like this (they should rarely be allowed but I think this is a reasonable case), enclose them with `#if 0` and `#endif`.

Better yet would be to write `EXPECT_THROW(this->build_and..., celeritas::RuntimeError);` since for now we should get a "not implemented" when converting.

src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,55 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
...
+{
+    auto const& solid = dynamic_cast<G4Trap const&>(solid_base);
+
+    real_type sinphi, cosphi;
+    sincos(solid.GetPhi(), &sinphi, &cosphi);
+    auto tanTheta = std::tan(solid.GetTheta());
```
sethrj: Regarding the build error: see https://github.com/celeritas-project/celeritas/pull/830 . I think for now (until we discuss https://github.com/celeritas-project/celeritas/issues/1132 ) we can just apply @drbenmorgan 's fix to this piece of code.

Also please follow the style guide and use `tan_theta`.

src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,55 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
...
+
+    auto hz = scale_(solid.GetZHalfLength());
+    auto hy1 = scale_(solid.GetYHalfLength1());
+    auto hy2 = scale_(solid.GetYHalfLength2());
+    auto hx1 = scale_(solid.GetXHalfLength1());
+    auto hx2 = scale_(solid.GetXHalfLength2());
+    auto hx3 = scale_(solid.GetXHalfLength3());
+    auto hx4 = scale_(solid.GetXHalfLength4());
+    auto dxdzDz = hz * tanTheta * cosphi;
+    auto dydzDz = hz * tanTheta * sinphi;
```
sethrj: I think some comments would help clarify the meaning of the 1,2 in y versus 1,2,3,4 in x. Also is there a significance in the capital `D` here? Please adhere to the variable naming guide.

src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,55 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
...
+    auto dxdyDy1 = scale_(solid.GetTanAlpha1() * solid.GetYHalfLength1());
+    auto dxdyDy2 = scale_(solid.GetTanAlpha2() * solid.GetYHalfLength2());
```
sethrj: I think it's clearer to reuse the locals `hy1` and `hy2` instead of re-fetching it and scalingn it:
```suggestion
    auto dxdyDy1 = y1 * solid.GetTanAlpha1();
    auto dxdyDy2 = y2 * solid.GetTanAlpha2();
```

sethrj (changes requested):
Looking good Guilherme! A few comments.

test/orange/g4org/SolidConverter.test.cc
mrguilima: I've added the EXPECT_THROW() suggestion.
Thanks for pointing that out.

src/orange/g4org/SolidConverter.cc
mrguilima: I have added documentation, referencing the Geant4 guide for CSG shapes, and renamed variables replacing D with _h for adherence to the naming guide.

mrguilima:
I've applied all suggestions.

src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,19 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
 auto SolidConverter::generictrap(arg_type solid_base) -> result_type
 {
     auto const& solid = dynamic_cast<G4GenericTrap const&>(solid_base);
-    CELER_DISCARD(solid);
-    CELER_NOT_IMPLEMENTED("generictrap");
+
+    auto const vtx = solid.GetVertices();
```
sethrj:
```suggestion
    auto const& vtx = solid.GetVertices();
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -336,8 +336,19 @@ auto SolidConverter::genericpolycone(arg_type solid_base) -> result_type
...
+
+    std::vector<GenTrap::Real2> lower(4), upper(4);
+    for (auto i : range(4))
+    {
+        lower[i] = scale_.to<Cone::Real2>(vtx[i].x(), vtx[i].y());
+        upper[i] = scale_.to<Cone::Real2>(vtx[i + 4].x(), vtx[i + 4].y());
+    }
+    real_type hh = scale_(solid.GetZHalfLength());
+
+    return make_shape<GenTrap>(solid, hh, lower, upper);
```
sethrj:
```suggestion
    return make_shape<GenTrap>(solid, hh, std::move(lower), std::move(upper));
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -569,12 +580,74 @@ auto SolidConverter::torus(arg_type solid_base) -> result_type
 }
 
 //---------------------------------------------------------------------------//
-//! Convert a generic trapezoid
+/** Convert a trapezoid.
```
sethrj:
```suggestion
/*!
 * Convert a trapezoid.
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -569,12 +580,74 @@ auto SolidConverter::torus(arg_type solid_base) -> result_type
...
     auto const& solid = dynamic_cast<G4Trap const&>(solid_base);
-    CELER_DISCARD(solid);
-    CELER_NOT_IMPLEMENTED("trap");
+
+    real_type sin_phi{}, cos_phi{}, tan_theta{};
+#if G4VERSION_NUMBER < 1100
+    // Geant4 10.7 and earlier - axis = (sinth.cosphi, sinth.sinphi, costh)
+    // Note: both sinth,costh >= 0 since theta is in [0, pi/2] for a G4Trap
+    auto axis = solid.GetSymAxis();
+    auto sin_theta = std::sqrt(real_type(1.0) - axis.z() * axis.z());
```
sethrj, line 617: Since all the Geant4 data are doubles, let's leave these as doubles too
```suggestion
    double sin_phi{}, cos_phi{}, tan_theta{};
#if G4VERSION_NUMBER < 1100
    // Geant4 10.7 and earlier - axis = (sinth.cosphi, sinth.sinphi, costh)
    // Note: both sinth,costh >= 0 since theta is in [0, pi/2] for a G4Trap
    auto const& axis = solid.GetSymAxis();
    auto sin_theta = std::sqrt(1.0 - axis.z() * axis.z());
```

sethrj (commented):
A couple of minor final comments. Thanks for the updates!

sethrj (approved):
Nice work! Thanks @mrguilima !

Issue #1211, opened 2024-04-29
Better statistics from celer-sim


sethrj:
The alive/initializers counts are nice to have for GPU but usually way too big to be useful for CPU (and will be totally useless if we move toward a single-particle mode). It would be great to "always" have a few of the basic diagnostics used by the regression suite:
- total number of step iterations
- total number of steps (sum of active tracks, amount of actual work done)
- maximum track initializer count (to inform the buffer size for the next run)
- aborted (aka unconverged, the number of alive plus initializers at the end of the simulation)

These should be small and cheap to tally (probably in the Stepper?) Maybe make it a new `Output` class so that it can be used as a "result" and shared between celer-sim and celer-g4?

PR #1214, opened 2024-04-30
Add always-on basic diagnostics to celer-sim

amandalund:
The diagnostics for #1211 are easy to add to our existing runner output, but I can try to think of how we could refactor this to work with celer-g4 as well if you'd like.


sethrj (approved):
So quick! Thanks @amandalund . I think we can refactor to be more "general" in a follow-on. I more urgently want to backport this to 0.4.x and compare compilers.

PR #1215, opened 2024-04-30
Fix library macro compatibility wrappers

sethrj:
This fixes the compatibility wrappers introduced in #1104 so that a downstream user code targeting Celeritas 0.4 will still work with 0.5.


stognini (approved):

stognini:
Looks good. Thanks!

PR #1217, opened 2024-05-01
Split optical pre-generator action into separate Cerenkov and scintillation actions

amandalund:
This splits the action to generate optical distribution data into separate actions for Cerenkov and scintillation so things will still work if only one of the processes is enabled (this is also now tested).


src/celeritas/optical/detail/ScintPreGenExecutor.hh
```diff
@@ -0,0 +1,83 @@
...
+    scintillation_dist = {};
+
+    auto sim = track.make_sim_view();
+    bool inactive = sim.status() == TrackStatus::inactive;
+    auto optmat_id = inactive ? OpticalMaterialId{}
+                              : track.make_material_view()
+                                    .make_material_view()
+                                    .optical_material_id();
+
+    if (inactive || !optmat_id || sim.step_length() == 0)
```
sethrj: This semi-wacky code is now repeated across the two pre-executors... maybe a helper function would be good:
```c++
inline CELER_FUNCTION OpticalMaterialId get_optical_material(
    CoreTrackView const& track)
{
    if (track.make_sim_view().status() == TrackStatus::inactive)
        return {};

    return track.make_material_view()
        .make_material_view()
        .optical_material_id();
}
```
and `(inactive || !optmat_id` would just become `(!optmat_id`

sethrj (approved):
Looks good to me. There's not *too* much code duplication and I don't think templating would save much.

src/celeritas/optical/detail/ScintPreGenExecutor.hh
amandalund: Yeah, good call.

PR #1218, opened 2024-05-02
Add ORANGE converter for G4Trd

mrguilima:
The new ORANGE converter for G4Trd is based on the generic trap shape.

G4Trd corners are needed, and not directly available from the public G4Trap interface, hence they are re-calculated as needed for the generic trap construction.


src/orange/g4org/SolidConverter.cc
```diff
@@ -656,8 +656,24 @@ auto SolidConverter::trap(arg_type solid_base) -> result_type
...
+    lower[0] = GenTrap::Real2{-hx1, -hy1};
+    lower[1] = GenTrap::Real2{+hx1, -hy1};
+    lower[2] = GenTrap::Real2{+hx1, +hy1};
+    lower[3] = GenTrap::Real2{-hx1, +hy1};
+    upper[0] = GenTrap::Real2{-hx2, -hy2};
+    upper[1] = GenTrap::Real2{+hx2, -hy2};
+    upper[2] = GenTrap::Real2{+hx2, +hy2};
+    upper[3] = GenTrap::Real2{-hx2, +hy2};
+
+    return make_shape<GenTrap>(solid, hz, lower, upper);
```
sethrj: I'd prefer (for the sake of reusability and testability) for the work of constructing these points be moved into a `GenTrap` static helper function that takes the half-height, the lower half-widths, and the upper half-widths and returns a `GenTrap`:

```suggestion
    return make_shape<GenTrap>(solid, GenTrap::from_rectangles(hz,{hx1, hy1}, {hx2, hy2}));
```

and add a test to `ConvexRegions.test.cc` with this constructor

sethrj (changes requested):
Thanks Guilherme! Reusing the GenTrap makes sense. Should we rename `GenTrap` to `Trapezoid` then since we aren't going to implement a more specific one?

src/orange/g4org/SolidConverter.cc
mrguilima: GenTrap in Geant4 potentially has twisted sides!  It is an extension of the standard trap, which contains only flat sides.

sethrj: Yep! I guess my point is that this `GenTrap` convex region is now general enough to encompass several object types... "generic trapezoid" with flat faces is the first one; the code here removes the need for an "optimized" `Trd` shape. And honestly the way it's implemented, it could be any sort of generic polygon with flat top and bottom...

mrguilima: Regarding your suggestion to write more constructors with different algorithms, do you intend to map Geant4 constructors one-to-one? 

For instance, your proposal above (in green, line 666) is appropriate for the G4Trd, which has [this constructor](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/geometry/solids/CSG/include/G4Trd.hh#L66).  But your helper function is only useful for the TRD.  So are you proposing to implement one helper function per G4 trap-like constructor?

Then we'll need another helper function for the [full G4Trap constructor](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/geometry/solids/CSG/include/G4Trap.hh#L114), with 3 angles _in addition_ to 7 half-lengths (hz, hy1, hy2, hx1, hx2, hx3, hx4) -- is that the idea?

sethrj: Aha, sorry I missed that there are three different trapezoids. Yes I think it will improve testing and reuse to have a distinct constructor for each of the cases.

sethrj: (But to clarify: we don't need a constructor for every Geant4 constructor: was just want one for each of the trapezoid "types".)

sethrj: Oh wait, G4Trap is the "genTrap". Maybe to avoid further confusion we should meet briefly on zoom...

mrguilima: I will defer the full G4Trap constructor for later, after the twisted planes are implemented.

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,31 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
...
+    CELER_VALIDATE(hz_ > 0, << "nonpositive halfheight: " << hz_);
+
+    lo_[0] = GenTrap::Real2{-hxy1[0], -hxy1[1]};
+    lo_[1] = GenTrap::Real2{+hxy1[0], -hxy1[1]};
+    lo_[2] = GenTrap::Real2{+hxy1[0], +hxy1[1]};
+    lo_[3] = GenTrap::Real2{-hxy1[0], +hxy1[1]};
+    hi_[0] = GenTrap::Real2{-hxy2[0], -hxy2[1]};
+    hi_[1] = GenTrap::Real2{+hxy2[0], -hxy2[1]};
+    hi_[2] = GenTrap::Real2{+hxy2[0], +hxy2[1]};
+    hi_[3] = GenTrap::Real2{-hxy2[0], +hxy2[1]};
```
sethrj: Since this code is inside the class now, you don't have to do `GenTrap::`, and in fact you don't even need the `::Real2` since it's explicit.

I think the code would be a bit safer if instead of sizing it on 376 and writing directly to the elements, you leave off the `lo_` and `hi_` and write this as an initializer list:
```suggestion
    lo_ = {{-hxy1[0], -hxy1[1]},
           {+hxy1[0], -hxy1[1]},
           {+hxy1[0], +hxy1[1]},
           {-hxy1[0], +hxy1[1]}};
    hi_ = {{-hxy2[0], -hxy2[1]},
           {+hxy2[0], -hxy2[1]},
           {+hxy2[0], +hxy2[1]},
           {-hxy2[0], +hxy2[1]}};
```

I'm actually not even sure if `, lo_{4}, hi_{4}` in your constructor is *resizing* it to 4 or *initializing* it with an initializer list... the fact that `vector` can be constructed with `std::initializer_list` changes the semantics in weird ways...

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,31 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
     }
 }
 
+/*!
```
```suggestion
//---------------------------------------------------------------------------//
/*!
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -609,55 +609,56 @@ auto SolidConverter::trap(arg_type solid_base) -> result_type
...
 
-    real_type sin_phi{}, cos_phi{}, tan_theta{};
+    real_type tan_theta{};
+    Turn phi{};
 #if G4VERSION_NUMBER < 1100
     // Geant4 10.7 and earlier - axis = (sinth.cosphi, sinth.sinphi, costh)
     // Note: both sinth,costh >= 0 since theta is in [0, pi/2] for a G4Trap
     auto axis = solid.GetSymAxis();
-    auto sin_theta = std::sqrt(real_type(1.0) - axis.z() * axis.z());
+    auto sin_theta = std::sqrt(1.0 - axis.x() * axis.x() + axis.y() * axis.y());
```
sethrj: Might want to put a `max` in here in case of floating point roundoff, and use `ipow`
```suggestion
    auto sin_theta = std::sqrt(max(0.0, 1.0 - ipow<2>(axis.x()) + ipow<2>(axis.y()));
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,75 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
...
+    CELER_VALIDATE(lo[1] > 0, << "nonpositive lower y half-edge: " << lo[1]);
+    CELER_VALIDATE(hi[1] > 0, << "nonpositive upper y half-edge: " << hi[1]);
+    CELER_VALIDATE(halfz > 0, << "nonpositive half-height: " << halfz);
+
+    VecReal2 lower
+        = {{-lo[0], -lo[1]}, {lo[0], -lo[1]}, {lo[0], lo[1]}, {-lo[0], lo[1]}};
+    VecReal2 upper
+        = {{-hi[0], -hi[1]}, {hi[0], -hi[1]}, {hi[0], hi[1]}, {-hi[0], hi[1]}};
+
+    return std::move(GenTrap{halfz, std::move(lower), std::move(upper)});
```
sethrj: To fix the build error (you don't need to `move` when returning a local variable or constructing an rvalue):

```suggestion
    return GenTrap{halfz, std::move(lower), std::move(upper)};
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,75 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
...
+ */
+GenTrap GenTrap::from_trd(real_type halfz, Real2 const& lo, Real2 const& hi)
+{
+    CELER_VALIDATE(lo[0] > 0, << "nonpositive lower x half-edge: " << lo[0]);
+    CELER_VALIDATE(hi[0] > 0, << "nonpositive upper x half-edge: " << hi[0]);
+    CELER_VALIDATE(lo[1] > 0, << "nonpositive lower y half-edge: " << lo[1]);
+    CELER_VALIDATE(hi[1] > 0, << "nonpositive upper y half-edge: " << hi[1]);
+    CELER_VALIDATE(halfz > 0, << "nonpositive half-height: " << halfz);
+
+    VecReal2 lower
```
sethrj, line 331:
```suggestion
    // Construct points counterclockwise from lower left
    VecReal2 lower
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,75 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
...
+    VecReal2 upper
+        = {{-hi[0], -hi[1]}, {hi[0], -hi[1]}, {hi[0], hi[1]}, {-hi[0], hi[1]}};
+
+    return std::move(GenTrap{halfz, std::move(lower), std::move(upper)});
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a general trap
+ */
```
```suggestion
```

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -368,6 +368,75 @@ GenTrap::GenTrap(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
...
+    VecReal2 lower = {{-dxdz_hz - dxdy_hy1 - hx1, -dydz_hz - hy1},
+                      {-dxdz_hz - dxdy_hy1 + hx1, -dydz_hz - hy1},
+                      {-dxdz_hz + dxdy_hy1 + hx2, -dydz_hz + hy1},
+                      {-dxdz_hz + dxdy_hy1 - hx2, -dydz_hz + hy1}};
+    VecReal2 upper = {{+dxdz_hz - dxdy_hy2 - hx3, +dydz_hz - hy2},
+                      {+dxdz_hz - dxdy_hy2 + hx3, +dydz_hz - hy2},
+                      {+dxdz_hz + dxdy_hy2 + hx4, +dydz_hz + hy2},
+                      {+dxdz_hz + dxdy_hy2 - hx4, +dydz_hz + hy2}};
+
+    return std::move(GenTrap{hz, std::move(lower), std::move(upper)});
```
```suggestion
    return GenTrap{hz, std::move(lower), std::move(upper)};
```

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -216,10 +216,48 @@ class GenTrap final : public ConvexRegionInterface
     using VecReal2 = std::vector<Real2>;
     //!@}
 
+    struct TrapFace
```
sethrj, line 220:
```suggestion
    //! Construction argument for building from regular trapezoidal top/bottom polygons
    struct TrapFace
```

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -216,10 +216,48 @@ class GenTrap final : public ConvexRegionInterface
...
+    {
+        real_type hy_{};  // vertical distance between horizontal edges
+        real_type hx_lo_{};  // top horizontal edge
+        real_type hx_hi_{};  // botton horizontal edge
```
sethrj: Since these are public data, omit the trailing `_`, and also use doxygen markup

```suggestion
        real_type hy_{};  //!< half the vertical distance between horizontal edges
        real_type hx_lo_{};  //!< top horizontal edge half-length
        real_type hx_hi_{};  //!< bottom horizontal edge half-length
```

src/orange/orangeinp/ConvexRegion.hh
```diff
@@ -216,10 +216,48 @@ class GenTrap final : public ConvexRegionInterface
...
+                 real_type tan_alpha)
+            : hy_{hy}, hx_lo_{hx_lo}, hx_hi_{hx_hi}, tan_alpha_{tan_alpha}
+        {
+            CELER_VALIDATE(hx_lo > 0,
+                           << ": nonpositive lower x half-edge: " << hx_lo);
+            CELER_VALIDATE(hx_hi > 0,
+                           << ": nonpositive lower x half-edge: " << hx_hi);
+            CELER_VALIDATE(hy > 0,
+                           << ": nonpositive y half-projection: " << hy);
+        }
```
sethrj: Instead of validating this in the constructor (which doesn't catch errors if you modify the public data after the fact!), either declare an `explicit operator bool()` (and then `CELER_VALIDATE(lo, ...);` in `GenTrap::from_trap`; or define a `void validate() const;` or (maybe best?) simply put the `CELER_VALIDATE` in the `GenTrap::from_trap` (looping over `for (TrapFace const* tf : {&lo, &hi})`)

sethrj (changes requested):
Thanks @mrguilima ! Hopefully this one last round and we'll have our Trds and Traps 😄

src/orange/g4org/SolidConverter.cc
mrguilima: Done, with a correction: should be sqrt(1.0 - ipow<2>(axis.z())).

sethrj (approved):
I made some final changes (which I'd already mostly done but just stashed) to fix the failing test. Thanks!

PR #1219, opened 2024-05-02
Fix minor dependency issues

sethrj:
We've encountered several instances in the past (usually with nlohmann/json) where the CI builds pass but individual builds fail, because the CI uses spack views so that including one dependency will propagate include paths for all of them. This fixes such an issue with JSON ~~and changes the "spack" CI so that it loads dependencies from separate installation directories~~ (no spack changes for now since it's not straightforward like I thought). It also should fix the ability to link against a Geant4 build directory, which @whokion noticed was broken.


mrguilima (approved):
The errors I reported earlier are fixed.
Thanks!

PR #1220, opened 2024-05-07
Add CPU track parallelism config option

esseivaju:
Add a CMake option to disable track-level parallelism in the CPU ActionLauncher as this significantly slows down execution when using one track slot. A few other `pragma omp parallel for` in the codebase should also be investigated, i.e. should we also turn them off with the same CMake option?

I also bundled an AlignedAllocator implementation. It's not used yet so it could be removed, but this can be useful to guarantee that allocation falls on a cache line boundary.


sethrj:
Arg you didn't see the other commits I had on the "improve the timer" branch... https://github.com/sethrj/celeritas/tree/single-thread-qol

esseivaju:
Ah, yes I didn't see the other commits. Feel free to close that PR

PR #1056, opened 2023-12-08
Warn and set default threads when OpenMP and Geant4 MT might collide

sethrj:
If using an OpenMP-enabled Celeritas with a multithreaded Geant4 application, (OMP_NUM_THREADS)*(Geant4 num threads) actual threads will be spawned without it being obvious. Generally it would be best to avoid building OpenMP when used with multithreaded Geant4, but it is convenient to be able to reuse the same library/executable for celer-sim and celer-g4.

This prints a warning if CPU+OpenMP+G4MT is in use; it will set the thread count to 1 if the `OMP_NUM_THREADS` environment variable isn't set (otherwise it defaults to "maximum threads").

@esseivaju or @amandalund could you please verify that your standalone tests show the "correct" behavior of 1 Celeritas thread to Geant4 thread when this is in use?

PR #1222, opened 2024-05-07
Add configure-time `CELERITAS_OPENMP` switch to change threading

sethrj:
Event-level OpenMP was added in #774 to more accurately represent the behavior of Geant4 integration. Unfortunately we found that OpenMP isn't up to the task of efficiently threading both track-parallel and event-parallel (see also #1073). This adds a configure option to switch between track-level parallelism (inside a kernel) and event-level (for now, only in celer-sim). Since track-level parallelism isn't ideal for CPU (striding, integration, etc.) the default is set to *event*-level, and furthermore a warning is emitted at configure time if track-level parallelism is enabled alongside Geant4 MT (see #1056).

Supersedes #1220


src/celeritas/track/detail/TrackSortUtils.cc
```diff
@@ -114,7 +114,9 @@ void count_tracks_per_action(
     std::fill(offsets.begin(), offsets.end(), ThreadId{});
     auto const size = states.size();
     // if get_action(0) != get_action(1), get_action(0) never gets initialized
-#pragma omp parallel for
+#if CELERITAS_OPENMP == CELERITAS_OPENMP_THREAD
```
amandalund:
```suggestion
#if CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
```

src/corecel/math/Atomics.hh
```diff
@@ -28,7 +28,7 @@ CELER_FORCEINLINE_FUNCTION T atomic_add(T* address, T value)
 #else
     CELER_EXPECT(address);
     T initial;
-#    ifdef _OPENMP
+#    if CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
```
amandalund: Do we need to keep the `#ifdef _OPENMP` to suppress the unknown pragmas warnings?
```
                 from /home/alund/celeritas_project/celeritas/test/corecel/data/StackAllocator.test.cc:8:
/home/alund/celeritas_project/celeritas/src/corecel/math/Atomics.hh:33: warning: ignoring ‘#pragma omp atomic’ [-Wunknown-pragmas]
   33 | #        pragma omp atomic capture
      | 
/home/alund/celeritas_project/celeritas/src/corecel/math/Atomics.hh:83: warning: ignoring ‘#pragma omp atomic’ [-Wunknown-pragmas]
   83 | #        pragma omp atomic capture
      | 
/home/alund/celeritas_project/celeritas/src/corecel/math/Atomics.hh:106: warning: ignoring ‘#pragma omp atomic’ [-Wunknown-pragmas]
  106 | #        pragma omp atomic capture
...
```

amandalund (commented):
Nice, looks good to me.

src/corecel/math/Atomics.hh
sethrj: Hmmm actually this points out that we're only "atomic" if OpenMP is enabled for a translation unit... that's a little surprising. If we forget to link the OpenMP target in one module and access the "atomic" data it'll be non-atomic. I can't think where this would happen in real life though...

src/celeritas/global/ActionLauncher.hh
```diff
@@ -34,7 +35,7 @@ void launch_action(ExplicitActionInterface const& action,
                    F&& execute_thread)
 {
     MultiExceptionHandler capture_exception;
-#ifdef _OPENMP
+#if CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
```
amandalund:
```suggestion
#if defined(_OPENMP) && CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
```
Here too:
```
In file included from /home/alund/celeritas_project/celeritas/test/celeritas/track/MockInteractAction.cc:16:
/home/alund/celeritas_project/celeritas/src/celeritas/global/ActionLauncher.hh:39: warning: ignoring ‘#pragma omp parallel’ [-Wunknown-pragmas]
   39 | #    pragma omp parallel for
```

amandalund (approved):

PR #1223, opened 2024-05-07
Write celer-sim results as 'null' when they're disabled

sethrj:
This writes `null` for track counters and step times if they're disabled. It also removes the call to `std::chrono::high_resolution_clock::now()` when step timers are disabled (see #1172).


app/celer-sim/simple-driver.py
```diff
@@ -141,6 +142,12 @@ def strtobool(text):
     json.dump(j, f, indent=1)
 print("Results written to", outfilename, file=stderr)
 
-time = j['result']['runner']['time'].copy()
-time.pop('steps')
+run_output =j['result']['runner']
+time = run_output['time'].copy()
+steps = time.pop('steps')
+if use_device:
+    assert(len(steps) == run_output.num_steps[0])
```
amandalund:
```suggestion
    assert(len(steps[0]) == run_output['num_step_iterations'][0])
```

amandalund (approved):
Looks good @sethrj!

app/celer-sim/simple-driver.py
sethrj: Clearly I'm too used to the Jenkins job failing...

sethrj:
https://cloud.cees.ornl.gov/jenkins-ci/job/celeritas/job/PR-1223/lastBuild/testReport/

app/celer-sim/simple-driver.py
```diff
@@ -141,6 +142,12 @@ def strtobool(text):
     json.dump(j, f, indent=1)
 print("Results written to", outfilename, file=stderr)
 
-time = j['result']['runner']['time'].copy()
-time.pop('steps')
+run_output =j['result']['runner']
+time = run_output['time'].copy()
+steps = time.pop('steps')
+if use_device:
+    assert(len(steps) == run_output['num_step_iterations'][0])
```
amandalund, line 149: Shouldn't it be `len(steps[0])` since there's an array of step times for each event?

sethrj: Damn it why didn't I just hit "commit suggestion" 😩

PR #1225, opened 2024-05-07
Propagate parent bounding boxes to ensure objects have finite extents

sethrj:
The BIH cost function fails in the presence of semi-infinite objects, which can happen given certain shapes. In theory, though, if the global boundary is finite, every bounding box should be finite. In fact we should be able to use this to determine strict bounding boxes on general combinatorial quadric surfaces using Millman 2014[^1].

With this change:
- Global boundaries are validated to be finite (e.g. you can't use a negated sphere or a shape that doesn't yet have a strict bbox implementation)
- Daughter universe bounding boxes start off as null but are *expanded* by their placement in parent volumes (using a parent-to-daughter transformation of the bounding box)
- Objects being constructed take an "initial" bounding box (the unit's boundary) so that they're no larger than that box

[^1]: Millman, David L., David P. Griesheimer, Brian R. Nease, and Jack Snoeyink. “Computing Numerically-Optimal Bounding Boxes for Constructive Solid Geometry (CSG) Components in Monte Carlo Particle Transport Calculations.” In SNA + MC 2013 - Joint International Conference on Supercomputing in Nuclear Applications + Monte Carlo, edited by D. Caruge, C. Calvin, C.M. Diop, F. Malvagi, and J.-C. Trama, 02506. Paris, France: EDP Sciences, 2014. https://doi.org/10.1051/snamc/201402506.


src/orange/detail/BIHBuilder.hh
```diff
@@ -53,36 +54,42 @@ class BIHBuilder
...
 
-    VecBBox bboxes_;
-    VecReal3 centers_;
-    Storage* storage_;
+    Temporaries temp_;
+
+    CollectionBuilder<FastBBox> bboxes_;
+    CollectionBuilder<LocalVolumeId> local_volume_ids_;
+    CollectionBuilder<BIHInnerNode> inner_nodes_;
+    CollectionBuilder<BIHLeafNode> leaf_nodes_;
```
sethrj, line 82: To be more consistent with the other builders, I've made `foo_` a builder that targets `data.foo`. To avoid the conflict with the temporary class data bboxes/centers I put those in a custom class.

Later we can turn these `CollectionBuilders` to `DedupeCollectionBuilders`.

src/orange/detail/BIHBuilder.cc
```diff
@@ -18,59 +18,68 @@ namespace detail
...
         {
             // Null bbox (background volume) is unreachable by volume
             // initialization
         }
-        else if (is_infinite(bboxes_[i]))
+        else if (is_infinite(temp_.bboxes[i]))
         {
+            // Infinite in *every* direction
+            // TODO: make an exception for "EXTERIOR" volume and remove the
+            // "infinite volume" exceptions?
```
sethrj, line 64: After this refactor, *only* the exterior (well, and the stuff coming from SCALE for the moment) should have an infinite volume.

src/orange/transform/VariantTransform.cc
```diff
@@ -97,5 +97,18 @@ apply_transform(VariantTransform const& transform, BBox const& bbox)
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the inverse of a transform.
+ */
+[[nodiscard]] VariantTransform calc_inverse(VariantTransform const& transform)
+{
+    CELER_ASSUME(!transform.valueless_by_exception());
+    // Dispatch to bounding box utils or "monostate" case above
```
sethrj:
```suggestion
```

src/orange/detail/BIHPartitioner.cc
```diff
@@ -152,9 +152,13 @@ BIHPartitioner::make_partition(VecIndices const& indices,
         }
     }
 
+    CELER_ASSERT(!p.indices[Edge::left].empty());
```
elliottbiondo, line 154: Minor point, but could you also change the `CELER_EXPECT(!indices.empty());` at the beginning of this function and at the beginning of `operator()` to `CELER_EXPECT(indices.size()>1);`. I think I should have been more explicit at indicating how we arrive a leaf nodes.

src/orange/BoundingBoxUtils.hh
```diff
@@ -27,17 +27,33 @@ class Transformation;
...
+            values.begin(), values.end(), [rhs](T lhs) { return lhs == rhs; });
+    };
+    constexpr T inf = numeric_limits<T>::infinity();
+
+    return all_equal(bbox.lower(), -inf) && all_equal(bbox.upper(), inf);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Check if a bounding box is finite.
```
elliottbiondo: "is finite in every direction", just be explicit and also consistent with the `is_infinite` doc string

elliottbiondo (approved):

src/orange/detail/BIHPartitioner.cc
sethrj: Alas but it fails:
```
[ RUN      ] BIHBuilderTest.basic
unknown file: Failure
C++ exception with description "/Users/seth/Code/celeritas-temp/src/orange/detail/BIHPartitioner.cc:56:
celeritas: precondition failed: indices.size() > 1" thrown in the test body.

[  FAILED  ] BIHBuilderTest.basic (0 ms)
```

sethrj: but the `make_partition` check update is good.

PR #1227, opened 2024-05-08
Fix trapezoid construction for negative phi and older Geant4

sethrj:
Thanks to @mrguilima for pointing out that there was an error in the calculation.

Also fixes https://github.com/celeritas-project/celeritas/pull/1223#discussion_r1592931865 .


app/celer-sim/simple-driver.py
```diff
@@ -146,8 +146,9 @@ def strtobool(text):
 time = run_output['time'].copy()
 steps = time.pop('steps')
 if use_device:
-    assert(len(steps) == run_output['num_step_iterations'][0])
+    assert len(steps) == 1
```
amandalund: I think it's 3 😉

sethrj: 🤪 how many times can I get this wrong!!

sethrj: But I thought with "merge events" it would only have 1? The dimension being number of streams...

amandalund: Isn't `merge_events` false in the test?

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -356,9 +356,9 @@ GenTrap GenTrap::from_trap(
     CELER_VALIDATE(theta >= zero_quantity() && theta < Turn{0.25},
                    << "invalid angle " << theta.value()
                    << " [turns]: must be in the range [0, 0.25)");
-    CELER_VALIDATE(phi >= zero_quantity() && phi < Turn{1.},
+    CELER_VALIDATE(phi > Turn{-1} && phi < Turn{1},
```
mrguilima: This is probably enough:
```suggestion
    CELER_VALIDATE(phi > Turn{-0.5} && phi <= Turn{0.5},
```
and it matches the image range of atan2 function.

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -356,9 +356,9 @@ GenTrap GenTrap::from_trap(
...
                    << "invalid angle " << phi.value()
-                   << " [turns]: must be in the range [0, 1)");
+                   << " [turns]: must be in the range (-1, 1)");
```
mrguilima:
```suggestion
                   << " [turns]: must be in the range (-0.5, 0.5]");
```

mrguilima (dismissed):

src/orange/orangeinp/ConvexRegion.cc
sethrj: True... maybe we should update the inf wedge to have the same bounds (I think I fmod it between 0 and 1). But since we just take trig functions of the value, anything is valid honestly... i could imagine a user calling with 3/4 instead of -1/4

sethrj:
@mrguilima I've removed the validation checking from phi:
- Any value is valid and just wraps around as one would expect from sin/cos.
- Unlike InfWedge, the value is used to *construct* the shape as opposed to actually being *stored* on the shape.

If the value were part of the shape's data, we *should* regularize it so that there's a one-to-one correspondence between values and shapes produced. However, since phi just gets preprocessed into the same points, there's no need to impose extra effort on the user (e.g. it's ok if they want to specify the rotation as 3/4 turn instead of -1/4 turn).

sethrj:
Is the proposed removal of validation satisfactory @mrguilima ?

mrguilima:
Yes, it looks good!
Thanks

mrguilima (approved):
LGTM.

Issue #1207, opened 2024-04-26
Parameterized source distributions


sethrj:
For testing it would be good to have a little more complexity in our RNG-determined input that would allow us to have a reasonably realistic parameterized set of "primaries" that could come from a Geant4 problem. To do this we should:
- Enable `offload_output_file` to ROOT and set up test beam inputs for HGCal, TileCal (and maybe ttbar inputs for full CMS?)
- Save a GB or so of primaries from each case
- Calculate some histograms for angle, particle type, energy, position (r/z?)
- ???
- Implement a new PrimaryGenerator that can sample from a relatively complex distribution
- Use that distribution to sample primaries in our regression tests

sethrj:
Close enough to closed by #1228.

PR #1228, opened 2024-05-08
Add RootPrimaryGenerator sampler to celer-sim

stognini:
This PR is a follow up on issue #1207  and adds the capability to sample primaries from an existing ROOT file with offloaded primaries generated by RootEventWriter

Main additions:

- New `RootEventSampler` class, derived from `EventReaderInterface`
- Expanded json input options. Now if a ROOT file with offloaded primaries is provided as `event_file`, a `file_sampling_options` struct can be used:
  ```json
  "event_file": "offloaded-primaries.root",
  "file_sampling_options": {
      "num_events": 5,
      "num_merged": 1
  },
  ```
  If either of those numbers is `0`, the `RootEventReader` is constructed and reads it on an event-by-event basis. If both  numbers are positive, the `RootEventSampler ` is constructed and will sample primaries as requested.


stognini:
Currently `RootPrimaryGenerator` is in `phys/`, but maybe should be moved to `io/` which is where `RootEventWriter/Reader` are?

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+    scoped_root_error.throw_if_errors();
+
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return a Real3 from leaves containing `std::array<double, 3>`.
+ */
+Real3 RootPrimaryGenerator::from_array_leaf(TLeaf const& leaf)
```
sethrj: This doesn't depend on class data so should be a free function (you can put it up in an anonymous namespace at the top of the file). In fact since we do something similar in a couple of places, maybe create `celeritas/ext/Convert.root.hh` analogous to `geocel/g4/Convert.geant.hh`.

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+{
+    CELER_EXPECT(leaf.GetLen() == 3);
+    return {static_cast<real_type>(leaf.GetValue(0)),
```
sethrj: If there's a single root call to return a std::array, it would be more efficient to do that in one call and then convert as you return.

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+    if (event_count_.get() == num_events_)
+    {
+        return {};
+    }
+
+    ScopedRootErrorHandler scoped_root_error;
+
+    // Fetch ROOT leaves
+    // TODO: Make these private members?
+    auto lpid = ttree_->GetLeaf("particle");
```
sethrj: It looks from the ROOT documentation like these are all pointers: capture pointers as `auto *` for improved legibility, or maybe even use `TLeaf *` since it's short and has a well-defined API.

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+    CELER_ENSURE(tfile_->IsOpen());
+
+    ttree_.reset(tfile_->Get<TTree>("primaries"));
+    CELER_ENSURE(ttree_);
+    CELER_VALIDATE(ttree_->GetEntries() > 0,
+                   << "TTree `primaries` in '" << tfile_->GetName()
+                   << "' has zero entries");
+    scoped_root_error.throw_if_errors();
+
+    entry_selector_ = std::uniform_int_distribution<>(0, primaries_per_event);
```
sethrj: name function-like class instances like you would name a function
```suggestion
    select_entry_ = std::uniform_int_distribution<>(0, primaries_per_event);
```

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+                                           size_type primaries_per_event,
+                                           unsigned int seed)
+    : params_(std::move(params))
+    , num_events_{num_events}
+    , primaries_per_event_(primaries_per_event)
+{
+    CELER_EXPECT(!filename.empty());
+    CELER_VALIDATE(num_events > 0 && primaries_per_event > 0,
+                   << "'num_events' and 'primaries_per_event options in input "
+                      "file must be set and positive");
```
sethrj: Validate message should first say what's wrong, not what it should be.
```suggestion
    CELER_VALIDATE(num_events > 0, << "invalid 'num_events' (should be positive)");
    CELER_VALIDATE(primaries_per_event > 0, << "invalid 'primaries_per_event'  (should be positive)");
```

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+
+    TrackId track_id{0};
+    result_type result(primaries_per_event_);
+    for (auto i : range(primaries_per_event_))
+    {
+        ttree_->GetEntry(entry_selector_(rng_));
+
+        result[i].particle_id
+            = params_->find(PDGNumber{static_cast<int>(lpid->GetValue())});
+        CELER_ENSURE(result[i].particle_id);
```
sethrj: I'd make this a `CELER_VALIDATE` since this is essentially checking runtime input is easy to screw up

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+{
+    CELER_EXPECT(!filename.empty());
+    CELER_VALIDATE(num_events > 0 && primaries_per_event > 0,
+                   << "'num_events' and 'primaries_per_event options in input "
+                      "file must be set and positive");
+
+    ScopedRootErrorHandler scoped_root_error;
+    tfile_.reset(TFile::Open(filename.c_str(), "read"));
+    CELER_ENSURE(tfile_->IsOpen());
+
```
sethrj: I realize that when I added the "core units" to Celeritas I never added any ROOT metadata to write/validate that the units are the same. Would it be possible (maybe in a follow-on PR) to add that? Maybe @pcanal can suggest the best way to add metadata that's just a single string.

src/celeritas/phys/RootPrimaryGenerator.cc
```diff
@@ -0,0 +1,121 @@
...
+    auto ldir = ttree_->GetLeaf("dir");
+    CELER_ASSERT(ldir);
+    auto ltime = ttree_->GetLeaf("time");
+    CELER_ASSERT(ltime);
+
+    TrackId track_id{0};
+    result_type result(primaries_per_event_);
+    for (auto i : range(primaries_per_event_))
+    {
+        ttree_->GetEntry(entry_selector_(rng_));
```
sethrj: I don't think we should be sampling individual primaries from the file since the primaries are highly correlated inside the event. I was imagining we'd sample *events* from the input, with a user parameter describing how many "ROOT file events" per "RPG event". This way we could run individual pions for the test beam, and then sample 700 different showers for a single "14 TeV" event.

Perhaps @amandalund has an opinion?

sethrj (changes requested):

src/celeritas/phys/RootPrimaryGenerator.cc
amandalund: Yeah, that would definitely be a more realistic input. We might even be able to do that in the `RootEventReader` (optionally return a random subset of events)?

sethrj: To be honest I'd forgotten about RootEventReader 😞 which could perhaps have saved some work here...

src/celeritas/phys/RootEventSampler.cc
```diff
@@ -0,0 +1,63 @@
...
+    CELER_EXPECT(!filename.empty());
+    CELER_EXPECT(params);
+
+    reader_ = std::make_unique<RootEventReader>(filename, std::move(params));
+    CELER_EXPECT(num_clumped_events_ > 0
+                 && num_clumped_events_ <= reader_->num_events());
+
+    rng_.seed(seed);
+    select_event_
+        = std::uniform_int_distribution<std::size_t>(0, reader_->num_events());
```
amandalund: Should we be sampling events without replacement?

sethrj: I think replacement is fine for the intended purpose

amandalund: Discussed with @stognini offline: this should sample with replacement but not return empty events after the number of events in the root file have been sampled.

app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -157,6 +158,8 @@ void to_json(nlohmann::json& j, RunnerInput const& v)
     LDIO_SAVE(geometry_file);
     LDIO_SAVE(physics_file);
     LDIO_SAVE_OPTION(event_file);
+    LDIO_SAVE_WHEN(num_clumped_events,
+                   !v.event_file.empty() && ends_with(v.event_file, ".root"));
```
sethrj: The `ends_with` function knows that if the string isn't long enough for the suffix, the result is "no" rather than "segfault" 😄 
```suggestion
                   ends_with(v.event_file, ".root"));
```

src/celeritas/io/RootEventReader.cc
```diff
@@ -30,38 +30,102 @@ RootEventReader::RootEventReader(std::string const& filename,
...
     scoped_root_error.throw_if_errors();
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Read a specific single event from the primaries tree.
+ */
+auto RootEventReader::operator()(EventId event_id) -> result_type
+{
+    CELER_EXPECT(event_id <= num_entries_);
```
sethrj:
```suggestion
    CELER_EXPECT(event_id < num_events_);
```
Right?

src/celeritas/io/RootEventReader.cc
```diff
@@ -30,38 +30,102 @@ RootEventReader::RootEventReader(std::string const& filename,
...
+            // Cached event entry; Load entry and return event
+            entry_count_ = event_to_entry_[event_id.get()];
+            return this->operator()();
+        }
+        else
+        {
+            // Continue from latest entry count
+            entry_count_ = event_to_entry_.back();
+        }
+    }
```
sethrj: I think you can simplify this by setting `event_to_entry_ = {0}`: no special case needed for the first one.

src/celeritas/io/RootEventReader.cc
```diff
@@ -30,38 +30,102 @@ RootEventReader::RootEventReader(std::string const& filename,
...
+    ttree_->GetEntry(entry_count_);
+    auto expected_evt_id
+        = EventId{from_leaf<size_type>(*ttree_->GetLeaf("event_id"))};
+
+    result_type result;
+    for (; entry_count_ < num_entries_; entry_count_++)
+    {
+        ttree_->GetEntry(entry_count_);
+        auto entry_evt_id
+            = EventId{from_leaf<size_type>(*ttree_->GetLeaf("event_id"))};
```
sethrj: You're repeating the first `GetEntry` here. I think this should be a `do {} while ()` loop instead of `for`.

src/celeritas/io/RootEventReader.cc
```diff
@@ -30,38 +30,102 @@ RootEventReader::RootEventReader(std::string const& filename,
...
+
+        if (entry_evt_id != expected_evt_id)
+        {
```
sethrj, line 85: Definitely want to assert that the newly found event is expected + 1; otherwise your `event_to_entry_` index won't match the event ID.

src/celeritas/phys/RootEventSampler.cc
```diff
@@ -0,0 +1,65 @@
...
+ */
+auto RootEventSampler::operator()() -> result_type
+{
+    if (event_count_.get() == num_sampled_events_)
+    {
+        return {};
+    }
+
+    result_type result;
+    for ([[maybe_unused]] auto i : range(num_clumped_events_))
```
sethrj: I'm not sure how the ROOT file access patterns work, but it might be more efficient to construct a temporary vector of "selected" events, then sort it, then read. @pcanal do you think the performance of replacing random versus sequential reads (a few hundred MB?) is negligible?

src/celeritas/phys/RootEventSampler.cc
```diff
@@ -0,0 +1,65 @@
...
+    {
+        // Select a random event and overwrite its event id
+        auto event = reader_->operator()(EventId{select_event_(rng_)});
+        std::for_each(event.begin(), event.end(), [&](auto& primary) {
+            primary.event_id = event_count_;
```
sethrj: It's rarely a bad idea to be explicit about lambda captures:
```suggestion
        std::for_each(event.begin(), event.end(), [ec=event_count_](auto& primary) {
            primary.event_id = ec;
```

src/celeritas/phys/RootEventSampler.hh
```diff
@@ -0,0 +1,76 @@
...
+    //! \name Type aliases
+    using SPConstParticles = std::shared_ptr<ParticleParams const>;
+    using UPRootEventReader = std::unique_ptr<RootEventReader>;
+    using result_type = std::vector<Primary>;
+    //!@}
+
+  public:
+    // Construct with input data for RootEventReader and sampling information
+    RootEventSampler(std::string const& filename,
+                     SPConstParticles params,
```
sethrj:
```suggestion
                     SPConstParticles particles,
```

src/celeritas/phys/RootEventSampler.cc
```diff
@@ -0,0 +1,65 @@
...
+#include <algorithm>
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct \c RootEventReader , and initialize sampling conditions.
+ */
+RootEventSampler::RootEventSampler(std::string const& filename,
+                                   SPConstParticles params,
```
sethrj:
```suggestion
                                   SPConstParticles particles,
```
(and in usage below) to avoid ambiguity

app/celer-sim/RunnerInput.hh
```diff
@@ -52,6 +52,11 @@ struct RunnerInput
     std::string physics_file;  //!< Path to ROOT exported Geant4 data
     std::string event_file;  //!< Path to input event data
 
+    // Optional setup options when event_file is a ROOT file that should be
+    // used for sampling combinations of events as opposed to just reading them
+    size_type num_sampled_events{};  //!< Events to be sampled from ROOT file
+    size_type num_clumped_events{};  //!< Events to be joined from ROOT file
```
sethrj: Since these options are specific to the ROOT "sampling" input, perhaps it would be better to have a separate struct for managing them?

sethrj (commented):
Looking good @stognini ! The word `clumped` has an unpleasant feel to it; perhaps since we use `merged` elsewhere that would be a better name?

app/celer-sim/RunnerInput.hh
stognini: I originally thought about having something like
```json
"root_sampler": {
    "num_events": 10
    "num_merged": 2
}
```
and ended up keeping it simpler. I can go for this route if preferred.

stognini:
> The word clumped has an unpleasant feel to it

Totally agree. Somehow it felt like 'merged' was a bit off, and 'concatenated' was long... But I guess merged is fine.

app/celer-sim/RunnerInput.hh
sethrj: I think it would improve clarity to have
```c++
struct EventFileSampling {
size_type num_events{}; //!< Total number of events to sample
size_type num_merged{}; //!< ROOT file events per sampled event
explicit operator bool() const { return num_events > 0; }
};
```
and
```c++
EventFileSampling event_file_sampling;
```
so you can write `if (input. event_file_sampling)` and then dig into the values.

sethrj (approved):
Great work! Thanks @stognini .

sethrj (changes requested):
Actually @stognini it looks like there are a few unresolved comments... could you take a look?

sethrj (approved):
Nice!

PR #1229, opened 2024-05-08
Add generic grid inserter to simplify construction

hhollenb:
Add a simple inserter for generic grids, similar to `ValueGridInserter`. I use this for constructing the step limit grids for optical processes, but it can probably be reviewed separately since we'll want to use it in more than one place. Since `ValueGridBuilder` handles grids differently than `GenericGridBuilder`, we can't just copy the `ValueGridInserter` implementation.

The main idea is to handle the `GenericGridBuilder` and `CollectionBuilder` that the grids are being inserted into as a single object, making it easy to pass this data to processes and have them build their grids directly instead of having to manage both the mutable scalar host data (which is an implementation detail) as well as constructing the generic grid collection. Currently I have it templated on the collection index type so it's usable by more collections.

I wrote a trivial change for `CerenkovParams.cc` to use `GenericGridInserter` as an example and to make sure it passed a simple test. The same can't be done for `ScintillationParams.cc` since per particle grids are encapsulated in another data structure before being added to a particle-indexed collection.


src/celeritas/optical/CerenkovParams.cc
```diff
@@ -33,8 +32,7 @@ CerenkovParams::CerenkovParams(SPConstProperties properties)
     auto const& host_ref = properties->host_ref();
 
     HostVal<CerenkovData> data;
-    GenericGridBuilder build_angle_integral(&data.reals);
-    CollectionBuilder angle_integral(&data.angle_integral);
+    GenericGridInserter build_angle_integral(&data.reals, &data.angle_integral);
```
sethrj: Perhaps `insert_angle_integral` now?

src/celeritas/optical/CerenkovParams.cc
```diff
@@ -58,10 +56,9 @@ CerenkovParams::CerenkovParams(SPConstProperties properties)
                                 * (1 / ipow<2>(refractive_index[i - 1])
                                    + 1 / ipow<2>(refractive_index[i]));
         }
-        angle_integral.push_back(
-            build_angle_integral(make_span(energy), make_span(integral)));
+
+        build_angle_integral(make_span(energy), make_span(integral));
     }
-    CELER_ASSERT(angle_integral.size() == host_ref.refractive_index.size());
```
sethrj, line 64: I think this should be restored, and you can use `data.angle_integral` instead of the now-deleted builder

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,139 @@
...
+    return grids_.push_back(grid_builder_(grid, values));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add an empty grid. Useful for when there's no imported grid present for
+ * a given material.
+ */
+template<typename Index>
+auto GenericGridInserter<Index>::operator()(void) -> Index
```
sethrj: Explicit void arguments are a legacy C thing that are discouraged in modern C++
```suggestion
auto GenericGridInserter<Index>::operator()() -> Index
```

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,139 @@
...
+    Index operator()(SpanConstFlt grid, SpanConstFlt values);
+
+    //! Add a grid of generic data with linear interpolation
+    Index operator()(SpanConstDbl grid, SpanConstDbl values);
+
+    //! Add an imported physics vector as a grid
+    Index operator()(ImportPhysicsVector const& vec);
+
+    //! Add an empty grid (no data present)
+    Index operator()(void);
```
sethrj: Also remove the `!` for the declarations above: doxygen docs should go with the *definition* (whether it's a class or a function).

```suggestion
    // Add an empty grid (no data present)
    Index operator()();
```

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,139 @@
...
+ * Add an imported physics vector as a generic grid to the collection.
+ * Returns the id of the inserted grid, or an empty id if the vector is
+ empty.
+ */
+template<typename Index>
+auto GenericGridInserter<Index>::operator()(ImportPhysicsVector const& vec)
+    -> Index
+{
+    if (vec.x.empty())
+        return Index{};
```
sethrj: Is this expected to happen? It seems like this *could* lead to propagating unusual behavior if e.g. a physics vector isn't initialized. If the "insert an empty" is an exceptional case that only would apply in certain builders, I think it'd be better to `CELER_EXPECT(!vec.x.empty());`, and then have the caller explicitly do `if (!vec.x.empty()) { insert(vec); } else { insert(); }`.

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,139 @@
...
+    if (grid.empty())
+        return Index{};
+
+    return grids_.push_back(grid_builder_(grid, values));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add an empty grid. Useful for when there's no imported grid present for
+ * a given material.
```
sethrj:
```suggestion
 * Add an empty grid.
 *
 * Useful for when there's no imported grid present for a given material.
```

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,139 @@
...
+    : grid_builder_(real_data), grids_(grid)
+{
+    CELER_EXPECT(real_data && grid);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add an imported physics vector as a generic grid to the collection.
+ * Returns the id of the inserted grid, or an empty id if the vector is
+ empty.
```
sethrj:
```suggestion
 * Add an imported physics vector as a generic grid to the collection.
 *
 * Returns the id of the inserted grid, or an empty id if the vector is
 * empty.
```

sethrj (commented):
Nice addition!

src/celeritas/grid/GenericGridInserter.hh
hhollenb: For some reason I thought I was mimicing a check I saw somewhere else, but it appears I was just imagining it. Changing to use `CELER_EXPECT`.

src/celeritas/grid/GenericGridInserter.hh
```diff
@@ -0,0 +1,136 @@
...
+ * the specified grid collection.
+ *
+ * \code
+    GenericGridInserter insert(&data->reals, &data->generic_grids);
+    std::vector<GenericGridIndex> grid_ids;
+    for (material : range(MaterialId{mats->size()}))
+        grid_ids.push_back(insert(physics_vector[material.get()]));
+   \endcode
+ */
+template<typename Index>
```
sethrj: Didn't notice this, but for style consistency we use "class" for the template argument, "typename" is only for resolving templates type aliases etc

sethrj (commented):
One last comment! And adding a tiny test to the generic grid ones would be a nice touch.

src/celeritas/grid/GenericGridInserter.hh
hhollenb: I only see a `GenericGridCalculator.test.cc` file -- I can add some tests there or also add make some `GenericGridBuilder` and `GenericGridInserter` like for the value grids

sethrj: Would you mind renaming `celeritas/grid/GenericCalculator.test.cc` to `GenericGrid.test.cc` and adding it there?

sethrj (approved):
Looks good; I'd be fine with combining all the GenericGrid tests (the GenericCalculator test effectively tests the GridBuilder as is).

The failure on the CI is:
```
[ RUN      ] GenericGridBuilderTest.build_vec
unknown file: Failure
C++ exception with description "/home/runner/work/celeritas/celeritas/src/celeritas/grid/GenericGridBuilder.cc:49:
celeritas: precondition failed: pvec.vector_type == ImportPhysicsVectorType::free" thrown in the test body.
```

sethrj (approved):
Thanks for fighting through all the float/double issues!

PR #1230, opened 2024-05-10
Support combined single and multiple Coulomb  scattering

amandalund:
This refactors and updates the single Coulomb scattering code in preparation for adding the Wentzel VI MSC model and combined single and multiple scattering. The main changes are:
- Refactor most of the Coulomb scattering model data into a separate Wentzel OK&VI data and params class. This data will also be used by the Wentzel VI MSC model, which samples single scattering for small step lengths and needs the parameters for the quantities calculated in the Wentzel helper class (cross sections, screening factor, etc.)
- Sample the scattering angle and calculate the Coulomb cross sections for arbitrary min/max scattering angles instead of assuming polar angle limits for single scattering. With combined SS and MSC, Wentzel VI will be used to sample scatterings with angles below a given polar angle limit, and single Coulomb scattering will be used above the limit.
- Calculate the electron and nuclear cross sections in addition to just the xs ratio (Wentzel VI will need the total xs)
- Import the MSC theta limit, angle limit factor, and nuclear form factor type from Geant4


src/celeritas/Types.hh
```diff
@@ -160,6 +160,17 @@ enum class MscStepLimitAlgorithm
     size_,
 };
 
+//---------------------------------------------------------------------------//
+//! Nuclear form factor model for Coulomb scattering
+enum class NuclearFormFactorType
```
hhollenb, line 165: If we're moving nuclear form factors to a more general namespace, should we also move their calculation out of `WentzelDistribution` so there's a single implementation?

src/celeritas/em/distribution/WentzelDistribution.hh
```diff
@@ -66,20 +67,28 @@ class WentzelDistribution
   private:
     //// DATA ////
 
-    // Shared model data
-    CoulombScatteringRef const& data_;
+    // Shared Coulomb scattering data
+    NativeCRef<WentzelOKVIData> const& wentzel_;
+
+    // Helper for calculating xs ratio and other quantities
+    WentzelHelper const helper_;
```
sethrj: Should this be a const reference?

src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -0,0 +1,113 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Parameters used in both single and multiple Coulomb scattering models.
+ *
+ * When combined single and multiple scattering is enabled, the Wentzel VI MSC
+ * model is used to sample scatterings with angles below the polar angle limit,
+ * and the single Coulomb scattering model is used for angles above the limit.
```
whokion: I am a bit confused here.  What do you mean by "combined single and multiple scattering is enabled" here?  It sounds like that there is a combination of the single Coulomb scattering model and the WentzelVI model in the process level.  Do you mean "combining single scattering and multiple scattering inside WentzelVI"?  May it be better to say "Parameters used in both single Coulomb scattering and Wentzel VI multiple scattering models"?

src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -0,0 +1,113 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Per-element data used by the Coulomb scattering and Wentzel VI models.
+ *
+ * The matrix of coefficients used to approximate the ratio of the Mott to
+ * Rutherford cross sections was developed in
+ * T. Lijian, H. Quing and L. Zhengming, Radiat. Phys. Chem. 45 (1995),
+ *   235-245
+ * and
+ * M. J. Boschini et al. arXiv:1111.4042
```
whokion:
```suggestion
 * The matrix of coefficients used to approximate the ratio of the Mott to
 * Rutherford cross sections was developed in T. Lijian, H. Quing and 
 * L. Zhengming, Radiat. Phys. Chem. 45 (1995), 235-245 and
 * M. J. Boschini et al. arXiv:1111.4042.
```

src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -0,0 +1,113 @@
...
+    // Constant prefactor for the squared momentum transfer [(MeV/c)^-2]
+    IsotopeItems<real_type> nuclear_form_prefactor;
+
+    // Per element form factors
+    ElementItems<MottElementData> elem_data;
+
+    // Check if the data is initialized
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return params && !elem_data.empty();
```
whokion:
```suggestion
        return params && !nuclear_form_prefactor.empty() && !elem_data.empty();
```

src/celeritas/em/distribution/WentzelDistribution.hh
```diff
@@ -113,17 +123,24 @@ class WentzelDistribution
...
     , target_(target)
-    , element_data_(element_data)
-    , helper_(particle, target.atomic_number(), data, cutoff)
+    , el_id_(el_id)
+    , costheta_min_(costheta_min)
+    , costheta_max_(costheta_max)
 {
+    CELER_EXPECT(el_id_ < wentzel_.elem_data.size());
+    CELER_EXPECT(costheta_min_ >= -1 && costheta_min_ <= 1);
+    CELER_EXPECT(costheta_max_ >= -1 && costheta_max_ <= 1);
```
whokion:
```suggestion
    CELER_EXPECT(costheta_max_ >= -1 && costheta_max_ <= 1);
    CELER_EXPECT(costheta_max_ <= costheta_min_);
```
if costheta_max_ == cos(theta_max) and costheta_min_ == cos(theta_min)

src/celeritas/em/distribution/WentzelDistribution.hh
```diff
@@ -134,33 +151,35 @@ template<class Engine>
...
     real_type cos_theta = 1;
-    if (BernoulliDistribution(helper_.calc_xs_ratio())(rng))
+    if (BernoulliDistribution(
+            helper_.calc_xs_ratio(costheta_min_, costheta_max_))(rng))
     {
         // Scattered off of electrons
-        cos_theta = this->sample_cos_t(helper_.costheta_max_electron(), rng);
+        real_type const ct_max_elec = helper_.costheta_max_electron();
+        real_type ct_min = max(costheta_min_, ct_max_elec);
+        real_type ct_max = max(costheta_max_, ct_max_elec);
```
whokion: Just for double-checking that these are logically correct.

src/celeritas/em/distribution/WentzelDistribution.hh
```diff
@@ -241,28 +261,39 @@ CELER_FUNCTION real_type WentzelDistribution::mom_transfer_sq(real_type cos_t) c
...
 {
     // Sample scattering angle [Fern] eqn 92, where cos(theta) = 1 - 2*mu
-    // For incident electrons / positrons, theta_min = 0 always
-    real_type const mu = real_type{0.5} * (1 - cos_t_max);
+    real_type const mu1 = real_type{0.5} * (1 - costheta_min);
+    real_type const mu2 = real_type{0.5} * (1 - costheta_max);
     real_type const xi = generate_canonical(rng);
+    // TODO: (1 - xi) --> xi and update tests (kept for now to ensure tests
+    // don't change)
+    real_type const w = (1 - xi) * (mu2 - mu1);
```
whokion:
```suggestion
    // TODO: (1 - xi) --> xi and update tests (kept for now to ensure tests
    // don't change)
    real_type const w = (1 - generate_canonical(rng)) * (mu2 - mu1);
```

src/celeritas/em/interactor/CoulombScatteringInteractor.hh
```diff
@@ -92,21 +98,31 @@ class CoulombScatteringInteractor
...
-                   shared.elem_data[el_id],
-                   // TODO: Use proton when supported
-                   cutoffs.energy(shared.ids.electron),
-                   shared)
+    , helper_(particle,
+              material,
+              target.atomic_number(),
+              wentzel.params,
+              shared.ids,
+              // TODO: Use proton when supported
```
whokion: Protons will follow G4hCoulombScatteringModel, so this model will not support the Coulomb scattering for protons.

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -0,0 +1,929 @@
...
+    }
+    CELER_ENSURE(host_data.nuclear_form_prefactor.size()
+                 == materials.num_isotopes());
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the constant prefactors of the squared momentum transfer.
+ *
+ * This factor is used in the exponential and Gaussian nuclear form models: see
```
whokion:
```suggestion
 * This factor is used in the exponential and Gaussian nuclear form factor types: see
```

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -0,0 +1,929 @@
...
+ * Eqs. 2.262--2.264 of [LR15].
+ *
+ * Specifically, it calculates \f$ (r_n/\bar h)^2 / 12 \f$. A special case is
```
whokion:
```suggestion
 * Specifically, it calculates \f$ (r_n/\bar h)^2 / 12 \f$ where \f$ r_n f$ is .... A special case is
```

src/celeritas/em/params/WentzelOKVIParams.hh
```diff
@@ -0,0 +1,85 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+class MaterialParams;
+struct ImportData;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct and store shared Coulomb scattering data.
```
whokion:
```suggestion
 * Construct and store shared Coulomb and multiple scattering data.
```

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -95,31 +134,76 @@ class WentzelHelper
...
+ */
+CELER_FUNCTION real_type WentzelHelper::calc_xs_electron(
+    real_type costheta_min, real_type costheta_max) const
+{
+    costheta_min = max(costheta_min, costheta_max_elec_);
+    costheta_max = max(costheta_max, costheta_max_elec_);
+    if (costheta_min <= costheta_max)
+    {
+        return 0;
+    }
```
whokion: So, no scattering if `cos(theta_min) <= cos(theta_max)`, but confusing names as it can be read as `(cos_theta)_min <= (cos_theta)_max`. May be (cos_thetamin <= cos_thetamax) or (cos_theta_min <= cos_theta_max) better?  Or, put variable definitions with a comment to clarify.

test/celeritas/em/CoulombScattering.test.cc
```diff
@@ -370,28 +424,33 @@ TEST_F(CoulombScatteringTest, simple_scattering)
 
 TEST_F(CoulombScatteringTest, distribution)
 {
-    CoulombScatteringHostRef const& data = model_->host_ref();
+    auto const material = this->material_track().make_material_view();
+    IsotopeView const isotope
+        = material.make_element_view(elcomp_id_).make_isotope_view(isocomp_id_);
+
+    // TODO: Use proton ParticleId{2}
```
whokion, line 431: Just curious where this proton cut used? Is this cut used for the production recoiled nucleus/ions?

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -0,0 +1,929 @@
...
+    bool wentzel = has_msc_model(data, ImportModelClass::wentzel_vi_uni);
+    bool coulomb = has_model(data, ImportModelClass::e_coulomb_scattering);
+    if (!(wentzel || coulomb))
+    {
+        // No Wentzel VI MSC or Coulomb scattering present
+        return nullptr;
+    }
+
+    // Use combined single and multiple Coulomb scattering if both the single
+    // scattering and the Wentzel VI models are present
```
whokion: Again, this description is somewhat confusing as it sounds like that two models are combined together.  The actual combination is to implement for cases for the single scattering (not the Coulomb scattering model itself) and the WentzelVI multiple scattering using the common WentzelOKVI cross section calculation, which are I think technically different.

whokion (approved):
Very good!  Left some trivial comments just for clarification.  Otherwise, good to go for now and hope that the WentzelVI msc model is available soon.  We may also need more low/high-level tests/verification/validation for the Coulomb scattering, which belongs to me and @hhollenb. Thanks for the update.

src/celeritas/em/interactor/CoulombScatteringInteractor.hh
amandalund: I think this todo is for changing the _cutoff_ to be for protons instead of electrons: https://github.com/celeritas-project/celeritas/pull/861#discussion_r1288852391

test/celeritas/em/CoulombScattering.test.cc
amandalund: Referring back to https://github.com/celeritas-project/celeritas/pull/861#discussion_r1288852391

src/celeritas/em/data/WentzelOKVIData.hh
amandalund: I removed this check since this should be sized as `num_isotopes` and we don't require isotopes to be defined in a material.

src/celeritas/em/interactor/CoulombScatteringInteractor.hh
whokion: Okay - protons will  not be supported in this model, but the recoil nucleus (with the minimal proton production cut) should be included properly when when (?? maybe never) we support generic ions in the future.  So, the better comment may be "Use the proton production cutoff when the recoiled nucleus production is supported".

amandalund: Thanks, updated the comment.

src/celeritas/em/data/WentzelOKVIData.hh
amandalund: You're right, that's confusing... I meant when the Wentzel VI and Coulomb scattering models are used together. I'll try to clarify the comment.

src/celeritas/em/distribution/WentzelDistribution.hh
amandalund: I believe so: https://github.com/Geant4/geant4/blob/dda54bbdcfe0cc76177eaa03a657592aac0b1eb8/source/processes/electromagnetic/standard/src/G4WentzelOKandVIxSection.cc#L344

src/celeritas/em/xs/WentzelHelper.hh
amandalund: Good idea to be more explicit. I had the same confusion with the names.

amandalund:
Thanks @whokion for the helpful feedback! I agree about adding more testing, and I also still need to set the correct lower energy limit for the Coulomb scattering model, which I'm planning do in a follow-up PR.

src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -0,0 +1,112 @@
...
+    //! Whether to use combined single and multiple scattering
+    bool is_combined;
+    //! Polar angle limit between single and multiple scattering
+    real_type costheta_limit;
+    //! Factor for the screening coefficient
+    real_type screening_factor;
+    //! Factor used to calculate the maximum scattering angle off of a nucleus
+    real_type a_sq_factor;
+    // Model for the form factor to use
+    NuclearFormFactorType form_factor_type;
```
sethrj: Missing default initializers

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -0,0 +1,929 @@
...
+    opts.is_combined = wentzel && coulomb;
+    opts.polar_angle_limit = [&]() -> real_type {
+        if (!coulomb)
+        {
+            // Set the maximum scattering angle for Wentzel VI MSC
+            return constants::pi;
+        }
+        if (!wentzel)
+        {
+            // Set the minimum scattering angle for Coulomb single scattering
```
sethrj, line 48: Is this :
```suggestion
            // Set the maximum scattering angle for Coulomb single scattering
```
?
So for wentzel you can fully backscatter, but with coulomb scattering you can't change more than 90 degrees?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -919,8 +942,12 @@ ImportEmParameters import_em_parameters()
 #else
     CELER_DISCARD(len_scale);
 #endif
+    import.msc_theta_limit = g4.MscThetaLimit();
+    import.angle_limit_factor = g4.FactorForAngleLimit();
     import.apply_cuts = g4.ApplyCuts();
     import.screening_factor = g4.ScreeningFactor();
+    import.angle_limit_factor = g4.FactorForAngleLimit();
```
sethrj: Duplicate assignment

src/celeritas/mat/MaterialView.hh
```diff
@@ -267,6 +270,15 @@ CELER_FUNCTION real_type MaterialView::radiation_length() const
     return this->material_def().rad_length;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Inverse effective atomic mass to the 2/3 power [1/mass^2/3].
```
sethrj: Is this quantity something that's going to be used across models? If not then it could go in the okwenzel params.

sethrj (approved):
That's a lot of work, thanks! Just have a couple small comments.

src/celeritas/em/params/WentzelOKVIParams.cc
amandalund: It should be the maximum scattering angle for wentzel VI (180° if there's no single coulomb, or the msc theta limit from the EM params if there is) and the minimum scattering angle for coulomb (0° if there's no wentzel VI, else the msc theta limit).

One thing I'm still unsure about is the default value of the msc theta limit... from the literature ([e.g. here](https://iopscience.iop.org/article/10.1088/1742-6596/219/3/032045/pdf)) it sounds like it should be around 0.2 radians, but as far as I can tell in the EM params it's pi?

src/celeritas/mat/MaterialView.hh
amandalund: I don't think it will be used by other models. I had it in okwvi originally and then moved it to the material params for some reason I can't recall 🤷 but yeah I think it makes more sense to have it there.

sethrj (approved):

sethrj:
Thanks @amandalund ! @hhollenb Do you want to take a look at this before I merge?

hhollenb (approved):
Everything looks really good! With these changes it looks like I'm getting better agreement with Geant4 for just the single Coulomb scattering.

src/celeritas/Types.hh
amandalund: I moved the enum here because we're now also using it in the import data structs, a lot of our other enums live here, and we shouldn't need additional changes for the SWIG build (#1159), but there wouldn't be any problem with leaving in the wentzel data. I think as long as the form factor calculations are only used in the wentzel distribution we can leave them there.

amandalund:
Thanks @hhollenb, that's great!

PR #1231, opened 2024-05-11
Make accel "auto flush" threshold configurable

amandalund:
Currently in our local transporter we offload tracks to Celeritas after `num_track_slots` have been buffered; however, it may be beneficial in some cases to buffer more than the number of available track slots before starting the stepping loop. This change lets the user specify the "auto flush" threshold (it will still default to the number of track slots). I haven't tested many configurations, but for example I saw a 15-25% performance improvement for the CMS run 3, 32 ttbar event problem on an A100 by decreasing the number of track slots from 2^19 to 2^16 but increasing the buffer size to 2^20:
speedup-a100

### DIFF
--- a/app/celer-g4/RunInput.hh
+++ b/app/celer-g4/RunInput.hh
@@ -67,6 +67,7 @@ struct RunInput
     size_type max_steps{unspecified};
     size_type initializer_capacity{};
     real_type secondary_stack_factor{};
+    size_type auto_flush{};
     bool sync{false};
     bool default_stream{false};  //!< Launch all kernels on the default stream
 
--- a/app/celer-g4/RunInput.hh
+++ b/app/celer-g4/RunInput.hh
@@ -67,7 +67,7 @@ struct RunInput
     size_type max_steps{unspecified};
     size_type initializer_capacity{};
     real_type secondary_stack_factor{};
-    size_type auto_flush{};
+    size_type auto_flush{};  //!< Defaults to num_track_slots
     bool sync{false};
     bool default_stream{false};  //!< Launch all kernels on the default stream
 
--- a/src/accel/LocalTransporter.cc
+++ b/src/accel/LocalTransporter.cc
@@ -50,7 +50,7 @@ namespace celeritas
  */
 LocalTransporter::LocalTransporter(SetupOptions const& options,
                                    SharedParams const& params)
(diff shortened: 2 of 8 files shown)


app/celer-g4/RunInput.hh
```diff
@@ -67,6 +67,7 @@ struct RunInput
     size_type max_steps{unspecified};
     size_type initializer_capacity{};
     real_type secondary_stack_factor{};
+    size_type auto_flush{};
```
sethrj:
```suggestion
    size_type auto_flush{}; //!< Defaults to num_track_slots
```

src/accel/SetupOptions.hh
```diff
@@ -126,6 +126,8 @@ struct SetupOptions
     size_type initializer_capacity{};
     //! At least the average number of secondaries per track slot
     real_type secondary_stack_factor{3.0};
+    //! Number of tracks to buffer before offloading
```
sethrj:
```suggestion
    //! Number of tracks to buffer before offloading (when unset: max num tracks)
```

src/accel/LocalTransporter.cc
```diff
@@ -50,7 +50,7 @@ namespace celeritas
  */
 LocalTransporter::LocalTransporter(SetupOptions const& options,
                                    SharedParams const& params)
-    : auto_flush_(options.max_num_tracks)
+    : auto_flush_(options.auto_flush)
```
sethrj: I don't think this is defaulting correctly when run through a user application (other than `celer-g4`)
```suggestion
    : auto_flush_(options.auto_flush ? options.auto_flush : options.max_num_tracks)
```

sethrj (approved):
That's a great improvement! Hopefully the performance with number of track slots will scale better with an improved geometry engine.

src/accel/LocalTransporter.cc
amandalund: Ah good catch.

esseivaju:
@amandalund If you buffer more primaries than we have track slots, where is that overflow handled?

amandalund:
The primaries are converted to track initializers and stored in that buffer while they are waiting to be transported, so as long as the "auto flush" threshold is less than the initializer capacity there shouldn't be any overflow.

PR #1232, opened 2024-05-11
Support twisted sides for the gentrap shape

mrguilima:
Added some tests and removed the temporary throws related to this feature not being implemented.

There might still be an issue with properly assigning the Sense of the twisted sides.


src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -471,13 +449,42 @@ void GenTrap::build(ConvexSurfaceBuilder& insert_surface) const
...
 
         // Calculate outward normal by taking the cross product of the edges
-        auto normal = make_unit_vector(cross_product(jlo - ilo, ihi - ilo));
+        auto lo_normal = make_unit_vector(cross_product(jlo - ilo, ihi - ilo));
+        auto hi_normal = make_unit_vector(cross_product(ihi - jhi, jlo - jhi));
 
-        // Insert the plane
-        insert_surface(Sense::inside, Plane{normal, ilo});
+        // Insert appropriate service, planar or "twisted"
+        if (soft_equal(dot_product(lo_normal, hi_normal), real_type{1}))
```
sethrj, line 460: Hm, I think we should expose the user-selected tolerance through the `ConvexSurfaceBuilder` and use that for the soft equal here. I can do that in a follow-on PR.

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -471,13 +449,42 @@ void GenTrap::build(ConvexSurfaceBuilder& insert_surface) const
...
+            // Insert a real plane
+            insert_surface(Sense::inside, Plane{lo_normal, ilo});
+        }
+        else
+        {
+            // Determine coefficients
+            auto alo = jlo[1] - ilo[1];
+            auto ahi = jhi[1] - ihi[1];
+            auto blo = -(jlo[0] - ilo[0]);
+            auto bhi = -(jhi[0] - ihi[0]);
```
```suggestion
            auto blo = ilo[0] - jlo[0];
            auto bhi = ihi[0] - jhi[0];
```
Otherwise you introduce signed zeros.

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -724,11 +718,47 @@ TEST_F(GenTrapTest, DISABLED_tetrahedron)
...
+    EXPECT_EQ(expected_node, result.node);
+    EXPECT_VEC_EQ(expected_surfaces, result.surfaces);
+    EXPECT_FALSE(result.interior) << result.interior;
+    EXPECT_VEC_SOFT_EQ((Real3{-inf, -2, -4}), result.exterior.lower());
+    EXPECT_VEC_SOFT_EQ((Real3{inf, inf, 4}), result.exterior.upper());
+}
+
+TEST_F(GenTrapTest, full2)
+{
+    // General non-planar GenTrap with 'twisted' faces is not yet implemented
```
sethrj:
```suggestion
```

sethrj (approved):
Nice work @mrguilima ! I did look at the GQ "surface simplifier" and will have to add some code to "regularize" these surfaces (i.e. make sure two opposing ones are converted to the same surface with opposite senses, and to scale them so that two identical surfaces (e.g. two trapezoids with a coincident twisted surface but different sizes) have the same quadric values.

src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -471,13 +449,42 @@ void GenTrap::build(ConvexSurfaceBuilder& insert_surface) const
...
+            auto chi = jhi[0] * ihi[1] - ihi[0] * jhi[1];
+
+            Real3 abc{0, 0, 0}, def{0, 0, 0}, ghi{0, 0, 0};
+            auto factor = 0.5 / hz_;
+            def[0] = (ahi - alo) * factor;
+            def[1] = (bhi - blo) * factor;
+            ghi[2] = (chi - clo) * factor;
+            ghi[0] = 0.5 * (alo + ahi);
+            ghi[1] = 0.5 * (blo + bhi);
+            auto offset = 0.5 * (clo + chi);
```
sethrj: The source of the error:
```
/home/runner/work/celeritas/celeritas/src/orange/orangeinp/ConvexRegion.cc:486:58: error: non-constant-expression cannot be narrowed from type 'double' to 'celeritas::real_type' (aka 'float') in initializer list [-Wc++11-narrowing]
                           GeneralQuadric{abc, def, ghi, offset});
                                                         ^~~~~~
```
use 
```suggestion
            constexpr real_type half = 0.5;
            auto factor = half / hz_;
            def[0] = (ahi - alo) * factor;
            def[1] = (bhi - blo) * factor;
            ghi[2] = (chi - clo) * factor;
            ghi[0] = (alo + ahi) * half;
            ghi[1] = (blo + bhi) * half;
            auto offset = half * (clo + chi);
```

sethrj (approved):

PR #1234, opened 2024-05-13
Split ORANGE test into subtests

sethrj:
@mrguilima rightly pointed out that the ORANGE test file is getting unwieldy. This PR simply breaks the test into a couple of files based on requirements.


test/orange/CMakeLists.txt
```diff
@@ -30,6 +30,14 @@ if(CELERITAS_USE_Geant4)
   celeritas_get_g4libs(_g4_geo_libs geometry)
 endif()
 
+if(CELERITAS_USE_Geant4 AND CELERITAS_REAL_TYPE STREQUAL "double")
+  set(_use_g4org TRUE)
+  set(_needs_g4org)
+else()
+  set(_use_g4org FALSE)
```
mrguilima, line 37: Isn't ORANGE supposed to handle floats?
How about using doubles during the G4->org conversion/construction phase, and then use floats for the running phase?

mrguilima (approved):
Looks good.
Thanks!

test/orange/CMakeLists.txt
sethrj: Yes, it can handle them on its own but  right now there are some type issues for the double/float. I agree about using doubles for the CPU setup code and floats for runtime code. However, that would require some extra templating in the code (for classes that are accessed both at setup time and runtime) and so I'm not sure if that's worth the near-term effort.

PR #1235, opened 2024-05-14
Improve transport of single track on CPU

esseivaju:
With a single track slot, `ActionSequence::execute` is a hot path and very sensitive to overhead.

-  Only pring logs in `LocalTransporter::flush` when running on device
- `Stopwatch` adds a significant overhead: disable it by default in `ActionSequence::execute` and add the `action_times` config option to explicitly enable it (formerly sync)  
- Skip unnecessary post-step actions when transporting one track at a time


src/accel/LocalTransporter.cc
```diff
@@ -190,10 +191,6 @@ void LocalTransporter::Flush()
         return;
     }
 
-    CELER_LOG_LOCAL(info) << "Transporting " << buffer_.size()
-                          << " tracks from event " << event_id_.unchecked_get()
-                          << " with Celeritas";
-
```
esseivaju: I removed this entirely, but maybe we should guard it with `celeritas::device`?

sethrj: Yeah. Or at least make it `debug` level, or even something like printing it every N buffered tracks...

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -113,6 +113,13 @@ void ActionSequence<Params>::execute(Params const& params, State<M>& state)
         // Execute all actions and record the time elapsed
         for (auto i : range(actions_.size()))
         {
+            // if we're running on track slot on host, we can skip unnecessary
+            // post-step actions
+            if (M == MemSpace::host && state.size() == 1
+                && actions_[i]->order() == ActionOrder::post
+                && actions_[i]->action_id()
+                       != state.ref().sim.post_step_action[TrackSlotId{0}])
+                continue;
```
sethrj:
```suggestion
            if (M == MemSpace::host && state.size() == 1
                && actions_[i]->order() == ActionOrder::post
                && actions_[i]->action_id()
                       != state.ref().sim.post_step_action[TrackSlotId{0}])
            {
                // Running a single track slot on host:
                // Skip inapplicable post-step action
                continue;
            }
```

src/corecel/sys/Stopwatch.hh
```diff
@@ -61,9 +68,24 @@ double Stopwatch::operator()() const
...
     using DurationSec = std::chrono::duration<double>;
 
-    auto duration = Clock::now() - start_;
+    auto duration = Stopwatch::now() - start_;
     return std::chrono::duration_cast<DurationSec>(duration).count();
 }
 
+inline auto Stopwatch::now() -> TimePoint
+{
+    static bool const result = [] {
```
sethrj: best to avoid `static` storage duration in `inline` functions because it can make trouble for the linker... where else is the stopwatch being called that is adding this level of overhead? I guarded the timer in `celer-sim` 🤔 if it's just the "action" timing that counts, maybe we should make the counter there locally dependent on the `sync` variable (which, if off, disables GPU action timing)?

esseivaju: Have you noticed static storage in inline function being an issue before? By the standard the object is shared between all definitions.

The call in ActionSequence::execute is the problem, on CPU it will always call the Stopwatch. Enabling it only when we have the sync option on could be an alternative.

sethrj: I've been burned before by *template* static inlines before... but compilers/linkers were more primitive and we could've been doing something subtly wrong.

I think it would be a better solution to suppress the timing at a higher level than the stopwatch though, especially since with "sync" off for GPU, the `["time"]["action"]` block is empty, whereas for CPU it would show all zeros. Perhaps we could rename `sync` to `action_times`?

esseivaju: ah yeah, with template inline static is more tricky since each instantiation would define different functions.

I renamed sync to action_times, reverted the changes in Stopwatch, and only measure time when requested on host.

app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -100,7 +100,7 @@ void from_json(nlohmann::json const& j, RunnerInput& v)
     LDIO_LOAD_REQUIRED(initializer_capacity);
     LDIO_LOAD_REQUIRED(secondary_stack_factor);
     LDIO_LOAD_REQUIRED(use_device);
-    LDIO_LOAD_OPTION(sync);
+    LDIO_LOAD_OPTION(action_times);
```
sethrj, line 104: Can you add next to the other `LDIO_LOAD_DEPRECATED`:

```c++
    LDIO_LOAD_DEPRECATED(sync, action_times);
```

src/accel/SetupOptionsMessenger.cc
```diff
@@ -207,8 +207,9 @@ SetupOptionsMessenger::SetupOptionsMessenger(SetupOptions* options)
         add_cmd(&options->cuda_heap_size,
                 "heapSize",
                 "Set the CUDA per-thread heap size for VecGeom");
-        add_cmd(
-            &options->sync, "sync", "Sync the GPU at every kernel for timing");
+        add_cmd(&options->action_times,
+                "actionTimes",
+                "Sync the GPU at every kernel for timing");
```
sethrj:
```suggestion
                "Add timers around every action (may reduce performance)");
```

src/celeritas/global/detail/ActionSequence.cc
```diff
@@ -103,40 +103,51 @@ template<MemSpace M>
...
-            if (M == MemSpace::host && state.size() == 1
-                && actions_[i]->order() == ActionOrder::post
-                && actions_[i]->action_id()
-                       != state.ref().sim.post_step_action[TrackSlotId{0}])
-                continue;
-            ScopedProfiling profile_this{actions_[i]->label()};
-            Stopwatch get_time;
-            actions_[i]->execute(params, state);
-            if (M == MemSpace::device)
+            if (auto const& action = actions_[i]; !skip_post_action(action))
```
sethrj: Since these are shared pointers I think it would improve readability (and maybe micro-performance by reducing indirection?) to dereference the shared pointer and have `action` be the actual `Action` object rather than a shared pointer:
```suggestion
            if (auto const& action = *actions_[i]; !skip_post_action(action))
```

sethrj (approved):

app/celer-sim/RunnerInputIO.json.cc
esseivaju: should we do the same for celer-g4?

sethrj: It couldn't hurt. I don't think we have as many inputs for it though: we *do* want `celer-sim` to work across a couple different versions so that we can run our peformance regression tests.

esseivaju:
With one track slot and skipping post-step actions, Celeritas is ~5% faster than Geant4 on the ATLAS tilecal benchmark. If not skipping post-step actions, we're pretty much even. Batching tracks is still faster, even on CPU.

Issue #983, opened 2023-10-16
Add "region" support


sethrj:
Geant4 regions have custom:

- Production cuts [see G4ProductionCutsTable::CreateCoupleTables]
- Generic "user region information"
- User limits (step limiter, tracking cuts) [see G4LogicalVolume::GetUserLimits]
- Fast simulation managers [see G4FastSimulationManagerProcess::PostStepGetPhysicalInteractionLength]
- Field managers [see G4PropagatorInField::FindAndSetFieldManager]
- MSC safety calculation algorithm

But many of those can be overridden on a per-LV basis. Eventually we might want to map **each combination** of these from logical volumes to independent Celeritas regions.

Regions are used in Geant4 for:

- Region-dependent field (disable field outside of trackers), #538 
- Region-dependent field parameters (sharing a single propagation kernel): see #883 for description of CMS using it
- Fast simulation, for which we need "onloading" (https://github.com/celeritas-project/celeritas/issues/1894)
- Physics parameters

## Execution plan

**One PR for each major bullet point**

- [ ] Add `Regions` to `inp::Model` (this will look a lot like #1955)
  - [ ] New struct `Region` with label and `set<VolumeId>` to which it applies. (**Most attributes should be part of another class, and keying on region ID, rather than the Region having a bunch of different specifications.**)
  - [ ] Load region data in GeantGeoParams: `geant_to_id`/`id_to_geant` for `G4Region*` (add `vector<G4Region const*>` to `GeantGeoParams`)
  - [ ] Add tests: add regions to `multi-level.gdml`, `four-levels`, `lar-sphere`, test by extending  `GenericGeoModelInp` and the `IsRefEq`
- [ ] Integrate regions into runtime code (I'm not 100% sure the best way to structurally do this, but perhaps:)
  - [ ] Rename `GeoMaterial*` to `VolumeExtras*` (non-structural data about volumes)
  - [ ] Pass `inp::Regions` during construction?

Future work:
- Are regions in Geant4 fundamentally about geometry model structure, physics, or both? (From what I can see, it seems more like physics.)
- Should Celeritas try to keep the same regions as Geant4, or should we merge other physics treatments (like field definitions/parameters) into our "region"?
- If we want a broader region definition compared to Geant4, should we have `GeoRegionId` map to `G4Region*` and have a `PhysRegionId` that corresponds to different fields etc.?
[...]

Issue #538, opened 2022-11-01
Support different along-step kernels


sethrj:
Currently we use the same along-step kernel (including slowing down and field propagation) for all particle types, all energies, all regions. It's probably silly to use the field propagator to move a neutral particle (but at least it works!), we'll likely need to have different slowing-down models for different energy regimes, and we will perhaps someday need to support more extensible per-region parameters.

What I'd like to have is an efficient way of partitioning the multi-D {particle, energy, material} space into blocks of `ActionId`s (one per user along-step kernel). This will allow users to define easy host-side schemes to select a kernel (e.g., looping over particle types and setting different kernels for charged and neutral).

The `GridIdFinder` can help us map continuous variables (energy) to a discrete axis. We could define an `M × P × E` block to map to the action IDs, or we could add a level of indirection for the particle and material axes to group equivalent materials/particles into subspaces.

sethrj:
Note that we should probably add *regions* (set of volumes) instead of materials; and then refactor our cutoff/materials so that materials can be reused across regions with different cutoffs.

sethrj:
@amandalund This is what I had in mind for having separate MSC methods as a function of energy. With the speedup we saw from breaking the along-step kernels in to smaller kernels, though, we may want to group tracks by propagation sub-kernel first, and then have a secondary set of kernels for the different MSC regimes.

sethrj:
See also #459  for details on MSC refactor

sethrj:
Along-step construction:
- Fill "blocks" of R-P-E space
- Input block domain ("all" for each type if none specified):
  * Region list
  * Particle list, or function for masking particles (e.g. "charged" or
    "hadrons")
  * Energy grid (pre-step energy)
- Input attributes:
  * MSC model class
  * Propagator class
  * Along-step cuts (looping max, min particle energy)
  * Energy loss class (mean, fluctuation, none [neutral])

> **note**: energy grid could be "above low bound" so that we have a separate bin for at-rest tracks

> **question**:  under what circumstances do tracks enter the along-step part
> with zero energy? We have a `particle.is_stopped()` check at the start of the
> eloss applier.

Along-step action:
- Params data:
  - Map of `{region,particle,energy} -> {msc id,propagate id,eloss id}`
  - Flags for common behavior for the different kernels
- State data:
  - msc/propagate/eloss IDs
  - Propagation result (future work)
  - Energy deposition result (future work)
- Device: calls apply for all components
[...]

PR #1236, opened 2024-05-14
Add option to specify maximum number of substeps in field propagator

amandalund:
This allows the user to set the maximum number of field propagator substeps, which can have a significant impact on performance due to improved load balancing (see the plot below for the 32 ttbar CMS run3 results with three different `max_substep` values: 10, our default of 100, and Geant4's default of 1000).

I'm not really sure how best to safely access the user-specified `max_substeps` in the `SimParams`, which needs it to calculate the looping threshold values. In particular, through `accel` apps other than celer-g4 I'm not sure how to validate that the `max_substeps` in the setup options matches what's in the along-step factory. I'm also not entirely sure if changing this parameter has any consequence on the physics (msc?).

speedup-a100


sethrj (approved):
I think this is fine for now, thanks @amandalund . I think to overcome this issue, and solve https://github.com/celeritas-project/celeritas/issues/538 and https://github.com/celeritas-project/celeritas/issues/983, we should have a "propagation params" that manages the various possible along-step kernels and would handle the looping parameter, which currently is used only in `global/alongstep/detail/PropagationApplier.hh`.

sethrj:
Also, good work with figuring out this speedup! I wonder if we *should* adjust the default parameter based on this finding. I want to see whether the spikes in the time-per-step plot for CMS 2018+field disappear with this... which we have always suspected is due to looping.
time-per-step-cms2018+field+msc

amandalund:
I think adding a propagation params is a good idea. It would definitely be interesting to see how changing the max substeps affects that plot... I also need to experiment with adjusting this parameter in some of our other test problems.

amandalund:
@sethrj a different plot, but you can see how changing the max substeps affects the step times/variation in times. It definitely looks like those spikes are from looping tracks.
field-maxsubsteps

sethrj:
Whoa why does the maximum 10 cut off at 6000 but the 100 live much longer? Is it because we're clearing track slots for the looping tracks earlier?

amandalund:
It might just be a statistical fluctuation (though I was surprised too that the number of step iterations was so much hgher than with 10 or 1000)... I tried running the same problem with a different seed and it only took 5900 steps.

amandalund:
@sethrj a couple more observations:
- I ran 100 simulations with different seeds for each of these max substep values. Most of them finish in ~5500 step iterations, but there are some outliers that take many more. This is more of a problem when the substep limit is high: with 1000 max substeps there was one instance where the total number of step iterations was over 100k, with >95% of the steps having only one active track.
- It looks like this is because of occasional low energy e-/e+ tracks in low density materials taking many (thousands) of small steps (but never being flagged as looping). Even though this might be relatively uncommon, especially with a lower max substeps, we may still want a strategy for killing (or un-sticking?) these tracks.

total-step-iter

The last 20 steps of one of these tracks:
```console
track 16419589, vol 1985, mat 406, particle 1, energy 3.9923808482957686e-02, step 6949, looping 0, step length 1.3858658842098418e-02
track 16419589, vol 2751, mat 381, particle 1, energy 3.9850266767637074e-02, step 6950, looping 0, step length 4.7244629498538651e-02
track 16419589, vol 1985, mat 406, particle 1, energy 3.9814529158514730e-02, step 6951, looping 0, step length 1.8629256656536680e-01
track 16419589, vol 2751, mat 381, particle 1, energy 3.8257980851032464e-02, step 6952, looping 0, step length 4.9934856957369343e-02
track 16419589, vol 1985, mat 406, particle 1, energy 3.7893562258098423e-02, step 6953, looping 0, step length 7.3445125203530637e-01
track 16419589, vol 2751, mat 381, particle 1, energy 3.1588001694221136e-02, step 6954, looping 0, step length 3.0593442567106464e-02
track 16419589, vol 1985, mat 406, particle 1, energy 3.1159181280340318e-02, step 6955, looping 0, step length 3.0731719856365525e-01
track 16419589, vol 2751, mat 381, particle 1, energy 2.8875578978385202e-02, step 6956, looping 0, step length 9.8504012780081442e-03
track 16419589, vol 1985, mat 406, particle 1, energy 2.8764161298133990e-02, step 6957, looping 0, step length 3.1567462632067628e-01
track 16419589, vol 2752, mat 381, particle 1, energy 2.6652691197892056e-02, step 6958, looping 0, step length 1.0000004250248819e-06
track 16419589, vol 2752, mat 381, particle 1, energy 2.6652682076712491e-02, step 6959, looping 0, step length 2.1192793848139056e-13
track 16419589, vol 1985, mat 406, particle 1, energy 2.6652682076710559e-02, step 6960, looping 0, step length 2.0257033488857515e-01
track 16419589, vol 2751, mat 381, particle 1, energy 2.3507798832861856e-02, step 6961, looping 0, step length 1.4031100172018471e-01
track 16419589, vol 2751, mat 381, particle 1, energy 2.2150049606347002e-02, step 6962, looping 0, step length 1.7791966459317970e-01
track 16419589, vol 2751, mat 381, particle 1, energy 2.0742165110699450e-02, step 6963, looping 0, step length 2.1894447090401701e-02
track 16419589, vol 2751, mat 381, particle 1, energy 1.8300777647732802e-02, step 6964, looping 0, step length 5.7666532505831737e-02
track 16419589, vol 2751, mat 381, particle 1, energy 1.6257806981944168e-02, step 6965, looping 0, step length 1.7311275991309827e-01
track 16419589, vol 2751, mat 381, particle 1, energy 1.3513392719547898e-02, step 6966, looping 0, step length 1.3732918277251066e-01
track 16419589, vol 2751, mat 381, particle 1, energy 9.6172212560103210e-03, step 6967, looping 0, step length 1.6222676351235218e-01
track 16419589, vol 2751, mat 381, particle 1, energy 3.9073150446839004e-03, step 6968, looping 0, step length 4.9231860431891476e-02
```

sethrj:
Damn, it's looping between multiple volumes... maybe there's some heuristic with average step length/average energy loss per step that could be used...

PR #1237, opened 2024-05-14
Regularize quadric surfaces to deduplicate opposing surfaces

sethrj:
This fixes one of the two issues that could get in the way of deduplicating twisted surfaces. The sign of quadric equations are now regularized so that two opposite-signed quadric surfaces become the same quadric surface but with a flipped sense. ~~The quadrics are also *scaled* so that the largest coefficient is 1.~~ We don't do any scaling yet because that seems to affect the accuracy of the calculation...


test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -760,6 +766,84 @@ TEST_F(GenTrapTest, full2)
...
+        static char const expected_node[] = "all(+0, -1, +2, +3, -4, -6)";
+
+        EXPECT_EQ(expected_node, result.node);
+        EXPECT_VEC_SOFT_EQ((Real3{-inf, -1, -1}), result.exterior.lower());
+        EXPECT_VEC_SOFT_EQ((Real3{1, 1, 1}), result.exterior.upper());
+    }
+    {
+        // Scaled (broadened) right side with the same hyperboloid but
+        // different size
+        // TODO: the scaled GQ should be normalized
```
sethrj, line 817:
```suggestion
```

sethrj:
Ping @elliottbiondo / @mrguilima

elliottbiondo (approved):

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -764,6 +764,67 @@ TEST_F(GenTrapTest, full2)
...
+        EXPECT_VEC_SOFT_EQ((Real3{0.5, 1, 1}), result.exterior.upper());
+    }
+    {
+        // Right
+        auto result
+            = this->test(GenTrap(1,
+                                 {{0, -1}, {1, -1}, {1, 1}, {0, 1}},
+                                 {{0.5, -1}, {1, -1}, {1, 1}, {-0.5, 1}}));
+
+        static char const expected_node[] = "all(+0, -1, +2, -4, -6, -7)";
```
mrguilima: Why ae there no +/- expectations for surfaces 3 and 5?

src/orange/surf/SurfaceSimplifier.hh
```diff
@@ -79,8 +79,8 @@ class SurfaceSimplifier
              SimpleQuadric>
     operator()(SimpleQuadric const&);
 
-    // Quadric with no cross terms is simple
-    Optional<SimpleQuadric> operator()(GeneralQuadric const&);
+    // Quadric can be normalized or simplified
+    Optional<SimpleQuadric, GeneralQuadric> operator()(GeneralQuadric const&);
 
     //! Default: no simplifcation
```
mrguilima, line 85: -> simplification (typo: missing an 'i')

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -806,16 +806,16 @@ TEST_F(GenTrapTest, adjacent_twisted)
...
+        static char const expected_node[] = "all(+0, -1, +2, +3, -4, -6)";
+        static char const* const expected_surfaces[] = {"Plane: z=-1",
+                                                        "Plane: z=1",
+                                                        "Plane: y=-1",
+                                                        "GQuadric: {0,0,0} "
+                                                        "{0,0.5,0} {2,0.5,0} "
+                                                        "0",
+                                                        "Plane: y=1",
+                                                        "Plane: x=-1",
+                                                        "Plane: x=1"};
```
mrguilima: Previous formatting looks better, IMO.

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -806,16 +806,16 @@ TEST_F(GenTrapTest, adjacent_twisted)
...
-        static char const expected_node[] = "all(+0, -1, +2, -4, -6, -7)";
-        static char const* const expected_surfaces[]
-            = {"Plane: z=-1",
-               "Plane: z=1",
-               "Plane: y=-1",
-               "GQuadric: {0,0,0} {0,0.5,0} {2,0.5,0} 0",
-               "Plane: y=1",
-               "Plane: x=-1",
-               "Plane: x=1",
-               "GQuadric: {0,0,0} {0,-0.5,0} {-2,-0.5,0} 0"};
```
mrguilima: So this GQ disappeared because of the deduplication?  
If I don't expect it to disappear, I'd be debugging this like crazy to figure out why it did.
Would it be easy to print the original one instead?

src/orange/surf/GeneralQuadric.cc
```diff
@@ -17,12 +17,14 @@ namespace celeritas
 /*!
  * Construct with all coefficients.
  *
- * TODO: normalize so that largest eigenvalue is unity?
+ * TODO: normalize so that largest eigenvalue is unity? Or what? (It would be
+ * nice to have "slightly twisted planes" have order-epsilon cross terms as
+ * opposed to order 1/eps linear terms.)
```
mrguilima, line 22: I wonder whether normalizing such that the largest abs.value of the largest order coefficient becomes 1 would be a good choice?  
You've tried normalizing all coefficients, my suggestion is to normalize based only on the highest-order ones, rather than on all coefficients.

test/orange/orangeinp/ConvexRegion.test.cc
```diff
@@ -760,6 +762,85 @@ TEST_F(GenTrapTest, full2)
...
+     *
+     *
+     * Upper polygons:
+     *    x=-0.5
+     * +--+------+ y=1
+     * |   \     |
+     * |    \  R |
+     * |  L  \   |
+     * |      \  |
+     * +-------+-+ y=-1
```
mrguilima, line 785: How about saving some vertical space by having the Upper polygon sketched by the right-side of the Lower one?  Hope it always comes out in fixed-size fonts on the resulting docs!

mrguilima (commented):
Sorry for taking too long to review this.
Here are some comments.

test/orange/orangeinp/ConvexRegion.test.cc
sethrj: Those were the ones built from the earlier region in the same function (the unit builder and surfaces are persistent inside `TEST_F`).

test/orange/orangeinp/ConvexRegion.test.cc
sethrj: Yep that would've been smart.

test/orange/orangeinp/ConvexRegion.test.cc
sethrj: I can't tell what you're commenting on because this isn't a complete diff.

sethrj:
@mrguilima It looks like you keep somehow reviewing "partial" merge requests; please confirm that the upper left of the review box looks like this: 
Screenshot 2024-05-18 at 07 20 50

and not like this:
Screenshot 2024-05-18 at 07 20 08

This is why you've been having trouble making code suggestions.

I left the review open for several days with no response; if I ping for a review and someone else reviews it, please just leave a message saying you're still working on it. After the merge goes in, if you find errors in the documentation, please feel free to open a new pull request with the desired changes.

src/orange/surf/GeneralQuadric.cc
sethrj: I did try that actually, and reverted it because it introduced inaccuracies near the boundaries for cases where the surface is scaled so that the coefficients are large — e.g., an ellipsoid with 100,200,300 cm radii.

mrguilima:
Yeah, sometimes I feel like I am doing something not quite right...
Thanks for the suggestion, I will keep an eye on it.

PR #1238, opened 2024-05-16
Integrate Perfetto as backend for ScopedProfiling

### LINKED ISSUE #1224: Integrate Perfetto for optional CPU profiling
It looks like [Perfetto](https://github.com/google/perfetto), the Google performance/tracing tool, could be useful to integrate into Celeritas to measure performance. It might be straightforward to wrap their [tracing SDK](https://perfetto.dev/docs/instrumentation/tracing-sdk) underneath `ScopedProfiling`.

esseivaju:
Add Perfetto as an external dependency. It can be used as a backend for the existing `ScopedProfiling` when CUDA and HIP are disabled. The two main additions are:
*  ```celeritas::trace_counter``` to record the evolution of numeric values over time
* ```TracingSession``` as a wrapper around a perfetto session so that only corecel has to depend on Perfetto. A ```TracingSession``` needs to be created to record data when using Perfetto.

Since Perfetto system-level tracing requires Linux and root privileges, it can't be used on most clusters, so this also provides a Dockerfile for a development container that can be used on non-Linux hosts.


cmake/FindPerfetto.cmake
```diff
@@ -0,0 +1,13 @@
...
+    set(Perfetto_ROOT ${PERFETTO_ROOT} CACHE PATH "Perfetto.[cc|hh] directory")
+endif()
+
+find_path(Perfetto_INCLUDE_DIR perfetto.h
+            HINTS "${Perfetto_ROOT}")
+
+if(Perfetto_INCLUDE_DIR)
+    add_library(celeritas_perfetto STATIC ${Perfetto_INCLUDE_DIR}/perfetto.cc)
+    add_library(Celeritas::Perfetto ALIAS celeritas_perfetto)
+    target_include_directories(celeritas_perfetto INTERFACE ${Perfetto_INCLUDE_DIR})
```
sethrj: I think if we're going to add a library in our own namespace it shouldn't be in "find package"... maybe we should make an externals for like with g4vg: https://github.com/celeritas-project/g4vg/blob/main/external/CMakeLists.txt

src/corecel/CMakeLists.txt
```diff
@@ -123,6 +128,7 @@ celeritas_target_include_directories(corecel
   PUBLIC
     "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>"
     "$<BUILD_INTERFACE:${CELERITAS_HEADER_CONFIG_DIRECTORY}>"
+    "$<$<BOOL:${Perfetto_INCLUDE_DIR}>:${Perfetto_INCLUDE_DIR}>"
```
sethrj: Ideally the includes would be part of the `perfetto` target and would also be `PRIVATE`...

app/celer-sim/celer-sim.cc
```diff
@@ -76,7 +81,28 @@ int get_openmp_thread()
...
+        perfetto::TraceConfig cfg;
+        cfg.add_buffers()->set_size_kb(1024 * 512);
+        auto* ds_cfg = cfg.add_data_sources()->mutable_config();
+        ds_cfg->set_name("track_event");
+        ds_cfg->set_track_event_config_raw(track_event_cfg.SerializeAsString());
+        fd = open("celer-sim.perfetto-trace", O_RDWR | O_CREAT | O_TRUNC, 0660);
+        tracing_session->Setup(cfg, fd);
+        tracing_session->Start();
+    }
+#endif
```
sethrj: I assume this is more or less debugging/playing around for now? I think it would be sensible for all this code to go in ScopedProfiling... and perhaps even the initialization could go inside the `ScopedProfiling::use_profiling` IIFE ?

sethrj (commented):
Don't worry about comments until you have a sense that Perfetto could actually be useful and interesting for c++ development/profiling.

src/corecel/CMakeLists.txt
esseivaju: This should be fixed now

esseivaju:
I'm not sure why doxygen isn't happy with the documentation

src/corecel/sys/Counter.hh
```diff
@@ -8,27 +8,34 @@
...
-#include <string_view>
 #include <type_traits>
 
 #include "celeritas_config.h"
+#include "corecel/Macros.hh"
 
 namespace celeritas
 {
+/*!
+ * Simple tracing counter. Records a named value at the current timestamp which
```
sethrj:
```suggestion
//---------------------------------------------------------------------------//
/*!
 * Simple tracing counter.
 * Records a named value at the current timestamp which
```

src/corecel/sys/Counter.hh
```diff
@@ -8,27 +8,34 @@
...
+ * can then be displayed on a timeline. Only supported on host, this compiles
+ * but is a noop on device.
+ * See https://perfetto.dev/docs/instrumentation/track-events#counters
+ * @tparam T Arithmetic counter type
```
sethrj:
```suggestion
 * \tparam T Arithmetic counter type
```

cmake/FindPerfetto.cmake
sethrj: This is still true. Also if we're making a `STATIC` library doesn't it have to be after our global "everything is relocatable" setup?

scripts/docker/interactive/build.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+
+# shellcheck disable=SC2048
```
sethrj: I'm not familiar with `shellcheck`, is that something we should put in our CI? Also should we `set -e` and use `sh` instead of forcing `bash`?

scripts/docker/interactive/run-perfetto.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd "${PERFETTO_ROOT}" || exit
+
+tools/tmux -c celer.cfg -C out/linux -n
```
sethrj: Probably better to exec?
```suggestion
exec tools/tmux -c celer.cfg -C out/linux -n
```
Same comment about bash above.

scripts/docker/interactive/run.sh
```diff
@@ -0,0 +1,37 @@
...
+      ;;
+    esac
+done
+
+CELER_DIR=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")"/../../..)
+
+: "${TAG_NAME:=develop}"
+: "${HOST_MOUNT:=$CELER_DIR}"
+
+docker run --rm -it --mount source=celeritas_storage,target=/data \
```
sethrj: Same exec comment

scripts/docker/interactive/run.sh
```diff
@@ -0,0 +1,37 @@
...
+if ! args=$(getopt t:m: $*); then
+    echo "Usage: $0 [-t tag_name] [-m bind_mount]"
+    exit
+fi
+
+# shellcheck disable=SC2086
+set -- $args
+
+while :; do
+    case "$1" in
```
sethrj: Can't you use `while getopts`?

E.g.
```sh
OPTIND=1
while getopts "fb:p:G:" opt; do
  case "${opt}" in
    t) TAG_NAME =${OPTARG} ;;
    m) HOST_MOUNT ="${OPTARG}" ;;
    \?) echo "Invalid option: -$OPTARG" >&2; exit 1 ;;
    *) usage; exit 2; ;;
  esac
done
shift $((OPTIND-1))
```

src/corecel/CMakeLists.txt
```diff
@@ -92,6 +92,15 @@ if(CELERITAS_USE_OpenMP)
   list(APPEND PRIVATE_DEPS OpenMP::OpenMP_CXX)
 endif()
 
+if(CELERITAS_USE_Perfetto)
+  list(APPEND PRIVATE_DEPS Celeritas::Perfetto)
+  list(APPEND SOURCES
+          sys/Counter.perfetto.cc
+          sys/TracingSession.perfetto.cc
+          sys/ScopedProfiling.perfetto.cc
```
sethrj: indentation

src/corecel/sys/Counter.hh
```diff
@@ -0,0 +1,36 @@
...
+#if CELERITAS_USE_PERFETTO
+/*!
+ * Simple tracing counter. Records a named value at the current timestamp which
+ * can then be displayed on a timeline. Only supported on host, this compiles
+ * but is a noop on device.
+ * See https://perfetto.dev/docs/instrumentation/track-events#counters
+ * @tparam T Arithmetic counter type
+ */
+template<class T>
+CELER_FUNCTION void trace_counter(char const* name, T value);
```
sethrj, line 20: Should there be `extern template` declarations as well?

src/corecel/sys/ScopedProfiling.hh
```diff
@@ -12,10 +12,36 @@
...
 #include "corecel/Macros.hh"
+#include "corecel/io/Logger.hh"
+
+#include "Environment.hh"
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
+#if CELER_USE_DEVICE || CELERITAS_USE_PERFETTO
+// Whether profiling is enabled
```
sethrj, line 22: Add doxygen and note the env variable name

src/corecel/sys/ScopedProfiling.hh
```diff
@@ -12,10 +12,36 @@
...
+    static bool const result = [] {
+        if (!celeritas::getenv("CELER_ENABLE_PROFILING").empty())
+        {
+            CELER_LOG(info) << "Enabling profiling support since the "
+                               "'CELER_ENABLE_PROFILING' "
+                               "environment variable is present and non-empty";
+            return true;
+        }
+        return false;
+    }();
```
sethrj: I really would rather have the static data and getenv/logger calls in a .cc file. Maybe change `.perfetto.cc` to `.cc` and `#if` the perfetto parts?

src/corecel/sys/ScopedProfiling.hh
```diff
@@ -12,10 +12,36 @@
...
+    return result;
+}
+#else
+// Profiling is never enabled if CUDA/HIP/Perfetto isn't available
+constexpr inline bool use_profiling()
+{
+    return false;
+}
```
sethrj: I wonder if we should also check for the existence of the environment variable to warn users?

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -22,34 +22,6 @@
...
-{
-    static bool const result = [] {
-        if (!celeritas::getenv("CELER_ENABLE_PROFILING").empty())
-        {
-            if (!CELERITAS_HAVE_ROCTX)
-            {
-                CELER_LOG(warning) << "Disabling profiling support "
-                                      "since ROC-TX is unavailable";
-                return false;
-            }
```
sethrj, line 42: This warning code should go back in there for the device case

src/corecel/sys/ScopedProfiling.perfetto.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "corecel/io/Logger.hh"
+
+#include "Environment.hh"
+
+#include "detail/TrackEvent.perfetto.hh"
+
+namespace celeritas
+{
+
+/*!
```
sethrj:
```suggestion
{
//---------------------------------------------------------------------------//
/*!
```
and below

src/corecel/sys/TracingSession.hh
```diff
@@ -0,0 +1,95 @@
...
+    InProcess,  //!< Record in-process, writting to a file
+    System  //!< Record in a system daemon
+};
+
+#if CELERITAS_USE_PERFETTO
+/*!
+ * RAII wrapper for a tracing session.
+ *
+ * Constructors will only configure an initialize the session. It needs to
+ * be started explicitely by calling \c TracingSession::start
```
sethrj:
```suggestion
 * Constructors will only configure and initialize the session. It needs to
 * be started explicitly by calling \c TracingSession::start
```

src/corecel/sys/TracingSession.hh
```diff
@@ -0,0 +1,95 @@
...
+ * TODO: Support multiple tracing mode.
+ */
+class TracingSession
+{
+  public:
+    // Configure a system session recording to a daemon
+    TracingSession();
+    // Configure an in-process session recording to filename
+    explicit TracingSession(std::string_view filename);
+    // Terminate thte session and close open files
```
sethrj:
```suggestion
    // Terminate the session and close open files
```

src/corecel/sys/TracingSession.hh
```diff
@@ -0,0 +1,95 @@
...
+  public:
+    // noop
+    TracingSession() = default;
+    // noop
+    explicit TracingSession(std::string_view) {}
+
+    // noop
+    void start() {};
+};
+#endif
```
sethrj: Instead of declaring a second class, could you take the normal approach of providing `inline` `CELER_NOT_CONFIGURED (or null-op I guess?)` definitions?

src/corecel/sys/TracingSession.perfetto.cc
```diff
@@ -0,0 +1,136 @@
...
+    ds_cfg->set_track_event_config_raw(track_event_cfg.SerializeAsString());
+    return cfg;
+}
+//---------------------------------------------------------------------------//
+}  // namespace
+
+namespace celeritas
+{
+
+/*!
```
sethrj: Could you please recheck the merge request for formatting and proper doxygen markup? There are a lot of inconsistencies with the rest of the codebase.

src/corecel/sys/TracingSession.perfetto.cc
```diff
@@ -0,0 +1,136 @@
...
+TracingSession::TracingSession(std::string_view filename)
+    : session_{initialize_session(TracingMode::InProcess)}, fd_{[&] {
+        return session_
+                   ? open(filename.data(), O_RDWR | O_CREAT | O_TRUNC, 0660)
+                   : -1;
+    }()}
+{
+    if (session_)
+    {
+        session_->Setup(configure_session(), fd_);
```
sethrj: Since the fd already has a default initializer to -1, I think this is far more readable:
```suggestion
    : session_{initialize_session(TracingMode::InProcess)}
{
    if (session_)
    {
        fd_ = open(filename.data(), O_RDWR | O_CREAT | O_TRUNC, 0660);
        session_->Setup(configure_session(), fd_);
```

src/corecel/sys/detail/TrackEvent.perfetto.hh
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include <perfetto.h>
+namespace celeritas
+{
+namespace detail
+{
+// Global category name for celeritas
+// TODO: add more categories for fine-grained control of events to record
+constexpr auto* perfetto_track_event_category{"Celeritas"};
```
sethrj: Doesn't this need a static, inline, or a `.cc` file?

sethrj (changes requested):

src/corecel/sys/detail/TrackEvent.perfetto.hh
esseivaju: Right, I forgot inline. I don't think static is needed though.

src/corecel/sys/ScopedProfiling.hh
esseivaju: We can't make that function constexpr if we want to do this here.

I updated it but now the compiler can't tell that constructing a ScopedProfiling without CUDA/HIP/Perfetto is a no-op.

src/corecel/sys/TracingSession.hh
esseivaju: I think in that case, null-op is fine, maybe add a warning here is the environment variable is set.

src/corecel/sys/Counter.hh
esseivaju: I'm not sure it's needed: the definition is not accessible from the header so we have to rely on explicit instantiation anyway or we get a linker error

scripts/docker/interactive/build.sh
esseivaju: I forced bash instead of sh because I'm using `BASH_SOURCE` to find the script directory but we can change it to `dirname $0` it will work in most cases. 

I like to use `shellcheck` since bash has many pitfalls 😅 I suppose it wouldn't hurt to have it in the CI.

cmake/FindPerfetto.cmake
esseivaju: I introduced it in our namespace because it gets compiled with Celeritas but I'm not sure that's the best option. We're handling all our dependencies using find_package, so I also went with that.

This is after the relocatable setup no? Perfetto.cc gets compiler with `-fPIC`

src/corecel/sys/ScopedProfiling.hh
esseivaju: Ok, done. However, I like the inline static more than the conditional compilation. What's the reason for preferring logger calls in cc? compile time?

esseivaju:
Here's an example of application and system traces: [traces.tar.gz](https://github.com/celeritas-project/celeritas/files/15441356/traces.tar.gz)

I've noticed that on my laptop (16 cores) if I use the in-process mode and max out the CPU with 16 OpenMP threads, some data is lost. It doesn't happen when using the system mode.

CMakeLists.txt
```diff
@@ -331,15 +334,42 @@ if(CELERITAS_USE_OpenMP)
...
+              "Make sure that you downloaded an amalgamated release")
+    endif()
+  elseif(CELERITAS_PERFETTO_DOWNLOAD)
+    include(FetchContent)
+    FetchContent_Declare(
+            perfetto
+            GIT_REPOSITORY https://android.googlesource.com/platform/external/perfetto
+            GIT_TAG 7add4fded398ef47414a556b7b81d212a4482a66 # v45.0
+    )
+    FetchContent_MakeAvailable(perfetto)
```
esseivaju: can we require a newer CMake version (3.14) or is 3.12 still important to support?

esseivaju:
@sethrj I'll update the CMake integration once #1253 is merged. Other than that it should be ready for a second look.

CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ celeritas_optional_package(OpenMP "Enable CPU shared-memory parallelism")
 celeritas_optional_package(Python "Use Python to generate and preprocess")
 celeritas_optional_package(ROOT "Enable ROOT I/O")
 celeritas_optional_package(VecGeom "Use VecGeom geometry")
+celeritas_optional_package(Perfetto "Perfetto tracing library")
```
sethrj: Since it sounds like perfetto isn't ever externally installable, I think this should just be an `option(CELERIITAS_USE_Perfetto` like we (will) have for JSON.

CMakeLists.txt
```diff
@@ -329,6 +330,38 @@ if(CELERITAS_USE_OpenMP)
...
+  FetchContent_Declare(
+    perfetto
+    # v45 .0
+    URL https://android.googlesource.com/platform/external/perfetto/+archive/7add4fded398ef47414a556b7b81d212a4482a66.tar.gz
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+  )
+
+  if (NOT FETCHCONTENT_SOURCE_DIR_PERFETTO)
+    message(STATUS "Fetching perfetto")
+  endif()
```
sethrj: Does this show up only on the first configure or on every configure? (Also for cmake use `if(` instead of `if (` since it's a function 🙄 )

app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -161,6 +162,7 @@ void to_json(nlohmann::json& j, RunnerInput const& v)
     LDIO_SAVE_WHEN(primary_options, v.event_file.empty());
 
     LDIO_SAVE_OPTION(mctruth_file);
+    LDIO_SAVE_OPTION(tracing_file);
```
sethrj: Probably `LDIO_SAVE_WHEN(tracing_file, CELERITAS_USE_PERFETTO)`?

app/celer-sim/celer-sim.cc
```diff
@@ -87,6 +87,12 @@ void run(std::istream* is, std::shared_ptr<OutputRegistry> output)
 #else
     CELER_ASSERT_UNREACHABLE();
 #endif
+    TracingSession tracing_session{
+        run_input->tracing_file.empty()
+            ? TracingSession{}
+            : TracingSession{run_input->tracing_file}};
```
sethrj: Maybe the `tracing_file` run input should have the default parameter (if perfetto is active) or the constructor could be simplified?

src/corecel/sys/ScopedProfiling.cc
```diff
@@ -0,0 +1,78 @@
...
+
+#if CELERITAS_USE_PERFETTO
+#    include <perfetto.h>
+
+#    include "detail/TrackEvent.perfetto.hh"
+#endif
+
+namespace celeritas
+{
+#if !CELER_USE_DEVICE
```
sethrj: The `use_profiling` could probably be combined with the `.cuda.cc` version: we're just using a different profiler. Or maybe at this point we should just merge everything into a single file... thoughts?

src/corecel/sys/ScopedProfiling.hh
```diff
@@ -47,6 +54,9 @@ struct ScopedProfilingInput
  *
  * \note The AMD roctx implementation requires the roctx library, which may not
  * be available on all systems.
+ *
+ * \note The CPU implementation requires Perfetto. It is not supported when
+ * Celeritas is build with device support (CUDA/HIP)
```
sethrj:
```suggestion
 * Celeritas is built with device support (CUDA/HIP)
```

src/corecel/sys/ScopedProfiling.hip.cc
```diff
@@ -21,20 +21,19 @@
...
-
 //---------------------------------------------------------------------------//
 /*!
  * Whether profiling is enabled.
  *
  * This is true only if the \c CELER_ENABLE_PROFILING environment variable is
  * set to a non-empty value.
  */
-bool ScopedProfiling::use_profiling()
+bool use_profiling()
```
sethrj: Possibly could also be combined into the `.cc` file

src/corecel/sys/TracingSession.cc
```diff
@@ -0,0 +1,154 @@
...
+#include "ScopedProfiling.hh"
+
+#include "detail/TrackEvent.perfetto.hh"
+
+PERFETTO_TRACK_EVENT_STATIC_STORAGE();
+
+namespace
+{
+//---------------------------------------------------------------------------//
+using celeritas::TracingMode;
```
sethrj, line 26: Is there a reason not to just put this anonymous namespace inside the celeritas one like we usually do?

sethrj (commented):
This is really cool, I think this would be worth at least a poster session at CHEP or the like!

CMakeLists.txt
sethrj: I'm comfortable with requiring up to 3.20 (3 year support window).

CMakeLists.txt
esseivaju: it shows every time, I removed it

app/celer-sim/celer-sim.cc
esseivaju: I moved that check inside the constructor, a bit more ergonomic to use that way

src/corecel/sys/ScopedProfiling.cc
esseivaju: we can definitely combine `use_profiling`. I think I still like having separate files for the other functions, it makes it easier to find the implementation you're looking for, if anything I'd like to split `use_profiling` in a separate file and remove the `#if CELERITAS_USE_PERFETTO` in the `.cc`

sethrj (approved):
Looks good to me! We can wait till #1253 is merged and then move the external fetching into `external/`.

CMakeLists.txt
```diff
@@ -321,31 +321,15 @@ if(CELERITAS_USE_OpenMP)
...
 if(CELERITAS_USE_Perfetto)
-  include(FetchContent)
   if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
     celeritas_error_incompatible_option(
       "Perfetto is not supported with CUDA/HIP build"
       CELERITAS_USE_Perfetto
       OFF
     )
+  else()
+    celeritas_find_or_external_package(Perfetto)
```
esseivaju, line 331: Should we provide a FindPerfetto module here? Otherwise, find_package will never succeed since Perfetto doesn't support CMake.

sethrj: I guess we could, is it something users even have installed externally? I'm fine with `find_package` never succeeding....

sethrj (approved):

CMakeLists.txt
esseivaju: No, it shouldn't be externally installed.  Having ```FETCHCONTENT_SOURCE_DIR_PERFETTO``` use a local checkout instead of fetching it is fine. I should have added QUIET to the find_or_external call as we get a warning for not having a module and config file.

PR #1239, opened 2024-05-17
Update single and multiple Coulomb scattering configurations in `GeantPhysicsList`

amandalund:
- Allow single Coulomb scattering without MSC over all energies or with Wentzel VI MSC with the same energy limits
- Update the MSC model selection to be more consistent with the configurations in the Geant4 physics constructors

### DIFF
--- a/src/celeritas/ext/GeantPhysicsOptions.cc
+++ b/src/celeritas/ext/GeantPhysicsOptions.cc
@@ -35,10 +35,9 @@ char const* to_cstring(MscModelSelection value)
     static EnumStringMapper<MscModelSelection> const to_cstring_impl{
         "none",
         "urban",
-        "urban_extended",
-        "wentzel_vi",
-        "urban_wentzel",
-        "goudsmit_saunderson",
+        "wentzelvi",
+        "urban_wentzelvi",
+        "gs_wentzelvi",
     };
     return to_cstring_impl(value);
 }
--- a/src/celeritas/ext/GeantPhysicsOptions.cc
+++ b/src/celeritas/ext/GeantPhysicsOptions.cc
@@ -37,7 +37,6 @@ char const* to_cstring(MscModelSelection value)
         "urban",
         "wentzelvi",
         "urban_wentzelvi",
-        "gs_wentzelvi",
     };
     return to_cstring_impl(value);
 }
--- a/src/celeritas/ext/GeantPhysicsOptions.hh
+++ b/src/celeritas/ext/GeantPhysicsOptions.hh
@@ -30,8 +30,7 @@ enum class MscModelSelection
     none,
     urban,  //!< Urban for all energies
     wentzelvi,  //!< Wentzel VI for all energies
-    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI abpve
-    gs_wentzelvi,  //!< Goudsmit-Saunderson below 100 MeV, Wentzel VI above
+    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI above
     size_
 };
 
--- a/src/celeritas/ext/GeantPhysicsOptions.hh
+++ b/src/celeritas/ext/GeantPhysicsOptions.hh
@@ -28,11 +28,10 @@ enum class BremsModelSelection
 enum class MscModelSelection
 {
     none,
-    urban,
-    urban_extended,  //!< Use 100 TeV as upper bound instead of 100 MeV
-    wentzel_vi,
-    urban_wentzel,  //!< Urban for low-E, Wentzel_VI for high-E
-    goudsmit_saunderson,
+    urban,  //!< Urban for all energies
+    wentzelvi,  //!< Wentzel VI for all energies
+    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI abpve
+    gs_wentzelvi,  //!< Goudsmit-Saunderson below 100 MeV, Wentzel VI above
     size_
 };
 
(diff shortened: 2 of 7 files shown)


src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -28,11 +28,10 @@ enum class BremsModelSelection
...
 {
     none,
-    urban,
-    urban_extended,  //!< Use 100 TeV as upper bound instead of 100 MeV
-    wentzel_vi,
-    urban_wentzel,  //!< Urban for low-E, Wentzel_VI for high-E
-    goudsmit_saunderson,
+    urban,  //!< Urban for all energies
+    wentzelvi,  //!< Wentzel VI for all energies
+    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI abpve
```
sethrj:
```suggestion
    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI above
```

src/celeritas/ext/detail/GeantPhysicsList.cc
```diff
@@ -340,58 +335,94 @@ void GeantPhysicsList::add_e_processes(G4ParticleDefinition* p)
...
-        double msc_energy_limit = G4EmParameters::Instance()->MscEnergyLimit();
-
-        auto process = std::make_unique<G4CoulombScattering>();
-        auto model = std::make_unique<G4eCoulombScatteringModel>();
-        process->SetMinKinEnergy(msc_energy_limit);
-        model->SetLowEnergyLimit(msc_energy_limit);
-        model->SetActivationLowEnergyLimit(msc_energy_limit);
-        process->SetEmModel(model.release());
-        physics_list->RegisterProcess(process.release(), p);
+        if (options_.msc == MMS::urban)
```
stognini, line 347: Maybe we should decouple single and multiple scattering for registering processes. Our `GeantPhysicsOptions` has `coulomb_scattering` and `msc` as separate fields, so I feel like it also makes sense to have something like

```cpp
if (options_.coulomb_scattering)
{
    // Enable Coulomb Scattering
}
if (options_.msc)
{
    // Register physics according to selection and issue warnings for energy ranges as needed
}
```

amandalund: I think if we have both Coulomb and Wentzel VI enabled we should be using the same energy range for both, and if if only single Coulomb scattering is enabled it should cover the entire energy range. I'm not sure it makes sense to have Coulomb scattering enabled in the same energy range as either Urban MSC or GS (although I suppose we could allow it and issue a warning?). Maybe @whokion can comment?

sethrj (approved):
Code looks good to me, but I'll defer to one of the real physicists for correctness. Would it be possible for you to add a subsection under `physics` in `doc/main/overview.rst` describing a little bit better the complementary (and exclusive) coulomb scattering models?

src/celeritas/ext/detail/GeantPhysicsList.cc
whokion: What @amandalund  propose is consistent with practical user cases and the current Geant4 default physics list support.  In principle, SS (exclusively without any other other msc model) in all energy ranges is the most accurate  Coulomb scattering model, but may not be used in any LHC-like HEP production as it is computational very expensive (due to the high cross section in typical HEP calorimeters). So, it is mostly used for the purpose of model developments (i.e., to benchmark accuracy of other MSC models or validation) or special studies for detector design or optimization including back scatterings - one of exceptional examples of using SS alone in all range is in a sparse or low density medium (such as gaseous material).  The reason that SS is combined in the same energy range of the high energy MSC model is a kind of trade-off of adding the more accuracy with an affordable additional computing cost as the SS in the low energy is very prohibitively expensive - of course, the question whether the same energy cutoff (100MeV) with the high energy (WentzelVI) MSC model should be used or not is a practical choice for conveniences for the simpler setup for physics processes or lists.   As a summary, it will be good to support 1) [Urbarn]+[[WentzelVI]+[SS]] (Geant4 default), 2) [Urban] + [WentzelVI] (a.k.a. CMS EMM), 3) [SS] in the all range - of course [Urbarn] can be substituted by [GS] if GS is implemented and the more accurate simulation is needed for the additional cost of computing time.

src/celeritas/ext/GeantPhysicsOptions.cc
```diff
@@ -35,10 +35,9 @@ char const* to_cstring(MscModelSelection value)
     static EnumStringMapper<MscModelSelection> const to_cstring_impl{
         "none",
         "urban",
-        "urban_extended",
-        "wentzel_vi",
-        "urban_wentzel",
-        "goudsmit_saunderson",
+        "wentzelvi",
+        "urban_wentzelvi",
+        "gs_wentzelvi",
```
whokion: This may be dropped for now and added later when GS is actually implemented.

whokion (commented):
As long as MSC concerns, we may have logical combinations only for [none], [Urban] or [Urban] + [WentzelVI] for now as GS is not available and  SS can be configured/added as a separately process in physics lists.

src/celeritas/ext/detail/GeantPhysicsList.cc
amandalund: Thanks for the detailed answer @whokion!

src/celeritas/ext/GeantPhysicsOptions.cc
amandalund: True, I can drop GS for now. I may keep Wentzel VI in addition to Urban + Wentzel VI even though it may not be used alone in practice since it might be helpful in the (hopefully not too distant) future for validating that model separately from Urban.

amandalund:
> Would it be possible for you to add a subsection under `physics` in `doc/main/overview.rst` describing a little bit better the complementary (and exclusive) coulomb scattering models?

No problem!

sethrj (approved):
Perfect! Thanks for the overview.

PR #1241, opened 2024-05-17
Fix Spack CI

sethrj:
Replaces part of #1240. This is a quick update due to https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240514.2 which removed GCC 13, which our spack build implicitly depended on (via the libc++ installation apparently).

### DIFF
--- a/scripts/spack.yaml
+++ b/scripts/spack.yaml
@@ -2,7 +2,7 @@ spack:
   specs: 
     - cmake
     - doxygen
-    - "geant4@11 cxxstd=17"
+    - "geant4@11"
     - git
     - git-lfs
     - "googletest@1.10:"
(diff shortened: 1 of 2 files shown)


amandalund (approved):

scripts/spack.yaml
```diff
@@ -16,9 +16,9 @@ spack:
     - py-sphinx
     - py-sphinxcontrib-bibtex
     - py-sphinxcontrib-mermaid
-    - "root@6.24: cxxstd=17"
+    - "root@6.28:"
     - "swig@4.1:"
-    - "vecgeom@1.2.4: +gdml cxxstd=17"
+    - "vecgeom@1.2.4: +gdml"
```
mrguilima, line 21: Would it be too difficult to bump VecGeom version up to 1.2.8?
There are some important bug fixes after 1.2.4.

sethrj: This is just the minimum required to build as specified in the CMake file; Spack will build the newest it has. We have unit tests showing the failures before 1.2.8 in case anyone *does* try to build with an older version.

PR #1242, opened 2024-05-18
Add custom FTFP BERT physics list with Celeritas EM physics constructor

amandalund:
This adds a custom FTFP BERT physics list in Celeritas with the electron, positron, and gamma EM standard physics replaced by a EM physics constructor with Celeritas-supported physics. This will allow us to do more fair comparisons with our Geant4-integrated apps (most importantly, using Urban MSC for all energies).


src/celeritas/ext/detail/CelerEmStandardPhysics.cc
```diff
@@ -0,0 +1,396 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2020-2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/ext/detail/CelerEmStandardPhysics.cc
```
amandalund, line 6: I'm not sure how to make git recognize this is a rename of GeantPhysicsList.cc (with only a couple small changes).

sethrj: Git's a bit tricky about that (and I don't think there's fundamentally any metadata that says it's that way). The new class name is probably also messing up the diff. I wouldn't worry about it; I think it'll be obvious enough when viewing the history... the only thing to do would be to have separate commits for "move file", "find-replace name", and "actually change stuff".... but that would only really show up in the review rather than the final squash.

amandalund: It seems to have figured it out when I renamed the old physics list.

sethrj (approved):
LGTM, thanks!

amandalund:
@sethrj FYI I added a couple helper methods to get this working with older Geant4 versions.

sethrj:
Looks like this conflicts a bit with #1239...

PR #1243, opened 2024-05-18
Improve ccache use on github workflows

sethrj:
This PR:
- fixes ccache invalidation due to different runner host names
- increases ccache size for windows which I think is causing issues with invalidation
- adds a cron job so that every sunday morning all ccache entries are cleared and new ones regenerated
- Updates the `pr` filename for consistency with the workflow trigger
- Update the `merge_group` workflow so we can try it again


.github/workflows/clear-cache.yml
```diff
@@ -0,0 +1,29 @@
+name: clear-cache
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  clear-cache:
+    runs-on: ubuntu-latest
```
pcanal, line 9: Should it run on all the platform with CCache? (I.e. eg Windows?)

sethrj: Nope. All it's doing is making GitHub calls; the cache itself is platform agnostic.

sethrj:
Ping @pcanal

sethrj:
Ping @pcanal

PR #1244, opened 2024-05-18
Add external bounding box to generic trapezoid

sethrj:
As part of this I added an improved `grow` function for the bbox and realized while doing so that the existing `min/max` that  we've defined is suboptimal for certain architectures (viz., ARM) and that the CUDA floating point `min/max` are exactly equal to `fmin/fmax` which is supported across architectures.


src/orange/orangeinp/ConvexRegion.cc
```diff
@@ -487,6 +487,22 @@ void GenTrap::build(ConvexSurfaceBuilder& insert_surface) const
...
+        for (Real2 const& xy : *p)
+        {
+            for (auto ax : {Axis::x, Axis::y})
+            {
+                exterior_bbox.grow(ax, xy[to_int(ax)]);
+            }
+        }
+    }
+    exterior_bbox.grow(Axis::z, -hz_);
+    exterior_bbox.grow(Axis::z, hz_);
```
mrguilima: IIUC, it is okay to save some cycles this way:
```suggestion
    exterior_bbox.grow(Bound::lo, Axis::z, -hz_);
    exterior_bbox.grow(Bound::hi, Axis::z, hz_);
```

mrguilima (approved):
LGTM.
Thanks!

PR #1245, opened 2024-05-20
Rename ConvexRegion to IntersectRegion

sethrj:
After implementing twisted faces (#1232 ) the convex region is incorrectly named: it's really a *conjunction* (intersection) of half-spaces rather than a convex region. The important thing is that *crossing any surface exits the region*, and the nearest distance to any surface is the nearest distance to the region's boundary.

I've therefore renamed `ConvexRegion` to `IntersectRegion`. I considered "conjunctive" , "intersection", and "isect", and think this is the best combination of clear, correct, and concise.


mrguilima:
Is it true that `Shapes` are meant to be convex, and `Solids` may be made of unions of `Shapes`?
If so, and once this concept is well explained, maybe the ConvexRegion could be renamed as a ShapeRegion?
Just a suggestion, in case you're not so happy with your choice of `IntersectRegion`.

src/orange/orangeinp/Shape.hh
```diff
@@ -58,7 +58,7 @@ class ShapeBase : public ObjectInterface
 
 //---------------------------------------------------------------------------//
 /*!
- * Shape that holds a convex region and forwards construction args to it.
+ * Shape that holds a intersect region and forwards construction args to it.
```
mrguilima:
```suggestion
 * Shape that holds an intersect region and forwards construction args to it.
```

src/orange/orangeinp/detail/BuildIntersectRegion.cc
```diff
@@ -23,21 +23,21 @@ namespace detail
 {
 //---------------------------------------------------------------------------//
 /*!
- * Build a convex region.
+ * Build a intersect region.
```
mrguilima:
```suggestion
 * Build an intersect region.
```

src/orange/orangeinp/IntersectSurfaceBuilder.cc
```diff
@@ -182,7 +182,7 @@ void ConvexSurfaceBuilder::grow_interior(BBox const& bbox)
 /*!
  * Apply a convex surface builder to an unknown type.
```
mrguilima:
```suggestion
 * Apply an intersect surface builder to an unknown type.
```

src/orange/orangeinp/detail/BuildIntersectRegion.hh
```diff
@@ -15,24 +15,24 @@ namespace celeritas
...
+NodeId build_intersect_region(VolumeBuilder& vb,
+                              std::string&& label,
+                              std::string&& face_prefix,
+                              IntersectRegionInterface const& region);
 
-//! Build a convex region with no face prefix
-inline NodeId build_convex_region(VolumeBuilder& vb,
-                                  std::string&& label,
-                                  ConvexRegionInterface const& region)
+//! Build a intersect region with no face prefix
```
mrguilima:
```suggestion
//! Build an intersect region with no face prefix
```

mrguilima (changes requested):
LGTM, except for some article adjustments.
Thanks!

mrguilima (approved):
LGTM. Thanks!

PR #1246, opened 2024-05-21
Minor refactoring/niceties for upcoming polycone work

sethrj:
- Add structured binding support to `celeritas::array` by implementing tuple methods
- Move `monotonic_decreasing` and VectorUtils to corecel and genericize the underlying algorithm along the lines of "any of"
- Expose geometry tolerance to surface builder so that we can use "soft equivalence" for twisted faces

PR #1247, opened 2024-05-22
Implement polycone

sethrj:
**Chained on #1246**

This adds the notorious "polycone" shape (stacked cones/cylinders), at least for inside/outside testing and distance calculation. It allows cylinders to be constructed as degenerate cones since that's a common pattern.


test/orange/g4org/SolidConverter.test.cc
```diff
@@ -156,7 +156,7 @@ TEST_F(SolidConverterTest, cons)
 {
     this->build_and_test(
         G4Cons("Solid TubeLike #1", 0, 50, 0, 50, 50, 0, 360),
-        R"json({"_type":"shape","interior":{"_type":"cylinder","halfheight":5.0,"radius":5.0},"label":"Solid TubeLike #1"})json",
+        R"json({"_type":"shape","interior":{"_type":"cone","halfheight":5.0,"radii":[5.0,5.0]},"label":"Solid TubeLike #1"})json",
```
mrguilima, line 159: Is a type=cone correct in this case, where both radii are 5.0?

src/orange/orangeinp/PolySolid.hh
```diff
@@ -0,0 +1,174 @@
...
+namespace celeritas
+{
+namespace orangeinp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Radial extents and axial segments for a stacked solid.
+ *
+ * Axial grid points must be nondecreasing. If "inner" points are specified,
+ * they must be
```
mrguilima: Incomplete sentence.

src/orange/orangeinp/PolySolid.cc
```diff
@@ -0,0 +1,252 @@
...
+                                              detail::VolumeBuilder& vb,
+                                              NodeId result)
+{
+    if (auto const& sea = base.enclosed_angle())
+    {
+        // The enclosed angle is "true" (specified by the user to truncate the
+        // shape azimuthally): construct a wedge to be added or deleted
+        auto&& [sense, wedge] = sea.make_wedge();
+        NodeId wedge_id
+            = build_intersect_region(vb, base.label(), "angle", wedge);
```
mrguilima, line 102: Trying to understand the little differences:  why not ".wedge" instead of "angle"?
```suggestion
            = build_intersect_region(vb, base.label(), ".wedge", wedge);
```

src/orange/orangeinp/PolySolid.cc
```diff
@@ -0,0 +1,252 @@
...
+                   << z_.size() << ")");
+    CELER_VALIDATE(outer_.size() == z_.size(),
+                   << "inconsistent outer radius size (" << outer_.size()
+                   << "): expected " << z_.size());
+    CELER_VALIDATE(inner_.empty() || inner_.size() == z_.size(),
+                   << "inconsistent inner radius size (" << inner_.size()
+                   << "): expected " << z_.size());
+
+    CELER_VALIDATE(is_monotonic_nondecreasing(make_span(z_)),
+                   << "axial grid is not monotonically increasing");
```
mrguilima:
```suggestion
                   << "axial grid is not monotonically nondecreasing");
```

mrguilima (dismissed):
Looks great to me...  I do see a broken test in IntersectRegion, ConeTest.errors though, in my local build:
```
[ RUN      ] ConeTest.errors
/work1/celeritas/lima/work/cele/cele/test/orange/orangeinp/IntersectRegion.test.cc:187: Failure
Expected: Cone({1.0, 1.0}, 0.5) throws an exception of type RuntimeError.
  Actual: it throws nothing.
[  FAILED  ] ConeTest.errors (0 ms)
```

test/orange/g4org/SolidConverter.test.cc
sethrj: This is to simplify the polycone construction; instead of having to construct a special case, we might as well "identify" it as a degenerate cylinder if that's what the user's asked for....

src/orange/orangeinp/PolySolid.hh
sethrj: 🐿️

src/orange/orangeinp/PolySolid.cc
sethrj: The `.` gets joined anyway; the extra `.` above is because it additionally has the integer segment inside it. So the surfaces will be `polycone.1.interior` or `polycone.angle`. The shape isn't always a wedge (it's sometimes an anti-wedge) and the name is consistent with the Solid.

src/orange/orangeinp/PolySolid.cc
```diff
@@ -0,0 +1,252 @@
...
+    , segments_{std::move(segments)}
+    , enclosed_{std::move(enclosed)}
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return a polycone *or* a simplified version for only a single segment.
+ */
+auto PolyCone::or_solid(std::string&& label,
```
elliottbiondo, line 172: Personally I think `solid_or_shape` would be better than `or_solid`, going all the way back up to `Solid<T>::or_shape`.

elliottbiondo (approved):

src/orange/orangeinp/PolySolid.cc
sethrj: "Polycone or solid or shape"? 😉

src/orange/orangeinp/PolySolid.cc
```diff
@@ -0,0 +1,252 @@
...
+                   << z_.size() << ")");
+    CELER_VALIDATE(outer_.size() == z_.size(),
+                   << "inconsistent outer radius size (" << outer_.size()
+                   << "): expected " << z_.size());
+    CELER_VALIDATE(inner_.empty() || inner_.size() == z_.size(),
+                   << "inconsistent inner radius size (" << inner_.size()
+                   << "): expected " << z_.size());
+
+    CELER_VALIDATE(is_monotonic_nondecreasing(make_span(z_)),
+                   << "axial grid has increasing grid points");
```
mrguilima, line 145:
```suggestion
                   << "axial grid has decreasing grid points");
```

src/orange/orangeinp/PolySolid.hh
```diff
@@ -0,0 +1,175 @@
...
+{
+namespace orangeinp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Radial extents and axial segments for a stacked solid.
+ *
+ * Axial grid points must be nondecreasing. If "inner" points are specified,
+ * they must be less than the outer points and more than zero. The inner list
+ * is allowed to be empty indictating no inner (hollow) exclusion.
```
mrguilima, line 28:
```suggestion
 * is allowed to be empty indicating no inner (hollow) exclusion.
```

src/orange/orangeinp/PolySolid.cc
sethrj: Nope, "is not all nondecreasing" = "has increasing"

mrguilima: I respecfully disagree: increasing points is what is required, so it cannot be the message explaining a failed VALIDATE.
OTOH, the VALIDATE will fail if at least one i+1 point is actually decreasing w.r.t point i.  Otherwise the VALIDE will pass.

mrguilima: How about "... has one or more decreasing grid point(s)" ?

sethrj: It's possible to have two consecutive equal (non increasing) points; this is commonly used for the "stacked cylinders"

mrguilima: Yes, I know.  Still, my last two comments above hold.

sethrj: Ok I'll take a closer look 😅

sethrj: Goodness I am so obtuse. I am so sorry for doubting you, I clearly have not had enough sleep!! I'll fix it in the polyprism pull request.

PR #1249, opened 2024-05-23
Add support for using the build directory as an installation

sethrj:
This feature is based on a capability for Geant4 I learned from @whokion  (thanks!) and allows more rapid iteration when developing tightly integrated downstream applications such as [hgcal](https://github.com/celeritas-project/hgcal). You can now add the Celeritas build dir to your cmake prefix path and it will "just work".


drbenmorgan (approved):
Nice add @sethrj, LGTM!

PR #1251, opened 2024-05-24
Add polyprism (G4Polyhedra)

sethrj:
The G4Polyhedra is a z-aligned stack of regular polygon shapes. Currently this doesn't support sloped sides, but we can if needed at some point. I haven't found any of our geometry models that have sloped sides.

### DIFF
--- a/src/orange/orangeinp/PolySolid.cc
+++ b/src/orange/orangeinp/PolySolid.cc
@@ -142,7 +142,7 @@ PolySegments::PolySegments(VecReal&& inner, VecReal&& outer, VecReal&& z)
                    << "): expected " << z_.size());
 
     CELER_VALIDATE(is_monotonic_nondecreasing(make_span(z_)),
-                   << "axial grid has increasing grid points");
+                   << "axial grid has decreasing grid points");
     for (auto i : range(outer_.size()))
     {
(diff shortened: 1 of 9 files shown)


src/orange/orangeinp/PolySolid.cc
```diff
@@ -247,6 +247,117 @@ void PolyCone::output(JsonPimpl* j) const
...
+                                           orientation);
+    }
+
+    auto const [zlo, zhi] = segments.z(0);
+    real_type const hh = (zhi - zlo) / 2;
+
+    auto const [rlo, rhi] = segments.outer(0);
+    if (rlo != rhi)
+    {
+        CELER_NOT_IMPLEMENTED("prism with different lo/hi radii");
```
mrguilima, line 276: Is the plan to extend this PolyPrism class to handle segments with different lo/hi radii?
Or those will be handled by a different class (e.g. Polyhedron)?

mrguilima (approved):
It looks great!  
Are the tests for complex polyhedra (Nsegments > 1) coming later?

src/orange/orangeinp/PolySolid.cc
sethrj: I think we should extend the prism, but a cursory look showed none of our detector models had that feature...

PR #1252, opened 2024-05-24
Add gdml-to-dot and improve celer-geo

sethrj:
This adds the small ability to use `null` values as input defaults for geometry/namespace (see https://github.com/celeritas-project/celerpy/pull/3) and also adds a script to export a Geant4 logical volume hierarchy from a GDML to a GraphViz representation (useful for understanding the depth and complexity of a realistic problem such as hgcal). We might want to move that one to celer-viz since it's pure python.


amandalund (approved):
Nice! Looks good, just some build errors from the missing `#include "celeritas_test.hh"`.

sethrj:
Oops that's a rebase error

PR #1253, opened 2024-05-27
Add automatic external fetching for JSON and Googletest

sethrj:
This adds support for automatically adding json and googletest as private dependencies when they're not externally installed. This is useful for building with CI or rapidly setting up for personal development. It is also meant to serve as an example for installing Perfetto in #1238.


cmake/CeleritasUtils.cmake
```diff
@@ -268,6 +283,36 @@ endfunction()
...
+      mark_as_advanced(${_var})
+      unset(_found)
+    endif()
+  endif()
+  if(_use_external)
+    set(CELERITAS_EXTERNAL TRUE)
+  endif()
+  unset(_use_external)
+  unset(_var)
+endmacro()
```
esseivaju: Unless I missed some details, this could be replaced by:
```cmake
FetchContent_Declare(
package
FIND_PACKAGE_ARGS
)
```
which also tries ```find_package``` before downloading the dependency.

external/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+
+#-----------------------------------------------------------------------------#
+# nlohmann_json
+#-----------------------------------------------------------------------------#
+
+if(CELERITAS_EXTERNAL_nlohmann_json)
+  FetchContent_Declare(
+    nlohmann_json
+    URL https://github.com/nlohmann/json/releases/download/v3.11.3/json.tar.xz
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
```
esseivaju: should we update ```cmake_minimum_required``` since this was introduced in version 3.24 (and ```FetchContent_MakeAvailable``` below in version 3.14) ?

cmake/CeleritasUtils.cmake
sethrj: Hmm good point, I didn't see that when I first started looking at `FetchContent`. I suppose the difference is that if we do it that way, the content fetching will have to go in the top level of our cmakelists (since `find_package` declares "local" variables). It's nice to be able to scope the external packages inside of a subdirectory so you can build `external/all` and  define local variables to override those projects.

It also looks like this is new to 3.24 which is *very* new: I'd like to keep 3 years of compatibility for dependencies if possible, which means ~3.20 is the newest we can do.

external/CMakeLists.txt
sethrj: I think arguments like `DOWNLOAD_EXTRACT_TIMESTAMP` and `EXCLUDE_FROM_ALL` are ignored if the cmake version doesn't support them... should we set a policy instead of using extract timestamp, and I guess I can reimplement https://github.com/celeritas-project/g4vg/pull/6 ?

esseivaju: I happened to have a system with an old (3.23.2) version of CMake so I tried passing ```DOWNLOAD_EXTRACT_TIMESTAMP``` and I get a configure error which goes away without that argument: 
```log
CMake Error at /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/ExternalProject.cmake:2670 (message):
  At least one entry of URL is a path (invalid in a list)
Call Stack (most recent call first):
  /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/ExternalProject.cmake:3640 (_ep_add_download_command)
  CMakeLists.txt:15 (ExternalProject_Add)


-- Configuring incomplete, errors occurred!
See also "/bld4/home/esseivaj/devel/celeritas/build/_deps/perfetto-subbuild/CMakeFiles/CMakeOutput.log".

CMake Error at /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/FetchContent.cmake:1076 (message):
  CMake step for perfetto failed: 1
Call Stack (most recent call first):
  /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/FetchContent.cmake:1217:EVAL:2 (__FetchContent_directPopulate)
  /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/FetchContent.cmake:1217 (cmake_language)
  /bld4/opt/cmake/3.23.2/share/cmake-3.23/Modules/FetchContent.cmake:1260 (FetchContent_Populate)
  CMakeLists.txt:353 (FetchContent_MakeAvailable)
```

So we probably want to set the policy as you suggested:
```cmake
if(POLICY CMP0135)
  cmake_policy(SET CMP0135 OLD)
endif()
```

cmake/CeleritasUtils.cmake
esseivaju: make sense, if we want to support 3.20 let's keep the custom find_or_fetch

external/CMakeLists.txt
sethrj: Thanks!

sethrj: Actually I'm going to take the opportunity to update the policy version range in the top-level header and require CMake 3.18.

esseivaju: we still need to set the policy to the OLD behavior if we want ```DOWNLOAD_EXTRACT_TIMESTAMP TRUE``` for all versions of CMake

sethrj: I guess I don't really care if the default behavior changes, I was mostly trying to silence the warning

sethrj:
@esseivaju I think this is finally ready!

esseivaju (approved):
Looks good!

esseivaju:
ATLAS likes to fetch sources from their own server https://atlas-software-dist-eos.web.cern.ch/externals/ so eventually we might want to allow the user to override the FetchContent url

pcanal:
ATLAS should not rely on the version bundled in Celeritas but should install them as externals.

PR #1254, opened 2024-05-27
Require JSON

sethrj:
Follow-on to #1253 : we're using JSON increasingly and will require it for the python front end (https://github.com/celeritas-project/celerpy/) and now have the ability to automatically provide it if it's not on the system. Requiring JSON simplifies a lot of the code, and other packages in the typical HEP pipeline (viz., ROOT) require nlohmann-json.

The only downside I see is the increased chance of causing a conflict in a multi-library toolchain (in which case nljson is *hopefully* already found, since e.g. ROOT is loaded before the initial decision to use JSON as an external?) or for G4VG where JSON isn't ever used in practice but this would require it.


drbenmorgan:
I think requiring `nlohmann-json` is fine for the reasons you've outlined. Noted that this is still in Draft and has some CI failures, so will approve once these are fixed.

drbenmorgan (approved):
Ci good, so LGTM @sethrj!

PR #1255, opened 2024-05-28
Fix windows test builds

sethrj:
This activates the build for googletest on windows so that we can guarantee compatibility on windows systems.

Differences from looking at test failures:
- hashing `std::variant` does not account for the "type" of the object, so two integer variants will compare the same even if they're different types
- the urban energy loss fluctuation model produces slightly different (but statistically equivalent) results


.github/workflows/build-fast.yml
```diff
@@ -90,5 +90,56 @@ jobs:
...
+            -DCeleritas_GIT_DESCRIBE="${{github.event.pull_request
+              && format(';-pr.{0};', github.event.pull_request.number)
+              || format(';-{0};', github.ref_name)}}"
+      - name: Build all
+        run: |
+          cmake --build --preset=$Env:CMAKE_PRESET 
+      - name: Test all
+        continue-on-error: true
+        run: |
+          ctest --preset=$Env:CMAKE_PRESET
```
pcanal: This seems to expand properly:
```
ctest --preset=$Env:CMAKE_PRESET 
```
and the test are not actually run (likely because of the missing config flag).

pcanal:
```suggestion
          ctest --preset=$CMAKE_PRESET 
```

.github/workflows/build-fast.yml
```diff
@@ -90,5 +90,56 @@ jobs:
...
+        shell: pwsh
+        run: |
+          Copy-Item scripts/cmake-presets/ci-windows-github.json -Destination CMakeUserPresets.json
+          cmake --preset=$Env:CMAKE_PRESET `
+            -DCeleritas_GIT_DESCRIBE="${{github.event.pull_request
+              && format(';-pr.{0};', github.event.pull_request.number)
+              || format(';-{0};', github.ref_name)}}"
+      - name: Build all
+        run: |
+          cmake --build --preset=$Env:CMAKE_PRESET
```
pcanal, line 127:
```suggestion
          cmake --build --preset=$CMAKE_PRESET 
```

.github/workflows/build-fast.yml
pcanal: but then again it seems to work in the config step ...

sethrj: Maybe it needs to be `cmake --test --preset`...

sethrj: The tests are run, and I think it's expanding correctly; this is the windows way of doing it. The failing tests didn't kill the job because I had `continue-on-error` set to true.

pcanal: In the run I looked at, they had not.  I see that indeed are running now.

test/celeritas/em/Fluctuation.test.cc
```diff
@@ -271,16 +271,17 @@ TEST_F(EnergyLossDistributionTest, urban)
...
-    {
-        static real_type const expected_counts[]
-            = {0,   0,   12, 223, 1174, 2359, 2661, 1867, 898, 369,
-               189, 107, 60, 48,  20,   9,    2,    2,    0,   0};
-        EXPECT_VEC_SOFT_EQ(expected_counts, counts);
-        EXPECT_SOFT_EQ(0.0099918954960280353, sum / num_samples);
-        EXPECT_EQ(551188, rng.count());
-    }
+#ifdef _MSC_VER
+    // XXX
```
pcanal: Is this missing an update/commit?

test/orange/orangeinp/CsgTree.test.cc
```diff
@@ -22,6 +22,12 @@ namespace test
 //---------------------------------------------------------------------------//
 TEST(CsgTypes, hash)
 {
+#ifdef _MSC_VER
+    // XXX
```
pcanal: Missing text?

test/orange/orangeinp/CsgTree.test.cc
```diff
@@ -22,6 +22,12 @@ namespace test
...
+    GTEST_SKIP()
```
pcanal, line 27: Do we rely on the `variant_hash` for the library code to work properly?

test/celeritas/em/Fluctuation.test.cc
sethrj: Ah yes I should actually write why this is disabled, thanks!

test/orange/orangeinp/CsgTree.test.cc
sethrj: Nope. It just means we've got a worse hash function: we're going to have more collisions and the CSG tree construction might be a bit slower as a result.

pcanal: Then what is the point of the test?  If the hash is still functional on Windows and 'just' bad, isn't the test as useful as on Linux but can be made to work with different types/values? (which then can be used on both platform.)

test/celeritas/em/Fluctuation.test.cc
sethrj: Oh right, I wrote it in the "skip" message. But I'll change this to a "TODO" and add a comment.

test/orange/orangeinp/CsgTree.test.cc
sethrj: It was more of a "let's try to understand what the compiler/library are doing" test: if it's embedding the type in the hash, it's not necessary to do it for our hash function. If it's not, then it might be worth trying both ways.

pcanal: fair enough.

pcanal (approved):

PR #1256, opened 2024-05-30
Fix CSG tree simplification needed by tilecal

sethrj:
The ATLAS tile calorimeter raytrace is having some problems. The first one I identified by simplifying a detector until it was a couple of cylinder segments. Previously this failed to raytrace while crossing the inner cylinder's surface (from left to right):
image

---

This is the original CSG tree graph, where "16" is "outside `TileTBEnv`" and set to false:
[tilecal-bug-0.pdf](https://github.com/celeritas-project/celeritas/files/15505388/tilecal-bug-0.pdf)
Originally this simplified to:
[tilecal-bug-2.pdf](https://github.com/celeritas-project/celeritas/files/15505407/tilecal-bug-2.pdf)
which did *not* eliminate the inner cylinder surface 3 (which can be proven 'true' by being inside tiletbenv) due to early replacement of a node. Now it's fixed:
[tilecal-bug-1-fixed.pdf](https://github.com/celeritas-project/celeritas/files/15505437/tilecal-bug-1-fixed.pdf)

---

Unfortunately although the raytrace for tilecal has fewer errors, there's still a problem:

image


src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -34,27 +35,83 @@ namespace orangeinp
  * - surface: "true"
  * - constant: check for contradiction
  *
+ * This operation is at worst O((number of nodes) * (depth of graph)).
+ *
  * \return Node ID of the lowest node that required simplification.
  */
-NodeId replace_down(CsgTree* tree, NodeId n, Node repl)
+void replace_and_simplify(CsgTree* tree, NodeId repl_node, Node repl)
```
elliottbiondo: `NodeId repl_node, Node repl` is kind of counterintuitive. How about `NodeId repl_id, Node repl_node` or something?

sethrj:
@elliottbiondo Please don't forget this one either!

elliottbiondo (approved):

elliottbiondo:
@sethrj sorry, I actually reviewed this Tuesday night, but I accidentally left my one minor comment as "pending".

PR #1257, opened 2024-06-03
Fix Geant4-to-ORANGE transformation conversion

sethrj:
Because of some weirdness in how Geant4 stores affine transforms (the rotation matrix is *inverted*; the transform is daughter-to-parent ["object" reference] but the rotation is parent-to-daughter ["frame" reference]) the `Transformer` object was incorrectly inverting the rotation matrix for placed volumes. This fixes the frame of reference so that "object" accessors are used when possible, and the *affine* rotation matrix is transposed on conversion.

---

ATLAS tile calorimeter before:
left-before
After:
left-after

There's still some issues, one of which is a boundary crossing problem fixed by #1256:
right-after


elliottbiondo (approved):

src/orange/g4org/Transformer.hh
```diff
@@ -105,17 +105,20 @@ auto Transformer::operator()(G4ThreeVector const& trans,
...
 {
-    return Transformation{transposed_from_geant(rot), scale_.to<Real3>(trans)};
+    return Transformation{convert_from_geant(rot), scale_.to<Real3>(trans)};
 }
 
 //---------------------------------------------------------------------------//
 /*!
  * Create a transform from an affine transform.
+ *
+ * The affine transform's stored rotation matrix is \em inverted!
```
mrguilima, line 115: How did you make sure that this inversion w.r.t. Geant4 is needed?
Geant4 and VecGeom use same convention, but I could not convince myself which one is (theirs or ours) is correct.  Or whether this relative inversion is okay for orange.

mrguilima (approved):
Tried to do more checks with the inverse transform, but ran out of time... :-(

src/orange/g4org/Transformer.hh
sethrj: There are so many inversions and call indirections that it's hard to exactly tell what is happening, but I think it comes down to `G4AffineTransform::TransformPoint`:
```c++
  G4double vecx = vec.x(), vecy = vec.y(), vecz = vec.z();
  return { vecx*rxx + vecy*ryx + vecz*rzx + tx,
           vecx*rxy + vecy*ryy + vecz*rzy + ty,
           vecx*rxz + vecy*ryz + vecz*rzz + tz };
```
this is the operation $\mathbf{R}^T\mathbf{x} + \mathbf{t}$.

In contrast, `HepRotation::operator *` (aka G4RotationMatrix) is:
```c++
  double x = p.x();
  double y = p.y();
  double z = p.z();
  return Hep3Vector(rxx*x + rxy*y + rxz*z,
                    ryx*x + ryy*y + ryz*z,
                    rzx*x + rzy*y + rzz*z);
```
which is $\mathbf{R}\mathbf{x}$.

Celeritas uses  $\mathbf{R}\mathbf{x} + \mathbf{t}$ for daughter-to-parent and $\mathbf{R}^T(\mathbf{x} - \mathbf{t})$ for parent-to-daughter.

Basically I was trying to use the "non-deprecated" methods in Geant4 that describe whether the reference frame is "object" (child-to-parent) or "frame" (parent-to-child) and avoid doing extra operations like `GetInverse`. Ultimately though the true consistency test is to ensure Geant4, VecGeom, and Celeritas have the same behavior when tracking through displaced solids and transformed daughters; which we have now done.

PR #1258, opened 2024-06-03
Import and add nuclear binding energy to isotopes

whokion:
# Import the nuclear binding energy of isotopes from Geant4
The nuclear binding energy is needed for hadronic models and is added to MaterialData and IsotopeView
based on Geant4 [G4NucleiProperties](https://geant4.kek.jp/lxr/source/particles/management/src/G4NucleiProperties.cc#L229)
- Add the nuclear binding energy in the Isotope definition and import it from Geant4 
- Update IsotopeView and tests associated with the addition

### DIFF
--- a/app/celer-dump-data.cc
+++ b/app/celer-dump-data.cc
@@ -123,7 +123,7 @@ void print_isotopes(std::vector<ImportIsotope>& isotopes)
     cout << R"gfm(
 # Isotopes
 
-| Isotope ID | Name   | Atomic number | Atomic mass number | Nuclear mass (MeV) |
+| Isotope ID | Name   | Atomic number | Atomic mass number | Binding energy (MeV) | Nuclear mass (MeV) |
 | ---------- | ------ | ------------- | ------------------ | ------------------ |
 )gfm";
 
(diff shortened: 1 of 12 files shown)


app/celer-dump-data.cc
```diff
@@ -123,7 +123,7 @@ void print_isotopes(std::vector<ImportIsotope>& isotopes)
     cout << R"gfm(
 # Isotopes
 
-| Isotope ID | Name   | Atomic number | Atomic mass number | Nuclear mass (MeV) |
+| Isotope ID | Name   | Atomic number | Atomic mass number | Binding energy (MeV) | Nuclear mass (MeV) |
 | ---------- | ------ | ------------- | ------------------ | ------------------ |
```
sethrj: Line below needs  another column as well

sethrj (approved):
One minor detail, otherwise looks good to me!

src/celeritas/mat/MaterialParams.cc
```diff
@@ -24,6 +24,7 @@
 #include "celeritas/io/ImportData.hh"
 
 #include "MaterialData.hh"  // IWYU pragma: associated
+
```
stognini, line 27: Looks like one of these days might be worth running a clang-format everywhere...

stognini (approved):

stognini:
Thanks @whokion! Looks great.

app/celer-dump-data.cc
whokion: Fixed.

src/celeritas/mat/MaterialParams.cc
whokion: @sethrj, @stognini Thank you for quick reviews and approvals!

PR #1261, opened 2024-06-05
Refactor CMake test name generation to be based on path

sethrj:
Very late follow-on to #1105 (and #1099 before that). This removes the duplicate path names in generated tests, and it ensures that the CTest name matches exactly the test path.


pcanal (approved):
It seems to work as advertised (on a local run).

amandalund (approved):
Thanks for fixing these! :)

PR #1262, opened 2024-06-05
Improve GenTrap regularization and construction

sethrj:
This does a number of small enhancements in GenTrap while trying to debug the failing EMEC alignment:
- Adds consistent face names to the GenTrap faces
- Changes the Trap/Trd construction points so that the sides are ordered like a prism (right-pointing face first, then counterclockwise)
- Correctly implements regularization for quadric surfaces that have one negative, one positive, and one zero term
- Uses the same logic for the Plane which improves construction and eliminates a potential soft equivalence/simplify inconsistency
- Fixes an edge case in the polygon `is_convex` function (false negative when the first two points are collinear with the last two)
- Enforces that GenTrap twists are less than 90 degrees
- Adds additional testing for twisted surfaces showing that there's a failure when the top and bottom edges of a face are different lengths


sethrj:
@elliottbiondo This is mostly changes to @mrguilima 's work but he's on vacation for a couple more weeks so I was wondering if you could take a look. If needed I can break up a couple of the changes into separate PRs... but it was more convenient to change all the surface representations simultaneously.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -203,7 +203,11 @@ class Ellipsoid final : public IntersectRegionInterface
  *
  * A GenTrap, like VecGeom's equivalent and ROOT's Arb8, represents a general
  * trapezoidal volume with up to eight vertices, or two 4-point sets, sitting
- * on two parallel planes perpendicular to Z axis.
+ * on two parallel planes perpendicular to the Z axis.
+ *
+ * Trapezoid constructed from the helper functions will have sides that are
```
elliottbiondo: Trapezoids

src/orange/surf/FaceNamer.hh
```diff
@@ -110,5 +114,19 @@ std::string FaceNamer::operator()(Sense s, S const& surf)
     return result;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Apply with an explicit face name.
+ *
+ * This can be useful for instances such as "generic trapezoid" where the
+ * surface type can change based on the
```
elliottbiondo: ...

elliottbiondo (approved):

src/orange/surf/FaceNamer.hh
sethrj: 🐿️

Issue #1260, opened 2024-06-04
Support unions as universe boundaries


sethrj:
We've encountered an issue where the exterior boundary of a unit is a *union* so we cannot eliminate the surfaces from the boundary, resulting in a coincident (or nearly coincident) boundary in multiple levels. This results in a tracking error when crossing from the parent to the daughter because we end up in the "exterior" of the daughter universe. (It can also cause assertion errors when the daughter universe is complex because we start by validating that the point is "inside".) This shows up in the ATLAS tile calorimeter which has a union that contains another physical volume.

The proposed short-term solution is to keep logically eliminating boundary surfaces on the daughters when we can, but we will also support coincident surfaces (rather, the same surface appearing in multiple universe, potentially with transformations and negations applied).
- Change proto construction from top-down to bottom-up so that we build the child CSG trees before the potential enclosing units and therefore know whether there are any boundary surfaces that cannot be eliminated.
- Retain an *ordered* list of the boundary's surfaces so that a congruent shape but with a transformation applied will have the correct surface mapping. Because of surface regularization (flipping normal directions and senses) we will have to retain knowledge of whether the surface sense needs to be flipped as well when going from parent to child.
- Add data structure in input and runtime for storing this mapping on volumes. We should be able to just have a single new `Id` index for each volume: null if no mapping is needed, offset into an array of *daughter* local surfaces and bools for sense flipping. (We won't need to store as a map or store the length because each volume already has a "face" map that will be the same length if the mapping is set.) Faces that can be deleted from the daughters can be left as null IDs.
- Probably want to add a new accessor to the `VolumeView` for these mappings?
- Inside `find_next_step` and `cross_boundary`, when descending from the "current surface level", we have to interrogate the current volume's map to find whether the next lower level has a corresponding surface, then *updating* the on-surface (ID and sense) that level in the `LocalState` object for the daughter. (Both of these calls depend on the local surface state.)

Long term solution for the Geant4 representation might be to have only a stack of universe IDs, a local tracker, and a detailed mapping of parent/child universes: because for this geometry specification we do not need to worry about parents truncating the children via overlaps. This is something I'd thought about for being more efficient but the tricky case of not being able to eliminate daughter surfaces may make it a requirement.

Parent universe where the daughter is a union of two spheres:
[...]

PR #1264, opened 2024-06-07
Work around union/polysolid boundaries and fix transformation of explicit daughters

sethrj:
This adds a minimal reproducer demonstrating #1260 (which is still disabled), adds a minimal GDML failure (union volume with a daughter), and then:
- Forces all LVs with a union/polysolid boundary to be explicitly inlined into the parent volume
- Fixes an incorrect doubly applied transformation to "explicit" background cells (i.e. parents with children subtracted).

The force-inlining may be very inefficient but at least allows us a working solution while we implement a better longer-term alternative. It also creates universes with *many* more surfaces than before, which means that the surface hash acceleration from #1183 needs to be merged for performance reasons for nontrivial problems.

Tilecal top-level universe:
tiletb csg

Raytrace at y=0:
tilecal-xz

Raytrace at x=1.8m still shows some issues with the GenTrap's positioning somehow:
emec-yz-x1800mm


src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -90,20 +91,26 @@ void replace_and_simplify(CsgTree* tree, NodeId repl_key, Node repl_value)
...
     // Replace nonliterals
     for (auto n : range(CsgTree::false_node_id() + 1, NodeId{tree->size()}))
     {
+        auto repl_state = state[n.get()];
         if (std::holds_alternative<Surface>((*tree)[n]))
         {
-            continue;
+            if (repl_state == NodeReplacer::unknown)
+            {
+                // Track boundary surfaces that we can't prove, likely because
```
elliottbiondo: Keep track of?

src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -39,7 +39,8 @@ namespace orangeinp
  *
  * \return Node ID of the lowest node that required simplification.
```
elliottbiondo: This comment needs to be updated

test/orange/OrangeGeant.test.cc
```diff
@@ -198,6 +198,39 @@ TEST_F(SimpleCmsGeantTest, trace)
...
 
+//---------------------------------------------------------------------------//
+class TilecalPlugTest : public GeantOrangeTest
+{
+    std::string geometry_basename() const final { return "tilecal-plug"; }
+};
+
+TEST_F(TilecalPlugTest, trace)
+{
+    if (0)
```
elliottbiondo: Maybe add a TODO here to explain why this block is `if`-ed out, on the off chance that it is forgotten about in the future.

elliottbiondo (approved):

test/orange/OrangeGeant.test.cc
sethrj: Actually the test passes now. Good catch!

PR #1267, opened 2024-06-12
Add initial set of clang-tidy checks

esseivaju:
Add a subset of clang-tidy checks. This is not integrated with the CI (yet) as the set of enabled checks triggers many warnings but this can still be useful to run locally on individual files during development / PR review.


sethrj (approved):
The options look good to me, but I'd be happier if there were at least a small test (even a minimal no-geant4 version with `continue-on-error: true`) as part of the CI. Is that something you could add?

esseivaju:
Sure, I'll add that

esseivaju:
It should be ready now. Maybe at some point, if we want to require no warning, we can go through the existing warnings and either silence them with a `// NOLINT` comment or remove some of the checks that are too strict. For now, we can see the warnings in the build step logs.

sethrj:
@esseivaju Very cool. It's a bit slow but probably worth it: I see what could be a couple of legitimate bugs or unintentional bad practices.
```
/home/runner/work/celeritas/celeritas/src/celeritas/optical/OpticalCollector.cc:42:9: warning: 'host_data' used after it was moved [bugprone-use-after-move]
    if (host_data.cerenkov)
        ^
/home/runner/work/celeritas/celeritas/src/celeritas/optical/OpticalCollector.cc:34:21: note: move occurred here
    storage_->obj = {std::move(host_data), inp.num_streams};
                    ^
```

esseivaju:
yes, it is a bit slow. A solution would be to list the files changed in the PR and only check these.

esseivaju:
> @esseivaju Very cool. It's a bit slow but probably worth it: I see what could be a couple of legitimate bugs or unintentional bad practices.
> 
> ```
> /home/runner/work/celeritas/celeritas/src/celeritas/optical/OpticalCollector.cc:42:9: warning: 'host_data' used after it was moved [bugprone-use-after-move]
>     if (host_data.cerenkov)
>         ^
> /home/runner/work/celeritas/celeritas/src/celeritas/optical/OpticalCollector.cc:34:21: note: move occurred here
>     storage_->obj = {std::move(host_data), inp.num_streams};
>                     ^
> ```

yeah, a few warnings in there should probably be fixed.

PR #1269, opened 2024-06-12
Add nuclear zone data

whokion:
# Extend neutron inelastic data for the Bertini cascade nuclei model

This MR adds data for nuclear zone properties to NeutronInelasticData and associated updates:
 - Import nucleon loss energies in IsotopeView
 - Add options for the Bertini cascade model
 - Add nuclear zone data for the Bertini cascade model
 - Extend the NeutronInelastic test and adjust other tests due to updated in Isotope


whokion:
@sethrj Is the fail from windows real?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Integrate a nuclear potential numerically by adaptive quadrature.
+ */
+real_type NuclearZoneBuilder::integrate_potential(real_type rmin,
+                                                  real_type delta_r,
+                                                  real_type integral,
+                                                  real_type ws_shift) const
+{
+    constexpr int max_trials = 1e+3;
```
sethrj: @whokion This is the source of the windows error:
```
D:\a\celeritas\celeritas\src\celeritas\neutron\model\detail/NuclearZoneBuilder.hh(368): error C2220: the following warning is treated as an error
2024-06-12T13:55:56.9194698Z D:\a\celeritas\celeritas\src\celeritas\neutron\model\detail/NuclearZoneBuilder.hh(368): warning C4244: 'initializing': conversion from 'double' to 'const int', possible loss of data
```

Write it as `1000`. (In theory you could specify a real number that's out of range of `int` and then loses accuracy.)

whokion: Thanks!

src/celeritas/ext/GeantImporter.cc
```diff
@@ -425,6 +425,16 @@ std::vector<ImportIsotope> import_isotopes()
...
+        isotope.proton_loss_energy
+            = G4NucleiProperties::GetBindingEnergy(isotope.atomic_mass_number,
+                                                   isotope.atomic_number)
+              - G4NucleiProperties::GetBindingEnergy(
+                  isotope.atomic_mass_number - 1, isotope.atomic_number - 1);
+        isotope.neutron_loss_energy
+            = G4NucleiProperties::GetBindingEnergy(isotope.atomic_mass_number,
+                                                   isotope.atomic_number)
+              - G4NucleiProperties::GetBindingEnergy(
+                  isotope.atomic_mass_number - 1, isotope.atomic_number);
```
sethrj: This is causing the test failure:
```
error: expected a JSON object but got the following stdout:
G4NucleiProperties::GetMassExccess: Wrong values for A = 0 and Z = 0
G4NucleiProperties::GetMassExccess: Wrong values for A = 0 and Z = 1
```

We probably should add 
```c++
    ScopedGeantLogger scoped_logger;
```
to  `GeantImporter::operator()`.

I guess you'll have to guard these calls with `A > 1` and `Z > 1`?

whokion: Okay, thanks!  This still looks a bit strange as it should return 0 and exit from [G4NucleiProperties::GetBindingEnergy](https://geant4.kek.jp/lxr/source/particles/management/src/G4NucleiProperties.cc#L238) for those cases and should not call `G4NucleiProperties::GetMassExccess`.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -425,6 +425,29 @@ std::vector<ImportIsotope> import_isotopes()
         isotope.atomic_mass_number = g4isotope.GetN();
         isotope.binding_energy = G4NucleiProperties::GetBindingEnergy(
             isotope.atomic_mass_number, isotope.atomic_number);
+
+        // Binding energy difference for lossing a nucleon
```
sethrj:
```suggestion
        // Binding energy difference for losing a nucleon
```

src/celeritas/io/ImportElement.hh
```diff
@@ -24,7 +24,9 @@ struct ImportIsotope
     std::string name;  //!< Isotope label
     int atomic_number;  //!< Atomic number Z
     int atomic_mass_number;  //!< Atomic number A
-    double binding_energy;  //!< Nuclear binding energy [MeV]
+    double binding_energy;  //!< Nuclear binding energy (BE) [MeV]
+    double proton_loss_energy;  //!< BE(A, Z) - BE(A-1, Z-1) [MeV]
+    double neutron_loss_energy;  //!< BE(A, Z) - BE(A-1, Z) [MeV]
```
sethrj: All of the data in this struct would benefit from default initializers (missed that in earlier reviews)

src/celeritas/neutron/model/CascadeOptionsIO.json.cc
```diff
@@ -0,0 +1,81 @@
...
+ * Read options from JSON.
+ */
+void from_json(nlohmann::json const& j, CascadeOptions& opts)
+{
+#define FDO_INPUT(NAME)                    \
+    do                                     \
+    {                                      \
+        if (j.contains(#NAME))             \
+            j.at(#NAME).get_to(opts.NAME); \
+    } while (0)
```
sethrj: There's a little utility function that makes this easier (and slightly more efficient and slightly more flexible):

```suggestion
#define FDO_INPUT(NAME)  CELER_JSON_LOAD_OPTION(j, opts, NAME)
```
and include "corecel/io/JsonUtils.json.hh"

src/celeritas/neutron/model/NeutronInelasticModel.cc
```diff
@@ -76,6 +84,16 @@ NeutronInelasticModel::NeutronInelasticModel(ActionId id,
...
     CELER_ASSERT(data.xs_params.size() == data.nucleon_xs.size());
 
+    // Build (A, Z)-dependent nuclear zone data
+    detail::NuclearZoneBuilder zone_builder(
+        options, data.scalars, &data.nuclear_zones);
+
+    for (auto iso_id : range(IsotopeId{materials.num_isotopes()}))
+    {
+        zone_builder(materials.get(iso_id));
+    }
```
sethrj: Rename the instance to the function-like `build_zone` :)

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+    using AtomicMassNumber = AtomicNumber;
+    using MevMass = units::MevMass;
+    using Energy = units::MevEnergy;
+    using ComponentVec = std::vector<ZoneComponent>;
+    using Data = HostVal<NuclearZoneData>;
+    //!@}
+
+  public:
+    // Construct with cascade options and shared data
+    explicit inline NuclearZoneBuilder(CascadeOptions const& options,
```
sethrj: Doesn't need to be explicit (since there's more than one construction argument)

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+    zones_.push_back(nucl_zones);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return the number of nuclear zones.
+ */
+size_type NuclearZoneBuilder::num_nuclear_zones(AtomicMassNumber a)
+{
+    return (a.get() < 5) ? 1 : (a.get() < 100) ? 3 : 6;
```
sethrj: Might be a bit clearer as a couple of chained `if` statements, what do you think?
```suggestion
    if (a < AtomicMassNumber{5})
        return 1;
    if (a < AtomicMassNumber{100})
        return 3;
    return 6;
```

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+            real_type y = std::log((1 + skin_decay) / alpha[i] - 1);
+            components[i].radius = nuclear_radius + skin_depth_ * y;
+            ur[i + 1] = y;
+            integral[i] = this->integrate_woods_saxon(
+                ur[i], ur[i + 1], nuclear_radius);
+            total_integral += integral[i];
+        }
+    }
+
+    ur.clear();
```
sethrj: This and the `integral` clear are unnecessary as they'll be destroyed at the end of this function

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+    real_type nuclear_radius = this->calc_nuclear_radius(a);
+
+    real_type skin_ratio = nuclear_radius / skin_depth_;
+    real_type skin_decay = std::exp(-skin_ratio);
+
+    // Temporary data for the zone-by-zone density function
+    size_type num_of_zones = components.size();
+    std::vector<real_type> ur(num_of_zones + 1);
+    std::vector<real_type> integral(num_of_zones);
+    real_type total_integral{0};
```
sethrj: It would be a little less error prone to calculate the total after the big `if` statements, rather than accumulating in four different places:
```c++
real_type total_integral = std::accumulate(integral.begin(), integral.end(), real_type{0});
```

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+/*!
+ * Integrate a nuclear potential numerically by adaptive quadrature.
+ */
+real_type NuclearZoneBuilder::integrate_potential(real_type rmin,
+                                                  real_type delta_r,
+                                                  real_type integral,
+                                                  real_type ws_shift) const
+{
+    constexpr int max_trials = 1000;
+    constexpr real_type epsilon = 1e-3;
```
sethrj, line 362: This function is deep enough and complicated enough to merit a standalone class with testing, and maybe configurable parameters like the maximum trial depth. (Note also that a linear increase in the "max trials" results in a *quadratic* increase in the execution time due to the inner loop.)

If you make a function-like class `PotentialIntegrator` these could be construction arguments... and since this looks like it could be a generic recursive integrator, I think it would be even better as a standalone testable `NewtonCotesIntegrator` class in `corecel/math` templated on `class F` (which evaluates the potential). And I don't think the below is a very optimal implementation of it...

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+{
+    real_type delta_r = rmax - rmin;
+    real_type rmin_sq = ipow<2>(rmin);
+    real_type rmax_sq = ipow<2>(rmax);
+
+    real_type integral
+        = this->half() * delta_r
+          * (rmin_sq * std::exp(-rmin_sq) + rmax_sq * std::exp(-rmax_sq));
+
+    real_type result = integrate_potential(false, rmin, delta_r, integral);
```
sethrj: I think this is wrong: `false` is being promoted to `rmin = 0`, `rmin` is being passed as `delta_r`, etc.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+        for (int i = 0; i < depth; ++i)
+        {
+            r += interval;
+            fi += (ws_shift > 0) ? r * (r + ws_shift) / (1 + std::exp(r))
+                                 : ipow<2>(r) * std::exp(-ipow<2>(r));
+        }
+
+        result = this->half() * integral + fi * delta_r;
+
+        if (std::fabs(1 - integral / result) < epsilon)
```
sethrj: I think this would be better written with `SoftEqual`? (Use a function-local `constexpr SoftEqual<>{epsilon}`). The SoftEqual also has an absolute tolerance that might be useful (or might be harmful); what's the scale of the expected result?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+{
+    real_type skin_ratio = nuclear_radius / skin_depth_;
+
+    real_type ws_shift = 2 * skin_ratio;
+    real_type delta_r = rmax - rmin;
+
+    // Average value of \f$ r^{2} V(r) dr \f$ at boundaries
+    real_type integral = half() * delta_r
+                         * (rmin * (rmin + ws_shift) / (1 + std::exp(rmin))
+                            + rmax * (rmax + ws_shift) / (1 + std::exp(rmax)));
```
sethrj, line 321: This (and the gaussian case below) are just initial guesses for the integral... this should be trivial to incorporate into the adaptive quadrature solver itself, right?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+            {
+                real_type y = std::log((1 + skin_decay) / alpha[i] - 1);
+                components[i].radius = nuclear_radius + skin_depth_ * y;
+                ur[i + 1] = y;
+                integral[i] = this->integrate_woods_saxon(
+                    ur[i], ur[i + 1], nuclear_radius);
+                total_integral += integral[i];
+            }
+        }
+    }
```
sethrj: This would be easier to read with 
```c++
    else if (amass < 12)
    {
        constexpr Real3 sqrt_minus_log_alpha = {/* precompute these*/};
        /* * */
    }
    else if (amass < 100)
    {
        constexpr Real3 alpha = {0.7, 0.3, 0.01};
        /* * */
    }
```

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+
+    // Calculate nuclear radius
+    real_type nuclear_radius = this->calc_nuclear_radius(a);
+
+    real_type skin_ratio = nuclear_radius / skin_depth_;
+    real_type skin_decay = std::exp(-skin_ratio);
+
+    // Temporary data for the zone-by-zone density function
+    size_type num_of_zones = components.size();
+    std::vector<real_type> ur(num_of_zones + 1);
```
sethrj: I don't think you need a whole vector of `ur`: only `[i, i+1]` are used at a time...

sethrj (changes requested):
Everything but the NuclearZoneBuilder looks essentially perfect! great design and implementation.

I think the NZB needs to split out a templated numerical integrator class that has some additional testing on it.

src/celeritas/ext/GeantImporter.cc
whokion: Fixed.

src/celeritas/io/ImportElement.hh
whokion: Added  default initializers.

src/celeritas/neutron/model/CascadeOptionsIO.json.cc
whokion: Updated.

src/celeritas/neutron/model/NeutronInelasticModel.cc
whokion: Rename build_zone to build_nuclear_zones.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Removed explicit.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Okay, changed as suggested.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Removed.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Good suggestion!  Updated accordingly.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Great suggestion - actually I thought about how to pass a functor like object to `integrate_potential`, but did not think harder^^.  If you do not mind, I would like to do this as a separate MR or you may help me here as you already have a good idea.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Nice catch!  This is definitely a bug (left over from the earlier implementation).  I also realize that this Gaussian potential function is not tested (i.e, any A in [5, 11]), so the bug was not caught in validation (with respect to Geant4) in the test.  I will add a test case.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Updated as suggested - at least, good to see that there is no changes test results.

src/celeritas/neutron/data/NeutronInelasticData.hh
```diff
@@ -59,6 +63,73 @@ struct StepanovParameters
...
+    NucleonArray potential{0, 0};  //!< nucleon potentials [MeV]
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Data characterizing the nuclear zones.
+ */
+struct NuclearZones
+{
+    size_type num_zones{};
```
amandalund:
```suggestion
```
Since we can get the number from `zones.size()`, right?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: You are absolutely right. This was a lazy man's implementation, but definitely can be integrated into `PotentialIntegrator` that you suggested early.

src/celeritas/neutron/data/NeutronInelasticData.hh
```diff
@@ -59,6 +63,73 @@ struct StepanovParameters
...
+        return num_zones > 0 && !zones.empty();
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Device data for nuclear zone properties
+ */
+template<Ownership W, MemSpace M>
+struct NuclearZoneData
```
amandalund, line 98: I think the data could be simplified slightly by removing this struct and instead storing the `components` and `zones` directly in `NeutronInelasticData`.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Okay, I did not through about this pre-computation during refactoring Geant4 codes.  Good suggestion. Changed accordingly.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Nice catch! Removed this vector, but use a temporary variable (one double to keep the last value of y).

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+    }
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with cascade options and data.
+ */
+CELER_FUNCTION
```
amandalund:
```suggestion
```

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,410 @@
...
+    return (a.get() < 5) ? 1 : (a.get() < 100) ? 3 : 6;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate components of nuclear zone data: the nuclear zone radius, volume,
+ * density, Fermi momentum and potential function as in G4NucleiModel and as
+ * documented in section 24.2.3 of the Geant4 Physics Reference (release 11.2).
+ */
+CELER_FUNCTION
```
amandalund:
```suggestion
```

whokion:
@sethrj Thanks a lot for your review comments and suggestions, which I hope that I addressed most of them except the request for PotentialIntegrator (which may be a separate MR).  Is the CI failure something that I should worry about or a just glitch?

sethrj:
@whokion Just a glitch I think: rerunning now. I'll see if I can make a small PR with a generic numerical integrator and test...

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,408 @@
...
+
+    // Calculate nuclear radius
+    real_type nuclear_radius = this->calc_nuclear_radius(a);
+
+    real_type skin_ratio = nuclear_radius / skin_depth_;
+    real_type skin_decay = std::exp(-skin_ratio);
+
+    // Temporary data for the zone-by-zone density function
+    size_type num_of_zones = components.size();
+    std::vector<real_type> integral(num_of_zones);
```
sethrj, line 171: Instead of doing the "number of zones"  calculation in a separate function and expecting the size to match in the `if` statements below, I think it'd be better to resize the vector inside this function. Inside the `if` statements, resize the vector to 1, `y.size()`, `alpha.size()`, etc., and then loop over `integral.size()` rather than a separate `num_of_zones` values.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,408 @@
...
+{
+    AtomicNumber a = target.atomic_mass_number();
+    size_type num_zones = this->num_nuclear_zones(a);
+
+    std::vector<ZoneComponent> comp(num_zones);
+    this->calc_zone_component(target, comp);
+
+    NuclearZones nucl_zones;
+    nucl_zones.num_zones = num_zones;
+    nucl_zones.zones = components_.insert_back(comp.begin(), comp.end());
```
sethrj: `num_zones` is redundant with `zones.size()`, so maybe that data should be removed?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,408 @@
...
+        components[i].volume *= this->four_thirds_pi();
+    }
+
+    // Fill the nuclear zone density, fermi momentum, and potential
+    int const dim = ZoneComponent::NucleonArray::size();
+    int num_protons = target.atomic_number().get();
+    int num_nucleons[dim] = {num_protons, a.get() - num_protons};
+    real_type mass[dim] = {proton_mass_.value(), neutron_mass_.value()};
+    real_type dm[dim] = {target.proton_loss_energy().value(),
+                         target.neutron_loss_energy().value()};
```
sethrj:
```suggestion
    int num_protons = target.atomic_number().get();
    int const num_nucleons[] = {num_protons, a.get() - num_protons};
    real_type const mass[] = {proton_mass_.value(), neutron_mass_.value()};
    real_type const dm[] = {target.proton_loss_energy().value(),
                            target.neutron_loss_energy().value()};
    static_assert(std::size(mass) == ZoneComponent::NucleonArray::size());
    static_assert(std::size(dm) == std::size(mass));
```
This is a little safer

sethrj (commented):
One or two small suggestions; I'll let @amandalund review and then I can follow on after the merge with the numerical integrator.

amandalund (approved):
This looks great @whokion!

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -0,0 +1,408 @@
...
+        for (int i = 0; i < depth; ++i)
+        {
+            r += interval;
+            fi += (ws_shift > 0) ? r * (r + ws_shift) / (1 + std::exp(r))
+                                 : ipow<2>(r) * std::exp(-ipow<2>(r));
+        }
+
+        result = this->half() * integral + fi * delta_r;
+
+        if (soft_eq(1, integral / result))
```
sethrj, line 388: Does 
```suggestion
        if (soft_eq(result, integral))
```
not give the expected behavior?

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Not sure about this.  The number of zones is used for other properties (i.e., volumes, etc), so it may be still more readable keeping it as a common size across the whole code, which is called just once for a specific Isotope.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Okay, it is removed from the zone data (but keep the data structure as it was since more data may be added later).

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Okay. Changed as suggested.

src/celeritas/neutron/data/NeutronInelasticData.hh
whokion: Yes, this is now removed.

src/celeritas/neutron/data/NeutronInelasticData.hh
whokion: Actually, I originally did what you are suggesting, but I may want to reorganize data structure by grouping data separately needed for the process (i.e., macro_xs) and for the interactor (nucleon_xs, zone_data and other data which will be added later).  Although NuclearZoneData is too specific now and can be renamed by the more generic name as more data are added, I would keep the structure as it is now and  reorganize them later as more codes and data are added.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Removed.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Removed.

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
whokion: Okay, I also tried this earlier, which did not produce the same results compared to the current ones - as epsilon is the tolerance for the ratio, not on the absolute value.  Of course, we can re-adjust the value epsilon for the direct comparison with some optimization later.  It may be better to keep it as it is until we fully verify results with Geant4 - at this moment, these zone data are identical to those from Geant4 (which depends on the particular epsilon value), but certainly need more downstream codes for further validation and verification.

sethrj (approved):
Nice, thanks!

src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
sethrj: Sorry, i didn't see this wasn't addressed. Are you expecting `NuclearZoneBuilder::num_nuclear_zones` to be called outside of that helper class? The number of zones is implicitly available through the components size, so the separate assumptions between `num_nuclear_zones` and `calc_zone_component` is a dangerous redundancy.

Issue #1268, opened 2024-06-12
Add particle weight to simulations


sethrj:
The concept of "weight" is used for variance reduction to allow splitting and rouletting of particles. We ignore the weight parameter for G4Track, which defaults to zero and can therefore break some SD implementations.

Since Celeritas itself doesn't (yet) perform variance reduction, a first step we will require offloaded particles to have unit weight, and set the weight to unity for reconstructed tracks.

This should be trivially easy:
- Add a collection to `SimStateData`
- Add a weight field to `Primary`
- Propagate weights from the parent during creation of secondaries in `InteractionApplier`
- Multiply weights into the energy deposition
- Copy the weights back to the hit

**Not** in scope for this issue:
- Biasing manager (weight windows, russian roulette, splitting)
- Worrying about the unphysicality/bias of weights in this non-statistical context (a weight of 10 implies that ten particles happen to undergo exactly the same interactions at the same place, which might work for calorimeters but not in other spaces)

sethrj:
@esseivaju This will be an easy first thing to tackle before moving on to the G4Track stuff. It's one more feature we can leave on during the ATLAS simulation.

sethrj:
@esseivaju This could be an easy "first MR" for someone else who wants to dip into the Celeritas codebase (@rahmans1 ? @Rashika-Gupta ?) but I think we should add it sooner than later.

esseivaju:
Definitely is someone wants to work on this feel free to take it.

Rashika-Gupta:
I can give it a try!

esseivaju:
Is this closed by #1853?

sethrj:
Also needed fixes from #1940 and #1965 ... but now it's fully tested 🤪

PR #1270, opened 2024-06-12
Set SD track weight and warn if offload weight is not unity

sethrj:
This is a band-aid to issue #1268 that will at least fix callbacks to SDs in ATLAS simulations, which check the `G4Track`'s weight (which Geant4 defaults to zero). To ensure correctness, we log an "error" (i.e., the answer is wrong but we'll keep going) if the incident track doesn't have unit weight.


drbenmorgan (approved):
Thanks for the quick fix @sethrj! All looks good, so LGTM once CI passes.

PR #1271, opened 2024-06-12
Support degenerate faces for "GenTrap" and rename to GenPrism

sethrj:
This adds support for degenerate (coincident or collinear) points in a GenTrap allowing for tetrahedra, or pentahedra (the top face being a line). There's no reason to limit the number of points on the top or bottom face; they can each be arbitrary convex polygons. (We should for example be able to make a twisted hexagonal prism.) Thus I suggest renaming this class to `GenPrism`. It's more like an extruded and warped polygon... but I think that encapsulates the typical use cases.


sethrj:
Ping @mrguilima

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -744,23 +745,58 @@ TEST_F(GenTrapTest, triang_prism)
...
+    static real_type const expected_twist_angles[] = {0, 0, 0, 0};
+    EXPECT_VEC_SOFT_EQ(expected_twist_angles, this->get_twist_angles(trap));
+
     auto result = this->test(trap);
-    result.print_expected();
-    this->check_corners(result.node_id, trap, 0.1);
+    static char const expected_node[] = "all(+0, -1, +2, -3, -4, +5)";
+    static char const* const expected_surfaces[]
+        = {"Plane: z=-2",
+           "Plane: z=2",
```
mrguilima: Since the top polygon is degenerated into a line, can't this Plane at z=2 be removed?

mrguilima (approved):
Nice addition to support degenerate faces!
I was originally wondering about the "Prism" part of the name, but the shape description is quite appropriate.

test/orange/orangeinp/IntersectRegion.test.cc
sethrj: Ooh, good point! I'll see what I can do.

PR #1274, opened 2024-06-14
Add Wentzel macro xs calculator and fix `a_sq_factor`

amandalund:
- Fix an error in the calculation of `a_sq_factor` for Wentzel OK&VI
- Fix calculation of `inv_mass_cbrt_sq`
- Fix `a_sq_factor` units
- Add a Wentzel total xs calculator (to be used in the Wentzel VI msc model)
- Move some of the common code for the Urban and Wentzel VI tests into a msc test base.


src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -36,7 +36,7 @@ struct CoulombParameters
     //! Factor used to calculate the maximum scattering angle off of a nucleus
     real_type a_sq_factor{real_type(0.5)
                           * ipow<2>(constants::hbar_planck * constants::c_light
-                                    * units::femtometer)};
+                                    / units::femtometer)};
```
sethrj: Maybe a `Quantity` (FmSq?) would help elucidate this factor? The divide vs multiply has bitten me several times...

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -77,7 +77,7 @@ WentzelOKVIParams::WentzelOKVIParams(SPConstMaterials materials,
     host_data.params.a_sq_factor
         = real_type(0.5)
           * ipow<2>(options.angle_limit_factor * constants::hbar_planck
-                    * constants::c_light * units::femtometer);
+                    * constants::c_light / units::femtometer);
```
sethrj: Since the factor's always being set here (it looks like??) maybe just default the `CoulombParameters` to zero and reduce the duplication here.

sethrj (approved):
LGTM, but I'll let @whokion take a look before merging. I'm going to remove the `bug` label since this is mostly an enhancement and the bug hasn't been seen in non-testing use cases.

src/celeritas/em/data/WentzelOKVIData.hh
amandalund: I would also be grateful if someone (maybe @whokion?) could take a second look at this [factor](https://github.com/Geant4/geant4/blob/dda54bbdcfe0cc76177eaa03a657592aac0b1eb8/source/processes/electromagnetic/standard/src/G4WentzelOKandVIxSection.cc#L114) and where it's used (in [geant4](https://github.com/Geant4/geant4/blob/dda54bbdcfe0cc76177eaa03a657592aac0b1eb8/source/processes/electromagnetic/standard/src/G4WentzelOKandVIxSection.cc#L191) and [celeritas](https://github.com/celeritas-project/celeritas/blob/4fe3cf4e24fb9b7a729c940a741a0e37ae449b4c/src/celeritas/em/xs/WentzelHelper.hh#L334)) and make sure everything looks ok. I'm still not really sure where it comes from or if I have all the units right...

whokion: Mo problem, I will take a second look for this today.

whokion: Okay, it looks all good to me.  This factor is related to`the screening parameter`, $A \equiv \frac{1}{4}(\frac{\hbar}{pR})^2$ (where $p$ is the particle momentum) which is originated from the Wentzel SS differential cross section based on the first Born approximation using a simple potential $V(r) \sim \frac{exp(-r/R)}{r}$ where $R \sim 0.885 Z^{1/3} a_o$  is the `screening radius` in terms of the Bohr radius $a_o$ and the atomic (charge) number $Z$ (i.e, based on Thomas-Fermi model) and then is also a parameter of the angular distribution of the single scattering - please see eq 23, 26 of the paper by J.M. Fernrindez-Varea et al.. When the SS is combined with the WentzelVI multiple scattering, the maximum angle of the scattering is therefore limited by $1-A$ (i.e., the line that you pointed in Geant4).

whokion (approved):
LGTM too.

src/celeritas/em/data/WentzelOKVIData.hh
amandalund: Thanks for the explanation and reference! That makes more sense now.  I'll take another look and see if I can understand the differences between the screening parameter and the value that's used in the code to limit the angle.

amandalund: Ok I fixed another bug (in the calculation of the effective atomic mass^(-2/3)), but the quantity that we're subtracting [here](https://github.com/celeritas-project/celeritas/blob/4fe3cf4e24fb9b7a729c940a741a0e37ae449b4c/src/celeritas/em/xs/WentzelHelper.hh#L334) is still about 4 orders of magnitude smaller than the screening parameter. Should they be roughly the same? The screening parameter itself is also very small (like O(1e-10)) which means the maximum angle is going to be limited to near zero instead of the polar angle limit... so I think I'm still missing something.

whokion: Yes, I guess that the magnitude, 1-factorA^2/ [(mass number)^{2/3} * p^2] should be close to 1, but not too close to 1 - based on my rough calculation, it is about 0.95/0.98 for electrons of 100/200 MeV (kinetic) energy on Pb (A~200).

amandalund:
One more fix... I think `a_sq_factor` should have units MeV^2. The cos(theta) limit looks more reasonable now.

sethrj (approved):
Good work!

PR #1275, opened 2024-06-14
Refactor nuclear zone builder to separate numerical integration

sethrj:
This is a follow-on to discussion in #1269. It factors out the numeric integration, adds testing, and performs simplifications on the `NuclearZoneBuilder`.

### DIFF
--- a/src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
+++ b/src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
@@ -9,6 +9,7 @@
 
 #include <cmath>
 #include <numeric>
+#include <vector>
 
 #include "corecel/cont/Span.hh"
 #include "corecel/math/Integrator.hh"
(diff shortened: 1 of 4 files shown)


src/celeritas/neutron/model/detail/NuclearZoneBuilder.hh
```diff
@@ -242,162 +180,153 @@ void NuclearZoneBuilder::calc_zone_component(IsotopeView const& target,
...
-    real_type delta_r = rmax - rmin;
+    Integrator integrate_gauss{
+        [](real_type r) { return ipow<2>(r) * std::exp(-ipow<2>(r)); }};
 
-    // Average value of \f$ r^{2} V(r) dr \f$ at boundaries
-    real_type integral = half() * delta_r
-                         * (rmin * (rmin + ws_shift) / (1 + std::exp(rmin))
-                            + rmax * (rmax + ws_shift) / (1 + std::exp(rmax)));
+    // Precompute y = sqrt(-log(alpha)) where alpha[3] = {0.7, 0.3, 0.01}
+    real_type const y[] = {0.597223, 1.09726, 2.14597};
```
sethrj, line 241: It occurs to me that since the function we're integrating uses fixed intervals and is constant with respect to the nucleus, we could simply precompute the gaussian integral...

src/corecel/math/Integrator.hh
```diff
@@ -0,0 +1,129 @@
...
+#include "Algorithms.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+//! Solver options
+struct IntegratorOptions
+{
+    real_type epsilon{1e-3};  //!< Convergence criterion
+    int max_depth{30};  //!< Maximum number of outer iterations
```
sethrj: I reduced this from 1000 because each "trial" doubles the integer counter, so we'd overflow in many cases before reaching even 32.

whokion (approved):
Excellent work!  Good to merge!  Thanks a lot for the nice refactoring and the integrator!

PR #1276, opened 2024-06-14
Fix tests for Geant4 11.2

sethrj:
This activates (on CI) and fixes test failures with Geant4 11.2. it also tackles a couple (but not all) of smaller issues with Geant4 11.


amandalund (approved):
Thanks @sethrj!

PR #1277, opened 2024-06-14
Fix aarch64 compilation and add Grace Hopper preset

ptheywood:
+ Fixes compilation warnings on `aarch64` (i.e NVIDIA Grace) where `char` is implicitly `unsigned`
+ Adds CMake Preset for GH200 nodes in the UK Tier 2 Bede HPC facility
  + Includes psABI warning suppression, `-lineinfo` for profiling and a gpudebug configuration (`-G`)
+ Fixes invalid JSON in the `stognini` preset
+ Fixes a the `SetupOptionsMessenger` docstring to include the correct geant4 macro name
  + Changed in #1235

### DIFF
--- a/scripts/cmake-presets/bede.json
+++ b/scripts/cmake-presets/bede.json
@@ -1,0 +1,157 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0},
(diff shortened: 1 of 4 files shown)


sethrj (approved):
Looks perfect! Thanks again for doing this.

PR #1278, opened 2024-06-17
Add auxiliary extensible params and state

### LINKED ISSUE #392: Store extra state data on state rather than actions
There are a growing number of Actions that store data as part of the *class* rather than the *state*, as implemented by the `StreamStore` in #722:
- SimpleCalo 
- StepDiagnostic
- ActionDiagnostic
- OpticalGenStorage

These should all be stored dynamically as part of the `CoreState` rather than in each of the classes

Requirements:
- Initialization happens at `begin_run` (construct?) via `BeginRunActionInterface`?
- Has a "finalize" method to deallocate
- `std::any`-like behavior but specific to a core
- `DataId` for storage/retrieval from `CoreState`

*Eventually*: have "initializing" state for each track so that per-track data can be constructed/reset/etc.

sethrj:
For #886 it would be good to start storing state data alongside the core state rather than using the hacky stream store to make the params class mutable.
- Params classes (e.g. actions) can inherit from `UserParamsInterface`
- `UserParamsInterface` has a function for the params to create a state (it can be called multiple times with a memspace, stream ID, and size)
- A `UserParamsRegistry` holds an array of user params, each associated with a `UserId`
- A `UserStateVec` is created alongside the core state and it holds the stream-local state corresponding to each user params
- A `UserStateData` helper class can wrap a `FooStateData` (collection group)
- A `get` helper function for getting states from a user

A follow-up PR refactors the optical data using this new change set.

Closes #392 .

---

Non-critical follow-on work:  *how to combine and output params data with the results of multiple states*. Currently the states are *unique* pointers owned by the core state and created when we build the stepper, and outputs are usually *shared* pointers that are registered when the params are created. Maybe we should punt on the output for now (so we can get this into the optical state) and consider having a separate "multithread output" class, or using actions to integrate over the streams at the end of a run, or something...

The action diagnostic execution kernel looks like this:
```c++
void ActionDiagnostic::execute(CoreParams const& params,
                               CoreStateHost& state) const
{
    auto execute = make_active_track_executor(
        params.ptr<MemSpace::native>(),
        state.ptr(),
        detail::ActionDiagnosticExecutor{
            this->ref<MemSpace::native>(),
            state.user_state_data<ParticleTallyStateData>(user_id_)});
    return launch_action(*this, params, state, execute);
}
```


src/celeritas/global/CoreState.cc
```diff
@@ -18,6 +18,10 @@
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
+//! Support polymorphic deletion
+CoreStateInterface::~CoreStateInterface() = default;
```
esseivaju, line 23: Do we need the out-of-line definition?

src/corecel/data/UserInterface.cc
```diff
@@ -0,0 +1,21 @@
...
+//---------------------------------------------------------------------------//
+//! \file corecel/data/UserInterface.cc
+//---------------------------------------------------------------------------//
+#include "UserInterface.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+//! Default destructor.
+UserStateInterface::~UserStateInterface() = default;
```
esseivaju: Same here, is the out-of-line definition needed?

src/corecel/data/UserParamsRegistry.hh
```diff
@@ -0,0 +1,97 @@
...
+
+    // Access params at the given ID
+    inline SPParams const& at(UserId);
+    inline SPConstParams at(UserId) const;
+
+    // Get the label corresponding to user params
+    inline std::string const& id_to_label(UserId id) const;
+
+    // Find the ID corresponding to an label
+    UserId find(std::string const& label) const;
```
esseivaju:
```suggestion
    UserId find(std::string_view label) const;
```

src/corecel/data/UserStateVec.hh
```diff
@@ -0,0 +1,115 @@
...
+//---------------------------------------------------------------------------//
+// FREE FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Get a mutable item from a state vector efficiently and safely.
+ */
+template<class S>
+S& get(UserStateVec& vec, UserId uid)
+{
+    CELER_EXPECT(uid < vec.size());
```
esseivaju:
```suggestion
    static_assert(std::is_base_of_v<UserStateInterface, S>);
    CELER_EXPECT(uid < vec.size());
```

esseivaju (approved):
Nice! LGTM, just a few minor comments.

amandalund (approved):
This looks great, thanks @sethrj! The only thing I'm a little unsure about is the name `User`; it definitely fits for the diagnostics, but maybe not as much when we're using this for the optical data.

src/corecel/data/UserParamsRegistry.hh
sethrj: I think we already discussed a similar feedback in a previous PR: the `template<T> find` for `unordered_map` is implemented in C++20 but not 17, so this would convert strings to views back to strings to do the search. (For `std::map` surprisingly it was available since 14).

src/celeritas/global/CoreState.cc
sethrj: I've been doing this as a matter of habit for pure abstract classes for some time to avoid "weak vtables": 
In [GCC](https://gcc.gnu.org/onlinedocs/gcc/Vague-Linkage.html):
>  If the class declares any non-inline, non-pure virtual functions, the first one is chosen as the “key method” for the class, and the vtable is only emitted in the translation unit where the key method is defined.

I think it's less important if the hierarchy is only used within a single library, and more important in a case like this where a potentially arbitrary number of libraries subclasses an interface.

See https://stackoverflow.com/questions/12024642/placing-of-external-virtual-tables/12025816#12025816 , https://stackoverflow.com/a/23749273/118160 .

src/celeritas/global/CoreParams.hh
```diff
@@ -54,6 +55,7 @@ class CoreParams final : public ParamsDataInterface<CoreParamsData>
     using SPConstWentzelOKVI = std::shared_ptr<WentzelOKVIParams const>;
     using SPActionRegistry = std::shared_ptr<ActionRegistry>;
     using SPOutputRegistry = std::shared_ptr<OutputRegistry>;
+    using SPUserRegistry = std::shared_ptr<AuxParamsRegistry>;
```
amandalund, line 58:
```suggestion
    using SPAuxRegistry = std::shared_ptr<AuxParamsRegistry>;
```

test/celeritas/GlobalTestBase.hh
```diff
@@ -73,6 +74,7 @@ class GlobalTestBase : public Test
 
     using SPActionRegistry = SP<ActionRegistry>;
     using SPOutputRegistry = SP<OutputRegistry>;
+    using SPUserRegistry = SP<AuxParamsRegistry>;
```
amandalund, line 77:
```suggestion
    using SPAuxRegistry = SP<AuxParamsRegistry>;
```

amandalund (commented):
A couple `User`s left (in case you want to replace them in #1280).

PR #1279, opened 2024-06-17
Tweak documentation in CoreState

sethrj:
This standardizes the formatting of the documentation in CoreState relating to the action diagnostics. It also removes a few unhelpful `constexpr` annotations, and corrects a semantic issue by changing a string `data` accessor to `c_str`.

(Split off from #1278 )


esseivaju (approved):
LGTM! Thanks for that!

PR #1280, opened 2024-06-17
Refactor optical data to use extensible params/states

sethrj:
This refactors the optical collector to use "user" params and state rather than a state store.


src/celeritas/optical/detail/ScintPreGenAction.cu
```diff
@@ -30,13 +30,13 @@ namespace detail
...
+        = get<OpticalGenState<MemSpace::native>>(core_state.aux(), data_id_);
+
     TrackExecutor execute{core_params.ptr<MemSpace::native>(),
                           core_state.ptr(),
                           detail::ScintPreGenExecutor{
-                              scintillation_->device_ref(),
-                              storage_->obj.state<MemSpace::native>(
-                                  core_state.stream_id(), core_state.size()),
-                              storage_->size[core_state.stream_id().get()]}};
+                              scintillation_->device_ref(), state.store.ref()}};
```
amandalund: Is this missing the buffer size?
```suggestion
                              scintillation_->device_ref(), state.store.ref(), state.buffer_size}};
```

amandalund (approved):
Nice, thanks @sethrj!

src/celeritas/optical/detail/ScintPreGenAction.cu
sethrj: Goodness, why did that compile, and even worse how does the test pass? 👀 Good catch. I've checked the other call signatures.

esseivaju (approved):

PR #1283, opened 2024-06-20
Check units in JSON input

sethrj:
This writes units and version metadata to some of the JSON I/O objects in Celeritas. This helps ensure consistency between intended user input and the actual runtime Celeritas.

### DIFF
--- a/src/celeritas/ext/GeantPhysicsOptionsIO.json.cc
+++ b/src/celeritas/ext/GeantPhysicsOptionsIO.json.cc
@@ -18,7 +18,7 @@
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
-static char const format_str[] = "GeantPhysicsOptions";
+static char const format_str[] = "geant-physics";
 
 //---------------------------------------------------------------------------//
 void from_json(nlohmann::json const& j, MscModelSelection& value)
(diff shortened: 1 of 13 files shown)


src/celeritas/ext/GeantPhysicsOptionsIO.json.cc
```diff
@@ -17,6 +17,9 @@
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
+static char const format_str[] = "GeantPhysicsOptions";
```
stognini: DO you think using something like `typeid(class).name()` to setup the names would be easier, in case a class name gets refactored?

sethrj: No, for two reasons: 1. type_info::name isn't constexpr so can't be evaluated at compile time; 2. if we change the class name we'd still want to keep the input the same. But it's a good point that maybe we shouldn't name it exactly after the class: perhaps a plain text string "Geant4 physics options", or "geant-physics-options"? The other "formats" we have are "ORANGE" and "SCALE ORANGE", for example.

stognini: > 1. type_info::name isn't constexpr so can't be evaluated at compile time; 2. if we change the class name we'd still want to keep the input the same

Good points. Okay, having a plain text, as you suggested, might be a better option. It avoid a correlation like I thought of, that we tend to keep the string being equal the class name, and removes the chances of being mistankely replaced in a class refactoring.

stognini:
Thanks @sethrj . Looks like the tests failing are just missing the new json terms.

stognini (approved):

PR #1285, opened 2024-06-20
Update perlmutter build

esseivaju:
This updates the Perlmutter build script and fixes configure issues with xpmem module that is loaded by default. Also, use compiler wrapper and ccache.

In addition, fix a configure issue when using CMake 3.27 and VecGeom+cuda. CMake 3.27 introduces the policy CMP0146 removing the FindCuda module. Since #1253 we updated CMake policy support to 3.28, setting CMP0146 to NEW.  The latest VecCore (0.8.1) version still uses the old policy and will fail to find cuda with a recent version of CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,10 +367,14 @@ if(CELERITAS_USE_VecGeom)
   set(_min_vecgeom_version 1.2.4)
   if(NOT VecGeom_FOUND)
     # TODO: remove once we require a veccore version including https://github.com/root-project/veccore/commit/743566fac1e9b2eaeb0f0b63242442ba430e0cc0
-    cmake_policy(PUSH)
-    cmake_policy(SET CMP0146 OLD)
-    find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
-    cmake_policy(POP)
+    if(POLICY CMP0146 AND CELERITAS_USE_CUDA)
+      cmake_policy(PUSH)
+      cmake_policy(SET CMP0146 OLD)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -366,7 +366,15 @@ endif()
...
-    find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
+    # TODO: remove once we require a veccore version including https://github.com/root-project/veccore/commit/743566fac1e9b2eaeb0f0b63242442ba430e0cc0
+    if(POLICY CMP0146 AND CELERITAS_USE_CUDA)
+      cmake_policy(PUSH)
+      cmake_policy(SET CMP0146 OLD)
+      find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
+      cmake_policy(POP)
+    else()
+      find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
+    endif()
```
sethrj: Could this be moved into FindVecgeom.cmake? As it is, would we still get the warning/error from downstream packages using the `find_dependency` call (or is that OK because it doesn't have the newer cmake policy?)

esseivaju: We can move this to FindVecgeom.cmake. CMake >= 3.27 won't output any error/warning from downstream packages. CMake versions between 3.10 (included) and 3.27 (not included) don't know the policy so we'd get a warning whenever CMake decided to introduce one but this is also the case without these changes.

sethrj (approved):

PR #1286, opened 2024-06-22
Implement infix evaluator

esseivaju:
This is a simple implementation, we still pre-compute the short-circuited senses.

### DIFF
--- a/src/orange/OrangeTypes.hh
+++ b/src/orange/OrangeTypes.hh
@@ -224,11 +224,13 @@ namespace logic
 // The enum values are set to the highest 4 values of logic_int.
 enum OperatorToken : logic_int
 {
-    lbegin = logic_int(~logic_int(4)),
+    lbegin = logic_int(~logic_int(6)),
     ltrue = lbegin,  //!< Push 'true'
     lor,  //!< Binary logical OR
     land,  //!< Binary logical AND
     lnot,  //!< Unary negation
+    lpar_open,  //!< Open parenthesis
+    lpar_close,  //!< Close parenthesis
     lend
 };
 }  // namespace logic
@@ -472,7 +474,7 @@ namespace logic
 {
 inline constexpr char to_char(OperatorToken tok)
 {
-    return is_operator_token(tok) ? "*|&~"[tok - lbegin] : '\a';
+    return is_operator_token(tok) ? "*|&~()"[tok - lbegin] : '\a';
 }
 }  // namespace logic
 
--- a/src/orange/OrangeTypes.hh
+++ b/src/orange/OrangeTypes.hh
@@ -221,7 +221,7 @@ enum class BoundaryResult : bool
 namespace logic
 {
 //! Special logical Evaluator tokens.
-// The enum values are set to the highest 4 values of logic_int.
+// The enum values are set to the highest 6 values of logic_int.
 enum OperatorToken : logic_int
 {
     lbegin = logic_int(~logic_int(6)),
--- a/src/orange/OrangeTypes.hh
+++ b/src/orange/OrangeTypes.hh
@@ -225,12 +225,12 @@ namespace logic
 enum OperatorToken : logic_int
 {
     lbegin = logic_int(~logic_int(6)),
-    ltrue = lbegin,  //!< Push 'true'
+    lopen = lbegin,  //!< Open parenthesis
+    lclose,  //!< Close parenthesis
+    ltrue,  //!< Push 'true'
     lor,  //!< Binary logical OR
     land,  //!< Binary logical AND
     lnot,  //!< Unary negation
-    lpar_open,  //!< Open parenthesis
-    lpar_close,  //!< Close parenthesis
     lend
 };
 }  // namespace logic
@@ -474,7 +474,7 @@ namespace logic
 {
 inline constexpr char to_char(OperatorToken tok)
 {
-    return is_operator_token(tok) ? "*|&~()"[tok - lbegin] : '\a';
+    return is_operator_token(tok) ? "()*|&~"[tok - lbegin] : '\a';
 }
 }  // namespace logic
 
--- a/src/orange/univ/detail/InfixEvaluator.hh
+++ b/src/orange/univ/detail/InfixEvaluator.hh
@@ -63,7 +63,9 @@ CELER_FUNCTION bool InfixEvaluator::operator()(SpanConstSense values) const
 {
     bool result{true};
 
-    for (size_type par_depth{0}, i{0}; i < logic_.size(); ++i)
+    int par_depth{0};
+    size_type i{0};
+    while (i < logic_.size())
     {
         if (logic_int const lgc{logic_[i]}; !logic::is_operator_token(lgc))
         {
@@ -78,27 +80,28 @@ CELER_FUNCTION bool InfixEvaluator::operator()(SpanConstSense values) const
                 break;
             }
             --par_depth;
-            i = short_circuit(i);
+            i = this->short_circuit(i);
         }
         else if (lgc == logic::ltrue)
         {
             result = true;
         }
-        else if (lgc == logic::lpar_open)
+        else if (lgc == logic::lopen)
         {
             ++par_depth;
         }
-        else if (lgc == logic::lpar_close)
+        else if (lgc == logic::lclose)
         {
+            CELER_ASSERT(par_depth > 0);
             --par_depth;
         }
         else if (lgc == logic::lnot && i + 1 < logic_.size())
         {
-            auto next = logic_[++i];
             // negation of a sub-expression is not supported
-            CELER_EXPECT(!logic::is_operator_token(next));
-            result = !static_cast<bool>(values[next]);
+            CELER_EXPECT(!logic::is_operator_token(logic_[i + 1]));
+            result = !static_cast<bool>(values[logic_[++i]]);
         }
+        ++i;
     }
     return result;
 }
@@ -109,15 +112,18 @@ CELER_FUNCTION bool InfixEvaluator::operator()(SpanConstSense values) const
  */
 CELER_FUNCTION uint32_t InfixEvaluator::short_circuit(uint32_t i) const
 {
-    for (size_type par_depth{1}; par_depth > 0;)
+    int par_depth{1};
+    while (par_depth > 0)
     {
-        if (logic_int const lgc{logic_[++i]}; lgc == logic::lpar_open)
+        switch (logic_[++i])
         {
-            ++par_depth;
-        }
-        else if (lgc == logic::lpar_close)
-        {
-            --par_depth;
+            case logic::lopen:
+                ++par_depth;
+                break;
+            case logic::lclose:
+                --par_depth;
+                break;
+            default:;
         }
     }
     return i;
--- a/src/orange/univ/detail/InfixEvaluator.hh
+++ b/src/orange/univ/detail/InfixEvaluator.hh
@@ -18,6 +18,13 @@ namespace detail
 //---------------------------------------------------------------------------//
 /*!
  * Evaluate an explicit infix logical expression applied to a vector of senses.
+ *
+ * Exlicit infix notation explicitly spells out the intersection operator,
+ * i.e., \c logic::land wehereas implicit infix notation omits it.
+ *
+ * These two expressions are equivalent:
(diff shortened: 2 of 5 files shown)


src/orange/OrangeTypes.hh
```diff
@@ -221,14 +221,16 @@ enum class BoundaryResult : bool
...
 enum OperatorToken : logic_int
 {
-    lbegin = logic_int(~logic_int(4)),
+    lbegin = logic_int(~logic_int(6)),
     ltrue = lbegin,  //!< Push 'true'
     lor,  //!< Binary logical OR
     land,  //!< Binary logical AND
     lnot,  //!< Unary negation
+    lpar_open,  //!< Open parenthesis
+    lpar_close,  //!< Close parenthesis
```
sethrj: I think (?) these operators need to be consistent with the definitions in SCALE; can you put them before `ltrue`? And maybe for brevity just call them `lopen` and `lclose`

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,131 @@
...
+    using SpanConstLogic = LdgSpan<logic_int const>;
+    using SpanConstSense = Span<Sense const>;
+    //@}
+
+  public:
+    // Construct with a view to some logic definition
+    explicit CELER_FORCEINLINE_FUNCTION InfixEvaluator(SpanConstLogic logic);
+
+    // Evaluate a logical expression, substituting bools from the vector
+    inline CELER_FUNCTION bool operator()(SpanConstSense values) const;
```
sethrj: You could also define this proactively as a template function (we'll have to do this later) and pass `[values](FaceId i) {CELER_EXPECT(i < values.size()); return values[i]; }`

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,131 @@
...
+        }
+        else if ((lgc == logic::lor && result)
+                 || (lgc == logic::land && !result))
+        {
+            if (par_depth == 0)
+            {
+                break;
+            }
+            --par_depth;
+            i = short_circuit(i);
```
sethrj:
```suggestion
            i = this->short_circuit(i);
```

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+        {
+            ++par_depth;
+        }
+        else if (lgc == logic::lpar_close)
+        {
+            --par_depth;
+        }
+        else if (lgc == logic::lnot && i + 1 < logic_.size())
+        {
+            auto next = logic_[++i];
```
esseivaju: GCC doesn't like that with the statically sized test case and warns about out-of-bound access even though it only happens if the logic expression is invalid. It doesn't have any problem with the access in short-circuit though... and it's inconsistent: some configurations will trigger the warning, some won't. `CELER_EXPECT` isn't enough to fix the warning

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+        {
+            result = true;
+        }
+        else if (lgc == logic::lpar_open)
+        {
+            ++par_depth;
+        }
+        else if (lgc == logic::lpar_close)
+        {
+            --par_depth;
```
sethrj, line 106: Should have an assertion about this being positive. Maybe this should also be an int so that we don't risk overflowing?

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+#include "corecel/cont/Span.hh"
+#include "orange/OrangeTypes.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Evaluate an explicit infix logical expression applied to a vector of senses.
```
sethrj, line 20: Add a comment about what "explicit" here means—parens included? Add an example of a valid logic string.

src/orange/univ/detail/InfixEvaluator.hh
sethrj:
```suggestion
        else if (lgc == logic::lnot)
        {
            CELER_ASSUME(i + 1 < logic_.size());
            auto next = logic_[++i];
```
Perhaps `ASSUME` will work? It uses `__builtin_unreachable` to help the compiler figure out what to do, and it'll raise a debug error if violated.

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Evaluate a logical expression, substituting bools from the vector.
+ * TODO: values don't need to be pre-calculated
+ */
+CELER_FUNCTION bool InfixEvaluator::operator()(SpanConstSense values) const
+{
+    bool result{true};
+
+    for (size_type par_depth{0}, i{0}; i < logic_.size(); ++i)
```
sethrj: Since both `i` and `par_depth` are changing outside of the `for` expression, I think this should be a `while`.

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+  public:
+    // Construct with a view to some logic definition
+    explicit CELER_FORCEINLINE_FUNCTION InfixEvaluator(SpanConstLogic logic);
+
+    // Evaluate a logical expression, substituting bools from the vector
+    inline CELER_FUNCTION bool operator()(SpanConstSense values) const;
+
+  private:
+    // Short-circuit evaluation of the second operand
+    inline CELER_FUNCTION uint32_t short_circuit(uint32_t i) const;
```
sethrj: Instead of `uint32_t` should be `logic_int`?

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+        {
+            --par_depth;
+        }
+        else if (lgc == logic::lnot && i + 1 < logic_.size())
+        {
+            auto next = logic_[++i];
+            // negation of a sub-expression is not supported
+            CELER_EXPECT(!logic::is_operator_token(next));
+            result = !static_cast<bool>(values[next]);
+        }
```
sethrj, line 114: Maybe there should be an else unreachable?

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Short-circuit evaluation of the second operand.
+ */
+CELER_FUNCTION uint32_t InfixEvaluator::short_circuit(uint32_t i) const
+{
+    for (size_type par_depth{1}; par_depth > 0;)
```
sethrj: Also here `while` seems more appropriate

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,128 @@
...
+    {
+        if (logic_int const lgc{logic_[++i]}; lgc == logic::lpar_open)
+        {
+            ++par_depth;
+        }
+        else if (lgc == logic::lpar_close)
+        {
+            --par_depth;
```
sethrj, line 136: And an assertion here too

test/orange/univ/detail/InfixEvaluator.test.cc
```diff
@@ -0,0 +1,130 @@
...
+    EXPECT_EQ('|', to_char(lor));
+    EXPECT_EQ('&', to_char(land));
+    EXPECT_EQ('~', to_char(lnot));
+    EXPECT_EQ('(', to_char(lpar_open));
+    EXPECT_EQ(')', to_char(lpar_close));
+}
+
+TEST(InfixEvaluatorTest, evaluate)
+{
+    // Logic for alpha : !1 | 2 | !3 | 4 | !8
```
sethrj, line 60: What about combining the operator with parens? In practice we only have "any of..." or "all of..." and don't mix within a subexpression. This would reduce the string size at the cost of having a local Boolean stack of operators associated with the open parens, I think?

So: `&( 1 2 -3 4 ) |( 5 -6 &(7 8 ) )` where we have "land" and "lor " be combined operator plus open parens

src/orange/univ/detail/InfixEvaluator.hh
esseivaju: It would be reachable if `lor` or `land` don't short-circuit

src/orange/univ/detail/InfixEvaluator.hh
esseivaju: the assertion is built in the loop condition

src/orange/univ/detail/InfixEvaluator.hh
esseivaju: this is the index into the `logic_` span, so it would make more sense for it to be a `size_type`

test/orange/univ/detail/InfixEvaluator.test.cc
esseivaju: That's an option but then this would be the prefix notation, the point of the infix notation is to not require a stack

src/orange/univ/detail/InfixEvaluator.hh
esseivaju: Should we also update the `LogicEvaluator`to have the same API?

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,146 @@
...
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Evaluate an explicit infix logical expression applied to a vector of senses.
+ *
+ * Exlicit infix notation explicitly spells out the intersection operator,
```
sethrj:
```suggestion
 * Explicit infix notation explicitly spells out the intersection operator,
```

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -0,0 +1,146 @@
...
+        CELER_ASSUME(i + 1 < logic_.size());
+        switch (logic_[++i])
+        {
+            case logic::lopen:
+                ++par_depth;
+                break;
+            case logic::lclose:
+                --par_depth;
+                break;
+            default:;
```
sethrj: This looks a little weird... did you use `switch` instead of `if/else` to silence a compiler warning?

test/orange/univ/detail/InfixEvaluator.test.cc
sethrj: Taking another look at the paper, I think it would be more along the lines of semi-explicit infix notation. You still get the short circuiting (which is I think the *main* point of the infix notation) but have a stack to keep track of the operators. I think we should try this at some point since we're almost always going to be using `&` and using a string of identical rather than mixed logic operators. CNF and DNF forms (union of intersection and intersection of unions) of boolean expressions will only have a depth of 2, and we're unlikely to have nested parens deeper than 32 (and can always expand the tree as needed), so we can still use a single-int bit string as a stack.

Future work!

test/orange/univ/detail/InfixEvaluator.test.cc
```diff
@@ -0,0 +1,133 @@
...
+                                    land,
+                                    6,
+                                    lclose,
+                                    land,
+                                    lnot,
+                                    7,
+                                    lclose,
+                                    land,
+                                    8,
+                                    lclose};
```
sethrj: Clang-format does this awkward aligned column to me on occasion... adding a trailing comma will at least indent it so it's no longer aligned, if you're interested.

sethrj (approved):
Thanks Julien! Coming along nicely.

src/orange/univ/detail/InfixEvaluator.hh
esseivaju: It was mostly because switch over enum variants seems more appropriate than if but I agree that the empty default is a bit weird.

src/orange/univ/detail/InfixEvaluator.hh
sethrj: Either way is fine with me. Less effort to leave as-is?

PR #1287, opened 2024-06-22
Regenerate ROOT files during CI testing

sethrj:
This updates the Geant4 configuration file for the `four-steel-slabs` ROOT input used by the UrbanMsc testing so that the committed data matches the regenerated data. To ensure consistency in the future (i.e. regenerating the file won't break the tests) the CI now regenerates (but does not commit!) the files as part of testing.

I also noticed during testing that the neutron XS environment variable isn't being passed to the tests; so this fixes that and adds an additional check to fail the CI if the environment varibles are missing inside the test.


test/celeritas/phys/ProcessBuilder.test.cc
```diff
@@ -71,16 +71,26 @@ class ProcessBuilderTest : public Test
         CELER_ENSURE(particle() && material());
     }
 
+    static bool has_env(std::string const& var)
+    {
+        bool result = !celeritas::getenv("G4LEDATA").empty();
```
amandalund:
```suggestion
        bool result = !celeritas::getenv(var).empty();
```

amandalund (approved):
Thanks @sethrj!

stognini (approved):

Issue #907, opened 2023-08-25
Refactor physics tables for correct material-dependent bounds


sethrj:
The physics cross section data and applicability model was implemented very early in our project: the cross section construction uses some very awkward classes, the physics "construct host data" is super complicated, and most importantly the applicability doesn't allow us to limit correctly the lower energy of models based on their production cutoff, so we end up sampling inapplicable discrete models (most seriously in the case of stopped positrons).

---

Priority ordering for implementation parts:
- [x] Use one dE/dx grid per particle rather than per process (#253)
- [x] Composite multiple grids (or use generic grid) for Geant4 11.1 support (#588)
- [ ] Replace applicability to support correct model bounds per material as a function of energy (so that we can correctly treat positrons when slowing down, for example)
- [ ] Replace material lookup with a `{material, region} -> PhysMaterial` mapping (see #983)
- [ ] Combine secondary production/range cutoffs into physics
- [ ] ~~Possibly integrate "MSC model" into physics? (*not* the same type as other physics processes though)~~

Replace applicability:
- Model selection: grid of `[particle][material][energy bin] -> model`
- Physics lists should be the same "per region", so process/models should be a function of [region][particle][energy]
- Energy grid structure can be different for each material (and of course for each particle type)
- Energy grid will live on `[0, ∞)` where zero is applicable to at-rest processes

Replace cross section construction:
- Define different energy grid input classes (log, log+scaled, nonuniform) and use *variants* for the CPU interface
- Define visitors that construct the data on device (emplacing a grid type + sub-grid ID)

Build order:
- User supplies processes
- Process provides a set of materials (for now require all processes in all materials since that's true for EM)
- Process creates models (models are same for all materials but *may* have material-dependent parameters)
- Each model provides a set of applicable particles and energy range for each {particle type, material}: for 'imported' data, these ranges should be exactly the range of the existing cross sections
- Process+model detail construction ensures no gaps or overlaps between model ranges

If we do our own cross section generation...
- Model micro XS calculation: depends on secondary cutoff, loop over `{material, element, particle} -> {{slowing down, micro XS}, energy}`
- Model macro XS calculation + element CDF:  integrate micros over elements for each material (and discard micros for models that don't need them, e.g.  Compton), and normalize
- Process macro XS calculation: combine non-overlapping energy grids from models for XS and slowing down
- Accumulate slowing-down over all processes into a `{particle, material, energy}`-dependent field
- Numerically integrate dE/dx to get range
[...]

PR #1288, opened 2024-06-23
Differentiate "geometry" from "physics" materials in ImportData

sethrj:
This differentiates between true material compositions associated with the geometry, and "physics" materials which are modified by physics cutoffs inside a region that's applied to a particular volume. The "geometry materials" are equivalent to G4Material, and the "physics material" are the G4MaterialCutsCouple which is what we use in Celeritas (and need to generalize for #907, #983).


src/celeritas/io/ImportParameters.hh
```diff
@@ -17,6 +17,24 @@
...
 {
+//---------------------------------------------------------------------------//
+/*!
+ * User-imposed cutoffs and restrictions.
+ *
+ * Used as tracking cuts, see \c G4UserLimits , used by \c G4UserSpecialCuts.
+ * As "production" cuts, only for electron/positron/proton/gamma, see \c
+ * G4ProductionCuts .
+ */
+struct ImportCutoff
```
stognini: I'm pondering about this name. This is another set of cuts beyond the `ImportProductionCuts` which originally are loaded into our `CutoffParams`---which bears a strikingly similar name to this one. So maybe we need that to be `ProductionCutParams` instead?

sethrj: I guess these are tracking cuts; I think we should combine them with production cuts in the codebase. I've been trying to untangle the particle production cuts, regions, "cut couple", etc. I forgot that I didn't actually implement the reading for these cuts from G4UserLimits .... I'll just remove them for this PR and we can add them to an "extended cut" set later. We should also determine how to make the `apply_cuts` and `lowest_electron_energy` EM parameters consistent with these options.

stognini: Yep, you reached another question I had while going over it: that this `ImportCutoff` struct is not being used yet. Okay, that sounds good to me. And later on, I think it makes sense to keep a `CutoffParams` that encompasses both production cuts and extended cuts by adding both to the `CutoffParamsData`.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -709,38 +755,45 @@ import_materials(GeantImporter::DataSelection::Flags particle_flags)
...
+    CELER_ENSURE(!materials.empty());
+    return materials;
+}
 
-            // Add material's element information
-            material.elements.push_back(elem_comp);
-        }
+//---------------------------------------------------------------------------//
+/*!
+ * Return a populated \c ImportVolume vector.
```
amandalund:
```suggestion
 * Return a populated \c ImportRegion vector.
```

app/celer-dump-data.cc
```diff
@@ -55,8 +55,8 @@ void print_particles(ParticleParams const& particles)
     cout << R"gfm(
 # Particles
 
-| Name              | PDG Code    | Mass [MeV] | Charge [e] | Decay [1/s] |
-| ----------------- | ----------- | ---------- | ---------- | ----------- |
+| Name              | PDG Code    | Mass [MeV] | Charge [e] | Decay [1/time] |
+| ----------------- | ----------- | ---------- | ---------- | -------------- |
```
amandalund, line 60:
```diff
-             << setw(11) << setprecision(3) << p.decay_constant()
+             << setw(14) << setprecision(3) << p.decay_constant()
```

app/celer-dump-data.cc
```diff
@@ -176,19 +196,29 @@ void print_materials(std::vector<ImportMaterial>& materials,
...
     //// PRINT CUTOFF LIST ///
 
     cout << R"gfm(
-## Cutoffs
+# Material physics
 
-| Material                   | Particle  | Energy [MeV] | Range [cm] |
-| -------------------------- | --------- | ------------ | ---------- |
+| Material ID/Name                | Particle  | Energy [MeV] | Range [len] |
+| ------------------------------- | --------- | ------------ | ----------- |
```
amandalund, line 219:
```diff
-                 << setw(12) << cuts.energy << " | " << setw(10) << cuts.range
+                 << setw(12) << cuts.energy << " | " << setw(11) << cuts.range
```

amandalund (approved):
Looks good to me!

stognini (approved):

stognini:
Looks great. Thanks!

Issue #1282, opened 2024-06-18
Implement short-circuit volume evaluation using infix notation


sethrj:
Based off some [still unpublished but available research](http://arxiv.org/abs/2406.13030) by @paulromano et al., there could be substantial savings in the way that we evaluate point-in-volume for initialization and surface crossing in ORANGE. (It's also used for distance-to-boundary for "complex" and "background" volumes since internal surface crossings have to be evaluated.)

I think we can break it down into the following steps:
1. Develop an on-device `InfixEvaluator` that applies a function `f(literal) -> bool` (where `literal` usually means a surface index possibly with negation) and returns the evaulated result, using short circuit logic. (#1286)
2. Add a CSG "simplifier" that replaces negated `join`s with the opposite kind of `join` using DeMorgan's laws. (#1289)
3. Add a utility to generate infix notation from the simplified tree (#1530)
4. Add a utility to import from postfix notation  (external JSON) to CSG tree for SCALE compatibility (@sethrj ) (#1582)
5. Pass an infix logic representation rather than prefix through the OrangeInput into the params data (#1565)

After using the infix evalutor and comparing performance, then we can replace the "precalculate surface senses" operation with "on-the-fly surface senses". (#1539, #1569, #1576 )

sethrj:
Question for @paulromano : on-the-fly sense evaluation works for "point in volume" but might be difficult for "distance to boundary" in the case with internal surface crossings. To do that right now, we precalculate all the surfaces with the starting point, progressively move through nearest surfaces, flip the senses as we cross them, and evaluate "inside" at each one of those points. For the internal surface case, do you re-evaluate at every internal surface crossing? (I guess we could probably cache just a single surface state rather than all of them...?)

sethrj:
@esseivaju Regarding your questions today about infix:
1. Let's get the building blocks in to the point where we can throw a switch to activate it.
2. Add an enum field to the top-level ORANGE input for "infix"/"postfix"; if it's not the one that's used by the code, convert it when loading from disk in `OrangeParams::from_json`.
4. We'll get a couple of @elliottbiondo's test problems (https://github.com/celeritas-project/docs/issues/12) in before doing performance comparisons
5. Assuming infix is not slower (which would be a shock) we commit and close this out 😄

esseivaju:
> 1. Let's get the building blocks in to the point where we can throw a switch to activate it.

[...]

PR #1289, opened 2024-06-24
Implement operation to simplify negated join on CsgTree

esseivaju:
Required to generate infix notation for  #1286  as part of #1282. The goal is to eliminate negated `Join` operations as those cannot be represented with infix notation.


src/orange/orangeinp/CsgTree.cc
```diff
@@ -166,6 +166,65 @@ auto CsgTree::simplify(NodeId node_id) -> Simplification
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Simplify negated joins using De Morgan's law.
+ *
+ * This is required if the tree's logic expression is used with
+ * \c InfixEvaluator as negated joins are not supported.
+ */
+void CsgTree::simplify_negated_joins()
```
sethrj: Is it possible to make this an external `CsgTreeUtils` function rather than extending the class?

esseivaju:
@sethrj @elliottbiondo I still need to add tests and finalize the implementation but I think the logic is good now.

esseivaju:
I'll add more complex test cases but this is ready for an initial review

src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -165,6 +171,273 @@ void simplify(CsgTree* tree, NodeId start)
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Simplify negated joins using De Morgan's law.
+ *
+ * This is required if the tree's logic expression is used with
+ * \c InfixEvaluator as negated joins are not supported.
+ */
+CsgTree transform_negated_joins(CsgTree const& tree)
```
sethrj, line 176: This function is too long; is there a sensible way to break it up into a detail helper class? Probably making these maps as class data, maybe a couple helper functions that you could call in sequence, turn the lambda into a member function.

src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -165,6 +171,273 @@ void simplify(CsgTree* tree, NodeId start)
...
+    // Contains the node_id of Negated{Joined}. These don't need to be inserted
+    // in the new tree and its parent can directly point to the transformed
+    // Joined node.
+    std::unordered_map<NodeId, Negated const*> transformed_negated_nodes;
+
+    // A Negated{Joined{}} nodes needs to be mapped to the opposite Joined{}
+    // node. Parents need to point to the new Joined{} node. The old Joined{}
+    // node needs to be kept if it had parents other than the Negated node.
+    std::unordered_set<NodeId> orphaned_join_nodes;
+    std::unordered_set<NodeId> orphaned_negate_nodes;
```
sethrj: Since all the node ids are consecutive, a vector of bools might be more effective (maybe not worrying about performance wise)

src/orange/orangeinp/CsgTreeUtils.cc
```diff
@@ -165,6 +171,273 @@ void simplify(CsgTree* tree, NodeId start)
...
+
+    // Recursive lambda. Negated{node_id} should be added to the new tree.
+    // if node_id is a Joined node, recusively add its children as negated
+    // nodes.
+    // TODO: extract the lambda to a free function
+    auto add_new_negated_nodes = [&](NodeId node_id) {
+        auto impl = [&](auto const& self, NodeId node_id) -> void {
```
sethrj: Recursive lambdas before C++23??? You're a madman 😂

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -385,6 +385,119 @@ TEST_F(CsgTreeUtilsTest, calc_surfaces)
...
 
+TEST_F(CsgTreeUtilsTest, transform_negated_joins)
+{
+    auto s0 = this->insert(Surface{S{0}});
+    auto s1 = this->insert(Surface{S{1}});
+    auto n0 = this->insert(Negated{s1});
+    auto j0 = this->insert(Joined{op_and, {s0, n0}});
+    auto simplified = transform_negated_joins(tree_);
+
+    // check well-formed tree
```
sethrj: For consistency with the rest of the code, please try to capitalize the comments and write them as a sentence without the trailing `.`, unless it's an adjective phrase describing data or something
```suggestion
    // Check a well-formed tree
```

sethrj (changes requested):
Great commenting of what's going on! I know it's something of a work in progress, but a little restructuring would make this easier to review 😅 I think making the big function into a detail class (even a single-header one) will make it easier to read and process; that'll also make it easier to design a "visitor". Also don't forget the `Overload` class in `corecel/cont/VariantUtils.hh` as an alternative to writing a whole standalone class.

esseivaju:
> Great commenting of what's going on! I know it's something of a work in progress, but a little restructuring would make this easier to review 😅 I think making the big function into a detail class (even a single-header one) will make it easier to read and process; that'll also make it easier to design a "visitor". Also don't forget the `Overload` class in `corecel/cont/VariantUtils.hh` as an alternative to writing a whole standalone class.

I had to add comments or even I'd get lost during the implementation. This function is getting unwieldy and way too long as you mentioned, thanks largely to the rough variant interface...
I wanted to get a few more complex test cases working first, now that it's done it might be worth double-checking that the simplification is the logic you'd expect and that I didn't make any mistakes in the test strings. I had to get the pen and paper out for this, I think they are correct. 
Next, I'll refactor the function to make it  shorter and easier to parse.

src/orange/orangeinp/CsgTreeUtils.cc
esseivaju: let me make sure everything works before using a vector of bool 😅

esseivaju:
It's in a better state now. It could still use better comments and variable names but at least it's easier to review. I think the most important thing to check is the correctness and exhaustiveness of test cases.

sethrj (commented):
One thing I see is that the CSG nodes don't map directly from old to new (naturally), but regions/volumes/etc. in the CSG unit that we're simplifying reference the original node numbers. We'll need to return not only a new CSG tree but also a vector of NodeIds where `remapped[old.get()] == new`

esseivaju:
Is it possible to initialize the regions/volumes/etc after we simplify the tree so that they directly use the correct node_id? If an arbitrary node can be referenced by an external component, there's no guarantee it still is in the simplified tree. 
For example, if we have a `Negated{Joined{}}` and that joined node has no other parents, it will be converted to the opposite join. If someone was keeping a node_id pointing directly to that joined node, there is no equivalent of the original joined node in the simplified tree.

sethrj:
@esseivaju I should've thought to mention this before, but unfortunately it's not going to be that easy: consider the following, a union of two spheres inside a volume enclosed by a larger sphere:
```mermaid
flowchart TB
  n00["t"]
  n01["~"]
  n01 --> n00
  n02["[EXTERIOR]\nbound@s"]
  n03["bound\n~"]
  n03 --> n02
  n04["bottomsph@s"]
  n05["bottomsph\n~"]
  n05 --> n04
  n06["topsph@s"]
  n07["topsph\n~"]
  n07 --> n06
  n08["union\n|"]
  n08 --> n05
  n08 --> n07
  n09["~"]
  n09 --> n08
  n10["fill\n&"]
  n10 --> n03
  n10 --> n09
subgraph Volumes
  v02(["[EXTERIOR]@global"])
  v08(["union_vol"])
  v10(["fill_vol"])
end
  v02 <--> n02
  v08 <--> n08
  v10 <--> n10
```
We need to keep a reference both to `fill_vol` and `union_vol`.

Is the easy solution just to replace `~join(op, ...)` with `join(other_op, ~...)` on insertion to the CSG tree (which would allow us to skirt the "no backward substitution" requirement by injecting negated daughter nodes)? I was hoping we could do this with a "simplification" step but perhaps adding this to `CsgTree::insert` is the best way...

esseivaju:
I don't think moving that to `insert` would solve the problem. In your example, it's impossible to keep the union node and have an equivalent tree after simplification even if you do it during insertion, you'd create a new root. The `intersect_vol` volume would have to point to the negated union. If the graph below is valid after simplification, then it's easy to provide a mapping of old volume to new (negated) volume.

```mermaid
flowchart TB
  n00["t"]
  n01["~"]
  n01 --> n00
  n02["[EXTERIOR]\nbound@s"]
  n03["bound\n~"]
  n03 --> n02
  n04["bottomsph@s"]
  n06["topsph@s"]
  n08["intersect\n&"]
  n08 --> n04
  n08 --> n06
  n10["fill\n&"]
  n10 --> n03
  n10 --> n08
subgraph Volumes
  v02(["[EXTERIOR]@global"])
  v08(["intersect_vol"])
  v10(["fill_vol"])
end
  v02 <--> n02
  v08 <--> n08
  v10 <--> n10
```

esseivaju:
@sethrj where is the Volume<->NodeId mapping stored?

sethrj:
@esseivaju These aren't internally saved anywhere in the CSG tree, but there's a vector of volumes in `CsgUnit` at `orange/orangeinp/detail/CsgUnit.hh`:
```c++
    std::vector<NodeId> volumes;  //!< CSG node of each volume
```
These will all be final at the time when we simplify the tree.

esseivaju:
And is this not going to pose an issue to have that extra disjoint tree when the tree is converted to a logical expression? Maybe we'll need to select which roots we want to convert.

sethrj:
@esseivaju Yeah. We're creating the logical trees from *each* of the volumes as a root. Lots of the time, a given volume doesn't touch every node in the tree.

esseivaju:
Okay, I've moved the `volumes` to the `CsgTree` and updated the simplification example. I'll need to add more complex test cases. This now simplifies to the following tree:

```mermaid
flowchart TB
  n00["t"]
  n01["~"]
  n01 --> n00
  n02["topsph@s"]
  n03["bottomsph@s"]
  n04["topsph\n~"]
  n04 --> n02
  n05["bottomsph\n~"]
  n05 --> n03
  n06["intersect\n&"]
  n06 --> n02
  n06 --> n03
  n07["union\n|"]
  n07 --> n04
  n07 --> n05
  n08["[EXTERIOR]\nbound@s"]
  n09["bound\n~"]
  n09 --> n08
  n10["fill\n&"]
  n10 --> n06
  n10 --> n09
subgraph Volumes
  v02(["[EXTERIOR]@global"])
  v08(["union_vol"])
  v10(["fill_vol"])
end
  v02 <--> n08
  v08 <--> n07
  v10 <--> n10
```

src/orange/orangeinp/CsgTree.hh
```diff
@@ -77,6 +77,12 @@ class CsgTree
     // Simplify a single node in-place [O(1)]
     Simplification simplify(NodeId);
 
+    // Insert a new volume which root is the node id
+    inline void insert_volume(NodeId);
+
+    //! List of volumes defined in this tree
+    std::vector<NodeId> const& volumes() const { return volumes_; }
```
sethrj: Maybe add a type alias

src/orange/orangeinp/UnitProto.cc
```diff
@@ -124,7 +124,7 @@ void UnitProto::build(ProtoBuilder& input) const
 
     // Save unit's bounding box
     {
-        NodeId node_id = csg_unit.volumes[orange_exterior_volume.get()];
+        NodeId node_id = csg_unit.tree.volumes()[orange_exterior_volume.get()];
```
sethrj: Keeping a local reference `auto const& volumes = csg_unit.tree.volumes()` would make this function cleaner

src/orange/orangeinp/CsgTreeIO.json.cc
```diff
@@ -129,11 +129,12 @@ void to_json(nlohmann::json& j, CsgUnit const& unit)
     }();
 
     j["volumes"] = [&unit] {
-        CELER_ASSERT(unit.volumes.size() == unit.fills.size());
+        CELER_ASSERT(unit.tree.volumes().size() == unit.fills.size());
```
sethrj, line 132: I think we should also annotate the volumes in the CSG tree as part of its output, now that it actually contains the volumes.

esseivaju:
@sethrj I think this is ready for another look. The only thing I'm not convinced of yet is that `MatchingNodes::modified` is a bit overloaded and factorizing it into 3 variables would make the algorithm easier to follow.

src/orange/orangeinp/CsgTree.hh
```diff
@@ -114,6 +123,16 @@ auto CsgTree::insert(LocalSurfaceId s) -> Insertion
     return this->insert(orangeinp::Surface{s});
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Insert a new volume which root is the node id.
+ */
+inline void CsgTree::insert_volume(NodeId node_id)
```
sethrj: Redundant
```suggestion
void CsgTree::insert_volume(NodeId node_id)
```

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -0,0 +1,124 @@
...
+    // Perform the simplification
+    CsgTree operator()();
+
+  private:
+    //! Helper struct to translate ids from the original tree to ids in the
+    //! simplified tree
+    struct MatchingNodes
+    {
+        //! Set if a node has the exact same node in the simplified tree
+        std::optional<NodeId> unmodified;
```
sethrj: Since an unset `NodeId` is false (it's supposed to act rather  like "optional") is redundant: you should be able to just use `NodeId unmodified`, same with `modified`

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Perform the simplification. The state of the instance isn't cleared, so only
+ * call this once.
+ */
+CsgTree DeMorganSimplifier::operator()()
+{
+    find_join_negations();
+    return build_simplified_tree();
```
sethrj: Use `this->member_func()` for improved readability (and to make it more obvious why the lambdas need to capture `this`

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+        Overload{
+            [&](Joined const& joined) {
+                // each operand of the Joined node also needs to be
+                // kept
+                for (auto join_operand : joined.nodes)
+                {
+                    record_parent_for(join_operand);
+                }
+                // and the Joined node itself needs to be kept
+                CELER_EXPECT(std::holds_alternative<Joined>(tree_[node_id]));
```
sethrj: This should be completely redundant with the `std::visit` call, right? Since there's no precondition that could trigger the assertion it should be `CELER_ASSERT`. And really, since there's no programming error you could make that would trigger this, you should probably delete it...

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+                   },
+                   tree_[node_id]);
+    }
+
+    // Volume nodes act as tags on a NodeId indicating that it is the root of a
+    // volume, so these subtrees need to be preserved.
+    // Consider a "virtual" parent for these nodes
+    std::unordered_set<NodeId> volume_roots{tree_.volumes().cbegin(),
+                                            tree_.volumes().cend()};
+    for (auto node_id : volume_roots)
```
sethrj: No two volumes should point to the same node: that implies two volumes occupy the same exact region of space, which is a geometry error. We could in theory check this in the CSG tree itself. For now, I think the code will still work even if the same NodeId is shared by two volumes...
```suggestion
    for (auto node_id : tree_.volumes())
```

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+ * that all operands of the join operation have a \c Negated parent added.
+ *
+ * \param negated_id of a \c Negated node pointing to \c Joined node inside
+ * the tree_.
+ */
+void DeMorganSimplifier::record_join_negation(NodeId negated_id)
+{
+    CELER_EXPECT(negated_id < tree_.size());
+
+    CELER_ASSERT(std::holds_alternative<Negated>(tree_[negated_id]));
```
sethrj:
```suggestion
    CELER_EXPECT(std::holds_alternative<Negated>(tree_[negated_id]));
```
This one is a precondition on input :) You can also use `CELER_ASSUME` to help the compiler optimize when assertions are disabled.

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+    auto* negated_node = std::get_if<Negated>(&tree_[negated_id]);
+
+    simplified_negated_nodes_[negated_id.get()] = true;
+
+    CELER_ASSERT(std::holds_alternative<Joined>(tree_[negated_node->node]));
+
+    negated_join_nodes_[negated_node->node.get()] = true;
+    add_negation_for_operands(negated_node->node);
```
sethrj: With `CELER_ASSUME`, a compiler with optimizations won't even generate the exception handling code for `std::get`, so you should be able to write this more cleanly as
```suggestion
    simplified_negated_nodes_[negated_id.get()] = true;

    NodeId negated_join = std::get<Negated>(tree_[negated_id]).node_id;
    CELER_ASSUME(std::holds_alternative<Joined>(tree_[negated_join]));
    negated_join_nodes_[negated_join.get()] = true;
    this->add_negation_for_operands(negated_join);
```

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+ * \param node_id a \c Joined node_id with a \c Negated parent.
+ */
+void DeMorganSimplifier::add_negation_for_operands(NodeId node_id)
+{
+    CELER_EXPECT(std::holds_alternative<Joined>(tree_[node_id]));
+    auto* join_node = std::get_if<Joined>(&tree_[node_id]);
+
+    for (auto const& join_operand : join_node->nodes)
+    {
+        std::visit(Overload{[&](Joined const&) {
```
sethrj: I feel like a couple of chained `get_if`s may be more semantically appropriate for this case.

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        // We need to insert that node in the simplified tree so make a copy
+        Node new_node = tree_[node_id];
+        // We need to update the children's ids in the new tree.
+        std::visit(
+            Overload{
+                [&](Negated& negated) {
+                    negated.node
+                        = node_ids_translation_[negated.node].mod_unmod_or(
```
sethrj: Maybe `replaced` would be be a better name for this function?

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Perform the simplification. The state of the instance isn't cleared, so only
+ * call this once.
+ */
+CsgTree DeMorganSimplifier::operator()()
+{
+    find_join_negations();
```
sethrj: The ordering of the functions in this header is quite hard to follow... maybe order them in roughly the order that they could be called (sorted topological call graph)?
- construct
- operator()
- find join negations
- record_join_negation
- add negation for operand
- ....
- build simplified tree
- process_negated_joined_nodes

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+                // of a double negation, so we don't need to insert it and
+                // redirect to its child. It still needs to be added as a
+                // modified node so that when adding a Joined node, we
+                // correctly redirect the operand looking for it to the
+                // children
+                if (orphaned_nodes_[node_id.get()])
+                {
+                    CELER_EXPECT(
+                        std::holds_alternative<Negated>(tree_[node_id]));
+                    auto* orphan_node = std::get_if<Negated>(&tree_[node_id]);
```
sethrj: Here you could just just use the argument to the lambda function right?

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,443 @@
...
+                }
+                // this node is a Negated node with a join child, we have
+                // already inserted the opposite Joined node
+                if (simplified_negated_nodes_[node_id.get()])
+                {
+                    CELER_EXPECT(
+                        std::holds_alternative<Negated>(tree_[node_id]));
+
+                    auto& negated_node_child
+                        = std::get_if<Negated>(&tree_[node_id])->node;
```
sethrj: same:
```suggestion
                    auto& negated_node_child = negated.node;
```

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -385,6 +385,306 @@ TEST_F(CsgTreeUtilsTest, calc_surfaces)
...
+    // Complex case with a negated join with negated children
+    auto simplified = transform_negated_joins(tree_);
+    EXPECT_EQ(
+        "{0: true, 1: not{0}, 2: surface 0, 3: surface 1, 4: not{2}, 5: "
+        "not{3}, 6: all{2,3}, 7: any{4,5}, 8: surface 2, 9: not{8}, 10: "
+        "all{6,9}, }",
+        to_string(simplified));
+    // Check that the new volumes map to the correct node
+    auto& volumes = simplified.volumes();
+    EXPECT_EQ(volumes.size(), 3);
```
sethrj: Good to use this form to prevent the code crashing in case of an error...
```suggestion
    ASSERT_EQ(volumes.size(), 3);
```

sethrj (commented):
This is an impressive piece of work. Could you please address a couple of these comments which hopefully will continue to make it easier to review? Can you also post another couple of before-and-after CSG trees so we can visually verify?

src/orange/orangeinp/detail/DeMorganSimplifier.cc
esseivaju: Right, it is redundant I'll just remove that

src/orange/orangeinp/detail/DeMorganSimplifier.cc
esseivaju: it's not always replaced, e.g. if the child was a surface.

src/orange/orangeinp/CsgTreeIO.json.cc
esseivaju: Do you mean that you'd have 
```json
{
"tree:" { "nodes": [], "volumes": []}
}
```
instead of the "tree" key being a list?

sethrj: Either that, or add an additional component to the volume nodes: 
```json
["t",["~",0],["S",0],["S",1],["~",3,0],["S",2],["~",5],["&",[2,4,6], 1],["S",3],["~",8],["&",[2,4,9]],["~",7],["&",[10,11]],["S",4],["~",2],["|",[13,14]]]
```
but that's dumb, I think replacing the list with a CSG tree would be more sensible. Then we could even add a `_format` key...

esseivaju: mmh, an entry for a volume needs material data from the CsgUnit which we can't pass to the `to_json` serialization function

esseivaju: With the following changes, many tests fail because of the new structure. This would require many changes. It may be better to do that in a separate PR.
```diff
diff --git a/src/orange/orangeinp/CsgTreeIO.json.cc b/src/orange/orangeinp/CsgTreeIO.json.cc
index f355328c1..f3ec75bf1 100644
--- a/src/orange/orangeinp/CsgTreeIO.json.cc
+++ b/src/orange/orangeinp/CsgTreeIO.json.cc
@@ -65,11 +65,21 @@ void to_json(json& j, CsgTree const& tree)
 {
     ContainerVisitor<CsgTree const&, CsgTree::NodeId> visit_tree{tree};
 
-    j = json::array();
+    j = json::object();
+
+    json nodes = json::array();
     for (auto n : range(CsgTree::NodeId{tree.size()}))
     {
-        j.push_back(visit_tree(JsonConverter{}, n));
+        nodes.push_back(visit_tree(JsonConverter{}, n));
+    }
+    j["nodes"] = nodes;
+
+    json volumes = json::array();
+    for (auto id : tree.volumes())
+    {
+        volumes.push_back(id.get());
     }
+    j["volumes"] = volumes;
 }
 
 namespace detail
@@ -128,18 +138,19 @@ void to_json(nlohmann::json& j, CsgUnit const& unit)
         return result;
     }();
 
-    j["volumes"] = [&unit] {
-        CELER_ASSERT(unit.tree.volumes().size() == unit.fills.size());
-        json result = json::array();
-        for (auto i : range(unit.tree.volumes().size()))
+    j["fills"] = [&volumes = unit.tree.volumes(), &fills = unit.fills] {
+
+        CELER_ASSERT(volumes.size() == fills.size());
+        json result = json::object();
+        for (auto const& i : range(fills.size()))
         {
-            auto entry
-                = nlohmann::json{{"csg_node", unit.tree.volumes()[i].get()}};
-            if (auto* m = std::get_if<GeoMaterialId>(&unit.fills[i]))
+
+            auto entry = json::object();
+            if (auto* m = std::get_if<GeoMaterialId>(&fills[i]))
             {
                 entry["material"] = m->unchecked_get();
             }
-            else if (auto* d = std::get_if<Daughter>(&unit.fills[i]))
+            else if (auto* d = std::get_if<Daughter>(&fills[i]))
             {
                 entry["universe"] = d->universe_id.unchecked_get();
                 entry["transform"] = d->transform_id.unchecked_get();
@@ -149,15 +160,6 @@ void to_json(nlohmann::json& j, CsgUnit const& unit)
         return result;
     }();
 
-    j["fills"] = [&volumes = unit.tree.volumes()] {
-        json result = json::array();
-        for (auto const& v : volumes)
-        {
-            result.push_back(v.unchecked_get());
-        }
-        return result;
-    }();
-
     if (unit.background)
     {
         j["background"] = unit.background.get();
```

esseivaju:
This is the most complicated test problem, the ATLAS tilecal cylinder segment. Roots 16 and 27 are preserved so we have to duplicate many nodes.

Before simplification ([mermaidchart](https://www.mermaidchart.com/app/projects/98e19971-82da-477e-ab3b-dac122f9f53f/diagrams/9b72871c-2586-445a-bf7d-c696326fd8ac/version/v0.1/edit)):

```mermaid
flowchart TB
  n00["00:t"]
  n01["01:~"]
  n01 --> n00
  n02["02:S0"]
  n03["03:S1"]
  n04["04:~"]
  n04 --> n03
  n05["05:S2"]
  n06["06:~"]
  n06 --> n05
  n07["07:&"]
  n07 --> n02
  n07 --> n04
  n07 --> n06
  n08["08:S3"]
  n09["09:~"]
  n09 --> n08
  n10["10:&"]
  n10 --> n02
  n10 --> n04
  n10 --> n09
  n11["11:~"]
  n11 --> n10
  n12["12:S4"]
  n13["13:S5"]
  n14["14:&"]
  n14 --> n12
  n14 --> n13
  n15["15:&"]
  n15 --> n07
  n15 --> n11
  n15 --> n14
  n16["16:~"]
  n16 --> n15
  n17["17:S6"]
  n18["18:S7"]
  n19["19:~"]
  n19 --> n18
  n20["20:&"]
  n20 --> n06
  n20 --> n17
  n20 --> n19
  n21["21:&"]
  n21 --> n09
  n21 --> n17
  n21 --> n19
  n22["22:~"]
  n22 --> n21
  n23["23:S8"]
  n24["24:S9"]
  n25["25:&"]
  n25 --> n23
  n25 --> n24
  n26["26:&"]
  n26 --> n20
  n26 --> n22
  n26 --> n25
  n27["27:~"]
  n27 --> n26
  n28["28:&"]
  n28 --> n15
  n28 --> n27
```
after simplification ([mermaidchart](https://www.mermaidchart.com/app/projects/98e19971-82da-477e-ab3b-dac122f9f53f/diagrams/3cbc8f5c-863a-4fe5-bb27-3923c29f6b90/version/v0.1/edit)):

```mermaid
flowchart TB
  n00["00:t"]
  n01["01:~"]
  n01 --> n00
  n02["02:S0"]
  n03["03:~"]
  n03 --> n02
  n04["04:S1"]
  n05["05:~"]
  n05 --> n04
  n06["06:S2"]
  n07["07:~"]
  n07 --> n06
  n08["08:|"]
  n08 --> n03
  n08 --> n04
  n08 --> n06
  n09["09:&"]
  n09 --> n02
  n09 --> n05
  n09 --> n07
  n10["10:S3"]
  n11["11:~"]
  n11 --> n10
  n12["12:|"]
  n12 --> n03
  n12 --> n04
  n12 --> n10
  n13["13:&"]
  n13 --> n02
  n13 --> n05
  n13 --> n11
  n14["14:S4"]
  n15["15:~"]
  n15 --> n14
  n16["16:S5"]
  n17["17:~"]
  n17 --> n16
  n18["18:|"]
  n18 --> n15
  n18 --> n17
  n19["19:&"]
  n19 --> n14
  n19 --> n16
  n20["20:|"]
  n20 --> n08
  n20 --> n13
  n20 --> n18
  n21["21:&"]
  n21 --> n09
  n21 --> n12
  n21 --> n19
  n22["22:S6"]
  n23["23:~"]
  n23 --> n22
  n24["24:S7"]
  n25["25:~"]
  n25 --> n24
  n26["26:|"]
  n26 --> n06
  n26 --> n23
  n26 --> n24
  n27["27:|"]
  n27 --> n10
  n27 --> n23
  n27 --> n24
  n28["28:&"]
  n28 --> n11
  n28 --> n22
  n28 --> n25
  n29["29:S8"]
  n30["30:~"]
  n30 --> n29
  n31["31:S9"]
  n32["32:~"]
  n32 --> n31
  n33["33:|"]
  n33 --> n30
  n33 --> n32
  n34["34:|"]
  n34 --> n26
  n34 --> n28
  n34 --> n33
  n35["35:&"]
  n35 --> n21
  n35 --> n34
```

esseivaju:
And a smaller test case with volumes: 

Before:
```mermaid
flowchart TB
  n00["00:t"]
  n01["01:~"]
  n01 --> n00
  n02["02:S0"]
  n03["03:S1"]
  n04["04:~"]
  n04 --> n02
  n05["05:~"]
  n05 --> n03
  n06["06:|"]
  n06 --> n04
  n06 --> n05
  n07["07:~"]
  n07 --> n06
  n08["08:S2"]
  n09["09:~"]
  n09 --> n08
  n10["10:&"]
  n10 --> n07
  n10 --> n09
  subgraph Volumes
    v0["vol0"]
    v1["vol1"]
    v2["vol2"]
  end
  v0 --> n06
  v1 --> n10
  v2 --> n07
```

After simplification:

```mermaid
flowchart TB
  n00["00:t"]
  n01["01:~"]
  n01 --> n00
  n02["02:S0"]
  n03["03:S1"]
  n04["04:~"]
  n04 --> n02
  n05["05:~"]
  n05 --> n03
  n06["06:&"]
  n06 --> n02
  n06 --> n03
  n07["07:|"]
  n07 --> n04
  n07 --> n05
  n08["08:S2"]
  n09["09:~"]
  n09 --> n08
  n10["10:&"]
  n10 --> n06
  n10 --> n09
  subgraph Volumes
    v0["vol0"]
    v1["vol1"]
    v2["vol2"]
  end
  v0 --> n07
  v1 --> n10
  v2 --> n06
```
Without volumes, this tree simplifies to:
```mermaid
flowchart TB
  n00["00:t"]
  n01["01:~"]
  n01 --> n00
  n02["02:S0"]
  n03["03:S1"]
  n04["04:&"]
  n04 --> n02
  n04 --> n03
  n05["05:S2"]
  n06["06:~"]
  n06 --> n05
  n07["07:&"]
  n07 --> n04
  n07 --> n06
```

sethrj:
One thing I notice (not to do with your work or in scope for this PR) is that we have `join(and, [join(and, [..]), ...])` which should probably be "canonicalized" into a single `join`. I wonder if that can easily be implemented by `NodeSimplifier`...

(EDIT: Done in #1784)

esseivaju:
> I wonder if that can easily be implemented by NodeSimplifier...

If you know that there are no other parents (volume) for that nested join, then I suppose that the NodeSimplifier should be able to do that.

esseivaju:
@sethrj I reworked the code to remove `MatchingNodes::modified` and use multiple variables instead, being more explicit about which value is needed in a given situation. I think this makes the algorithm clearer.

esseivaju:
@sethrj I'm finally happy with the implementation so it should be good for review.

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -0,0 +1,143 @@
...
+
+    //! Helper struct to translate ids from the original tree to ids in the
+    //! simplified tree
+    struct MatchingNodes
+    {
+        //! Set if a node has the exact same node in the simplified tree
+        NodeId unmodified;
+
+        //! Indirections to new simplified join following DeMorgan's law
+        //! Set iif the original node is a negated node.
```
sethrj: Is this `iff` ("if and only if") or a typo?

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -0,0 +1,143 @@
...
+        explicit operator bool() const noexcept
+        {
+            return unmodified || simplified_to || opposite_join || new_negation;
+        }
+
+        // Lookup a node an equal node in the simplified tree
+        NodeId equivalent_node() const;
+    };
+
+    //! Rudimentary 2D square matrix view of a vector<bool>
```
sethrj, line 88: Should the matrix be square? I thought the number of nodes was likely to change. Also it means that the storage requirements (and cost?) scale quadratically with the number of CSG nodes... which could be unpleasant for some of the geometries. (A fair number have hundreds of volumes each of which has  a couple of faces.)

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -0,0 +1,143 @@
...
+    //! Set when we must insert a \c Negated parent for the given index
+    std::vector<bool> new_negated_nodes_;
+
+    //! Set when \c Joined nodes have a \c Negated parent, so we need to insert
+    //! an opposite join node with negated operands
+    std::vector<bool> negated_join_nodes_;
+
+    //! Parents matrix. For nodes n1, n2, if n1 * tree_.size() + n2 is set, it
+    //! means that n2 is a parent of n1
+    Matrix2D parents_;
```
sethrj, line 136: Yeah I really think this should be sparse...

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -0,0 +1,143 @@
...
+    CsgTree operator()();
+
+  private:
+    //! CsgTree node 0 is always True{} and can't be the parent of any node
+    //! so reuse that bit to tell that a given node is a volume
+    static constexpr auto is_volume_index{NodeId{0}};
+    //! CsgTree node 1 is always a Negated node parent of node 0, so we can
+    //! reuse that bit to tell if a node has a parent as it's never set for
+    //! node id >= 2
+    static constexpr auto has_parents_index{NodeId{1}};
```
sethrj: Even though these are constexpr, since they act like private member data should we give them underscores as well? When I saw the constant used I wasn't sure if it was a local variable, global constant, etc.

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,449 @@
...
+
+                // Check if the negation is a root or a volume. If so, we must
+                // insert it in the simplified tree
+                if (parents_[{node_id, is_volume_index}]
+                    || !parents_[{node_id, has_parents_index}])
+                {
+                    return true;
+                }
+
+                //
```
sethrj:
```suggestion
```

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -0,0 +1,449 @@
...
+                        operands.push_back([&] {
+                            auto& trans = node_ids_translation_[n.get()];
+                            CELER_EXPECT(trans.new_negation
+                                         || trans.opposite_join);
+                            if (trans.new_negation)
+                                return trans.new_negation;
+                            return trans.opposite_join;
+                        }());
+                    }
+                }
```
sethrj: This looks like a good candidate for another helper member function
```c++
std::vector<NodeId> DeMorganSimplifier::build_negated_nodes(std::vector<NodeId> const& nodes) const
```

sethrj (approved):
Great refactor @esseivaju , sorry it took me two weeks to review! We can postpone the "2D matrix" reconsideration to a follow-on PR since that's an optimization. I think we should (for testing) activate the simplifier as part of the UnitProto construction and compare before/after build times and memory usage for our suite of benchmark problems (including the super big ugly nasty ones).

src/orange/orangeinp/detail/DeMorganSimplifier.hh
esseivaju: typo

src/orange/orangeinp/detail/DeMorganSimplifier.hh
esseivaju: This is only used to represent the parent relationship in the original `CsgTree`so the number of nodes never changes. Storage does scale quadratically but at least vector<bool> is very compact. Under a certain number of CsgNode per tree, a sparse matrix could easily use more space since you'd need to store row/col indices though it would still be better to use it. The quadratic memory scaling starts to become significant around 1000 CsgNode (100Mb for that matrix of parents)

Issue #687, opened 2023-03-15
Implement on-device error reporting


sethrj:
With https://github.com/celeritas-project/celeritas/pull/685 we have the first case of detecting and killing a particle due to an erroneous condition. We'll probably have other cases (surface crossings, convergence issues, ...) that might stop a particle before its time, and rather than doing something arbitrary like removing or depositing its energy, we should alert the user like Geant4 does. We might consider allowing a set of action IDs that are flagged as error conditions, and before the secondary creation step (where killed tracks are destroyed) we dump the state of those tracks to CPU and display them.

PR #1290, opened 2024-06-24
Log and kill geometry/propagation errors

sethrj:
This replaces the "kill looping tracks" with an explicit kernel for killing tracks, and replaces a host-only "validate" for tracks being inside the geometry and having a valid material ID. It logs descriptions and error messages for each of these cases.

A follow-on PR will add internal error states to the geometry tracking so we can catch issues with the current ORANGE implementation.

See https://github.com/celeritas-project/celeritas/issues/687 .


src/celeritas/geo/detail/GeoErrorAction.cu
```diff
@@ -0,0 +1,39 @@
...
+
+#include "GeoErrorExecutor.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Launch the boundary action on device.
```
amandalund, line 23:
```suggestion
 * Launch the geometry error action on device.
```

src/celeritas/geo/detail/GeoErrorAction.cc
```diff
@@ -0,0 +1,57 @@
...
+ * Construct with action ID.
+ */
+GeoErrorAction::GeoErrorAction(ActionId aid)
+    : ConcreteAction(aid, "geo-kill", "kill a track due to a navigation error")
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Launch the boundary action on host.
```
amandalund, line 36:
```suggestion
 * Launch the geometry error action on host.
```

src/celeritas/geo/detail/GeoErrorExecutor.hh
```diff
@@ -0,0 +1,80 @@
...
+        << repr(geo.dir()) << ": ";
+    if (!geo.is_outside())
+    {
+        msg << "depositing " << deposited << " [" << Energy::unit_type::label()
+            << "] in "
+            << "volume " << geo.volume_id().unchecked_get();
+    }
+    else
+    {
+        msg << "lost " << deposited << " energy";
```
amandalund, line 73:
```suggestion
        msg << "lost " << deposited << " " << Energy::unit_type::label() << " energy";
```

src/celeritas/geo/detail/BoundaryExecutor.hh
```diff
@@ -54,7 +58,16 @@ BoundaryExecutor::operator()(celeritas::CoreTrackView const& track)
         // Update the material in the new region
         auto geo_mat = track.make_geo_material_view();
         auto matid = geo_mat.material_id(geo.volume_id());
-        CELER_ASSERT(matid);
+        if (CELER_UNLIKELY(!matid))
+        {
+#if !CELER_DEVICE_COMPILE
+            CELER_LOG_LOCAL(error) << "Track entered a volume without an "
+                                      "associated material";
```
amandalund, line 65: Under what circumstances might this occur?

src/celeritas/global/alongstep/detail/PropagationApplier.hh
```diff
@@ -13,12 +13,14 @@
...
-#    define CELER_CHECK_POSITION 1
+#define CELER_CHECK_POSITION 0
+#if !CELER_DEVICE_COMPILE
 #    include "corecel/io/Logger.hh"
-#    include "corecel/io/Repr.hh"
-#else
-#    define CELER_CHECK_POSITION 0
+#    if CELERITAS_DEBUG
+#        undef CELER_CHECK_POSITION
+#        define CELER_CHECK_POSITION 0
```
amandalund, line 21: Should this be defined in this case?
```suggestion
#        define CELER_CHECK_POSITION 1
```

src/celeritas/track/SimTrackView.hh
```diff
@@ -299,10 +299,13 @@ CELER_FUNCTION void SimTrackView::along_step_action(ActionId action)
...
 /*!
  * Set whether the track is active, dying, or inactive.
+ *
+ * For efficiency we require that the track status isn't changed redundantly,
+ * except for "killing" .
  */
 CELER_FUNCTION void SimTrackView::status(TrackStatus status)
 {
-    CELER_EXPECT(status != this->status());
+    CELER_EXPECT(status != this->status() || status == TrackStatus::killed);
```
amandalund, line 308: When might we set the status to `killed` more than once?

amandalund (commented):
Thanks @sethrj! I think this looks good overall. I'm not sure I would classify looping tracks as a "geometry/propagation error", except in the (hopefully rare) case a track that is actually stuck is marked as looping. I think it's really more of a tracking cut: there's nothing wrong with their behavior and we could continue transporting them to completion, but we choose to kill them because they make progress so slowly and the computational cost of transporting them would be too high. But anyway, nomenclature aside the logic for killing looping tracks or tracks with geometry errors should be the same.

src/celeritas/geo/detail/GeoErrorExecutor.hh
sethrj: Hmmm apparently I forgot to push!

src/celeritas/geo/detail/BoundaryExecutor.hh
sethrj: Our `GeoMaterial` input takes a map of volume name/material IDs and fills the rest with invalid IDs. Those IDs can be legitimate if the volume isn't reachable by the tracking routine (e.g., the `[EXTERIOR]` volume, or other "imaginary" volumes defined for convenience by vecgeom/g4). However if there's an error in the input or importing or something, you can end up with undefined materials...

src/celeritas/global/alongstep/detail/PropagationApplier.hh
sethrj: 🤦‍♂️

src/celeritas/track/SimTrackView.hh
sethrj: It happens if the track is killed during initialization; we can't leave it as "alive" if it has an undefined volume/material so I have to kill it there. This is kind of messy.

sethrj:
This was a bit of a half-baked idea, you're perfectly right about the looping track not being the same as a geometry error. Maybe we should just make this a "tracking cut" action? Should we print log messages immediately when the error occurs?

It also exposes some of the fragility in our implementation... there's still a test failure due to a hardcoded condition not being met...

src/celeritas/geo/detail/BoundaryExecutor.hh
amandalund: Thanks, makes sense. In what cases do you think we should be asserting/validating vs. logging an error and killing on the CPU/silently killing the track on the GPU?

src/celeritas/track/SimTrackView.hh
amandalund: Ok, so if it's killed during initialization and then killed again sometime later (I guess in either the interaction or eloss applier)?

amandalund:
Yeah, a "tracking cut" or more generic kind of "track killer" action might make more sense, same with printing the whole error message immediately.

sethrj:
OK I think I'm going to rework this: maybe a combination of
- a tracking cut action, plus
- a helper function that can choose between logging a message and throwing an exception on host

sethrj:
I'm going to close this in favor of a fresh PR.

PR #1291, opened 2024-06-26
Add Bethe-Bloch muon ionization model

amandalund:
This adds the high energy  muon ionization model (valid for incident mu-/mu+ above 200 keV).


src/celeritas/em/data/MuIonizationData.hh
```diff
@@ -0,0 +1,34 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * IDs used in muon ionization.
+ */
+struct MuIonizationIds
+{
+    ActionId action;
```
sethrj: It's been a long time since we needed to filter on action *inside* the model actions: nowadays that's handled by the action launcher. Perhaps this could all be inlined into the `MuIonizationData`?

src/celeritas/em/data/MuBetheBlochData.hh
```diff
@@ -0,0 +1,38 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Device data for creating an interactor.
+ */
+struct MuBetheBlochData
+{
+    //! Model/particle IDs
```
whokion:
```suggestion
    //! Model ID
```

src/celeritas/em/data/MuIonizationData.hh
whokion:
```suggestion
    //! Action/particle IDs
    ActionId action;
```
Just for consistency.

src/celeritas/em/data/MuIonizationData.hh
```diff
@@ -0,0 +1,34 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/em/data/MuIonizationData.hh
```
whokion: Is this data separated from MuBetheBlockData as it will be commonly used for low energy models too?  An alternative is to combine them anyway as mu- and mu+ have different low energy models anyway (required particle Ids will be different, so it will not be common for all models). Not a big deal, but just for a second thought.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+#include "detail/PhysicsConstants.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Perform the discrete part of the muon ionization process.
+ *
+ * This simulates the production of delta rays by incident mu- and mu+ with
+ * energies above 200 keV.
```
whokion, line 36: Probably this hard number may be a parameter.  The model validity (>100 keV?) and an actual low energy limit used (200 keV) may be different and can be settable - okay, this is detail and can be done later.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * \note This performs the same sampling routine as in Geant4's
+ * G4MuBetheBlochModel.
```
whokion: Please add PM Section 11.1 here too.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    real_type max_secondary_energy_;
+    // Whether to apply the radiative correction
+    bool use_rad_corr_;
+    // Envelope distribution for rejection sampling
+    real_type envelope_dist_;
+
+    //// CONSTANTS ////
+
+    //! Incident energy above which the radiative correction is applied [MeV]
+    static CELER_CONSTEXPR_FUNCTION Energy inc_energy_limit()
```
whokion:
```suggestion
    static CELER_CONSTEXPR_FUNCTION Energy rad_correction_limit()
```
or something similar as the current name is somewhat misleading (?, at least, to me)

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    {
+        return Energy{250};  //!< 250 MeV
+    }
+
+    //! Secondary energy above which the radiative correction is applied [MeV]
+    static CELER_CONSTEXPR_FUNCTION Energy secondary_energy_limit()
```
whokion: This name is also confusing with `secondary_cutoff_`.  Technically this is the low limit of the energy transfer from muon to the delta ray and radiative gammas. So, it may be close to the secondary electron kinetic energy, but still sounds like referring to a secondary production cutoff - the actual meaning is something like `lower_limit_of_radiative_correction_integral` (yes, it is too long^^).  Again, it is a minor detail, but for the better readability.

src/celeritas/em/interactor/detail/PhysicsConstants.hh
```diff
@@ -64,6 +64,12 @@ CELER_CONSTEXPR_FUNCTION units::MevEnergy seltzer_berger_limit()
     return units::MevEnergy{1e3};  //! 1 GeV
 }
 
+//! Energy threshold between muon ionization models. TODO: make configurable
+CELER_CONSTEXPR_FUNCTION units::MevEnergy mu_ionization_limit()
```
whokion: Okay, this addresses my earlier comment. But, the name is somewhat misleading again.  This 200 keV is the low limit of the BetheBlock model, not the limit of muon ionization. As lower models may be added later (?), may mu_bethe_block_limit or something similar be better?

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    // \c G4EmParameters::UseAngularGeneratorForIonisation())
+    real_type secondary_momentum = std::sqrt(
+        secondary_energy * (secondary_energy + 2 * electron_mass_));
+    real_type total_momentum = total_energy_ * std::sqrt(beta_sq_);
+    real_type costheta = secondary_energy * (total_energy_ + electron_mass_)
+                         / (secondary_momentum * total_momentum);
+    CELER_ASSERT(costheta <= 1);
+
+    // Sample and save outgoing secondary data
+    UniformRealDistribution<real_type> sample_phi(0, 2 * constants::pi);
```
whokion, line 212: This may go to the constructor.

src/celeritas/em/model/MuBetheBlochModel.cc
```diff
@@ -0,0 +1,111 @@
...
+    data_.ids.electron = particles.find(pdg::electron());
+    data_.ids.mu_minus = particles.find(pdg::mu_minus());
+    data_.ids.mu_plus = particles.find(pdg::mu_plus());
+
+    CELER_VALIDATE(
+        data_.ids.electron && data_.ids.mu_minus && data_.ids.mu_plus,
+        << "missing electron and/or muon particles (required for "
+        << this->description() << ")");
+
+    data_.electron_mass = particles.get(data_.ids.electron).mass();
```
whokion: Can we do the same way for muon mass (mass_) instead of setting it from particle.mass() in the constructor or is the latter more safer?

src/celeritas/em/model/MuBetheBlochModel.cc
```diff
@@ -0,0 +1,111 @@
...
+{
+    Applicability mu_minus_applic, mu_plus_applic;
+
+    mu_minus_applic.particle = data_.ids.mu_minus;
+    mu_minus_applic.lower = detail::mu_ionization_limit();
+    mu_minus_applic.upper = detail::high_energy_limit();
+
+    mu_plus_applic.particle = data_.ids.mu_plus;
+    mu_plus_applic.lower = detail::mu_ionization_limit();
+    mu_plus_applic.upper = detail::high_energy_limit();
```
whokion:
```suggestion
    mu_plus_applic = mu_minus_applic;
    mu_plus_applic.particle = data_.ids.mu_plus;
```

whokion (approved):
Good addition!  All comments are minor, but for little bit more clarify. Thanks for the nice work!

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    bool use_rad_corr_;
+    // Envelope distribution for rejection sampling
+    real_type envelope_dist_;
+
+    //// CONSTANTS ////
+
+    //! Incident energy above which the radiative correction is applied [MeV]
+    static CELER_CONSTEXPR_FUNCTION Energy inc_energy_limit()
+    {
+        return Energy{250};  //!< 250 MeV
```
sethrj:
```suggestion
        return units::MevEnergy{250};
```
since it's a hardcoded constant rather than an "internal representation of energy" (which is what the `Energy` typedef signifies)

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    real_type secondary_energy;
+    real_type target_dist;
+    do
+    {
+        real_type u = generate_canonical(rng);
+        secondary_energy
+            = electron_cutoff_ * max_secondary_energy_
+              / (electron_cutoff_ * (1 - u) + max_secondary_energy_ * u);
```
sethrj: Perhaps
```suggestion
        secondary_energy
            = electron_cutoff_ * max_secondary_energy_
              / UniformDistribution{electron_cutoff_, max_secondary_energy_}(rng);
```
?

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
sethrj: Even better: since we repeat this line and the next in so many places, we really need a helper function for it. Another helper function for some of the energy conservation calculations would be good too...

sethrj (approved):
LGTM! Couple of minor comments.

src/celeritas/em/data/MuIonizationData.hh
amandalund: Good point. I can submit a separate small PR to move the action IDs out of the model data.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
...
+    // Sample and save outgoing secondary data
+    UniformRealDistribution<real_type> sample_phi(0, 2 * constants::pi);
+    secondary->direction
+        = rotate(from_spherical(costheta, sample_phi(rng)), inc_direction_);
+    secondary->energy = Energy{secondary_energy};
+    secondary->particle_id = shared_.ids.electron;
+
+    Interaction result;
+    result.energy = Energy{inc_energy_ - secondary_energy};
+    for (int i = 0; i < 3; ++i)
```
stognini, line 220:
```suggestion
    for (auto i : range(3))
```

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -0,0 +1,257 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/em/interactor/MuBetheBlochInteractor.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <cmath>
```
stognini, line 10: This may be a silly question, but do we have a reason to use `cmath` vs. `math.h`?

Edit: nevermind, `math.h` is the deprecated math header, not the opposite, which is what led to my comment.

stognini (approved):

stognini:
Looks great! Thanks @amandalund !

src/celeritas/em/data/MuIonizationData.hh
amandalund: Yeah, that was my thinking, but I'd forgotten about the separate mu-/mu+models. I'll get rid of this and just put the IDs in the data.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
amandalund: Nice!

src/celeritas/em/model/MuBetheBlochModel.cc
amandalund: I thought we only stored a particle mass in the data when we weren't able to get it from the incident particle (e.g. if we needed electron mass in a gamma model), but I see now there are several exceptions. It's not _necessaary_ to store it in the model data, but is there a benefit to doing that vs. getting it directly from the particle?

src/celeritas/em/interactor/detail/PhysicsConstants.hh
amandalund: I initially thought `mu_bethe_block_limit` too, but worried since it wasn't consistent with our naming for the limit between the bremsstrahlung models it might be confusing (`mu_bethe_block_limit` would be the lower limit for the Bethe-Bloch model, but `seltzer_berger_limit` is the upper limit for SB). I agree `mu_ionization_limit` is misleading though... maybe we could rename `seltzer_berger_limit` to `rel_brems_limit`, or add `upper` and `lower` to the names or something?

src/celeritas/em/model/MuBetheBlochModel.cc
sethrj: I guess it's maybe slightly less indirection but more code and more work... let's keep to using the incident `Particle` insofar as possible.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
sethrj: yeah this is one that we've used inconsistently across the various interactions... a helper function for this would be better.

src/celeritas/em/model/MuBetheBlochModel.cc
whokion: I do not have a con or pro for this as they are essentially same.  Getting it directly from the particle input is assuming that the particle is, in this, either mu+ or mu- of which id will be assured later in the body of the constructor (which is of course self-satisfactory each other).  Agree to set a consistent rule for this if @sethrj has a final thought.

src/celeritas/em/interactor/detail/PhysicsConstants.hh
whokion: Adding low/upper definitely clarifies what the limit is referring to (of course, with a trade off of using the longer name).

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
amandalund: Geant4 calls it `limitKinEnergy`, so maybe `kin_energy_limit`? Also not great, but it would at least make it easier to compare to the Geant4 model, and it's short, and I don't have any better ideas. I'm open to any other suggestions, and will add your comment for clarity.

whokion: `kin_energy_limit` is fine to me. You may add a short description to clarify what it is, something like,
"the lower limit of the energy transfer from the muon to the secondary electron and radiative gammas" - okay,   let's not pull our hair out^^.

amandalund:
Thanks all for the reviews and good suggestions! I think I've addressed everything except the helper functions for the angle and direction calculations, which would probably be better as a separate PR (I can do that now or as a follow up).

sethrj:
Follow up sounds great!

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
amandalund: Actually, what did you have in mind for this helper function? Since it's just two lines of code I'm not sure what would help...

PR #1293, opened 2024-06-27
Store model action IDs in model class instead of in data

amandalund:
As @sethrj pointed out, the only place we currently use the action IDs for the physics models is in the model class itself (where it's returned). This removes the action IDs from the model device data and stores them directly in the model class instead.


sethrj (approved):
Excellent! We should give bonuses for net removal of code 😄

PR #1294, opened 2024-06-27
Add new track status and support user "initialization"

sethrj:
This defines an "error" status (to supporting a refactor of #1290) and an "initializing" status, which is set whenever a track slot starts its life. A new `user_start` action order allows users to set custom data that applies only to new tracks. I split the `Stepper` test into two, one of which requires Geant4 and the other which uses built-in data.

Changes:
- Rename `post_post` to `user_post`, and `post_pre` to `user_pre`
- "Not inactive" and "alive" are now different
- Slightly changed `TrackSortUtils`

Do we also want to add a `Stepper` output for "errored" tracks? 🤔


amandalund:
I think this is a good idea! Before I look too closely, are the stepper tests a straight up refactoring, or did you make changes?

sethrj:
The stepper tests basically got moved. There's one or two new tests in the Stepper, and almost everything got directly moved to StepperGeant which you can ignore

amandalund (approved):
Looks good to me, thanks @sethrj!

PR #1295, opened 2024-06-28
Add involute surface to support HFIR simulation

VHLM2001:
Adds involute surface to Orange, with additional solver and crests.

Involutes are curves are created by unwinding a chord from a parent shape. 

Involute

The involute implemented here is constructed from a circle and can be made to be clockwise (negative) or anti-clockwise (positive). 

Signed_Involutes

This involute is the same type of that found in HFIR, and is necessary for building accurate models.

Involutes can be represented by the following equation:

$$\vec{r} = r_b\begin{bmatrix}\cos\left(t+a\right)+t\sin\left(t+a\right) \\\ \sin\left(t+a\right)-t\cos\left(t+a\right) \\\ 0 \end{bmatrix}$$

Where $`\vec{r}`$ is the parametric form of the involute, $`r_b`$ is the radius of the circle of involute,  $`a`$ is the angular displacement of the involute from the x-axis, and $`t`$ is the angle from the displacement angle on describing the tangent that is used to construct the involute.

To determine the $`a`$ for this involute the tangent used for the construction of the involute for the point must be obtained. To obtain points instruction for the tangents that cross through that point, using the circle which the involute is constructed from and a second circle constructed with the radius being the distance from the origin to the point center point is needed, and shown bellow.

While the parameters define the curve itself have no restriction, except for the radius of involute being positive, the involute needs to be bounded for there to be finite solutions to the roots of the intersection points. Additionally this bound cannot exceed an interval of size of 2$`\pi`$ to be able to determine if whether a particle is in, out, or on the involute surface. Lastly the involute geometry is fixed to be z-aligned. 

To determine the sense of a point, first it is determined if it is within the radial conversion of the bounds given by:

$$r=\sqrt{t^2+1}$$

Then it is checked if the particle is on the surface. After which it is possible to check if a particle is inside the region by finding if there exists an $`a`$ in the interval of $`[a_0, a_0+t_{max}]`$, where $`a_0`$ is the displacement angle for the reference involute and $`t_{max}`$ is the upper bound, for which an involute that passes through the point can be constructed. 

This tangent can be obtained with circle of involute and constructing a second circle centered at the midpoint of the origin and the point with a radius equal to half the point's distance from the origin shown bellow. 

tangent_construction

Then to find the $`x\prime`$ coordinate of the intersection of the intersection the following relation is used:

$$ x\prime = \frac{d^2-r^2+r_b^2}{2d} $$

Since $`r=d`$ the relation is simplified to:

$$ x\prime = \frac{r_b^2}{2r} $$

Then the $`y\prime`$ coordinate can be obtained from (positive for positive involutes and negative for negative involutes):

$$ y\prime = \pm \sqrt{r_b^2-{x\prime}^2} $$

Then the coordinates can be transformed to the $`(x,y)`$ by rotating the coordinates by the angle between the $`x\prime`$ -axis and the $`x`$-axis.

sense

To calculate the intersections a root function of a line and involute is obtained bellow:

$$f(t) = r_b \left( v\cos\left(t+a\right) - u\sin\left(t+a\right) + t\left[\sin\left(t+a\right) + u\cos\left(t+a\right) \right] \right)-xu+yu$$

Where $`u`$ is the $`x`$ component of the direction of the particle, $`v`$ is the $`y`$ component of the direction of the particle, $`x`$ is the $`x`$ coordinate of the particle, and $`y`$ is the $`y`$ coordinate of the particle. Since this equation has no analytical solution an iteration scheme must be used. The method applied here is the Regula False Iteration due to its stability and described by:

$$ c_k= \frac{a_kf(b_k)-b_kf(a_k)}{f(b_k)-f(a_k) }$$

$`c_k`$ is used to update the bounds and find the root throughout the iteration, while initial values of $`a_k`$ and $`b_k`$ are the bounds the upper and lower bounds of the roots give by:

$$ \text{for } f(t)=0; t \in \\{0, \beta - a, \beta - a - \pi, \beta - a + \pi, \beta - a -2\pi, \beta -a + 2\pi, ...\\}$$

Where,

$$ \beta = \arctan\left(\frac{-v}{u}\right) $$

distance

Lastly to calculate the outward normal of a point on the involute the following equation can be used:

$$ \hat{n} =  s \begin{bmatrix} \sin\left(t+a\right) \\\ -\cos\left(t+a\right) \\\ 0 \end{bmatrix} $$

Where $`s`$ is the sign of the involute. 

outward_normal


sethrj:
Thanks for the contribution @VHLM2001 ! I'll be up at ORNL today and we can do a very rough first pass overview. In the meantime, can you please update the description to include:
- What is an involute and why does it matter than we can track on it?
- Are there any assumptions or restrictions about the surface?
- What methods are used to perform the intersection?

sethrj:
@VHLM2001 can you please:
- Run `git-clang-format $(git merge-base HEAD upstream/develop)` and commit the result
- Run `scripts/dev/install-commit-hooks.sh` to install an auto-formatter for future commits
- Build with `-Werror -Wall -Wextra -pedantic`
- Merge `upstream/develop`

Thanks!

src/orange/surf/Involute.hh
```diff
@@ -126,15 +132,18 @@ class Involute
...
+CELER_FUNCTION Involute::Involute(Real3 const& origin,
+                                  real_type radius,
+                                  real_type a,
+                                  real_type sign,
+                                  real_type tmin,
+                                  real_type tmax)
+    : origin_(origin), r_b_(radius), a_(a), sign_(sign), tmin_(tmin), tmax_(tmax)
 {
-    const double pi = 3.14159265358979323846;
+    double const pi = 3.14159265358979323846;
```
sethrj: Remove these in favor of the `pi` defined in `corecel/math/Constants.hh`.

src/orange/surf/Involute.hh
```diff
@@ -163,37 +176,38 @@ CELER_FUNCTION SignedSense Involute::calc_sense(Real3 const& pos) const
...
+        return SignedSense::on;
+    }
 
     /*
-     * Check if point is in defined bounds. 
+     * Check if point is in defined bounds.
      */
-    if (abs(tPoint2) < ipow<2>(tmin_) - tol) {
-       return SignedSense::outside;
+    if (abs(tPoint2) < ipow<2>(tmin_) - tol)
```
sethrj: We *really really* should not have tolerances when evaluating senses and distance to boundary. When crossing any surface, involute or not, the sense is updated analytically (i.e we flip a bit and *don't* recalculate senses near a surface).  You should never need a tolerance because we will always know whether we are inside, outside, or on.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+    ya = sqrt(ipow<2>(r_b_) - ipow<2>(xa));
+    yb = -sqrt(ipow<2>(r_b_) - ipow<2>(xa));
+    yc = sqrt(ipow<2>(r_b_) - ipow<2>(xb));
+    yd = -sqrt(ipow<2>(r_b_) - ipow<2>(xb));
+
+    b = -2 * ipow<2>(r_b_) * y;
+    c = ipow<4>(r_b_) - ipow<2>(x) * ipow<2>(r_b_);
+
+    yalpha = (-b + sqrt(ipow<2>(b) - 4 * a * c)) / (2 * a);
+    ybeta = (-b - sqrt(ipow<2>(b) - 4 * a * c)) / (2 * a);
```
sethrj: Is there a way to sensibly use the `detail::QuadraticSolver` for this calculation?

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+
+    // Calculate angle of tangent
+    real_type theta = std::acos(
+        point[0] / sqrt(point[0] * point[0] + point[1] * point[1]));
+    if (point[1] < 0)
+    {
+        theta = (pi - theta) + pi;
+    }
+    real_type a1 = theta - tPoint;
+    if (abs(theta) < abs(tmax_ + a_) + tol && a1 >= a_ - tol)
```
sethrj: Always use `std::fabs` instead of `abs`, since the latter has a flaw in C/C++ where including the wrong header files may result in accidentally truncating to `int` before taking the absolute value.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+        point = point1;
+    }
+    else
+    {
+        point = point2;
+    }
+
+    // Calculate angle of tangent
+    real_type theta = std::acos(
+        point[0] / sqrt(point[0] * point[0] + point[1] * point[1]));
```
sethrj: Use `std::` namespace for `sqrt` and all other std functions. If you find yourself using a lot of them in a single function, you can add `using std::sqrt` to the top of the function body.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,316 @@
...
+                                         real_type tmin_,
+                                         real_type tmax_);
+
+    // Solve fully general case
+    inline CELER_FUNCTION Intersections operator()(real_type x,
+                                                   real_type y,
+                                                   real_type z,
+                                                   real_type u,
+                                                   real_type v,
+                                                   real_type w) const;
```
sethrj: Use `Real3 const& pos` and `Real3 const& dir` if possible; it makes the function less error-prone since it logically separates the two variables rather than flattening them together.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,316 @@
...
+    real_type yInv = r_b_ * (std::sin(angle) - tPoint * std::cos(angle));
+    if (abs(x - xInv) < tolPoint && abs(y - yInv) < tolPoint)
+    {
+        dist[j] = 0;
+        j++;
+    }
+
+    real_type beta;
+    // Line angle parameter
+    if (u != 0.0)
```
sethrj: Since we allow ORANGE/Celeritas to work with both double *and* single precision numbers, use `0` here to automatically promote the int to `real_type` rather than promote `u` to `double`.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+    else if (sign_ < 0)
+    {
+        tUpper = 0;
+        tLower = beta - a_ + 2 * pi;
+        while (tLower >= 0.0)
+        {
+            tLower -= pi;
+        }
```
sethrj: You may be able to use `std::fmod` for this?

src/orange/surf/Involute.cc
```diff
@@ -0,0 +1,32 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+Involute Involute::at0origin(Real3 const& origin,
+                             real_type radius,
+                             real_type a,
+                             real_type sign,
+                             real_type tmin,
+                             real_type tmax)
+{
```
sethrj, line 31: I don't understand how this helper function is any different from the other? Did you mean to take `real_type z` and fill with `{0,0,z}`? I don't think that's worth the extra code.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+
+    //! Get bounds of the involute
+    CELER_FUNCTION real_type tmin() const { return tmin_; }
+    CELER_FUNCTION real_type tmax() const { return tmax_; }
+
+    //! Get a view to the data for type-deleted storage
+    CELER_FUNCTION StorageSpan data() const { return {&origin_[0], 4}; }
+
+    // // Helper function to get the origin as a 3-vector
+    // Real3 calc_origin() const;
```
sethrj: Make sure to remove "dead" code.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+ with
+ * radius \em r_b from a starting angle of \em a (\f$r/h\f$ for a finite cone.
+ */
+class Involute
+{
+  public:
+    //@{
+    //! \name Type aliases
+    using Intersections = Array<real_type, 3>;
+    using StorageSpan = Span<real_type const, 4>;
```
sethrj: By my count, this class has 8 reals, not 4, so reconstructing this surface from type-deleted data will fail.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,347 @@
...
+    real_type r_b_;
+    real_type a_;
+    real_type sign_;
+
+    // Bounds
+    real_type tmin_;
+    real_type tmax_;
+
+    //! Private default constructor for manual construction
+    Involute() = default;
```
sethrj: I don't think you'll need this once you remove the static constructor. It's kind of a rare use case.

sethrj (changes requested):
A few initial comments to consider... the `tolerance` one will require some major changes that we can discuss with @tarapandya .

src/orange/surf/detail/InvoluteSolver.hh
sethrj: Correction: the first case you might be able to use `eumod` (defined in corecel/math/Algorithms). The second one may  require something else.

src/orange/surf/Involute.hh
VHLM2001: `detail::QuadraticSolver` only results in real and positive solutions to the quadratic equation. However, since it is possible the point to construct the tangent to contain negatives, this method is required.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,348 @@
...
+    real_type const y = pos[1] - origin_[1];
+    real_type const z = pos[2] - origin_[2];
+
+    real_type const u = dir[0];
+    real_type const v = dir[1];
+    real_type const w = dir[2];
+
+    detail::InvoluteSolver solve(r_b_, a_, sign_, tmin_, tmax_);
+
+    return solve(Real3{x, y, z}, Real3{u, v, w});
```
sethrj:
```suggestion
    detail::InvoluteSolver solve(r_b_, a_, sign_, tmin_, tmax_);
    
    // Expand translated positions into 'xyz' coordinate system
    Real3 rel_pos{pos};
    rel_pos -= origin_;
    return solve(rel_pos, dir);
```

tarapandya:
@VHLM2001 It would be good to include the equations for the involute in the description of the PR as well as a reference to the specific Regular Falsi iteration scheme you are using.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,268 @@
...
+        return SurfaceType::inv;
+    }
+
+    // Safety
+    static CELER_CONSTEXPR_FUNCTION bool simple_safety() { return false; }
+
+  public:
+    //// CONSTRUCTORS ////
+    // Construct at (0,0,z)
+    // Redundant method, required due to causing an error when excluded.
```
tarapandya: Remove this comment

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,268 @@
...
+    static Involute involute(Real3 const& origin,
```
tarapandya: You can make this explicit

src/orange/surf/Involute.hh
tarapandya: Agreed, once you change that constructor to be explicit you should be able to remove this default construction.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,268 @@
...
+/*!
+ * Determine the sense of the position relative to this surface.
+ */
+CELER_FUNCTION SignedSense Involute::calc_sense(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    /*
+     * Calculate distance to origin and obtain t value for disance.
```
tarapandya: disance -> distance in this comment

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,268 @@
...
+/*!
+ * Calculate outward normal at a position on the surface.
+ */
+CELER_FORCEINLINE_FUNCTION Real3 Involute::calc_normal(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    /*
+     * Calculate distance to origin and obtain t value for disance.
```
tarapandya: disance -> distance

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2021-2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/surf/detail/QuadraticSolver.hh
```
tarapandya: Update this comment: InvoluteSolver

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+{
+using constants::pi;
+//---------------------------------------------------------------------------//
+/*!
+ * Find positive, real, nonzero roots for involute intersection function.
+ *
+ * The involute intersection equation \f[
+   r_b * [v{cos(t+a)+tsin(t+a)} + u{sin(t+a)-tcos(t+a)}] + xv - yu = 0
+ * \f]
+ * has n solutions mathematically, but we only want solutions where t is
```
tarapandya: remove `is` in this comment line

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+ */
+class InvoluteSolver
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Intersections = Array<real_type, 3>;
+
+  public:
+    // Construct with
```
tarapandya: Finish this comment

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+ * \c Tolerance::sqrt_quadratic and the derivation in \c CylAligned ).
+ *
+ * \return An Intersections array where each item is a positive valid
+ * intersection or the sentinel result \c no_intersection() .
+ */
+class InvoluteSolver
+{
+  public:
+    //!@{
+    //! \name Type aliases
```
tarapandya, line 47: Should this `\name` be here?

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+                    talpha = tgamma;
+                    ftalpha = root(talpha, x, y, u, v, r_b_, a_);
+                }
+            }
+            t = tgamma;
+            angle = t + a_;
+            xInv = r_b_ * (std::cos(angle) + t * std::sin(angle));
+            yInv = r_b_ * (std::sin(angle) - t * std::cos(angle));
+
+            // Check if point is interval
```
tarapandya: if point is `in` interval

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    Array<real_type, 3> dist;
+    real_type j = 0;
+
+    real_type convert = std::sqrt(ipow<2>(v) + ipow<2>(u) + ipow<2>(w))
+                        / std::sqrt(ipow<2>(v) + ipow<2>(u));
+
+    /*
+     * Define tolerances.
```
tarapandya: You should say something about why you chose these tolerances

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+     */
+    real_type const tolPoint = 1e-7;
+    real_type const tolConv = 1e-8;
+
+    // 0 distance of particle on invoute surface
```
tarapandya: What do you mean by 0 here?

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+        real_type c = r_b_ * (a - b);
+        return c + x * v - y * u;
+    };
+    /*
+     * Results initalization
+     */
+    Intersections result;
+    result = {no_intersection(), no_intersection(), no_intersection()};
+
+    if (u == 0 && v == 0)
```
tarapandya, line 168: And this means?  Add a comment

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,310 @@
...
+    CELER_EXPECT(std::fabs(tmax) < 2 * pi + std::fabs(tmin));
+}
+
+//---------------------------------------------------------------------------//
+
+/*!
+ * Find all positive roots for involute surfaces that are within the bounds.
+ */
+CELER_FUNCTION auto
+InvoluteSolver::operator()(Real3 const& pos,
```
tarapandya, line 142: This method needs more descriptive comments and explanation of the iteration method

src/orange/surf/detail/SurfaceTransformer.cc
```diff
@@ -114,6 +115,20 @@ Sphere SurfaceTransformer::operator()(Sphere const& other) const
                                   other.radius_sq());
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Transform a Involute.
```
tarapandya: a -> an

tarapandya (commented):
@VHLM2001 I've reviewed the main code and provided some comments, will review the tests next

src/orange/surf/detail/InvoluteSolver.hh
sethrj: Yeah doxygen uses that to group

src/orange/surf/detail/InvoluteSolver.hh
sethrj: We shouldn't be using them at all 😅 and if we really need a solver tolerance there is a better place for it

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: Don't forget about this guy

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: And this guy

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: Also this guy

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: And this guy

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: And this guy

src/orange/surf/detail/InvoluteSolver.hh
tarapandya: I assume you are going to add more comments throughout this function as well so someone can follow the solver procedure?

VHLM2001:
Add involute images

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -119,7 +118,7 @@ InvoluteSolver::operator()(Real3 const& pos,
     real_type const v = dir[1];
     real_type const w = dir[2];
 
-    // Lambda used for calculating the roots
+    // Lambda used for calculating the roots using Regular Flasi Iteration
```
tarapandya: Flasi -> Falsi

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -145,9 +144,12 @@ InvoluteSolver::operator()(Real3 const& pos,
         return result;
     }
 
+    // Initialize distance vecctor with root counter j
```
tarapandya: vecctor -> vector

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -145,9 +144,12 @@ InvoluteSolver::operator()(Real3 const& pos,
...
     Array<real_type, 3> dist;
     real_type j = 0;
 
+    // Convertion constant for 2-D distance to 3-D distance
```
tarapandya: Convertion -> Conversion

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -161,7 +163,7 @@ InvoluteSolver::operator()(Real3 const& pos,
     real_type const tolPoint = 1e-7;
     real_type const tolConv = 1e-8;
 
-    // Particle on surface has distance of 0 to surface.
+    // Check if particle is on a surface within tolerence, and add a 0 distance
```
tarapandya: tolerence -> tolerance

tarapandya:
@VHLM2001 Thanks for these updates.  I looked through your tests.  Be sure to reference the python files you added in the test files so someone knows they exists and what they are for.  Also, looks like you have a lot of spelling typos in your test comments, please fix.

@sethrj Looks like this PR is ready for a more thorough review

sethrj:
@VHLM2001 thanks for posting those plots! Would it be possible to regenerate them by calling `ax.set_aspect("equal")`? (so [circles look like circles](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_aspect.html)). I'll start taking another look at this today. Thanks!

sethrj:
@VHLM2001 I've updated your branch through github by merging the latest upstream/develop into it. Please run `git pull --rebase` on your local machine. Thanks!

sethrj:
@VHLM2001 And also take a look at and fix the CI failures: https://github.com/celeritas-project/celeritas/actions/runs/9810411562/job/27090497812?pr=1295 (or the "actions" tab at the top of this page). Whenever you push, those will re-run, and it may take a couple of iterations until all the tests pass, since there are several combinations of configurations.

src/orange/surf/Involute.cc
```diff
@@ -0,0 +1,26 @@
...
+//---------------------------------------------------------------------------//
+#include "Involute.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+Involute::Involute(Real3 const& origin,
+                   real_type radius,
+                   real_type a,
+                   real_type sign,
```
sethrj: This would be better as an enum class bool: clockwise/anticlockwise

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+ * \f[
+ *   x = r_b * (cos(t+a) + tsin(t+a)) + x_0
+ *   y = r_b * (sin(t+a) - tcos(t+a)) + y_0
+ *  \f]
+ *
+ * where \em t is the normal angle of the tangent to the circle of involute
+ with
+ * radius \em r_b from a starting angle of \em a (\f$r/h\f$ for a finite cone.
+
+high level what it does
```
sethrj: Fix

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * To calculate the intercept of two circles it is assumed that the two circles
+ * are \em xp axis. The coordinate along \em xp originating from the center of
+ * one of the circles is given by: /f[
+ * xp = (d^2 - r^2 + R^2)/(2d)
+ * /f]
+ * Where \em d is the distance between the center of the two circles, \em r is
+ * the radius of the displaced circle, and \em R is the radius of the circle at
+ * the orgin. Then the point \em yp can be obatined from: \f[
+ * yp = +- sqrt(R^2 - xp^2)
```
sethrj:
```suggestion
 * y_p = \pm \sqrt(R^2 - x_p^2)
```

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+ * \f]
+ * Then to convert ( \em xp , \em yp ) to ( \em x, \em y ) the following
+ * roation is applied:/f[ x = xp*cos(theta) - yp*sin(theta) y = yp*cos(theta) +
```
sethrj: Typo and also backslashes and underscores

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from raw data.
+ */
+template<class R>
+CELER_FUNCTION Involute::Involute(Span<R, StorageSpan::extent> data)
+    : origin_{data[0], data[1], data[2]}
+    , r_b_{data[3]}
+    , a_{data[4]}
+    , sign_{data[5]}
```
sethrj: We could in theory encode the sign into a known-positive value (such as `r_b_`) and save a whole 8 bytes of storage... something like
```c++
orientation_ = static_cast<Orientation>(r_b_ < 0);
r_b_ = std::fabs(r_b_);
```

where
```c++
enum class Orientation : bool
{
    counterclockwise,
    clockwise
};
```
and then your "flip" boolean in the solver will be `static_cast<bool>(orientation_)`.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+ */
+CELER_FUNCTION SignedSense Involute::calc_sense(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    /*
+     * Calculate distance to origin and obtain t value for distance.
+     */
+    real_type const rxy2 = ipow<2>(x) + ipow<2>(y);
```
sethrj: Avoid `2` as part of a variable name since it's not clear whether it's "second" or "squared" or ....
```suggestion
    real_type const r_sq = ipow<2>(x) + ipow<2>(y);
```

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    real_type const t_point2 = (rxy2 / ipow<2>(r_b_)) - 1;
+    real_type const t_point = std::sqrt(t_point2);
```
sethrj: Before doing this, I'd add an assertion that `t_point2` >= 0, otherwise this will be NaN and the rest of the function will likely be garbage.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+
+    if (x == x_inv && y == y_inv)
+    {
+        return SignedSense::on;
+    }
+
+    /*
+     * Check if point is in defined bounds.
+     */
+    if (std::fabs(t_point2) < ipow<2>(tmin_))
```
sethrj: You shouldn't have to take `fabs` here or below since the square of a real value is always nonnegative

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    if (std::fabs(t_point2) > ipow<2>(tmax_))
+    {
+        return SignedSense::outside;
+    }
+
+    /*
+     * Check if point is in interval
+     */
+
+    // Describ
```
sethrj: ...

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+
+    // Calculate tangents
+    real_type r = std::sqrt(ipow<2>(x) + ipow<2>(y));
+    real_type xp = (ipow<2>(r_b_)) / (2 * r);
```
sethrj:
```suggestion
    real_type xp = ipow<2>(r_b_) / (2 * std::sqrt(r));
```

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    Array<real_type, 2> point;
+    point[0] = xp * std::cos(theta) - yp * std::sin(theta);
+    point[1] = yp * std::cos(theta) + xp * std::cos(theta);
+
+    // Calculate angle of tangent
+    theta = std::acos(point[0]
+                      / std::sqrt(point[0] * point[0] + point[1] * point[1]));
+    if (point[1] < 0)
+    {
+        theta = (pi - theta) + pi;
```
sethrj:
```suggestion
        theta = 2 * pi - theta;
```

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    }
+    real_type a1 = theta - t_point;
+    if (std::fabs(theta) < std::fabs(tmax_ + a_) && a1 >= a_)
```
sethrj: `theta` should always be in `[0, 2pi)` so why the absolute values?

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+        theta += pi;
+    }
+
+    Array<real_type, 2> point;
+    point[0] = xp * std::cos(theta) - yp * std::sin(theta);
+    point[1] = yp * std::cos(theta) + xp * std::cos(theta);
+
+    // Calculate angle of tangent
+    theta = std::acos(point[0]
+                      / std::sqrt(point[0] * point[0] + point[1] * point[1]));
```
sethrj:
```suggestion
                      / norm(point));
```

src/orange/surf/Involute.cc
```diff
@@ -0,0 +1,26 @@
...
+                   real_type tmin,
+                   real_type tmax)
+    : origin_(origin), r_b_(radius), a_(a), sign_(sign), tmin_(tmin), tmax_(tmax)
```
sethrj: Are there limits in the relationship getween tmin and max? Does `t` mean theta here? (If so, I'd add it). Would it simplify the calculation later (i.e., some of the `while` loops etc). to constrain one or both of these to be in `[0, 2pi)`?

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    {
+        theta = (pi - theta) + pi;
+    }
+    real_type a1 = theta - t_point;
+    if (std::fabs(theta) < std::fabs(tmax_ + a_) && a1 >= a_)
+    {
+        return SignedSense::inside;
+    }
+
+    while (std::fabs(theta) < (std::fabs(tmax_ + a_)))
```
sethrj: You've got extra parens here (and I've seen unnecessary ones elsewhere too). Also you shoudl definitely constrain tmax: if the user puts in `100000 * pi` this loop will take an unneccessarily long time.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    /*
+     * Calculate distance to origin and obtain t value for distance.
+     */
+    real_type const rxy2 = ipow<2>(x) + ipow<2>(y);
+    real_type const t_point2 = (rxy2 / ipow<2>(r_b_)) - 1;
+    real_type const t_point = std::sqrt(t_point2);
+
+    /*
+     * Check if Point is on involute.
+     */
```
sethrj:
```suggestion
    // Check if Point is on involute
```
and below (since the comment applies to just a couple of contiguous lines)

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+ * /f]
+ */
+CELER_FORCEINLINE_FUNCTION Real3 Involute::calc_normal(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    /*
+     * Calculate distance to origin and obtain t value for distance.
+     */
```
sethrj: Make this a one-line comment too, and check the rest of the file for them

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,314 @@
...
+    // Calculate all possible straight-line intersections with this surface
+    inline CELER_FUNCTION Intersections calc_intersections(
+        Real3 const& pos, Real3 const& dir, SurfaceState on_surface) const;
+
+    // Calculate outward normal at a position
+    inline CELER_FUNCTION Real3 calc_normal(Real3 const& pos) const;
+
+  private:
+    // Location of the center of circle of involute
+    Real3 origin_;
```
sethrj: Since the surface is invariant along `z`, consider defining `Real2  = Array<real_type, 2>` and using that instead.

src/orange/surf/detail/SurfaceTransformer.cc
```diff
@@ -114,6 +115,21 @@ Sphere SurfaceTransformer::operator()(Sphere const& other) const
                                   other.radius_sq());
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Transform an Involute.
+ */
+Involute SurfaceTransformer::operator()(Involute const& other) const
+{
+    Involute invo{tr_.transform_up(other.origin()),
```
sethrj: This isn't correct: since the involute is *always* z-aligned, I think we have to check that the transformation is purely an x-y rotation, then adjust the theta parameters to match. We can make this a follow-on PR and for now replace this function with `CELER_NOT_IMPLEMENTED("transformed involutes")`

src/orange/surf/detail/SurfaceTranslator.cc
```diff
@@ -106,6 +106,21 @@ Sphere SurfaceTranslator::operator()(Sphere const& other) const
                                   other.radius_sq());
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a translated Involute.
+ */
+Involute SurfaceTranslator::operator()(Involute const& other) const
+{
```
sethrj: You should add a test for this function.

test/orange/surf/pythonInvoluteReference/inOut.py
```diff
@@ -0,0 +1,272 @@
+import numpy as np
```
sethrj: Please rename this file to `test/orange/surf/doc/involute-in-out.py`

test/orange/surf/pythonInvoluteReference/inOut.py
```diff
@@ -0,0 +1,272 @@
...
+import matplotlib.pyplot as plt
+import copy
+import math
+
+from involute import Involute
+
+# Add Fuzzy Logic
+
+class inOut:
```
sethrj:
```suggestion
class InOutTester:
```

test/orange/surf/pythonInvoluteReference/inOut.py
```diff
@@ -0,0 +1,272 @@
...
+        
+        
+        self.tmin = np.sqrt(rmin**2/self.rb**2-1) * self.sign
+        self.tmax = np.sqrt(rmax**2/self.rb**2-1) * self.sign
+
+            
+        self.rmin = rmin
+        self.rmax = rmax
+            
+    def isIn(self, x,y):
```
sethrj:
```suggestion
    def __call__(self, x,y):
```

test/orange/surf/pythonInvoluteReference/inOut.py
```diff
@@ -0,0 +1,272 @@
...
+        return False
+        
+# Test Involute      
+rb = 1.0
+a0 = np.pi*0.5
+t = np.linspace(0,4*np.pi,500)
+sign = 1
+involute = Involute(rb, a0, t, sign)
+
+inout = inOut(involute, 2.0, 9.5)
```
sethrj:
```suggestion
is_inside = InOut(involute, 2.0, 9.5)
```

test/orange/surf/pythonInvoluteReference/inOut.py
```diff
@@ -0,0 +1,272 @@
...
+plt.plot(xlin,ylin, color ="Orange", label="ThetaMax")
+
+
+# Origin Test
+print("-------------------\n"
+      "Testing Point (0,0)\nExpected Result: False")
+x = 0 
+y = 0
+print("Obtained Result : {}".format(inout.isIn(x, y)))
+if inout.isIn(x, y):
```
sethrj:
```suggestion
if is_inside(x, y):
```

test/orange/surf/pythonInvoluteReference/involute.py
```diff
@@ -0,0 +1,412 @@
+import numpy as np
```
sethrj: Add headers to this file too (run `celeritas-gen.py temp.py` from `scripts/dev`)

test/orange/surf/Involute.test.cc
```diff
@@ -0,0 +1,88 @@
...
+
+    auto dist = invo.calc_intersections(
+        Real3{-5.8653052986571326, -0.30468305643505367 + 1, 0},
+        Real3{u, v, w},
+        SurfaceState::on);
+
+    EXPECT_SOFT_EQ(0.0, dist[0]);
+    EXPECT_SOFT_EQ(8.2132658639260558, dist[1]);
+    EXPECT_SOFT_EQ(9.167603472624553 * convert, dist[2]);
+}
```
sethrj, line 171: This test is *very* sparse for the amount of code that's gone into the involute. You should have them for a combination of radii, and definitely for both CW/CCW.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+
+    /*
+     * Define tolerances.
+     * tol_point gives the tolerance level for a point and is set to 1e-7,
+     * account for the floating point error when performing square roots.
+     * tol_conv gives the tolerance for the Regular Falsi iteration,
+     * ensuring that the root found produces a result that within 1e-8 of 0.
+     */
+    real_type const tol_point = 1e-7;
+    real_type const tol_conv = 1e-8;
```
sethrj: I don't think the absolute tolerance should be hardcoded. We Maybe you should use `Tolerance<>::sqrt_quadratic()` (see the `Tolerance` class in `orange/OrangeTypes.hh`) for the relative (convergence) tolerance, and multiply by the radius to get the absolute tolerance. Another approach to the tolerance would be to use the values in `Tolerance<T>::from_default`...

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    // Initialize distance vector with root counter j
+    Array<real_type, 3> dist;
+    real_type j = 0;
+
+    // Conversion constant for 2-D distance to 3-D distance
+
+    real_type convert = std::sqrt(ipow<2>(v) + ipow<2>(u) + ipow<2>(w))
+                        / std::sqrt(ipow<2>(v) + ipow<2>(u));
```
sethrj: This is just the inverse of the polar cosine angle: u/v/w are already a unit vector, so the numerator is 1.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+
+    real_type beta;
+
+    if (u != 0)
+    {
+        // Standard method
+        beta = std::atan(-v / u);
+    }  // Edge case
+    else if (-v < 0)
+    {
```
sethrj: Comments go inside of the block that they apply to. Also check the negation here:
```suggestion
    } 
    else if (v > 0)
    {
         // Edge case
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+    t_upper += std::fmax(0, -std::floor(t_upper / pi)) * pi;
+
+    // Parameters that will be used in loop
+    int i = 1;
+    real_type t_alpha;
+    real_type t_beta;
+    real_type t_gamma;
+    real_type ft_gamma;
+    real_type ft_alpha;
+    real_type ft_beta;
```
sethrj: Declare these where you initialize them.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+            t_lower = t_upper;
+            t_upper += pi / i;
+            i++;
+        }
+
+        for (int k = 0; k < j; k++)
+        {
+            // Convert 2-D distances to 3-D distances
+            result[k] = dist[k] * convert;
+        }
```
sethrj: You only need to do this after the solve has converged

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    // Lambda used for calculating the roots using Regular Falsi Iteration
+    auto root = [](real_type t,
+                   real_type x,
+                   real_type y,
+                   real_type u,
+                   real_type v,
+                   real_type r_b_,
+                   real_type a_) {
```
sethrj: remove the underscores from the function parameters.  Better yet, just capture `this` and then you can use those guys without passing them in.
`[this](...)`

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        for (int k = 0; k < j; k++)
+        {
+            // Convert 2-D distances to 3-D distances
+            result[k] = dist[k] * convert;
+        }
+    }
+    return result;
+}
```
sethrj, line 249: This function is far too long and complicated: it has 200 lines and 13 different conditionals (if/while/for) in it. Can you think of ways to break it into smaller *testable* pieces, each of which has only a couple of conditionals?

sethrj (changes requested):
This is some great work! It's a very good first pull request, and even though it needs a little work to reduce the complexity, I think we'll be able to get it in without too much more work 😄

src/orange/surf/Involute.hh
VHLM2001: tmax is already constrained in the contractors for the Involute in Involute.cc, and InvoluteSolver InvoluteSolver.hh

sethrj: I see:
```c++
CELER_EXPECT(std::fabs(tmax) < 2 * pi + std::fabs(tmin));
```
but no condition on tmin? Thanks for pointing those out. But my question is also why tmax/tmin can be negative? (Why are you checking the absolute value rather than the value?)  Is it possible to simplify the logic by processing the originally constructed values?

VHLM2001: That was a left er from when I was solving the negative and positive involutes separately.  I am adding `CELER_EXCEPT` for ensuring they are positive

src/orange/surf/Involute.hh
VHLM2001: Think I got this, I'll push it by 3pm today along with the changes for some of the comments below.

src/orange/surf/Involute.hh
VHLM2001: Having trouble with the SurfaceTranslator with implementing this will revisit later.

src/orange/surf/Involute.cc
```diff
@@ -10,17 +10,35 @@
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
+/*
+ * \parma radius: radius of the circle of involute
```
sethrj:
```suggestion
/*!
 * \param radius: radius of the circle of involute
```
🇮🇹🧀

src/orange/surf/Involute.cc
```diff
@@ -10,17 +10,35 @@
...
-    CELER_EXPECT(radius > 0);
     CELER_EXPECT(a > 0);
-    CELER_EXPECT(std::fabs(tmax) < 2 * pi + std::fabs(tmin));
+    CELER_EXPECT(tmax > 0);
+    CELER_EXPECT(tmin > 0);
+    CELER_EXPECT(tmax < 2 * pi + tmin);
+
+    if (sign_)
+    {
+        a_ = -a + pi;
```
sethrj: Or save an operation...
```suggestion
        a_ = pi - a;
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -44,11 +44,18 @@ class InvoluteSolver
     //! \name Type aliases
     using Intersections = Array<real_type, 3>;
 
+    //! Enum defining chirality of involute
+    enum Sign
+    {
+        CLOCKWISE = 1 == 1,
+        ANTICLOCKWISE = 1 != 1
```
sethrj:
```suggestion
    enum Sign : bool
    {
        counterclockwise = 0,
        clockwise, //!< Apply symmetry when solving
```

I'm curious about the semantics of your definitions as originally written though...

(and by the way, counterclockwise/CCW is the American English spelling, anticlockwise is British, and it would be too much to have "locale" support for our codebase 😄)

src/orange/surf/detail/InvoluteSolver.hh
VHLM2001: Got it down to 120 lines and 7 conditionals (counting one else) in the main function

test/orange/surf/Involute.test.cc
VHLM2001: I'm not sure if you mean just for the three point case or just more in general?

sethrj:
@VHLM2001 Perhaps to simplify the review process, you could first create a pull request with *just* the core of the nonlinear solver? That could go in `corecel/math` provided it's sufficiently general, something like:
```c++
template<class F>
class RegulaFalsiRootFinder
{
  public:
    RegulaFalsiRootFinder(F&& func, real_type tolerance);

    // Solve between
    real_type operator()(real_type left, real_type right);
};
```

You'll then be able to run this locally in the function with:
```c++
auto calc_zeros = [&](real_type t) {
    return ...;
};
RegulaFalsiRootFinder find_root{calc_zeros, tol_conv.abs};
while (...)
{
    t_gamma = find_root(t_alpha, t_beta);
}
```

PS also is it "regular falsi" or [regula falsi](https://en.wikipedia.org/wiki/Regula_falsi)?

A class like this would be easy to test because you can give it a trivial 1D function instead of a complex involute function.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,382 @@
...
+    // found
+    real_type t_lower = 0;
+    real_type t_upper = beta - a_;
+
+    // Round t_upper to the first positive multiple of pi
+    t_upper += std::fmax(0, -std::floor(t_upper / pi)) * pi;
+
+    // Parameters that will be used in loop
+    int i = 1;
+
```
sethrj, line 193: With your new solver class, here you can define:
```c++
auto calc_theta_intersect = [&](real_type theta) {
    real_type sint, cost;
    sincos(t + a_, &sint, &cost);
    real_type alpha = u * sint - v * cost;
    real_type beta = t * (u * cost + v * sint);
    return r_b_ * (alpha - beta) + x * v - y * u;
};
RegulaFalsi find_root_between{calc_theta_intersect, tol_conv.abs};
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,382 @@
...
+                                              t_beta,
+                                              ft_alpha,
+                                              ft_beta,
+                                              r_b_,
+                                              a_,
+                                              x,
+                                              y,
+                                              u,
+                                              v,
+                                              tol_conv.abs);
```
sethrj: .... and this becomes
```suggestion
            // Regula Falsi Iteration
            real_type t_gamma = find_root_between(t_alpha, t_beta);
```
which is much more readable right?

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,382 @@
...
+ * Convert root to distance by calculating the point on the involute given by
+ * the root and then taking the distance to that point from the particle.
+ */
+CELER_FUNCTION real_type InvoluteSolver::calc_dist(real_type x,
+                                                   real_type y,
+                                                   real_type u,
+                                                   real_type v,
+                                                   real_type t,
+                                                   real_type r_b,
+                                                   real_type a,
```
sethrj: There's no need to pass these two since they're already class data

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -51,7 +51,7 @@ class RegulaFalsi
     real_type tol_;
 
     // Maximum amount of iterations
-    static constexpr inline int max_iters_ = 1000;
+    static constexpr inline int max_iters_ = 500;
```
sethrj: If this needs to change, perhaps it should be an input parameter? Also, 500 iterations is a *lot* on GPU (since one thread may stall the entire kernel). Is the reason to choose Regula Falsi over [Newton Rhapson](https://mathworld.wolfram.com/NewtonsMethod.html) the stability or the sign checking or the lack of analytic derivative? Newton's method gives quadratic convergence rather than linear so that would accelerate this quite a bit...

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -52,7 +52,7 @@ class RegulaFalsi
     real_type tol_;
 
     // Maximum amount of iterations
-    static constexpr inline int max_iters_ = 500;
+    static constexpr inline int max_iters_ = 50000;
```
sethrj: This is never going to work on GPU 😅

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,320 @@
...
+ * \f]
+ */
+CELER_FORCEINLINE_FUNCTION Real3 Involute::calc_normal(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    // Calculate distance to origin and obtain t value for distance
+    real_type const rxy_sq = ipow<2>(x) + ipow<2>(y);
+    real_type const t_point = std::sqrt((rxy_sq / (ipow<2>(r_b_))) - 1);
```
sethrj:
```suggestion
    real_type const t_point = std::sqrt(rxy_sq / ipow<2>(r_b_) - 1);
```
and we might actually want to combine it with the next expression and add a check for being barely inside the inner radius (which could conceivably happen due to floating point issues)

```c++
    real_type const t_point_sq = (ipow<2>(x) + ipow<2>(y)) / ipow<2>(r_b_) - 1;
    real_type const angle = std::sqrt(clamp_to_nonneg(t_point_sq)) + a_;
```

(using `corecel/math/Algorithms.hh`)

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,320 @@
...
+    {
+        theta = 2 * pi - theta;
+    }
+    real_type a1 = theta - t;
+    if (theta < tmax_ + a_ && a1 >= a_)
+    {
+        return SignedSense::inside;
+    }
+
+    while (theta < tmax_ + a_)
```
sethrj: Does this need to be a while loop? Some extra assertions inside the function would be good for understanding the domains of the operations, e.g. `CELER_ASSERT(theta >= 0 && theta <= pi + a_)` (or whatever).

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,320 @@
...
+    real_type y_prime = std::sqrt(ipow<2>(r_b_) - ipow<2>(x_prime));
+    real_type theta = std::atan(y / x);
+    if (x < 0)
+    {
+        theta += pi;
+    }
+
+    Array<real_type, 2> point;
+    point[0] = x_prime * std::cos(theta) - y_prime * std::sin(theta);
+    point[1] = y_prime * std::cos(theta) + x_prime * std::sin(theta);
```
sethrj: Can these trig operations be simplified? It looks like you're taking the arctan and then doing sin/cos on the result, which means you can probably use basic ratio trig identities instead. Evaluating irrational functions is one of the most expensive things you can do on a GPU.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,320 @@
...
+    real_type const y = pos[1] - origin_[1];
+
+    if (this->sign() == Sign::clockwise)
+    {
+        x = negate(x);
+    }
+
+    // Calculate distance to origin and obtain t value for distance.
+    real_type const rxy_sq = ipow<2>(x) + ipow<2>(y);
+    real_type const t_point_sq = (rxy_sq / ipow<2>(r_b_)) - 1;
```
sethrj: Try to avoid creating temporary stack variables that are only used for intermediate computations. It might make the compiler generate less efficient code (unlikely these days the way optimizers work), but more importantly it obscures to the reader which variables are needed later and why. For example, rxy_sq is never used except for calculating t_point_sq. If it doesn't mak it less clear (and I don't think it does since you do this same thing a number of places), just write
```suggestion
    // Calculate distance to origin and obtain t value for distance.
    real_type const t_point_sq = (ipow<2>(x) + ipow<2>(y)) / ipow<2>(r_b_) - 1;
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+    real_type t_upper = beta - a_;
+
+    // Round t_upper to the first positive multiple of pi
+    real_type floor_upper = -std::floor(t_upper / pi);
+    t_upper += max<real_type>(0, floor_upper) * pi;
+
+    // Parameters that will be used in loop
+    int i = 1;
+
+    // Lambda used for calculating the roots using Regular Falsi Iteration
```
sethrj: Regular -> Regula here and below
```suggestion
    // Parameters that will be used in loop
    int i = 1;

    // Lambda used for calculating the roots using Regula Falsi Iteration
```
Instead of "parameters" (not very descriptive) can you add exactly what `i` is?

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+        {
+            // Regular Falsi Iteration: Sometimes will slowly converge
+            real_type t_gamma = find_root_between(t_lower, t_upper);
+
+            // Convert root to distance and store if positive and in interval
+            real_type dist = calc_dist(x, y, u, v, t_gamma);
+            if (dist > tol_point
+                || (!(on_surface == SurfaceState::on) && (dist > 0)))
+            {
+                result[j] = convert * dist;
```
sethrj, line 231: You should definitely add an assertion that you're not going out of bounds
```suggestion
                CELER_ASSERT(j < result.size());
                result[j] = convert * dist;
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+                j++;
+            }
+
+            // Obtain next bounds
+            t_lower = t_upper;
+            t_upper += pi;
+        }
+        else
+        {
+            // Incremet interval slowly until root is in interval
```
sethrj, line 241:
```suggestion
            // Increment interval slowly until root is in interval
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+                                                          real_type v)
+{
+    using constants::pi;
+
+    if (u != 0)
+    {
+        // Standard method
+        return std::atan(-v / u);
+    }
+    else if (-v < 0)
```
sethrj, line 267:
```suggestion
    else if (v > 0)
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+/*!
+ * Convert root to distance by calculating the point on the involute given by
+ * the root and then taking the distance to that point from the particle.
+ */
+CELER_FUNCTION real_type InvoluteSolver::calc_dist(
+    real_type x, real_type y, real_type u, real_type v, real_type t) const
+{
+    real_type angle = t + a_;
+    real_type x_inv = r_b_ * (std::cos(angle) + t * std::sin(angle));
+    real_type y_inv = r_b_ * (std::sin(angle) - t * std::cos(angle));
```
sethrj: This function evaluation is repeated in at least three different places between here and the involute solver. Maybe a little helper class would be useful:
```c++
class NormalizedInvolute
{
inline CELER_FUNCTION Real2 calc_point(real_type theta) const;

real_type r_b_;
real_type a_;
};
```

test/orange/surf/detail/InvoluteSolver.test.cc
```diff
@@ -0,0 +1,654 @@
...
+namespace celeritas
+{
+namespace detail
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+using constants::pi;
+using Sign = InvoluteSolver::Sign;
+using SurfaceSate = celeritas::SurfaceState;
```
sethrj, line 23:
```suggestion
using SurfaceState = celeritas::SurfaceState;
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,313 @@
...
+
+    real_type convert = 1 / std::sqrt(ipow<2>(v) + ipow<2>(u));
+    u *= convert;
+    v *= convert;
+
+    // Remove 0 dist if particle is on surface
+    if (on_surface == SurfaceState::on)
+    {
+        result[j] = 0;
+        j++;
```
sethrj: No we don't want this, a zero distance-to-intercept is bad because that results in a zero movement of the particle. If we're already on the surface, there is no intersection at zero.

test/orange/surf/Involute.test.cc
```diff
@@ -0,0 +1,160 @@
...
+        real_type v = -0.11508335932330707 * std::sin(1);
+        real_type w = std::cos(1);
+        real_type convert = 1 / std::sqrt(ipow<2>(v) + ipow<2>(u));
+
+        auto dist = invo.calc_intersections(
+            Real3{-5.8653052986571326, -0.30468305643505367 + 1, 0},
+            Real3{u, v, w},
+            SurfaceState::on);
+
+        EXPECT_SOFT_EQ(0.0, dist[0]);
```
sethrj: Zero shouldn't be an intersection, especially if it's on the involute.

test/orange/surf/Involute.test.cc
```diff
@@ -0,0 +1,160 @@
...
+        real_type u = 1;
+        real_type v = 0;
+
+        auto dist = invo.calc_intersections(
+            Real3{x, y, 0}, Real3{u, v, 0.0}, SurfaceState::off);
+
+        EXPECT_SOFT_EQ(no_intersection(), dist[0]);
+        EXPECT_SOFT_EQ(no_intersection(), dist[1]);
+        EXPECT_SOFT_EQ(no_intersection(), dist[2]);
+    }
```
sethrj, line 170: Can you add a test case where it has three nonzero intersections?

src/orange/surf/Involute.hh
VHLM2001: Think I figured how to push theta so it does not need to go through the while loop.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,305 @@
...
+    real_type x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    if (this->sign() == Sign::clockwise)
+    {
+        x = negate(x);
+    }
+
+    // Calculate distance to origin and obtain t value for distance.
+    real_type const t_point_sq = (ipow<2>(x) + ipow<2>(y) / ipow<2>(r_b_)) - 1;
```
sethrj: Hey are you sure this isn't supposed to be 
```suggestion
    real_type const t_point_sq = (ipow<2>(x) + ipow<2>(y)) / ipow<2>(r_b_) - 1;
```
???

If it is, then you definitely need more tests of the `is_inside` function.

Maybe instead of separate `x,y` variables you should store them in a `Real2 xy` as well. Then you can replace `ipow<2>(x) + ipow<2>(y)` with `dot_product(xy, xy)`.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    // Check if point is in interval
+
+    // Calculate tangent point
+    real_type x_prime = ipow<2>(r_b_) / std::sqrt(ipow<2>(x) + ipow<2>(y));
+    real_type y_prime = std::sqrt(ipow<2>(r_b_) - ipow<2>(x_prime));
+
+    point[0] = (x_prime * x - y_prime * y) / std::sqrt(ipow<2>(x) + ipow<2>(y));
+    point[1] = (y_prime * x + x_prime * y) / std::sqrt(ipow<2>(x) + ipow<2>(y));
```
sethrj: On a follow-up PR, before we integrate this into the main tracking capabilities, we should evaluate whether the compiler is able to optimize out repeated expressions like the `hypot` call here (which shows up three times!).

src/orange/surf/Involute.cc
```diff
@@ -0,0 +1,49 @@
...
+                   Sign sign,
+                   real_type tmin,
+                   real_type tmax)
+    : origin_(origin), r_b_(radius), a_(displacement), tmin_(tmin), tmax_(tmax)
+{
+    CELER_EXPECT(r_b_ >= 0);
+    CELER_EXPECT(a_ >= 0);
+    CELER_EXPECT(tmin_ >= 0);
+    CELER_EXPECT(tmax_ > 0);
+    CELER_EXPECT(tmax_ < 2 * constants::pi + tmin_);
```
sethrj: Should it be this?
```suggestion
    CELER_EXPECT(tmin_ >= 0 && tmin_ < 2 * constants::pi);
    CELER_EXPECT(tmax_ > tmin_ && tmax_ < 2 * constants::pi + tmin_);
```

src/orange/surf/Involute.cc
```diff
@@ -0,0 +1,49 @@
...
+Involute::Involute(Real2 const& origin,
+                   real_type radius,
+                   real_type displacement,
+                   Sign sign,
+                   real_type tmin,
+                   real_type tmax)
+    : origin_(origin), r_b_(radius), a_(displacement), tmin_(tmin), tmax_(tmax)
+{
+    CELER_EXPECT(r_b_ >= 0);
+    CELER_EXPECT(a_ >= 0);
```
sethrj: Should there be an upper bound on `a` as well (maybe 2pi?)

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,305 @@
...
+
+    // Calculate angle of tangent
+    real_type theta = std::acos(point[0] / norm(point));
+    if (point[1] < 0)
+    {
+        theta = 2 * pi - theta;
+    }
+    theta += max<real_type>(0, std::floor((tmax_ + a_ - theta) / (2 * pi))) * 2
+             * pi;
```
sethrj: So here, the range of theta starts out in `[0, π]`, then is `[0, 2π]`. tmax is in `[0, 4π)`, a is in `[0, 2π)` maybe, so
`tmax_ + a_ - theta` is in `[-2π, 6π)`.  If it's negative, `std::floor` returns the more negative multiple (so -π gets rounded to -2π and then `max`ed to zero).  So I think this is really just
```c+++
theta += max(real_type{0}, std::fmod(tmax_ + a_ - theta, 2 * pi));
```

I think a little extension of the comment above it could help readability too.

src/orange/surf/Involute.hh
```diff
@@ -0,0 +1,305 @@
...
+ * \f]
+ */
+CELER_FORCEINLINE_FUNCTION Real3 Involute::calc_normal(Real3 const& pos) const
+{
+    real_type const x = pos[0] - origin_[0];
+    real_type const y = pos[1] - origin_[1];
+
+    // Calculate normal
+    real_type const angle
+        = std::sqrt((ipow<2>(x) + ipow<2>(y) / (ipow<2>(r_b_))) - 1) + a_;
```
sethrj:
```suggestion
        = std::sqrt((ipow<2>(x) + ipow<2>(y)) / ipow<2>(r_b_) - 1) + a_;
```
You're sure it's not this?  Again, if so, it probably needs another test or two.

src/orange/surf/detail/InvolutePoint.hh
```diff
@@ -0,0 +1,86 @@
...
+
+/*!
+ * Calculate the point on an involute.
+ */
+CELER_FUNCTION Real2 InvolutePoint::operator()(real_type theta) const
+{
+    real_type angle = theta + a_;
+    Real2 point;
+    point[0] = r_b_ * (std::cos(angle) + theta * std::sin(angle));
+    point[1] = r_b_ * (std::sin(angle) - theta * std::cos(angle));
```
sethrj, line 81: This is another place where we can test whether repeating the same operations results in extra work.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,306 @@
...
+        u = -u;
+    }
+
+    /*
+     * Define tolerances.
+     * tol_point gives the tolerance level for a point,
+     * account for the floating point error when performing square roots.
+     * tol_conv gives the tolerance for the Regular Falsi iteration.
+     */
+    real_type tol_conv = r_b_ * tol();
```
sethrj: Move these tolerances closer to where they're being used. You can just inline the `tol_conv` into the solver definition.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,306 @@
...
+    // Slow down factor to increment bounds when a root cannot be found
+    int i = 1;
+
+    // Lambda used for calculating the roots using Regula Falsi Iteration
+    auto calc_t_intersect = [&](real_type t) {
+        real_type alpha = u * std::sin(t + a_) - v * std::cos(t + a_);
+        real_type beta = t * (u * std::cos(t + a_) + v * std::sin(t + a_));
+        return r_b_ * (alpha - beta) + x * v - y * u;
+    };
+    IllinoisRootFinder find_root_between{calc_t_intersect, tol_conv};
```
sethrj:
```suggestion
    IllinoisRootFinder find_root_between{calc_t_intersect, r_b_ * tol()};
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -0,0 +1,306 @@
...
+        // Only iterate when roots have different signs
+        if (signum<real_type>(ft_lower) != signum<real_type>(ft_upper))
+        {
+            // Regular Falsi Iteration: Sometimes will slowly converge
+            real_type t_gamma = find_root_between(t_lower, t_upper);
+
+            // Convert root to distance and store if positive and in interval
+            real_type dist = calc_dist(x, y, u, v, t_gamma);
+            if (dist > tol_point
+                || (!(on_surface == SurfaceState::on) && (dist > 0)))
```
sethrj: Better to write the more common case first:
```suggestion
            if ((dist > 0 && on_surface != SurfaceState::on)
                || dist > tol_point)
```
you could also combine this expression with
```suggestion
            if (dist > (on_surface == SurfaceState::on ? tol_point : 0))
```
which actually, since the `on_surface` and `tol_point` are loop invariant, means you should just set
```c++
tol_point = (on_surface == SurfaceState::on ? r_b_ * tol() * 100 : 0);
```
 outside of the loop, and simplify this comparison to just
```suggestion
            if (dist > tol_point)
```

sethrj (commented):
@VHLM2001 Sorry to give more comments, but I took another deep look at some of the expressions and need to run them by you again. It also looks like some of the previous comments weren't addressed: you may have to click the "Load more..." button to see them.

src/orange/surf/Involute.cc
VHLM2001: tmin_ shouldn't have any upper bound, but the rest seems fine

src/orange/surf/Involute.hh
```diff
@@ -221,34 +223,39 @@ CELER_FUNCTION SignedSense Involute::calc_sense(Real3 const& pos) const
     detail::InvolutePoint calc_point{this->r_b(), a_};
     Real2 point = calc_point(clamp_to_nonneg(t_point_sq));
 
-    if (x == point[0] && y == point[1])
+    if (xy[0] == point[0] && xy[1] == point[1])
```
sethrj: alternatively, `xy == point`

src/orange/surf/Involute.hh
```diff
@@ -221,34 +223,39 @@ CELER_FUNCTION SignedSense Involute::calc_sense(Real3 const& pos) const
...
     {
         return SignedSense::on;
     }
 
     // Check if point is in interval
 
     // Calculate tangent point
-    real_type x_prime = ipow<2>(r_b_) / std::sqrt(ipow<2>(x) + ipow<2>(y));
+    // Efficient compiler check
+    real_type x_prime = ipow<2>(r_b_) / std::sqrt(dot_product(xy, xy));
```
sethrj: Here and below:
```suggestion
    real_type x_prime = ipow<2>(r_b_) / norm(xy);
```
also same comment about comment having a `TODO:`

src/orange/surf/detail/InvolutePoint.hh
```diff
@@ -76,6 +76,7 @@ CELER_FUNCTION Real2 InvolutePoint::operator()(real_type theta) const
 {
     real_type angle = theta + a_;
     Real2 point;
+    // Efficient compiler check
```
sethrj: The comment here makes it read like you're performing a check of the compiler efficiency. Remember to write it as though it's not a note to yourself but a note to someone else:
```suggestion
    // TODO: check that compiler avoids recomputing trig functions
```

src/orange/surf/Involute.hh
```diff
@@ -285,12 +292,13 @@ Involute::calc_intersections(Real3 const& pos,
...
-    real_type const x = pos[0] - origin_[0];
-    real_type const y = pos[1] - origin_[1];
+    Real2 xy;
+    xy[0] = pos[0] - origin_[0];
+    xy[1] = pos[1] - origin_[1];
 
     // Calculate normal
     real_type const angle
-        = std::sqrt((ipow<2>(x) + ipow<2>(y) / (ipow<2>(r_b_))) - 1) + a_;
+        = std::sqrt((dot_product(xy, xy) / ipow<2>(r_b_)) - 1) + a_;
```
sethrj: There's a chance the point is going to be slightly inside but not exactly on the surface
```suggestion
        = std::sqrt(clamp_to_nonneg(dot_product(xy, xy) / ipow<2>(r_b_) - 1)) + a_;
```

test/orange/surf/Involute.test.cc
```diff
@@ -59,6 +59,8 @@ TEST(InvoluteTest, sense)
         {1, 0}, 1.0, 0.5 * pi, ccw, 1.732050808, 1.732050808 + 1.99 * pi};
     EXPECT_EQ(SignedSense::outside, invo.calc_sense({0, 0, 1}));
     EXPECT_EQ(SignedSense::outside, invo.calc_sense({0, 1.0, 1}));
+    EXPECT_EQ(SignedSense::outside, invo.calc_sense({2.999, 0, 1}));
+    EXPECT_EQ(SignedSense::inside, invo.calc_sense({3.001, 0, 1}));
```
sethrj: I reverted the involute test to the "buggy" version (where only the y component was being divided by the radius) and these tests still pass: a key thing during development is when you catch an error, you make sure you add a test that fails before and passes after, otherwise that error may someday creep back in. Maybe you should use python to sample a bunch of random points on a disk between an inner and outer radius for an involute, feed them into the unit test and output whether they're "inside" or "outside", and then plot the points in python (with the involute shape) and make sure the results are all sensible.

src/orange/surf/detail/InvoluteSolver.hh
sethrj: This comment didn't get addressed yet.

sethrj (approved):
All right! Nice work @VHLM2001 ! Thanks for incorporating all the feedback.

PR #1297, opened 2024-06-28
Use pre-step optical material ID in optical pre-generators

amandalund:
@sethrj [noticed](https://github.com/celeritas-project/celeritas/pull/1184#discussion_r1655375535) that we are using the post-boundary-crossing material in the optical pre-generators when we should be using the pre-step material. This fixes that.


src/celeritas/optical/OpticalGenData.hh
```diff
@@ -90,11 +90,12 @@ struct OpticalPreStepData
     units::LightSpeed speed;
     Real3 pos{};
     real_type time{};
+    OpticalMaterialId opt_mat;
 
     //! Check whether the data are assigned
     explicit CELER_FUNCTION operator bool() const
     {
-        return speed > zero_quantity();
+        return speed > zero_quantity() && opt_mat;
```
sethrj: It might be slightly more efficient to flip the order of these? Or even just test for `opt_mat`. Since we're going to be evaluating this on every thread once or twice per step...

src/celeritas/optical/detail/CerenkovPreGenExecutor.hh
```diff
@@ -59,8 +57,9 @@ CerenkovPreGenExecutor::operator()(CoreTrackView const& track)
     cerenkov_dist = {};
 
     auto sim = track.make_sim_view();
-    auto optmat_id = get_optical_material(track);
-    if (!optmat_id || sim.step_length() == 0)
+    auto step = state.step[tsid];
```
sethrj:
```suggestion
    auto const& step = state.step[tsid];
```

And I also noticed that `CerenkovPreGenerator` and `ScintillationPreGenerator` are missing const references to the pre-step data.

src/celeritas/optical/detail/ScintPreGenExecutor.hh
```diff
@@ -56,8 +54,9 @@ CELER_FUNCTION void ScintPreGenExecutor::operator()(CoreTrackView const& track)
     scintillation_dist = {};
 
     auto sim = track.make_sim_view();
-    auto optmat_id = get_optical_material(track);
-    if (!optmat_id || sim.step_length() == 0)
+    auto step = state.step[tsid];
```
sethrj:
```suggestion
    auto const& step = state.step[tsid];
```

sethrj (approved):
Subject to a couple of missing const references, looks good! Thanks! Ideally we would also have an "optical collector" test with multiple materials that would catch a problem like this...

src/celeritas/optical/OpticalGenData.hh
amandalund: I'll flip the order; we'll need to check the speed to make sure the particle didn't start the step with zero energy (I removed the `sim.step_length() == 0` check and replaced it with this).

amandalund:
Yeah, we'll definitely want expand the testing...

test/celeritas/optical/Scintillation.test.cc
```diff
@@ -300,15 +300,16 @@ TEST_F(ScintillationTest, basic)
...
     // Pre-generate optical distribution data
     ScintillationPreGenerator generate(
         this->make_particle_track_view(post_energy_, pdg::electron()),
         this->make_sim_track_view(step_length_),
         post_pos_,
-        opt_matid_,
         edep_,
         data,
-        this->build_pre_step());
+        pre_step);
```
sethrj, line 312: Good catch by the way—I realized this is why one of my tests was spitting out `NaN`...

PR #1298, opened 2024-06-28
Improve documentation for physics, RNG, stepping loop

sethrj:
This adds some documentation to the user manual. I also made a minor change to the RNG engine to parallelize reseeding if track-parallel OpenMP is in use and to split the RNG initialization code for ease of documentation.


doc/main/overview.rst
```diff
@@ -186,8 +198,49 @@ and querying the geometry state.
...
+      track = track_queue.pop()
+      while track.alive:
+         for apply_action in [pre, along, post]:
+            apply_action(track)
+         track_queue += track.secondaries
+
+There is effectively a data dependency between the track at step *i* and step
+*i + 1* that prevents vectorization. The approach Celeritas takes to
+"vectorize" the stepping loop on GPU is to have an outer loop over "step
+iterations" and an inner loop over "track slots":
```
stognini: Just double checking: do we have a definition of a track slot in the documentation?

doc/main/api/celeritas.rst
```diff
@@ -117,13 +124,31 @@ These classes correspond to JSON input files to the field setup.
...
-to sample a random number from a distribution. Distributions are function-like
+to sample a random number from a distribution.
+
+Engines
+~~~~~~~
+
+Celeritas defaults to using an in-house implementation of the XORWOW
+:cite:`marsaglia_xorshift_2003` bit shifting generator. Each thread's state is
+seeded at runtime (or at the start of each event when integrating with Geant4)
+by filling the state with bits generated from a 32-bit
```
amandalund: 🐿️ 
```suggestion
by filling the state with bits generated from a 32-bit MT19937 engine.
```
(although at the start of each event we skip ahead in the sequence)

amandalund (approved):
Looks good!

src/celeritas/random/XorwowRngData.cc
```diff
@@ -18,23 +18,21 @@ namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Resize and initialize with the seed stored in params.
+ * Initialize XORWOW states with well-distributed random data.
+ *
+ * cuRAND and hipRAND implement functions that permute the original (but
+ * seemingly arbitrary) seeds given in Marsaglia with 64 bits of seed data. We
```
stognini, line 24: Maybe worth placing the DOI or link to make it a bit easier to access? (Even though this ref is now in the references.bib)

doc/main/overview.rst
```diff
@@ -170,6 +172,16 @@ with the single Coulomb scattering model, is an implementation of the mixed
 simulation algorithm. It is the default model in Geant4 above 100 MeV and
 currently under development in Celeritas.
 
+Optical Physics
+===============
+
+TODO:
```
amandalund, line 178: This is a TODO for later and not something you meant to add in this PR, correct?

doc/main/api/celeritas.rst
sethrj: Yep, went back to look at the code docs for this and got sidetracked...

stognini (approved):

stognini:
Looks great. Thanks!

doc/main/overview.rst
sethrj: Yeah, just something we need to remember to do — I think even now some of the basic ordering descriptions would be helpful but maybe we should fill this in once we have the initial optical loop in place.

src/celeritas/random/XorwowRngData.cc
sethrj: Yeah, I was just thinking it would be nice to have a better way to document actual references inline in the code... I'm not sure if there's a good way at all to turn "inline references" into a bibtex document without some hacky preprocessing or something, unless we can insert ```:cite:`foo` ``` in the code and hope the associated reference shows up in the doxygen ...

doc/main/overview.rst
sethrj: Added. Thanks!

src/celeritas/random/XorwowRngData.cc
stognini: Yeah, I was not too hopeful to have a hyperlink that ends up working on Doxygen to be honest, so I just imagined a DOI (or similar) as plain text. So even if we can't click to be redirected, we can copy/paste it (or select, right click -> Search with Google).

sethrj: Oh I guess a hyperlink would work... I feel like that could be a separate issue where we do this for all our references.

PR #1299, opened 2024-06-29
Refactor GenericGrid classes to more closely mimic behavior of ValueGrid classes

hhollenb:
Initially, the `GenericGridBuilder` was designed to accept a scalar `Collection` that it would immediately populate with generic grid data, and return a `GenericGridData`. This is the opposite of the `ValueGridBuilder` which we use in `StepLimitBuilders`, which caches the grids to be built and then populates a `Collection` when a `ValueGridInserter` is supplied. I later added the `GenericGridInserter` on top of this to combine the steps of building a `GenericGridData` grid and putting it into a grid `Collection`, returning its new ID in the collection.

To have optical physics more closely mimic the conventions of core physics, I've refactored some of the `GenericGrid` classes. All classes called `Builder` are now intermediate objects which will later supply their grids to classes called `Inserter`, which are responsible for the actual population of grids in a given `Collection`. 

* Previously `GenericGridBuilder` --> now `GenericGridSingleInserter`. The term `Single` is meant to signify that only the real scalar `Collection` is populated, and the direct `GenericGridData` object itself is returned.
* `GenericGridInserter` has mainly the same behavior. Instead of using `GenericGridBuilder` previously it's now updated to use the `GenericGridSingleInserter`.
* New `GenericGridBuilder`. As with `ValueGridBuilder` it caches the raw data for a generic grid and can be called later to populate a `Collection` in the usual `StepLimitBuilder` algorithm. Its `build` method is templated based on the inserter, so that a `GenericGridInserter` that uses different ID types will return the correct ID.

Appropriate changes in existing code have been made to use the right classes.

### DIFF
--- a/src/celeritas/optical/OpticalPropertyParams.cc
+++ b/src/celeritas/optical/OpticalPropertyParams.cc
@@ -16,8 +16,7 @@
 #include "corecel/math/Algorithms.hh"
 #include "celeritas/Quantities.hh"
 #include "celeritas/Types.hh"
-#include "celeritas/grid/GenericGridBuilder.hh"
-#include "celeritas/grid/GenericGridData.hh"
+#include "celeritas/grid/GenericGridInserter.hh"
 #include "celeritas/io/ImportData.hh"
 
 namespace celeritas
@@ -55,8 +54,8 @@ OpticalPropertyParams::from_import(ImportData const& data)
 OpticalPropertyParams::OpticalPropertyParams(Input const& inp)
 {
     HostVal<OpticalPropertyData> data;
-    CollectionBuilder refractive_index{&data.refractive_index};
-    GenericGridBuilder build_grid(&data.reals);
+    GenericGridInserter insert_refractive_index{&data.reals,
+                                                &data.refractive_index};
     for (auto const& mat : inp.data)
     {
         // Store refractive index tabulated as a function of photon energy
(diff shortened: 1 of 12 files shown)


sethrj:
Oh nooooo I thought we talked about this at the hackathon, but the "inserter" stuff is legacy crap that we used to populate data before we had Collections. I'll take a look at this later but we want to change the *physics* code (which is basically the oldest in our code base) to be more like the more contemporary code... this is the danger of not updating (or at least warning about!) obsolete code 😞

src/celeritas/optical/OpticalPropertyParams.cc
```diff
@@ -77,9 +76,10 @@ OpticalPropertyParams::OpticalPropertyParams(Input const& inp)
                        << "refractive index values are not monotonically "
                           "increasing");
 
-        refractive_index.push_back(build_grid(ri_vec));
+        // refractive_index.push_back(build_grid(ri_vec));
+        insert_refractive_index(ri_vec);
```
sethrj, line 80: I do like the way this looks, actually... a helper function that encapsulates both the backend and the grid index is a good idea. I just freaked out a bit about reverting to the physics inserter.

sethrj (commented):
Sorry Hayden, we should've discussed how the optical process+model would build cross sections, especially in the context of https://github.com/celeritas-project/celeritas/issues/253 . Perhaps this is a good time to experiment with a better design for constructing cross sections. I'll make an issue for this and we can discuss in the group meeting Monday?

PR #1301, opened 2024-06-30
Add helper functions for sampling/calculating direction in interactors

amandalund:
As discussed in #1291, this adds a couple helper functions for the physics interactors. Any suggestions for better names would be welcome.


src/celeritas/em/interactor/detail/Utils.hh
```diff
@@ -0,0 +1,56 @@
...
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate exiting direction via conservation of momentum.
+ */
+inline CELER_FUNCTION Real3 calc_exiting_direction(real_type inc_momentum,
+                                                   real_type out_momentum,
+                                                   Real3 const& inc_dir,
+                                                   Real3 const& out_dir)
```
sethrj: What do you think about a little detail helper:
```c++
struct Momentum {
    real_type magnitude;
    Real3 const& direction;
};
```
so the components are more strongly associated:
```c++
result.direction =  calc_exiting_direction(
    {inc_momentum_.value(), inc_direction_},
    {gamma_energy_.value(), secondary_->direction});
```

src/celeritas/em/interactor/detail/Utils.hh
```diff
@@ -0,0 +1,56 @@
...
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample azimuthal angle, calculate Cartesian vector, and rotate direction.
+ */
+struct CartesianTransformSampler
```
sethrj: Maybe something like ...
```suggestion
 * Sample an exiting direction from a polar cosine and incident direction.
 *
 * Combine an already-sampled change in polar cosine (dot product of incident
 * and exiting) with a sampled uniform azimuthal direction, and apply that
 * rotation to the original track's incident direction.
 */
struct ExitingDirectionSampler
```

sethrj (approved):
Thanks Amanda! It's definitely an improvement to unify these implementations (though there might be a couple in the neutron physics too, and eventually optical as well, so perhaps we should move the helper utils to `phys`?)

src/celeritas/em/interactor/detail/Utils.hh
amandalund: I like that.

src/celeritas/em/interactor/detail/Utils.hh
amandalund: Better, thanks!

src/celeritas/phys/PhysicsUtils.hh
```diff
@@ -3,54 +3,62 @@
 // See the top-level COPYRIGHT file for details.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file celeritas/em/interactor/detail/Utils.hh
+//! \file celeritas/phys/PhysicsUtils.hh
```
sethrj: What do you think about InteractionUtils? 😅 or is that the final "bring me a rock" that will break the camel's back...

src/celeritas/phys/PhysicsUtils.hh
```diff
@@ -0,0 +1,64 @@
...
+    Real3 const& direction;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate exiting direction via conservation of momentum.
+ */
+inline CELER_FUNCTION Real3 calc_exiting_direction(Momentum inc_momentum,
+                                                   Momentum out_momentum)
+{
```
sethrj: Should we take the opportunity to add assertions on `magnitude > 0`? (Or >= 0?)

sethrj (approved):

src/celeritas/phys/PhysicsUtils.hh
amandalund: I did ask for it... sounds good!

sethrj (approved):
Thank you! That's an itch that I've wanted to scratch since we finished the initial round of EM physics.

PR #1304, opened 2024-07-03
Add debug output for track states

sethrj:
While debugging an exception in the status checker I got tired of trying to dig into the raw state data to access the current track's debug information. This adds JSON I/O, and lldb-callable `debug_print` functions, for the track.

There's now a hacky piece of code to set the "currently executing params" from the action sequence. @pcanal you're the resident expert on global variables, so perhaps you could take a look at the global variable? I only intend this functionality to be used in single-thread mode for interactive debugging, so perhaps it doesn't need to be thread-local (and it might be bad if we *do* thread-local because the main thread can catch an exception run by an openmp thread?). Later we might extend the JSON I/O to allow `KernelContextException` to merge the core track state with the core params data *without* a global variable/lock...

Calling from an `Executor` that has an assertion failure:
```
(lldb) call celeritas::debug_print(track)
{
 "geo": {
  "dir": [
   0.9998302826766889,
   0.010529089939196719,
   0.015117675340624488
  ],
  "is_on_boundary": false,
  "is_outside": false,
  "pos": [
   -2.135075225174846,
   0.0,
   0.0
  ],
  "volume_id": "inner@0x60000350ada0"
 },
 "particle": {
  "energy": [
   96.45131706860465,
   "MeV"
  ],
  "particle_id": "electron"
 },
 "sim": {
  "along_step_action": "along-step-neutral",
  "event_id": 2,
  "num_steps": 0,
  "parent_id": 0,
  "status": "alive",
  "step_length": null,
  "time": 6.626225658693907e-10,
  "track_id": 1
 },
 "thread_id": 26,
 "track_slot_id": 26
}
```


src/celeritas/global/DebugIO.json.cc
```diff
@@ -0,0 +1,167 @@
...
+        ASSIGN_TRANSFORMED(track_id, to_int);
+        ASSIGN_TRANSFORMED(parent_id, to_int);
+        ASSIGN_TRANSFORMED(event_id, to_int);
+        ASSIGN_TRANSFORMED(num_steps, passthrough);
+        ASSIGN_TRANSFORMED(event_id, to_int);
+        ASSIGN_TRANSFORMED(time, passthrough);
+        ASSIGN_TRANSFORMED(step_length, passthrough);
+    }
+
+    ASSIGN_TRANSFORMED_IF(num_looping_steps, passthrough, bool);
```
amandalund: Just curious why this one is assigned for inactive tracks too?

src/celeritas/global/Debug.cc
```diff
@@ -0,0 +1,68 @@
...
+    nlohmann::json j = view;
+    std::clog << j.dump(1) << std::endl;
+}
+
+//---------------------------------------------------------------------------//
+}  // namespace
+
+//---------------------------------------------------------------------------//
+/*!
+ * During an execute call, for interactive debugging, a pointer.
```
amandalund: Almost a haiku...

amandalund (approved):
Cool, looks good from my end!

src/celeritas/global/DebugIO.json.cc
sethrj: Mistake ❌

src/celeritas/global/Debug.cc
```diff
@@ -0,0 +1,68 @@
...
+    nlohmann::json j = view;
+    std::clog << j.dump(1) << std::endl;
+}
+
+//---------------------------------------------------------------------------//
+}  // namespace
+
+//---------------------------------------------------------------------------//
+/*!
+ * In the stepping loop, for interactive breakpoints, a debug pointer.
```
amandalund, line 38: 😆 💯

PR #1306, opened 2024-07-03
Add final state helper for ionization interactions

amandalund:
This adds a helper class for updating the final state of the incident particle and secondary electron for ionization and refactors the ionization interactors to use it.

### DIFF
--- a/src/celeritas/em/distribution/MuBBEnergyDistribution.hh
+++ b/src/celeritas/em/distribution/MuBBEnergyDistribution.hh
@@ -1,0 +1,169 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/em/distribution/MuBBEnergyDistribution.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <cmath>
+
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+#include "corecel/math/Algorithms.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/random/distribution/BernoulliDistribution.hh"
+#include "celeritas/random/distribution/UniformRealDistribution.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the energy of the delta ray for the MuBetheBloch ionization model.
+ */
+class MuBBEnergyDistribution
+{
+  public:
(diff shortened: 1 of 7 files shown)


src/celeritas/em/distribution/MuBBEnergyDistribution.hh
```diff
@@ -0,0 +1,169 @@
...
+{
+    real_type energy;
+    real_type target;
+    do
+    {
+        energy = electron_cutoff_ * max_secondary_energy_
+                 / UniformRealDistribution(electron_cutoff_,
+                                           max_secondary_energy_)(rng);
+        target = 1 - beta_sq_ * energy / max_secondary_energy_
+                 + ipow<2>(energy) / (2 * ipow<2>(total_energy_));
```
sethrj:
```suggestion
                 + real_type{0.5} * ipow<2>(energy / total_energy_);
```
Save a division and a multiplication 😉 (probably decreases our FLOP/s performance 😄)

src/celeritas/em/interactor/detail/BremFinalStateHelper.hh
```diff
@@ -49,16 +48,16 @@ class BremFinalStateHelper
...
+    Momentum inc_momentum_;
     // Exiting energy
-    Energy const exit_energy_;
+    Energy exit_energy_;
     // Bremsstrahlung photon id
-    ParticleId const gamma_id_;
+    ParticleId gamma_id_;
     // Exiting gamma energy
-    Energy const gamma_energy_;
+    Energy gamma_energy_;
```
sethrj, line 57: Yeah we were a bit ambitious with `const`ing everything 😅

src/celeritas/em/interactor/detail/IoniFinalStateHelper.hh
```diff
@@ -0,0 +1,128 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Update the final state of the incident particle and secondary electron.
+ */
+template<class Engine>
+CELER_FUNCTION Interaction IoniFinalStateHelper::operator()(Engine& rng)
+{
+    // TODO: If the \c UseAngularGeneratorFlag is set (false by default), use
+    // the angular generator interface to sample the secondary direction (see
+    // \c G4EmParameters::UseAngularGeneratorForIonisation())
```
sethrj: It seems like the flag in Geant4 basically lets users directly sample `secondary_->direction`... are we really ever going to support that? It would essentially require templating everything up through the model.

sethrj (approved):
LGTM! Thanks for the refactor.

src/celeritas/em/interactor/detail/IoniFinalStateHelper.hh
amandalund: Yeah almost certainly not :)

PR #1309, opened 2024-07-05
Add Bragg and ICRU73QO ionization models

amandalund:
This adds the Bragg and ICRU73QO ionization models used for muons and hadrons in the low energy region. Since the sampling of the secondary and final state is the same in the two models, a single interactor is used for both.


src/celeritas/em/model/BraggModel.cc
```diff
@@ -0,0 +1,97 @@
...
+ * Construct from model ID and other necessary data.
+ *
+ * TODO: This model also applies to hadrons.
+ */
+BraggModel::BraggModel(ActionId id, ParticleParams const& particles)
+    : ConcreteAction(id, "ioni-bragg", "interact by muon ionization (Bragg)")
+{
+    CELER_EXPECT(id);
+
+    data_.inc_particle = particles.find(pdg::mu_plus());
```
sethrj, line 36: Yeah I guess we will eventually want to have the user pass a vector of particle types for this model to apply to?

whokion:
Thanks for adding these two low energy models and an associated interactor for the muon ionization process.  However, it is not clear whether this addition is actually necessary as core pieces of these two models are already absorbed inclusively into the discrete delta ray cross section and the dE/dx table (i.e., there are no separated lambda and energy loss related tables associated with these two models, unlike other models).  The common interactor for sampling the delta ray final state also can be shared by the existing muon Bethe-Block interactor (with general treatments for the mass scaling factor of the maximum energy transfer to the delta ray, the low energy limits of model validity, and the conditional radiative correction).  Furthermore, the current approach may add some additional complication to handle the process and its associated discrete actions depending on the energy and charge of muons, which may be arguable though. The situation will be same for he charged hadron Bethe-Block ionization process and model (if we add them later) of which low energy model implementation can be again absorbed into common imported data and its interactor as a similar way.  I do not have a strong objection for this MR, but just want to discuss alternatives before a detailed review.

sethrj (approved):
LGTM but perhaps we should discuss with @whokion as suggeted?

amandalund:
Thanks @whokion and @sethrj! Yes, maybe we can discuss at our meeting today?

sethrj:
@amandalund can you also please update the physics doc page to include the new models?

amandalund:
@whokion following up on our discussion about combining the `BraggICRU73QO` and `MuBetheBloch` interactors into a single interactor, I think I'm leaning toward keeping them separate for now: 
 - Most of the duplication was eliminated with the addition of the `IoniFinalStateHelper`. The calculation of the max energy transfer is still duplicated, but pretty much everything else is different:
   - Incident particle type(s)
   - Calculation of the minimum energy of the delta ray
   - Lower and upper model limits
   - Rejection sampling of the delta ray energy (different target and envelope distributions, radiative correction in Bethe-Bloch)
 - We would need to use the same data for all 3 models: I'm not sure what the best way would be to handle the different incident particle types between MuBetheBloch and Bragg/ICRU73QO

whokion:
@amandalund Okay, fine with me.  Nevertheless, as there will be no separated processes for these two models, it may be better 1) to generalize them for other applicable charged hadrons and 2) to use the interactor inside MuBetheBlockInteractor (or combined them together?) so that it is actually called - still not sure the actual role of these two models as the MuIonizationProcess (which is not implemented yet) will import all associated Geant4 data including cross sections and energy losses from corresponding Geant4 models at the low energy.  I still think that all items that you listed may be consolidated into the MuBetheBlockInteractor with different combinations of charges and energy regions even though it is still arguable which way is the better approach.

amandalund:
@whokion here is how they should be used (but please correct me if I am misunderstanding your comment):
 - We will add a `MuIonizationProcess` which will construct the three models (`Bragg`, `ICRU73QO`, and `MuBetheBloch`)
 - As you said, the process will provide the imported per-process cross section and energy loss tables
 - The models will return an `Applicabiilty` describing the particle types and energy range they apply to, which will be used to build the `ModelGroup`s in the physics data (energy regions and the corresponding model IDs that apply in those regions)
 - When a discrete process is sampled, the model that applies at the incident particle's energy will be selected from the process's model group
 - The model class will launch a kernel to apply the correct interactor to all tracks that are undergoing that discrete interaction

Then as you said we will need to generalize for other applicable charged hadrons: the Bragg and ICRU models should be modified to accept the applicable particle types as input.

sethrj:
Sounds like Soon is advocating something like the "combined brems" model, which we never ended up upgrading to support multiple elements per material...

whokion:
Yes, I had the similar workflow, but my point was that all those trivial features can be simply absorbed into the muon ionization process/model/interactor.  Anyway, the MR is good for now as we do a short cut anyway and may not simulate any charged hadron in the foreseeable future.

amandalund:
Ok, then I'll leave this as is for now, but we can revisit combining the interactors in the future (especially e.g. if we add the hadron Bethe-Bloch which is also very similar).

src/celeritas/em/model/BraggModel.cc
```diff
@@ -0,0 +1,97 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Particle types and energy ranges that this model applies to.
+ */
+auto BraggModel::applicability() const -> SetApplicability
+{
+    Applicability applic;
+    applic.particle = data_.inc_particle;
+    applic.lower = zero_quantity();
```
whokion, line 57: should this be  `applic.lower = data_.lowest_kin_energy`;?

src/celeritas/em/model/ICRU73QOModel.cc
```diff
@@ -0,0 +1,97 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Particle types and energy ranges that this model applies to.
+ */
+auto ICRU73QOModel::applicability() const -> SetApplicability
+{
+    Applicability applic;
+    applic.particle = data_.inc_particle;
+    applic.lower = zero_quantity();
```
whokion, line 57: same here?

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
```diff
@@ -0,0 +1,189 @@
...
+    CutoffView const& cutoffs,
+    Real3 const& inc_direction,
+    StackAllocator<Secondary>& allocate)
+    : allocate_(allocate)
+    , inc_direction_(inc_direction)
+    , inc_energy_(particle.energy())
+    , inc_momentum_(particle.momentum())
+    , inc_mass_(particle.mass())
+    , beta_sq_(particle.beta_sq())
+    , electron_mass_(shared.electron_mass)
```
whokion, line 116: Just for a consistency check, why do we use data here instead of constants::electron_mass (reverse argument for constants::proton_mass - I know our internal convention, but it is still bit confusing.

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
```diff
@@ -0,0 +1,189 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Simulate discrete muon ionization.
+ */
+template<class Engine>
+CELER_FUNCTION Interaction BraggICRU73QOInteractor::operator()(Engine& rng)
+{
+    if (min_secondary_energy_ >= max_secondary_energy_)
```
whokion, line 136: Wonder whether this actually happens?

whokion (commented):
Sorry for the delayed review.  LGTM too.

whokion:
Did not know that this was already merged (probably my browser might not be automatically refreshed) and sorry for the delayed review - no problem not to address my comments.

sethrj:
Whoops, sorry I misunderstood the comments and merged too soon... 😔

whokion:
My fault! Should have reviewed on time.

amandalund:
Thanks @whokion! And no worries, I can address any comments in my next PR.

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
amandalund: Actually I've wondered this before too (why do we store the electron mass from the particle params in our interactor data instead of using the constant)... maybe @sethrj can clarify? The reason I used `constants::proton_mass` here is because we're not currently importing proton data.

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
amandalund: I think it's possible; I added a [test](https://github.com/celeritas-project/celeritas/blob/011b8daaa62753700b775c95ed09c4a87abb5210/test/celeritas/em/BraggICRU73QO.test.cc#L165) where the incident energy is just above the production cut and we do hit this case.

src/celeritas/em/model/BraggModel.cc
amandalund: I don't think so, since it's not the model energy limit, just used in the interactor to calculate the minimum delta ray energy. I think these models should be valid at all energies below 200 keV.

whokion: Okay, then the comment,

>  //! Kinetic energy limit for the Bragg model 

 Is somewhat misleading.  Probably, should it be `the kinetic energy limit of the delta ray` or something related to that (the minimum energy transfer from muon to electron?)?  We know that the process is valid down to 100eV which is the lower limit of the lambda and dE/dx, but how can the muon below the limit can transfer the kinetic energy above the limit to generate a delta ray? Anyway, I agree that it is not the lower limit of the model, which I confused from the comment - same for the ICRU model.

amandalund: That's a good point, I'll update the documentation as you suggested. The minimum energy transfer from the incident particle to the delta ray is the smaller of the cutoff energy and `kin_energy_limit * inc_mass / proton_mass`, so I don't thiink we can use the latter as the lower model energy limit since it could still be larger than the cutoff.

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
sethrj: Well, we *could* import protons now.

Regarding the inconsistency though: CLHEP and Celeritas both define "experimentally measured" physical constants in the `constants` namespace; and those come from different sources. The cross sections and kinematics we get are from Geant4 which is using their constants, which I assume are fed into the particle data. I've tried to "consistently" use those imported values for masses and such.

The proton mass is an odd case of why this is a problem. We use the CODATA 2018 value, 1.67262192369e-24 gram = `938.272 089 43(29)` MeV. CLHEP claims to use the PDG 2019 value, which is `938.272 088 16(29)` MeV (so they're about 3 sigma off). **HOWEVER**, the value CLHEP actually uses is `938.272 013` MeV. Where does this value come from? I don't know. It could be from an older PDG evaluation that they decided not to adjust because then answers would change.

But I feel like until we generate our own cross sections, it's better to keep using whatever approximate constants Geant4 uses. Importing our "experimental constants" at runtime also allows us to compare two different code versions with the same data, or vice versa, which is an important capability for future validation studies.

amandalund: That makes sense, thanks for the explanation @sethrj! So yes, we probably should be importing the proton data. I can make a note to update this bit of code once we do. (And yes, it looks like 938.272 013(23) MeV is the 2006 CODATA value.)

PR #1310, opened 2024-07-05
Add a helper function to copy a "collection group" back to the host

sethrj:
This refactors collection assignment semantics (the initial implementation wasn't allowed to use C++17's `if constexpr`) and adds a helper function to copy an entire *state collection group* back to the host for debugging.


sethrj:
Since this contains some rather low level changes to device code, I *did* run unit tests on wildstyle (CUDA) with debug-orange and ndebug-vecgeom.

amandalund (approved):
Nice, thanks @sethrj! I've definitely wished we had that functionality before for testing/debugging.

PR #1311, opened 2024-07-06
Add "tracking cut" to handle errors and kill tracks

sethrj:
This replaces #1290 by defining a "tracking cut" that, if set as a post-step action, will kill a track and deposit its energy locally. It is currently used both for "looping" tracks and for initialization errors, replacing a host-only validation/assertion with a host+device error status. Only the host warns the user: we should perhaps add another "track counter" that uses an atomic increment/copy or "mapped" memory to inform the user of the total number of errors per step.

I'm very open to suggestions for using a different terminology... we're using the Geant4 "cut" term, but elsewhere we apply particle physics "cutoff" values, but this class really just kills a track to deposit its energy so it's not exactly a cutoff.


src/celeritas/global/alongstep/detail/PropagationApplier.hh
```diff
@@ -127,17 +129,17 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView const& track)
...
                 << " leading to distance " << repr(p.distance)
                 << (p.boundary  ? " (boundary hit)"
                     : p.looping ? " (**LOOPING**)"
                                 : "")
-                << " failed to change position at " << repr(orig_pos)
-                << " with ending direction "
-                << repr(track.make_geo_view().dir());
+                << " failed to change position";
+            sim.status(TrackStatus::errored);
+            sim.post_step_action(track.tracking_cut_action());
```
amandalund, line 140: Should we also be checking this/setting an error status on device?

src/celeritas/phys/detail/TrackingCutExecutor.hh
```diff
@@ -0,0 +1,94 @@
...
+    {
+        auto geo = track.make_geo_view();
+        auto msg = CELER_LOG_LOCAL(error);
+        msg << "Tracking error (track ID " << sim.track_id().get()
+            << ", track slot " << track.track_slot_id().get() << ") at "
+            << repr(geo.pos()) << " along " << repr(geo.dir()) << ": ";
+        if (!geo.is_outside())
+        {
+            msg << "depositing " << deposited << " ["
+                << Energy::unit_type::label() << "] in "
```
amandalund, line 80:
```suggestion
            msg << "depositing " << deposited << " "
                << Energy::unit_type::label() << " energy in "
```
(just to be consistent with the message below)

amandalund (approved):
Thanks for this @sethrj, looks great! I agree that having the capability to output information on the number of errors on both the host and device would be valuable.

src/celeritas/global/alongstep/detail/PropagationApplier.hh
sethrj: Not yet I think: this is only in debug mode, it takes three extra full precision registers, and worst of all I know why it's usually happening: our field integration is taking place in global coordinates. (The integration sub-substep sizes are so small it's no wonder we step without changing position.) We should perhaps refactor field integration (from the substepper/integrator and lower, at least) so that we accumulate in a "from the start of the step" coordinate system, and then transform to the global system when calculating the local field.

src/celeritas/phys/detail/TrackingCutExecutor.hh
sethrj: I'm gonna add brackets to the other guy: we use this form in accel/ExceptionConverter.cc, RZ map debug output, etc.

src/celeritas/global/alongstep/detail/PropagationApplier.hh
amandalund: Ahh ok, good we understand what's causing it. If it's just in debug we don't really care about the extra registers, do we? Why not have the same behavior on host and device?

sethrj: I see what you mean. Sure, why not.

PR #1314, opened 2024-07-09
Add optical sub-namespace

hhollenb:
Add the sub-namespace `celeritas::optical` to help separate core and optical code. Most files and classes prefixed with `Optical` were renamed to either not include the prefix, or use a more descriptive prefix. Added the prefix `PreGen` to classes and files used by the Cerenkov and Scintillation pre-generators to specify they're in-between core and optical physics.

It may be worth adding a `celeritas::core` sub-namespace to ensure there's no ambiguity between similarly named types.


src/celeritas/optical/PreGenDistributionData.hh
```diff
@@ -36,14 +38,14 @@ struct OpticalStepData
 /*!
  * Input data for sampling optical photons.
  */
-struct OpticalDistributionData
+struct PreGenDistributionData
```
amandalund: Maybe something like `PrimaryDistributionData` or `GeneratorDistributionData` would be clearer since this is used by the generators, not pre-generators, to sample optical primaries?  (We could probably eventually also improve on the name `PreGenerator`...)

src/celeritas/optical/PreGenData.hh
```diff
@@ -39,7 +41,7 @@ struct OpticalBufferSize
  *
  * At least one of cerenkov and scintillation must be enabled.
  */
-struct OpticalGenSetup
+struct PreGenOptions
```
amandalund: Similarly for some of these other `Optical*` --> `PreGen*`, perhaps e.g. `Generator*` or `PrimaryGen*`?

hhollenb: So far I've only found that `OpticalDistributionData` is used by `ScintillationGenerator` or `CerenkovGenerator`. The other classes I've marked as `PreGen` I've found to be only used on the core physics side of the code, the `Generators` only on the optical side, and `OpticalDistributionData` and `OpticalCollector` to be the interface between the two. It may be worthwhile to leave such classes in the core namespace and have only the ones operating in the optical loop be in the optical namespace?

src/celeritas/optical/PreGenDistributionData.hh
hhollenb: Maybe `Dispatcher` instead of `PreGenerator`? In the sense that optical photons are dispatched to another part of the code (can't use off-loader or transporter). Can also try `Deferer`, `Forwarder`, `Memoizer`, etc.

src/celeritas/optical/PreGenData.hh
amandalund: Yeah I think that makes sense, I'd be good with leaving the classes used on the core side/interface out of the optical namespace.

sethrj:
@amandalund I'll let you review and merge this one?

hhollenb:
Naming convention changes after the recent commits:
* Classes which operate in the core physics loop are in the `celeritas` namespace
* Classes which operate in the optical physics loop are in the `celeritas::optical` namespace
* `PreGenerator` --> `Dispatcher`: these are classes in the core loop which are responsible for dispatching ionization and radiation optical secondaries to the optical loop.
* `Generator`: these classes take the dispatched optical photons and insert them into the optical loop

There's some classes which live at the interface of core and optical. In general I place them inside the optical namespace, but I don't have a particular preference.

src/celeritas/optical/OpticalCollector.hh
```diff
@@ -86,19 +88,20 @@ class OpticalCollector
...
     //// DATA ////
 
-    SPOpticalGenParams gen_params_;
+    SPDispatcherParams gen_params_;
 
     SPGatherAction gather_action_;
-    SPCerenkovPreGenAction cerenkov_pregen_action_;
-    SPScintPreGenAction scint_pregen_action_;
+    SPCerenkovDispatcherAction cerenkov_pregen_action_;
+    SPScintDispatcherAction scint_pregen_action_;
```
amandalund: We can update these and `gen_params_` above to reflect the new names:
```suggestion
    SPCerenkovDispatcherAction cerenkov_dispatcher_action_;
    SPScintDispatcherAction scint_dispatcher_action_;
```

src/celeritas/optical/TrackData.hh
```diff
@@ -45,7 +47,7 @@ void resize(OpticalPhysicsStateData<Ownership::value, M>*,
 /*!
  * Memspace-independent core variables.
  */
-struct OpticalScalars
+struct PhysicsScalars
```
amandalund: We might want to keep this `OpticalScalars` since these variables aren't necessarily related to the physics.

src/celeritas/optical/detail/DispatcherGatherAction.hh
```diff
@@ -24,7 +24,7 @@ struct OpticalGenStorage;
 /*!
  * Generate optical distribution data.
  */
-class PreGenGatherAction final : public ExplicitCoreActionInterface
+class DispatcherGatherAction final : public ExplicitCoreActionInterface
```
amandalund: Should we update the label for this action (`"optical-dispatcher-gather"`)?

src/celeritas/optical/detail/DispatcherParams.hh
```diff
@@ -59,8 +59,8 @@ class OpticalGenParams final : public AuxParamsInterface,
 template<MemSpace M>
 struct OpticalGenState : public AuxStateInterface
```
amandalund: Should this now be:
```suggestion
struct DispatcherState : public AuxStateInterface
```
?

src/celeritas/optical/GeneratorDistributionData.hh
```diff
@@ -16,11 +16,13 @@
...
 namespace celeritas
 {
+namespace optical
+{
 //---------------------------------------------------------------------------//
 /*!
  * Pre- and post-step data for sampling optical photons.
  */
-struct OpticalStepData
+struct DispatcherStepData
```
amandalund: Do you think we should call this `GeneratorStepData` since it's stored in the `GeneratorDistributionData` and used by the optical generators?

amandalund (commented):
Thanks @hhollenb, looking great!

@sethrj do you have any opinion on the new names? The biggest difference is that `PreGenerator` (and the `OpticalGen` data) are now `Dispatcher`.

src/celeritas/optical/TrackData.hh
sethrj: Yeah I guess they're analogous to `CoreScalars` (celeritas/global/CoreTrackData.hh
): so maybe just call them that?

sethrj:
@amandalund @hhollenb Regarding dispatcher: I really like the idea that  the "pre-generator" name is focused on creating "optical loop" generator data from the "core loop" step data. However, "dispatch" to me (and in computer science and in C++ context discussions thereof) usually means forwarding a command or instruction (see [wikipedia multiple dispatch for example](https://en.wikipedia.org/wiki/Multiple_dispatch#Understanding_dispatch)) so  I think that word is potentially confusing.

Since we usually refer to the analogous process from EM tracks as offloading (from Geant4 to the Celeritas core tracking loop), what do you think about *offloading* (from the core tracking loop to the optical tracking loop) so we end up with `optical::OffloadXXX` and can call the kernels "optical-offload-gather" for example?

hhollenb:
I like "dispatch" because it's (in a sense) forwarding the optical secondary production command from the core loop to the optical loop.

My main concern with "offload" was potential confusion with track offloading from CPU to GPU. If we use offloading to just mean Geant4 to Celeritas then I'd be happy with that, but I want to avoid any potential confusion about moving tracks between devices. Also, would we eventually need an optical track offloader from Geant4 to Celeritas optical loop? Or will they just be handled as a separate case when we're given an arbitrary track to manage? That might cause another conflict if we want offloading to be moving alive tracks from Geant4 to Celeritas versus Celeritas moving distribution data (and not tracks) internally.

sethrj:
Well, we're likely going to use the same "pregenerator"/"dispatcher"/"offload" to send optical photon distributions from Geant4 to Celeritas. And since theoretically we could run the celeritas core loop on CPU while still using an optical GPU loop, we might end up doing the "offloading" there.

hhollenb:
Renamed "Dispatcher" to "Offload" and fixed the discrepancies @amandalund pointed out. A simple grep check shows no instances of "pregen" or "dispatcher" in the src or test optical directories.

src/celeritas/optical/TrackData.hh
```diff
@@ -113,7 +113,7 @@ struct OpticalStateData
 
     GeoStateData<W, M> geometry;
     Items<OpticalMaterialId> materials;
-    PhysicsStateData<W, M> physics;
+    CoreStateData<W, M> physics;
```
amandalund: This (and the `CoreParamsData` above) should still be `Physics*Data`, right?

amandalund (approved):
Nice work @hhollenb! Thanks for bearing with us with all the renaming :)

PR #1315, opened 2024-07-09
Skip indirection array for sorted track slots

sethrj:
This updates the core state to avoid allocating or accessing a track slot array in the usual case that we're not sorting tracks. It also defines a constructor directly from track slot ID which we can use to simplify the track initialization kernel. Finally, it adds a core track helper function to simultaneously mark a track as "errored" and to set the post-step action to kill it. I've made this function non-const, which requires updating a number of function signatures to make it mutable.

### DIFF
--- a/src/celeritas/global/CoreTrackView.hh
+++ b/src/celeritas/global/CoreTrackView.hh
@@ -264,7 +264,7 @@ CELER_FUNCTION auto CoreTrackView::make_rng_engine() const -> RngEngine
  */
 CELER_FORCEINLINE_FUNCTION ThreadId CoreTrackView::thread_id() const
 {
-    CELER_ENSURE(thread_id_);
+    CELER_EXPECT(thread_id_);
     return thread_id_;
 }
 
--- a/src/celeritas/global/CoreTrackView.hh
+++ b/src/celeritas/global/CoreTrackView.hh
@@ -25,6 +25,8 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 /*!
  * Helper class to create views from core track data.
+ *
+ * TODO: const correctness? (Maybe have to wait till C++23's "deducing this"?)
  */
 class CoreTrackView
 {
@@ -40,6 +42,11 @@ class CoreTrackView
     inline CELER_FUNCTION CoreTrackView(ParamsRef const& params,
                                         StateRef const& states,
                                         ThreadId thread);
+
+    // Construct directly from a track slot ID
+    inline CELER_FUNCTION CoreTrackView(ParamsRef const& params,
+                                        StateRef const& states,
+                                        TrackSlotId slot);
 
     // Return a simulation management view
     inline CELER_FUNCTION SimTrackView make_sim_view() const;
@@ -71,8 +78,8 @@ class CoreTrackView
     // Return an RNG engine
     inline CELER_FUNCTION RngEngine make_rng_engine() const;
 
-    //! Get the index of the current thread in the current kernel
-    CELER_FUNCTION ThreadId thread_id() const { return thread_; }
+    // Get the index of the current thread in the current kernel
+    inline CELER_FUNCTION ThreadId thread_id() const;
 
     // Get the track's index among the states
     inline CELER_FUNCTION TrackSlotId track_slot_id() const;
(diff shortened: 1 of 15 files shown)


src/celeritas/global/CoreTrackView.hh
```diff
@@ -218,18 +252,29 @@ CELER_FUNCTION auto CoreTrackView::make_rng_engine() const -> RngEngine
...
- *              ? thread_.get()
- *              : states_.track_slots[thread_]}
- * \endcode
+ * \pre The thread ID is only set if the class is initialized with the thread
+ * ID (e.g. from \c TrackExecutor ), which is not the case in track
+ * initialization (where the "core track" is constructed from a vacancy).
+ */
+CELER_FORCEINLINE_FUNCTION ThreadId CoreTrackView::thread_id() const
+{
+    CELER_ENSURE(thread_id_);
```
amandalund:
```suggestion
    CELER_EXPECT(thread_id_);
```

src/celeritas/track/detail/InitTracksExecutor.hh
```diff
@@ -79,28 +80,20 @@ CELER_FUNCTION void InitTracksExecutor::operator()(ThreadId tid) const
...
-        return data.vacancies[idx];
-    }();
+    // View to the new track to be initialized
+    CoreTrackView vacancy{
+        *params, *state, [&] {
+            TrackSlotId idx{index_before(counters.num_vacancies, tid)};
+            return data.vacancies[idx];
+        }()};
 
     // Initialize the simulation state
```
amandalund:
```suggestion
    // Initialize the simulation state and particle
```

amandalund (approved):
Thanks @sethrj! The initialization looks nicer with the new track view constructor.

Maybe you could also update the diagnostic test results in this PR, which changed in #1311 when the "kill looping" action was replaced by the tracking cut:
```diff
diff --git a/test/celeritas/user/Diagnostic.test.cc b/test/celeritas/user/Diagnostic.test.cc
index 55efd9866..ea182b4dd 100644
--- a/test/celeritas/user/Diagnostic.test.cc
+++ b/test/celeritas/user/Diagnostic.test.cc
@@ -249,7 +249,7 @@ TEST_F(TestEm3DiagnosticTest, TEST_IF_CELER_DEVICE(device))
         EXPECT_VEC_EQ(expected_steps, result.steps);
 
         EXPECT_JSON_EQ(
-            R"json({"_category":"result","_index":["particle","action"],"_label":"action-diagnostic","actions":[[0,0,0,0,0,0,0,3,0,0,0,0,0,0,0,0,0,0,0,0,0,9,0],[0,0,0,996,0,0,2,0,0,0,0,20,509,0,0,0,0,0,0,0,0,521,0],[0,0,0,902,0,0,10,0,0,0,9,20,577,0,0,0,0,0,0,0,0,518,0]]})json",
+            R"json({"_category":"result","_index":["particle","action"],"_label":"action-diagnostic","actions":[[0,0,0,0,0,0,0,3,0,0,0,0,0,0,0,0,0,0,0,0,9,0,0],[0,0,0,996,0,0,2,0,0,0,0,20,509,0,0,0,0,0,0,0,521,0,0],[0,0,0,902,0,0,10,0,0,0,9,20,577,0,0,0,0,0,0,0,518,0,0]]})json",
             this->action_output());
         EXPECT_JSON_EQ(
             R"json({"_category":"result","_index":["particle","num_steps"],"_label":"step-diagnostic","steps":[[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],[0,0,0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],[0,0,5,2,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]]})json",
```

src/celeritas/global/CoreTrackView.hh
sethrj: Yeah, I was thinking about it in terms of "it's the core track view's responsibility to return a valid value" rather than "it's the caller's responsibility to require that the view was initialized using a thread"—e.g. if the assertion failed, is it the caller or the core track view that needs fixing... I think the latter condition is more easily fixable and the precondition is stated in the comments, so this should indeed be EXPECT.

esseivaju: If `thread_id_` is invalid, can we not return `track_slot_id_`? Or set `thread_id_ = track_slot_id_` in `CoreTrackView(ParamsRef const&, StateRef const&, TrackSlotId)` That is essentially what the previous implementation was doing

sethrj: Actually we could.... the only reason I avoided this is that we only (!) use the `thread_id` accessor in one place, the "pre-step" executor (which resets the stack allocator on thread zero). Sorry I forgot to let you look at this before I merged!

esseivaju: Ok, I suppose it's a detail since the assert would catch any misuse but I'd argue that having a more coherent interface is worth the cost of setting one register.

PR #1316, opened 2024-07-09
Detect and gracefully handle ORANGE geometry errors during tracking

sethrj:
This adds a *temporary scope* `failed` flag to the geometry. Rather than storing this in global memory across kernels, it's stored in the track view as a temporary boolean. If initialization or surface crossing fails for ORANGE, this flag is set and the track initialization/boundary crossing kills the track.


sethrj:
Ping @elliottbiondo

amandalund (approved):
I think this looks good @sethrj! One thing I'm still a little unclear on is when we should be validating/asserting vs. marking tracks as errored and killing them. Do we understand why these failures are occuring? E.g., would these errors most likely result from problems in the geometry definition, or are we possibly hiding bugs in our implementation? It would definitely be helpful now to have some output quantifying how many tracks we are killing.

sethrj:
@amandalund I think we'll be doing most of the initial debugging on CPU where we do get the output.  At the beginning we're going to hit bugs in the implementation I'm sure; but I think this mechanism will have to still be in place for actual user errors as well (e.g. slightly overlapping regions). I would definitely like to have diagnostics for the number of "errored" tracks on GPU...

elliottbiondo (approved):

PR #1317, opened 2024-07-10
Add optical absorption and Rayleigh interactors

hhollenb:
Add optical absorption and Rayleigh interactors and respective tests to be able to do initial testing of optical models. Also includes an optical `Interaction` class for the results of discrete optical interactions.

New files are placed in the `celeritas::optical` namespace, so the class and file names do not have the usual `Optical` prefix. Both interactors don't depend on any physics or material data, so they're rather simple.

The `optical::TackView` class is just a placeholder for testing, and will need to be changed later for when the optical state and track code is set up. I'm including it now so that `RayleighInteractor`'s function signature will match the usual pattern of passing a track view, and also removes ambiguity of passing two `Real3` arrays to a function.


src/celeritas/optical/Interaction.hh
```diff
@@ -0,0 +1,92 @@
...
+    enum class Action
+    {
+        scattered,  //!< Still alive, state has changed
+        absorbed,  //!< Absorbed by the material
+        unchanged,  //!< No state change, no secondaries
+        failed,  //!< Ran out of memory during sampling
+    };
+
+    Real3 direction;  //!< Post-interaction direction
+    Real3 polarization;  //!< Post-interaction polarization
```
sethrj, line 35: Even though we'll eventually need it, let's leave out polarization for now as discussed.

src/celeritas/optical/Interaction.hh
```diff
@@ -0,0 +1,92 @@
...
+ */
+struct Interaction
+{
+    //! Interaction result category
+    enum class Action
+    {
+        scattered,  //!< Still alive, state has changed
+        absorbed,  //!< Absorbed by the material
+        unchanged,  //!< No state change, no secondaries
+        failed,  //!< Ran out of memory during sampling
```
sethrj: I think we can leave out failure too since we're not worried about large showers of secondaries.

src/celeritas/optical/model/RayleighInteractor.hh
```diff
@@ -0,0 +1,111 @@
...
+{
+    CELER_EXPECT(is_soft_unit_vector(inc_dir_));
+    CELER_EXPECT(is_soft_unit_vector(inc_pol_));
+    CELER_EXPECT(soft_zero(dot_product(inc_dir_, inc_pol_)));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the scattered direction and polarization for a single optical
+ * Rayeliegh interaction.
```
sethrj:
```suggestion
 * Rayliegh interaction.
```

Wait, so Rayleigh interaction is just isotropic in the exiting direction if polarization isn't considered?

test/celeritas/optical/Rayleigh.test.cc
```diff
@@ -0,0 +1,140 @@
...
+{
+namespace optical
+{
+namespace test
+{
+using namespace ::celeritas::test;
+
+//---------------------------------------------------------------------------//
+
+class RayleighInteractorTest : public InteractorHostTestBase
```
sethrj, line 23: This old test base (again, one of the first ones we wrote in celeritas) is probably not suitable for optical physics. I think we could potentially use the optical test base...

test/celeritas/optical/Rayleigh.test.cc
```diff
@@ -0,0 +1,140 @@
...
+    static double const expected_pol_angle[] = {
+        0.46914077892327, 0.31484033374691, 0.82815845476385, 0.50516788333603};
+
+    EXPECT_VEC_SOFT_EQ(expected_dir_angle, dir_angle);
+    EXPECT_VEC_SOFT_EQ(expected_pol_angle, pol_angle);
+}
+
+TEST_F(RayleighInteractorTest, stress_test)
+{
+    int const num_samples = 1'000;
```
sethrj, line 82: Fancy! First use of the digits separator in our code 😄 we should consider doing this for the other large integers in use...

src/celeritas/optical/model/RayleighInteractor.hh
hhollenb: Geant4 uses the Rayleigh scattering model from: G.O. Depaola / Nuclear Instruments and Methods in Physics Research A 512 (2003) 619–630

It treats molecules as isotropic, and the only angular dependence in the differential XS is the angle between the old and new polarizations. So without polarizatoins, Rayleigh will just be isotropic scattering. Anisotropic scattering would require a different model (e.g. https://arxiv.org/pdf/2202.12156 ) but would still probably need polarizations.

test/celeritas/optical/Rayleigh.test.cc
hhollenb: I'll admit I was being a bit lazy with this one and wanted to get tests out and working -- I'll throw together an optical interactor test base

src/celeritas/optical/Interaction.hh
```diff
@@ -0,0 +1,92 @@
...
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * The result of a discrete optical interaction.
+ *
+ * Discrete optical interactions are always elastic. The wavelength of a
+ * a photon is only changed through absorption re-emission processes.
```
whokion:
```suggestion
 * photon is only changed through absorption re-emission processes.
```

src/celeritas/optical/Interaction.hh
whokion: For a future reference, `reemitted` may be added for the wavelength swift interaction.

src/celeritas/optical/model/AbsorptionInteractor.hh
```diff
@@ -0,0 +1,45 @@
...
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Sample an absorption interaction.
+ */
+CELER_FUNCTION Interaction AbsorptionInteractor::operator()() const
+{
+    return Interaction::from_absorption();
```
whokion: We shortly discussed before, but should the energy of absorbed photon be deposited locally (which may provide, at least, MC truth like information inside radiators as well as in the sensitive detectors such as PMT or SiPM) -  of course, the energy deposition can be done elsewhere (at the high level), but followed our standard action for physics interactions.

src/celeritas/optical/Interaction.hh
```diff
@@ -0,0 +1,92 @@
...
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * The result of a discrete optical interaction.
+ *
+ * Discrete optical interactions are always elastic. The wavelength of a
```
whokion: Not sure whether this statement is correct as absorption and WLS are not elastic.  Probably, do you mean that `all photon interactions are discrete`?

src/celeritas/optical/Interaction.hh
whokion: Have we also decided not to allocate any storage for secondaries (which should be fine for now as we do not have WLS yet)?  Just wondering though.

src/celeritas/optical/model/AbsorptionInteractor.hh
```diff
@@ -0,0 +1,45 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/optical/model/AbsorptionInteractor.hh
```
whokion: Should we create a separate directory for interactors (i.e, option/interactor )?

test/celeritas/optical/Rayleigh.test.cc
```diff
@@ -0,0 +1,140 @@
...
+    std::vector<real_type> pol_angle;
+
+    for ([[maybe_unused]] int i : range(num_samples))
+    {
+        Interaction result = interact(rng_engine);
+        this->sanity_check(result);
+
+        dir_angle.push_back(dot_product(result.direction, inc_direction_));
+        pol_angle.push_back(
+            dot_product(result.polarization, track_view().polarization()));
```
whokion: A good additional test is the product of the polarization vector and the direction.

whokion (approved):
Very good start.  LGTM with several minor comments!

src/celeritas/optical/Interaction.hh
sethrj: Correct, the first iteration let's go for absolute minimum functionality (as long as it doesn't prevent us from adding the new capabilities going forward).

src/celeritas/optical/model/AbsorptionInteractor.hh
hhollenb: The interaction result here will just be the proposed change to the photon track. We definitely need a way to say that the photon has been absorbed because the track will need to be killed. Whether or not that will perform a local energy deposition will be at a higher level where the `Interaction` result gets applied to each track. 

As for recording the actual energy deposit, I've currently written in the assumption that the energy of optical photons never changes, so when an absorption result is returned, the higher level loop should use the photon's current energy (if we decide to deposit it).

src/celeritas/optical/Interaction.hh
hhollenb: You're very right, I may have been too caught up in thinking about Rayleigh scattering at the time...

test/celeritas/optical/Rayleigh.test.cc
hhollenb: The sanity checks should always check that the polarization and direction are perpendicular for each interaction. Do you mean the initial direction - final polarization?

src/celeritas/optical/model/AbsorptionInteractor.hh
whokion: Yes, we can assume that the energy of optical photons is constant along a step. As long as the photon energy is deposited somewhere in the higher level, it should be okay.  Extended to WLS, the energy is not conversed (as some of energy will be dissipated through phonons or heat during absorption and reemission), so the difference in energy may be deposited locally even though this delta energy is not currently deposited in Geant4.  So, I was thinking that any energy deposition may be handled inside physics interactors consistently (if we support fine detailed optical simulation inside layers of a fancy photo-tube, like SIPM).  Either way probably be okay for now.

test/celeritas/optical/Rayleigh.test.cc
whokion: I overlooked the sanity check, which may also include  `dot_product(final direction, final polarization) = 0`

src/celeritas/optical/model/AbsorptionInteractor.hh
hhollenb: Thanks, that's good to know.

If we have access to the secondaries from the interaction, we should be able to calculate the energy deposition at a higher level, and so we don't have to check energy conservation across every interactor. I don't have a strong opinion either way at the moment.

We may also want to consider adding WLS now since it seems to have more complications to keep track of than Rayleigh or absorption.

whokion: Good. It would be better to merge this MR first so that we move to integrate and move to the rest of the optical photon loop.  WLS implementation needs more brain-storming for how to support multiple time structures which is poorly designed in Geant4 (i.e., WLS, WLS2 which are just same codes, but with different reemission functions, which can be combined within one model).

src/celeritas/optical/Interaction.hh
amandalund: We don't have to decide in this PR, but if the track initialization stuff will be similar to what's in our main stepping loop it's possible it will be easier to include secondaries from the start rather than removing that logic and adding it back in later.

sethrj: Perhaps... looking at the code though, the "locate tracks" executor uses core params and physics step view, and the initialization-from-secondaries is likewise complicated...

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -0,0 +1,111 @@
...
+        {
+            // Project polarization onto plane perpendicular to new direction
+            new_pol = make_unit_vector(inc_pol_ - projected_pol * new_dir);
+            new_pol *= BernoulliDistribution{real_type{0.5}}(rng) ? 1 : -1;
+        }
+
+        // Perform rejection sampling for with the probability of the
+        // scattered polarization overlap squared
+    } while (
+        BernoulliDistribution{ipow<2>(dot_product(new_pol, inc_pol_))}(rng));
```
amandalund: Should this be
```suggestion
        !BernoulliDistribution{ipow<2>(dot_product(new_pol, inc_pol_))}(rng));
```
?

hhollenb: In terms of Geant4's code, the while loop goes over
```
do {
...
} while (std::pow(cos_theta, 2) < G4UniformRand());
```
so they're accepting with probability $\cos^2 \Theta$. This matches the equations in their references, but differs from the equations in the Geant physics reference manual, which is confusing. I'll reword the documentation to make the sampling method clearer, because the definition of the angles is a bit confusing.

sethrj:
Since the generator/pregenerator code already has polarization in there, I'm taking back what I said about removing it—sorry @hhollenb about the flip-flop and sorry this PR has taken so long. Thanks again for the great work!

sethrj:
@amandalund Could you give a last quick once-over and hit the auto-merge button if it looks good? Thanks!

src/celeritas/optical/interactor/RayleighInteractor.hh
amandalund: Sorry if I'm being dense, but isn't `ipow<2>(dot_product(new_pol, inc_pol_))` equal to $\cos^2 \theta$? Which would mean we're currently _rejecting_ with probability $\cos^2 \theta$ (since `BernoulliDistribution{p_true}(rng)` is equivalent to `u < p_true`)?

test/celeritas/optical/Rayleigh.test.cc
```diff
@@ -0,0 +1,123 @@
...
+        real_type pol_angle = dot_product(result.polarization,
+                                          this->photon_track().polarization());
+        pol_angle_1st_moment += pol_angle;
+        pol_angle_2nd_moment += pol_angle * pol_angle;
+    }
+
+    dir_angle_1st_moment /= num_samples;
+    dir_angle_2nd_moment /= num_samples;
+    pol_angle_1st_moment /= num_samples;
+    pol_angle_2nd_moment /= num_samples;
```
amandalund: We could also check the average number of RNG samples.

src/celeritas/optical/interactor/RayleighInteractor.hh
sethrj: Perhaps this is my fault for insisting on using the non intuitive "Bernoulli distribution". Perhaps a helper function that does "reject less than" would be more readable and less error prone. We've already caught one or two of these...

hhollenb: > Sorry if I'm being dense, but isn't `ipow<2>(dot_product(new_pol, inc_pol_))` equal to cos2⁡θ? Which would mean we're currently _rejecting_ with probability cos2⁡θ (since `BernoulliDistribution{p_true}(rng)` is equivalent to `u < p_true`)?

Nope you're absolutely right, I mixed up that passing the Bernoulli sample was a rejection instead of accepting

test/celeritas/optical/Rayleigh.test.cc
hhollenb: Is there a way to query how many times an engine has been sampled, or would I need to make a wrapper?

amandalund: The `DiagnosticRngEngine` used in these tests is a wrapper that does that, so you should be able to call `rng_engine.count()` at the end to get it.

src/celeritas/optical/interactor/RayleighInteractor.hh
amandalund: Yeah we’ve run into this rejection/acceptance confusion with the Bernoulli distribution more than a few times now… you've said before we need a rejection sampler, I think you're right :)

sethrj: Yep https://github.com/celeritas-project/celeritas/pull/1138#pullrequestreview-1914662560 . I'll put that together right quick...

sethrj: https://github.com/celeritas-project/celeritas/pull/1354

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -0,0 +1,112 @@
...
+ * the new direction.
+ */
+class RayleighInteractor
+{
+  public:
+    //! Construct interactor from an optical track
+    inline CELER_FUNCTION
+    RayleighInteractor(TrackView const& particle, Real3 const& direction);
+
+    //! Sample an interaction with the given RNG
```
sethrj:
```suggestion
    // Construct interactor from an optical track
    inline CELER_FUNCTION
    RayleighInteractor(TrackView const& particle, Real3 const& direction);

    // Sample an interaction with the given RNG
```

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -0,0 +1,112 @@
...
+    Real3& new_pol = result.polarization;
+
+    IsotropicDistribution sample_direction{};
+    do
+    {
+        new_dir = sample_direction(rng);
+
+        auto projected_pol = dot_product(new_dir, inc_pol_);
+
+        if (projected_pol == 0)
```
sethrj: Should we use a soft equivalence here?

sethrj (approved):

sethrj:
I went ahead and updated with the rejection sampler to speed this up (and take a little work off your sholders!): @hhollenb the reduced rejection rate here looks correct to me, but I don't know about the updated scattering angles: https://github.com/celeritas-project/celeritas/pull/1317/commits/39c7a511521ce4c1cf6627506318f8244e1bb845

hhollenb:
> I went ahead and updated with the rejection sampler to speed this up (and take a little work off your sholders!): @hhollenb the reduced rejection rate here looks correct to me, but I don't know about the updated scattering angles: [39c7a51](https://github.com/celeritas-project/celeritas/commit/39c7a511521ce4c1cf6627506318f8244e1bb845)

Made a quick plot of the scattered polarization angles (e.g. `dot_product(new_pol, inc_pol_))`) and compared it to the expected $\cos^2\theta$ probability distribution and it looks like they're fairly similar:
rayleigh_dist
There is a little shape difference but I'm not entirely certain if that's within the error bounds or just an artifact of rejection sampling method.

sethrj:
@hhollenb thanks for the plot. It's not surprising the distributions are different since we changed "accept when" to "reject when", as @amandalund suggested. The question is whether the new distribution is correct...

Issue #1313, opened 2024-07-09
Debug parallel crashes running with multiple streams on Frontier


sethrj:
We discovered that ROCm 5.7.1 and higher hang during multithreaded Geant4 runs. The problem appears to be a regression in the async memory allocation that results in a race condition, or possibly a bug in thrust: we've seen some cases where a kernel launch on one thread and an async malloc/free on another cause the app to lock up.

TODO: fill this in from OLCF help tickets

sethrj:
Worked around with using #1318

Issue #1263, opened 2024-06-06
Cleanup tasks for v1.0


sethrj:
- [ ] Standardize nomenclature: #721 
- [ ] Reorganize files
  - [x] params/helper/distribution for `celeritas/random`? see #1356
- [ ] Revisit namespaces (setup vs runtime?)
- [x] Move stream operators alongside the classes (e.g., merge ArrayIO.hh into Array.hh) now that we're no longer worried about propagating headers into CUDA/HIP #2265 
- [ ] Replace `lock_guard` with `scoped_lock`
- [ ] Add numeric separators `1'234'567` where it's clean/meaningful to do so?
- [x] Upgrade `Span` with CTAD and smarter constructors to eliminate `make_span`?
- [x] [Trapezoid integrator](https://github.com/celeritas-project/celeritas/pull/1507/files/c3d68d1f70d871c29d1a91c898fb2532982a7e83#diff-2835de3c9a75074efcd522cf0204c744e75ccaa2097cd57b452639ed860517c7) (#1537 )
- [ ] [CDF sampler](https://github.com/celeritas-project/celeritas/pull/1507/files#discussion_r1844973621)
- [ ] [Use C indexing for cyl field](https://github.com/celeritas-project/celeritas/pull/1662#discussion_r1999707750)
- [ ] [use std::filesystem](https://github.com/celeritas-project/celeritas/pull/1703/files/2abdebba44433a9605351cc030ba700a8af2ae78#diff-328a9dcb6564cb931cfc5756d37f009bb23cc590e9b6f2925b864983f5eb909b)
- [x] ArrayUtils has a few Real3-specific/geometric utils (e.g. cross product), but sampling methods are in geocel: just move real3 and isotropic distribution to corecel
- [x] GeantGeoTrackView in single precision loses accuracy from `convert_from_geant` but maybe it doesn't matter
- [ ] Move `CELER_DEVICE_API_CALL` to `DeviceRuntimeApi.hh`
- [ ] Move `CELER_MPI_CALL` to `Mpi.hh`

LSchwiebert:
Where appropriate, the code for creating and invoking device kernels and related device operations should be more consistent between the global and optical tracks.

PR #1318, opened 2024-07-10
Add environment variable to disable broken HIP asynchronous allocation

sethrj:
This adds a runtime selectable environment variable `CELER_DEVICE_ASYNC` to let the user *force* synchronous memory allocations, since asynchronous malloc seems to be broken on newer ROCm versions (see #1313).

It also fixes the "status checker" test on HIP since it seems like "abort" on HIP platforms results in a SIGSEGV.

NOTE: this also fixes a CI failure with the latest GitHub actions image: apparently `--option=value` isn't valid in CTest and it's just been ignoring the argument until now. See https://gitlab.kitware.com/cmake/cmake/-/issues/24227, https://gitlab.kitware.com/cmake/cmake/-/issues/24215

### DIFF
--- a/src/corecel/sys/Environment.cc
+++ b/src/corecel/sys/Environment.cc
@@ -7,13 +7,26 @@
 //---------------------------------------------------------------------------//
 #include "Environment.hh"
 
+#include <algorithm>
 #include <cstdlib>
 #include <mutex>
 
 #include "corecel/Assert.hh"
+#include "corecel/io/Logger.hh"
 
 namespace celeritas
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+std::mutex& getenv_mutex()
+{
+    static std::mutex mu;
+    return mu;
+}
+//---------------------------------------------------------------------------//
+}  // namespace
+
 //---------------------------------------------------------------------------//
 // FREE FUNCTIONS
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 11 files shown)


src/corecel/sys/Environment.cc
```diff
@@ -34,11 +47,56 @@ Environment& environment()
  */
 std::string const& getenv(std::string const& key)
 {
-    static std::mutex getenv_mutex;
-    std::lock_guard<std::mutex> scoped_lock{getenv_mutex};
+    std::lock_guard<std::mutex> scoped_lock{getenv_mutex()};
```
esseivaju: `std::scoped_lock` is the most recent implementation.

```suggestion
    std::scoped_lock<std::mutex> scoped_lock{getenv_mutex()};
```

src/corecel/sys/Environment.cc
```diff
@@ -34,11 +47,56 @@ Environment& environment()
...
+ * for providing a diagnostic message to the user.
+ *
+ * - Allowed true values: <code>"1", "t", "yes", "true", "True"</code>
+ * - Allowed false values: <code>"0", "f", "no", "false", "False"</code>
+ * - Empty value returns the default
+ * - Other value warns and returns the default
+ */
+GetenvFlagResult getenv_flag(std::string const& key, bool default_val)
+{
+    std::lock_guard<std::mutex> scoped_lock{getenv_mutex()};
```
esseivaju:
```suggestion
    std::scoped_lock<std::mutex> scoped_lock{getenv_mutex()};
```

src/corecel/sys/Environment.cc
```diff
@@ -34,11 +47,56 @@ Environment& environment()
...
+ * environment variable or default value, and (2) an "insertion" flag
+ * specifying whether the default was used. The insertion result can be useful
+ * for providing a diagnostic message to the user.
+ *
+ * - Allowed true values: <code>"1", "t", "yes", "true", "True"</code>
+ * - Allowed false values: <code>"0", "f", "no", "false", "False"</code>
+ * - Empty value returns the default
+ * - Other value warns and returns the default
+ */
+GetenvFlagResult getenv_flag(std::string const& key, bool default_val)
```
esseivaju, line 73: maybe this should be renamed because this will insert the queried key into the environment, whereas the analogous function `getenv` doesn't do that.

src/corecel/sys/Stream.cc
```diff
@@ -100,6 +116,41 @@ void AsyncMemoryResource<Pointer>::do_deallocate(pointer p,
     }
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Whether asynchronous operations are supported.
+ *
+ * This is true if CUDA or HIP>=5.2 is in use, and can be disabled by setting
+ * the \c DEVICE_DISABLE_ASYNC environment variable.
```
esseivaju:
```suggestion
 * This is true if CUDA or HIP (5.2 <= HIP_VERSION < 5.7) is in use, and can be disabled by setting
 * the \c CELER_DEVICE_ASYNC=0 environment variable.
```

src/corecel/sys/Stream.cc
```diff
@@ -34,27 +35,42 @@ namespace
...
     return ptr;
 }
 
 //---------------------------------------------------------------------------//
 //! Free asynchronously for CUDA and newer HIP versions
 void free_async_impl(void* ptr, Stream::StreamT s)
 {
+    if (Stream::async())
+    {
 #if CELER_STREAM_SUPPORTS_ASYNC
```
esseivaju, line 62: `Stream::async` already checks `CELER_STREAM_SUPPORTS_ASYNC`so I don't think it needs to be rechecked here.

src/corecel/sys/Stream.cc
```diff
@@ -34,27 +35,42 @@ namespace
 void* malloc_async_impl(std::size_t bytes, Stream::StreamT s)
 {
     void* ptr{};
+    if (Stream::async())
+    {
 #if CELER_STREAM_SUPPORTS_ASYNC
```
esseivaju, line 40: same comment as in `free_async_impl`

esseivaju (commented):
Looks good! Just a few comments

src/corecel/sys/Environment.cc
sethrj: Done for this file, added to https://github.com/celeritas-project/celeritas/issues/1263 for the rest of the codebase.

src/corecel/sys/Environment.cc
sethrj: Actually the other one *does* insert it... that way we export what environment variables "mattered" during the execution, and we can cache the results with `static const`.  I added documentation to this effect.

src/corecel/sys/Stream.cc
sethrj: It's necessary to disable it for HIP < 5.2 right? The API call isn't available so this would error out.

esseivaju: Ah right it's still needed to prevent compile error.

esseivaju (approved):

PR #257, opened 2021-06-21
Add MuBremsstrahlungInteractor

DoaaDeeb:
This commit includes the following new files:

src/physics/em/BetheBlochModel.cc
src/physics/em/BetheBlochModel.hh
src/physics/em/detail/BetheBloch.cu
src/physics/em/detail/BetheBloch.hh
src/physics/em/detail/BetheBlochInteractor.hh
src/physics/em/detail/BetheBlochInteractor.i.hh
test/physics/em/BetheBloch.test.cc

and it updates the following files:

src/CMakeLists.txt
test/CMakeLists.txt

PR #1319, opened 2024-07-11
Fix muon bremsstrahlung energy limits and sampling 

amandalund:
This fixes a few things in the muon bremsstrahlung interactor:
 - Remove erroneous min/max model energy limits
 - Add cutoff energy to interactor
 - Add missing minimum cutoff energy limit
 - Fix sampling of the gamma energy

I also slightly modified the brems final state helper so we can use it for this interactor too.


src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
```diff
@@ -98,16 +121,23 @@ CELER_FUNCTION MuBremsstrahlungInteractor::MuBremsstrahlungInteractor(
...
+                    value_as<Energy>(cutoffs.energy(shared.gamma))))
 {
-    CELER_EXPECT(particle_.energy() >= shared_.min_incident_energy()
-                 && particle_.energy() <= shared_.max_incident_energy());
-    CELER_EXPECT(particle.particle_id() == shared_.ids.mu_minus
-                 || particle.particle_id() == shared_.ids.mu_plus);
+    CELER_EXPECT(particle.particle_id() == shared_.mu_minus
+                 || particle.particle_id() == shared_.mu_plus);
+    CELER_EXPECT(cutoffs.energy(shared.gamma) >= min_cutoff_energy());
+    CELER_EXPECT(particle_.energy() > cutoffs.energy(shared.gamma));
```
whokion, line 135: These conditions are reasonable and practically should work.  Nevertheless, these may be too strict as 1) the `cutoffs` is material-dependent (and also a user may set it to very low which can be lower than the magic threshold 0.9keV), 2) the valid energy range of this model (down to zero) is independent to the gamma cutoffs - do we have a logical protection that this interactor will not be chosen as a discrete process during the process competition or is excluded by the continuous energy loss action (by the stopping range)? - just wonder whether it is fully guaranteed.

test/celeritas/em/MuBremsstrahlung.test.cc
```diff
@@ -115,12 +139,12 @@ TEST_F(MuBremsstrahlungInteractorTest, basic)
...
-    double const expected_energy[] = {
-        1012.99606184083, 1029.80705246907, 1010.52595539471, 1010.77666768483};
-    double const expected_costheta[] = {0.968418002240112,
-                                        0.999212413725981,
-                                        0.998550042495312,
-                                        0.983614606590488};
+    static double const expected_energy[] = {0.0065836962047077,
+                                             0.072737465588407,
+                                             0.0046101810148469,
+                                             0.0047801585825437};
```
whokion: Nice fix! Good to see reasonable brem energy here. Now, these seems to be relatively small (expecting that the mean brem energy is roughly a couple of percent of the incident muon kinetic energy. It could be simply a statistical effect (i.e., following a steep falling spectrum) as only 4 interactions are sampled here or the gamma cutoff (1keV here) may be too small.  We will do a separate physics validation anyway, but can you sample something like 10 interactions here to see whether there is any brem energy above MeV?

test/celeritas/em/MuBremsstrahlung.test.cc
```diff
@@ -161,41 +189,82 @@ TEST_F(MuBremsstrahlungInteractorTest, stress_test)
...
+                                                         9.0945,
+                                                         10.259,
+                                                         8.5959,
+                                                         8.1658,
+                                                         8.1053};
+    static double const expected_avg_energy[] = {0.0038567895491616,
+                                                 0.019186468482337,
+                                                 0.098553972746449,
+                                                 1.6852671846735,
+                                                 229.53138001979,
```
whokion: Okay, this average number seems much more reasonable!  A good additional test may be a comparison of the transverse momentum along the line of flight between gamma and muon after interaction, which should be same. This is just for a note, so it is okay without the test as we will validate angular distribution too.

whokion (approved):
Nice fix and update! Just a couple of comments for future physics validations - may be it is a good time to add/run @stognini's validation suite for muon processes too.

sethrj:
Follow-on to #257 and #1309 :)

src/celeritas/em/interactor/CombinedBremInteractor.hh
```diff
@@ -158,18 +162,14 @@ CELER_FUNCTION Interaction CombinedBremInteractor::operator()(Engine& rng)
...
-
     // Update kinematics of the final state and return this interaction
-    return sample_interaction(rng);
+    return detail::BremFinalStateHelper(inc_energy_,
+                                        inc_direction_,
+                                        inc_momentum_,
+                                        shared_.rb_data.ids.gamma,
+                                        gamma_energy,
+                                        sample_costheta_(rng),
+                                        secondaries)(rng);
```
sethrj, line 172: This is fine, but we want to be very careful not to ever do `h(f(rng), g(rng))` because that results in unspecified behavior (as to which argument gets the first invocation of the RNG).

src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
```diff
@@ -177,14 +193,12 @@ CELER_FUNCTION real_type MuBremsstrahlungInteractor::sample_cos_theta(
 
 //---------------------------------------------------------------------------//
 
-CELER_FUNCTION real_type MuBremsstrahlungInteractor::differential_cross_section(
-    real_type gamma_energy) const
+CELER_FUNCTION real_type
+MuBremsstrahlungInteractor::calc_differential_xs(real_type gamma_energy) const
```
sethrj, line 197: This should probably be moved to a standalone helper class since we will likely need it one day for generating our own cross sections. It would also be good for testing...

src/celeritas/em/interactor/SeltzerBergerInteractor.hh
```diff
@@ -149,17 +153,14 @@ CELER_FUNCTION Interaction SeltzerBergerInteractor::operator()(Engine& rng)
...
-                                                    secondaries);
-
     // Update kinematics of the final state and return this interaction
-    return sample_interaction(rng);
+    return detail::BremFinalStateHelper{inc_energy_,
+                                        inc_direction_,
+                                        inc_momentum_,
+                                        shared_.ids.gamma,
+                                        sample_photon_energy_(rng),
+                                        sample_costheta_(rng),
```
sethrj, line 162: Ooh this can't be done reproducibly. There was once a hard to track discrepancy in our CI between intel/gcc because intel would sample one function first and GCC would sample the other.

test/celeritas/em/MuBremsstrahlung.test.cc
```diff
@@ -161,41 +189,82 @@ TEST_F(MuBremsstrahlungInteractorTest, stress_test)
...
+                                                         8.6038,
+                                                         8.1621,
+                                                         8.1053,
+                                                         8.1199,
+                                                         8.4714,
+                                                         9.0945,
+                                                         10.259,
+                                                         8.5959,
+                                                         8.1658,
+                                                         8.1053};
```
sethrj: If you add a comma to the end of the list, it'll undo the ridiculous indentation and just indent it one tab...

sethrj (commented):
Only error I see is the subtle one about the RNG; please feel free to merge once that's fixed!

src/celeritas/em/interactor/SeltzerBergerInteractor.hh
amandalund: Ah, I thought because I used brace initialization here they would always be sampled in the order they appear?

test/celeritas/em/MuBremsstrahlung.test.cc
amandalund: Nice trick :)

src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
amandalund: Already started this, actually!

test/celeritas/em/MuBremsstrahlung.test.cc
amandalund: I had the same concern, and when I increased the number of samples to see if it was a statistical effect I did see some larger secondary energies (above 1 MeV). I added the average energy check in the stress test with many more samples so we have an idea what the average secondary energy is for different incident energies.

src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
amandalund: I _think_ the cutoff value will be set to the max of the cutoff and this magic threshold (see [here](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/utils/src/G4EmModelManager.cc#L521) and [here](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/muons/src/G4MuBremsstrahlungModel.cc#L109), but I've only skimmed the code). If not, this check should catch it (and we should modify the interactor to not sample an interaction below the threshold). Because the process cross section is zero below the cutoff, we should be guaranteed to never sample this interaction if the incident energy is less than the cutoff.

src/celeritas/em/interactor/SeltzerBergerInteractor.hh
sethrj: I think that's only true for initializer lists, not constructor arguments... could be wrong though!

amandalund: From the standard:
> 4   Within the initializer-list of a braced-init-list, the initializer-clauses, including any that result from pack
    expansions (14.5.3), are evaluated in the order in which they appear. That is, every value computation and
    side effect associated with a given initializer-clause is sequenced before every value computation and side
    effect associated with any initializer-clause that follows it in the comma-separated list of the initializer-list.
    [ Note: This evaluation ordering holds regardless of the semantics of the initialization; for example, it applies
    when the elements of the initializer-list are interpreted as arguments of a constructor call, even though
    ordinarily there are no sequencing constraints on the arguments of a call. -- end note ]

So from the note it sounds like it should be true (I initially did make this mistake, but caught it when the SB and RB tests started failing).

sethrj: Wow, that's a deep cut. Good find! Hopefully the compilers all follow this standard 😅

sethrj (approved):

Issue #1312, opened 2024-07-09
Optimize track vector data layout for particle types


sethrj:
I have another optimization idea related to sorting but maybe less intrusive, along the lines of how AdePT has separate state vectors for each particle type. To do something similar, what about we instantiate charged and neutral particles at opposite ends of the state vector? This will naturally partition them to some extent (though we'd need to consider cases like annihilation and pair production, where the track slot basically changes type, and how to handle when the two stacks meet in the middle). We don't have to worry about "mixing" breaking anything due to how we launch kernels.

@amandalund suggested sorting the last `num_vacancies` track initializers by particle type before initializing which should be equivalent.

I think there has to be an additional toggle between "filling" and "filled": that suggestion would work during the "filled" stage but we'd still have to push to opposite ends of the vector while filling.

See related work from AdePT: https://cds.cern.ch/record/2856531

PR #1322, opened 2024-07-14
Optimize track vector data layout for particle types

amandalund:
I started sketching out a possible implementation of #1312 (still a WIP). This partitiions the last `num_new_tracks` initializers by charged/neutral and initializes neutral and charged tracks at opposite ends of the track vector.

For now I've disabled the immediate initialization of. a secondary in the parent's track slot (since that results in quite a bit more mixing) and the copying of the parent's geometry state (allowing this would require partitioning the parent IDs as well, which we could do). FWIW I didn't see a ton of speedup from those geometry optimizations anyway (maybe a few percent at most, in cms2018).


test/celeritas/track/TrackSort.test.cc
```diff
@@ -450,5 +521,167 @@ TEST_F(TestActionCountEm3StepperD, TEST_IF_CELER_DEVICE(device_count_actions))
...
+            "NNNNNNNNNNNNNNNNNNNNNNNN_N_________________CCCCCCC_C_CCCCCCCCCCC",
+            "NNNNNNNNNN_NNNNNN_NNNNNNNN_NNN_NN_NNN______CCCCCCCCCCC__CCCCCCCC",
+            "NNNNNN_NNN_N__NNNNNNNNNNNNNNNNNN_NNNN_N____CCCCCCCCC_CCCCCCCCCCC",
+            "NNNNNNNNNNNNN_NNNNNNNNNNNNNNNNNN_NNNN__NNNCCCCCCCC_CCCCCCCCCCCCC",
+            "NNNNNNNNNNN_NNNNN_NNNNNNNNNNNNNNNNNNN_NNNN_CCCCCCCNC_CCCCCCCCCCC",
+            "NNNNNNNNNNNNNNNNNN_NNNNNNNNN_NN_NNN_NNNNNNNCCCCCC_N_NC_CC_CCCCCC",
+            "NNNNNNNNNNNNNNNN_NNNNNNNNNNNNNN_NNNNNNNNNNNCCCCCC_NNNCNCCCCCCCCC",
+            "NN_N__NNNNN_NNNNNNNNNNNNNNNNNNNNNN_NNNNNNNNCCC_CCCNN_CN_C_CCCCCC",
+            "NNNNNNNNNN__NNNNN_NNNNNN_NNNNN_NNNNNNNNNNNNCCCNC_CNNNCNN__CCCCCC",
+            "NNNNNNNNNNN_NN_NN__NNNNNNNNNNNNNNNNNNNNNNNNCCCNCNC_NNCNNN_CCCCCC",
```
sethrj, line 631: Super cool!

sethrj:
I might test this today... also the test segfaults are caught by clang asan:
```
[ RUN      ] PartitionDataTest.init_primaries_host
=================================================================
==2683==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7f277607cc58 at pc 0x00000038c3b0 bp 0x7ffc76ad7e70 sp 0x7ffc76ad7e68
READ of size 8 at 0x7f277607cc58 thread T0
```

amandalund:
@sethrj @esseivaju I think this is ready for a look and some testing/profiling now. I haven't done much myself yet, but in some quick initial comparisons saw ~25% improvement on testem3 and ~5% on cms2018 with celer-sim. With `merge_events` off I saw only a 15% improvement with testem3 and a performance degradation with cms; I haven't looked at the performance of any of the Geant4-integrated apps yet.

sethrj:
throughput
Seems to be barely any bump in performance for Frontier/AMD. @esseivaju if you run the `v0.5.0-dev.partitioned` branch I'll generate the equivalent plot for perlmutter for the whole run of problems.

src/celeritas/track/InitializeTracksAction.cc
```diff
@@ -61,6 +63,17 @@ void InitializeTracksAction::execute_impl(CoreParams const& core_params,
         = std::min(counters.num_vacancies, counters.num_initializers);
     if (num_new_tracks > 0 || core_state.warming_up())
     {
+        if (core_params.init()->host_ref().track_order
+            == TrackOrder::partition_charge)
```
sethrj: Perhaps add a `track_order()` accessor to the `TrackInitParams`?

esseivaju:
I'll try to run this tomorrow

sethrj:
rel-throughput

Copying the output from the slack channel. We're obviously not yet copying the track sort parameter into the GPU setup for Celeritas. It's very odd that the CPU "sort" performance dips in a couple of cases.

src/celeritas/track/detail/ProcessSecondariesExecutor.hh
```diff
@@ -123,7 +123,8 @@ ProcessSecondariesExecutor::operator()(TrackSlotId tid) const
             ti.particle.energy = secondary.energy;
             CELER_ASSERT(ti);
 
-            if (!initialized && sim.status() != TrackStatus::alive)
+            if (!initialized && sim.status() != TrackStatus::alive
+                && params->init.track_order != TrackOrder::partition_charge)
```
sethrj, line 131: It would be good to update the comment to note that we're skipping in-place initialization to do a better job of partitioning the tracks. (And in fact, if we end up adding decay or something where we can kill a charged particle and add a new one in place, we might want to update this to `&& old.charged() != new.charged()`)

sethrj (approved):
Superb work @amandalund ! We really need to publish some of this 😅

sethrj:
@amandalund I think this is good to go aside from the minor comments and the conflict 😄 I know it's probably been lost in the pile of pull requests 😅

amandalund:
Thanks @sethrj! I'd modified this to try to reuse the parent's geo state when possible for improved performance, but got sidetracked when I added an assertion that the parent's position match the initializer position and several unit tests started failing. Should be able to wrap this up pretty quickly now that those fixes are in.

amandalund:
Alright, I've updated this so we're now partitioning an array of indices (rather than track initializers) which we can also use to access the parent track slot IDs to copy the geometry state. We'll still have to reinitialize the geometry more often than we would when not sorting (we're not doing the in-place initialization, so we can't reuse the state if the track was killed but produced secondaries), but hopefully it should still help some. @esseivaju if you wouldn't mind rerunning the regression problems with the latest changes and @sethrj updating the performance plot again I'd appreciate it!

sethrj:
Rerunning frontier now...

src/celeritas/random/distribution/RejectionSampler.hh
```diff
@@ -62,7 +62,9 @@ class RejectionSampler
 
     //! Construct when the distribution's maximum is normalized
     explicit CELER_FUNCTION RejectionSampler(real_type f)
-        : RejectionSampler{f, 1} {}
+        : RejectionSampler{f, 1}
+    {
+    }
```
sethrj: Looks like someone *ahem* 😳 forgot to clang-format their PR and ended up making everyone else's PRs harder. Perhaps I should run a big clang-format now?

sethrj:
rel-throughput
Frontier's seeing a solid 10% speedup with this 😄  nice work!

sethrj:
rel-throughput
Perlmutter's also seeing similar speedups for testem3. The Run 2 geometry minus MSC shows that same anomaly...

sethrj:
@esseivaju did you get a chance to re-run develop for the CMS/run2/no-msc case? It would be good to get this in but I'd like to understand the performance regression first.

amandalund:
I also ran it actually (on an A100 + AMD EPYC 7532):
rel-throughput

sethrj:
Weird...

amandalund:
It's possible the worse performance for cms+msc could still be the penalty of having to reinitialize the geometry state more often than on develop...

sethrj:
That makes a lot of sense, good thought @amandalund . The cost of a step iteration without MSC or field is relatively low compared to *with MSC and field*, and the cost of initialization with vecgeom/CMS is much higher than testem3. We could potentially do a better job of serializing the geometry state for fast reconstruction by storing an additional "unique volume ID" (aka VecGeom navindex, or expanded ORANGE volume ID) on the track and defining a new initializer that reconstructs the logical geometry state more efficiently than searching based on the point.

amandalund:
Yeah it definitely seems worthwhile to explore ways of speeding up that initialization.

amandalund:
Sorry @sethrj I only just realized that M means !MSC 😅. Maybe we should invert that?

sethrj:
There's supposed to be a tilde on top; matplotlib didn't render it correctly. Since the results aren't correct without msc I figured it was better to start denoting "not included" rather than "accurate"...

amandalund:
It does seem better to better to mark that some problems are missing key physics than the other way around... though on the other hand $F$ for field and $M$ for msc is much more obvious when glancing at a plot than $F$ for field and $\tilde{M}$ for no msc.

sethrj:
I'm cool with changing this. Maybe better (if we want to go crazy) would be to replace the "no MSC" mode with full coulomb scattering? I have no idea if that'll increase our run times by 1000x though...

PR #1324, opened 2024-07-16
Add Regula Falsi root solver

VHLM2001:
Implements non linear root solver for general calculations of non-trivial roots, such as the intersection of an involute and a line.
Method applied here is the Regula Falsi methodology described by the following relation:

$$ c_k= \frac{a_kf(b_k)-b_kf(a_k)}{f(b_k)-f(a_k) }$$

Where $`f(t)`$ is the root function of interest, $`a_k`$ is the left bound, $`b_k`$ is the right bound, $`c_k`$ is the approximated root.  $`c_k`$ is used to update the bonds based on the sign $`f(c_k)`$ and whether it matches the sign of $`f(a_k)`$ or $`f(b_k)`$.  Performing this update allows for the iteration on the root, using the convergence criteria based on $`f(c_k)`$ proximity to $`0`$.


src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+namespace celeritas
+{
+template<class F>
+class RegulaFalsi
+{
+  public:
+    // Contructpr of Regula Falsi
+    inline CELER_FUNCTION RegulaFalsi(F&& func, real_type tol);
+
+    // Solve between
```
sethrj:
```suggestion
    // Solve for a root between two points
```

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+    real_type operator()(real_type left, real_type right);
+
+  private:
+    F&& func_;
```
sethrj:
```suggestion
    F func_;
```

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+// TEMPLATE DEDUCTION
+//---------------------------------------------------------------------------//
+
+template<class F, class... Args>
+RegulaFalsi(F&&, Args...) -> RegulaFalsi<F>;
+
+//---------------------------------------------------------------------------//
```
sethrj:
```suggestion
//---------------------------------------------------------------------------//
// TEMPLATE DEDUCTION
//---------------------------------------------------------------------------//

template<class F, class... Args>
RegulaFalsi(F&&, Args...) -> RegulaFalsi<F>;

//---------------------------------------------------------------------------//
// INLINE DEFINITIONS
//---------------------------------------------------------------------------//
```

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+/*!
+ * Construct from function.
+ */
+template<class F>
+CELER_FUNCTION RegulaFalsi<F>::RegulaFalsi(F&& func, real_type tol)
+    : func_{func}, tol_{tol}
```
sethrj: the `F&&` templating stuff is really weird so it's not surprising you didn't get it right on the first try: it took me probably 2 years to figure it all out even after 10 years of C++ experience 😆
```suggestion
CELER_FUNCTION RegulaFalsi<F>::RegulaFalsi(F&& func, real_type tol)
    : func_{celeritas::forward<F>(func)}, tol_{tol}
```

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+    while (std::fabs(f_root) > tol_)
+    {
+        // Calcuate root
+        root = (left * f_right - right * f_left) / (f_right - f_left);
+
+        // Root function value of root
+        f_root = func_(root);
+
+        // Update bounds with iterated root
+        if ((0 < f_left) - (f_left < 0) == (0 < f_root) - (f_root < 0))
```
sethrj: What exactly are you trying to say here?

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from function.
+ */
+template<class F>
+CELER_FUNCTION RegulaFalsi<F>::RegulaFalsi(F&& func, real_type tol)
+    : func_{func}, tol_{tol}
+{
+    CELER_EXPECT(tol > 0);
```
sethrj:
```suggestion
    CELER_EXPECT(tol > 0 && tol < 1);
```

sethrj (commented):
@VHLM2001 The timeout was in the Regula Falsi solver (more than 15 seconds) in *single-precision* mode, which likely means your solver failed to converge. (There's also no termination critera for  iteration count which is typical for this kind of solver to prevent infinite loops.)

test/corecel/math/RegulaFalsiRootFinder.test.cc
```diff
@@ -0,0 +1,86 @@
...
+
+namespace test
+{
+
+// Solve: (x-2)(x+2) = 0
+TEST(RegulaFalsi, root_two)
+{
+    auto root = [](real_type t) { return (t - 2) * (t + 2); };
+
+    RegulaFalsi find_root{root, 1e-13};
```
sethrj: These tolerances are too tight for a single-precision solve... perhaps include the `corecel/math/SoftEqual.hh` header and use `constexpr real_type tol = SoftEqual<>{}.rel();`?

src/corecel/math/RegulaFalsiRootFinder.hh
sethrj: Actually the tolerance isn't relative so perhaps it should just be `> 0` after all

test/corecel/math/RegulaFalsiRootFinder.test.cc
```diff
@@ -0,0 +1,86 @@
...
+#include "celeritas_test.hh"
+
+namespace celeritas
+{
+using constants::pi;
+
+namespace test
+{
+
+// Solve: (x-2)(x+2) = 0
```
sethrj, line 27: Great tests by the way!

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,110 @@
...
+ * parameters defined in \em params.
+ */
+template<class F>
+CELER_FUNCTION real_type RegulaFalsi<F>::operator()(real_type left,
+                                                    real_type right)
+{
+    // Initialize Iteration parameters
+    real_type f_left = func_(left);
+    real_type f_right = func_(right);
+    real_type f_root = 1;
```
sethrj, line 88: Since `f(x) can be unbounded`, then perhaps either the loop should be `do {} while ();` or initialized to `numeric_limits<real_type>::infinity()` (`corecel/math/NumericLimits.hh` for the device-compatible version).

sethrj (commented):
Sorry I hit enter too soon!

src/corecel/math/RegulaFalsiRootFinder.hh
VHLM2001: The check is comparing the sign of the root function at the left bound and at the approximated root.

VHLM2001: Does "Update the bound which produces the same sign as the root" clarify what I meant?

sethrj: aha, so here's what you might do then: define a little helper function `CELER_FORCEINLINE int signum(real_type x) { return (0 < a) - (a < 0); }` in `Algorithms.hh`, add the comment as you suggest, and write:
```suggestion
        if (signum(f_left) == signum(f_root))
        {
            // Left bound has the same sign as the current root estimate
```

src/corecel/math/Algorithms.hh
```diff
@@ -602,6 +602,16 @@ CELER_CONSTEXPR_FUNCTION T eumod(T numer, T denom)
     return r;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the sign of a number.
```
sethrj, line 607:
```suggestion
 * Calculate the sign of a number.
 *
 * \return -1 if negative, 0 if exactly zero (or NaN), 1 if positive
```

Nice going adding a test! You're a star.

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+ * Perform a Regula Falsi Iteration given a root function \em func and
+ * toolerance \em tol .
+ *
+ * Using a \em left and \em right bound a Regula Falsi approximates the \em
+ * root as: \f[ root = (left * func(right) - right * func(left)) / (func(right)
+ * - func(left)) \f]
+ *
+ * Then value of \em func at the root is calculated compared to values of
+ * \em func at the bounds. The bound which has the same sign \em func(root)
+ */
```
sethrj:
```suggestion
//! \file corecel/math/RegulaFalsiRootFinder.hh
```
and move the documentation to the class.

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+#include "corecel/Assert.hh"
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+#include "corecel/math/Algorithms.hh"
+
+#include "detail/MathImpl.hh"
+
+namespace celeritas
+{
+template<class F>
```
sethrj, line 39:
```suggestion
/*! Perform a Regula Falsi Iteration given a root function \em func and
 * toolerance \em tol .
 *
 * Using a \em left and \em right bound a Regula Falsi approximates the \em
 * root as: \f[ root = (left * func(right) - right * func(left)) / (func(right)
 * - func(left)) \f]
 *
 * Then value of \em func at the root is calculated compared to values of
 * \em func at the bounds. The bound which has the same sign \em func(root)
 * FIXME: finish the docs?
 */
 template<class F>
```

I think you also got distracted midsentence like I am wont to do.

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+CELER_FUNCTION RegulaFalsi<F>::RegulaFalsi(F&& func, real_type tol)
+    : func_{celeritas::forward<F>(func)}, tol_{tol}
+{
+    CELER_EXPECT(tol > 0);
+}
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+
```
sethrj:
```suggestion
```

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+            left = root;
+            f_left = f_root;
+        }
+        else
+        {
+            right = root;
+            f_right = f_root;
+        }
+        i++;
+    } while (std::fabs(f_root) > tol_ && i <= 100);
```
sethrj: Instead of a magic value of 100, define a `static constexpr inline int max_iters_ = 100;` in the class. In addition, a trick I learned to make checking the termination criterion slightly more efficient in some circumstances, is to initialize them with the maximum value and then decrement until the value is negative:

```suggestion
    } while (std::fabs(f_root) > tol_ && --remaining_iters > 0);
```

we could also consider changing the loop criterion to an assertion, otherwise you're silently returning an unconverged value....

sethrj (commented):
Sorry this "simple" pull request is so much effort to review. It's not because you've done anything bad or sloppy: you're really doing some outstanding work. It's that we have really high expectations 😅 but I think this is the last round of comments!

src/corecel/math/RegulaFalsiRootFinder.hh
VHLM2001: Do you mean like putting an ensure statement after the loop to make sure the convergence criteria is met?

sethrj (approved):
Awesome, thanks @VHLM2001 for the thorough job! A well-tested fundamental class like this is a necessity for building more complex algorithms like the involute solve because it means you can really understand and catch the edge cases quickly.

PR #1326, opened 2024-07-17
Add cascade collider

whokion:
# Add a helper for intra-nucleus nucleon-nucleon collisions
 - Add TwodGridBuilder and angular distribution data for sampling cos\theta of the intra-nucleus nucleon-nucleon scattering.
 -  Add CascadeCollider and CascadeParticle for sampling the final state of the ntra-nucleus cascade collision and an associated test.
 - Add namespace lorentz for FourVector and a couple of utility functions

### DIFF
--- a/src/celeritas/grid/TwodGridBuilder.hh
+++ b/src/celeritas/grid/TwodGridBuilder.hh
@@ -3,7 +3,7 @@
 // See the top-level COPYRIGHT file for details.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file celeritas/grid/GenericGridBuilder.hh
+//! \file celeritas/grid/TwodGridBuilder.hh
 //---------------------------------------------------------------------------//
 #pragma once
 
--- a/src/celeritas/grid/TwodGridBuilder.hh
+++ b/src/celeritas/grid/TwodGridBuilder.hh
@@ -1,0 +1,58 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/grid/GenericGridBuilder.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "corecel/data/Collection.hh"
+#include "corecel/data/CollectionBuilder.hh"
+#include "corecel/data/DedupeCollectionBuilder.hh"
+#include "corecel/grid/TwodGridData.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a generic 2D grid.
+ *
+ * This uses a deduplicating inserter for real values to improve cacheing.
+ */
+class TwodGridBuilder
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using Items = Collection<T, Ownership::value, MemSpace::host>;
+    using TwodGrid = TwodGridData;
+    using SpanConstFlt = Span<float const>;
+    using SpanConstDbl = Span<double const>;
+    //!@}
+
+  public:
+    // Construct with pointers to data that will be modified
+    explicit TwodGridBuilder(Items<real_type>* reals);
+
+    // Add a 2D grid of generic data with linear interpolation
+    TwodGrid
+    operator()(SpanConstFlt grid_x, SpanConstFlt grid_y, SpanConstFlt values);
+
+    // Add a 2D grid of generic data with linear interpolation
+    TwodGrid
+    operator()(SpanConstDbl grid_x, SpanConstDbl grid_y, SpanConstDbl values);
+
+  private:
+    DedupeCollectionBuilder<real_type> reals_;
+
+    // Insert with floating point conversion if needed
+    template<class T>
+    TwodGrid insert_impl(Span<T const> grid_x,
+                         Span<T const> grid_y,
+                         Span<T const> values);
+};
+
+//---------------------------------------------------------------------------//
+}  // namespace celeritas
--- a/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
+++ b/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
@@ -107,14 +107,14 @@ CascadeCollider::CascadeCollider(NeutronInelasticRef const& shared,
     , sample_phi_(0, 2 * constants::pi)
 {
     // Initialize the boost velocity and momentum in the center of mass frame
-    FourVector sum_vec4 = lorentz::add(bullet_.vec4, target_.vec4);
-    cm_velocity_ = lorentz::boost_vector(sum_vec4);
-    cm_p_ = calc_cm_p(sum_vec4);
+    FourVector sum_four_vec = bullet_.four_vec + target_.four_vec;
+    cm_velocity_ = boost_vector(sum_four_vec);
+    cm_p_ = calc_cm_p(sum_four_vec);
 
     // Calculate the kinetic energy in the target rest frame
-    FourVector bullet_p = bullet_.vec4;
-    lorentz::boost(-lorentz::boost_vector(target_.vec4), &bullet_p);
-    kin_energy_ = bullet_p.energy - lorentz::norm(bullet_p);
+    FourVector bullet_p = bullet_.four_vec;
+    boost(-boost_vector(target_.four_vec), &bullet_p);
+    kin_energy_ = bullet_p.energy - norm(bullet_p);
 }
 
 //---------------------------------------------------------------------------//
@@ -175,10 +175,10 @@ CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
     // Find the final state of outgoing particles
     FinalState result = {bullet_, target_};
 
-    FourVector cm_vec4 = target_.vec4;
-    lorentz::boost(-cm_velocity_, &cm_vec4);
+    FourVector cm_momentum = target_.four_vec;
+    boost(-cm_velocity_, &cm_momentum);
 
-    Real3 cm_dir = make_unit_vector(-cm_vec4.mom);
+    Real3 cm_dir = make_unit_vector(-cm_momentum.mom);
     real_type vel_parallel = dot_product(cm_velocity_, cm_dir);
 
     // Degenerated if velocity perpendicular to the c.m. momentum is small
@@ -192,20 +192,21 @@ CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
         Real3 vxcm = make_unit_vector(cross_product(cm_dir, cm_velocity_));
         if (norm(vscm) > this->epsilon() && norm(vxcm) > this->epsilon())
         {
-            result[0].vec4
+            result[0].four_vec
                 = {{fv.mom[0] * vscm + fv.mom[1] * vxcm + fv.mom[2] * cm_dir},
                    fv.energy};
         }
     }
 
-    result[1].vec4 = {{-result[0].vec4.mom},
-                      std::sqrt(dot_product(mom, mom)
-                                + ipow<2>(value_as<MevMass>(target_.mass)))};
+    result[1].four_vec
+        = {{-result[0].four_vec.mom},
+           std::sqrt(dot_product(mom, mom)
+                     + ipow<2>(value_as<MevMass>(target_.mass)))};
 
     // Convert the final state to the lab frame
     for (auto i : range(2))
     {
-        lorentz::boost(cm_velocity_, &result[i].vec4);
+        boost(cm_velocity_, &result[i].four_vec);
     }
 
     return result;
@@ -219,7 +220,7 @@ CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
 CELER_FUNCTION real_type CascadeCollider::calc_cm_p(FourVector const& v) const
 {
     // The total energy in c.m.
-    real_type m0 = lorentz::norm(v);
+    real_type m0 = norm(v);
 
     real_type m1 = value_as<MevMass>(bullet_.mass);
     real_type m2 = value_as<MevMass>(target_.mass);
--- a/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
+++ b/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
@@ -65,11 +65,11 @@ class CascadeCollider
 
     // Id for intra-nuclear channels
     ChannelId ch_id_;
-    // Booster vector in the center of mass frame
+    // Boost vector in the center of mass frame [1/c]
     Real3 cm_velocity_;
-    // Momentum magnitude in the center of mass frame
+    // Momentum magnitude in the center of mass frame [MeV/c]
     real_type cm_p_;
-    // Kinetic energy in the target rest frame
+    // Kinetic energy in the target rest frame [MeV]
     real_type kin_energy_;
 
     // Sampler
@@ -77,7 +77,7 @@ class CascadeCollider
 
     //// CONSTANTS ////
 
-    //! A criteria for coplanarity in the Lorentz transformation
+    //! A criteria [1/c] for coplanarity in the Lorentz transformation
     static CELER_CONSTEXPR_FUNCTION real_type epsilon()
     {
         return real_type{1e-10};
@@ -134,7 +134,7 @@ CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
 {
     // Sample cos\theta of outgoing particles in the center of mass frame
     real_type cdf = generate_canonical(rng);
-    TwodGridData cdf_grid = shared_.angular_cdf[ch_id_];
+    TwodGridData const& cdf_grid = shared_.angular_cdf[ch_id_];
     Grid energy_grid(cdf_grid.x, shared_.reals);
     real_type cos_theta{0};
 
@@ -181,21 +181,28 @@ CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
     Real3 cm_dir = make_unit_vector(-cm_momentum.mom);
     real_type vel_parallel = dot_product(cm_velocity_, cm_dir);
 
-    // Degenerated if velocity perpendicular to the c.m. momentum is small
-    bool degenerated
-        = (dot_product(cm_velocity_, cm_velocity_) - ipow<2>(vel_parallel)
-           < this->epsilon());
+    Real3 vscm = cm_velocity_;
+    axpy(-vel_parallel, cm_dir, &vscm);
+    vscm = make_unit_vector(vscm);
 
-    if (!degenerated)
+    if (norm(vscm) > this->epsilon())
     {
-        Real3 vscm = make_unit_vector(cm_velocity_ - vel_parallel * cm_dir);
         Real3 vxcm = make_unit_vector(cross_product(cm_dir, cm_velocity_));
-        if (norm(vscm) > this->epsilon() && norm(vxcm) > this->epsilon())
+        if (norm(vxcm) > this->epsilon())
         {
-            result[0].four_vec
-                = {{fv.mom[0] * vscm + fv.mom[1] * vxcm + fv.mom[2] * cm_dir},
-                   fv.energy};
+            for (int i = 0; i < 3; ++i)
+            {
+                result[0].four_vec.mom[i] = fv.mom[0] * vscm[i]
+                                            + fv.mom[1] * vxcm[i]
+                                            + fv.mom[2] * cm_dir[i];
+            }
+            result[0].four_vec.energy = fv.energy;
         }
+    }
+    else
+    {
+        // Degenerated if velocity perpendicular to the c.m. momentum is small
+        result[0].four_vec = fv;
     }
 
     result[1].four_vec
--- a/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
+++ b/src/celeritas/neutron/interactor/detail/CascadeCollider.hh
@@ -109,7 +109,7 @@ CascadeCollider::CascadeCollider(NeutronInelasticRef const& shared,
     // Initialize the boost velocity and momentum in the center of mass frame
     FourVector sum_four_vec = bullet_.four_vec + target_.four_vec;
     cm_velocity_ = boost_vector(sum_four_vec);
-    cm_p_ = calc_cm_p(sum_four_vec);
+    cm_p_ = this->calc_cm_p(sum_four_vec);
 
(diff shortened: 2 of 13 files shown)


src/celeritas/phys/FourVector.hh
```diff
@@ -62,4 +64,26 @@ inline CELER_FUNCTION void boost(Real3 const& v, FourVector* p)
     p->energy = gamma * (p->energy + vp);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Add two four-vectors.
+ */
+inline CELER_FUNCTION FourVector add(FourVector const& a, FourVector const& b)
```
sethrj: Instead of `add` we might as well make it `operator+`, but perhaps first we should do `operator+=`?

src/celeritas/phys/FourVector.hh
```diff
@@ -28,6 +28,8 @@ struct FourVector
     real_type energy{0};  //!< Particle energy
 };
 
+namespace lorentz
+{
```
sethrj: I don't think we need a whole namespace for this... and even if we did the entire file (especially the FourVector) should be in it.

src/celeritas/phys/FourVector.hh
```diff
@@ -62,4 +64,26 @@ inline CELER_FUNCTION void boost(Real3 const& v, FourVector* p)
...
+    result.mom = a.mom + b.mom;
+    result.energy = a.energy + b.energy;
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the magnitude of a four vector.
+ */
+inline CELER_FUNCTION real_type norm(FourVector const& a)
```
sethrj, line 89: This should get a test... I actually don't see any tests for this file at all.

src/celeritas/neutron/model/NeutronInelasticModel.cc
```diff
@@ -65,25 +66,35 @@ NeutronInelasticModel::NeutronInelasticModel(ActionId id,
...
         // Add nucleon-nucleon cross section parameters and data
-        ChannelXsData const& channel_data = this->get_channel_xs(channel_id);
+        ChannelData const& channel_data = this->get_channel_data(channel_id);
         CELER_ASSERT(channel_data.par.slope > 0);
         xs_params.push_back(channel_data.par);
 
         GenericGridBuilder build_grid{&data.reals};
         make_builder(&data.nucleon_xs)
-            .push_back(build_grid(bins, make_span(channel_data.xs)));
+            .push_back(build_grid(xs_energy_bins, make_span(channel_data.xs)));
```
sethrj: To allow deduplication of cross section grids etc., the `build_grid` (and the other builders) should be outside the `for` loop. I'd also replace `make_builder(&data.nucleon_xs)` inside the loop with `CollectionBuilder nucleon_xs{&data.nucleon_xs};` outside of it. (The `make_builder` command was before we had C++17 CTAD.)

src/celeritas/neutron/model/NeutronInelasticModel.hh
```diff
@@ -74,18 +74,21 @@ class NeutronInelasticModel final : public Model, public ConcreteAction
...
     using HostXsData = HostVal<NeutronInelasticData>;
-    using ChannelArray = Array<double, 13>;
 
-    struct ChannelXsData
+    struct ChannelData
     {
         StepanovParameters par;
-        ChannelArray xs;
+        Array<double, 13> xs;
+        Array<double, 19 * 6> cdf;  //! c.d.f[cos_theta_bins][cdf_energy_bins]
```
sethrj: Since these are fixed dimensions it would be more sensible to have:
```c++
using ArrayEnergy = Array<double, 6>;
Array<ArrayEnergy, 19> cdf; // [angle][energy]
```

Also since the constants each have only 4 digits of precision we could probably use float... or store even a 16-bit integer if we first multiply them by 10000 😆 it's fine to keep them as double though.

src/celeritas/neutron/interactor/detail/CascadeParticle.hh
```diff
@@ -0,0 +1,39 @@
...
+        unknown,
+        proton,
+        neutron,
+        gamma
+    };
+
+    ParticleType type{ParticleType::unknown};  //!< Particle type
+    units::MevMass mass;  // !< Particle mass
+    FourVector vec4;  //!< Four momentum in natural MevMomentum and
+                      //!< MevEnergy units
```
sethrj: I'm not a big fan of the combined text+number; could it just be "momentum"?

whokion:
@sethrj Thanks a lot for all review comments. Will follow up later today!  Regarding to vec4 --> momentum, I will think an alternative as  a component of FourVector is `mom (three vector)`, which is confusing with the proposed momentum (four vector).  Maybe changing FourVector(mom, energy) --> (p, e) and then vec4 --> momentum or mom. Welcome to the 4-momentum world^^.

src/celeritas/phys/FourVector.hh
whokion: Okay, I added operator+ and the assignment operator+=.  I also realize that it may be better to template FourVector with the precision type, which I may do in another MR.  I did not add other assignment/arithmetic operators as they are not likely used.

src/celeritas/phys/FourVector.hh
whokion: Removed - I though that it might help for clarify.

src/celeritas/phys/FourVector.hh
whokion: I added a minimal basic test.  Other tests may be also added as need later.

src/celeritas/neutron/model/NeutronInelasticModel.cc
whokion: Good to know!  Updated as suggested. Thanks.

src/celeritas/neutron/model/NeutronInelasticModel.hh
whokion: Okay, I keep them as they are for now.

src/celeritas/neutron/interactor/detail/CascadeParticle.hh
whokion: Change vec4 to four_vec.

src/celeritas/grid/TwodGridBuilder.hh
```diff
@@ -0,0 +1,58 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/grid/GenericGridBuilder.hh
```
amandalund:
```suggestion
//! \file celeritas/grid/TwoGridBuilder.hh
```

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+        boost(cm_velocity_, &result[i].four_vec);
+    }
+
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the momentum magnitude of outgoing particles in the center of mass
+ * (c.m.) frame. See Geant4's 4G4VHadDecayAlgorithm::TwoBodyMomentum method.
```
amandalund:
```suggestion
 * (c.m.) frame. See Geant4's G4VHadDecayAlgorithm::TwoBodyMomentum method.
```

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+ */
+CELER_FUNCTION real_type CascadeCollider::calc_cm_p(FourVector const& v) const
+{
+    // The total energy in c.m.
+    real_type m0 = norm(v);
+
+    real_type m1 = value_as<MevMass>(bullet_.mass);
+    real_type m2 = value_as<MevMass>(target_.mass);
+
+    real_type pc_squre = diffsq(m0, m1 - m2) * diffsq(m0, m1 + m2);
```
amandalund:
```suggestion
    real_type pc_sq = diffsq(m0, m1 - m2) * diffsq(m0, m1 + m2);
```
?

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+    , bullet_(bullet)
+    , target_(target)
+    , ch_id_(ChannelId{
+          static_cast<size_type>((bullet_.type == target_.type) ? 0 : 1)})
+    , sample_phi_(0, 2 * constants::pi)
+{
+    // Initialize the boost velocity and momentum in the center of mass frame
+    FourVector sum_four_vec = bullet_.four_vec + target_.four_vec;
+    cm_velocity_ = boost_vector(sum_four_vec);
+    cm_p_ = calc_cm_p(sum_four_vec);
```
amandalund:
```suggestion
    cm_p_ = this->calc_cm_p(sum_four_vec);
```

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+    kin_energy_ = bullet_p.energy - norm(bullet_p);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the final state of outgoing particles from the two-body intra-nucleus
+ * nucleon-nucleon collision in the center of mass (c.m.) frame and return
+ * them after converting momentum to the lab frame. This performs the same
+ * sampling routine as in Geant4's G4ElementaryParticleCollider, mainly
+ * implemented in collider and generateSCMfinalState methods. The cos\theta
```
amandalund:
```suggestion
 * implemented in collide and generateSCMfinalState methods. The cos\theta
```

amandalund (approved):
Nice addition @whokion! Thanks! Just a couple of tiny documentation fixes.

whokion:
@amandalund Thanks for the review and suggestions!

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+        real_type cdf_lower = 1;
+
+        do
+        {
+            cdf_upper = cdf_lower;
+            cdf_lower = calc_cdf({kin_energy_, cos_grid[idx]});
+        } while (cdf_lower > cdf && idx-- > 0);
+
+        real_type frac = (cdf - cdf_lower) / (cdf_upper - cdf_lower);
+        cos_theta = fma(frac, cos_grid[idx + 1] - cos_grid[idx], cos_grid[idx]);
```
sethrj, line 158: Even though this algorithm is only ten lines, it's a little complicated, I think we'll see it a lot in tabulated sampling,  and it could use some extra assertions/testing—after this merge request I'll create a helper function that encapsulates it.

src/celeritas/neutron/model/NeutronInelasticModel.hh
sethrj: @whokion Forgetting about the precision type: the documentation is wrong since the array is accessed and built with `{energy, angle}`
```suggestion
        Array<double, 6 * 19> cdf;  //! [energy][angle]
```

src/celeritas/phys/FourVector.hh
```diff
@@ -26,10 +27,29 @@ struct FourVector
...
 // INLINE UTILITY FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Add two four-vectors.
+ */
+inline CELER_FUNCTION FourVector operator+(FourVector const& lhs,
+                                           FourVector const& rhs)
+{
+    FourVector result = lhs;
+    return std::move(result += rhs);
```
sethrj: This is a "pessimization" (and I'm surprised you don't see a warning) : the move isn't necessary and prevents return-value optimization
```suggestion
    return result += rhs;
```

test/celeritas/phys/FourVector.test.cc
```diff
@@ -0,0 +1,53 @@
...
+
+    // Test norm
+    EXPECT_DOUBLE_EQ(std::sqrt(ipow<2>(c.energy) - dot_product(c.mom, c.mom)),
+                     norm(c));
+    EXPECT_SOFT_EQ(2999.9741665554388, norm(c));
+
+    // Test boost_vector
+    Real3 expected_boost_vector = {0.001, 0.002, 0.003};
+    EXPECT_EQ(a.mom / a.energy, boost_vector(a));
+    EXPECT_EQ(expected_boost_vector, boost_vector(a));
```
sethrj:
```suggestion
    EXPECT_VEC_SOFT_EQ(a.mom / a.energy, boost_vector(a));
    EXPECT_VEC_SOFT_EQ(expected_boost_vector, boost_vector(a));
```
probably?

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+ * distribution function (c.d.f) data in the kinetic energy and cos\theta bins
+ * which are implemented in G4CascadeFinalStateAlgorithm::GenerateTwoBody and
+ * G4NumIntTwoBodyAngDst::GetCosTheta methods.
+ */
+template<class Engine>
+CELER_FUNCTION auto CascadeCollider::operator()(Engine& rng) -> FinalState
+{
+    // Sample cos\theta of outgoing particles in the center of mass frame
+    real_type cdf = generate_canonical(rng);
+    TwodGridData cdf_grid = shared_.angular_cdf[ch_id_];
```
sethrj:
```suggestion
    TwodGridData const& cdf_grid = shared_.angular_cdf[ch_id_];
```

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+    Grid energy_grid(cdf_grid.x, shared_.reals);
+    real_type cos_theta{0};
+
+    if (kin_energy_ < energy_grid.back())
+    {
+        // Find cos\theta from tabulated angular data for a given c.d.f.
+        Grid cos_grid(cdf_grid.y, shared_.reals);
+        TwodGridCalculator calc_cdf(cdf_grid, shared_.reals);
```
sethrj, line 145: I think the 2D calculator may be a bit overkill here. All the CDFs are being evaluated exactly on the cosine grid point but interpolated in energy. So I think could consider building a specialization of the `TwodSubgridCalculator` that evaluates on y *indices* rather than y *values*. This is an optimization but would improve the semantics of the grid calculation.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+
+    if (!degenerated)
+    {
+        Real3 vscm = make_unit_vector(cm_velocity_ - vel_parallel * cm_dir);
+        Real3 vxcm = make_unit_vector(cross_product(cm_dir, cm_velocity_));
+        if (norm(vscm) > this->epsilon() && norm(vxcm) > this->epsilon())
+        {
+            result[0].four_vec
+                = {{fv.mom[0] * vscm + fv.mom[1] * vxcm + fv.mom[2] * cm_dir},
+                   fv.energy};
```
sethrj: This is exactly why I didn't want to make "array operators": even though this reads like a simple algebra expression, C++ is creating an insane number of temporaries here. I created [some test kernels](https://github.com/celeritas-project/cuda-test-snippets/tree/main/celer-vec-add) replicating this operation: as written, this takes 35 registers and 280 (!) bytes of stack space. Rewriting inside a `for` loop reduces the usage to 32 registers and 88 bytes of stack space, or even less (depending on how it's written).

```suggestion
                for (int i = 0; i < 3; ++i)
                {
                    result[0].four_vec[i] = fv.mom[0] * vscm[i]
                                            + fv.mom[1] * vxcm[i]
                                            + fv.mom[2] * cm_dir[i];
                }
                result[0].four_vec.energy = fv.energy;
```

Remember that 200 bytes of overflow per thread is 200MB of memory for the expected use case.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
```
sethrj, line 201: If it's not degenerate, `result[0]` is zero?

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+    real_type vel_parallel = dot_product(cm_velocity_, cm_dir);
+
+    // Degenerated if velocity perpendicular to the c.m. momentum is small
+    bool degenerated
+        = (dot_product(cm_velocity_, cm_velocity_) - ipow<2>(vel_parallel)
+           < this->epsilon());
+
+    if (!degenerated)
+    {
+        Real3 vscm = make_unit_vector(cm_velocity_ - vel_parallel * cm_dir);
```
sethrj: This should use an `axpy` intead of array operators.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+    CascadeParticle const& target_;
+
+    // Id for intra-nuclear channels
+    ChannelId ch_id_;
+    // Booster vector in the center of mass frame
+    Real3 cm_velocity_;
+    // Momentum magnitude in the center of mass frame
+    real_type cm_p_;
+    // Kinetic energy in the target rest frame
+    real_type kin_energy_;
```
sethrj:
```suggestion
    // Boost vector in the center of mass frame [1/c]
    Real3 cm_velocity_;
    // Momentum magnitude in the center of mass frame [MeV / c]
    real_type cm_p_;
    // Kinetic energy in the target rest frame [MeV]
    real_type kin_energy_;
```

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
```diff
@@ -0,0 +1,235 @@
...
+
+    Real3 cm_dir = make_unit_vector(-cm_momentum.mom);
+    real_type vel_parallel = dot_product(cm_velocity_, cm_dir);
+
+    // Degenerated if velocity perpendicular to the c.m. momentum is small
+    bool degenerated
+        = (dot_product(cm_velocity_, cm_velocity_) - ipow<2>(vel_parallel)
+           < this->epsilon());
+
+    if (!degenerated)
```
sethrj: Are you sure this expression is right? It's comparing the difference in the square of speeds? Maybe it would be better to calculate the `vscm` and then check the magnitude of that...

Also it appears that `epsilon` here needs to have units of `[1/c^2]`? Is that right?

sethrj (changes requested):
Some of the arithmetic opertions in the collider need to be checked/clarified/simplified...

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Great!  I originally tried an array of GenericGridData [energy bin], so that we can swap x an y after find a corresponding cos_grid_data for each energy bin, then use the GenericGridCalculator to sample cos(theta). I believe that two approaches are basically equivalent, but using Two2GridData is much more compact and cleaner.  If we can encapsulate the inverse sampling inside Twod(Sub)gridCalculator, it would be even great!  Thanks for your help.

src/celeritas/phys/FourVector.hh
whokion: Okay, Dropped std::move.

test/celeritas/phys/FourVector.test.cc
whokion: Changed as suggested.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Added const&.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Yes, it is good idea - again, thank you for your help here. Actually, we may parameterize data further as the cdf might be a smooth function (probably a simple functional form may fit to parameterize by cos(theta) in each energy bin - need several parameters, but essentially remove tabulation in cos_theta. This is a detail optimization/refactoring data which may be explored after validating the final interactor with respect to Geant4.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Good to know!  I thought that they are very much equivalent. Anyway, changed as suggested.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Thanks! Nice catch for a bug.  It should be set to the initial un-rotated four vector. Fixed.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
sethrj: Yeah, unfortunately C++ forces the creation of temporaries that all store local copies of the intermediate compilation. The only ways to get around it are to manually unroll like this or to use a complicated template metaprogramming math library like Eigen or Blitz++ (neither of which are CUDA compatible).

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
sethrj: Agreed, a potential optimization for future work.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Do you mean `axpy(-vel_parallel, cm_dir, &cm_velocity_)` which will change value for cm_velocity?

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: I agree that they should be equivalent, i.e., `norm(vscm) == sqrt(dot_product(cm_velocity_, cm_velocity_) - ipow<2>(vel_parallel))`.  epsilon is small arbitrary magic number from Geant4 and used for condition for the degeneracy [1/c^2] and vscm [1/c] as c = 1 in Geant4.  Okay, let's use norm(vscm) as the condition for the degeneracy, which should be equivalent for the final logic.

src/celeritas/neutron/interactor/detail/CascadeCollider.hh
whokion: Never mind.  I got it.

whokion:
@sethrj Thanks again for all the comments and suggestions! I think that I addressed them for now.

sethrj (approved):
Thanks Soon! Looks great.

PR #1327, opened 2024-07-20
Refactor types to reduce duplication and cyclic dependencies

sethrj:
While rearranging some IO subroutines I inadvertently caused a circular dependency between collection/ldg/span/quantity/opaqueID. This moves all LDG related definitions and specializations to the "most downstream" compared to the primary implementation. I also reduced some type duplication in `geocel`/`orange` by defining Real2/Size2 in `geocel/Types.hh`. This also fixes an inadvertent find/replace error in GeoParamsInterface in #1000 .

### DIFF
--- a/src/corecel/OpaqueId.hh
+++ b/src/corecel/OpaqueId.hh
@@ -98,7 +98,7 @@ class OpaqueId
     CELER_CONSTEXPR_FUNCTION size_type unchecked_get() const { return value_; }
 
     //! Access the underlying data for more efficient loading from memory
-    size_type const* data() const { return &value_; }
+    CELER_CONSTEXPR_FUNCTION size_type const* data() const { return &value_; }
 
   private:
     size_type value_;
--- a/src/corecel/OpaqueId.hh
+++ b/src/corecel/OpaqueId.hh
@@ -17,12 +17,6 @@
 
 namespace celeritas
 {
-namespace detail
-{
-template<class, class>
-struct LdgLoader;
-}  // namespace detail
-
 //---------------------------------------------------------------------------//
 /*!
  * Type-safe index for accessing an array or collection of data.
@@ -103,6 +97,9 @@ class OpaqueId
     //! Get the value without checking for validity (atypical)
     CELER_CONSTEXPR_FUNCTION size_type unchecked_get() const { return value_; }
 
+    //! Access the underlying data for more efficient loading from memory
+    size_type const* data() const { return &value_; }
+
   private:
     size_type value_;
 
@@ -113,7 +110,6 @@ class OpaqueId
     {
         return static_cast<size_type>(-1);
     }
-    friend detail::LdgLoader<OpaqueId<value_type, size_type> const, void>;
 };
 
 //---------------------------------------------------------------------------//
@@ -139,6 +135,7 @@ CELER_DEFINE_OPAQUEID_CMP(>=)
 
 #undef CELER_DEFINE_OPAQUEID_CMP
 
+//---------------------------------------------------------------------------//
 //! Allow less-than comparison with *integer* for container comparison
 template<class V, class S, class U>
 CELER_CONSTEXPR_FUNCTION bool operator<(OpaqueId<V, S> lhs, U rhs)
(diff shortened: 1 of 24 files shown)


src/corecel/math/Quantity.hh
```diff
@@ -121,10 +114,11 @@ class Quantity
 #undef CELER_DEFINE_QACCESS
     //!@}
 
+    //! Access the underlying data for more efficient loading from memory
+    value_type const* data() const { return &value_; }
```
esseivaju:
```suggestion
    CELER_CONSTEXPR_FUNCTION value_type const* data() const { return &value_; }
```

src/corecel/OpaqueId.hh
```diff
@@ -103,6 +97,9 @@ class OpaqueId
     //! Get the value without checking for validity (atypical)
     CELER_CONSTEXPR_FUNCTION size_type unchecked_get() const { return value_; }
 
+    //! Access the underlying data for more efficient loading from memory
+    size_type const* data() const { return &value_; }
```
esseivaju:
```suggestion
    CELER_CONSTEXPR_FUNCTION size_type const* data() const { return &value_; }
```

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -72,20 +81,13 @@ struct LdgLoader<OpaqueId<I, T> const, void>
...
     using reference = value_type;
 
     CELER_CONSTEXPR_FUNCTION static reference read(pointer p)
     {
-#if CELER_DEVICE_COMPILE
-        return value_type{__ldg(&p->value_)};
-#else
-        return value_type{p->value_};
-#endif
+        return ldg(p->data());
```
esseivaju:
```suggestion
        return value_type{ldg(p->data())};
```

src/corecel/data/detail/LdgIteratorImpl.hh
```diff
@@ -14,30 +14,46 @@
...
+#endif
+}
+
 //---------------------------------------------------------------------------//
 /*!
  * Reads a value T using __ldg builtin and return a copy of it
  */
 template<class T, typename = void>
 struct LdgLoader
 {
```
esseivaju, line 49:
```suggestion
struct LdgLoader
{
    static_assert(std::is_const_v<T>, "Only const data are supported by __ldg");
```

src/geocel/Types.hh
```diff
@@ -18,9 +18,15 @@ namespace celeritas
...
-//! Fixed-size array for 3D space
+//! Three-dimensional cartesian coordinates
 using Real3 = Array<real_type, 3>;
 
+//! Two-dimensional cartesian coordinates
+using Real2 = Array<real_type, 2>;
+
+//! Two-dimensional extents
+using Size2 = Array<size_type, 2>;
+
```
esseivaju, line 29: There are still a few to `using ... = Array<real_type, 2>` and `using ... = Array<size_type, 2>` in Orange if you want to consolidate everything

src/corecel/data/detail/LdgIteratorImpl.hh
sethrj: Weird that it compiled with this one...

src/geocel/Types.hh
sethrj: The `Array<real_type, 2>` are all `Intersections`, where 2 is the number of intersections rather than a spatial coordinate—I don't consider it having the same semantics as a Real2. I've updated the Size2 ones though, thanks :)

src/corecel/data/detail/LdgIteratorImpl.hh
esseivaju: yeah, on Perlmutter it didn't. Maybe the CI failed earlier because of the missing `__host__ __device__`

esseivaju (approved):
LGTM!

src/corecel/data/detail/LdgIteratorImpl.hh
sethrj: Maybe it didn't fail on host-only because this class doesn't get selected by the traits?? 🤔

PR #1328, opened 2024-07-20
Remove celeritas include in corecel

esseivaju:
A few files in `corecel` were depending on `celeritas`.


sethrj (approved):
Whoops! Thanks for this! I wonder also if we should put this directly into the Interpolator class since it's not really used outside of that...

esseivaju:
It's probably better to keep it in `Types.hh` since it defines a `to_cstring` function that I forgot to move.

sethrj (approved):

PR #1329, opened 2024-07-21
Add git pre-commit hook

esseivaju:
Some errors are easy to overlook during review but could be validated with a git hook. Add a pre-commit script that will parse staged files in `corecel` for includes of any upstream library. This could easily be extended with more checks. Alternatively, this could be added to the CI

### DIFF
--- a/scripts/dev/pre-commit.validate-source
+++ b/scripts/dev/pre-commit.validate-source
@@ -1,9 +1,7 @@
 #!/bin/sh -e
-###############################################################################
-# File  : script/developer/pre-commit.validate-source
-#
-# Called by $SCALE/.git/hooks/pre-commit to apply various preliminary checks
-###############################################################################
+# Copyright 2021-2024 UT-Battelle, LLC, and other Celeritas developers.
+# See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 exit_with_message() {
   printf "\r\e[0;31m%s\e[0m\n" "$1"
--- a/scripts/dev/pre-commit.validate-source
+++ b/scripts/dev/pre-commit.validate-source
@@ -11,12 +11,12 @@ exit_with_message() {
 }
 
 check_corecel_dependencies() {
-  CMD="grep -n -E"
+  CMD="grep -nHE"
   IDX_FILE=0
   IDX_LINE_NO=1
   IDX_LINE_TXT=2
 
-  # use ripgrep is available, it's faster
+  # use ripgrep if available, it's faster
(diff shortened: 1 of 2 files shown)


scripts/dev/pre-commit.validate-source
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+###############################################################################
+# File  : script/developer/pre-commit.validate-source
+#
+# Called by $SCALE/.git/hooks/pre-commit to apply various preliminary checks
+###############################################################################
```
sethrj:
```suggestion
#!/bin/sh -e
# Copyright 2021-2024 UT-Battelle, LLC, and other Celeritas developers.
# See the top-level COPYRIGHT file for details.
# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```

sethrj (approved):
Thanks, this looks like a definite step in the right direction. I wonder if we should start using/requiring https://pre-commit.com which can manage git hooks more cleanly than the script I hacked together?

scripts/dev/pre-commit.validate-source
esseivaju: I didn't set `-e` because `rg`/`grep` return `1` if they don't find any match and we don't want to fail if that's the case. Does `sh` always supports read into an array variable?

esseivaju:
> Thanks, this looks like a definite step in the right direction. I wonder if we should start using/requiring https://pre-commit.com which can manage git hooks more cleanly than the script I hacked together?

This looks interesting but C++ is not in the list of supported languages so we'd only use that to manage our custom git hooks?

scripts/dev/pre-commit.validate-source
sethrj: Not sure, but `bash` is very old on modern mac systems and may be eventually deleted...

sethrj:
@esseivaju I think https://github.com/pre-commit/mirrors-clang-format?tab=readme-ov-file ?

scripts/dev/pre-commit.validate-source
esseivaju: Actually, `-e` has no effect currently because of how the function is called before a `||`: `check_corecel_dependencies || exit_with_message "..."`

sethrj: Yeah I usually run with `-e` because if some other part of teh script has an unexpected error, the program continues in a rather undefined state...

esseivaju:
> @esseivaju I think https://github.com/pre-commit/mirrors-clang-format?tab=readme-ov-file ?

Ah right, yes I think it would be nice to integrate this at some point.

PR #1330, opened 2024-07-22
Update frontier toolchain and backward compatibility

sethrj:
A recent [system update on Frontier](https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#id17) required an updated toolchain. For some reason even though the same ROCm and HIP versions are being used, some new warnings have started showing up, but there are no changes in the answers.

While doing some very basic profiling of a tilecal run I noticed that `celeritas::fma` was showing up as a symbol, which means we're not inlining it like we should.

Finally to get the regression suite to run as is, I have temporarily restored (with a warning) the "geant physics" string option.


amandalund (approved):
Thanks @sethrj!

PR #1331, opened 2024-07-23
Add unit test function wrapper for diagnosing solver convergence

sethrj:
@VHLM2001 This takes a function $` f\colon\mathbb{R^1}\to\mathbb{R^1} `$ and wraps it with a counter so that we can test convergence properties in solvers.


sethrj:
@VHLM2001 Could you take a look at the convergence counts in the updated unit test and check that they seem reasonable?

VHLM2001:
> @VHLM2001 Could you take a look at the convergence counts in the updated unit test and check that they seem reasonable?

From what I can tell they seem reasonable and consistent with what I am seeing when with involutes in python, assuming that the counts is reset to zero after each retrieval in the RegulaFalsiRootFinder test cases.

PR #1332, opened 2024-07-23
Add optical collector to celer-sim

stognini:
This PR builds the `OpticalCollector` object needed to pre-generate optical photons and is a part of #886 .

The main additions are:
- Add a new `shared_ptr<OpticalCollector>` to the `celer-sim` `Runner` class.
- Add optical physics options to `GeantPhysicsOptions`.
- Add the capability to create different `celer-sim` tests with different geometries and inputs. This is not used yet, but will be important when the optical stepping loop is enabled, allowing us to run different instances, such as `celer-sim-cms` for `simple-cms.gdml`, and `celer-sim-lar` for `lar-sphere.gdml`.

The new `celer-sim-lar` test is not enabled yet since the optical photons are pre-generated but not transported. This leads to two failed assertions in the existing main transport loop:
1. During `pre-step` in `PhysicsStepUtils.hh:76`:
```cpp
CELER_ASSERT(total_macro_xs > 0 || !particle.is_stopped());
```
2. During `along-step-neutral` in `AlongStepNeutralImpl.hh:34`:
```cpp
CELER_ASSERT(step > 0);
```


app/celer-sim/Runner.cc
```diff
@@ -293,24 +297,27 @@ void Runner::build_core_params(RunnerInput const& inp,
...
     params.cutoff = CutoffParams::from_import(
-        imported, params.particle, params.material);
+        imported_, params.particle, params.material);
 
     // Construct shared data for Coulomb scattering
-    params.wentzel = WentzelOKVIParams::from_import(imported, params.material);
+    params.wentzel = WentzelOKVIParams::from_import(imported_, params.material);
+
+    // FIX ME: can't pass private member to lambda so I made a copy
+    auto const imported = imported_;
```
sethrj: sure you can, you just either have to pass  `this`  or do e.g. `imported = std::ref(imported_)`

app/celer-sim/Runner.hh
```diff
@@ -84,9 +86,11 @@ class Runner
 
     //// DATA ////
 
+    ImportData imported_;
```
sethrj: Instead of this big guy as member data, how about a unique_ptr/shared_ptr that we clear at the end of construction? Or better yet, forward declare `ImportData` in the header, make a new helper function  `ImportData load_import_data(RunnerInput const&);`, keep the result by value locally within the constructor, and pass it into the optical/core functions that need it. That way the scope is nicely managed.

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -86,6 +86,17 @@ struct GeantPhysicsOptions
...
     //!@}
 
+    //!@{
+    //! \name Optical photon physics
+    //! Enable Cerenkov physics
+    bool cerenkov{true};
+    //! Enable scintillation
+    bool scintillation{true};
+    //! Enable optical Rayleigh scattering
+    bool optical_rayleigh{true};
```
sethrj: Do these do anything? I don't see any changes in GeantImporter itself.

src/celeritas/optical/OpticalCollector.hh
```diff
@@ -78,7 +77,7 @@ class OpticalCollector
 
   public:
     // Construct with core data and optical params
-    OpticalCollector(CoreParams const&, Input&&);
+    OpticalCollector(SPConstCore, Input&&);
```
sethrj: Why is this change needed? The optical collector shouldn't keep a copy of the core. It doesn't look like you're using it as a shared pointer.

test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -223,35 +223,35 @@ auto LArSpherePreGenTest::run(size_type num_tracks, size_type num_steps)
...
+                  auto const& pre = dist.points[StepPoint::pre];
+                  auto const& post = dist.points[StepPoint::post];
+                  EXPECT_GT(pre.speed, zero_quantity());
+                  EXPECT_NE(post.pos, pre.pos);
+                  EXPECT_GT(dist.step_length, 0);
+                  EXPECT_EQ(0, dist.material.get());
+              }
+              result.charge.insert(
+                  result.charge.end(), charge.begin(), charge.end());
+          };
```
sethrj: What changed here?

src/celeritas/optical/OpticalCollector.hh
stognini: CoreParams is a shared_ptr in the Runner class. Passing the shared_ptr seemed more natural than the object's reference, but I can revert that.

test/celeritas/optical/OpticalCollector.test.cc
stognini: Oh, the change in the `OpticalCollector` constructor from `const&` to `shared_ptr` (line 163) led to a full clang-formatting of the file.

src/celeritas/ext/GeantPhysicsOptions.hh
stognini: They are used in `Runner::build_optical_collector`. Though `optical_rayleigh` is not used yet (the optical collector input does not load Rayleigh data).

app/celer-sim/Runner.hh
stognini: The imported data comes from the lambda function in `build_core_params`, which sets up what Importer type we use. Since we don't want to initialize Geant4 more than once, maybe I should set up the importer in the constructor and pass the world volume and import data as needed to the `build_[x]` helper functions?

src/celeritas/optical/OpticalCollector.hh
sethrj: No, the runner holds a copy of it because it uses the core data. The optical collector merely *uses* the core params to construct its equivalent.

test/celeritas/optical/OpticalCollector.test.cc
sethrj: OK, makes sense. It can be reverted then.

app/celer-sim/Runner.cc
stognini: Oh I see, using `this` works, adn I only considered `this->imported_`.

app/celer-sim/Runner.hh
sethrj: What if we save the world volume as class member data, do the import in the constructor, and pass the import data to the various parts that need it?

stognini: That was one of my first thoughts actually, when dealing with passing the imported data from `build_core_params` to `build_optical_data`. I would either need a direct G4 include in our `Runner.hh` or have something like the `UPExtern` we use for ROOT. Which led me to add the data itself, despite being large. Do you think we should go for the unique ptr?

sethrj: No, since it's a raw pointer, you just have to forward declare:
```c++
class G4VPhysicalVolume;  // IWYU pragma: keep
```

stognini: Oh, if you're okay with the raw pointer, then yes, that's easy. Come to think of it, the runner's lifetime is basically the celer-sim one, but the Geant4 application should end after we're done with the importing (currently in `build_core_params`). Wouldn't  this poses a problem with our world raw pointer (i.e. Geant4 deleting the geometry when it goes out of scope)?

app/celer-sim/Runner.cc
```diff
@@ -121,15 +121,40 @@ size_type calc_num_streams(RunnerInput const& inp, size_type num_events)
...
+    using SPImporter = std::shared_ptr<ImporterInterface>;
+
     CELER_EXPECT(output);
 
     this->setup_globals(inp);
 
+    // Possible Geant4 world volume so we can reuse geometry
+    G4VPhysicalVolume const* g4world{nullptr};
+
+    // Import data and load geometry
```
sethrj, line 133: Since we were both confused about this (and I wrote it!) perhaps extend the comment to note that if loading through Geant4, the lifetime of the Geant4 geometry etc. are scoped by the importer.

app/celer-sim/Runner.cc
```diff
@@ -546,6 +550,44 @@ void Runner::build_step_collectors(RunnerInput const& inp)
...
+                                     ImportData const& imported)
+{
+    if (imported.optical.empty())
+    {
+        // No optical data loaded
+        return;
+    }
+
+    // TODO: Rethink if statements after implementing CelerOpticalPhysicsList
+    CELER_EXPECT(core_params_);
```
sethrj, line 574: Move to the top of the function, before the `if`.

app/celer-sim/Runner.cc
```diff
@@ -546,6 +550,44 @@ void Runner::build_step_collectors(RunnerInput const& inp)
     }
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct optical collector. This *MUST* be called *AFTER*
+ * \c this->build_core_params , as it depends on it.
```
sethrj:
```suggestion
 * Construct optical collector.
 *
 * \pre Must be called after \c build_core_params
```

app/celer-sim/Runner.cc
```diff
@@ -546,6 +550,44 @@ void Runner::build_step_collectors(RunnerInput const& inp)
...
+    auto const& optical_data = imported.optical.begin()->second;
+    if (static_cast<bool>(optical_data.properties))
+    {
+        oc_inp.properties = OpticalPropertyParams::from_import(imported);
+    }
+    if (static_cast<bool>(optical_data.properties))
+    {
+        oc_inp.cerenkov = std::make_shared<CerenkovParams>(oc_inp.properties);
+    }
+    if (static_cast<bool>(optical_data.scintillation))
```
sethrj: You don't need the static casts:
```suggestion
    if (optical_data.properties)
    {
        oc_inp.properties = OpticalPropertyParams::from_import(imported);
    }
    if (optical_data.properties)
    {
        oc_inp.cerenkov = std::make_shared<CerenkovParams>(oc_inp.properties);
    }
    if (optical_data.scintillation)
```
Also I don't know if the first properties/cerenkov should be combined. Maybe we need an optical model/process list...

app/celer-sim/simple-driver.py
```diff
@@ -47,6 +47,9 @@ def strtobool(text):
     'msc': "urban" if core_geo == "vecgeom" else "none",
     'eloss_fluctuation': True,
     'lpm': True,
+    'cerenkov': True,
+    'scintillation': True,
+    'optical_rayleigh': True,
```
sethrj: Shouldn't these be conditional on some testing input option?

sethrj (approved):
Looks good after these small updates. Thanks!

app/celer-sim/CMakeLists.txt
```diff
@@ -40,88 +40,104 @@ if(NOT BUILD_TESTING)
...
-set_tests_properties("app/celer-sim:cpu" PROPERTIES
-  ENVIRONMENT "${_env};${CELER_G4ENV};${CELER_OMP_ENV}"
-  REQUIRED_FILES "${_driver};${_gdml_inp};${_hepmc3_inp}"
-  LABELS "app;nomemcheck"
-  ${CELER_PROCESSORS}
-  ${_needs_deps}
-)
+# TODO: enable celer-sim-lar[cpu:gpu] test when optical stepping loop is
+#       available. Decide if lar-sphere.gdml should be copied from
+#       test/geocel/data to app/data (as simple-cms.gdml)
```
amandalund: It looks like the two simple-cms geometries are slightly different (the one in app/data has tiny gaps between the cylinders, while the one in test/geocel/data does not). Maybe that's why we have two copies?

app/celer-sim/simple-driver.py
```diff
@@ -91,6 +94,7 @@ def strtobool(text):
     'num_track_slots': num_tracks,
     'max_steps': max_steps,
     'initializer_capacity': 100 * max([num_tracks, num_primaries]),
+    'optical_buffer_capacity': 100 * max([num_tracks, num_primaries]),
```
amandalund:
```suggestion
    'optical_buffer_capacity': 3 * max_steps * num_tracks,
```
Does the test still fail if you increase this buffer size so it's large enough to hold all the distribution data? Since we don't actually convert the distributions into optical photons yet the buffers will just keep growing, and in addition to the failures you mention I am seeing:
```console
/home/alund/celeritas_project/celeritas/app/celer-sim/celer-sim.cc:251: critical: While running input at <stdin>: celeritas: runtime error: insufficient capacity (4500) for buffered scintillation distribution data (total capacity requirement of 4521)
/home/alund/celeritas_project/celeritas/src/celeritas/optical/detail/ScintPreGenAction.cc:84: 'buffer_size + core_state.size() <= buffer.size()' failed
```

Looks like you'll also need to make sure `simple_calo` is disabled for the lar-sphere test.

amandalund (commented):
Nice @stognini!

app/celer-sim/CMakeLists.txt
stognini: That is a good point. Could be worth keeping one of them then. The differences most likely come from when the simple-cms geometry was written. I had placed a tolerance between cylinders so we could test boundary crossings between G4 and VecGeom, so we had it updated during geometry tests. Nowadays these numbers are set in our gdml-generator app (see [here](https://github.com/celeritas-project/utils/blob/820af6db40b658793088ab55bba87be91aed899c/gdml-generator/src/SimpleCmsDetector.hh#L51-L57)).

app/celer-sim/simple-driver.py
stognini: Oh good catch, since GeantPhyscisOptions was reverted after our discussion, these should have been too. They were originally just set to true due to our initial planning: if the data is not available from the gdml, the optical physics will not be initialized. So it would work for both simple-cms and for an optical test, such as the lar-sphere.

app/celer-sim/CMakeLists.txt
sethrj: I don't think we need the one with the "tolerance": geometry errors etc. should primarily be unit tested rather than the subject of integration tests.

app/celer-sim/CMakeLists.txt
```diff
@@ -40,88 +40,103 @@ if(NOT BUILD_TESTING)
...
+  endif()
+  set(CELER_GEO ${CELERITAS_CORE_GEO})
+  if(CELERITAS_CORE_GEO STREQUAL "ORANGE" AND NOT CELERITAS_USE_Geant4)
+    # Signal to the app that we have to use the JSON input
+    set(CELER_GEO "${CELER_CORE_GEO}-JSON")
+  endif()
+  
+  add_test(NAME "app/celer-sim-${test_ext}:cpu"
+    COMMAND "${CELER_PYTHON}"
+    "${_driver}" "${_gdml_inp}" "${_hepmc3_inp}" "${mctruth_name}"
```
amandalund: I think this should be:
```suggestion
    "${_driver}" "${_gdml_inp}" "${_hepmc3_inp}" "${mctruth_out}"
```

app/data/lar-sphere.gdml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amandalund, line 1: One thing to keep in mind if we add a copy of the gdml to the app directory is we may need to update both versions as we add more optical processes/properties.

amandalund (approved):

PR #1333, opened 2024-07-23
Add more root finding algorithms

VHLM2001:
Adds bisection method and a modified Regula Falsi method (Illinois method) to find roots of function.

Bisection method is linear, guarantees a decently fast convergence for when other methods are slow to converge. This method follows by calculating whether midpoint is left or right of the root and updating bound on the same side to be the midpoint.

The Illinois method addresses the degenerate case of the Regula Falsi method in which one bound becomes fixed in the iterations leading to extremely slow convergence. This is done by checking if the approximated root calculated in the current iteration is on the same side of the true root as the previous iteration. When this occurs the value of the root function at the bound of the other side is halved.


src/corecel/math/BisectionRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+#include <cmath>
+#include <type_traits>
+
+#include "celeritas_config.h"
+#include "corecel/Assert.hh"
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+#include "corecel/math/Algorithms.hh"
+
+#include "detail/MathImpl.hh"
```
sethrj: I suspect this got pulled in by vscode, nothing you use is in there...

src/corecel/math/BisectionRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+    real_type f_left = func_(left);
+    real_type f_root = 1;
+    real_type root = 0;
+    int remaining_iters = max_iters_;
+
+    // Iterate on root
+    do
+    {
+        // Estimate root and update value
+        root = 0.5 * (left + right);
```
sethrj:
```suggestion
        root = real_type(0.5) * (left + right);
```
Otherwise you're doing a float->double->float conversion potentially

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,122 @@
...
+#include "detail/MathImpl.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Perform Regula Falsi (see RegulaFalsi for more details) iterations given a
+ * root function \em func and tolerance \em tol using the Illinois method.
+ *
+ * Illonois method modifies the standard approach by comparing the sign of
```
sethrj:
```suggestion
 * Illinois method modifies the standard approach by comparing the sign of
```

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,122 @@
...
+ * \em func(root) approximation in the current iteration with the previous
+ * approximation. If both iterations are on the same side then the \em func at
+ * the bound on the other side is halved.
+ */
+template<class F>
+class Illinois
```
sethrj:
```suggestion
class IllinoisRootFinder
```

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,122 @@
...
+        // Estimate root and update value
+        root = (left * f_right - right * f_left) / (f_right - f_left);
+        f_root = func_(root);
+
+        // Update the bound which produces the same sign as the root
+        if (signum(f_left) == signum(f_root))
+        {
+            left = root;
+            f_left = f_root;
+            if (side == -1)
```
sethrj: This looks like it should be a boolean enum rather than a real value.

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,122 @@
...
+            {
+                f_right *= 0.5;
```
sethrj:
```suggestion
                f_right *= real_type(0.5);
```
and belowas well

test/corecel/math/IllinoisRootFinder.test.cc
```diff
@@ -0,0 +1,96 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2020-2024 UT-Battelle, LLC, and other Celeritas developers.
```
sethrj: Since all these files are new, you can just mark them as 2024 :) They're normally generated automatically with `scripts/dev/celeritas-gen.py`.

test/corecel/math/RegulaFalsiRootFinder.test.cc
```diff
@@ -65,29 +65,30 @@ TEST(RegulaFalsi, trigometric)
...
+TEST(RegulaFalsi, exponential_intersect)
 {
-    double x = 0.0;
-    double y = 2.0;
-    double u = 0.0;
-    double v = -1.0;
+    double x = 1.5;
+    double y = 0.5;
+    double u = -0.7071067812;
+    double v = 0.7071067812;
```
sethrj: While you're at it (see #1331) change these to `real_type`, otherwise you're making the lambdas do work in double precision

sethrj (changes requested):
Only a few small suggestions. Nice work, and isn't it cool how this modular strategy lets us change the involute solver performance with a single line (in the solver itself)?

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,130 @@
...
+ * Illinois method modifies the standard approach by comparing the sign of
+ * \em func(root) approximation in the current iteration with the previous
+ * approximation. If both iterations are on the same side then the \em func at
+ * the bound on the other side is halved.
+ */
+template<class F>
+class IllinoisRootFinder
+{
+  public:
+    //! Enum defining side of aproximated root to true root
```
sethrj: This enum can be private, and I think you can even define it local to the function

src/corecel/math/BisectionRootFinder.hh
```diff
@@ -0,0 +1,113 @@
...
+ */
+template<class F>
+class BisectionRootFinder
+{
+  public:
+    // Contruct with function to solve and solution tolerance
+    inline CELER_FUNCTION BisectionRootFinder(F&& func, real_type tol);
+
+    // Solve for a root between two points
+    inline real_type operator()(real_type left, real_type right);
```
sethrj: Missing `CELER_FUNCTION`

src/corecel/math/IllinoisRootFinder.hh
```diff
@@ -0,0 +1,130 @@
...
+        init = 0,
+        right = 1
+    };
+
+  public:
+    // Contruct with function to solve and solution tolerance
+    inline CELER_FUNCTION IllinoisRootFinder(F&& func, real_type tol);
+
+    // Solve for a root between two points
+    inline real_type operator()(real_type left, real_type right);
```
sethrj: Missing `CELER_FUNCTION`

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -37,11 +35,11 @@ namespace celeritas
  * proximity to 0.
  */
 template<class F>
-class RegulaFalsi
+class RegulaFalsiRootFinder
```
sethrj, line 38: Oops, I didn't even notice on the first review 😄

src/corecel/math/RegulaFalsiRootFinder.hh
```diff
@@ -37,11 +35,11 @@ namespace celeritas
...
 {
   public:
     // Contruct with function to solve and solution tolerance
-    inline CELER_FUNCTION RegulaFalsi(F&& func, real_type tol);
+    inline CELER_FUNCTION RegulaFalsiRootFinder(F&& func, real_type tol);
 
     // Solve for a root between two points
     inline real_type operator()(real_type left, real_type right);
```
sethrj: Might as well add `CELER_FUNCTION` here as well (you did this in the involute branch)

sethrj (approved):
Excellent work! Thanks!

PR #1337, opened 2024-07-26
Fix GenPrism surface shapes to match Geant4

mrguilima:
Add new GenTrap, with the goal of perfectly matching Geant4's G4GenericTrap solid.

Unlike GenPrism, which linearly varies both linear and angular coefficients from each edge (from -z to +z), which ends up with non-straight edges,
this new GenTrap forces these "vertical" edges to be straight lines between corresponding corners (from -z to +z).

Only tested without CUDA so far, so it is a [WIP].


sethrj:
@mrguilima , what I was trying to explain on Slack is that there's no reason to have a "curvy edged" shape at all. I renamed GenTrap to GenPrism because GenTrap allows degenerate trapezoidal shapes (tetrahedron, envelope) and our implementation would allow it to be trivially extended.

Does the new GenTrap have the same capability or is it limited to exactly four sides? Can you delete the existing GenPrism and rename your GenTrap to GenPrism so that we can use `git` to do a diff rather than having to diff two copy-pasted classes?

mrguilima:
Yes, I can get rid of old implementation, and move the new on into the GenPrism class.

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -794,6 +796,9 @@ TEST_F(GenPrismTest, odd_tetrahedron)
     EXPECT_FALSE(result.interior) << result.interior;
     EXPECT_VEC_SOFT_EQ((Real3{-1, -1, -3}), result.exterior.lower());
     EXPECT_VEC_SOFT_EQ((Real3{2, 1, 3}), result.exterior.upper());
+
+    GTEST_SKIP() << "GenPrism::odd_tetrahedron fails check_corners()test!";
```
sethrj: I think it's OK that this test doesn't work, because the `check_corners` bumps toward the origin, and since the lower point is well off the origin, the "expected inside" position really is outside.

Instead of skipping, just add a comment that we can't run `odd_tetrahedron` because one of the faces doesn't enclose the origin.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -200,13 +200,16 @@ class Ellipsoid final : public IntersectRegionInterface
...
+ * The faces have an orientation and ordering so that \em twisted faces can be
+ * constructed by joining corresponding points using straight-line "vertical"
+ * edges, directly matching the G4GenericTrap definition, but using a generic
+ * quadric expression for each twisted face.
  *
  * Trapezoids constructed from the helper functions will have sides that are
  * same ordering as a prism: the rightward face is first (normal is along the
  * +x axis), then the others follow counterclockwise.
+ * TODO: is this last statement is correct? e.g. is the +x face always the
+ *       first face to be constructed?
```
sethrj: Yeah, for both `GenPrism::from_trap` and `GenPrism::from_trd` the ordering starts at the lower right corner and proceeds counterclockwise. This gives it the same orientation as the regular prism.

sethrj (approved):
This is great @mrguilima ! The EMEC now is consistent across all three geometry implementations. Nice work!
emec-yz-x1800mm

sethrj (approved):
Awesome! Thanks guilherme.

PR #1338, opened 2024-07-26
Reduce string duplication from assertion messages

sethrj:
Inspired by @esseivaju 's example in #1325 , this defines external constant strings for "runtime error" types to both reduce library size and to avoid "magic constants" (i.e. requiring that special string values in one file match string values in another).

HIP/frontier (ORANGE ndebug):
| Library        | Before   | After    |
| -------------- | -------- | -------- |
| libcorecel.a   |  1613066 |  1606148 |
| libgeocel.a    |   902558 |   898838 |
| liborange.a    |  4122526 |  4095566 |
| libceleritas.a | 11385000 | 11295664 |
| libaccel.a     |   869100 |   863828 |

CUDA/wildstyle (ORANGE reldeb):
| Library        | Before    | After     |
| -------------- | --------- | --------- |
| libcorecel.a   |  17431736 |  17415056 |
| libgeocel.a    |  10777392 |  10776984 |
| liborange.a    |  54714576 |  54682904 |
| libceleritas.a | 182853472 | 181899976 |
| libaccel.a     |  21292144 |  21290984 |

Unfortunately I'm not able to use `static inline constexpr` for the CUDA/HIP printf statements:
```
/home/s3j/Code/celeritas/src/corecel/Assert.hh(516): error: An inline __device__/__constant__/__managed__ variable must have internal linkage when the program is compiled in whole program mode (-rdc=false)
```
see [this llvm bug report](https://reviews.llvm.org/D88786?id=295997) for details.

### DIFF
--- a/src/corecel/Assert.hh
+++ b/src/corecel/Assert.hh
@@ -512,15 +512,17 @@ class RichContextException : public std::exception
 // INLINE FUNCTION DEFINITIONS
 //---------------------------------------------------------------------------//
 
+#if (defined(__CUDA_ARCH__) && defined(NDEBUG)) || defined(__HIP__)
+inline constexpr char device_assertion_str[]
+    = "%s:%u:\nceleritas: internal assertion failed: %s\n";
+#endif
+
 #if defined(__CUDA_ARCH__) && defined(NDEBUG)
 //! Host+device definition for CUDA when \c assert is unavailable
 inline __attribute__((noinline)) __host__ __device__ void device_debug_error(
     DebugErrorType, char const* condition, char const* file, int line)
 {
-    printf("%s:%u:\nceleritas: internal assertion failed: %s\n",
-           file,
-           line,
-           condition);
+    printf(device_assertion_str, file, line, condition);
     __trap();
 }
 #elif defined(__HIP__)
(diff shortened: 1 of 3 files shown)


src/corecel/Assert.hh
```diff
@@ -476,6 +478,13 @@ class RuntimeError : public std::runtime_error
     //! Access detailed information
     RuntimeErrorDetails const& details() const { return details_; }
 
+    //!@{
+    //! String constants for "which" error message
+    static char const validate_err_str[];
+    static char const not_config_err_str[];
+    static char const not_impl_err_str[];
```
esseivaju, line 485: What about this
```suggestion
    static constexpr char validate_err_str[] {"runtime"};
    static constexpr char not_config_err_str[] {"configuration"};
    static constexpr char not_impl_err_str[] {"implementation"};
```

or using `string_view`?

sethrj: That would require making those constants `inline` and making the linker do the work. I think it's fine to keep the definition of those constants in the `.cc` file where they could be changed without forcing the whole codebase to recompile.

What's the advantage of using string_view? The `RuntimeError.which` now takes a `char const*` C string so a string_view would technically conflict with that.

esseivaju:
> Unfortunately I'm not able to use `static inline constexpr` for the CUDA/HIP printf statements:

They also mention it [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#inline-variable), maybe that will be supported one day

src/corecel/Assert.hh
esseivaju: No big advantage for string_view here, it just provides more of a c++ interface at a cost of a few extra bytes.

esseivaju (approved):

PR #1340, opened 2024-07-29
Fix HepMC3 tests to work with 3.3.0

sethrj:
I think our "no vertex" tests for HepMC3 (which were an odd use case) relied on unspecified behavior. @mrguilima discovered that they fail with 3.3.0, and it seems to be that previously the default position/time for each event reset to zero, and now the default position/time if unspecified is set to that of the previous event.


amandalund (approved):
Thanks @sethrj!

PR #1341, opened 2024-07-29
Use spack-based workflow for address sanitizer

sethrj:
This adds an address sanitizer and leak checker to the spack builds and removes from the docker build. With geant4 off we can use leak detection.


esseivaju (approved):
LGTM! We can consider adding more sanitizer at some point.

PR #1342, opened 2024-07-30
Integrate involute surface into ORANGE construction

VHLM2001:
Bring the implemented involute geometry from pull [210ce4422](https://github.com/celeritas-project/celeritas/pull/1295) into the runtime. Before pulling analysis of the different sections of the code should be looked at for compilation efficiency and checks to ensure that the involute code does not slow down other non-related sections when not called for.


test/orange/surf/SoftSurfaceEqual.test.cc
```diff
@@ -192,6 +192,64 @@ TEST_F(SoftSurfaceEqualTest, general_quadric)
         softeq_(ref, SurfaceTranslator(Translation{{large, 0, 0}})(ref)));
 }
 
+TEST_F(SoftSurfaceEqualTest, involute)
+{
+    Involute ref_ccw{{1.0,0.0},1.0, 2.0, Involute::Sign::counterclockwise, 1.0, 2.0};
+    Involute ref_cw{{1.0,0.0}, 1.0 , 2.0, Involute::Sign::clockwise, 1.0, 2.0};
```
sethrj: Add a test that ref_ccw != ref_cw 😉 

Also the function would be more readable if you define:
```c++
constexpr auto ccw = Involute::Sign::counterclockwise;
constexpr auto cw =Involute::Sign::clockwise;
```
and use those in your constructors.

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -1202,6 +1206,182 @@ TEST_F(GenPrismTest, adjacent_twisted)
...
+
+        static char const expected_node[] = "all(+1, +2, -3, -4, +5, -6)";
+
+        EXPECT_EQ(expected_node, result.node);
+        EXPECT_VEC_SOFT_EQ((Real3{-4, -4, -1}), result.exterior.lower());
+        EXPECT_VEC_SOFT_EQ((Real3{4, 4, 1}), result.exterior.upper());
+    }
+
+    static char const* const expected_surfaces[] = {
+        "Involute: r, a, sign, tmin, tmax =1 3.1416 1 1.7321 4.3652 at {0,0}",
```
sethrj: Could you change the output to look like
```suggestion
        "Involute ccw: r=1, a=3.1416, t=[1.7321,4.3652) at x=0, y=0",
```

src/orange/surf/detail/SurfaceTranslator.cc
```diff
@@ -106,6 +107,27 @@ Sphere SurfaceTranslator::operator()(Sphere const& other) const
...
+ */
+Involute SurfaceTranslator::operator()(Involute const& other) const
+{
+    using constants::pi;
+    using Real2 = Involute::Real2;
+    Real3 other_origin = {other.origin()[0], other.origin()[1], 0};
+    Real3 other_origin_tr_3 = tr_.transform_up(other_origin);
+    Real2 other_origin_tr = {other_origin_tr_3[0], other_origin_tr_3[1]};
+
+    Involute invo{other_origin_tr,
```
sethrj:
```suggestion
    Real3 origin = tr_.transform_up({other.origin()[0], other.origin()[1], 0});

    Involute invo{{origin[0], origin[1]},
```

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -152,7 +152,7 @@ InvoluteSolver::operator()(Real3 const& pos,
     real_type v = dir[1];
 
     // Mirror systemm for counterclockwise involutes
-    if (sign_)
+    if (sign_==Sign::clockwise)
```
sethrj: Please run `git-clang-format` on your branch

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -68,6 +68,21 @@ bool SoftSurfaceEqual::operator()(SphereCentered const& a,
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two centered involutes for near equality.
+ */
+bool SoftSurfaceEqual::operator()(Involute const& a, Involute const& b) const
+{
+    return this->soft_eq_sq(a.r_b(), b.r_b()) && this->soft_eq_sq(a.a(), b.a())
+           && this->soft_eq_sq(a.sign() == Involute::Sign::clockwise, 
+                               b.sign() == Involute::Sign::clockwise)
```
sethrj:
```suggestion
           && a.sign() == b.sign()
```

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -68,6 +68,21 @@ bool SoftSurfaceEqual::operator()(SphereCentered const& a,
...
+           && this->soft_eq_sq(a.tmin(), b.tmin())
+           && this->soft_eq_sq(a.tmax(), b.tmax())
+           && this->soft_eq_sq(a.origin()[0], b.origin()[0])
+           && this->soft_eq_sq(a.origin()[1], b.origin()[1]);
```
sethrj:
```suggestion
           && this->soft_eq_distance({a.origin()[0], a.origin()[1], 0}, {b.origin()[0], b.origin()[1], 0});
```

src/orange/surf/SoftSurfaceEqual.cc
```diff
@@ -68,6 +68,21 @@ bool SoftSurfaceEqual::operator()(SphereCentered const& a,
     return this->soft_eq_sq(a.radius_sq(), b.radius_sq());
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Compare two centered involutes for near equality.
+ */
+bool SoftSurfaceEqual::operator()(Involute const& a, Involute const& b) const
+{
+    return this->soft_eq_sq(a.r_b(), b.r_b()) && this->soft_eq_sq(a.a(), b.a())
```
sethrj: I don't think any of these "soft equivalent"s should be square... `soft_eq_sq` is for testing soft equivalence on the square of values.

src/orange/surf/FaceNamer.cc
```diff
@@ -111,6 +111,15 @@ std::string FaceNamer::Impl::operator()(Sphere const&) const
     return "s";
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a name for an involute.
+ */
+std::string FaceNamer::Impl::operator()(Involute const&) const
+{
+    return "inv";
```
sethrj, line 120: Since we can have multiple involutes per intersect region, you should add a `num_inv` to the state and do like the `Plane` does.

src/orange/orangeinp/detail/SurfaceHashPoint.hh
```diff
@@ -47,6 +47,8 @@ struct SurfaceHashPoint
         return std::sqrt(s.radius_sq());
     }
 
+    real_type operator()(Involute const& s) const { return s.r_b(); }
```
sethrj: I don't think this what you want. The "surface hash" should return "most likely to be well distributed" values for different objects in a unit. All the involutes in a HFIR core for example will have the same `r_b`. You probably want to return the displacement angle.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -84,6 +86,49 @@ class Box final : public IntersectRegionInterface
...
+    void build(IntersectSurfaceBuilder&) const final;
+
+    // Output to JSON
+    void output(JsonPimpl*) const final;
+
+    //// ACCESSORS ////
+
+    //! Radius
+    Real3 radii() const { return radii_; }
+    Real2 a() const { return a_; }
```
sethrj: `a` isn't a very descriptive name (I think we should also rename it on the surface).  Can you rename to `displacement_angle` or something?

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -84,6 +86,49 @@ class Box final : public IntersectRegionInterface
...
+    // Build surfaces
+    void build(IntersectSurfaceBuilder&) const final;
+
+    // Output to JSON
+    void output(JsonPimpl*) const final;
+
+    //// ACCESSORS ////
+
+    //! Radius
+    Real3 radii() const { return radii_; }
```
sethrj: Expand the documentation: what are the three radii? They sure aren't x/y/z.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -84,6 +86,49 @@ class Box final : public IntersectRegionInterface
...
+    Real2 a() const { return a_; }
+    real_type tmin() const { return tmin_; }
+    real_type tmax() const { return tmax_; }
```
sethrj: I think these would be better as `Real2 t_bounds`

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -84,6 +86,49 @@ class Box final : public IntersectRegionInterface
...
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * An involute centered on the origin.
+ *
+ * \note Be aware there's also an involute *surface* at orange/surf/Involute.hh 
+ * in a different namespace.
+ */
+class Involute final : public IntersectRegionInterface
```
sethrj: Please order this alphabetically in the file

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -84,6 +86,49 @@ class Box final : public IntersectRegionInterface
...
+
+    //// ACCESSORS ////
+
+    //! Radius
+    Real3 radii() const { return radii_; }
+    Real2 a() const { return a_; }
+    real_type tmin() const { return tmin_; }
+    real_type tmax() const { return tmax_; }
+    Sign sign() const { return sign_; }
+    real_type hh() const { return hh_; }
```
sethrj: Rename to `halfheight()` like the other accessors in this file

src/orange/OrangeTypes.hh
```diff
@@ -126,6 +126,7 @@ enum class SurfaceType : unsigned char
     kz,  //!< Cone parallel to Z axis
     sq,  //!< Simple quadric
     gq,  //!< General quadric
+    inv,  //!< Involute
```
sethrj, line 134: See the note in the documentation for  "These are ordered by number of coefficients needed for their representation"? The involute requires 6 coefficients to store, so it should be above `sq`.

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -1202,6 +1206,182 @@ TEST_F(GenPrismTest, adjacent_twisted)
     EXPECT_VEC_EQ(expected_node_strings, node_strings);
 }
 
+//---------------------------------------------------------------------------//
+// Involute
```
sethrj: So close to alphabetical.. `INVO` comes after `INFW`.

test/orange/surf/SoftSurfaceEqual.test.cc
```diff
@@ -192,6 +192,64 @@ TEST_F(SoftSurfaceEqualTest, general_quadric)
         softeq_(ref, SurfaceTranslator(Translation{{large, 0, 0}})(ref)));
 }
 
+TEST_F(SoftSurfaceEqualTest, involute)
```
sethrj, line 195: Since we're reordering the surface type enum for consistency, we might as well update the ordering here too...

src/orange/orangeinp/detail/SurfaceHashPoint.hh
sethrj: Also move this before `SimpleQuadric` for consistent ordering.

sethrj (changes requested):
I think this is pretty close. Please check for consistency in ordering of all the surface and object variant helpers and tests.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -646,6 +698,69 @@ void InfWedge::output(JsonPimpl* j) const
...
+        insert_surface(Sense::outside,
+                       celeritas::Involute{
+                           {0, 0},
+                           radii_[0],
+                           eumod(displacement_angle_[1], 2 * constants::pi),
+                           sign_,
+                           t_bounds_[0],
+                           t_bounds_[1] + displacement_angle_[1]
+                               - displacement_angle_[0]});
+    }
```
sethrj: This can be simplified quite a bit. I think this should work:
```suggestion
    // Make an inside and outside involute
    auto sense = (sign_ == Sign::clockwise ? Sense::outside : Sense::inside);
    for (auto i : range(2))
    {
        insert_surface(sense,
                       celeritas::Involute{
                           {0, 0},
                           radii_[0],
                           eumod(displacement_angle_[i], 2 * constants::pi),
                           sign_,
                           t_bounds_[0],
                           t_bounds_[1] + displacement_angle_[1]
                               - displacement_angle_[0]});
        sense = flip_sense(sense);
    }
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -646,6 +698,69 @@ void InfWedge::output(JsonPimpl* j) const
...
+    insert_surface(Sense::outside, PlaneZ{-hh_});
+    insert_surface(Sense::inside, PlaneZ{hh_});
+    insert_surface(Sense::inside, CCylZ{radii_[2]});
+    insert_surface(Sense::outside, CCylZ{radii_[1]});
```
sethrj: For consistency with the other shapes, order these as (lower z, upper z, inner cyl, outer cyl, involute 0, involute 1)

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -13,6 +13,8 @@
 #include "corecel/math/Turn.hh"
 #include "orange/OrangeTypes.hh"
 
+#include "../surf/detail/InvoluteSolver.hh"
```
sethrj: Let's not include the involute solver detail in this unrelated class. Let's add a new enum to `OrangeTypes`:
```c++
/*!
 * Chirality of a twirly object (currently only Involute).
 */
enum class Chirality : bool
{
    left, //!< Sinistral, spiraling counterclockwise
    right, //!< Dextral, spiraling clockwise
};
```

You can update the existing typedefs to reduce the amount of code you have to change:
```c++
    using Sign = Chirality;
    using ccw = Chirality::left;
    using cw = Chirality::right;
```

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -317,6 +319,52 @@ class InfWedge final : public IntersectRegionInterface
     Turn interior_;
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * An involute centered on the origin.
+ *
```
sethrj: More documentation please 😄 
```suggestion
 * An involute "blade" centered on the origin.
 *
 * This is the intersection of two parallel involutes with a cylindrical shell.
 * The three radii, which must be in ascending order, are that of the involute,
 * the inner cylinder, and the outer cylinder.
```

src/orange/surf/Involute.hh
```diff
@@ -98,7 +98,10 @@ class Involute
     CELER_FUNCTION real_type r_b() const { return std::fabs(r_b_); }
 
     //! Displacement angle
-    CELER_FUNCTION real_type a() const { return a_; }
+    CELER_FUNCTION real_type displacement_angle() const
+    {
+        return displacement_angle_;
```
sethrj: All I really care about is the public accessors for these (what you'd see from someone calling this class): you can leave the internal implementation as `a_` since it's well described internally.

src/orange/surf/SurfaceIO.cc
```diff
@@ -71,6 +71,24 @@ std::ostream& operator<<(std::ostream& os, GeneralQuadric const& s)
...
+        os << "Involute cw: r=" << s.r_b()
+           << ", a=" << constants::pi - s.displacement_angle() << ", t={"
+           << s.tmin() << ',' << s.tmax() << '}' << " at " << s.origin();
+    }
+    else
+    {
+        os << "Involute ccw: r=" << s.r_b() << ", a=" << s.displacement_angle()
+           << ", t={" << s.tmin() << ',' << s.tmax() << '}' << " at "
+           << s.origin();
+    }
```
sethrj: Avoid copy-pasting.
```suggestion
    real_type a = s.displacement_angle();
    os << "Involute ";
    if (s.sign() == Involute::Sign::clockwise)
    {
        os << "cw";
        a = constants::pi - s.displacement_angle();
    }
    os << ", a=" << a << ", t={" << s.tmin() << ',' << s.tmax() << "} at x=" << s.origin()[0] << ", y=" << s.origin()[1];
```

src/orange/surf/Involute.hh
```diff
@@ -59,17 +59,19 @@ class Involute
     //! \name Type aliases
     using Intersections = Array<real_type, 3>;
     using StorageSpan = Span<real_type const, 6>;
-    using Sign = detail::InvoluteSolver::Sign;
+    using Sign = Chirality;
+    Sign ccw = Chirality::left;
+    Sign cw = Chirality::right;
```
sethrj: You've accidentally declared these as member data, they should be `static constexpr inline`.

src/orange/surf/detail/InvoluteSolver.hh
```diff
@@ -50,11 +50,9 @@ class InvoluteSolver
...
 
     //! Enum defining chirality of involute
-    enum Sign : bool
-    {
-        counterclockwise = 0,
-        clockwise,  //!< Apply symmetry when solving
-    };
+    using Sign = Chirality;
+    Sign ccw = Chirality::left;
+    Sign cw = Chirality::right;
```
sethrj: Same here

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -35,6 +36,10 @@ namespace orangeinp
 {
 namespace test
 {
+//! Enum defining chirality
+using Sign = celeritas::Chirality;
+Sign ccw = celeritas::Chirality::left;
+Sign cw = celeritas::Chirality::right;
```
sethrj:
```suggestion
inline constexpr Sign ccw = celeritas::Chirality::left;
inline constexpr Sign cw = celeritas::Chirality::right;
```

sethrj (commented):
Forgot to post these before the end of your internship. I'll polish this PR off so we can merge it. Thanks and good luck!

sethrj:
Well, the GPU performance (even when not using involutes) seems to take a substantial hit (~40% longer for testem3 without involutes!) from this branch. It's easy to see why from looking at the occupancy change in the `initialize-tracks` kernel:

```diff
 {
 "const_mem": 0,
 "heap_size": 8388608,
-"local_mem": 152,
-"max_blocks_per_cu": 5,
+"local_mem": 160,
+"max_blocks_per_cu": 2,
 "max_threads_per_block": 256,
-"max_warps_per_eu": 40,
+"max_warps_per_eu": 16,
 "name": "initialize-tracks",
-"num_regs": 48,
-"occupancy": 0.625,
+"num_regs": 92,
+"occupancy": 0.25,
 "print_buffer_size": 5242880,
 "stack_size": 1024,
 "threads_per_block": 256
```
and along-step-neutral:
```
 {
 "const_mem": 0,
 "heap_size": 8388608,
-"local_mem": 0,
+"local_mem": 288,
 "max_blocks_per_cu": 2,
 "max_threads_per_block": 256,
 "max_warps_per_eu": 16,
 "name": "along-step-neutral",
-"num_regs": 123,
+"num_regs": 128,
 "occupancy": 0.25,
 "print_buffer_size": 5242880,
 "stack_size": 1024,
```
The local memory usage goes way up and the occupancy goes way down (where it can). For now I'll try disabling the runtime code paths and see if performance goes back to normal.

sethrj:
@VHLM2001 Great work this summer! As we discussed last Friday, it's disappointing but unsurprising that this ends up being a burden on the GPU. Next week (?) I can see if we can disable the code paths on GPU so that we can get this merged (and let someone else work on optimization).

sethrj:
I just ran before and after this branch; it seems the runtime "not reachable" is doing its job, but the CPU still suffers a runtime hit; not sure why since it should hit a "throw" into a cold section of the code. There should be no difference...
rel-throughput

sethrj:
OK with the latest update all is well at runtime!

rel-throughput

sethrj (approved):

PR #1348, opened 2024-07-31
Implement physics constructor for OpticalPhysics

drbenmorgan:
Celeritas intends to implement optical photon offload from Geant4 via custom `G4VProcess`es for photon-generating processed (Cerenkov, Scintillation). Within Geant4, this means they will need to be added to the relevant particles in application physics lists both to enable the offload and for optical data construction/availablity (as with the EM processes).

Following the pattern for used for EM physics in `CelerEMStandardPhysics`, the changes here add a preliminary concrete `G4VPhysicsConstructor` to handle particle and process construction for Optical processes. Though Celeritas will not generate and track G4OpticalPhotons, this initial implement largely follows the core `G4OpticalPhysics` constructor just to get something in place with `G4OpticalPhoton`. 

Comments are used, though it's fairly obvious, to note where the custom Celeritas processes will go. Adding this constructor in `CelerEmPhysicsList` and/or other lists and `celer-g4` can be done, but wanted to submit a WIP just to check this is on the right base.


drbenmorgan:
Though adding this to `CelerEmPhysicsList` and `CelerFTFPBert` could proceed at this point, it might be worth thinking about how best to handle the optical physics parameters. One gotcha the last few commits show is that `G4OpticalParameters` (analogous to `G4EmParameters`) is only present from Geant4 10.7. Before this, the parameters are handled locally in `G4OpticalPhysics`, which `CelerOpticalPhysics` attempts to emulate.

I think all of the optical parameters can be handled/transfered in the same way as the EM ones are in the `CelerEmStandardPhysics` constructor:

https://github.com/celeritas-project/celeritas/blob/cc2d1df1b3d609c00c12561e2dec60d179e7c012/src/celeritas/ext/detail/CelerEmStandardPhysics.cc#L98-L107

i.e. obtain them from the input `struct` and:

1. For Geant4 < 10.7, hold this locally and configure processes locally (see [how Geant4 did it](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-10.6-release/source/physics_lists/constructors/electromagnetic/src/G4OpticalPhysics.cc?ref_type=heads#L207-217))
2. Otherwise, call `G4OpticalParameters` setters as appropriate.

Some of this may of course not be needed later if we don't need to set these processes up to get data etc for Celeritas from them.

However, I don't think we can/should add the optical parameters as direct members of `GeantPhysicsOptions` as that's really focussed on EM options. It feels like we should maybe have a separate `GeantOpticalPhysicsOptions` struct that's a member of `GeantPhysicsOptions`? Longer term, could also consider a larger refactor so we'd have something like:

```c++
struct GeantPhysicsOptions
{
    GeantEmPhysicsOptions electromagnetic;
    GeantOpticalPhysicsOptions optical;
    // GeantHadronicPhysicsOptions hadronic;
};
```

but could just start with the optical parameter struct as a member?

sethrj:
@drbenmorgan I think your plan for a separate struct (and later creating another one for EM parameters) is perfect. Thanks for this!

drbenmorgan:
Struct in place, but as you'll see, the implementation of `CelerOpticalPhysics` is now a mess due to the changes in how optical physics was configured between Geant4 10.5-10.7. I'll have a think how to improve that.

Next is to add `GeantOpticalPhysicsOptions` as a member of `GeantPhysicsOptions`, but ran into some test failures as naturally the generated JSON changes. I did think about having it as a `std::optional` as we probably don't always want optical physics on. However, need to play about with that as I don't think `nlohmann::json` supports this directly (and not sure on Celeritas policy on use of `std::optional`). Advice/ideas welcome!

sethrj:
@drbenmorgan To make life easier for now, let's just stick to the newer versions, and we can throw a "not implemented" for the older versions if the experiments still need it by the time we've got sufficient optical physics implemented 😅

drbenmorgan:
Modulo fixing any errors in the CI, I think this is close to ready. I guess the last step is unit tests for when the optical physics are enabled (they are off by default), but am not 100% sure where to add these. [`GeantImporterTest.cc`](test/celeritas/ext/GeantImporterTest.cc) and the `LarSphere` test would appear to be appropriate, but I then get failures on `G4Cerenkov` and `G4Scintillation` not being exportable by `GeantImporter.cc`. I'm not sure about the higher level `celeb-sim/celer-g4` yet.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -0,0 +1,135 @@
...
+    //! Attach scintillation interaction information to generated photon
+    bool scint_track_info{false};
+    //!@}
+
+    //!@{
+    //! \name Boundary physics options
+    //! Invoke Geant4 SD at post step point if photon deposits energy
+    bool invoke_sd{false};
+    //!@}
+
```
sethrj: I think this would belong on the higher level setup rather than the physics options...

drbenmorgan:
Having played a bit with the tests as above, one slight interface smell (depending on point of view) with the use of `std::optional` is that setting options through the top level `GeantPhysicsOptions` feels a bit messy:

```c++
    GeantPhysicsOptions opts; // default, no optical options
    opts.optical_options.cerenkov_radiation = on; // whoops, not the correct interface, and is nullopt anyway

    // so have to either do
    GeantOpticalOptions optical; // gets you the defaults
    optical.cerenkov_radiation = false;
    opts.optical_options = optical;

    // or
    opts.optical_options = GeantOpticalOptions();
    (*opts.optical_options).cerenkov_radiation = false; // or .value()...
```

So I'm leaning towards @sethrj's suggestion to define an `explicit operator bool() const` for `GeantOpticalOptions` which would return `false` if none of the processes are activated. JSON to/from can work just the same (I think), but need to decide what default construction behaviour is, and from that the need for a "`build_default`" or "`deactivate`" member function to provide a simple switch.

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,277 @@
...
+    else if (process_name == "OpAbsorption")
+        return options_.absorption;
+    else if (process_name == "OpBoundary")
+        return options_.boundary;
+    else if (process_name == "OpMieHG")
+        return options_.mie_scattering;
+    else if (process_name == "OpRayleigh")
+        return options_.rayleigh_scattering;
+    else if (process_name == "OpWLS")
+        return options_.wavelength_shifting != WLSTimeProfileSelection::none;
```
sethrj: This  makes me think we might want to move toward changing both EM and optical physics to use a `std::set` of process names rather than an increasing number of bools... makes sense as-is for now though!

sethrj (commented):
Thanks @drbenmorgan , "optional optical options" does sound a little silly 😅 even though that's what it actually is.

I tend to agree with you that a standalone `operator bool` is the way to go. A couple thoughts:
- We should probably default the optical physics to `off` 
- We can add a `static OpticalPhysicsOptions all_physics()` so that users could assign `optical = ...`

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -0,0 +1,135 @@
...
+/*!
+ * Construction options for Geant optical physics.
+ *
+ * These options attempt to default to our closest match to \c
+ * G4OpticalPhysics from Geant4 10.5 onwards.
+ */
+struct GeantOpticalPhysicsOptions
+{
+    //!@{
+    //! \name Optical photon physics
```
stognini: Is this a placeholder or was it duplicated by mistake?

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,277 @@
...
+    auto scint = std::make_unique<G4Scintillation>();
+#if G4VERSION_NUMBER < 1070
+    scint->SetStackPhotons(options_.scint_stack_photons);
+    scint->SetTrackSecondariesFirst(options_.scint_track_secondaries_first);
+    scint->SetScintillationByParticleType(options_.scint_by_particle_type);
+    scint->SetFiniteRiseTime(options_.scint_finite_rise_time);
+    scint->SetScintillationTrackInfo(options_.scint_track_info);
+    // These two are not in 10.7 and newer, but defaults should be sufficient
+    // for now
+    // scint->SetScintillationYieldFactor(fYieldFactor);
```
stognini: Wait, so we're keeping these commented for now as a reminder to set them up later on? Is the scintillation yield factor here the same as, on newer G4 versions, the scintillation yield that is set via the material properties table?

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
drbenmorgan: That makes sense, though perhaps we leave for now as a reminder (I can comment out) until there's a mechanism to do this in Celeritas? I'd guess it'll eventually go in `SDSetupOptions` or similar.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
drbenmorgan: Yes, a duplication be mistake, so removed!

src/celeritas/ext/detail/CelerOpticalPhysics.cc
drbenmorgan: Possibly, I think @sethrj suggested that we focus on 10.7 and newer and revisit here if and only if experiments positively require support for the earlier versions. I'd have to look through the code to check on the yield being the same.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
drbenmorgan: O.k., latest commit implements the `operator bool` retaining the previous behaviour, so

- default construction enables all Optical processes
- The static function is "deactivated" rather than "all_physics"

So opposite to what you suggested @sethrj, but can easily flip this. It kind of felt more natural for default construction to enable everything, but I don't have a strong preference.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -0,0 +1,156 @@
...
+    //! \name Cerenkov physics options
+    //! Enable generation of Cerenkov photons
+    bool cerenkov_stack_photons{true};
+    //! Track generated photons before parent
+    bool cerenkov_track_secondaries_first{true};
+    //! Maximum number of photons that can be generated before limiting step
+    int cerenkov_max_photons{100};
+    //! Maximum percentage change in particle \f$\beta\f$  before limiting step
+    double cerenkov_max_beta_change{10.0};
+    //!@}
```
sethrj: We could avoid the scintillation/cerenkov bools and the prefixes here by doing another "optional" struct... but that might be too complicated (or we would end up taking it to the logical conclusion of a set of physics option classes in a hierarchy of their own 🤪)

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -136,6 +138,10 @@ struct GeantPhysicsOptions
 
     //! Print detailed Geant4 output
     bool verbose{false};
+
+    //! Optical physics options
+    GeantOpticalPhysicsOptions optical_options{
```
sethrj: Maybe optical? So we could end up with `options.optical.cerenkov.max_photons`?

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,277 @@
...
+{
+    auto* process_manager
+        = G4OpticalPhoton::OpticalPhoton()->GetProcessManager();
+    CELER_ASSERT(process_manager);
+
+    // Add Optical Processes
+    // TODO: Celeritas will eventually implement these directly (no
+    // G4OpticalPhotons) so how to set up on "Celeritas-side"
+    auto absorption = std::make_unique<G4OpAbsorption>();
+    if (process_is_active("OpAbsorption"))
```
sethrj: Does geant4 error if there's a typo in these names or just return "false"? It would be safer to have an enum plus string mapping since these strings are also repeated in `process_is_active`. We have something similar in `celeritas/io/ImportProcess.hh`...

Maybe for now we could make string constants local to this file just to avoid the repetition?

sethrj (commented):
Looks good @drbenmorgan , this is a somewhat tedious yet tricky and important task! I think the only substantial change to make would be to split off process-dependent options into another level of `struct`.

src/celeritas/ext/detail/CelerEmPhysicsList.cc
```diff
@@ -32,6 +33,14 @@ CelerEmPhysicsList::CelerEmPhysicsList(Options const& options)
     // Celeritas-supported EM Physics
     auto em_standard = std::make_unique<CelerEmStandardPhysics>(options);
     RegisterPhysics(em_standard.release());
+
+    // Celeritas-supported Optical Physics
```
stognini: Nitpick:

```suggestion
    if (options.optical_options)
    {
        // Celeritas-supported Optical Physics
```

src/celeritas/ext/detail/CelerOpticalPhysics.cc
stognini: It should return "NoProcess" (associated with enum `G4OpticalProcessIndex::kNoProcess`).
I think the enum string mapping makes sense too.

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,277 @@
...
+        CELER_LOG(debug) << "Loaded Optical wavelength shifting with G4OpWLS "
+                            "process";
+    }
+
+#if G4VERSION_NUMBER >= 1070
+    auto wls2 = std::make_unique<G4OpWLS2>();
+    if (process_is_active("OpWLS2"))
+    {
+        process_manager->AddDiscreteProcess(wls2.release());
+        // I need to check how this differs from G4OpWLS...
```
amandalund, line 262: It looks identical except that it uses separate optical properties... so I guess it lets you have two different wavelength shifting processes in the same material?

stognini: I did not realize WLS2 was akin to the scintillation components. That makes sense and we might consider generalizing it like we did with the scintillation components. 

PS: As an e.g., the NOvA detectors use a mineral oil with a dopant that scintillates and at least one other that produces a wavelength shifting. The scintillation/shifted photons get captured by an optical fiber that is also a wavelength-shifting fiber, so technically at least 2 WLS processes happen between scintillation and capture by the fiber that leads to a sensor.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
drbenmorgan: I think that's reasonable as it'd only be for the Cerenkov, Scintillation and Boundary processes. Happy to go ahead and do this if that's the prefered way.

src/celeritas/ext/detail/CelerOpticalPhysics.cc
drbenmorgan: [c4ba704](https://github.com/celeritas-project/celeritas/pull/1348/commits/c4ba704312bd749c8d5b4c0f8fbbea6b9b96a9be) should implement a _rough_ version of this. Everything's internal to `CelerOpticalPhysics.cc` for now, with only the minimal enum-enum/string mapping implemented.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
drbenmorgan: O.k. went ahead and did this as it was a pretty simple copy-paste exercise!

sethrj: Excellent, thanks Ben!

sethrj (approved):
I think this is great, @drbenmorgan , and a lot of painstaking work! Much appreciated.

amandalund (approved):
Nice work @drbenmorgan, thanks for this addition!

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,337 @@
...
+
+#include "corecel/Assert.hh"
+#include "corecel/io/Logger.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+namespace
+{
```
stognini, line 36:
```suggestion
{
//---------------------------------------------------------------------------//
/*!
 * Optical physics process type.
 *
 * See Geant4's \c G4OpticalProcessIndex in G4OpticalParameters.hh for the equivalent enum.
 */
```

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -0,0 +1,211 @@
...
+           && a.stack_photons == b.stack_photons
+           && a.track_secondaries_first == b.track_secondaries_first
+           && a.by_particle_type == b.by_particle_type
+           && a.finite_rise_time == b.finite_rise_time
+           && a.track_info == b.track_info;
+    // clang-format on
+}
+
+//---------------------------------------------------------------------------//
+//! Optical Bounrary process options
```
stognini, line 97:
```suggestion
//! Optical Boundary process options
```

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,337 @@
...
+        case OpticalProcessType::rayleigh:
+            return options.rayleigh_scattering;
+        case OpticalProcessType::mie_hg:
+            return options.mie_scattering;
+        case OpticalProcessType::boundary:
+            return bool(options.boundary);
+        case OpticalProcessType::wavelength_shifting:
+            return options.wavelength_shifting != WLSTimeProfileSelection::none;
+        case OpticalProcessType::wavelength_shifting_2:
+            // Not supported pre 10.7
```
stognini, line 117: I know we're not worrying about any version < 10.7, but should we throw a `CELER_ASSERT_UNREACHABLE();` here?

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+#if G4VERSION_NUMBER >= 1070
+    auto wls2 = std::make_unique<G4OpWLS2>();
+    if (process_is_active(OpticalProcessType::wavelength_shifting_2, options_))
+    {
+        process_manager->AddDiscreteProcess(wls2.release());
+        // I need to check how this differs from G4OpWLS...
+        CELER_LOG(debug)
+            << "Loaded Optical wavelength shifting V2 with G4OpWLS2 "
```
stognini, line 264: Since WLS and WLS2 are identical, maybe

```suggestion
            << "Loaded second optical wavelength shifting with G4OpWLS2 process "
```

drbenmorgan:
@stognini I'll implement your suggestions in a fresh PR!

sethrj:
Sorry I pulled the trigger too soon!

stognini:
No worries. Everything was looking great, so was just having a final look in case I spotted any minor detail after yesterday's commits. Thanks a lot @drbenmorgan !

PR #1350, opened 2024-08-01
Automatically build and deploy github documentation

### LINKED ISSUE #698: Improve CI infrastructure
- ~~[Dockerhub is getting rid of the free teams tier](https://www.docker.com/blog/we-apologize-we-did-a-terrible-job-announcing-the-end-of-docker-free-teams/) but maybe our images will stay live?~~ Apparently docker is [no longer ending it](https://www.docker.com/blog/no-longer-sunsetting-the-free-team-plan/)
- Kokkos builds block the single HIP executor and each one of those seems to take 2–6 hours (@dalg24 ) whereas our pipelines take <10 minutes
- Jenkins/blueocean annoys the shit out of me with the number of clicks it takes to get to a broken test and how it autorefreshes in the background
- Our github pages website requires manual updating

Ideally we could:
- [x] Auto-publish documentation (moved from https://github.com/celeritas-project/celeritas/issues/592)
- [ ] View results on CDash instead of jenkins/blue ocean
- [x] Use github actions for better github integration (#1020)
- [x] Not have to maintain our own machines for HIP and CUDA, nor to share a single machine with Kokkos

Resources:
- [Dockerfile support for GitHub Actions](https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions)
- [Creating a Docker container action](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action)
- [Apply for docker open source tier](https://www.docker.com/community/open-source/application/)
- [Building CI with CDash](https://embeddeduse.com/2022/05/02/building-a-ci-pipeline-with-ctest-and-cdash/)

TODO for Github CI:
- [ ] Fix the issues with examples
  * [ ] asan build needs `-fsanitize=address` in downstream example
  * [ ] rocm+ndebug (static libs) fails to propagate the HIP library linkage
  * [x] SimpleOffload::Build crashes because ~~MTLogger is called before run manager completes initialization~~ default run manager is "task" which returns zero threads on the first call
  * [ ] VecGeom build on Ubuntu requires `-Wl,--no-as-needed` which we provide to Celeritas via `CMAKE_{SHARED|EXE}_LINKER_FLAGS`: maybe @pcanal 's RDC code needs to propagate these? Should this be vecgeom's responsibility?
- [x] Use a shallow clone and `git describe` or some better way of fetching tags
- [ ] Improve docker images
  * Add ccache to the build image and use it
  * Avoid custom user ID; GitHub requires root
  * Install all supported geant4 versions for a small build matrix
  * Fix entry point so that we don't have to manually source `/etc/profile` and re-export variables
- [x] Add separate build matrix for geant4 versions (to test minimum and maximum supported)
- [x] Add ~~nightly~~ minimal build matrix for various compilers on raw hardware (earliest GCC through latest, multiple clang versions)
- [x] Add automatic clang formatting and/or fix?
- [x] Add documentation
- [x] Only do a simple build for "draft" PRs

sethrj:
This fixes a couple of small doxygen issues, and it adds a github action that combines the user and dev docs and publishes them to https://celeritas-project.github.io/celeritas/ when the `develop` branch is pushed.

Closes https://github.com/celeritas-project/celeritas/issues/698


.github/workflows/doc.yml
```diff
@@ -97,6 +98,4 @@ jobs:
           overwrite: true
           retention-days: 7
```
pcanal: Should it also go down to 1 day?

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,55 @@
...
+  build-docs:
+    uses: ./.github/workflows/doc.yml
+  build-site:
+    runs-on: ubuntu-latest
+    needs: build-docs
+    steps:
+      - name: Check out Celeritas doc base
+        uses: actions/checkout@v4
+        with:
+          ref: doc/gh-pages-base
```
pcanal, line 24: Do we have a note of this branch existence in a onboarding doc for (new) developers?

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/download-artifact@v4
+      - name: Rename artifacts
+        run: |
+          mv doc-dev dev
+          mv doc-user user
+      - name: Build with Jekyll
+        uses: actions/jekyll-build-pages@v1
+        with:
+          source: ./
+          destination: ./_site
```
pcanal: Out of curiosity (our source does not seem to have mention of `_site`), whose convention is this name from? (humm https://github.com/actions/jekyll-build-pages seems to say that this is the default value anyway).

.github/workflows/doc.yml
sethrj: I figured the user documentation, since it's thinner than the dev (which contains perhaps hundreds of MB of include/use maps), can stay around longer. It's also more likely to be kept and uploaded with version releases.

.github/workflows/deploy-pages.yml
sethrj: I've added a note about it to the administration section. Basically no one should need to know about or touch this branch, since it is only there to redirect to the dev/user docs. Someday we could update it to include release notes, people description, etc.

.github/workflows/deploy-pages.yml
sethrj: This was part of the default build template github provided...

sethrj:
Ping @pcanal :)

.github/workflows/deploy-pages.yml
pcanal: That ends up confusing me :).  By specifying it, one would assume it is not the default and then the next step look 'weird' (needs lookup of the doc to understand whether it works or not.

I would recommend to either note explicitly it is the default (possibly even not set it, just say where it will be created) **or** also add the corresponding setting in the upload-pages-artifact part.

.github/workflows/deploy-pages.yml
pcanal: > I've added a note about it to the administration section.

Thanks.

> Basically no one should need to know about or touch this branch

Which also means the project will eventually forgot about it existence until something breaks :)  (But this is fixes by adding to the administration section).

.github/workflows/doc.yml
pcanal: Ok.  It might help our future self to add some comments justify the choices.

sethrj:
Thanks for the feedback @pcanal ! I also added documentation about the "approval status" so that in the future you can treat the approval button as "accept with minor revisions" and save me from having to ping you quite so often 😉

pcanal (approved):

PR #1352, opened 2024-08-02
Fix clang-tidy checks, including some use-after-move

esseivaju:
Fix a few `clang-tidy` checks. There were a few use-after-move in `LocalTransporter.cc`, `Solid.cc` and `ValueGridBuilder.cc` that could lead to bugs. In addition, ignore some checks making a lot of noise. The remaining warnings are mostly false alerts and could use a `// NOLINT` comment explaining why the code is fine.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -12,13 +12,21 @@ Checks: |
   -cppcoreguidelines-pro-bounds-pointer-arithmetic
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
+  -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+  -cppcoreguidelines-avoid-const-or-ref-data-members
+  -cppcoreguidelines-pro-bounds-constant-array-index
   bugprone-*
   -bugprone-sizeof-expression
   -bugprone-narrowing-conversions
   -bugprone-macro-parentheses
+  -bugprone-easily-swappable-parameters
+  -bugprone-implicit-widening-of-multiplication-result
 CheckOptions:
   cppcoreguidelines-macro-usage.CheckCapsOnly: true
   cppcoreguidelines-avoid-do-while.IgnoreMacros: true
+  cppcoreguidelines-rvalue-reference-param-not-moved.IgnoreNonDeducedTemplateTypes: true
 HeaderFilterRegex: ''
 FormatStyle: file
 ...
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -307,7 +307,7 @@ void DetectorConstruction::ConstructSDandField()
  * Apply a function to the range of volumes for each detector.
  */
 template<class F>
-void DetectorConstruction::foreach_detector(F const& apply_to_range) const
+void DetectorConstruction::foreach_detector(F&& apply_to_range) const
 {
     auto start = detectors_.begin();
     while (start != detectors_.end())
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -319,7 +319,7 @@ void DetectorConstruction::foreach_detector(F&& apply_to_range) const
             ++stop;
         } while (stop != detectors_.end() && start->first == stop->first);
 
-        apply_to_range(start, stop);
+        std::forward<F>(apply_to_range)(start, stop);
         start = stop;
     }
 }
--- a/app/celer-g4/DetectorConstruction.cc
+++ b/app/celer-g4/DetectorConstruction.cc
@@ -307,7 +307,7 @@ void DetectorConstruction::ConstructSDandField()
  * Apply a function to the range of volumes for each detector.
  */
 template<class F>
-void DetectorConstruction::foreach_detector(F&& apply_to_range) const
+void DetectorConstruction::foreach_detector(F const& apply_to_range) const
 {
     auto start = detectors_.begin();
     while (start != detectors_.end())
@@ -319,7 +319,7 @@ void DetectorConstruction::foreach_detector(F&& apply_to_range) const
             ++stop;
         } while (stop != detectors_.end() && start->first == stop->first);
 
-        std::forward<F>(apply_to_range)(start, stop);
+        apply_to_range(start, stop);
         start = stop;
     }
 }
--- a/src/geocel/ScopedGeantLogger.cc
+++ b/src/geocel/ScopedGeantLogger.cc
@@ -113,7 +113,7 @@ class GeantLoggerAdapter : public G4coutDestination
   public:
     // Assign to Geant handlers on construction
     GeantLoggerAdapter();
-    ~GeantLoggerAdapter();
+    ~GeantLoggerAdapter() override;
 
     // Handle error messages
     G4int ReceiveG4cout(G4String const& str) final;
--- a/src/geocel/ScopedGeantLogger.cc
+++ b/src/geocel/ScopedGeantLogger.cc
@@ -108,12 +108,12 @@ G4int log_impl(G4String const& str, LogLevel level)
 /*!
  * Send Geant4 log messages to Celeritas' world logger.
  */
-class GeantLoggerAdapter : public G4coutDestination
+class GeantLoggerAdapter final : public G4coutDestination
 {
   public:
     // Assign to Geant handlers on construction
     GeantLoggerAdapter();
-    ~GeantLoggerAdapter() override;
+    ~GeantLoggerAdapter() final;
 
     // Handle error messages
     G4int ReceiveG4cout(G4String const& str) final;
--- a/src/orange/orangeinp/detail/ProtoBuilder.cc
+++ b/src/orange/orangeinp/detail/ProtoBuilder.cc
@@ -24,16 +24,16 @@ namespace detail
  */
 ProtoBuilder::ProtoBuilder(OrangeInput* inp,
                            ProtoMap const& protos,
-                           Options&& opts)
+                           Options const& opts)
     : inp_{inp}
     , protos_{protos}
-    , save_json_{std::move(opts.save_json)}
+    , save_json_{opts.save_json}
     , bboxes_{protos_.size()}
 {
     CELER_EXPECT(inp_);
     CELER_EXPECT(opts.tol);
 
-    inp_->tol = std::move(opts).tol;
+    inp_->tol = opts.tol;
     inp_->universes.reserve(protos_.size());
 }
 
--- a/src/orange/orangeinp/detail/ProtoBuilder.cc
+++ b/src/orange/orangeinp/detail/ProtoBuilder.cc
@@ -33,7 +33,7 @@ ProtoBuilder::ProtoBuilder(OrangeInput* inp,
     CELER_EXPECT(inp_);
     CELER_EXPECT(opts.tol);
 
-    inp_->tol = opts.tol;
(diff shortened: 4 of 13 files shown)


app/celer-g4/DetectorConstruction.cc
```diff
@@ -319,7 +319,7 @@ void DetectorConstruction::foreach_detector(F&& apply_to_range) const
             ++stop;
         } while (stop != detectors_.end() && start->first == stop->first);
 
-        apply_to_range(start, stop);
+        std::forward<F>(apply_to_range)(start, stop);
```
sethrj: Since we're applying this multiple times, I don't think we want to forward it

src/geocel/ScopedGeantLogger.cc
```diff
@@ -113,7 +113,7 @@ class GeantLoggerAdapter : public G4coutDestination
   public:
     // Assign to Geant handlers on construction
     GeantLoggerAdapter();
-    ~GeantLoggerAdapter();
+    ~GeantLoggerAdapter() override;
```
sethrj: Why is this override? It's not like you can declare a destructor "final"...

src/orange/orangeinp/detail/ProtoBuilder.cc
```diff
@@ -33,7 +33,7 @@ ProtoBuilder::ProtoBuilder(OrangeInput* inp,
     CELER_EXPECT(inp_);
     CELER_EXPECT(opts.tol);
 
-    inp_->tol = opts.tol;
+    inp_->tol = std::move(opts).tol;
```
sethrj: Maybe we should just make the options `const&` since there's no heavyweight data in it anyway.

src/geocel/ScopedGeantLogger.cc
esseivaju: To document/verify that the base class has a virtual dtor. This is following the discussion https://github.com/isocpp/CppCoreGuidelines/pull/1448 `virtual` instead of `override` is also fine.

app/celer-g4/DetectorConstruction.cc
esseivaju: Do we need the forwarding reference then?

sethrj: If F is a class with mutable operator() I don't think constructing ref will work?

src/geocel/ScopedGeantLogger.cc
sethrj: That's weird. The parent class doesn't have a `~GeantLoggerAdapter` method. Maybe next c++ version will name the destructor `__del__` so you can actually tell it's overriding... I think I prefer `virtual` here since it says "polymorphic and safe to destruct" rather than "I'm replacing what the parent classdoes".

esseivaju: The parent `G4coutDestination` still has a virtual destructor. Even though technically destructor override their parent's I agree that semantically `virtual` makes sense since they are chained and don't replace the parent's behavior.  However, the advantage of `override` is that it will result in an error if the parent dtor is no longer virtual. Using `virtual` would silently redeclare it virtual in the derived which would be confusing.

Anyway, in that specific case, it's not really important since this is a TU-local class but in general I prefer the destructor be declared override if it needs to be defined again in the derived class for the reason above. 

Maybe for that case we could declare the class final? It's local anyway and would silence the warning

sethrj: Declaring as `final` sounds good to me, and now that I know about this behavior I think `override` for virtual destructors is sensible. (It's partly weird because the daughter classes will automatically define virtual `override` destructors...)

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
...
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+  -cppcoreguidelines-avoid-const-or-ref-data-members
+  -cppcoreguidelines-pro-bounds-constant-array-index
   bugprone-*
   -bugprone-sizeof-expression
   -bugprone-narrowing-conversions
   -bugprone-macro-parentheses
+  -bugprone-easily-swappable-parameters
+  -bugprone-implicit-widening-of-multiplication-result
```
sethrj, line 25: ✅ Most of these seem to be about comparing a `std::size_t` against a `celeritas::size_type`:
```
/home/runner/work/celeritas/celeritas/src/celeritas/field/RZMapFieldParams.cc:47:31: warning: performing an implicit widening conversion to type 'unsigned long' of a multiplication performed in type 'unsigned int' [bugprone-implicit-widening-of-multiplication-result]
        inp.field_z.size() == inp.num_grid_z * inp.num_grid_r,
```

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
...
+  -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+  -cppcoreguidelines-avoid-const-or-ref-data-members
+  -cppcoreguidelines-pro-bounds-constant-array-index
   bugprone-*
   -bugprone-sizeof-expression
   -bugprone-narrowing-conversions
   -bugprone-macro-parentheses
+  -bugprone-easily-swappable-parameters
```
sethrj, line 24: ✅ These seem over the top:
```
/home/runner/work/celeritas/celeritas/src/orange/orangeinp/IntersectRegion.cc:217:20: warning: 2 adjacent parameters of 'Cylinder' of similar type ('celeritas::real_type') are easily swapped by mistake [bugprone-easily-swappable-parameters]
Cylinder::Cylinder(real_type radius, real_type halfheight)
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
   -cppcoreguidelines-pro-bounds-pointer-arithmetic
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
+  -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+  -cppcoreguidelines-avoid-const-or-ref-data-members
+  -cppcoreguidelines-pro-bounds-constant-array-index
```
sethrj, line 19: I don't see this last warning showing up anywhere in the output 😕

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
   -cppcoreguidelines-pro-bounds-pointer-arithmetic
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
+  -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
```
sethrj, line 17: ✅ These are silly and mostly seem to be about decaying a `static char const arr[]` 
```
/home/runner/work/celeritas/celeritas/src/corecel/io/BuildOutput.cc:31:34: warning: do not implicitly decay an array into a pointer; consider using gsl::array_view or an explicit cast instead [cppcoreguidelines-pro-bounds-array-to-pointer-decay]
    obj["version"] = std::string(celeritas_version);
                                 ^
```

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
   -cppcoreguidelines-pro-bounds-pointer-arithmetic
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
+  -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-pro-type-const-cast
```
sethrj, line 16: ✅ Can't help these:
```
/home/runner/work/celeritas/celeritas/src/geocel/GeantGeoUtils.cc:128:19: warning: do not use const_cast [cppcoreguidelines-pro-type-const-cast]
    auto& touch = const_cast<GeantTouchableBase&>(*pnh.touch);
                  ^
```

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
   -cppcoreguidelines-pro-bounds-pointer-arithmetic
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
+  -cppcoreguidelines-avoid-do-while
```
sethrj, line 15: I know I saw these once but don't see them in the latest [clang-tidy output from develop](https://productionresultssa4.blob.core.windows.net/actions-results/1e27f446-2749-4de8-934c-49550066eb6b/workflow-job-run-0f44d42c-80fb-5772-bde9-fe33f2c238e7/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-08-05T11%3A39%3A37Z&sig=Q%2FXD0qidw1S%2B4D770tQPN7ZxQpV%2FtCB3xtZO9WedILE%3D&ske=2024-08-05T19%3A29%3A48Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-08-05T07%3A29%3A48Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-05-04&sp=r&spr=https&sr=b&st=2024-08-05T11%3A29%3A32Z&sv=2024-05-04)...

.clang-tidy
```diff
@@ -12,13 +12,21 @@ Checks: |
...
+  -cppcoreguidelines-pro-type-const-cast
+  -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+  -cppcoreguidelines-avoid-const-or-ref-data-members
```
sethrj, line 18: I don't see this warning in the output either...

sethrj (approved):
Thanks @esseivaju ! Looks like there's only 20 warnings left, after those are gone we can make "warning" fatal 😄

.clang-tidy
esseivaju: that one seems present:
```
/home/runner/work/celeritas/celeritas/src/celeritas/em/params/WentzelOKVIParams.cc:948:12: warning: do not use array subscript when the index is not an integer constant expression [cppcoreguidelines-pro-bounds-constant-array-index]
    return mott_coeffs[index];
```

.clang-tidy
esseivaju: also see that locally, e.g.:
```
/home/esseivaj.linux/devel/celeritas/src/celeritas/geo/GeoMaterialParams.cc:142:22: warning: member 'geo_' of type 'const GeoParams &' (aka 'const celeritas::OrangeParams &') is a reference [cppcoreguidelines-avoid-const-or-ref-data-members]
  142 |     GeoParams const& geo_;
```

.clang-tidy
esseivaju: Mmh right, I'm not sure none is showing up, I see it locally: 
```
[144/338] Building CXX object src/celeritas/CMakeFiles/celeritas_hepmc.dir/io/EventReader.cc.o
/home/esseivaj.linux/devel/celeritas/src/celeritas/io/EventReader.cc:53:9: warning: avoid do-while loops [cppcoreguidelines-avoid-do-while]
   53 |         do
```

I'm using clang 17 so this might explain the differences.

.clang-tidy
esseivaju: Right, this one has quite a [few options](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/easily-swappable-parameters.html), maybe at 5 or 6 parameter it is a bit less over the top

esseivaju (commented):
Even though some checks are not showing up on the CI, they'll show up in my environment. If we don't want these checks I suppose we can keep them disabled for future-proofing when we switch to ubuntu-24.04 runners.

PR #1354, opened 2024-08-05
Refactor rejection sampling using a helper class

sethrj:
This defines a `RejectionSampler` class that can be used as a more intuitive replacement for `BernoulliDistribution` during rejection sampling loops. Instead of "accepting" when the sample is within bounds (necessitating an easy-to-miss negation in the `while` condition) we "reject" by returning true unless the sampled value is accepted.


src/celeritas/random/distribution/RejectionSampler.hh
```diff
@@ -0,0 +1,99 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using real_type = RealType;
+    using result_type = real_type;
+    //!@}
+
+  public:
+    // Construct on an arbitrary interval
+    inline CELER_FUNCTION RejectionSampler(real_type a, real_type b);
```
amandalund:
```suggestion
    // Construct with acceptance probability f/fmax
    inline CELER_FUNCTION RejectionSampler(real_type f, real_type fmax);
```

src/celeritas/random/distribution/RejectionSampler.hh
```diff
@@ -0,0 +1,99 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Return whether a rejection loop needs to continue trying.
+ *
+ * A common implementation of sampling from a "difficult" (non-analytically
+ * invertible) probability distribution function is to bound the difficult
+ * distribution \em f(x) with another easily sampled function \em g(x) . Given
+ a
+ * maximum value \em M over the \em x interval being sampled, it is equivalent
```
amandalund, line 28: Adjust formatting?

test/celeritas/random/distribution/RejectionSampler.test.cc
```diff
@@ -0,0 +1,74 @@
...
+
+#include "DiagnosticRngEngine.hh"
+#include "celeritas_test.hh"
+
+namespace celeritas
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+// Target PDF with maximum 1.5
```
amandalund:
```suggestion
// Target PDF with maximum 2
```

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
```diff
@@ -149,15 +149,14 @@ CELER_FUNCTION Interaction BraggICRU73QOInteractor::operator()(Engine& rng)
...
     real_type energy;
-    real_type target;
     do
     {
         energy = min_secondary_energy_ * max_secondary_energy_
                  / UniformRealDistribution(min_secondary_energy_,
                                            max_secondary_energy_)(rng);
-        target = 1 - beta_sq_ * energy / max_secondary_energy_;
-    } while (!BernoulliDistribution(target)(rng));
+    } while (BernoulliDistribution(energy * acceptance)(rng));
```
amandalund: `acceptance` might be confusing since `energy * acceptance` is the rejection probability, right?

amandalund (approved):
Thanks @sethrj! Hopefully this helps prevent any more rejection sampling bugs going forward.

src/celeritas/em/interactor/BraggICRU73QOInteractor.hh
sethrj: Yeah that was a dumb change, I'm reverting

PR #1357, opened 2024-08-06
Improve use of distributions and simplify interactions

sethrj:
Following on to #1354 , this PR improves readability of the physics models by adding a new distribution and refactoring some "manual" distributions in terms of other tested distributions. I also found one or two more places that could use a RejectionDistribution, which will also catch any errors we might've made in the envelope assumption (e.g. if the lowest muon cross sections is *not* the largest one).

- Add an `InverseSquareDistribution` to sample $`1/E^2`$
- Replace MuBB, BraggICRU73QO with inverse square distribution
- Replace MuBremsstrahlung with reciprocal distribution, which allows us to mathematically eliminate the "min cutoff energy"
- Fix variable name in KleinNishinaInteractor
- Remove unused class data from moller-bhabha
- Replace log canonical with existing exponential sampler in optical scintillation generator
- Replace bernoulli with rejection there as well
- Refactor and add scintillation tests for the "zero rise time" case and made sure that the above changes were correct


src/celeritas/random/distribution/InverseSquareDistribution.hh
```diff
@@ -0,0 +1,87 @@
...
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+
+#include "UniformRealDistribution.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample \f$ 1/x^2 \f$ over a givendomain.
```
amandalund:
```suggestion
 * Sample \f$ 1/x^2 \f$ over a given domain.
```

src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
```diff
@@ -80,21 +81,13 @@ class MuBremsstrahlungInteractor
...
+    ReciprocalDistribution<> sample_energy_;
     // Envelope distribution for rejection sampling of gamma energy
     real_type envelope_;
 
     //// CONSTANTS ////
 
-    //! Minimum allowed secondary cutoff energy [MeV]
-    static CELER_CONSTEXPR_FUNCTION Energy min_cutoff_energy()
-    {
-        return units::MevEnergy{9e-4};  //!< 0.9 keV
```
amandalund, line 95: Just curious why this constant was removed? Is the reciprocal distribution equivalent to what's used here to sample the energy if the production cut is larger than this value?.

amandalund (commented):
Looks great @sethrj, nice improvements! Just one little question about the mu brems interactor.

src/celeritas/em/interactor/MuBremsstrahlungInteractor.hh
sethrj: The reciprocal distribution is exactly mathematically equivalent: it turns out the min cutoff doesn't factor into the sample at all. I just refactored it a bit and simplified it so that it was the reciprocal sampling expression, and surprisingly that value simply disappeared. (The original expression had a couple of extra unnecessary `log`s.)

amandalund: gotcha, I see it now, thanks!

amandalund (approved):

PR #1358, opened 2024-08-06
Add optical physics documentation

sethrj:


doc/main/api/celeritas.rst
```diff
@@ -247,6 +247,105 @@ as properties of any secondary particles produced.
...
+
+     geant4-direct -->|primaries| photons
+     scintillation-gen -->|primaries| photons
+     cerenkov-gen -->|primaries| photons
+
+
+Optical materials
+~~~~~~~~~~~~~~~~~
+
+Optical volume and surface properties are imported from Geant4 (see below).
```
stognini, line 304: Should we discuss our plan on having a better generalization overall? We have generalized the number of scintillation components for both materials and particle types, but arbitrary limitations that we may find on other optical processes will be redesigned as well if that provides more capabilities to the users.

src/celeritas/optical/detail/OffloadGatherExecutor.hh
```diff
@@ -35,6 +35,9 @@ struct OffloadGatherExecutor
 //---------------------------------------------------------------------------//
 /*!
  * Gather pre-step data needed to generate optical distributions.
+ *
+ * TODO: check optical material first, skip storing if it's false. Also maybe
+ * skip storing for tracks that can't lose energy over the step?
```
amandalund, line 40: True we can do that, as long as we clear the step data in those cases.

amandalund (approved):
Thanks @sethrj!

src/celeritas/optical/detail/OffloadGatherExecutor.hh
sethrj: Yeah if we set the optical material to false then we don't have to load and store the other data? Maybe not bad in the big scheme of things but every fewer memory transaction counts 😉

doc/main/api/celeritas.rst
sethrj: There's plenty of stuff we *could* put in, but my priority for this PR is documenting what we *have*, then next the stuff we *will*, and eventually what we *could* 😅

PR #1359, opened 2024-08-07
Automatically add Zotero publications to project site

sethrj:
I've updated `celeritas-project.github.io` to redirect to this code's github page. I also found that Zotero has a JSON API that can be used to extract publication data. Combining that with github actions, we can now get an updated publication list on our web site every time we push to develop 🎉 

Preview [the updated web site](https://sethrj.github.io/celeritas/index.html)

Note that most of the implementation of this is in [a script in the docs/gh-pages-base branch](https://github.com/celeritas-project/celeritas/blob/doc/gh-pages-base/scripts/generate-pubs.py)

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -37,8 +37,10 @@ jobs:
           ZOTERO_GROUP: 2380941
         run: |
           python scripts/generate-pubs.py -g $ZOTERO_GROUP -o publications.md
+          echo "DIFF_COUNT=$(git diff -- publications.md | wc -l)" >> $GITHUB_OUTPUT
       - name: Push updated publications
-        if: ${{steps.update.outcome != 'failure'}}
+        if: ${{steps.update.outcome != 'failure'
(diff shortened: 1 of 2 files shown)


esseivaju:
That's nice! A few minor comments not directly related to that PR:

- I would keep the same font size when hovering a link
- DOE code page is out of date
- A dark theme option to avoid the flashbang would be great 😄

sethrj:
@esseivaju the DOE code page is the "official" code citation (though I notice now that the url/DOI doesn't show up, I'll fix that) and I'll see if I can update the membership on that. Maybe we should start looking at [zenodo releases](https://zenodo.org)?

The weird font change and the lack of dark/light are part of the github-provided theme (and it's an improvement from the [previous one](https://pages-themes.github.io/slate/), I think). Can you find one that's better? Happy to set it as default.

amandalund (approved):
That's cool! I think that theme looks best out of some [not](https://pages-themes.github.io/architect/) [great](https://pages-themes.github.io/merlot/) options ;). The old one's not bad either, but it's nice that this one has a spot for the logo.

PR #1360, opened 2024-08-08
Add documentation for config options, improve unit docs, and use doxygen todos

sethrj:
This also implements one of the to-do notes by changing the value of `RunnerInput::unspecified`.


amandalund (approved):
Thanks @sethrj!

PR #1362, opened 2024-08-09
Add oriented bounding zone class to be used for safety distance calculations

elliottbiondo:
This MR creates an `OrientedBoundingZone` class. This class, like a standard `BoundingZone`, consists of an inner and outer box, with the difference being that both boxes are transformed the same transformation. The orientation of the box should mean that tighter bounding boxes (w.r.t. each volume) can be stored during runtime and used for safety distance calculations. The safety distance is the minimum distance from a given point to the inner or outer box. When a point is between the two boxes, the safety distance is zero.


elliottbiondo:
@sethrj this is ready for your review

elliottbiondo:
(minor floating point issue causing one test failure in one build configuration, will fix in morning)

src/orange/OrangeData.hh
```diff
@@ -306,6 +306,40 @@ struct BIHTreeData
...
+template<Ownership W, MemSpace M>
+struct OrientedBoundingZoneData
+{
+    template<class T>
+    using Items = Collection<T, W, M>;
+
+    // Low-level storage
+    Items<Real3> half_widths;
+    Items<TransformRecord> transforms;
+    Items<real_type> reals;
```
sethrj: I don't think we want a fully separate bounding zone data from the orange params, because that prevents us from sharing the transforms+reals with the universes. (We'd have to duplicate everything.) I recall we discussed the merits of using double- or single-precision data for safety distances: we could potentially also use doubles for the transformation (which would let us share with the main part of the code) and singles for the half widths.

Currently the transformations aren't templated (they're all "real" type) so for now we can stick with double?

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/detail/OrientedBoundingZone.cc
```
sethrj: I think it's been too long since you wrote code in Celeritas 😉  this whole file has to be inlined into the header and can't use host-only `std` functions.

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+real_type OrientedBoundingZone::safety_distance_inside(Real3 pos)
+{
+    CELER_EXPECT(this->is_inside_outer(pos));
+
+    auto trans_pos = this->quadrant_zero(this->translate(pos));
+    auto inner_hw = storage_->half_widths[inner_hw_id_];
+
+    if (!this->is_inside(trans_pos, inner_hw))
+    {
+        return 0.;
```
sethrj: Avoid double-precision literals
```suggestion
        return 0;
```

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+
+    auto outer_hw = storage_->half_widths[outer_hw_id_];
+    return std::accumulate(
+        range(Axis::size_).begin(),
+        range(Axis::size_).end(),
+        std::numeric_limits<real_type>::infinity(),
+        [&](real_type min_dist, auto ax) {
+            return std::min(min_dist,
+                            outer_hw[to_int(ax)] - trans_pos[to_int(ax)]);
+        });
```
sethrj: We can't use `std::accumulate` since this has to be device compatible, and I think it's semantically wrong here. You're not accumulating a result, you're finding the minimum of a range of possibilities: use `min_element` (from corecel/math/Algorithms) maybe? Or just a simple loop.

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * Case 1: the point is between the inner and outer boxes, resulting in a
+ * safety distance of zero.
+ *
+ * Case 2: the point is outside both the inner and outer boxes, in which case
+ * the safety distance is the minimum distance from the given point to any
+ * point on the inner box. This can be calculated as:
+ *
+ * d = sqrt(max(0, px - hx)^2 + max(0, py - hy)^2 + max(0, pz - hz)^2)
+ *
```
sethrj:
```suggestion
 * \f[
 * d = \sqrt{\max(0, p_x - h_x)^2 + \max(0, p_y - h_y)^2 + \max(0, p_z - h_z)^2}
 * \f]
```

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+real_type OrientedBoundingZone::safety_distance_outside(Real3 pos)
+{
+    CELER_EXPECT(!this->is_inside_inner(pos));
+
+    auto trans_pos = this->quadrant_zero(this->translate(pos));
+    auto outer_hw = storage_->half_widths[outer_hw_id_];
+
+    if (this->is_inside(trans_pos, outer_hw))
+    {
+        return 0.;
```
sethrj: Same comment about double

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    auto inner_hw = storage_->half_widths[inner_hw_id_];
```
sethrj: Instead of repeating this expression (which assumes a particular data layout that we may want to change later), make a helper member function that returns a `Span<real_type const, 3>` or `Real3 const&` .

src/orange/detail/OrientedBoundingZone.cc
```diff
@@ -0,0 +1,190 @@
...
+ */
+Real3 OrientedBoundingZone::translate(Real3 const& pos)
+{
+    Real3 trans_pos;
+    TransformVisitor apply_transform(storage_->transforms, storage_->reals);
+    auto transform_down
+        = [&pos, &trans_pos](auto&& t) { trans_pos = t.transform_down(pos); };
+
+    apply_transform(transform_down, transform_id_);
+    return trans_pos;
```
sethrj:
```suggestion
    TransformVisitor apply_transform(storage_->transforms, storage_->reals);
    auto transform_down
        = [&pos, &trans_pos](auto&& t) { return t.transform_down(pos); };
    return apply_transform(transform_down, transform_id_);
```
It might also be good to make a templated helper member function `auto apply_transform(F&&)` that locally creates the visitor so you can refactor the backend without changing the entire code. We don't need to keep the transform visitor around since we're only ever doing one transform per function call.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,75 @@
...
+
+    // Calculate the safety distance for any position outside the inner box
+    real_type safety_distance_outside(Real3 pos);
+
+    // Determine if a position is inside the inner box
+    bool is_inside_inner(Real3 const& pos);
+
+    // Determine if a position is inside the outer box
+    bool is_inside_outer(Real3 const& pos);
+
```
sethrj: To avoid having to repeat the transform on the point, I think we should have a single `SignedSense calc_sense(Real3 const& pos)`  that returns whether it's in/out/on. We should be able to use this during initialization and boundary crossings to find out whether we need to test the target volume.

test/orange/detail/OrientedBoundingZone.test.cc
```diff
@@ -0,0 +1,77 @@
...
+
+    ref_storage_ = storage_;
+
+    OrientedBoundingZone obz(inner_id, outer_id, transform_id, &ref_storage_);
+
+    // Test is_inside functions
+    ASSERT_EQ(true, obz.is_inside_inner({10.5, 20.5, 30.5}));
+    ASSERT_EQ(true, obz.is_inside_outer({10.5, 20.5, 30.5}));
+
+    ASSERT_EQ(false, obz.is_inside_inner({11.5, 21.5, 31.5}));
```
sethrj:
```suggestion
    EXPECT_TRUE(obz.is_inside_inner({10.5, 20.5, 30.5}));
    EXPECT_TRUE(obz.is_inside_outer({10.5, 20.5, 30.5}));

    EXPECT_FALSE(obz.is_inside_inner({11.5, 21.5, 31.5}));
```
etc... `ASSERT` is only for "the code will crash after this if the test fails". For example, if you're checking vector elements you need to first assert on the size.

sethrj (changes requested):

src/orange/detail/OrientedBoundingZone.hh
elliottbiondo: Between inner and outer box would come back as `SignedSense::on`. Is that close enough semantically?

sethrj: Yep, exactly. The bounding zone is like having a surface that's (often much) thicker than 1 ULP.

test/orange/detail/OrientedBoundingZone.test.cc
```diff
@@ -0,0 +1,77 @@
...
+    ASSERT_EQ(false, obz.is_inside_outer({12.5, 22.5, 32.5}));
+
+    // Test safety distance functions
+    EXPECT_SOFT_EQ(1.43, obz.safety_distance_inside({10.12, 20.09, 30.57}));
+
+    EXPECT_SOFT_EQ(std::sqrt(1.1 * 1.1),
+                   obz.safety_distance_outside({10, 20, 32.1}));
+    EXPECT_SOFT_EQ(std::sqrt(0.2 * 0.2 + 1.1 * 1.1),
+                   obz.safety_distance_outside({10, 18.8, 32.1}));
+    EXPECT_SOFT_EQ(std::sqrt(0.3 * 0.3 + 0.2 * 0.2 + 1.1 * 1.1),
```
sethrj: The three-term `hypot` was introduced in C++17:
```suggestion
    EXPECT_SOFT_EQ(std::hypot(0.3, 0.2, 1.1),
```

test/orange/detail/OrientedBoundingZone.test.cc
```diff
@@ -0,0 +1,77 @@
...
+    ASSERT_EQ(false, obz.is_inside_inner({11.5, 21.5, 31.5}));
+    ASSERT_EQ(true, obz.is_inside_outer({11.5, 21.5, 31.5}));
+
+    ASSERT_EQ(false, obz.is_inside_inner({12.5, 22.5, 32.5}));
+    ASSERT_EQ(false, obz.is_inside_outer({12.5, 22.5, 32.5}));
+
+    // Test safety distance functions
+    EXPECT_SOFT_EQ(1.43, obz.safety_distance_inside({10.12, 20.09, 30.57}));
+
+    EXPECT_SOFT_EQ(std::sqrt(1.1 * 1.1),
```
sethrj:
```suggestion
    EXPECT_SOFT_EQ(1.1,
```

test/orange/detail/OrientedBoundingZone.test.cc
```diff
@@ -0,0 +1,77 @@
...
+                   obz.safety_distance_outside({10, 20, 32.1}));
+    EXPECT_SOFT_EQ(std::sqrt(0.2 * 0.2 + 1.1 * 1.1),
```
sethrj:
```suggestion
    EXPECT_SOFT_EQ(std::hypot(0.2, 1.1),
```

src/orange/detail/OrientedBoundingZone.cc
elliottbiondo: Discussed in-person. This is already a helpful function, so the TransformVisitor appears only once.

elliottbiondo:

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+    //!@}
+
+  public:
+    // Construct from inner/outer half-widths and a corresponding transform
+    inline CELER_FUNCTION
+    OrientedBoundingZone(OrientedBoundingZoneRecord const* obz_record,
+                         Storage const* storage);
+
+    // Calculate the safety distance for any position inside the outer box
+    inline CELER_FUNCTION real_type safety_distance_inside(Real3 pos);
```
sethrj: Didn't we say we'd name this like the corresponding method in `OrangeTrackView`?  Also use references for Real3
```suggestion
    inline CELER_FUNCTION real_type find_safety_inside(Real3 const& pos);
```

Or `calc_safety_inside` to make it read like the surface methods.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using StorageItems
+        = Collection<T, Ownership::const_reference, MemSpace::native>;
+
+    using ObzReal3 = OrientedBoundingZoneRecord::Real3;
+    using ObzReal3Id = OrientedBoundingZoneRecord::Real3Id;
```
sethrj: Call it `FastReal3` or something? Or call it `Real3` and template the accessors (?)

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+        StorageItems<TransformRecord> const* transforms;
+        StorageItems<real_type> const* reals;
+    };
+    //!@}
+
+  public:
+    // Construct from inner/outer half-widths and a corresponding transform
+    inline CELER_FUNCTION
+    OrientedBoundingZone(OrientedBoundingZoneRecord const* obz_record,
+                         Storage const* storage);
```
sethrj: The guideline we have is to pass pointers only for mutable data; these should be const references.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+CELER_FUNCTION real_type OrientedBoundingZone::safety_distance_outside(Real3 pos)
+{
+    CELER_EXPECT(this->calc_sense(pos) != SignedSense::inside);
+
+    auto trans_pos = this->quadrant_zero(this->translate(pos));
+    auto outer_hw = this->get_hw(obz_record_->outer_hw_id);
+
+    // Case 1
+    if (this->is_inside(trans_pos, outer_hw))
+    {
```
sethrj: Instead of indirecting us to the more detailed (but excellent!) documentation, summarize it. Also put comments about conditionals *inside the context where they apply*.
```suggestion
    if (this->is_inside(trans_pos, outer_hw))
    {
        // Case 1: point is between the inner and outer boxes
```

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+    // Case 2
+    auto inner_hw = this->get_hw(obz_record_->inner_hw_id);
+
+    real_type min_squared = 0;
+    for (auto ax : range(Axis::size_))
+    {
+        auto temp = celeritas::max(
+            real_type{0},
+            trans_pos[to_int(ax)]
+                - static_cast<real_type>(inner_hw[to_int(ax)]));
```
sethrj: Should we be doing the operation (and computing the transformed point) in  "fast" float arithmetic to reduce register usage?

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+        min_squared += temp * temp;
```
sethrj:
```suggestion
        min_squared += ipow<2>(temp);
```

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Reflect a position into quadrant zero.
+ */
+CELER_FUNCTION Real3 OrientedBoundingZone::quadrant_zero(Real3 const& pos)
+{
+    Real3 temp;
+    for (auto ax : range(Axis::size_))
+    {
+        temp[to_int(ax)] = fabs(pos[to_int(ax)]);
```
sethrj:
```suggestion
        temp[to_int(ax)] = std::fabs(pos[to_int(ax)]);
```

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+    return true;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get half-widths for a given index.
+ */
+CELER_FUNCTION OrientedBoundingZone::ObzReal3
+OrientedBoundingZone::get_hw(ObzReal3Id hw_id)
+{
```
sethrj, line 284: This would be a good place to add
```suggestion
{
    CELER_EXPECT(hw_id < storage_.half_widths->size());
```

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,283 @@
...
+    // Calculate the safety distance for any position outside the inner box
+    inline CELER_FUNCTION real_type safety_distance_outside(Real3 pos);
+
+    // Determine the sense of position with respect to the bounding zone
+    inline CELER_FUNCTION SignedSense calc_sense(Real3 const& pos);
+
+  private:
+    // >> DATA
+    OrientedBoundingZoneRecord const* obz_record_;
+    Storage const* storage_;
```
sethrj: Since `Storage` here is actually pointers and not a Collection, I think it would be better to rename as `Input` or `StoragePointers` and you definitely should store it by value.

sethrj (commented):
Sorry for the long delay in my reviewing this!

src/orange/detail/OrientedBoundingZone.hh
elliottbiondo: I specifically avoided calling it `FastReal3` because it might *not* be a fast: it is whatever is specified `OrientedBoundingZoneRecord`. I can name it `FastReal3` if you want though.

elliottbiondo: Changed to `FastReal3` but I am still interested in hearing your perspective on why this is preferred.

elliottbiondo:
@sethrj this should be ready

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,291 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using StorageItems
+        = Collection<T, Ownership::const_reference, MemSpace::native>;
+
+    using FastReal3 = OrientedBoundingZoneRecord::Real3;
+    using FastReal3Id = OrientedBoundingZoneRecord::Real3Id;
+    using fast_real_3 = FastReal3::value_type;
```
sethrj: What is this? Should be `fast_real_type` right??

src/orange/detail/OrientedBoundingZone.hh
sethrj: Well, I think we either need to template OBZ or assume it's an "accelerating" structure like the BIH. In that case we want to emphasize that the types here may not be the same as the real types in the rest of the code.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,291 @@
...
+    // Calculate the safety distance for any position outside the inner box
+    inline CELER_FUNCTION real_type find_safety_outside(Real3 const& pos);
+
+    // Determine the sense of position with respect to the bounding zone
+    inline CELER_FUNCTION SignedSense calc_sense(Real3 const& pos);
+
+  private:
+    // >> DATA
+    OrientedBoundingZoneRecord const& obz_record_;
+    StoragePointers const& sp_;
```
sethrj: I think you missed the comment "you definitely should store it by value." This is basically a pointer to (likely on-the-stack) pointers, whereas it should be more like references that are part of the data, as though it were e.g.,
```c++
    OrientedBoundingZoneRecord const& obz_record_;
    struct {
        StorageItems<FastReal3> const* half_widths;
        StorageItems<TransformRecord> const* transforms;
        StorageItems<real_type> const* reals;
    } storage_;
```
which would let you use `*storage_.reals`, etc.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -0,0 +1,291 @@
...
+ * Case 1: the point is between the inner and outer boxes, resulting in a
+ * safety distance of zero.
+ *
+ * Case 2: the point is inside both the inner and outer boxes, in which case
+ * the safety distance is the minimum distance from the given point to any
+ * point on the outer box. This is calculated by finding the minimum of the
+ * distances to each half width.
+ */
+CELER_FUNCTION real_type
+OrientedBoundingZone::find_safety_inside(Real3 const& pos)
```
sethrj: You know, I thought a bit more about the `find_` vs `calc_` which I mentioned as an alternative spelling: I think in general we want `find_` for algorithms that do a search of some kind (e.g. find_next_distance) and a `calc_` for algorithms that have nontrivial but perhaps still constant time. Is that too annoying/arbitrary do you think?

src/orange/detail/OrientedBoundingZone.hh
elliottbiondo: Yeah it makes sense from an accelerating structure perspective.

src/orange/detail/OrientedBoundingZone.hh
elliottbiondo: Nope, I prefer `calc_` as well

elliottbiondo: And agree with your delineation

src/orange/detail/OrientedBoundingZone.hh
elliottbiondo: Oops

sethrj (approved):
Thanks @elliottbiondo !

PR #1363, opened 2024-08-12
Refactor user manual

sethrj:
This breaks apart several of the long Sphinx RST files into smaller ones (so that instead of being organized according to the code directories, it's broken apart by subject matter). It flattens the structure to make it easier to jump to and read certain parts of the documentation. Finally it improves usability of the within-document sidebar.

Latest manual: https://sethrj.github.io/celeritas/user/index.html


doc/introduction/installation.rst
```diff
@@ -88,6 +88,9 @@ will error out in the configure if their required
 dependencies are missing, but they will update the CMake cache variable so that
 the next configure will succeed (with that component disabled).
 
+
+.. _configuration:
```
esseivaju, line 93: Since you're reworking the doc, I forgot to add Perfetto in the dependency table above when we added it.

src/celeritas/em/interactor/BetheHeitlerInteractor.hh
```diff
@@ -240,7 +240,7 @@ CELER_FUNCTION Interaction BetheHeitlerInteractor::operator()(Engine& rng)
                     auto lpm = calc_lpm_functions_(epsilon);
                     g = lpm.xi
                         * ((2 * lpm.phi + lpm.g) * screening.phi1
-                           - lpm.g * screening.phi2 - lpm.phi * f_z)
+                           - lpm.g * screening.phi2 - (0 + lpm.phi) * f_z)
```
amandalund: What's the reason for this change?

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
```diff
@@ -161,6 +161,9 @@ CELER_FUNCTION Interaction MuBetheBlochInteractor::operator()(Engine& rng)
 //---------------------------------------------------------------------------//
 /*!
  * Calculate maximum kinetic energy of the secondary electron.
+ *
+ * TODO this should be precomputed and used as part of the "applicability" per
+ * particle type, not included in the interactor. See celeritas#907 .
```
amandalund, line 166: How do we do that if it depends on the incident particle energy?

amandalund (approved):
LGTM!

src/celeritas/em/interactor/BetheHeitlerInteractor.hh
sethrj: Oops, I put that in there while I got distracted trying to investigate the difference between the two LPM corrections to the rejection functions. The g1 and g2 LPM corrections look *very* similar after multiplying the g2 factor by 2; except that the first one is missing an `lpm.g` where the zero is now. I'll revert this but it (and frankly lots of the mysterious implementations) bug me.

src/celeritas/em/interactor/MuBetheBlochInteractor.hh
sethrj: The "applicability" for interactions should have a lower bound for the incident particle energy based on conservation rules etc. So, given a  particle type and the electron cutoff for a material, we should be able to solve for the lowest incident muon energy that creates an electron higher than the secondary cutoff. Right? The reason we can't do this now is because we don't have per-material applicability (https://github.com/celeritas-project/celeritas/issues/907).

amandalund: Ah yes, that's right. Actually, looking again I think the max secondary energy might already be used to limit the cross section/energy loss calculation in Geant4, in which case we shouldn't need to do anything except remove the (unnecessary) check in the interactor.

PR #1364, opened 2024-08-13
Improve sincospi availability and implementation

sethrj:
I've had reports from one user building on a macOS machine with Homebrew-installed Clang, and another using a Conda-provided clang, of failures to build `Algorithms.hh`:
```
celeritas/src/corecel/math/Algorithms.hh:515:1: error: no member named '__sincospi' in the global namespace
CELER_WRAP_MATH_FLOAT_DBL_PTR_2(__, sincospi)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
celeritas/src/corecel/math/Algorithms.hh:399:18: note: expanded from macro 'CELER_WRAP_MATH_FLOAT_DBL_PTR_2'
        return ::PREFIX##FUNC(value, a, b);                                  \
               ~~^
<scratch space>:100:1: note: expanded from here
__sincospi
^
```

The `__sincospi` is an apple extension provided in their SDK, which isn't included by other compilers.

This PR includes two enhancements:
- A cmake "try compile" functionality to allow arbitrarily prefixed system-specific `sincospi` functions, and to correctly get that without using `__APPLE__` or implementation-defined macros
- A full and correct implementation of sincospi and sincospif from https://stackoverflow.com/a/42792940/118160 for systems that don't provide the built-ins.


src/corecel/math/Algorithms.hh
```diff
@@ -648,8 +648,9 @@ CELER_WRAP_MATH_FLOAT_DBL_1(, sinpi)
 CELER_WRAP_MATH_FLOAT_DBL_1(, cospi)
 CELER_WRAP_MATH_FLOAT_DBL_PTR_2(, sincospi)
 CELER_WRAP_MATH_FLOAT_DBL_PTR_2(, sincos)
-#elif __APPLE__ && !defined(__CLING__)
-// Apple defines __sinpi, __sinpif, __sincospi, ...
+#elif (__MAC_OS_X_VERSION_MIN_REQUIRED > 1090) \
+    || (__IPHONE_OS_VERSION_MIN_REQUIRED > 70000)
```
pcanal: iphone, really?  humm ... I guess soon enough its gpu will be strong enough to run Celeritas :) :)

src/corecel/math/Algorithms.hh
```diff
@@ -648,8 +648,9 @@ CELER_WRAP_MATH_FLOAT_DBL_1(, sinpi)
 CELER_WRAP_MATH_FLOAT_DBL_1(, cospi)
 CELER_WRAP_MATH_FLOAT_DBL_PTR_2(, sincospi)
 CELER_WRAP_MATH_FLOAT_DBL_PTR_2(, sincos)
-#elif __APPLE__ && !defined(__CLING__)
-// Apple defines __sinpi, __sinpif, __sincospi, ...
+#elif (__MAC_OS_X_VERSION_MIN_REQUIRED > 1090) \
```
pcanal: it looks odd to use a `__` macro, isn't there a "public" version we can use here?

src/corecel/math/Algorithms.hh
sethrj: I copied the condition from Apple's `math.h`... might as well be "exactly right", no? 😉

src/corecel/math/Algorithms.hh
sethrj: Alas no, I tried. They're defined in [Availability.h](https://opensource.apple.com/source/xnu/xnu-7195.81.3/EXTERNAL_HEADERS/Availability.h.auto.html) and it says they're "for use in OS header files" but at first reading I didn't notice the "OS" — so maybe we're not allowed to use them ☹️ the fallback still works though...

The only other thing I could do is add a cmake compiler test instead but that seems like overkill.

src/corecel/math/Algorithms.hh
pcanal: Yes but (especially because it is apple) it might change without warning :(

pcanal: eg.  The version on my system is:
```
#if (defined __MAC_OS_X_VERSION_MIN_REQUIRED && __MAC_OS_X_VERSION_MIN_REQUIRED < 1090) || \
    (defined __IPHONE_OS_VERSION_MIN_REQUIRED && __IPHONE_OS_VERSION_MIN_REQUIRED < 70000)
```

src/corecel/math/Algorithms.hh
pcanal: > The only other thing I could do is add a cmake compiler test instead but that seems like overkill.

it is however the 'right' thing to do ... and it even catches other compiler/os that supports it.

src/corecel/math/Algorithms.hh
sethrj: Well right, but the "else" conditional is the one that has sincospi. So if the `min required` is undefined then we have the condition  0 > 1090, etc., which is a nice fallback for the macros going undefined.

src/corecel/math/Algorithms.hh
sethrj: Thinking about it, `__APPLE__`  and `__CLING__` are both underscores too...

CMakeLists.txt
```diff
@@ -490,7 +490,7 @@ elseif(CELERITAS_USE_HIP)
   find_library(ROCTX_LIBRARY roctx64)
   if(ROCTX_LIBRARY)
     set(CELERITAS_HAVE_ROCTX ON)
-    target_link_libraries(celeritas_device_toolkit INTERFACE ${ROCTX_LIBRARY})
+    target_link_libraries(celeritas_device_toolkit INTERFACE "${ROCTX_LIBRARY}")
```
pcanal, line 493: Why are the quote needed here?

cmake/try-sincospi.cc
```diff
@@ -0,0 +1,33 @@
...
+//! \file try-sincospi.cc
+//---------------------------------------------------------------------------//
+
+#include <cmath>
+
+#ifndef CELERITAS_SINCOSPI_PREFIX
+#    error "Expected CELERITAS_SINCOSPI_PREFIX to be defined"
+#endif
+
+#define CELER_CONCAT(PREFIX, FUNC) PREFIX##FUNC
```
pcanal: If it is important that those macro stays in sync with future changes in the original, we may want to note where they come from.

CMakeLists.txt
```diff
@@ -510,6 +510,18 @@ install(TARGETS celeritas_device_toolkit
   EXPORT celeritas-targets
 )
 
+#----------------------------------------------------------------------------#
+if(APPLE)
```
pcanal: Why limit this test to apple?  We think only Apple provides it now, but other platforms may be providing now or later without us knowing.

CMakeLists.txt
```diff
@@ -510,6 +510,18 @@ install(TARGETS celeritas_device_toolkit
...
+  set(_prefix "__")
+  try_compile(CELERITAS_TRY_${_prefix}SINCOSPI #${CMAKE_BINARY_DIR}
+    SOURCES "${PROJECT_SOURCE_DIR}/cmake/try-sincospi.cc"
+    COMPILE_DEFINITIONS "-DCELERITAS_SINCOSPI_PREFIX=${_prefix}"
+  )
+  if(CELERITAS_TRY_${_prefix}SINCOSPI)
+    set(CELERITAS_SINCOSPI_PREFIX "${_prefix}")
+  endif()
+endif()
```
```suggestion
foreach(__prefix "" "__" )
  try_compile(CELERITAS_TRY_${_prefix}SINCOSPI #${CMAKE_BINARY_DIR}
    SOURCES "${PROJECT_SOURCE_DIR}/cmake/try-sincospi.cc"
    COMPILE_DEFINITIONS "-DCELERITAS_SINCOSPI_PREFIX=${_prefix}"
  )
  if(CELERITAS_TRY_${_prefix}SINCOSPI)
    set(CELERITAS_SINCOSPI_PREFIX "${_prefix}")
    break()
  endif()
endforeach()
```

src/corecel/math/Turn.hh
```diff
@@ -35,7 +35,13 @@ struct HalfPi
 };
 
 //---------------------------------------------------------------------------//
-//! Quantity denoting a full turn
+/*!
+ * Quantity denoting a full turn.
+ *
+ * Turns are a useful way of representing angles without the historical
```
pcanal, line 41: This is a good improvement.  Is it related to the issue of this PR?

CMakeLists.txt
sethrj: In the unlikely case that there's a space in the path: it's a good practice.

src/corecel/math/Turn.hh
sethrj: Yeah, this is all about sincospi which lets you use rational values for turns.

CMakeLists.txt
sethrj: OK, I'll expand (again) the scope of the PR...

src/corecel/math/detail/Sincospi.hh
```diff
@@ -0,0 +1,206 @@
...
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file corecel/math/detail/Sincospi.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <cmath>
+#include <cstdint>
+
```
pcanal, line 12:
```suggestion
#include "corecel/Config.hh"

#ifdef CELERITAS_SINCOSPI_PREFIX
#error "sincospi is provided natively on this platform. Do not include Sincospi.hh"
#endif
```

sethrj: What if we want to test the implemented functions? You should only `#error` to prevent further build errors...

CMakeLists.txt
pcanal: Fair enough.  There are other place where we forgot to do this :)

src/corecel/math/detail/Sincospi.hh
pcanal: The problem is that if a user 'inadvertently' use this, there will be duplicate symbols or worse simple silent break of ODR (eg. if all the sincospi get inlined).  If one wants to test those function on a platform that already has `sincospi` additional care/gymnastic is anyway needed and it might make sense for that test to have to explicitly `#undef` the macro.

sethrj: No there won't; the `detail` functions in this header are uniquely defined. The only potential ODR violation will be if `::celeritas::sincospi` (for example) calls one implementation in one translation unit and another implementation in another. No macro will fix that, and since they're forceinlined it probably won't double-define anyway.

pcanal: > since they're forceinlined it probably won't double-define anyway.

That the worst case scenario (it silent weird result because the 2 implementation give slightly different result).

> No macro will fix that

Sort of.  I indeed completely missed that this header is defining `sincospi_impl` rather than`sincospi`.  So indeed it would take a corrupted build using 2 distinct config header file to trigger the problem.

So we can leave it as is.

pcanal (approved):

PR #1366, opened 2024-08-15
Accept single CsgTree from stdin 

esseivaju:
The use case for reading from stdin is to print a single tree dumped by `to_json_string`. The list check only makes sense when reading from a `.csg.json` file where we can have multiple csg universes.

### DIFF
--- a/scripts/user/orange-csg-to-dot.py
+++ b/scripts/user/orange-csg-to-dot.py
@@ -133,7 +133,7 @@ def run(infile, outfile, gencls, universe):
         # Load from a .csg.json debug file
         csg_unit = tree[universe]
     else:
-        if isinstance(tree, list):
+        if isinstance(tree, list) and infile is not sys.stdin:
             num_univ = len(tree)
             print("Input tree is a CSG listing: please rerun with -u N "
                   f"where 0 ≤ N < {num_univ}",
--- a/scripts/user/orange-csg-to-dot.py
+++ b/scripts/user/orange-csg-to-dot.py
@@ -11,8 +11,9 @@ import json
 import sys
 
 class DotGenerator:
-    def __init__(self, f):
+    def __init__(self, f, args):
         self.f = f
+        self.print_ids = args.print_ids
         self.write = f.write
         self.vol_edges = []
 
@@ -34,7 +35,11 @@ edge [color=gray, dir=both]
         self.write("}\n}\n")
 
     def write_node(self, i, value):
-        self.write(f"{i:02d} [label=\"{value}\"]\n")
+        if self.print_ids:
+            id_format = f"{i:02d}:"
+        else:
+            id_format = ""
+        self.write(f"{i:02d} [label=\"{id_format}{value}\"]\n")
 
     @contextmanager
     def write_volumes(self):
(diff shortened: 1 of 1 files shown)


scripts/user/orange-csg-to-dot.py
```diff
@@ -133,7 +133,7 @@ def run(infile, outfile, gencls, universe):
         # Load from a .csg.json debug file
         csg_unit = tree[universe]
     else:
-        if isinstance(tree, list):
+        if isinstance(tree, list) and infile is not sys.stdin:
```
sethrj: I think this condition I set was wrong: sorry. Maybe we need to check the contents of the list to see whether it's a ".csg.json" file? Perhaps we should just have the `.csg.json` be a dict that has a format and version? Or for now can we just see if the list's first item is a dict that has "metadata" as a key?

scripts/user/orange-csg-to-dot.py
```diff
@@ -72,7 +72,7 @@ def __exit__(self, type, value, traceback):
             self.write(f"  v{i:02d} <--> n{i:02d}\n")
 
     def write_node(self, i, value):
-        self.write(f"  n{i:02d}[\"{value}\"]\n")
+        self.write(f"  n{i:02d}[\"{i:02d}:{value}\"]\n")
```
sethrj: Can you post what the CSG output looks like with this? The node number probably only matters for debugging CSG utilities, not for generally visualizing the contents of a CSG unit. Maybe this should be an option...

esseivaju: It looks like this. It makes sense to have it as an option.
```mermaid
flowchart TB
  n00["t"]
  n01["~"]
  n01 --> n00
  n02["2:S0"]
  n03["3:S1"]
  n04["4:~"]
  n04 --> n03
  n05["5:S2"]
  n06["6:~"]
  n06 --> n05
  n07["7:&"]
  n07 --> n02
  n07 --> n04
  n07 --> n06
  n08["8:S3"]
  n09["9:~"]
  n09 --> n08
  n10["10:&"]
  n10 --> n02
  n10 --> n04
  n10 --> n09
  n11["11:~"]
  n11 --> n10
  n12["12:S4"]
  n13["13:S5"]
  n14["14:&"]
  n14 --> n12
  n14 --> n13
  n15["15:&"]
  n15 --> n07
  n15 --> n11
  n15 --> n14
  n16["16:~"]
  n16 --> n15
  n17["17:S6"]
  n18["18:S7"]
  n19["19:~"]
  n19 --> n18
  n20["20:&"]
  n20 --> n06
  n20 --> n17
  n20 --> n19
  n21["21:&"]
  n21 --> n09
  n21 --> n17
  n21 --> n19
  n22["22:~"]
  n22 --> n21
  n23["23:S8"]
  n24["24:S9"]
  n25["25:&"]
  n25 --> n23
  n25 --> n24
  n26["26:&"]
  n26 --> n20
  n26 --> n22
  n26 --> n25
  n27["27:~"]
  n27 --> n26
  n28["28:&"]
  n28 --> n15
  n28 --> n27
```

scripts/user/orange-csg-to-dot.py
esseivaju: yeah, I think for now just checking if the first element in a csg unit dict should be enough

scripts/user/orange-csg-to-dot.py
sethrj: Option definitely: for CSG trees it's convenient but for "labeled" geometry I think it would be just noise.

sethrj (approved):
Good enough for government work!

PR #1367, opened 2024-08-16
Properly clear state counters and data for reuse

amandalund:
This fixes an unsual case for celer-sim when multiple events are run one at a time on a single stream and `max_steps` is specified (unlikely to be done in practice, but how we run our CPU app test). Because the core state counters, track status, and vacancies are not reset between events, stale state data is reused in the next event. This adds a helper method to reset the state data, as well as:
 - Add a generic function to fill a collection with a sequence
 - Move track shuffle utils to global/detail
 - Remove unecessary initialization from the sim state


sethrj (approved):
Nice catch, thanks for the refactoring!

PR #1368, opened 2024-08-16
Add inverse helpers to generic calculator

sethrj:
This is in preparation for making "views" to the optical physics data: we have at least one place where we need to "invert" a generic grid calculator.
- Complete a 'todo' for renaming Data→Record
- Refactor the "uniform grid" to directly access the underlying collection rather than storing a temporary span
- Add  "from_inverse" and "make_inverse" functions to GenericCalculator


src/corecel/grid/NonuniformGrid.hh
```diff
@@ -73,12 +82,12 @@ class NonuniformGrid
  */
 template<class T>
 CELER_FUNCTION
-NonuniformGrid<T>::NonuniformGrid(ItemRange<value_type> const& values,
-                                  Values const& data)
-    : data_(data[values])
+NonuniformGrid<T>::NonuniformGrid(ItemRangeT const& values, Values const& data)
+    : values_{data}, offset_{values}
```
amandalund: The names are a tiny bit confusing with all the `values` ;)

src/celeritas/grid/GenericCalculator.hh
```diff
@@ -21,51 +21,81 @@ namespace celeritas
...
 // INLINE DEFINITIONS
 //---------------------------------------------------------------------------//
 /*!
- * Construct from grid data and backend values.
+ * Construct by \em inverting a monotonicially increasing generic grid.
+ */
+CELER_FUNCTION GenericCalculator GenericCalculator::from_inverse(
+    GenericGridRecord const& grid, Reals const& storage)
+{
+    return GenericCalculator{storage, grid.value, grid.grid};
```
amandalund, line 86: Do we need to check somewhere that the values actually are monotonically increasing? Or is that the responsibility of whatever's calling this?

amandalund (approved):
Looks great, thanks @sethrj!

src/corecel/grid/NonuniformGrid.hh
sethrj: Oh goodness yes... I thought I cleaned that up (maybe only in generic calculator). Sorry!

src/celeritas/grid/GenericCalculator.hh
sethrj: It's too expensive to check that in the constructor, but at least the `NonUniformGrid` will check that the first value is greater than the last... I'll add a warning/precondition to the documentation.

PR #1369, opened 2024-08-16
Fix initialization of geometry state when inserting multiple sets of primaries

amandalund:
This fixes a bug where a new track's geometry was detail-initialized using the geometry state of the incorrect parent track. This occurred when new primaries were inserted in the middle of a simulation and initialized using the parent IDs of secondaries produced in the previous step.

### DIFF
--- a/src/celeritas/track/ExtendFromPrimariesAction.cc
+++ b/src/celeritas/track/ExtendFromPrimariesAction.cc
@@ -9,6 +9,7 @@
 
 #include "corecel/Assert.hh"
 #include "corecel/Macros.hh"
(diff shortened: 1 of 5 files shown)


sethrj (approved):
Wow, good catch. I've always been a bit afraid of the way we're doing initialization 😓 perhaps now that we have the `initializing` track status we can refactor the code so that initialization happens in a way that's easier to follow...

amandalund:
Yeah, that was something we added very early on and haven't touched much, and clearly we've since added support for running in different ways that break it unexpectedly...

PR #1371, opened 2024-08-17
Allow normal distribution to preserve spare value

sethrj:
This allows a normal distribution's "spare" value (from the Box-Muller method) to be reused with different normal distributions. This will be used in scintillation where the wavelength is sampled from a component-dependent gaussian.


amandalund (approved):
Nice, thanks @sethrj!

PR #1372, opened 2024-08-17
Clang format develop with clang 18

sethrj:
Hopefully this reduces the automatic reformatting that happens when merging in develop 😅

### DIFF
--- a/src/celeritas/io/ImportUnits.cc
+++ b/src/celeritas/io/ImportUnits.cc
@@ -71,19 +71,21 @@ double mass_scale(UnitSystem sys)
  */
 char const* to_cstring(ImportUnits value)
 {
-    static EnumStringMapper<ImportUnits> const to_cstring_impl{"unitless",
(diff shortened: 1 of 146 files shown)


src/celeritas/io/ImportUnits.cc
```diff
@@ -82,8 +82,9 @@ char const* to_cstring(ImportUnits value)
         "len^2",
         "time",
         "1/len^3",
-        "len-time^2/mass",
-        "1/MeV"
+        "len-time^2/"
+        "mass",
```
amandalund: Hmm what happened here...

amandalund (approved):
And a code reduction as well :)

src/celeritas/io/ImportUnits.cc
sethrj: It first formatted it all the way to the right, then I added a comma to unindent it, and I didn't notice it inserted this horrible line break...

PR #1373, opened 2024-08-19
Sample cerenkov photons one at a time using rejection

sethrj:
This is the first half of a refactoring of the generators to sample photons one at a time: the next will be a refactor of the scintillation generator.

- Refactor generator so one instance can generate an arbitrary number of photons, rather than only being callable once and returning an array. (This could let us e.g. stop sampling early if we don't have enough primary space.)
- Refactor the cerenkov rejection loop into a "reject on unphysical energy" (i.e. raw spectrum) and "reject as part of a modified algorithm". This saves a few RNG samples in the simplest case.
- Add helper functions for converting between wavelength and energy, to be applied to scintillation in next PR
- Add `exchange_count` helper to RNG which returns and clears count at the same time


src/celeritas/optical/CerenkovGenerator.hh
```diff
@@ -54,12 +54,11 @@ class CerenkovGenerator
     inline CELER_FUNCTION
     CerenkovGenerator(MaterialView const& material,
                       NativeCRef<CerenkovData> const& shared,
-                      GeneratorDistributionData const& dist,
-                      Span<Primary> photons);
+                      GeneratorDistributionData const& dist);
 
     // Sample Cerenkov photons from the distribution
```
stognini:
```suggestion
    // Sample a single Cerenkov photon
```

amandalund (approved):
Thanks @sethrj!

test/celeritas/optical/Cerenkov.test.cc
```diff
@@ -113,10 +114,9 @@ Span<double const> get_refractive_index()
     return make_span(refractive_index);
 }
 
-double convert_to_energy(double wavelength)
+real_type convert_to_energy(real_type wavelength)
```
stognini: Isn't it better to just use `OpticalUtils::wavelength_to_energy` directly and drop this one?

stognini (approved):
Looks good to me! Thanks!

test/celeritas/optical/Cerenkov.test.cc
sethrj: yeah I'll just do that.

sethrj: I think it's nicer to have a local utility function for the test to make stuff more readable actually.

PR #1376, opened 2024-08-19
Sample scintillation photons individually

sethrj:
This updates the scintillation photon generator to create a single photon at a time by sampling from material yields instead of splitting the selected photons among them. It also fixes an error in the sampling (which assumed that component index for a material was the same as the component ID).

It keeps us from  losing "stratified sampling" of the original implementation but it no longer unfairly rounds.

I refactored some of the data structures and builders to make this happen, and to reduce the amount of work I've removed the unused scintillate-by-particle construction code (which @stognini added and I believe @whokion mildly objected to because it may not be needed). The data structures and related import code are still there; we can either remove those at some future point or re-add the by-particle construction and implement sampling by particle.


src/celeritas/optical/ScintillationGenerator.hh
```diff
@@ -57,12 +60,11 @@ class ScintillationGenerator
     // Construct from scintillation data and distribution parameters
     inline CELER_FUNCTION
     ScintillationGenerator(GeneratorDistributionData const& dist,
-                           NativeCRef<ScintillationData> const& shared,
-                           Span<Primary> photons);
+                           NativeCRef<ScintillationData> const& shared);
 
     // Sample Scintillation photons from the distribution
```
stognini: Just so it looks equivalent to the previous refactoring

```suggestion
    // Sample a scintillation photon from the distribution
```

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -23,7 +23,7 @@ namespace celeritas
  */
 struct ImportScintComponent
 {
-    double yield_per_energy{};  //!< Yield for this material component [1/MeV]
+    double yield_frac{};  //!< Fraction of total scintillation yield
```
stognini, line 26: I am pondering about the renaming here, since this has the same meaning as in `MatScintSpectrumRecord::yield_per_energy`, right? I.e. both are in [1/MeV], and are not a fraction of the total yield (e.g. the total is 200 photons per MeV for the material, and this component is 30% of that).

sethrj: We don't use these to calculate the total material yield in the import vector... if we got rid of that it would make sense to use non normalized values for these.

sethrj: I originally was going to check that the yields summed to the total material yield, but the couple of examples I saw, it was just fractional.

stognini: Ohh right, the components are always fractional, and the material-only yield is the actual photons / MeV (which is `ImportMaterialScintSpectrum::yield_per_energy`). Okay, that makes sense.

stognini (approved):
Looks good to me!

src/celeritas/optical/ScintillationData.hh
```diff
@@ -20,71 +20,73 @@ namespace optical
...
+    real_type lambda_sigma{};  //!< Standard deviation of wavelength
     real_type rise_time{};  //!< Rise time
     real_type fall_time{};  //!< Decay time
 
     //! Whether all data are assigned and valid
     explicit CELER_FUNCTION operator bool() const
     {
-        return yield_frac > 0 && yield_frac <= 1 && lambda_mean > 0
-               && lambda_sigma > 0 && rise_time >= 0 && fall_time > 0;
+        return lambda_mean > 0 && lambda_sigma >= 0 && rise_time >= 0
```
amandalund: Did you mean to change this to `lambda_sigma >= 0` (the normal distribution requires a positive stddev, and in the params construction we still validate that it's positive)?

amandalund (approved):
This looks great @sethrj, thanks!

src/celeritas/optical/ScintillationData.hh
sethrj: Oh I figured that for testing a delta function could be useful, but I didn't think to check that it would work with the normal distribution. Perhaps we should allow that as a degenerate case too?

amandalund: Yeah, I know we'd considered that on a couple past PRs but held off: https://github.com/celeritas-project/celeritas/pull/1108#discussion_r1491233708, https://github.com/celeritas-project/celeritas/pull/1086#discussion_r1462255701

sethrj: Wow, history repeats itself... thanks for the reminder.

sethrj: I've reverted this, as usual we can add "=0" when we encounter a use case 🤪

PR #1378, opened 2024-08-27
Add volume-to-material mapping and change import data

sethrj:
This adds to the `OpticalMaterial` a mapping of volume ID to optical material ID; it is allowable to have no ID assigned (so that another part of the code can kill the track etc.) I've also changed the import material data so that the optical material ID is associated with the physical material ID, since it's conceivable we'd want to turn on/off optical physics in different regions of the problem. 

A few changes are:
- Instead of looking for scintillation etc. in the first optical material, we look at all materials, since the user might want to turn off processes on a per-material basis
- Any material with an RINDEX property is considered "optical"; the other physics properties are optional.
- I've added an "optical" data selection to the geant importer that defaults to on; we should think about detecting the presence of optical processes to specify whether cerenkov (e.g.) is enabled 

Currently


src/celeritas/ext/GeantImporter.cc
```diff
@@ -147,6 +153,8 @@ struct ProcessFilter
         {
             case G4ProcessType::fElectromagnetic:
                 return (which & DataSelection::em);
+            case G4ProcessType::fOptical:
```
stognini, line 156: I still have to go through most of the PR, but fun fact: scintillation is tagged as `fElectromagnetic` in Geant4. So we'll have to be careful with our physics flags if we plan on saying that scintillation (though physically correct to be an EM process) belongs to optical, as it emits optical photons.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -250,6 +260,8 @@ fill_vec_import_scint_comp(MatPropGetter& get_property,
                             ImportUnits::time);
         if (comp)
         {
+            // TODO: should we error if the user defines some but not all
```
stognini: Looking at our generator it seems like we need all, except for the rise time, that can be set to zero. Everything else must be non-zero to avoid either odd behaviors or straight up errors: if the fall time is not set we end up with a division by zero in `ScintillationGenerator::148`.

src/celeritas/ext/GeantImporter.cc
sethrj: Yeesh. Thanks for the heads up. Guess that's because it's affecting EM tracks and not optical photons??

stognini: Probably? That would be my first guess too... Also, for scintillation you need ionization on (you need dE/dx), so that may be another reason?

src/celeritas/ext/GeantImporter.cc
sethrj: It makes sense that we need it all, but since these properties aren't really validated by Geant4 perhaps we should throw an error saying "hey we expected all but found these". Or perhaps that should be done inside the scintillation generator itself from the import data.

amandalund: Yeah, we do validate the imported components in the scintillation params...

amandalund (approved):
Looks great, thanks @sethrj!

src/celeritas/optical/ScintillationParams.cc
```diff
@@ -30,66 +30,68 @@ std::shared_ptr<ScintillationParams>
...
 
     Input input;
     input.resolution_scale.resize(num_optmats);
     if (data.optical_params.scintillation_by_particle)
     {
         // Collect ScintillationParticleIds
         input.pid_to_scintpid.resize(data.particles.size());
-        ScintillationParticleId scintpid{0};
-        for (auto const& [matid, iom] : data.optical)
+        ScintillationParticleId::size_type scintpid{0};
```
stognini, line 53: Just so I am in the same page, let me know if this statement is correct: the plan is to keep the scintillation by particle import code, but we're not implementing that yet in our generator. So we keep per-particle data on the import side (and `ScintillationData`, though will be empty), but do not enable it in our `ScintillationParams`, leaving the stepping loop to be material-only. And if/when we do add scintillation by particle type, all the backend and data storage is already in place, since the current extra bits shouldn't make any real difference in the stepping loop.

sethrj: Yeah, that's about the sum of it. Since I don't think we've yet found any sincillation-by-particle detector we could probably delete it and restore it later if needed... 🤔

PR #1379, opened 2024-08-27
Count number of optical primaries to be generated from distributions

amandalund:
This small update adds a reduction to the optical offload actions to count the number of primaries that would be generated from the distribution data. This will be used to determine when to generate the primaries and start the optical tracking loop.

### DIFF
--- a/src/celeritas/optical/detail/OpticalGenAlgorithms.cc
+++ b/src/celeritas/optical/detail/OpticalGenAlgorithms.cc
@@ -16,11 +16,14 @@ namespace celeritas
 {
 namespace detail
 {
+namespace
+{
 //---------------------------------------------------------------------------//
+
 struct AccumNumPhotons
 {
     // Accumulate the number of optical photons from the distribution data
-    CELER_FUNCTION size_type
+    size_type
     operator()(size_type count,
                celeritas::optical::GeneratorDistributionData const& data) const
     {
@@ -29,15 +32,16 @@ struct AccumNumPhotons
 };
 
 //---------------------------------------------------------------------------//
+}  // namespace
+
+//---------------------------------------------------------------------------//
 /*!
  * Remove all invalid distributions from the buffer.
  *
- * This returns the total number of valid distributions in the buffer.
+ * \return Total number of valid distributions in the buffer
  */
 size_type
-remove_if_invalid(Collection<celeritas::optical::GeneratorDistributionData,
-                             Ownership::reference,
-                             MemSpace::host> const& buffer,
+remove_if_invalid(GeneratorDistributionRef<MemSpace::host> const& buffer,
                   size_type offset,
                   size_type size,
                   StreamId)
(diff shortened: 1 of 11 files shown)


src/celeritas/optical/detail/OpticalGenAlgorithms.hh
```diff
@@ -44,6 +44,23 @@ remove_if_invalid(Collection<celeritas::optical::GeneratorDistributionData,
                   size_type,
                   StreamId);
 
+//---------------------------------------------------------------------------//
+// Count the number of optical photons in the distributions.
+size_type
+count_num_photons(Collection<celeritas::optical::GeneratorDistributionData,
+                             Ownership::reference,
+                             MemSpace::host> const&,
```
sethrj: Perhaps it would be helpful to define
```c++
template<MemSpace M>
using GeneratorDistributionRef = Collection<::celeritas::optical::GeneratorDistributionData, Ownership::reference, M>;
```
in this header file

src/celeritas/optical/detail/OpticalGenAlgorithms.cc
```diff
@@ -8,16 +8,31 @@
...
 
 namespace celeritas
 {
 namespace detail
 {
+//---------------------------------------------------------------------------//
+struct AccumNumPhotons
+{
+    // Accumulate the number of optical photons from the distribution data
+    CELER_FUNCTION size_type
```
sethrj:
```suggestion
    size_type
```
since it's a `.cc`, and you can put it in an anonymous namespace too

src/celeritas/optical/detail/OpticalGenAlgorithms.cc
```diff
@@ -8,16 +8,31 @@
...
+    {
+        return count + data.num_photons;
+    }
+};
+
 //---------------------------------------------------------------------------//
 /*!
  * Remove all invalid distributions from the buffer.
+ *
+ * This returns the total number of valid distributions in the buffer.
```
sethrj: could use the doxygen markup here 😉 
```suggestion
 *
 * \return the total number of valid distributions in the buffer
```

sethrj (approved):
Great next step!

src/celeritas/optical/detail/OpticalGenAlgorithms.hh
amandalund: Yeah that looks a bit neater.

test/celeritas/global/StepperGeant.test.cc
```diff
@@ -800,7 +800,6 @@ TEST_F(LeadBox, host)
     // tracking cut
     EXPECT_EQ(1, result.calc_avg_steps_per_primary());
     EXPECT_EQ(1, result.num_step_iters());
-    EXPECT_SOFT_EQ(1, result.calc_avg_steps_per_primary());
```
sethrj, line 803: Was this intentional? Not sure how this would be affected...

sethrj (approved):

test/celeritas/global/StepperGeant.test.cc
amandalund: Yeah, it doesn't have anything to do with this PR, it was just a duplicate check I'd left in by mistake. I can revert that and clean it up later if you'd prefer.

sethrj: Ah I see it now. Works for me!

PR #1380, opened 2024-08-28
Replace SWIG with JSON (via ROOT export) for data exploration

sethrj:
This adds an option in `celer-export-geant` to export our "import data" to JSON using ROOT's helpful reflection/introspection tools. Since it's easier to read and parse this format compared to ROOT or SWIG, I think continuing to move toward JSON for I/O is a smart move.


src/celeritas/ext/RootJsonDumper.cc
```diff
@@ -0,0 +1,46 @@
...
+#include "corecel/io/Logger.hh"
+#include "celeritas/io/ImportData.hh"
+
+#include "RootFileManager.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from path to ROOT file.
```
stognini:
```suggestion
 Construct with output stream object.
```

Likely a remnant copy-paste from RootImporter. Same for the `operator()` below.

src/celeritas/ext/RootJsonDumper.hh
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * \code
+ *  RootJsonDumper dump(&std::cout);
+ *  dump(my_import_data);
+ * \endcode
+ */
+class RootJsonDumper
+{
+  public:
+    // Construct with ROOT file name
```
stognini: Same here. :)

test/celeritas/CMakeLists.txt
```diff
@@ -163,6 +163,9 @@ celeritas_add_test(ext/GeantImporter.test.cc
 celeritas_add_test(ext/GeantVolumeMapper.test.cc ${_needs_geant4}
   LINK_LIBRARIES ${Geant4_LIBRARIES} Celeritas::orange)
 celeritas_add_test(ext/RootImporter.test.cc ${_needs_root})
+celeritas_add_test(ext/RootJsonDumper.test.cc ${_needs_root}
+  LINK_LIBRARIES nlohmann_json::nlohmann_json
```
stognini: This may be a silly question, but why do we need to link nlohmann here? ROOT classes are already linked against it, no?

app/celer-export-geant.cc
```diff
@@ -76,6 +78,55 @@ GeantPhysicsOptions load_options(std::string const& option_filename)
...
+        // Write ROOT file
+        RootExporter export_root(out_filename.c_str());
+        export_root(imported);
+    }
+    else if (ends_with(out_filename, ".json"))
+    {
+        // Write JSON to file
+        CELER_LOG(info) << "Opening JSON output at " << out_filename;
+        std::ofstream os(out_filename);
+        RootJsonDumper dump_root(&os);
```
stognini:
```suggestion
        RootJsonDumper export_json(&os);
```
so it reads more naturally compared to the previous `if` statement block.

stognini (approved):
Looks great! Thanks!

test/celeritas/CMakeLists.txt
sethrj: Oh I intended to modify the JSON export directly but ended up changing my mind. You're right that it's not needed now.

src/celeritas/ext/RootJsonDumper.cc
sethrj: Sorry about not reviewing this more thoroughly before sending it on to you! Dumb copy paste errors.

stognini: Don't worry. Happy to try and catch these.

PR #1382, opened 2024-08-29
Allow sim params without looping

sethrj:
Small pull request to allow construction of SimParams without looping counters for problems without a magnetic field or without charged particles (i.e., optical physics).

### DIFF
--- a/src/celeritas/track/SimData.hh
+++ b/src/celeritas/track/SimData.hh
@@ -110,7 +110,10 @@ struct SimTrackInitializer
  * Data storage/access for simulation states.
  *
  * Unless otherwise specified, units are in the native system (time = s for
- * CGS).
+ * CGS, step length = cm).
+ *
+ * \c num_looping_steps will be empty if params doesn't specify any looping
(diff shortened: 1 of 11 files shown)


src/celeritas/track/SimData.hh
```diff
@@ -75,7 +75,7 @@ struct SimParamsData
     //// METHODS ////
 
     //! Whether the data are assigned
-    explicit CELER_FUNCTION operator bool() const { return !looping.empty(); }
+    explicit CELER_FUNCTION operator bool() const { return true; }
```
amandalund, line 78: Looks like we also need to relax the requirement that `num_looping_steps` is allocated in the state `operator bool()`.

amandalund (approved):
Looks good @sethrj!

src/celeritas/track/SimData.hh
sethrj: Yeah, this is a rebase error 😅 will fix in a minute, thanks!

PR #1383, opened 2024-08-29
Add always-on celer-sim diagnostic to count total number of tracks

amandalund:
Could be useful sometimes to get the total number of tracks simulated without e.g. enabling the step diagnostic.

### DIFF
--- a/app/celer-sim/Transporter.cc
+++ b/app/celer-sim/Transporter.cc
@@ -10,10 +10,12 @@
 #include <algorithm>
 #include <csignal>
 #include <memory>
+#include <numeric>
 #include <utility>
 
 #include "corecel/Assert.hh"
 #include "corecel/cont/Range.hh"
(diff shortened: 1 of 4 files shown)


app/celer-sim/Transporter.cc
```diff
@@ -158,6 +160,9 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries) -> TransporterResult
         record_step_time();
     }
 
+    auto counters = copy_to_host(stepper_->state_ref().init.track_counters);
+    result.num_tracks = std::accumulate(
+        counters.data().get(), counters.data().get() + counters.size(), 0);
```
sethrj:
```suggestion
        counters.data().get(), counters.data().get() + counters.size(), size_type(0));
```
otherwise it's accumulating a signed int and then implicitly casting the result to a size_type.

sethrj (approved):

PR #1386, opened 2024-08-31
Add skeleton optical core params and launch action

sethrj:
This adds the basic structure for the optical core params and state to be used in the tracking loop, and it adds the  "optical launch action" that can access both the offload data and the launch state.


src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -0,0 +1,152 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Launch the optical tracking loop.
+ */
+template<MemSpace M>
+void OpticalLaunchAction::execute_impl(CoreParams const& core_params,
+                                       CoreState<M>& core_state) const
+{
+    auto& offload_state = get<OpticalOffloadState<M>>(
+        core_state.aux(), offload_params_->aux_id());
```
amandalund, line 140: I might still need some help understanding how we'll access the offload state from the generator actions (currently two explicit optical actions for creating scintillation/Cerenkov primaries from the distributions). Looks like the plan is to construct them in `optical::CoreParams::build_actions()`, and they'll execute once at the start of the optical tracking loop (similar to "extend from primaries"). Since they'll execute with the optical core params/state, how should we access the regular core state aux data within them? Should they instead be part of the main stepping loop, constructed in the optical collector with both the offload and optical aux IDs?

sethrj: I see what you're getting at; my thought was that the optical loop would only contain primaries and the "optical launcher" would construct optical primaries from the main loop, but that's not ideal (though perhaps good enough for now).

I think what would be best long-term is if we have an action-like host interface that can be passed to the optical loop: in the same way that we pass a span of primaries to our core loop, we pass one or more generator C++ classes : cerenkov, scintillation, optical map voxel, etc. Should I sketch out what I'm thinking?

amandalund: Sure, if you could elaborate on or sketch out what you're thinking that would be helpful :).

I think it makes sense to construct the optical primaries outside the optical loop. If we can build the generator actions with the optical offload and core state aux IDs then we could make them part of the main stepping loop, so for now they would look something like:

```c++
template<MemSpace M>
void CerenkovGeneratorAction::execute_impl(CoreParams const& core_params,
                                           CoreState<M>& core_state) const
{
    auto& offload_state
        = get<OpticalOffloadState<M>>(core_state.aux(), offload_id_);
    auto& optical_state
        = get<optical::CoreState<M>>(core_state.aux(), optical_id_);

    auto& num_primaries = optical_state.counters().num_primaries;
    auto& num_new_primaries = offload_state.buffer_size.num_primaries;

    if (num_primaries + num_new_primaries < auto_flush_)
        return;

    auto primaries_size = optical_state.ref().init.primaries.size();
    CELER_VALIDATE(num_primaries + num_new_primaries <= primaries_size,
                   << "insufficient capacity (" << primaries_size
                   << ") for optical photon primaries (total capacity "
                      "requirement of "
                   << num_primaries + num_new_primaries << ")");

    auto& offload = offload_state.store.ref();
    auto& buffer_size = offload_state.buffer_size.cerenkov;

    // Calculate the starting indices in the primary buffer for each thread to
    // generate optical photons
    auto count = exclusive_scan_primaries(
        offload.cerenkov, offload.offsets, buffer_size, core_state.stream_id());

    // Generate the optical primaries from the distribution data
    this->generate(core_params, core_state);

    num_primaries += count;
    num_new_primaries -= count;
    buffer_size = 0;
}
```

(with a similar separate action for scintillation). Then we'd check if there are any buffered primaries and launch the optical loop if so.

sethrj: All right all right, I think this is coming together. What if we start exploring this in the core loop by generalizing the `ExtendFromPrimariesAction`, adding another "track initializer generator" interface class that has "num pending initializers" accessor; and the primaries themselves would just be aux data for that class? We could add some helper classes that would let us generalize the `PrimaryGenerator` class and do the sampling of primaries on device with the device RNG rather than a separate host RNG.

sethrj:
@amandalund Even though this doesn't have enough functionality to be useful, I think it's at a decent merging point... I think we could split up the next steps once this is in?

amandalund (approved):
Works for me! This is a great starting point, and now that the skeleton’s in place we can move forward with the stepping loop and modify these classes as needed as we go. Thanks!

PR #1387, opened 2024-09-01
Rename ExplicitAction to StepAction and add templating

sethrj:
This is a follow-on to #1160 for the optical stepping loop.
- Rename explicit -> step action interfaces and step action order
- Change step action interface to be a template, removing physics dependence from it (and allowing it to be moved to corecel)
- Template and rename begin-run action interface
- Add mutable action interface to support BeginRun action
- Move action registry to corecel/sys now that it's no longer physics dependent
- Add deprecated aliases


amandalund (approved):
Looks good to me! Thanks @sethrj.

PR #1388, opened 2024-09-03
Address minor documentation, contract fixes in CelerOpticalPhysics

drbenmorgan:
Some small review comments in #1348 were left dangling due to auto-merge, so this PR implements them. Per that review these are:

- Typos and Doxygen/Logging clarification fixes
- Mark WLS2 process as unreachable in Geant4 < 10.7
  - Even though a user could request WLS2 be active in Geant4 < 10.7 the setup in `CelerOpticalPhysics` should never trigger construction/use so this is a failsafe more than anything else (but good to have just in case we do end up supported pre-10.7...).


stognini (approved):
Thanks @drbenmorgan !

PR #1389, opened 2024-09-03
Add clang 18 and gcc 14 builds to CI

sethrj:
Since github actions now includes [beta support for ubuntu 24.04 noble numbat](https://github.com/actions/runner-images?tab=readme-ov-file), we can add newer compilers using the vanilla apt-get servers. This adds the latest GCC (14) and Clang (18) versions available on the repository.


sethrj:
@drbenmorgan This passes with GCC 14 (and showed the same spurious warnings as GCC 13) so I think that fits the build?

drbenmorgan (approved):
Looks good to me @sethrj, LGTM!

Issue #1391, opened 2024-09-05
Fix Coulomb scattering applicabiliity and discrete model selection


amandalund:
When trying to run our app with Coulomb scattering enabled, there are some issues which stem from the model/process energy limits:
 - The Coulomb scattering model cross section energy limits are set using [`G4VEmModel::MinPrimaryEnergy`](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/standard/src/G4eCoulombScatteringModel.cc#L134) and are _material dependent_. (The lower bounds are also quite high: order tens of MeV, but still as far as I’ve seen below the 100 MeV MSC energy limit. It looks like the physics list for single Coulomb scattering by default uses a different model that is [valid down to 10 eV.](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/standard/include/G4eDPWACoulombScatteringModel.hh#L117)) This should hopefully not be a problem when we use Wentzel VI MSC with SS where the model limits are set to 100 MeV, but if we do want to run in SS mode we won't have the correct model limits untili we add material-dependent applicability (#907).
 - The model limits are also _different_ from the limits for the Coulomb process cross section.  Unless the MSC energy limit is set for the process/model, the process grid bounds extend below the model bounds, which means even with the correct model limits we could sample the process but have no applicable model. Again, this should only be a problem with SS.
 - Finally, the Coulomb scattering process has the "start from null" flag set to false when building the lambda table. This means the first cross section value will be nonzero. Geant4 and Celeritas extrapolate cross sections outside the grid, so the process cross section will be nonzero for energies below the first grid point. As a result, even with consistent model and process energy limits (e.g. when we use MSC with SS above 100 MeV), with our current discrete model selection method it’s _still_ possible to sample the CS process and have no applicable model.

PR #1392, opened 2024-09-05
Fix Geant4 setup when single Coulomb scattering is enabled

amandalund:
This fixes a few little issues I encountered when trying to run celer-sim with Coulomb scattering enabled (but not the problems with the model limits described in #1391):
 - Construct `G4Proton` in the celer EM physics list when Coulomb scattering is enabled (fixes `Geant4 error: G4ParticleDefinition should be created in PreInit state G4ParticleDefintion::G4ParticleDefinition(): 'PART101' failed`)
 - Always construct `G4eCoulombScatteringModel` with the default value of `isCombined` (avoiding all zero cross sections).
 - Set the polar angle limit between SS and MSC to zero if single Coulomb scattering is used.


src/celeritas/em/distribution/WentzelDistribution.hh
```diff
@@ -178,9 +177,10 @@ CELER_FUNCTION real_type WentzelDistribution::operator()(Engine& rng) const
                                       std::sqrt(particle_.beta_sq()));
         real_type xs = mott_xsec(cos_theta)
                        * ipow<2>(this->calculate_form_factor(cos_theta));
-        if (RejectionSampler(xs, helper_.mott_factor())(rng))
+        if (xs < helper_.mott_factor() * generate_canonical(rng))
```
sethrj: I've spent about 2 hours digging into this and I think the actual issue might be that the Mott ratio has substantially different forms between electron and positron. From https://doi.org/10.1016/j.radphyschem.2013.04.020 :

The electron Mott ratios as a function of angle are:
1-s2 0-S0969806X13002454-gr1_lrg

but for positrons they're bounded by 1:
1-s2 0-S0969806X13002454-gr2_lrg

I looked at Lijian et al [1995], which is the only thing we use for the Mott ratio calculator, and that reference only describes electrons; the newer reference includes both. So I think we need to implement a new calculator based on the 2013 reference. Note a comment from that article:
> Thus extending the treat- ment to positrons (not discussed in Lijian et al., 1995) and to the interaction of electrons with higher Z nuclei. It is worthwhile to remark that the previously available coefficients for electrons, as already noted by Jun et al. (2009), exhibit many typographic errors among those listed in Table 1 of Lijian et al. (1995).

src/celeritas/em/model/CoulombScatteringModel.cc
```diff
@@ -60,6 +60,10 @@ CoulombScatteringModel::CoulombScatteringModel(ActionId id,
         = imported_.energy_grid_bounds(data_.ids.electron, MaterialId{0});
 
     // Check that the bounds are the same for all particles/materials
+    // \todo This is only expected when using Coulomb scattering with the
```
sethrj: the `\todo` command is only parsed as part of a doxygen block, so this would need to be annotated with `/*! ...`

test/celeritas/em/CoulombScattering.test.cc
```diff
@@ -438,46 +438,88 @@ TEST_F(CoulombScatteringTest, distribution)
         = this->cutoff_params()->get(mat_id_).energy(ParticleId{0});
 
     std::vector<real_type> avg_angles;
+    std::vector<real_type> avg_engine_samples;
 
-    for (real_type energy : {1, 50, 100, 200, 1000, 13000})
-    {
-        this->set_inc_particle(pdg::electron(), MevEnergy{energy});
+    auto sample = [&](PDGNumber pdg) {
```
sethrj: This lambda is getting out of hand 😝

src/celeritas/em/distribution/WentzelDistribution.hh
amandalund: Oh that's fantastic, thanks for digging into that and for the reference! Let me see if I can make that update to the Mott ratios first and then revert this change.

sethrj (approved):

Issue #1390, opened 2024-09-03
Refactor primary construction into aux data + generator


sethrj:
In the optical transport loop, we want to be able to use Cerenkov/Scintillation "generators" (which have signature `(RNG&) -> Primary`) to produce optical photons on device. Like buffered track initializers, we'd like to be able to fill up track slots, pause partway (if there aren't enough track slots), and then fill any empty slots on the next loop. We'll also want to do this to support sampling for optical maps: a generator could be initialized with a volume ID and a region of space and only accept samples within that region.
We should consider doing something similar for *core* tracks, since there is a memory usage/bandwidth advantage to initializing tracks on device rather than on host. (We'd also be able to use a consistent RNG.)

I propose we move the StepActionOrder "generate" before the "initialize" that fills empty track slots:
- Change "extend from primaries" to be the only class that handles primaries; it would take them directly from the users
- Refactor the "track initializer" buffer to likewise be an action that will fill empty slots
- New future "generator" actions can be added to replace `phys/PrimaryGenerator.hh` or define new generalized distributions (e.g., sample N tracks from a gaussian)

Perhaps we could also use this refactor to think of a way to make track IDs deterministic (currently, tracks from secondaries get their event ID-associated track count using an `atomic_add`). This would involve *removing the track ID from the track initializer* and instead constructing new track IDs in the initialization stage.

### GeneratorInterface

- Has a GeneratorId created at setup
- Given an auxiliary vec (so that each generator class stores the remaining count as part of the state), set the number of tracks left to initialize to an index `GeneratorId` in a shared array of "number tracks left to generate".
- We could consider also someday having a "priority" return as well, maybe returning the amount of expected secondaries from each one (using the total energy)?
- For charge partitioning, we'll also need to split these into two counts: front and back

Interface instances:
- The ExtendFromPrimariesAction is now a generator, its aux state contains buffered primaries, all for a single event
- ExtendFromSecondaries at the end of a step does in-place replacement, and its initializers should store the "touchable" (unique volume ID) and parent ID
- A more tightly integrated Geant4 initializer could also include touchable information (and even other metadata) to reduce initialization costs
- CherenkovGeneratorAction, ScintGeneratorAction are also generator interfaces

### GeneratorRegistry

- Generators are added to it during construction, using `GeneratorId next_id`

### SlotPartitioner

This should take the place of most of the track initialization action (aside from the actual track initialization).

- Runs at the beginning of the stepping loop
[...]

PR #1393, opened 2024-09-06
Refactor primaries as `aux` data

sethrj:
Almost none of the stepping loop has to access the primaries, and right now the storage and usage of them is split across the CoreState and the ExtendFromPrimariesAction (EFPA). This PR refactors the primary storage, removing it from the main state and adding it as "aux" data to the EFPA. It's expected that the higher-level code (whether optical transport or core transport) determines whether primaries are to be used as the source of track initializers: for example, we could use a generalized source description to generate track initializers on device.

This completes the first part of #1390.


sethrj:
@amandalund This is finally all fixed and working!

amandalund:
Sorry for the delay! Looking at this today.

sethrj:
No worries!

amandalund (approved):
Nice refactor! All looks good to me, thanks @sethrj.

PR #1394, opened 2024-09-06
Add separate configure option for on-device assertions

sethrj:
Assertions and print statements use a fair bit of local memory and drastically slow NVCC as a result. This adds an option `CELERITAS_DEVICE_DEBUG` to disable debug assertions only on device to improve build times by a factor of ~3 on multicore systems.

|  | Before     | After | 
| --- | --- | --- |
| real | 2m37.039s  | 0m48.314s |
| user | 33m41.604s | 25m49.679s |

Part of the slowdown is just because CMake doesn't "schedule" the build commands to put the slower-building .cu.o files before the .cc.o files, leading to a much higher real time...


esseivaju (approved):
Looks good!

.github/workflows/build-docker.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - name: Check out Celeritas
         uses: actions/checkout@v3
         with:
-          fetch-depth: 255
+          fetch-depth: 383
```
pcanal, line 68: What is the 'cost'/'loss' if this number is too low?

sethrj: Cost is finding and downloading extra Git objects, potentially slowing down the CI. I didn't think we'd ever reach 255 TBH.... we should've released 0.5 months ago 😅instead of waiting for "just around the corner" optical physicsl.

pcanal (approved):

.github/workflows/build-docker.yml
pcanal: What about the cost/loss of functionality in the reverse case, i.e. in the extreme:
```
fetch-depth: 1
```
?

sethrj: Then we get CI errors like this because it can't detect the version because the tag isn't within the last 1 depth: https://github.com/celeritas-project/celeritas/actions/runs/10724334134/job/29739743532

pcanal: humm ... too bad you can't do `git clone --depth <since this tag reference> ....` :)

And actually you can!
```
git clone --depth 1 git@github.com:celeritas-project/celeritas.git
cd celeritas 
git fetch --shallow-exclude v0.5.0-dev
git fetch --deepen=1
git log v0.5.0-dev
```
```
commit b1e52f2001c73fd887ab26dfa100480b5b146167 (grafted, tag: v0.5.0-dev)
Author: Amanda Lund <alund@anl.gov>
Date:   Wed Jan 17 17:26:02 2024 -0600

    Add Cerenkov distribution and generator (#1080)
```
The "grafted" and total size shows things are as expected.

So with those 3 commands (assuming the tag that is needed is known), we can have the minimal download size and no risk of missing commits :)

PR #1395, opened 2024-09-08
Extend Mott coefficients to include positrons and higher Z elements

amandalund:
As @sethrj [pointed out](https://github.com/celeritas-project/celeritas/pull/1392#discussion_r1748069859) in #1392, the coefficients used to calculate the ratio of the Mott to Rutherford cross sections are provided in Boschini et al (2013) for 1 <= Z <= 118 for electrons and positrons. Geant4 includes that data in G4MottData.hh, but only for electrons and only for Z up to 92; this is what we were previously doing as well. This PR extends the data to include coefficients for higher Z elements and positrons.


src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -179,772 +181,1241 @@ real_type WentzelOKVIParams::calc_nuclear_form_prefactor(IsotopeView const& iso)
...
+ * (2013), 39-66. Geant4 provides the same data in G4MottData.hh, but only for
+ * 1 <= Z <= 92.
+ */
+auto WentzelOKVIParams::get_electron_mott_coeffs(AtomicNumber z) -> CoeffMat
+{
+    CELER_EXPECT(z);
+
+    // clang-format off
+    static CoeffMat const electron_mott_coeffs[]
+        = {{{{1.00000e+00, 2.67363e-08, 7.11530e-08, -9.77030e-08, -6.69132e-07, -3.09263e-07},
```
amandalund, line 197: I could also format this differently...

sethrj: I think the formatting is fine! It's not like they're readable anyway...

sethrj (approved):
LGTM! Where did you get these numbers in tabulated form?? I also noticed in the Boschini paper a casual remark about the previous papers having typographical errors... and finally, do we know whether Geant4 uses these? Looking at  the "single scattering" physics list on master I see
```c++
  if(param->UseMottCorrection()) {
    ss = new G4eDPWACoulombScatteringModel();
  } else {
    ss = new G4eCoulombScatteringModel(false);
  }
```

amandalund:
I pulled the tables from [here](https://www.sciencedirect.com/science/article/pii/S0969806X13002454?via%3Dihub#t0015).  It looks like Geant4 is also using the values from Boschini, not Lijian: they reference a 2011 Boschini paper in G4MottData.hh and that the coefficients were recalculated using the same procedure as in Lijian. And yeah, I also noticed when looking into #1391 that Geant4 by default uses that other model in the single Coulomb scattering physics list, which I don't think uses these?

sethrj:
Sounds good. I was more wondering if you OCR/copy-pasted/manually transcribed them, or whether you found a database that the authors made available. Thanks!

amandalund:
Yeah I didn't see any database, so copy-pasted + some python parsing/reformatting.

sethrj:
Cool thanks! Hopefully Celeritas can be a resource in the future for anyone who needs to use those coefficients in the future...

PR #1398, opened 2024-09-10
Test and fix nuclear form factors

sethrj:
This refactors some of the Wentzel distribution internals to make it more testable, and fixes the uniform-uniform nuclear form factor (the units were not correct, and it used nuclear mass in MeV instead of atomic mass number). I also went back through the original citations to derive the "magic" number used as a special case by hydrogen (which is actually wrong in the fourth digit) : see 59ab10fe2aba669894bb36aaf9451a555bbcc319 . With the testing it's easy to verify that the nuclear form factors are similar in value, since they should all be approximating the same thing.

Based on the limited results for numerical stability, it looks like some of these form factors will be more appropriate for different regimes in Z and momentum transfer, so it's likely not ideal to have a single global form factor.


src/celeritas/em/distribution/NuclearFormFactors.hh
```diff
@@ -0,0 +1,286 @@
...
+ *      Comments on multiple scattering of high-energy muons in thick layers,
+ *      Nuclear Instruments and Methods in Physics Research Section A:
+ *      Accelerators, Spectrometers, Detectors and Associated Equipment 488
+ *      (2002) 282–294. https://doi.org/10.1016/S0168-9002(02)00478-3.
+ *
+ * \todo Instead of using this coarse parameterization, we should add nuclear
+ * radius to the isotope properties for a more accurate treatment, and
+ * construct these classes directly from the atomic radius.
+ */
+class ExpNuclearFormFactor : public NuclearFormFactorTraits
```
sethrj: I've used this inheritance as a shortcut but could easily add type aliases to each of these classes.

src/celeritas/em/distribution/NuclearFormFactors.hh
```diff
@@ -0,0 +1,286 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Gaussian nuclear form factor.
+ *
+ * This nuclear form factor corresponds \c NuclearFormFactorType::gaussian and
+ * assumes the nuclear charge decays exponentially from its center. Its
+ * prefactor has the same value as the exponential.
+ */
+class GaussianNuclearFormFactor : public ExpNuclearFormFactor
```
sethrj: If we do as my 'todo' suggests and make these functionally dependent on nuclear radius instead of "prefactor" then we should remove this inheritance.

test/celeritas/em/CoulombScattering.test.cc
```diff
@@ -368,96 +380,101 @@ TEST_F(CoulombScatteringTest, wokvi_transport_xs)
...
-                                             wentzel_->host_ref(),
-                                             this->particle_track(),
-                                             this->direction(),
-                                             material,
-                                             isotope,
-                                             el_id_,
-                                             cutoffs);
-
-        for ([[maybe_unused]] int i : range(num_samples))
+        for (real_type energy : {0.05, 0.1, 0.5, 1.0, 10.0})
```
amandalund: Should we be testing energies above the 100 MeV limit?

src/celeritas/em/distribution/NuclearFormFactors.hh
```diff
@@ -0,0 +1,290 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the gaussian folded form factor.
+ */
+CELER_FUNCTION real_type
+GaussianNuclearFormFactor::operator()(MomentumSq target_momsq) const
+{
+    CELER_EXPECT(target_momsq >= zero_quantity());
+    return std::exp(-2 * value_as<InvMomentumSq>(this->prefactor()) * target_momsq.value());
```
amandalund: Might need a clang-format on this file...

src/celeritas/em/distribution/NuclearFormFactors.hh
```diff
@@ -0,0 +1,290 @@
...
+  private:
+    real_type prefactor_;  // Function of nuclear radius [(MeV/c)^{-2}]
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Gaussian nuclear form factor.
+ *
+ * This nuclear form factor corresponds \c NuclearFormFactorType::gaussian and
+ * assumes the nuclear charge decays exponentially from its center. Its
```
amandalund: Has a Gaussian charge distribution?

amandalund (approved):
Nice! Thanks for the fix and additional documentation @sethrj.

src/celeritas/em/distribution/NuclearFormFactors.hh
sethrj: Caught me copy-pasting 😬

test/celeritas/em/CoulombScattering.test.cc
sethrj: Yeah good point, and probably below 50 keV too. I chose these since they were the energies used in the paper showing Mott factors, but of course we're going to be using the scattering kernel outside of that.

test/celeritas/em/distribution/NuclearFormFactors.test.cc
```diff
@@ -0,0 +1,103 @@
...
+    // clang-format on
+    EXPECT_VEC_SOFT_EQ(expected_result, result);
+}
+
+TEST_F(NuclearFormFactorsTest, uniform_uniform_folded)
+{
+    if (CELERITAS_REAL_TYPE != CELERITAS_REAL_TYPE_DOUBLE)
+    {
+        GTEST_SKIP() << "UU suffers from numerical errors for anything below "
+                        "~2 MeV";
```
sethrj: FYI, the numerical errors in this approximation ended up leading to other RejectionSampler failures for positrons when using the "flat" form factor.

sethrj:
@hhollenb I'll wait to merge till you review this since you did your initial implementation. I think the way that having separate "nuclear form factor" classes allows us to test their correctness, and indeed discover that the numeric instabilities for small radii/small momentum lead to errors, shows the real value of having lots of small classes in the codebase.

test/celeritas/em/distribution/NuclearFormFactors.test.cc
amandalund: Good to know! The energy limits for this Coulomb model seem pretty high at least, so hopefully something we'd be unlikely to run into.

sethrj:
Ping @hhollenb , not urgent but if you're bored.... 😅

PR #1399, opened 2024-09-10
Add action to generate optical primaries from Cerenkov distributions

amandalund:
This adds an action to create optical primaries from the Cerenkov distribution data; I'll follow up with the scintillation action after. We'll likely need to refactor some of this once the primary construction refactor is complete. I still need to add a better test, but this should otherwise be ready for a look.


src/celeritas/optical/CoreState.hh
```diff
@@ -34,8 +34,7 @@ struct CoreStateCounters
 {
     // Initialization input
     size_type num_vacancies{};  //!< Number of unused track slots
-    size_type num_primaries{};  //!< Number of primaries to be converted
-    size_type num_initializers{};  //!< Number of track initializers
+    size_type num_primaries{};  //!< Number of primaries
```
sethrj: I wonder if we should be calling these "initializers" rather than primaries, and the simplification versus the main stepping loop is that we don't define a user-friendly "Primary" class but just reuse the initializers? (And maybe instead of worrying about having separate track/sim/geo initializer data structures, we just use the single existing optical::Primary class (but perhaps rename it).)

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
```diff
@@ -0,0 +1,101 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Generate Cerenkov photons from optical distribution data.
+ */
+struct CerenkovGeneratorExecutor
+{
+    //// DATA ////
+
+    RefPtr<CoreStateData, MemSpace::native> state;
+    NativeCRef<celeritas::optical::MaterialParamsData> const material;
```
sethrj, line 34: The materials should be part of the core optical state I think?

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
```diff
@@ -0,0 +1,101 @@
...
+        // Find the distribution this thread will generate from
+        size_type dist_idx = find_distribution_index(offsets, primary_idx);
+        CELER_ASSERT(dist_idx < size.cerenkov);
+        auto const& dist = offload_state.cerenkov[DistId(dist_idx)];
+        CELER_ASSERT(dist);
+
+        // Generate one primary from the distribution
+        optical::MaterialView opt_mat{material, dist.material};
+        celeritas::optical::CerenkovGenerator generate(opt_mat, cerenkov, dist);
+        optical_state->init.primaries[PrimaryId(primary_idx)] = generate(rng);
```
sethrj: As a future optimization for CPU we should consider generating consecutive primaries from the same generator object... even though that would eliminate GPU/CPU reproducibility.

src/celeritas/optical/detail/OpticalGenAlgorithms.cc
```diff
@@ -69,6 +69,38 @@ count_num_photons(GeneratorDistributionRef<MemSpace::host> const& buffer,
...
+    size_type size,
+    StreamId)
+{
+    CELER_EXPECT(!buffer.empty());
+    CELER_EXPECT(size > 0 && size <= buffer.size());
+    CELER_EXPECT(offsets.size() == buffer.size());
+
+    auto* data = static_cast<celeritas::optical::GeneratorDistributionData*>(
+        buffer.data());
+    auto* result = static_cast<size_type*>(offsets.data());
```
sethrj: The static cast is a little confusing (I had to dig back into Collection to realize this is a memspace-safe `ObserverPtr`, as opposed to casting for an inheritance hierarchy). Maybe use `.get()` (unchecked but safe here since the code's not templated and in a .cc file) or add a comment about casting from observer pointer?

src/celeritas/optical/detail/OpticalGenAlgorithms.cu
```diff
@@ -82,6 +83,37 @@ count_num_photons(GeneratorDistributionRef<MemSpace::device> const& buffer,
...
+    auto stop = thrust::transform_inclusive_scan(thrust_execute_on(stream),
+                                                 data,
+                                                 data + size,
+                                                 result,
+                                                 GetNumPhotons{},
+                                                 thrust::plus<size_type>());
+    CELER_DEVICE_CHECK_ERROR();
+
+    // Copy the last element (accumulated total) back to host
+    return *(stop - 1);
```
sethrj: Out of curiosity, is this a synchronous call or does the iterator get attached to the inclusive scan's stream?

src/celeritas/optical/detail/OpticalUtils.hh
```diff
@@ -44,4 +45,46 @@ inline CELER_FUNCTION real_type energy_to_wavelength(units::MevEnergy energy)
...
+ */
+inline CELER_FUNCTION size_type calc_local_work(ThreadId tid,
+                                                size_type num_threads,
+                                                size_type total_work)
+{
+    CELER_EXPECT(tid);
+    CELER_EXPECT(num_threads > 0);
+    CELER_EXPECT(total_work > 0);
+
+    return total_work / num_threads + (tid < total_work % num_threads ? 1 : 0);
```
sethrj: This could be a nice little algorithm to add next to `ceil_div` in `corecel/math/Algorithms.hh`... I've seen similar for parallelizing over MPI as well. (It'd also be nice to add a test to it.) I could also see writing it as a functional object `LocalWorkCalculator` where it's constructed with the "global" arguments and called with the "local" ID.

src/celeritas/optical/detail/OpticalUtils.hh
```diff
@@ -44,4 +45,46 @@ inline CELER_FUNCTION real_type energy_to_wavelength(units::MevEnergy energy)
...
+    CELER_EXPECT(!offsets.empty());
+
+    auto iter = celeritas::lower_bound(offsets.begin(), offsets.end(), value);
+    CELER_ASSERT(iter != offsets.end());
+
+    if (value == *iter)
+    {
+        ++iter;
+    }
+    return iter - offsets.begin();
```
sethrj, line 70: This would be good to add a test for.

sethrj (approved):
Just a few minor things!

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
amandalund: So the original way of having each thread generate all the primaries from a single distribution?

src/celeritas/optical/CoreState.hh
amandalund: I think that makes more sense.

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
sethrj: Yep exactly. Future optimization though, let's get it working efficiently on GPUs first and foremost as always.

amandalund: Yeah, once we we have at least an initial implementation and some test problems and can study the performance. Should be pretty straightforward though: we can change the inclusive scan to exclusive and add a separate `Executor` for the CPU that does things that way and the rest should be unchanged.

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
amandalund: Hmm but we'd need the optical params which I don't think we have access to, right?

src/celeritas/optical/detail/OpticalGenAlgorithms.cu
amandalund: Oh that is a very good question. I'm not positive, but I suspect it is synchronous...

We do this also in our core track init algorithms. A little while back I had refactored those to get rid of the `thrust::remove_if`, which @esseivaju had shown in his profiling requires synchronizing with the device. I didn't end up seeing much improvement, but this might explain why. I'm going to revisit that but using our async `Copier` to get the result of the scan instead...

sethrj: Cool! This is pretty subtle... and if this actually does improve performance/remove the synchronize, then maybe a little `Copier`-like class that simply returns a simple element (via an async copy) could be useful.

src/celeritas/optical/detail/CerenkovGeneratorExecutor.hh
sethrj: Ah right, I forgot that the core state above was the *main* core state, not the *optical* one. Thanks!

amandalund:
Thanks @sethrj, really helpful feedback!

PR #1402, opened 2024-09-11
Store OBZ data during runtime

elliottbiondo:
In this MR, OBZ data is extracted during construction and stored on a `OrientedBoundingZoneRecord` object during runtime. This allows `OrientedBoundingZone` objects to be constructed during runtime, which will facilitate safety distance calculations as well as accelerated distance-to-boundary calls in background volumes.

In addition, the `OrientedBoundingZone` class has been refactors such that transformation operations are never repeated.


elliottbiondo:
@sethrj there might be a few CI errors for me to look into, but this is otherwise ready for your review.

elliottbiondo:
@sethrj I am getting weird Windows error related to adding to collections in `UnitInserter`. Could you took at `process_obz_record` and let me know if I am doing anything non-standard?

sethrj:
@elliottbiondo Sorry I didn't see the message earlier; I think it's because the specialization `std::hash<Real3>` doesn't exist, and that's used under the hood for the dedupe builder. The question in my mind is, why do GCC and Clang work...

sethrj:
The simplest solution is to store the Real3 directly on the `ObzRecord` rather than trying to hash it. Using the deduplication as an "optimization" could be future work... and in that case you could simply use the existing `fast_real` storage and just insert a `Span<real_type>`.

elliottbiondo:
@sethrj I changed it to a normal, non-dedup, `CollectionBuilder`, but let me know if you prefer a different approach.

elliottbiondo:
@sethrj CI fixed. You can remove draft status and review.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -18,21 +18,34 @@ namespace celeritas
...
+ * The safety distance can be approximated as the minimum distance to the inner
+ * or outer box. For points lying between the inner and outer boxes, the safety
  * distance is zero.
  */
 class OrientedBoundingZone
 {
   public:
-    //!@{
-    //! \name Type aliases
+    //!@{ ! \name Type aliases
```
sethrj: Something messed up the comment here, also strangely deleting a line in the class comment above.

src/orange/detail/OrientedBoundingZone.hh
```diff
@@ -67,29 +80,53 @@ class OrientedBoundingZone
     inline CELER_FUNCTION SignedSense calc_sense(Real3 const& pos);
 
   private:
+    //!@{
+    //! \name Type aliases
+
+    //! Enum to distinquish between inner and outer bboxes
```
sethrj:
```suggestion
    //! Enum to distinguish between inner and outer bboxes
```
Q->G 😉

src/orange/detail/UnitInserter.cc
```diff
@@ -373,6 +382,38 @@ VolumeRecord UnitInserter::insert_volume(SurfacesRecord const& surf_record,
...
+{
+    CELER_EXPECT(obz_input);
+
+    OrientedBoundingZoneRecord obz_record;
+
+    // Set half widths
+    auto inner_hw_id = fast_real3s_.push_back(
+        calc_half_widths(calc_bumped_(obz_input.inner)));
+    auto outer_hw_id = fast_real3s_.push_back(
+        calc_half_widths(calc_bumped_(obz_input.outer)));
```
sethrj: There's no point in having an extra indirection into a separate vector of Real3 if you're not deduplicating or saving space by optimizing for the "no bbox" case with a false ID. Just replace `Array<Real3Id, 2> hw_ids;` in the OBZ record with `Array<Real3, 2> hw_ids;` and you'll save 2 ints per record and reduce memory accesses by avoiding the extra indirection.

src/orange/detail/UnitInserter.cc
```diff
@@ -373,6 +382,38 @@ VolumeRecord UnitInserter::insert_volume(SurfacesRecord const& surf_record,
...
+
+    // Set offsets
+    auto inner_offset_id = insert_transform_(VariantTransform{
+        std::in_place_type<Translation>, calc_center(obz_input.inner)});
+    auto outer_offset_id = insert_transform_(VariantTransform{
+        std::in_place_type<Translation>, calc_center(obz_input.outer)});
+    obz_record.offset_ids = {inner_offset_id, outer_offset_id};
+
+    // Set transformation
+    obz_record.transform_id = obz_input.transform_id;
```
sethrj, line 411: I don't understand why there's an extra set of variant transforms, especially on top of an *additional* transform. The offsets are only ever translations (or zero).

src/orange/OrangeData.hh
```diff
@@ -141,16 +142,22 @@ struct ConnectivityRecord
...
+
+    // Transformation from the OBZ coordinate system to the unit coordinate
+    // system
     TransformId transform_id;
 
     //! True if assigned
     explicit CELER_FUNCTION operator bool() const
     {
-        return inner_hw_id && outer_hw_id && transform_id;
+        return !half_widths.empty() && !offset_ids.empty() && transform_id;
```
sethrj: The first two tests are trivially false (array is fixed at size 2), so perhaps instead:
```suggestion
        return offset_ids[0] && offset_ids[1] && transform_id;
```

src/orange/detail/UnitInserter.cc
```diff
@@ -191,6 +191,19 @@ UnitInserter::UnitInserter(UniverseInserter* insert_universe, Data* orange_data)
...
+    // from a boundary so that the bbox will enclose the point even after a
+    // potential bump
+    calc_bumped_ = BoundingBoxBumper<fast_real_type, real_type>(
+        [&tol = orange_data_->scalars.tol] {
+            Tolerance<real_type> bbox_tol;
+            bbox_tol.rel = 2 * tol.rel;
+            bbox_tol.abs = 2 * tol.abs;
+            CELER_ENSURE(bbox_tol);
+            return bbox_tol;
+        }());
```
sethrj: Better to put this as a member data initializer rather than assigning it later; you could turn the lambda into an anonymous-namespace helper function to make it a bit cleaner.

src/orange/detail/UnitInserter.cc
```diff
@@ -233,18 +233,25 @@ UniverseId UnitInserter::operator()(UnitInput&& inp)
...
-            bboxes.push_back(calc_bumped(inp.volumes[i].bbox));
+            bboxes.push_back(calc_bumped_(inp.volumes[i].bbox));
         }
         else
         {
             bboxes.push_back(BoundingBox<fast_real_type>::from_infinite());
         }
 
+        // Add oriented bounding zone record
+
```
sethrj:
```suggestion
```

sethrj (commented):
One last round of comments.

src/orange/OrangeData.hh
elliottbiondo: Any reason not to check half-widths as well?

sethrj: Since they're `Real3` and not `OpaqueId` there's nothing meaningful to check

elliottbiondo:
@sethrj this should be ready pending CI

sethrj (approved):

Issue #1296, opened 2024-06-28
Replace StreamStore and helpers with reduction function


sethrj:
The `StreamStore` is no longer needed since #1278, *except* that diagnostic classes still use it for doing a reduction over multiple states for output. I think we need to break apart this functionality since we really don't want the params to be mutable and keep access to multiple states.

1. Add an `EndRunActionInterface` that takes a the core params, a core state, and a `Span<CoreState<M> const*>` of all states for performing a reduction in a multithread context. The action itself should know whether to do a global reduction or otherwise: probably it should always reduce to `StreamId{0}`. The action should check that `state.stream_id() < all_states.size() && all_states[state.stream_id().get()] == &state`. We probably want to add an MPI communicator to the core params so that we can perform reductions with dynamic parallelism.
2. Define an output adapter that can take a state (or aux state data plus memspace?) and write that at the end of the program. This requires lifetime considerations: we either want the state itself to be shared outside of the stepper, or the aux state vec should become a shared pointer. I'm leaning toward the latter...

- Once we do this we can also get rid of the `max_streams` parameter.
- The reduction capability can later be extended so that in addition to begin/end run, we can have begin/end batch: see #809 
- With reductions, we may need to be more careful about parallel execution. The `begin_run` action is taken when a `Stepper` is created, which (for celer-sim) may happen separately in warmup and inside the parallel loop, or (for the `accel` interface) will happen during the run manager's `BeginOfRunAction` call. We should make sure the `end_run` action is executed in parallel...

sethrj:
This is going to be troublesome for the different ways that we execute across threads. It's easy if we're doing OpenMP and know that all states are going to be starting and stopping at a synchronization point, it's easy to send a vector of state references when each thread is finished but the states are still allocated. However, if we're running through Geant4 MT, the "EndOfRunAction" will be called individually on each thread and then on the "master" thread. But we have to deallocate the state *on the original thread*, which creates an ordering issue.

Perhaps instead of trying to make our destructor ordering work with Geant4's threading model, we add special cases for anything that has to add Geant4 objects:
- Hit manager
- Navigation states (if using geant geometry)

Now that we have a `LocalTransporter` we could have it manage shared pointers to the hit processors, and then give weak pointers (for safety) + raw pointers (for performance, knowing that since the hit processes are only "shared" within a single thread we don't have to use locking) to the hit manager.

[...]

PR #1403, opened 2024-09-13
Change LocalTransporter deallocation semantics

sethrj:
In preparation for a multi-stream "end run gather" (https://github.com/celeritas-project/celeritas/issues/1296) that will apply to timers, calorimeters, etc., this refactors the local transporter and hit processor so that the *hit processor* is owned by the local thread and *referenced* by the hit manager, as opposed to being *owned* by hit manager. This means that when the local transport is allocated/deallocated (which happens at Begin/EndOfRun on the worker thread) the hit processor will also be allocated/deallocated, necessary to make Geant4 thread-local allocators play nicely. To make this safer, we keep a weak pointer on the hit manager that does a debug check for hit processor being alive.

As part of the refactoring I devised a slightly cleaner way of setting up the output registry to keep the detector output, eliminating a couple of "hack: remove in v0.5 " bits of code.

While looking through the debug output I also noticed that the "looping track" warning was not very clear, so I refactored it and now have the tracking cut action print at a debug level even if the track is not in an "error" state:
```
[1/2] ../src/celeritas/global/alongstep/detail/PropagationApplier.hh:164: error: Track (pid=1, E=1.17434 MeV) is looping after 100 steps
[1/2] ../src/celeritas/phys/detail/TrackingCutExecutor.hh:72: debug: Killing track (ID 1926 of event 5, track slot 935) at {-17.844418516047, -25.727498848885, -1072.3128574325} [cm] along {-0.096423674800004, 0.44785031455201, -0.88889401544488}: depositing 1.15475 [MeV] in volume 7
```


src/accel/detail/HitManager.cc
```diff
@@ -243,20 +255,17 @@ void HitManager::setup_particles(ParticleParams const& par)
 
 //---------------------------------------------------------------------------//
 /*!
- * Ensure the local hit processor exists, and return it.
+ * Return teh local hit processor.
```
amandalund:
```suggestion
 * Return the local hit processor.
```

src/celeritas/global/alongstep/detail/PropagationApplier.hh
```diff
@@ -161,7 +161,12 @@ PropagationApplierBaseImpl<MP>::operator()(CoreTrackView& track)
                 && sim.is_looping(particle.particle_id(), particle.energy()))
             {
 #if !CELER_DEVICE_COMPILE
-                CELER_LOG_LOCAL(error) << "Killing looping track";
+                CELER_LOG_LOCAL(error)
```
amandalund: Should this be a debug message since these tracks don't get an error status?

amandalund (approved):
Thanks for the refactor (and for getting rid of some of those hacks)!

src/celeritas/global/alongstep/detail/PropagationApplier.hh
sethrj: Yeah, since we're printing the tracking cut at a "debug" level I agree this should get the same.

PR #1405, opened 2024-09-15
Improve multithread GPU performance by removing subtle device sync

amandalund:
@esseivaju found in his profiling that in our "extend from secondaries" action there was a device synchronization:

> that’s what I was saying about having to synchronize each step. You can see that in the extend-from-secondaries action::remove-if-alive scope we have to synchronize with the device. So, we cannot push the last few kernels for that step or move on to the next step before the GPU has finished executing the current step. You can see in the top line there is a gap of ~1ms in the GPU utilization at the end of a step / between steps
image

@sethrj [questioned recently](https://github.com/celeritas-project/celeritas/pull/1399#discussion_r1757094890) whether the way we are copying the result of a scan to the host might be synchronous, and I think that is the culprit. This adds a little helper class to copy a single value to the host and does the copy of the scan result asynchronously.

I ran the regression problems with `merge_events` off; here’s the speedup in the throughput with this change relative to develop:
rel-throughput-remove-sync

### DIFF
--- a/src/celeritas/track/detail/TrackInitAlgorithms.cu
+++ b/src/celeritas/track/detail/TrackInitAlgorithms.cu
@@ -74,7 +74,7 @@ size_type exclusive_scan_counts(
     CELER_DEVICE_CHECK_ERROR();
 
     // Copy the last element (accumulated total) back to host
-    return *(stop - 1);
+    return ItemCopier<size_type>{stream_id}(stop.get() - 1);
 }
 
 //---------------------------------------------------------------------------//
--- a/src/corecel/data/Copier.hh
+++ b/src/corecel/data/Copier.hh
@@ -52,6 +52,31 @@ class Copier
 };
 
 //---------------------------------------------------------------------------//
+/*!
+ * Copy a value from device to host.
+ *
+ * The source of the data to copy is the function argument.
+ */
+template<class T>
+class ItemCopier
+{
+    static_assert(std::is_trivially_copyable<T>::value,
+                  "Data is not trivially copyable");
+
+  public:
+    //! Default constructor
+    ItemCopier() = default;
+
+    //! Also construct with a stream ID to use for async copy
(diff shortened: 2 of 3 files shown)


esseivaju:
I'm pretty sure the copy is still synchronous because we're copying back to [pageable memory](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior), we'd have to allocate pinned for the counters to make the copy truly asynchronous. From your plot, it's still a nice improvement as the copy is now issued in a stream instead of using the default stream which synchronizes with the device! 👍

sethrj:
Good lord that's a shocking improvement. I'll see what frontier does...

amandalund:
Yes sorry that's right, not fully asynchronous, but no longer on the default stream.

sethrj:
@amandalund Frontier shows no performance improvement at all, unfortunately. @esseivaju would you be able to replicate this on Perlmutter?

amandalund:
@sethrj just making sure, did you update `omp_threads` in the script (right now it's set to 1 when running on the device)? I guess you should still see improvement with celer-g4 either way though...

sethrj:
Oh, this was with the default values. So I guess that's only using 1 CPU per GPU so it shouldn't show much speedup. Sorry, I didn't even notice `merge_events` being off 😅

sethrj (approved):
The implementation is slick and the results are sick! 😄

esseivaju:
To follow on my comment on that PR, as this resurged working on the CHEP proceedings. I tried using pinned memory for the transfer in extend-from-secondaries. See [that commit](https://github.com/esseivaju/celeritas/commit/bb3242ba1d50ee1b8abdc6e80176ac9a8022d499). It doesn't make a difference because we immediately need to value on the host, so we need to perform a stream synchronization anyway.

PR #1409, opened 2024-09-17
Fix construction of single-volume Geant4 with ORANGE and activate examples

sethrj:
When #1180 was completed, I forgot to enable the downstream examples for ORANGE geometry too. Running them led to a debug assertion failure that shows up in the simplest case possible where the world volume is the only one that exists.

I also updated the Geant4 error messages to include the "which" type; otherwise some of the messages (specifically, "not implemented") don't make sense as displayed.

Before:
```
*** G4Exception : celer0001
      issued by : celeritas/phys/PhysicsParams.cc:97
physics

```

After:
```
*** G4Exception : celer0001
      issued by : celeritas/phys/PhysicsParams.cc:97
Celeritas implementation error: feature is not yet implemented: physics

```


sethrj:
Ping @drbenmorgan @elliottbiondo ; should be an easy review.

src/corecel/io/StringUtils.cc
```diff
@@ -50,7 +52,21 @@ bool is_ignored_trailing(unsigned char c)
 }
 
 //---------------------------------------------------------------------------//
-// Return a string view with leading and trailing whitespace removed
+/*!
+ * Test C strings for equality allowing null pointers for LHS.
```
elliottbiondo: Document that if both LHS and RHS are null, it returns false

src/orange/g4org/ProtoConstructor.cc
```diff
@@ -42,12 +44,24 @@ SPConstObject make_explicit_background(LogicalVolume const& lv,
...
+        interior = std::make_shared<AnyObjects>(lv.name + ".children",
+                                                std::move(children));
+    }
+    else if (children.size() == 1)
+    {
+        interior = std::move(children.front());
+    }
+    else
+    {
+        // Rare case (world is the only volume!)
```
elliottbiondo: Is there an opportunity here to use a `LIKELY` macro above?

elliottbiondo (approved):

PR #1410, opened 2024-09-17
Refactor action sequence into action group plus extras

sethrj:
This refactors the grouping functionality "action sequence" into a separate class that's reusable for optical stepping as it doesn't contain core-specific components.


amandalund (approved):
Looks good to me, thanks @sethrj!

PR #1414, opened 2024-09-18
Add action for generating scintillation photons from distributions

amandalund:
Mostly boilerplate and mostly identical to the Cerenkov actiion.


sethrj (approved):
Looks good. For future work we should define a template interface that handles the common parts of the generators.

Also I was thinking while writing the optical stepping loop: we really should have the generator actions be *optical* not *core*, since we might want to generate more initializers without allocating the extra space for them. (I.e., we could get away with a relatively small number of initializers *and* optical track slots while still having a lot of EM tracks.) For now of course let's just allocate as many initializers as we need.

amandalund:
Agreed, I originally had them in the optical loop, but after our discussion in https://github.com/celeritas-project/celeritas/pull/1386#discussion_r1740208258 moved them to the core loop so we can access all the necessary data (but figured we'd be doing quite a bit of refactoring).

sethrj:
Right, I'm so forgetful 🫠

PR #1415, opened 2024-09-19
Simplify UnitProto after building a CsgUnit

esseivaju:
Add an option to build a `CsgUnit` using demorgan's boolean simplification and making the `CsgTree` suitable for infix evaluator.

Requires #1418


src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -17,6 +17,17 @@ namespace celeritas
 {
 namespace orangeinp
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Result of a DeMorgan simplification.
+ */
+struct DeMorganSimplifierResult
```
sethrj: This could be a general "simplification" mapping. Perhaps it should be moved to CsgTree itself (imagine a function that runs "simplify" on the tree itself)?

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -17,6 +17,17 @@ namespace celeritas
...
+{
+    CsgTree tree;
+    std::vector<NodeId> equivalent_nodes;
```
sethrj: Old node ID is the index, new node ID is the value? (Maybe `new_nodes`?)

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -391,7 +391,7 @@ TEST_F(CsgTreeUtilsTest, transform_negated_joins)
     auto s1 = this->insert(Surface{S{1}});
     auto n0 = this->insert(Negated{s1});
     auto j0 = this->insert(Joined{op_and, {s0, n0}});
-    auto simplified = transform_negated_joins(tree_);
+    auto simplified = transform_negated_joins(tree_).tree;
```
sethrj: You should test the mapping result too...

esseivaju:
@sethrj This has been open for a while, I'm trying to close the loop on that. We wanted to compare the runtime/memory usage for constructing `CsgUnit` w/wo simplification. What would be the easiest way to test that? Writing a separate unit test?

src/orange/orangeinp/detail/DeMorganSimplifier.hh
esseivaju: `CsgTree::simplify` only updates a single node, do we really want a vector with `tree.size()` elements for such a case?

sethrj:
Sorry for the long delay @esseivaju , I'm going to need to dedicate a morning (tomorrow?) to try to get back into this since the context switching into geometry is pretty high.

src/orange/OrangeTypes.hh
```diff
@@ -272,6 +272,16 @@ enum class ZOrder : size_type
...
 
+/*!
+ * Optional transformations to apply when building a CsgUnit.
+ */
+enum class UnitSimplification : size_type
+{
+    none = 0,  //!< No simplification
+    csgtree_infix_logic,  //!< CsgTree suitable for infix logic evaluation
+    size_
+};
```
sethrj: This is specifically to construction (rather than construction + runtime), so please move to `orange/orangeinp/CsgTypes.hh`. You should drop the `csgtree_` as well.

src/orange/orangeinp/detail/DeMorganSimplifier.hh
sethrj: I was thinking a different function that simplifies the whole tree—but that's dumb since we have those advanced algorithms in the CsgTreeUtils.

Could you move this class into CsgTreeUtils.hh (maybe rename `SimplifiedCsgTree`?)  since it's not in the detail namespace?

src/orange/orangeinp/detail/CsgUnitBuilder.cc
```diff
@@ -193,6 +194,47 @@ void CsgUnitBuilder::fill_volume(LocalVolumeId v,
...
+ * Apply DeMorgan simplification to use the \c CsgUnit in infix evaluation.
+ * \c NodeId indexing in the \c CsgTree are invalidated after calling this,
+ * \c CsgUnit data is updated to point to the simplified tree \c NodeId but any
+ * previously cached \c NodeId is invalid.
+ */
+void CsgUnitBuilder::simplifiy_joins()
+{
+    auto& tree = unit_->tree;
+    auto simplification = transform_negated_joins(tree);
+    CELER_EXPECT(tree.size() == simplification.new_nodes.size());
```
sethrj: It's not a precondition so:
```suggestion
    CELER_ASSERT(tree.size() == simplification.new_nodes.size());
```

src/orange/orangeinp/detail/CsgUnitBuilder.cc
```diff
@@ -193,6 +194,47 @@ void CsgUnitBuilder::fill_volume(LocalVolumeId v,
...
+    std::vector<std::set<CsgUnit::Metadata>> md;
+    md.resize(simplification.tree.size());
+
+    std::map<NodeId, CsgUnit::Region> regions;
+
+    for (auto node_id : range(tree.size()))
+    {
+        if (auto equivalent_node = simplification.new_nodes[node_id];
+            equivalent_node)
```
sethrj:
```suggestion
        if (auto equivalent_node = simplification.new_nodes[node_id])
```

src/orange/orangeinp/detail/CsgUnitBuilder.cc
```diff
@@ -193,6 +194,47 @@ void CsgUnitBuilder::fill_volume(LocalVolumeId v,
...
+        {
+            CELER_EXPECT(equivalent_node < md.size());
+            md[equivalent_node.unchecked_get()] = unit_->metadata[node_id];
+            regions[equivalent_node] = unit_->regions[NodeId{node_id}];
```
sethrj: Can we `move` them? (Nodes in the CSG tree shouldn't get duplicated, so this should be one-to-one right?)
```suggestion
            md[equivalent_node.unchecked_get()] = std::move(unit_->metadata[node_id]);
            regions[equivalent_node] = std::move(unit_->regions[NodeId{node_id}]);
```

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -65,6 +68,27 @@ class CsgTreeUtilsTest : public ::celeritas::test::Test
         return eval_sense(n);
     }
 
+    DeMorganSimplifierResult
+    validate_simplification(std::string_view expected,
+                            std::vector<NodeId> const& expected_node_map)
+    {
+        auto [simplified, node_map] = transform_negated_joins(tree_);
+        EXPECT_EQ(expected, to_string(simplified));
+        EXPECT_VEC_EQ(expected_node_map, node_map);
```
sethrj: The downside of `EXPECT_EQ` in a function is that the failure messages point here instead of the calling location where you want them...

sethrj (commented):
One last round of changes! Sorry for the long review cycle... the scidac review threw me for a loop

src/orange/orangeinp/detail/CsgUnitBuilder.cc
esseivaju: yep should be fine to move since we reassign `unit_->metadata` and `unit_->regions` later.

test/orange/orangeinp/CsgTreeUtils.test.cc
esseivaju: yep, I've been using the debugger to pin-point which expect was failing but it's easier to have the source location listed in the logs.

sethrj (approved):
Looks great! Thanks Julien.

PR #1416, opened 2024-09-19
Define "static" concrete action for those with unique name/description

sethrj:
This is a small refactor to use string views instead of heap-allocated strings for actions that should only exist once in the stepping loop.

### DIFF
--- a/src/corecel/sys/ActionInterface.hh
+++ b/src/corecel/sys/ActionInterface.hh
@@ -231,7 +231,7 @@ class ConcreteAction : virtual public ActionInterface
 
     // Default destructor
     ~ConcreteAction() noexcept;
-    CELER_DEFAULT_COPY_MOVE(ConcreteAction);
+    CELER_DELETE_COPY_MOVE(ConcreteAction);
 
     //! ID of this action for verification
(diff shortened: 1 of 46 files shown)


src/corecel/sys/ActionInterface.hh
```diff
@@ -210,15 +211,27 @@ class StepActionInterface : public ActionTypeTraits<P, S>,
       using ConcreteAction::ConcreteAction;
   };
  * \endcode
+ *
+ * The \c noexcept declarations improve code generation for the
+ * common use case of multiple inheritance.
+ *
+ * \note Use this class when multiple instances of the class may coexist in the
+ * same stepping loop.
```
amandalund, line 219: Do we have any such actions in our loop?

amandalund (approved):
Looks good!

src/corecel/sys/ActionInterface.hh
sethrj: I thought one of the "calo" or step classes did but apparently not. Since the "step collector" class has different use cases (hit manager and root step writer collect different data), that will eventually create different actions I think.

PR #1417, opened 2024-09-19
Add initial stepping loop to optical offload

sethrj:
This adds a very basic stepping loop directly into the optical offload class, and it adds a skeleton placeholder boundary action to demonstrate the loop being called.


test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -536,8 +538,23 @@ TEST_F(LArSphereOffloadTest, host_generate)
...
+        "Celeritas optical state initialization complete",
+        "Celeritas core state initialization complete",
+        "Boundary action is not implemented",
+        "Boundary action is not implemented",
+        "Exceeded step count of 2: aborting optical transport loop with 0 tracks and 324193 queued",
+    };
+    EXPECT_VEC_EQ(expected_log_messages, scoped_log_.messages());
+    static char const* const expected_log_levels[]
+        = {"status", "status", "error", "error", "error"};
+    EXPECT_VEC_EQ(expected_log_levels, scoped_log_.levels());
```
amandalund, line 557: Looks like the `device_generate` test still needs to be updated too.

amandalund (approved):
Nice, thanks @sethrj! One more step forward :)

PR #1418, opened 2024-09-19
Support aliased nodes during CsgTree simplification

esseivaju:
This is necessary when building `CsgUnit`.

### DIFF
--- a/src/orange/orangeinp/detail/DeMorganSimplifier.cc
+++ b/src/orange/orangeinp/detail/DeMorganSimplifier.cc
@@ -96,18 +96,16 @@ CsgTree DeMorganSimplifier::operator()()
 
 //---------------------------------------------------------------------------//
 /*!
- * Equivalent to \c std::get_if except that for \c Aliased nodes, we type check
- * the target.
+ * Find the real node id to access by dereferencing Aliased nodes.
  */
-template<class T>
-T const* DeMorganSimplifier::node_is(Node const& node) const
+NodeId DeMorganSimplifier::dealias(NodeId node_id) const
 {
-    Node const* target = &node;
-    while (auto* alias = std::get_if<Aliased>(target))
+    NodeId dealiased{node_id};
+    while (auto const* aliased = std::get_if<Aliased>(&tree_[dealiased]))
     {
-        target = &tree_[alias->node];
+        dealiased = aliased->node;
     }
-    return std::get_if<T>(target);
+    return dealiased;
 }
 
 //---------------------------------------------------------------------------//
@@ -119,24 +117,25 @@ void DeMorganSimplifier::find_join_negations()
 {
     for (auto node_id : range(NodeId{tree_.size()}))
     {
-        if (auto const& node = tree_[node_id];
-            auto* negated = this->node_is<Negated>(node))
+        auto const* node = &tree_[this->dealias(node_id)];
+        if (auto* negated = std::get_if<Negated>(node))
         {
-            parents_[{negated->node, node_id}]
-                = parents_[{negated->node, has_parents_index_}] = true;
-            if (this->node_is<Joined>(tree_[negated->node]))
+            parents_[{negated->node, node_id}] = true;
+            parents_[{negated->node, has_parents_index_}] = true;
+            if (std::holds_alternative<Joined>(
+                    tree_[this->dealias(negated->node)]))
             {
                 // This is a negated join node
                 negated_join_nodes_[negated->node.get()] = true;
                 this->add_negation_for_operands(negated->node);
             }
         }
-        else if (auto* joined = this->node_is<Joined>(node))
+        else if (auto* joined = std::get_if<Joined>(node))
         {
             for (auto const& join_operand : joined->nodes)
             {
-                parents_[{join_operand, node_id}]
-                    = parents_[{join_operand, has_parents_index_}] = true;
+                parents_[{join_operand, node_id}] = true;
+                parents_[{join_operand, has_parents_index_}] = true;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -103,34 +119,26 @@ void DeMorganSimplifier::find_join_negations()
...
+            parents_[{negated->node, node_id}]
+                = parents_[{negated->node, has_parents_index_}] = true;
+            if (this->node_is<Joined>(tree_[negated->node]))
+            {
+                // This is a negated join node
+                negated_join_nodes_[negated->node.get()] = true;
+                this->add_negation_for_operands(negated->node);
+            }
+        }
+        else if (auto* joined = this->node_is<Joined>(node))
```
sethrj: Instead of traversing a potential "aliased" chain more than once, what about a function that de-aliases the node `dealias(NodeId) -> NodeId`, and then keeping the original code? (Or replacing with `get_if` if you prefer, which does look a bit cleaner...)

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -103,34 +119,26 @@ void DeMorganSimplifier::find_join_negations()
...
+        {
+            for (auto const& join_operand : joined->nodes)
+            {
+                parents_[{join_operand, node_id}]
+                    = parents_[{join_operand, has_parents_index_}] = true;
```
sethrj: I think this would be clearer as two separate `true` assignments... didn't notice in the original code since it was so indented by `Overload` 🙃

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -193,35 +201,36 @@ CsgTree DeMorganSimplifier::build_simplified_tree()
         // This node needs to be inserted in the simplified tree, but we need
         // to update the node ids of its children
         Node new_node = tree_[node_id];
+        // deref aliased nodes, we don't want to insert them in the new tree
+        while (auto const* aliased = std::get_if<Aliased>(&new_node))
+        {
+            new_node = tree_[aliased->node];
+        }
```
sethrj: This is the `dealias` function 😄

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -193,35 +201,36 @@ CsgTree DeMorganSimplifier::build_simplified_tree()
...
-                        CELER_EXPECT(
-                            node_ids_translation_[op.get()].equivalent_node());
-                        op = node_ids_translation_[op.get()].equivalent_node();
-                    }
-                },
-                // Other nodes don't have children so nothing to do
-                [](auto&&) {},
-            },
-            new_node);
+        CELER_EXPECT(!std::holds_alternative<Aliased>(new_node));
```
sethrj:
```suggestion
        CELER_ASSERT(!std::holds_alternative<Aliased>(new_node));
```

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -277,72 +286,69 @@ CsgTree DeMorganSimplifier::build_simplified_tree()
...
+
+        // Check if the negation is a root or a volume. If so, we must
+        // insert it in the simplified tree
+        if (parents_[{node_id, is_volume_index_}]
+            || !parents_[{node_id, has_parents_index_}])
+        {
+            return true;
+        }
+
+        for (auto p : range(NodeId{2}, NodeId{tree_.size()}))
```
sethrj: Instead of the `NodeId{2}` maybe there should be a static constexpr indicating the first meaningful node?

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -277,72 +286,69 @@ CsgTree DeMorganSimplifier::build_simplified_tree()
...
-                    node_ids_translation_[node_id.get()].opposite_join
-                        = std::move(new_id);
-                }
-                return this->should_insert_join(node_id);
-            },
-            // Other nodes need to be inserted
-            [](auto&&) { return true; },
-        },
-        tree_[node_id]);
+    if (auto const& node = tree_[node_id];
```
sethrj: It's weird that you have the local `node` in this `if` branch but use it in the next chained conditional! I didn't know that was legal, but it's definitely confusing. Just move it outside the `if` since the scope barely changes.

esseivaju: yeah, the init statement essentially desugars to a scope around the if statement

src/orange/orangeinp/detail/DeMorganSimplifier.cc
esseivaju: hmm we could but we have to pass it around to many different functions, I'm not sure if that's cleaner

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -94,6 +94,20 @@ CsgTree DeMorganSimplifier::operator()()
...
+ * Find the real node id to access by dereferencing Aliased nodes.
+ */
+NodeId DeMorganSimplifier::dealias(NodeId node_id) const
+{
+    NodeId dealiased{node_id};
+    while (auto const* aliased = std::get_if<Aliased>(&tree_[dealiased]))
+    {
+        dealiased = aliased->node;
+    }
+    return dealiased;
```
sethrj: Add some assertions:
```suggestion
{
    CELER_EXPECT(node_id < tree->size())
    NodeId dealiased{node_id};
    while (auto const* aliased = std::get_if<Aliased>(&tree_[dealiased]))
    {
        dealiased = aliased->node;
        CELER_ASSERT(dealiased < tree->size())
    }
    return dealiased;
```

sethrj (approved):
A couple more assertions would be welcome, otherwise looks good!

PR #1419, opened 2024-09-21
Import muon pair production sampling table from Geant4

amandalund:
This imports the table used for sampling the electron-positron pair energy from the Geant4 muon pair production model.

### DIFF
--- a/src/celeritas/ext/detail/CelerEmStandardPhysics.cc
+++ b/src/celeritas/ext/detail/CelerEmStandardPhysics.cc
@@ -466,6 +466,13 @@ void CelerEmStandardPhysics::add_e_processes(G4ParticleDefinition* p)
  * | Multiple scattering          | G4WentzelVIModel             |
  *
  * \note Currently all muon processes are disabled by default
+ *
+ * \todo Prior to version 11.1.0, Geant4 used the \c G4BetheBlochModel for muon
+ * ionization between 200 keV and 1 GeV and the \c G4MuBetheBlochModel above 1
+ * GeV. Since version 11.1.0, the \c G4MuBetheBlochModel is used for all
+ * energies above 200 keV.
+ *
+ * \todo Implement energy loss fluctuation models for muon ionization.
  */
 void CelerEmStandardPhysics::add_mu_processes(G4ParticleDefinition* p)
 {
--- a/src/celeritas/io/ImportMuPairProductionTable.hh
+++ b/src/celeritas/io/ImportMuPairProductionTable.hh
@@ -1,0 +1,45 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/io/ImportMuPairProductionTable.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <vector>
+
+#include "ImportPhysicsVector.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sampling table for electron-positron pair production by muons.
+ *
+ * This 3-dimensional table is used to sample the energy transfer to the
+ * electron-positron pair, \f$ \epsilon_p \f$. The outer grid stores the atomic
+ * number using 5 equally spaced points in \f$ \log Z \f$; the x grid tabulates
+ * the ratio \f$ \log \epsilon_p / T \f$, where \f$ T \f$ is the incident muon
+ * energy; the y grid stores the incident muon energy using equal spacing in
+ * \f$ \log T \f$.
+ */
+struct ImportMuPairProductionTable
+{
+    //!@{
+    //! \name Type aliases
+    using ZInt = int;
+    //!@}
+
+    std::vector<ZInt> atomic_number;
+    std::vector<ImportPhysics2DVector> physics_vectors;
+
+    explicit operator bool() const
+    {
+        return !atomic_number.empty()
+               && physics_vectors.size() == atomic_number.size();
+    }
+};
+
+//---------------------------------------------------------------------------//
+}  // namespace celeritas
--- a/src/celeritas/io/ImportPhysicsVector.hh
+++ b/src/celeritas/io/ImportPhysicsVector.hh
@@ -47,6 +47,27 @@ struct ImportPhysicsVector
 };
 
 //---------------------------------------------------------------------------//
+/*!
+ * Store imported 2D physics vector data (see Geant4's G4Physics2DVector.hh).
+ *
+ * This stores a 2D grid of generic data with linear interpolation.
+ */
+struct ImportPhysics2DVector
+{
+    std::vector<double> x;  //!< x grid
+    std::vector<double> y;  //!< y grid
+    std::vector<double> value;  //!< [x][y]
+
+    explicit operator bool() const
+    {
+        return !x.empty() && !y.empty() && value.size() == x.size() * y.size();
+    }
+};
+
+// Equality operator, mainly for debugging
+bool operator==(ImportPhysics2DVector const& a, ImportPhysics2DVector const& b);
+
+//---------------------------------------------------------------------------//
 // FREE FUNCTIONS
 //---------------------------------------------------------------------------//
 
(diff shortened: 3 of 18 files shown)


src/celeritas/io/ImportPhysicsVector.hh
```diff
@@ -46,6 +46,27 @@ struct ImportPhysicsVector
     }
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Store imported 2D physics vector data (see Geant4's G4Physics2DVector.hh).
+ *
+ * This stores a 2D grid of generic data with linear interpolation.
+ */
+struct ImportPhysics2DVector
```
sethrj, line 55: Can you also replace the `ImportSBTable` with this structure? It's got the same data+layout.

src/celeritas/io/ImportMuPairProductionTable.hh
```diff
@@ -0,0 +1,45 @@
...
+
+    std::vector<ZInt> atomic_number;
+    std::vector<ImportPhysics2DVector> physics_vectors;
+
+    explicit operator bool() const
+    {
+        return !atomic_number.empty()
+               && physics_vectors.size() == atomic_number.size();
+    }
+};
```
sethrj, line 42: FWIW the import SB table is kept at a high level as `map<int, table>`, which makes it compatible with `celeritas/io/ImportedElementalMapLoader.hh` and lets us choose to read it from a file *or* from the imported data. I think we should have the same approach for both. Perhaps keep the comment but use:
```suggestion
using ImportMuPairProductionTable = ImportPhysics2DVector;
```
since the SBTable is for a single element? Then 
```c++
    using ImportMuPPMap = std::map<ZInt, ImportMuPairProductionTable >;
```
in the IO data.

amandalund: Sure... it's a bit different from the other data because it can _only_ be imported from the model, not read from a file, and interpolates on a fixed Z grid rather than building the tables for each element in the problem, but we can do that.

sethrj: Can you explain what you mean by interpolating on fixed Z? As in, the model hardcodes a few different Z values that don't depend on the materials in the problem, and linearly interpolates on that? In that case yeah I agree let's keep the layout as-is (but still replace ImportSBTable with ImportPhysics2DVector).

amandalund: Yeah, exactly. From the PRM:
> a constant increment in ln𝑍 is chosen. The shape of the sampling distribution does depend on 𝑍, but very weakly, so that eight points in the range 1 ≤ 𝑍 ≤ 128 are sufficient

(although in the code there are only 5 points: `zdat[5] = {1, 4, 13, 29, 92}`).

sethrj: Might be worth going back to the original source for the table since they might have the full Z... also I wonder if we should be only importing Z values based on the range of materials in the problem (or does it matter...)

amandalund: They're constructed in the model from the differential cross section, which is calculated from a formula rather than tables. Maybe someday we can implement the cross section calculator and build the tables ourselves...

True, we could restrict the table based on the elements in the problem... I can add that if you think it's worth it.

src/celeritas/ext/detail/CelerEmStandardPhysics.cc
```diff
@@ -466,6 +466,13 @@ void CelerEmStandardPhysics::add_e_processes(G4ParticleDefinition* p)
  * | Multiple scattering          | G4WentzelVIModel             |
  *
  * \note Currently all muon processes are disabled by default
+ *
+ * \todo Prior to version 11.1.0, Geant4 used the \c G4BetheBlochModel for muon
+ * ionization between 200 keV and 1 GeV and the \c G4MuBetheBlochModel above 1
+ * GeV. Since version 11.1.0, the \c G4MuBetheBlochModel is used for all
+ * energies above 200 keV.
```
amandalund, line 473: One slight complication is that different Geant4 versions use different models for muon ionization...

src/celeritas/io/ImportMuPairProductionTable.hh
sethrj: Building the cross sections sounds like a better investment of our time, and there's plenty of stuff better than building cross sections 😅

sethrj:
Ummmm....
```
155/265 Test #157: celeritas/ext/GeantImporter:FourSteelSlabs* .....................................***Exception: SegFault  0.39 sec
Celeritas version 0.5.0-dev.277+20f1e85
Note: Google Test filter = FourSteelSlabs*
[==========] Running 17 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 17 tests from FourSteelSlabsEmStandard
[ RUN      ] FourSteelSlabsEmStandard.em_particles

```

src/celeritas/ext/detail/CelerEmStandardPhysics.cc
sethrj: Is the newer model available in both? Or should we just default to the same models available in the selected version of G4?

sethrj (approved):
Great! Nice work.

amandalund:
Yeah I'm not sure what's up with that test, and I don't have Celeritas built with Geant4 11.2 yet...

sethrj:
The good news is it's reproducible on my machine with 11.2.1; the bad news is that `GetElement2DData` returns garbage for Z=12... perhaps the semantics of that call have changed?

amandalund:
Ah excellent, thanks @sethrj! I'll look into it...

sethrj:
Wait wait wait, it's not too late, but we almost missed the opportunity to call that Element Table a `MuPPET`!

amandalund:
Ok it looks like in v11.2+ `G4ElementData` can be [constructed with the maximum number of elements](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/materials/src/G4ElementData.cc#L43) (which should be greater than the largest Z). `G4MuPairProductionModel` [now uses this constructor](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/muons/src/G4MuPairProductionModel.cc#L202),  but calls it with `NZDATPAIR`,  which is the _size_ of the Z array, which is 5. It looks like there should be some kind of [error when initializing the data](https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/materials/src/G4ElementData.cc#L108) though for Z >= 5...

sethrj:
@amandalund Yeah, so the ElementTable is designed to have the index be the element ID, but the MuPPET is fixed at size 5 (an implementation detail) and those are just indices (index of ZDATPAIR) rather than Z number (value in ZDATPAIR). There appears to be no way to get the size (the private data member `maxNumElm`). ~~Also it seems to use the Z number to access the muppet in some places, and uses the index to set it.~~ OK it looks like universally it just *misuses* the table, accessing by Z index rather than Z number. I don't know how this has worked before now ... but I do find the British slang appropriately ironic 😉   https://clip.cafe/lock-stock-two-smoking-barrels-1998/i-dont-wanna-know-you-use/

I think the solution is for the muppet code to hard-code Z = 5? I can submit a PR to add an accessor to the table limits.

amandalund:
Haha lord, ok, using the index to set it rather than Z (and access it... sometimes?) is also new since v11.2. Thanks for looking into that more; I'll update the import for the newer versions and I guess just hardcode the size... 🐸 🐷

sethrj:
Oh, so on older versions the table actually is indexed by element (and just has null for the ones that it's interpolating between?) That makes more sense and is a totally different implementation than the current one. See my update that it *is* actually using just entries 0 through 4; the current version does not expose the `{1, 4, 13, 29, 92}` element numbers in any way. Guess we will have to switch on version number... or reimplement the cross sections ourselves  🤪

amandalund:
Yeah, the older versions used the table in a more intuitive way/how it was designed to be used.

> or reimplement the cross sections ourselves 🤪

I'm tempted at this point ;)

amandalund:
> ~~Also it seems to use the Z number to access the muppet in some places, and uses the index to set it.~~ OK it looks like universally it just misuses the table, accessing by Z index rather than Z number.

It is still erroneously accessing the tables by the Z number in `G4MuPairProductionModel::StoreTables()`, right?

sethrj:
Ah indeed, I got confused because I'd changed some lines but not all. Clearly no one used or tested that method...

sethrj (approved):

PR #1420, opened 2024-09-21
Improve PRINT_EXPECTED formatting for long strings and vectors

sethrj:
This does two things to `repr` (used by PRINT_EXPECTED and other such utilities) to improve clang-formatting of the output:
1. Automatically format long strings (> 70) as [raw string literals](https://en.cppreference.com/w/cpp/language/string_literal) to keep clang-format from breaking them into multiple concatenated strings.
2. Add a trailing comma to containers that have a lot of characters, to force clang-format to indent them rather than to align them to the assignment operator.

### DIFF
--- a/test/corecel/io/Repr.test.cc
+++ b/test/corecel/io/Repr.test.cc
@@ -14,6 +14,7 @@
 
 #include "corecel/cont/Array.hh"
 #include "corecel/cont/Span.hh"
+#include "corecel/io/StringUtils.hh"
 
 #include "celeritas_test.hh"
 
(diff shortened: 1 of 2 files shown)


test/corecel/io/Repr.test.cc
```diff
@@ -73,6 +78,11 @@ TEST(ReprTest, container)
 
     std::string const strings[] = {"a", "", "special\nchars\t"};
     EXPECT_EQ("{\"a\", \"\", \"special\\nchars\\t\"}", repr_to_string(strings));
+
+    auto long_vec_str
+        = repr_to_string(std::vector<std::string>(10, std::string(75, 'x')));
+    cout << long_vec_str;
```
amandalund:
```suggestion
```

test/corecel/io/Repr.test.cc
```diff
@@ -73,6 +78,11 @@ TEST(ReprTest, container)
...
+    EXPECT_TRUE(ends_with(long_vec_str, ")\",}")) << long_vec_str;
```
amandalund: Even though it shows up with `repr_to_string`, for some reason I'm not seeing the trailing comma with `PRINT_EXPECTED`.

sethrj: I think this is because PRINT_EXPECTED writes directly to stdout, and `std::cout` must not have `tellp` (i.e. the console outputs don't keep track of how many characters they've written). Let me think about a way around this...

amandalund (approved):
Sweet!

PR #1421, opened 2024-09-21
Report and annotate test failures in CI

sethrj:
This uploads detailed unit test information in JUnit format and adds two PR postprocessors to (1) comment on the PR showing the number of test passed/skipped/failed and (2) annotate google tests in the PR that are now failing. It splits up tests into "app" (run directly through CMake and exported using its `JUnit` generator) and non-"app" (assumed to be googletest and exported via its generator using the `GTEST_OUTPUT` environment variable).

It's a little messy because, due to our breadth of CI testing, multiple instances of the same test case are uploaded (so the same failure can show up multiple times, or it's not clear from the annotation which exact instance is failing). The annotations for google tests also show the line number of the "test case" rather than the actual test assertion. Usually it's close though.

The comments/failures won't show up on this PR because the new `pull_request_completed` workflow has to first be pushed to develop. See https://github.com/sethrj/celeritas/pull/2 for an example of the output.

Note the weirdness of the repeated
```yaml
            always()
            && (steps.unittest.outcome == 'success'
```
syntax is due to [GHA expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-of-failure-with-conditions): without the `always()`, failing tests will not be uploaded.

Test output and failure messages are demonstrated on https://github.com/sethrj/celeritas/pull/3 .


.github/workflows/build-docker.yml
```diff
@@ -115,5 +118,22 @@ jobs:
...
+        # upgrade the image
+        if: >-
+          ${{
+            false &&
+            always()
+            && (steps.test.outcome == 'success'
+                || steps.test.outcome == 'failure')
+            && (env.CMAKE_PRESET != 'ndebug-vecgeom')
+          }}
+        uses: actions/upload-artifact@v3
```
pcanal, line 131: Why `v3` here and `v4` later?

.github/workflows/build-docker.yml
```diff
@@ -115,5 +118,22 @@ jobs:
             export LDFLAGS=-Wl,--no-as-needed ${LDFLAGS}
           fi
           ./scripts/ci/test-examples.sh
+      - name: Upload test results
+        # Note: upload-v3 doesn't seem to pick up the paths: delete the "false" line below once we
```
pcanal, line 122: I assume the `always()` should also go?

.github/workflows/build-docker.yml
sethrj: The docker image is stuck on an old version of centos/ubuntu that doesn't support the newer version of node.js ... we're overdue for regenerating these images but I just haven't made it a priority since I've been able to patch them along.

.github/workflows/build-docker.yml
sethrj: No, `always` should be there after this is working, in order to upload if the tests failed.

.github/workflows/build-docker.yml
```diff
@@ -115,5 +118,22 @@ jobs:
             export LDFLAGS=-Wl,--no-as-needed ${LDFLAGS}
           fi
           ./scripts/ci/test-examples.sh
+      - name: Upload test results
+        # Note: upload-v3 doesn't seem to pick up the paths: delete the "false" line below once we 
+        # upgrade the image
```
pcanal:
```suggestion
        # upgrade the centos/ubuntu docker images
```

.github/workflows/build-docker.yml
```diff
@@ -115,5 +118,22 @@ jobs:
...
+        if: >-
+          ${{
+            false &&
+            always()
+            && (steps.test.outcome == 'success'
+                || steps.test.outcome == 'failure')
+            && (env.CMAKE_PRESET != 'ndebug-vecgeom')
```
pcanal: Why is this build excluded from the uploads?

sethrj: Hmm I think I was playing around with why the test was failing and said hey, maybe it's because it didn't have any google tests? But I think the real issue is the old uploader version. I'll delete this line.

pcanal (approved):
Looks good.  Thanks.

Issue #1412, opened 2024-09-18
Add support for surface-based VecGeom


sethrj:
VecGeom is implementing a new surface-based navigation mechanism designed to be more efficient on GPU:  see [10.2172/2204991](https://doi.org/10.2172/2204991) and associated AdePT integration https://github.com/apt-sim/AdePT/issues/298 . This issue will track the progress of that integration.

@mrguilima please document the issues you're seeing as part of this update (upload log files of tests, etc.) Thanks!

mrguilima:
The first three failing tests need support for Orb (FourLevels*) and many other shapes (solids.gdml).
 54 - geocel/vg/Vecgeom:FourLevelsGeantTest.* (SEGFAULT)  <<== needs Orb
 55 - geocel/vg/Vecgeom:SolidsGeantTest.* (SEGFAULT)     <<== solids.gdml, needs support for ~all shapes
157 - celeritas/ext/GeantVolumeMapper (Failed)  <<== also uses solids.gdml, needs support for ~all shapes

sethrj:
> Thanks! Did you see Severin's comment about recent changes negating the need for the stack size? How recent is the build of vecgeom you're using?

Yes, as I said yesterday, the stack size issue appears only in Debug mode, not in Release mode.

> For the missing "orb" we should be able to trivially build a sphere instead, and for the unsupported volumes (do you have a list?) we can likewise modify the geometry converter.

I can apply this change.

mrguilima:
The next group of tests are failing because they load more than one geometry, but the `reset_geometry()` function is not working properly for the Surface infrastructure.  Using the --gtest_filter=Cmse* etc.. works for most of the geometries though (exception: Cmse), so I assume that the BVHNavigator from VecGeom is working fine at this point.

162 - celeritas/field/FieldPropagator (SEGFAULT)
165 - celeritas/geo/Geometry (SEGFAULT)

All these work well:
    test/celeritas/field_FieldPropagator --gtest_filter=TwoBox*
    test/celeritas/field_FieldPropagator --gtest_filter=LayersTest*
    test/celeritas/field_FieldPropagator --gtest_filter=SimpleCmsTest*
    test/celeritas/field_FieldPropagator --gtest_filter=CmseTest*

    test/celeritas/geo_Geometry --gtest_filter=SimpleCmsTest*
    test/celeritas/geo_Geometry --gtest_filter=ThreeSpheres*

and this one fails:
[...]

PR #1422, opened 2024-09-23
Support surface-based VecGeom 2.x navigator

mrguilima:
Adapting Celeritas to use the latest surface-based model from VecGeom 2.x  (see #1412).

The main change is to use VecGeom's `vgbrep::protonav::BVHSurfNavigator`. Since there are interface differences between this one and the volume-based `BVHNavigator`, the code-guards `VECGEOM_USE_SURF`, in analogy to AdePT.
Alternatively, `VECGEOM_VERSION >= 0x020000` could be used instead, as the VecGeom 1.2.x series should always be used for volume-based navigation anyway.

There are also distinct interfaces for ABBoxManager (template vs. non-template).

Note: in draft mode, as some tests currently fail.  Under investigation.


src/geocel/vg/VecgeomParams.cc
```diff
@@ -11,6 +11,7 @@
 #include <vector>
 #include <VecGeom/base/Config.h>
 #include <VecGeom/base/Cuda.h>
+#include <VecGeom/base/Version.h>
```
sethrj, line 17: I've verified that this header (and the `VECGEOM_VERSION` macro therein) are present in v1.2.4, which is the minimum vecgeom required by Celeritas.

src/geocel/vg/detail/SurfNavigator.hh
```diff
@@ -79,9 +88,9 @@ class SurfNavigator
             return step_limit;
         }
 
-        int exit_surf = 0;
-        auto step = vgbrep::protonav::ComputeStepAndHit(
-            globalpoint, globaldir, in_state, out_state, exit_surf, step_limit);
+	long hitsurf_id = 0;
+        auto step = vgbrep::protonav::BVHSurfNavigator<Precision>::ComputeStepAndNextSurface(
+            globalpoint, globaldir, in_state, out_state, hitsurf_id, step_limit);
```
sethrj: Something here is misbehaving compared to the volume version:
```
54: /Users/seth/Code/celeritas-temp/test/geocel/vg/Vecgeom.test.cc:164: Failure
54: Value of: next.boundary
54:   Actual: true
54: Expected: false
```
for the calls
```
    auto next = geo.find_next_step(from_cm(100));
    EXPECT_SOFT_EQ(100, to_cm(next.distance));
    EXPECT_FALSE(next.boundary);
```

So it's claiming to have a boundary for the next surface, even though it doesn't hit it within the step limit.

sethrj:
Without GPU on my laptop (clang 15 debug), on VecGeom ca34e35ea5fc822fcffd3c7c5d117e6bc671ce32 (reldeb), the following tests fail:
```
	 54 - geocel/vg/Vecgeom:SimpleCmsTest.* (Failed)
	 55 - geocel/vg/Vecgeom:FourLevelsGeantTest.* (Failed)
	 56 - geocel/vg/Vecgeom:SolidsGeantTest.* (SEGFAULT)
	 57 - geocel/vg/Vecgeom:ZnenvGeantTest.* (Failed)
	166 - celeritas/field/FieldPropagator (SEGFAULT)
	167 - celeritas/field/LinearPropagator (Failed)
	169 - celeritas/geo/Geometry (SEGFAULT)
	170 - celeritas/geo/GeoMaterial (Failed)
	171 - celeritas/global/AlongStep:-Em3*:SimpleCms*:LeadBox* (SEGFAULT)
	172 - celeritas/global/AlongStep:Em3AlongStepTest.nofluct_nomsc (Failed)
	173 - celeritas/global/AlongStep:Em3AlongStepTest.msc_nofluct (Failed)
	175 - celeritas/global/AlongStep:Em3AlongStepTest.fluct_nomsc (Failed)
	176 - celeritas/global/AlongStep:SimpleCmsAlongStepTest.msc_field (Subprocess killed)
	178 - celeritas/global/AlongStep:SimpleCmsRZFieldAlongStepTest.msc_rzfield (Failed)
	180 - celeritas/global/KernelContextException (Failed)
	181 - celeritas/global/Stepper (Failed)
	182 - celeritas/global/StepperGeant:-TestEm*:LeadBox* (Failed)
	183 - celeritas/global/StepperGeant:TestEm3Compton.* (Failed)
	184 - celeritas/global/StepperGeant:TestEm3NoMsc.* (Failed)
	185 - celeritas/global/StepperGeant:TestEm3Msc.* (Failed)
	186 - celeritas/global/StepperGeant:TestEm3MscNofluct.* (Failed)
	187 - celeritas/global/StepperGeant:TestEm15FieldMsc.* (Failed)
	203 - celeritas/optical/OpticalCollector (Failed)
	238 - celeritas/track/TrackSort (Failed)
	241 - celeritas/user/Diagnostic:-TestEm3* (Failed)
	242 - celeritas/user/Diagnostic:TestEm3* (Failed)
	243 - celeritas/user/StepCollector (SEGFAULT)
```

The simplest failure is just that it seems to report aa boundary when not actually hitting a boundary:
```
[ RUN      ] SimpleCmsTest.track
/Users/seth/Code/celeritas-temp/test/geocel/vg/Vecgeom.test.cc:164: Failure
Value of: next.boundary
  Actual: true
Expected: false

/Users/seth/Code/celeritas-temp/test/geocel/vg/Vecgeom.test.cc:176: Failure
  "si_tracker"
  this->volume_name(geo)
    Which is: "world"

[  FAILED  ] SimpleCmsTest.track (0 ms)
```

Full failure log:
[LastTest.log](https://github.com/user-attachments/files/17130055/LastTest.log)

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -282,18 +365,19 @@ class FourLevelsTest : public VecgeomVgdmlTestBase
         return this->load_vgdml("four-levels.gdml");
     }
 
-    SpanStringView expected_log_levels() const final
-    {
-        if (vecgeom_version >= Version{2})
+    /*  SpanStringView expected_log_levels() const final
```
sethrj: I suspect you mean this to be temporary, but it's a good habit to start using `#if 0` or (even better since you mean it *just* for the surface case)  use a conditional based on the surface settings macro.

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -984,8 +1070,8 @@ TEST_F(CmseTest, trace)
             180, 910, 24000, 6000};
         EXPECT_VEC_SOFT_EQ(expected_distances, result.distances);
         static real_type const expected_hw_safety[] = {100, 2.1499999999997,
-            0.52499999999986, 13.023518051922, 6.95, 6.95, 13.023518051922,
-            0.52499999999986, 2.15, 100, 5, 8, 100, 100, 100};
+            10.3027302206745, 13.023518051922, 6.95, 6.95, 13.023518051922,
+            10.3027302206745, 2.15, 100, 5, 8, 100, 100, 100};
```
sethrj: It's great that you've recorded the updated safety values: these are now consistent with the ones returned by Geant4. I suppose we'll have to "edit" these when using the volume-based vecgeom.

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -1034,7 +1120,8 @@ TEST_F(CmseTest, trace)
         static real_type const expected_distances[] = {12.495, 287.505, 530,
             920};
         EXPECT_VEC_SOFT_EQ(expected_distances, result.distances);
-        static real_type const expected_hw_safety[] = {1, 1, 242, 460};
+        static real_type const expected_hw_safety[] = {6.2475, 47.9499999999998,
+            242, 460};
```
sethrj: Guess we'll have to change the label on this one 😄 since it now agrees between the two

sethrj:
@mrguilima I think we can merge this after #1652 and https://github.com/spack/spack/pull/49313 are merged! I'll pull those in and update the CI when those are in.

The one outstanding issue is the sensitive behavior of TestEM3 ... but we can disable that for the moment.

sethrj (approved):
Hooray! Surface crossing issues (when a direction change happens) to be revisited in the next follow-on.

PR #1423, opened 2024-09-24
Refactor action launcher components into kernel launcher

sethrj:
This splits up the action launcher and traits into a generic "Kernel launcher", which contains all the complexity of thread ranges, block sizes, etc.; and "action launcher" which only applies to core data and extracts the state size/stream. The kernel launcher can be reused for things that *aren't* actions (e.g., raytrace and step collectors and **optical physics kernels**), and the kernel traits can now be safely reused without violating our "detail" convention.


esseivaju (approved):
No comment, everything looks good to me!

sethrj:
Thanks for the quick review @esseivaju !

PR #1425, opened 2024-09-25
Add Bethe-Bloch model and refactor muon ionization

amandalund:
This adds the Bethe-Bloch ionization model. Older versions of Geant4 use Bethe-Bloch at intermediate energies, so this will let us default to using whatever models are in the selected Geant4 version (see https://github.com/celeritas-project/celeritas/pull/1419#discussion_r1769651975).

The Bethe-Bloch interactor is more or less identical to the other three muon ionization models, so most of the changes here are just refactoring the muon ionization to reduce some of the duplicate code and add more flexibility:
- Use a single interactor and executor templated on the model's energy distribution
- Move duplicated methods to helper functions
- Construct the ionization models with the applicability to allow for different incident particle types (e.g. hadrons) and energy bounds


sethrj:
Nice refactor! Still working on it, but in the meantime could you update the docs? https://github.com/celeritas-project/celeritas/actions/runs/11035158855/job/30650454651?pr=1425#step:7:26

src/celeritas/em/model/detail/MuHadIonizationBuilder.hh
```diff
@@ -0,0 +1,98 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using Energy = units::MevEnergy;
+    using SetApplicability = std::set<Applicability>;
+    //!@}
+
+  public:
+    // Construct with shared particle data and model label
+    explicit inline MuHadIonizationBuilder(ParticleParams const& particles,
```
sethrj: Explicit isn't needed here

src/celeritas/em/distribution/BetheBlochEnergyDistribution.hh
```diff
@@ -0,0 +1,127 @@
...
+    InverseSquareDistribution sample_energy(value_as<Energy>(min_energy_),
+                                            value_as<Energy>(max_energy_));
+    real_type energy;
+    do
+    {
+        // Sample 1/E^2 from Emin to Emax
+        energy = sample_energy(rng);
+        /*!
+         * \todo Adjust rejection functions if particle has positive spin
+         */
```
sethrj, line 119: So the MuBB model is this the same as the BetheBloch but with rejection functions? The MuBBEnergyDistribution could use a touch more documentation describing the difference between these... it might also be fun to rig up a little test that lets us plot the exiting energy distributions for a couple of different incident particle energies.

src/celeritas/em/distribution/MuBBEnergyDistribution.hh
```diff
@@ -100,22 +105,17 @@ class MuBBEnergyDistribution
...
+MuBBEnergyDistribution::MuBBEnergyDistribution(ParticleTrackView const& particle,
                                                Energy electron_cutoff,
-                                               Energy max_secondary_energy)
-    : inc_energy_(value_as<Energy>(inc_energy))
-    , inc_mass_(value_as<Mass>(inc_mass))
-    , total_energy_(inc_energy_ + inc_mass_)
-    , beta_sq_(beta_sq)
+                                               Mass electron_mass)
+    : inc_mass_(value_as<Mass>(particle.mass()))
+    , total_energy_(value_as<Energy>(particle.energy()) + inc_mass_)
```
sethrj, line 133: I wonder if `total_energy` should be a `ParticleTrackView` accessor too?

src/celeritas/em/interactor/detail/PhysicsConstants.hh
```diff
@@ -64,28 +64,22 @@ CELER_CONSTEXPR_FUNCTION units::MevEnergy seltzer_berger_upper_limit()
     return units::MevEnergy{1e3};  //! 1 GeV
 }
 
-//! Energy threshold between muon ionization models. TODO: make configurable
-CELER_CONSTEXPR_FUNCTION units::MevEnergy mu_bethe_bloch_lower_limit()
+//! Maximum energy for the Bragg and ICRU73QO models
+CELER_CONSTEXPR_FUNCTION units::MevEnergy bragg_icru73qo_upper_limit()
```
sethrj: With your refactor, is this *only* now used in the construction of the models by the process? If so that's fantastic: it always bothered me to have a constexpr arbitrary value like this used in several places. Moving these to the process options would allow the values to be configurable as well.

sethrj (approved):
Great refactor! For some future performance paper we could reinvestigate the cost of multiple kernels versus a single "combined" one (like the combined SB+rel brem) since most of the code in the resulting kernels looks to be the same.

src/celeritas/em/distribution/BetheBlochEnergyDistribution.hh
amandalund: Right, for muons it's just different rejection functions (for hadrons after sampling the energy the Bethe-Bloch model will also sample whether or not to produce a high energy delta ray). A distribution test is a good idea.

src/celeritas/em/interactor/detail/PhysicsConstants.hh
amandalund: Yep, only in the process... I'll move them there.

src/celeritas/em/distribution/MuBBEnergyDistribution.hh
amandalund: True, we use that in several places. I can follow up with a little PR for that.

sethrj: That would be great! 🙇‍♂️

amandalund:
Thanks @sethrj! I agree about investigating the performance with a single combined vs. multiple interactor kernels (I don't think we ever did with brems, did we?). It's nice to have this modularity, but not if it's going to come at a cost.

Could you also help me with the CI build errors?

sethrj:
I think at the very beginning of the project we saw that the combined kernel didn't save us much: but it's something of a moot point for our field-dominated problems. It'll be worth another look for our muon physics problems. Taking a look at the CUDA builds now...

sethrj:
Oh wait, I think this is because of #1423 (where I reordered the kernel arguments) and the PR is running on the "merged" version. Shame on me for not keeping a "deprecated" version, sorry about that! I'll push a fix in just a minute.

amandalund:
Omg yes that's it, thanks! Was staring at my unmerged branch for 10 mins with no idea what I was missing 🤦

sethrj:
Sorry again for that! Teaches me a lesson that the "deprecated" annotations aren't completely fatuous.

PR #1426, opened 2024-09-25
Add slot particle diagnostic

sethrj:
This adds a class that extends @amandalund 's neat ASCII art in #1322 to exporting particle IDs at every step for postprocessing.
- [x] Add slot diagnostic, JSON output, and test
- [x] Add python postprocessing script
- [x] Add interface to front end

TestEM3 demo (from Amanda's unit test):
em3slot-long-demo-plot

Celer-g4 offloading unit test (without partitioning) showing concatenated events:
slot-diag-cpu-plot


amandalund:
Super cool! I was wishing we had something like this to see how the separation evolves over time for some of the benchmark problems. Will take a look soon.

sethrj:
Yeah! After this gets merged I was going to run it (but I'm happy to let you!) to make pretty plots for the upcoming presentations.

sethrj:
Just as a side note, I ran the same problem on GPU and got identical behavior. One of these days we should do a more thorough GPU/CPU equivalence comparison.
em3slot-long-demo-plot

amandalund (approved):
Looks great, cool addition @sethrj!

app/celer-g4/GlobalSetup.cc
```diff
@@ -167,10 +167,7 @@ void GlobalSetup::ReadInput(std::string const& filename)
         options_->cuda_heap_size = input_.cuda_heap_size;
         options_->action_times = input_.action_times;
         options_->default_stream = input_.default_stream;
-
-        // TODO: add track_order I/O
-
-        options_->slot_diagnostic_prefix = input_.slot_diagnostic_prefix;
```
amandalund: Did you mean to delete this or is this a merge error?

sethrj: Oops, merge error. Thanks!

PR #1427, opened 2024-09-26
Calculate safety distances using OBZ when simple safety is not available

elliottbiondo:
The MR "hooks up" the OBZ safety distance calculation capability by using it within `SimpleUnitTracker` to calculate safety distances when simple safety is not available.

### DIFF
--- a/src/orange/univ/SimpleUnitTracker.hh
+++ b/src/orange/univ/SimpleUnitTracker.hh
@@ -11,6 +11,7 @@
 #include "corecel/math/Algorithms.hh"
 #include "orange/OrangeData.hh"
 #include "orange/detail/BIHTraverser.hh"
+#include "orange/detail/OrientedBoundingZone.hh"
 #include "orange/surf/LocalSurfaceVisitor.hh"
 
 #include "detail/InfixEvaluator.hh"
@@ -100,6 +101,7 @@ class SimpleUnitTracker
 
     ParamsRef const& params_;
     SimpleUnitRecord const& unit_record_;
+    detail::OrientedBoundingZone::StoragePointers obz_storage_;
 
     //// METHODS ////
 
@@ -143,7 +145,9 @@ class SimpleUnitTracker
  */
 CELER_FUNCTION
 SimpleUnitTracker::SimpleUnitTracker(ParamsRef const& params, SimpleUnitId suid)
-    : params_(params), unit_record_(params.simple_units[suid])
+    : params_(params)
+    , unit_record_(params.simple_units[suid])
+    , obz_storage_({&params_.transforms, &params_.reals})
 {
     CELER_EXPECT(params_);
 }
@@ -301,8 +305,12 @@ CELER_FUNCTION real_type SimpleUnitTracker::safety(Real3 const& pos,
     if (!vol.simple_safety())
     {
         // Has a tricky surface: we can't use the simple algorithm to calculate
-        // the safety, so return a conservative estimate.
-        return 0;
+        // the safety, use the oriented bounding zone
+        auto const& obz_id
+            = params_.volume_records[unit_record_.volumes[volid]].obz_id;
+        detail::OrientedBoundingZone obz(params_.obz_records[obz_id],
+                                         obz_storage_);
+        return obz.calc_safety_inside(pos);
     }
 
     // Calculate minimim distance to all local faces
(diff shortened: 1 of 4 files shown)


elliottbiondo:
@sethrj this is ready for your review.

I believe everything in this MR and my previous MR are correct, but I am suspicious of the OBZ inner bbox that is being created upon construction. In the GDML file I added here, I have a cone of r=13 and z=50 centered at the origin. As you can see, the outer bbox is correct. However, I don't under why the inner bbox spans z = (-12.5, 0).  No matter what is chosen for the inner bbox x and y widths, the inner bbox should always start at z=-25. Obviously the way we are calculating bboxs by truncating all space with planes does not always yield the analytic solution, but I don't see where the -12.5 could be coming from.

OBZ for the cone volume
Inner: Half widths: (4.596195, 4.596195, 12.500001), offset: (0.000000, 0.000000, -12.500000)
Outer: half widths: (13.000001, 13.000001, 25.000002),  offset: (0.000000, 0.000000, 0.000000)
Overall tranform (0., 0., 0.)

elliottbiondo:
CI failures from `src/celeritas/optical/action/BoundaryAction.cu` appear to be unrelated

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -100,6 +101,7 @@ class SimpleUnitTracker
 
     ParamsRef const& params_;
     SimpleUnitRecord const& unit_record_;
+    detail::OrientedBoundingZone::StoragePointers obz_storage_;
```
sethrj, line 104: This should probably be created on the fly since we keep `params_` around

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -301,8 +305,12 @@ CELER_FUNCTION real_type SimpleUnitTracker::safety(Real3 const& pos,
...
     {
         // Has a tricky surface: we can't use the simple algorithm to calculate
-        // the safety, so return a conservative estimate.
-        return 0;
+        // the safety, use the oriented bounding zone
+        auto const& obz_id
+            = params_.volume_records[unit_record_.volumes[volid]].obz_id;
+        detail::OrientedBoundingZone obz(params_.obz_records[obz_id],
+                                         obz_storage_);
+        return obz.calc_safety_inside(pos);
```
sethrj, line 313: Unfortunately it's not going to be this simple: we need the minimum over both "inside current volume" and "outside all directly enclosed volumes". I'm not sure if we have that metadata yet, in fact...

test/geocel/data/cone.gdml
```diff
@@ -0,0 +1,89 @@
...
+      <position ref="origin"/>
+    </physvol>
+    <physvol>
+      <volumeref ref="vol2"/>
+      <position ref="origin"/>
+    </physvol>
+    <physvol>
+      <volumeref ref="vol3"/>
+      <position ref="origin"/>
+    </physvol>
```
sethrj, line 79: I think these three enclosed volumes are overlapping. It might be more natural for you to make a geometry with `orange/orangeinp/UnitProto.test.cc` using the `InputBuilderTest` harness.

test/orange/univ/SimpleUnitTracker.test.cc
```diff
@@ -967,6 +973,22 @@ TEST_F(FiveVolumesTest, TEST_IF_CELERITAS_DOUBLE(heuristic_init))
     }
 }
 
+TEST_F(ConeTest, safety)
+{
```
sethrj, line 977: Oh crap, I forgot that the Geant input gets scaled into the native unit system (unlike the .org.json files which are precalculated). This test is with the CLHEP unit system which scales everything by a factor of 10. I think it would be better to build a geometry  with the InputBuilder test...  sorry 😞

sethrj (changes requested):

sethrj:
There's also a build error due to `StoragePointers`: it should have `CELER_FUNCTION explicit operator bool()`, not just operator bool.

elliottbiondo:
Closing this MR. Discussing with @sethrj, this approach will not work in the short term. The OBZ inner bounding box does not exclude inner volumes (this is a good thing because if it did, it would be too small to be useful in many cases). A consequence of this is that when checking inner safety distances, we would have to check the outer safety distances of internal volumes. Volumes don't currently have metadata denoting what internal volumes are contained. While this could be remedied when converting from G4 (in which the concept of internal volumes is a baked into the geometry format), it is not clear how it could be done during construction in the general CSG case. As we implement more specialized features we may need to split things up so that ORANGE differentiates between HEP and nuclear (i.e., general CSG) geometries, which would be a considerable undertaking.

Issue #261, opened 2021-07-19
Add MPI support to parallelize over events


sethrj:
To run large jobs on leadership-class machines, we'll need to parallelize across nodes by distributing events using MPI. A first attempt could just equally partition among tasks, but we should also consider investigating whether any MPI framework has a queue-like model that can dispatch events to waiting nodes (or "rebalance" events in case some processors get unlucky).

sethrj:
We're not going to have this in time to do any big runs on Summit for the SciDAC proposal, so let's defer to Q3.

PR #1428, opened 2024-09-26
Add global MPI communicator and refactor some internals

sethrj:
This is in preparation for an "end run gather" and other possible synchronization points for multiprocess tallying using MPI in addition to OpenMP/CUDA. (See #261 )  I've adjusted the code to hide the dependence on MPI from downstream codes, and added a shared "comm_world()" as well as an MPI communicator that stays with the shared params.


sethrj:
@tmdelellis This is in preparation for some of the work to be done with @sphamil.

tmdelellis (approved):
Looks fine to me

PR #1424, opened 2024-09-24
Add optical core track view and launcher

sethrj:
This defines optical kernel launchers, track view, and executor. It uses these to implement a boundary crossing kernel for optical tracks.

PR #1429, opened 2024-09-27
Hotfix: untested CUDA code in #1424

sethrj:
When I merged #1424 I told it to skip the full checks because the only change from the previous "successful" run was a typo. This was dumb because the last test ran when it was draft, which excluded the docker builds. I tested this build on wildstyle and will merge when all the tests pass...


amandalund (approved):
Oops, I should have checked that too.

Issue #1334, opened 2024-07-23
Improve ORANGE memory usage and performance


sethrj:
On frontier, something catastrophic has happened to the performance of ORANGE in a couple of problems. Here's the relative throughput of 0.5.0-dev.202+141cd4928 compared to v0.4.4:
![throughput](https://github.com/user-attachments/assets/0d4d8ade-8c6e-4b17-b77e-8e023f72a67c)

This is not due to a change in the number of steps (so the amount of work is the same):
![work](https://github.com/user-attachments/assets/18e9dde2-84b1-412f-a192-ba4292449bd8)

As a reminder, the problem complexity and abbreviations:
![problem-complexity](https://github.com/user-attachments/assets/1d68c0ec-4be1-4e52-a6d6-be997e6b2a3e)
![names](https://github.com/user-attachments/assets/3e302a50-14c1-4b58-88cf-c11119bb310c)
and 'M̃' means MSC is disabled.

sethrj:
@esseivaju 's runs on perlmutter confirm this is an ORANGE issue:
![throughput](https://github.com/user-attachments/assets/7186257c-dd9b-4d33-99cf-fac7b4846667)

sethrj:
OK, the difference was from changing from the pregenerated `.org.json` to the built-in GDML generator... not clear why

sethrj:
@esseivaju has been doing some profiling of the along-step kernels. Long story short: the T- cases ("EM flat") are doing the same number of operations but the memory access are much more expensive.

It seems the main difference is the empty "calorimeter" volume, which is a box with all the inner boxes subtracted out: the manually constructed version doesn't have this. Even though the volume is never actually entered, it results in a logic expression for the volume that's huge, leading to a major change in "max faces" and "max intersections" which are used by the ORANGE state data:
```diff
 "orange": {
 "scalars": {
 "max_depth": 1,
-"max_faces": 12,
-"max_intersections": 12,
-"max_logic_depth": 3,
+"max_faces": 111,
+"max_intersections": 111,
+"max_logic_depth": 2,
 "tol": {
-"abs": 1e-08,
-"rel": 1e-08
+"abs": 1.5e-09,
+"rel": 1.5e-08
 }
 },
 "sizes": {
 "bih": {
 "bboxes": 102,
-"inner_nodes": 0,
-"leaf_nodes": 1,
+"inner_nodes": 99,
+"leaf_nodes": 100,
 "local_volume_ids": 102
 },
[...]

PR #1430, opened 2024-09-27
Add a hack to override ORANGE max intersections/faces

sethrj:
See https://github.com/celeritas-project/celeritas/issues/1334 . This should allow us to restore the performance of TestEM3 temporarily, although we should disclose that this hack is being used when we present our test problems.

This will only help if *all the visited volumes* have fewer faces/intercepts than the given value: otherwise it will crash.

I'm going to test this before merging.


src/orange/detail/UnitInserter.cc
```diff
@@ -174,6 +180,58 @@ BoundingBoxBumper<fast_real_type, real_type> make_bumper(Tolerance<> const& tol)
...
+        {
+            return ForceMax{};
+        }
+        CELER_LOG(warning)
+            << "Using a temporary, unsupported, and dangerous hack to "
+               "override maximum faces and intersections in ORANGE: "
+            << mfi_hack_envname << "='" << mfi << "'";
+
+        ForceMax result;
+        static std::regex const mfi_regex{R"re(^(\d+),(\d+)$)re"};
```
elliottbiondo, line 213: In the context of this being a presumably temporary hack I don't know if this matters, but it the regex validation is complex enough to get its own function in anonymous namespace.

elliottbiondo (approved):

src/orange/detail/UnitInserter.cc
sethrj: Yeah this option shouldn't live long (the "cheat" itself isn't tested anywhere except by me manually for our regression problems). The regex `^(\d+),(\d+)$` isn't that complicated (whole text, capture two nonnegative integers separated by a comma) but perhaps it's harder to read because of the literal `R"re()re"` string.

sethrj:
OK, using the "flat" file changes the results because of MSC:
rel-work

But the throughput for the "identical geometry in memory" goes up:
rel-throughput

PR #1431, opened 2024-09-27
Fix intermittent CI failures due to corrupt XML

sethrj:
Currently the CI fails occasionally due to multiple tests (the ones with "filter") writing to the same XML filename chosen by googletest's default.

Instead of using that filename, a new CMake option passed to the test generator tells google test the exact file name for each test including filter and processor count. Test outputs now look like:
```
celeritas_global_StepperGeant_TestEm_LeadBox_.xml
celeritas_global_StepperGeant_TestEm3Compton_.xml
celeritas_global_StepperGeant_TestEm3Msc_.xml
celeritas_global_StepperGeant_TestEm3MscNofluct_.xml
celeritas_global_StepperGeant_TestEm3NoMsc_.xml
celeritas_global_StepperGeant_TestEm15FieldMsc_.xml
```
instead of:
```
global_StepperGeant_1.xml
global_StepperGeant.xml
```
where multiple files may write to `_1.xml` at the same time due to the race condition.


sethrj:
@pcanal This is ready. I've verified both the updated test names and the reduced artifact download at https://github.com/sethrj/celeritas/actions/runs/11072978145/job/30768477521
.

cmake/CeleritasAddTest.cmake
```diff
@@ -407,10 +407,16 @@ function(celeritas_add_test SOURCE_FILE)
       # Launch with MPI directly
       _celeritasaddtest_mpi_cmd(_test_cmd "${_np}" "${_EXE_NAME}")
 
-      set(_test_args "${_EXE_ARGS}")
+      set(_test_args)
```
sethrj, line 411: There is no `_EXE_ARGS` defined; that was just a leftover artifact from some previous iteration

cmake/CeleritasAddTest.cmake
```diff
@@ -407,10 +407,16 @@ function(celeritas_add_test SOURCE_FILE)
...
       if(_filter)
         list(APPEND _test_args "--gtest_filter=${_filter}")
       endif()
+      if(CELERITAS_TEST_XML)
+        string(REGEX REPLACE "[^a-zA-Z0-9_]+" "_" _xml_name "${_TEST_NAME}")
```
pcanal: I suspect we do not have the case but technically this can lead to duplicate name unless the same characters are always use in the same place.  I.e. the save way is to replace each characters by its own mangling (``*` -> `_st_` etc.)

pcanal: but the potential list is long :(
```
      switch(c) {
         case '+': repl = "pL"; break;
         case '-': repl = "mI"; break;
         case '*': repl = "mU"; break;
         case '/': repl = "dI"; break;
         case '&': repl = "aN"; break;
         case '%': repl = "pE"; break;
         case '|': repl = "oR"; break;
         case '^': repl = "hA"; break;
         case '>': repl = "gR"; break;
         case '<': repl = "lE"; break;
         case '=': repl = "eQ"; break;
         case '~': repl = "wA"; break;
         case '.': repl = "dO"; break;
         case '(': repl = "oP"; break;
         case ')': repl = "cP"; break;
         case '[': repl = "oB"; break;
         case ']': repl = "cB"; break;
         case '{': repl = "lB"; break;
         case '}': repl = "rB"; break;
         case ';': repl = "sC"; break;
         case '#': repl = "hS"; break;
         case '?': repl = "qM"; break;
         case '`': repl = "bT"; break;
         case '!': repl = "nO"; break;
         case ',': repl = "cO"; break;
         case '$': repl = "dA"; break;
         case ' ': repl = "sP"; break;
         case ':': repl = "cL"; break;
         case '"': repl = "dQ"; break;
         case '@': repl = "aT"; break;
         case '\'': repl = "sQ"; break;
         case '\\': repl = "fI"; break;
      }
```

sethrj: You can always break that kind of a system unless you have a character that's reserved solely for escaping, right? I.e., "+foo" and "pLfoo" are the same. You'd have to do something crazy like use `_` as an escape and mangle user-provided `_` to `_Us` and prefixing an underscore to all those mangled guys above.

But no there really is no practical use case where the system in place breaks.

pcanal: > You can always break that kind of a system unless you have a character that's reserved solely for escaping, right? 

True.

> But no there really is no practical use case where the system in place breaks.

Maybe.  Concretely I see we are addressing:
```
celeritas/global/StepperGeant:-TestEm*:LeadBox*
```
What is the semantic of `:-`, `*:` and `*` there?  Why is there no risk of seeing (at some point later when we forgot about this problem of course :) ):
```
celeritas/global/StepperGeant:+TestEm*:LeadBox*
```
(or some other variation where the `-` and `*` are replaced?)

sethrj: The first `:` is the separator between the test name and the "filter". The second is the googletest glob filter syntax: there is no `+`, and in practice we'll never see `-Foo` and `*Foo` because then there would be overlap between the two cases. I don't see any alternative here as being worth the effort/cost due to the primitive nature of CMake commands.

sethrj: It's a good thought, and if we were using Python/C I'd say sure let's figure out a clean way to represent the special characters rather than coarsely filtering them all. But given the limitations in place, I think this is solution is acceptable.

pcanal:
```suggestion
        string(REGEX REPLACE "\*" "_all_" _xml_name "${_TEST_NAME}")
        string(REGEX REPLACE "\+" "_pl_" _xml_name "${_xml_name}")
        string(REGEX REPLACE "-" "_exc_" _xml_name "${_xml_name}")
        string(REGEX REPLACE "[^a-zA-Z0-9_]+" "_" _xml_name "${_xml_name}")
```
but on the other hand the gtest test selection is indeed probably rigid enough that the code as-is could be fine.

sethrj: Maybe easier is to allow those characters since they're legal filenames...

pcanal: Legal but 'hard' to use.

pcanal (approved):

cmake/CeleritasAddTest.cmake
sethrj: It only gets globbed up and parsed by the script 🤷‍♂️

Issue #1233, opened 2024-05-13
Remove hard limit on events


sethrj:
Problems:
- Event "ids" are effectively 64-bit unique identifiers used for global reproducibility. They are not consecutive from zero as we assumed when we first made them.
- When using multiple streams, events are usually partitioned among streams as opposed to being consecutive, so each stream will have `num_events * (num_threads - 1) / num_threads` zeros for its track counter.
- Using atomic counters for each event to determine track IDs results in nonreproducible track IDs.
- Having an event ID higher than `max_event` raises an assertion.

We could simplify reproducibility by requiring all tracks in flight to be from the same event (i.e., the usual way we integrate into Geant4). We could store the single event ID on the "core state" object since we don't need access to it on GPU. (Or, as of #1447, we just use the "unique event ID" to reset the RNG reproducibly.)

If we want to run multiple events simultaneously, the "event IDs" should be more like event *slots* so we can have up to N events in flight simultaneously, and once that event finishes we send an end-of-event "action" and let the event slot manage a new unique event. This methodology could let us have a single CPU thread + GPU simultaneously handle multiple Geant4 workers.

sethrj:
From @amandalund :
> with our standalone app I’ve typically seen better performance transporting multiple events at a time (e.g. about ~2x faster for  the cms2018+field+msc-vecgeom-gpu  regression problem with merge_events=true ). Our regression suite merges events for celer-sim on the gpu, right? I don’t think I’ve run those problems in single-event mode in a while, but it would be good to compare. I’d still probably prefer to keep that capability.

My follow up:
> fair enough! I think it'd also be good to run with a few more events (to reduce the load imbalance from each core having only a single event) and make sure that the merge_events=false case has optimal track slots

amandalund:
Here's where the speedup running one event at a time stands with all the latest performance improvements (but not partiitioning by charge when running with `merge_events` off, since it doesn't help there):
![rel-throughput-merge-events](https://github.com/user-attachments/assets/1ae3957e-f00b-4c7a-a623-f2e52b346d74)

Still a bit slower when not combining events, but I haven't tried increasing the total number of events or optimizing the number of track slots (which may also be problem dependent).

EDIT: Here's the same plot but with 4x the number of track slots when `merge_events` is off:
![rel-throughput-merge-events-4xts](https://github.com/user-attachments/assets/3e7430dd-3c5d-4ff5-b85d-e5fc3d92a024)

[...]

PR #1433, opened 2024-09-28
Add `track_order` option to celer-g4 and default to partitioning by charge on GPU

amandalund:
This add the `track_order` option to celer-g4 and changes the default track order in celer-sim and celer-g4 from unsorted to partitioning by charge when running on the GPU (based on the results in #1322, #1233, and below). This is the change in throughput for celer-g4 with partitioning:
rel-throughput-4xts-pc-g4

### DIFF
--- a/app/celer-g4/RunInputIO.json.cc
+++ b/app/celer-g4/RunInputIO.json.cc
@@ -13,28 +13,10 @@
 #include "corecel/io/StringEnumMapper.hh"
 #include "corecel/sys/Environment.hh"
 #include "celeritas/Types.hh"
+#include "celeritas/TypesIO.hh"
 #include "celeritas/ext/GeantPhysicsOptionsIO.json.hh"
 #include "celeritas/field/FieldDriverOptionsIO.json.hh"
 #include "celeritas/phys/PrimaryGeneratorOptionsIO.json.hh"
-
-namespace celeritas
-{
-//---------------------------------------------------------------------------//
-void from_json(nlohmann::json const& j, TrackOrder& value)
-{
-    static auto const from_string
-        = StringEnumMapper<TrackOrder>::from_cstring_func(to_cstring,
-                                                          "track order");
-    value = from_string(j.get<std::string>());
-}
-
-void to_json(nlohmann::json& j, TrackOrder const& value)
-{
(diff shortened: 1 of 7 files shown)


app/celer-g4/RunInputIO.json.cc
```diff
@@ -17,6 +17,25 @@
 #include "celeritas/field/FieldDriverOptionsIO.json.hh"
 #include "celeritas/phys/PrimaryGeneratorOptionsIO.json.hh"
 
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+void from_json(nlohmann::json const& j, TrackOrder& value)
```
sethrj: Since we use this in celer-sim as well, maybe we should make these in `celeritas/TypesIO.json.hh`? We could make it small and inline...

app/celer-sim/RunnerInput.hh
```diff
@@ -105,7 +105,7 @@ struct RunnerInput
     bool action_times{};
     bool merge_events{false};  //!< Run all events at once on a single stream
     bool default_stream{false};  //!< Launch all kernels on the default stream
-    bool warm_up{CELER_USE_DEVICE};  //!< Run a nullop step first
+    bool warm_up{false};  //!< Run a nullop step first
```
sethrj, line 108: As future work: maybe we should start using `std::optional` for some of these. Our whole input thing needs a refactor anyway to reduce duplication across the two apps.

sethrj (commented):
We ended up thinking too similarly again... I was actually working on this yesterday 🤪  https://github.com/sethrj/celeritas/commit/44722f4b9f03945ef7f2267d8af647b08641856c

LGTM though, this is nice to have by default!

sethrj (approved):

amandalund:
Haha oops... your slot diagnostic reminded me ;)

app/celer-g4/RunInputIO.json.cc
amandalund: Yeah good call.

app/celer-sim/RunnerInput.hh
amandalund: That could be nice... but yes, definitely needs a refactor, so perhaps I'll wait for that.

PR #1436, opened 2024-09-30
Fix weak vtables causing dynamic_cast error in apple clang 16

sethrj:
There's something wrong with the dynamic cast when using `-O1` or higher on apple clang 16 (i.e. XCode 16 aka the one targeting Sequoia):
```
test/orange/orangeinp/Solid.test.cc:309: Failure
Value of: dynamic_cast<ConeShape const*>(shape.get())
  Actual: false
Expected: true
actual shape: celeritas::orangeinp::Shape<celeritas::orangeinp::Cone>
```
for code
```c++
    {
        auto shape = ConeSolid::or_shape(
            "cone", Cone{{1, 2}, 10.0}, std::nullopt, SolidEnclosedAngle{});
        EXPECT_TRUE(shape);
        EXPECT_TRUE(dynamic_cast<ConeShape const*>(shape.get()))
            << "actual shape: " << demangle_shape(*shape);
    }
```

Moving the dynamic-cast code from the test to the library works:
```c++
ConeShape const* dyncast_cone_shape(ObjectInterface const* obj)
{
    return dynamic_cast<ConeShape const*>(obj);
}
```
and generates identical assembly to the function if added to the test binary:
```asm
celeritas::orangeinp::test::dyncast_cone_shape_2(celeritas::orangeinp::ObjectInterface const*): ; @_ZN9celeritas9orangeinp4test20dyncast_cone_shape_2EPKNS0_15ObjectInterfaceE
Lfunc_begin77:
	cbz	x0, LBB77_2
	ldr	x8, [x0]
	adrp	x9, vtable for celeritas::orangeinp::Shape<celeritas::orangeinp::Cone>@GOTPAGE
	ldr	x9, [x9, vtable for celeritas::orangeinp::Shape<celeritas::orangeinp::Cone>@GOTPAGEOFF]
	add	x9, x9, #16
	cmp	x8, x9
	b.eq	LBB77_3
LBB77_2:
	.loc	1 0 12 is_stmt 0                ; ./test/orange/orangeinp/Solid.test.cc:0:12
	mov	x0, #0                          ; =0x0
LBB77_3:
	.loc	1 304 5                         ; ./test/orange/orangeinp/Solid.test.cc:304:5
	ret
```
Note the above code loads the vtable for the object, and simply compares the vtable address to that of the pointer. Interestingly, the vtable shows up in both the library and the test: so we conclude that the vtable for the pointer created by the library is *not* the same vtable known to the unit test!!

This seems to be related to a new Clang optimization, https://reviews.llvm.org/D154658 (aka https://github.com/llvm/llvm-project/commit/9d525bf94b255df89587db955b5fa2d3c03c2c3e), perhaps also the cause of https://github.com/llvm/llvm-project/issues/71196 . Adding `-fno-assume-unique-vtables` as documented by the new feature causes the tests to work again.

This seems to be because the class being compared is `final` but since it's templated, *because it's instantiated in the main library but not declared external*, the vtable appears both in the library and test so the address is different between the two of them. Adding an `extern template` declaration fixes this.

### DIFF
--- a/src/celeritas/global/Stepper.hh
+++ b/src/celeritas/global/Stepper.hh
@@ -92,6 +92,9 @@ class StepperInterface
     //!@}
 
   public:
+    // Default virtual destructor
+    virtual ~StepperInterface();
+
     // Warm up before stepping
     virtual void warm_up() = 0;
 
@@ -114,8 +117,8 @@ class StepperInterface
     virtual SPState sp_state() = 0;
 
   protected:
-    // Protected destructor prevents deletion of pointer-to-interface
-    ~StepperInterface() = default;
+    StepperInterface() = default;
+    CELER_DEFAULT_COPY_MOVE(StepperInterface);
 };
 
 //---------------------------------------------------------------------------//
--- a/src/celeritas/global/Stepper.hh
+++ b/src/celeritas/global/Stepper.hh
@@ -192,4 +192,11 @@ class Stepper final : public StepperInterface
 };
 
 //---------------------------------------------------------------------------//
+// EXPLICIT INSTANTIATION
+//---------------------------------------------------------------------------//
+
+extern template class Stepper<MemSpace::host>;
+extern template class Stepper<MemSpace::device>;
+
+//---------------------------------------------------------------------------//
 }  // namespace celeritas
--- a/src/celeritas/global/CoreState.hh
+++ b/src/celeritas/global/CoreState.hh
@@ -230,4 +230,11 @@ auto& CoreState<M>::native_action_thread_offsets()
 }
 
 //---------------------------------------------------------------------------//
+// EXPLICIT INSTANTIATION
+//---------------------------------------------------------------------------//
+
+extern template class CoreState<MemSpace::host>;
+extern template class CoreState<MemSpace::device>;
+
+//---------------------------------------------------------------------------//
 }  // namespace celeritas
--- a/src/celeritas/global/Stepper.cc
+++ b/src/celeritas/global/Stepper.cc
@@ -54,6 +54,9 @@ ScopeExit(F&& func) -> ScopeExit<F>;
 }  // namespace
 
 //---------------------------------------------------------------------------//
+StepperInterface::~StepperInterface() = default;
+
+//---------------------------------------------------------------------------//
 /*!
  * Construct with problem parameters and setup options.
  */
--- a/src/celeritas/user/detail/StepGatherAction.hh
+++ b/src/celeritas/user/detail/StepGatherAction.hh
@@ -93,5 +93,12 @@ void step_gather_device(DeviceCRef<CoreParamsData> const&,
                         DeviceRef<StepStateData>&);
 
 //---------------------------------------------------------------------------//
+// EXPLICIT INSTANTIATION
+//---------------------------------------------------------------------------//
+
+extern template class StepGatherAction<StepPoint::pre>;
+extern template class StepGatherAction<StepPoint::post>;
+
+//---------------------------------------------------------------------------//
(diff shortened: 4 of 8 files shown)


esseivaju:
Interesting digging on that issue. Could this originate from the fact that we had explicit instantiation definitions of these template classes without matching explicit instantiation declarations? So we'd get duplicate vtables during explicit instantiation and implicit instantiation when the template is used.

EDIT: that's what your last paragraph says, I didn't read it carefully enough 😅

src/celeritas/global/Stepper.hh
```diff
@@ -191,5 +191,12 @@ class Stepper final : public StepperInterface
     std::shared_ptr<ActionSequenceT> actions_;
 };
 
+//---------------------------------------------------------------------------//
+// EXPLICIT INSTANTIATION
+//---------------------------------------------------------------------------//
+
+extern template class Stepper<MemSpace::host>;
+extern template class Stepper<MemSpace::device>;
```
esseivaju, line 202: This class has a public non-virtual dtor.

esseivaju (approved):
Looks good, nice fix! Small unrelated comment if you want to bundle that in that PR.

sethrj:
Yeah but the template functions should all be weak so I think it's legal? 😕

src/celeritas/global/Stepper.hh
sethrj: It's `final` so it doesn't need to be explicitly virtual? The base class is protected.

sethrj: But there's no good reason not to make the base destructor virtual so I'll do that.

pcanal: > It's final so it doesn't need to be explicitly virtual?

`Stepper`'s destructor has apriori no reason to be more virtual than it's base class.  However since the base `StepperInterface` exists, users could have tried to call the operator delete on a pointer to the base class ... except for the fact that that the destructor of the base class was declared protected ... but this prevents only external calls to the base class operator delete but does not prevent calls to that same operator from any of the `Stepper` classes ... and this misuse would not be caught with the current code ... to prevent it we would need to declare the operator delete itself private (we can't declare the destructor private since it is needed by the derived classes).

TLDR:  Seth is right, just make the base class destructor virtual since it already has a virtual table.

esseivaju:
yes it should be legal. From

> When the destination is a final class type that does derive from the source type, emit a direct comparison against the corresponding base class vptr value(s).

During the compilation of the test suit, they replace the `dynamic_cast` with a comparison of vtable pointers. During linking, the linker resolves the weak symbols, keeps only the first definition (the explicit instantiation), and discards the implicit instantiation from the test suit. So that's why the vtable pointers are not the same at runtime. I take that's what happens but I'm not sure why the linker can't get the correct vtable offset when resolving weak symbols. Maybe the code emitted after compilation uses the raw offset instead of a symbol, i.e. removing the explicit instantiation declaration and definition would also fix this (assuming there are no other implicit instantiations in the program).

src/celeritas/global/Stepper.hh
sethrj: @pcanal If anyone tries to submit a PR that contains `delete this`, they're off the project 😂

pcanal: This could happens a little bit easier :)
```
class Base { protected: ~Base(); };
class Derived { Base *other; public: ~Derived() { delete other; } };
```

sethrj: Ha, good point. That (and any other use of `delete` of Celeritas objects) would only merit a stern talking-to 😉

PR #1438, opened 2024-10-02
Add optical track initialization action

amandalund:
This add the track initialization action for the optical loop as well as optical simulation and particle data and views. I'll follow up with the action to locate vacancies.


src/celeritas/optical/CoreTrackView.hh
```diff
@@ -88,6 +96,42 @@ CoreTrackView::CoreTrackView(ParamsRef const& params,
...
+    }
+
+    // Initialize the particle state
+    this->particle()
+        = ParticleTrackView::Initializer{init.energy, init.polarization};
+
+    //! \todo Add physics view and clear physics state
+
+    return *this;
+}
```
sethrj, line 133: I like it! Future work for the main core track view?  (And would we need a second version for the "detailed initializer" that has the full volume state?)

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -37,8 +37,8 @@ class RayleighInteractor
 {
   public:
     // Construct interactor from an optical track
-    inline CELER_FUNCTION
-    RayleighInteractor(TrackView const& particle, Real3 const& direction);
+    inline CELER_FUNCTION RayleighInteractor(ParticleTrackView const& particle,
+                                             Real3 const& direction);
```
sethrj, line 41: Clang-format differences 😒 really gotta get this into the CI, don't worry about it here

sethrj (approved):
Excellent!

src/celeritas/optical/CoreTrackView.hh
amandalund: Yep! And yeah, gotta think about the best way to handle the special geo initialization...

amandalund:
Actually @sethrj if you want to hold off merging this, I might just add the "locate vacancies" to this track initialization action and refactor that part once we're producing secondaries.

sethrj:
Sure thing! Separate PR works just as well :)

amandalund:
Ok cool, then maybe I'll just do it separate since I think we do want it at the end of the loop so the counters are updated correctly.

PR #1439, opened 2024-10-03
Separate ImportOpticalModels from ImportOpticalMaterials

hhollenb:
# Separate `ImportOpticalModel`s from `ImportOpticalMaterial`s

Mean free paths (MFPs) used by optical models in optical physics are imported from Geant4 through the `G4MaterialPropertiesTable`, and have been previously stored in `ImportOpticalMaterials`. This PR moves the MFP tables to be stored instead in `ImportOpticalModels`, which allows a common way to access MFP tables used by discrete optical models, as well as extensibility for creating mock data or user optical models.

Discrete optical model data in `ImportOpticalMaterial`s now contains model specific data (e.g. parameters for WLS or parameters for Rayleigh scattering to build its own MFP tables). 

Optical models are imported from Geant4 in mostly the same way that EM processes are imported. When iterating over `G4VProcess`es, if they're a recognized optical process then they're loaded as an optical model. This now also respects the `GeantOpticalPhysicsOptions` for `absorption`, `rayleigh_scattering`, and `wavelength_shifting`, and supports `CelerOpticalPhysics` physics lists. Therefore we can set these options in Json files used by `celer-export-data` and only those specified models will be propagated to the root file. 

In a similar manner, it may be useful to handle the optical generators (Cerenkov, scintillation) in the import process loop so all the Geant4 processes are handled in the same way.


src/celeritas/io/ImportOpticalModel.hh
```diff
@@ -0,0 +1,56 @@
...
+ * This enum is used to identify the optical model that imported model MFP
+ * tables correspond to.
+ */
+enum class ImportModelClass
+{
+    other,
+    absorption,
+    rayleigh,
+    wls,
+    //
```
sethrj:
```suggestion
```

src/celeritas/io/ImportOpticalModel.hh
```diff
@@ -0,0 +1,56 @@
...
+}  // namespace optical
+
+//---------------------------------------------------------------------------//
+/*!
+ * Imported data for an optical physics model.
+ */
+struct ImportOpticalModel
+{
+    optical::ImportModelClass model_class;
+    std::vector<ImportPhysicsVector> mfps;  //!< per optical material MFPs
```
sethrj, line 51: I kind of regret that we added the plural `s` suffix to our other vectors 😅 it is consistent but it looks a little weird

src/celeritas/ext/detail/GeantOpticalModelImporter.cc
```diff
@@ -0,0 +1,119 @@
...
+    switch (imc)
+    {
+        case IMC::absorption:
+            return ImportOpticalModel{imc, this->import_mfps("ABSLENGTH")};
+        case IMC::rayleigh:
+            return ImportOpticalModel{imc, this->import_mfps("RAYLEIGH")};
+        case IMC::wls:
+            return ImportOpticalModel{imc, this->import_mfps("WLSABSLENGTH")};
+        default:
+            return ImportOpticalModel{IMC::size_, {}};
```
sethrj: Perhaps instead:

```suggestion
            CELER_ASSERT_UNREACHABLE();
```

sethrj (approved):
This looks fantastic! One or two super minor comments; and do you have any idea how the ROOT file decrease 45% in size? 🤔

app/celer-dump-data.cc
```diff
@@ -775,19 +775,31 @@ void print_atomic_relaxation_data(
...
  */
-void print_optical_materials(std::vector<ImportOpticalMaterial> const& iom)
+void print_optical_materials(std::vector<ImportOpticalModel> const& io_models,
+                             std::vector<ImportOpticalMaterial> const& io_mats)
 {
-    if (iom.empty())
+    if (io_models.empty())
+    {
+        CELER_LOG(info) << "Optical model data not available";
+        return;
```
amandalund, line 846: Could there still be optical material data to print below even if there are no imported models?

test/celeritas/data/lar-sphere.geant.json
```diff
@@ -0,0 +1,63 @@
...
+  "wavelength_shifting2": "none",
+  "boundary": {
+   "enable": false
+  },
+  "absorption": true,
+  "rayleigh_scattering": true,
+  "mie_scattering": false,
+  "verbose": false
+ }
+}
```
amandalund, line 62: You could also add "lar-sphere" to the data update in test/celeritas/CMakeLists.txt so we can run `make update-root-test-data` to regenerate that root file.

amandalund (approved):
This is great @hhollenb!

hhollenb:
> do you have any idea how the ROOT file decrease 45% in size? 🤔

Looking at `test/celeritas/ext/RootImporter.test.cc`, the test description says to use the option `reader_data = false` in `app/celer-export-geant` to reduce the size of the root files. Doing a quick test shows with `read_data = true`:
```
-rw------- 1 kcr8wx users 166K Oct  3 12:35 /home/kcr8wx/celeritas-project/celeritas-2/test/celeritas/data/four-steel-slabs.root
-rw------- 1 kcr8wx users 102K Oct  3 12:35 /home/kcr8wx/celeritas-project/celeritas-2/test/celeritas/data/lar-sphere.root
```
And with `read_data = false`:
```
-rw------- 1 kcr8wx users 76K Oct  3 12:33 /home/kcr8wx/celeritas-project/celeritas-2/test/celeritas/data/four-steel-slabs.root
-rw------- 1 kcr8wx users 66K Oct  3 12:33 /home/kcr8wx/celeritas-project/celeritas-2/test/celeritas/data/lar-sphere.root
```
The value of `read_data` is hard-coded to be true in `celer-export-geant`, so I'm guessing that at some point the root files started getting auto generated with the larger size and never got caught since both sizes pass tests.

stognini:
That sounds reasonable. Also, since we reorganized things over time, the comment in `RootImporter.test.cc` could be updated from 
```
* GeantImporter::DataSelection::reader_data = false
```
to something like
```
* GeantImportDataSelection::reader_data = false in GeantImporter.hh
```

stognini (approved):
Thanks @hhollenb ! Looks great!

hhollenb:
I added an extra optional flag to `celer-export-geant` called `--gen-test` so that we can set data selection flags / other options for root files that will be stored in test data directories. I added it to the `update-root-test-data` so test data will be automatically generated with the smaller files.

PR #1440, opened 2024-10-03
Add optical absorption model with imported data

hhollenb:
The base optical `Model` class with `AbsorptionModel` as an example case. `ImportedModels` is a list of `ImportOpticalModel`s with a map for built-in model classes to their imported identifier, and can be built directly or from imported data with a `ModelImporter` (included in a later PR). `ImportedModelAdapter` can then be used by models to access their MFP tables.

MFP tables are built with `detail::MfpBuilder` which internally handles the grid IDs that a model builds. It should probably be refactored when we refactor the rest of the grid builders.

For testing I created a `MockImportData` class to handle a bunch of common mock data and functions that the model classes use. I find that it makes writing the tests easy and fairly consistent, but could probably use a good critique.


src/celeritas/optical/Model.hh
```diff
@@ -0,0 +1,56 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Base class for discrete, volumetric optical models.
+ *
+ * For optical physics, there is precisely one particle (optical photons)
+ * and one energy range (optical wavelengths), so only models and no processes
+ * are used in optical physics.
+ */
+class Model : public StepActionInterface<CoreParams, CoreState>,
```
sethrj:
```suggestion
class Model : public OpticalStepActionInterface,
```
in `celeritas/optical/action/ActionInterface.hh`. I think we could move all the models to the `action` directory too?

src/celeritas/optical/Model.hh
```diff
@@ -0,0 +1,56 @@
...
+              public ConcreteAction
+{
+  public:
+    using ConcreteAction::ConcreteAction;
+
+    //! Virtual destructor for polymorphic deletion
+    virtual ~Model() = default;
```
sethrj: This is unnecessary since the base class is already a virtual destructor.

src/celeritas/optical/model/AbsorptionModel.cc
```diff
@@ -0,0 +1,62 @@
...
+/*!
+ * Build the mean free paths for the model.
+ */
+void AbsorptionModel::build_mfps(detail::MfpBuilder& build) const
+{
+    for (auto mat : range(OpticalMaterialId{imported_.num_materials()}))
+    {
+        build(imported_.mfp(mat));
+    }
+}
```
sethrj: I think we'll have a lot of implementations look like this: perhaps we *should* have `build_mfp(OpticalMaterialId, MfpBuilder&)`? Or analogously to the ORANGE proto builder, have the `MfpBuilder` store the current material ID being built?

src/celeritas/optical/detail/MfpBuilder.hh
```diff
@@ -0,0 +1,93 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Helper class for optical models to build MFP tables.
+ *
+ * Tracks individual grid IDs that have been built, and returns them
+ * as an ItemRange which may be used by model MFP tables. Can likely
+ * be refactored into GenericGridInserter.
+ */
+class MfpBuilder
```
sethrj: Since all volumetric optical models (internal or external) will need to use this class to initialize, it shouldn't be in the detail namespace. I don't think we'll want to refactor into the generic grid inserter; instead this class may "have a" grid inserter.

src/celeritas/optical/model/AbsorptionModel.hh
```diff
@@ -0,0 +1,42 @@
...
+    AbsorptionModel(ActionId id, ImportedModelAdapter imported);
+
+    // Build the mean free paths for this model
+    void build_mfps(detail::MfpBuilder&) const override final;
+
+    // Execute the model with host data
+    void step(CoreParams const&, CoreStateHost&) const override final;
+
+    // Execute the model with device data
+    void step(CoreParams const&, CoreStateDevice&) const override final;
```
sethrj: `override final` is redundant:
```suggestion
    void build_mfps(detail::MfpBuilder&) const final;

    // Execute the model with host data
    void step(CoreParams const&, CoreStateHost&) const final;

    // Execute the model with device data
    void step(CoreParams const&, CoreStateDevice&) const final;
```

Also marking the class as `final` will improve error checking (and, potentially, allow some small optimizations to take place).

src/celeritas/optical/ImportedModelAdapter.hh
```diff
@@ -0,0 +1,98 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using IMC = ImportModelClass;
+    using ImportedModelId = OpaqueId<ImportOpticalModel>;
+    //!@}
+
+  public:
+    // Built-in imported model classes
+    static std::set<IMC> const& builtin_model_classes();
```
sethrj: I don't think it's necessary to have builtin model classes since it's not possible for users to extend the enumerations. For now let's assume that the only models are the ones we've "hardcoded" (i.e., `range(IMC::size_)`)  into the import data, and worry later about extensibility.

src/celeritas/optical/ImportedModelAdapter.cc
```diff
@@ -0,0 +1,177 @@
...
+
+        // Check if IMC is built-in
+        auto iter = builtin_id_map_.find(imc);
+        if (iter != builtin_id_map_.end())
+        {
+            // Update imported model ID
+            ImportedModelId& mapped_id = iter->second;
+            if (!mapped_id)
+            {
+                mapped_id = ImportedModelId(model_id);
```
sethrj: Instead of the separate insert and find, you should be able to do this by `insert`ing and checking the `iterator, bool` result I think. But again, perhaps for now we should just try to simplify this to include only a one-to-one mapping for the builtin models?

sethrj (commented):
Sorry it took me a couple of days to review this with ongoing travel!

src/celeritas/optical/model/AbsorptionModel.cc
hhollenb: I think `build_mfp(OpticalMaterialId, MfpBuilder&)` is nicer because it makes `MfpBuilder` less stateful and less coupled to optical models. Looping in the `build_mfps` was a byproduct of not having an immediate way to get the number of optical materials in `optical::PhysicsParams::build_mfps`, but that should be easily resolved.

src/celeritas/optical/model/AbsorptionModel.hh
hhollenb: I didn't realize `final` fails for non-virtual functions that aren't overrides, that's good to know.

src/celeritas/optical/detail/MfpBuilder.hh
hhollenb: The refactoring was in reference to discussions about changing how core physics does their `GridBuilder`s. This is really just the same pattern that we do when building core XS tables:
 * Build data vector in reals
 * Put item ranges of built data into `GridRecord`
 * Build the grid in `...Data::grids`
 * After building grids for each material, put the range of the grid IDs into some table of type `ItemRange<GridId>`

The only difference is that in core physics the builder is cached and evaluated later, while in optical physics the grids are built when the `Model::build_mfps` function is called. There's nothing really specific to optical physics or MFPs in this code, so it feels like there should be some more common way to do this with core physics.

src/celeritas/optical/ImportedModelAdapter.cc
hhollenb: If we're limiting to just built-ins, then an `EnumArray` should probably work fine.

amandalund:
Sorry for the delay @hhollenb --this looks great at first glance, I'll go over it in detail first thing tomorrow!

src/celeritas/optical/detail/MfpBuilder.hh
sethrj: Definitely, but as I've emphasized, the core physics cross section calculation is some of our earliest attempts to move generic data onto device, and so we don't necessarily want to emulate it: this is our chance to prototype a better way of doing it (based on some of the iterations we've made in the materials, geometry, optical properties, ...) without the baggage of needing to refactor the other physics simultaneously.

For now, let's keep it as-is so long as we get something working, and then *before* adding further models we'll consider refactors/improvements.

src/celeritas/optical/ImportedModelAdapter.hh
```diff
@@ -0,0 +1,95 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using ImportedModelId = typename ImportedModels::ImportedModelId;
+    using SPConstImported = std::shared_ptr<ImportedModels const>;
+    //!@}
+
+  public:
+    // Create an adapter for the given model identifier
+    ImportedModelAdapter(ImportedModelId id, SPConstImported imported);
```
amandalund, line 77: Instead of constructing this with the `ImportModelId`, which is somewhat arbitrary, would it make sense to do what we do with the EM processes and construct it with the `ImportModelClass` instead? Then the models would be constructed with a shared pointer to `ImportedModels` and build the model adapter with their `ImportModelClass`.

src/celeritas/optical/ImportedModelAdapter.cc
```diff
@@ -0,0 +1,140 @@
...
+        builtin_id_map_.begin(), builtin_id_map_.end(), ImportedModelId{});
+
+    // Load all built-in IMCs into the map
+    for (auto model_id : range(models_.size()))
+    {
+        auto const& model = models_[model_id];
+
+        // Check imported data is consistent
+        CELER_EXPECT(model.model_class != IMC::size_);
+        CELER_EXPECT(model.mfps.size() == models_.front().mfps.size());
```
amandalund: Maybe we want these to be `VALIDATE`s since we're checking input data? And we might also want to check the first vector is non-empty.

test/celeritas/optical/MockImportedData.hh
```diff
@@ -0,0 +1,86 @@
...
+    //!@}
+
+    //!@{
+    //! \name Access mock data
+    static std::vector<ImportOpticalModel> const& import_models();
+    static std::vector<ImportOpticalMaterial> const& import_materials();
+
+    static ImportedModelId absorption_id();
+    static ImportedModelId rayleigh_id();
+    static ImportedModelId wls_id();
```
amandalund: If we change how the `ImportedModelAdapter` and models are constructed, we might also not need to hardcode these `ImportModelId`s in the tests.

amandalund (commented):
Really nice work @hhollenb!

src/celeritas/optical/ImportedModelAdapter.cc
hhollenb: I believe `models_` will have to be non-empty since the check is in a loop over `models_` (the first check will be redundant).

src/celeritas/optical/model/AbsorptionModel.cc
```diff
@@ -0,0 +1,61 @@
...
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct the model from imported data.
+ */
+AbsorptionModel::AbsorptionModel(ActionId id, ImportedModelAdapter imported)
+    : Model(id, "absorption", "interact by optical absorption")
+    , imported_(std::move(imported))
```
amandalund: I was thinking we could construct these similar to our EM `Process`s:
```c++
using SPConstImported = std::shared_ptr<ImportedModels const>;
...
AbsorptionModel(ActionId, SPConstImported);
```
```suggestion
AbsorptionModel::AbsorptionModel(ActionId id, SPConstImported model_data)
    : Model(id, "absorption", "interact by optical absorption")
    , imported_(ImportModelClass::absorption, model_data)
```
unless there's a need to have the `ImportedModelAdapter` outside of the model.

src/celeritas/optical/ImportedModelAdapter.cc
```diff
@@ -0,0 +1,159 @@
...
+    CELER_EXPECT(mid < imported_->num_models());
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Create an adapter from imported models for the given model class.
+ */
+ImportedModelAdapter::ImportedModelAdapter(ImportModelClass imc,
+                                           SPConstImported imported)
+    : model_id_(), imported_(imported)
```
amandalund:
```suggestion
    : imported_(imported)
```

src/celeritas/optical/ImportedModelAdapter.cc
amandalund: Ah sorry, I meant the first MFP physics vector.

src/celeritas/optical/model/AbsorptionModel.cc
hhollenb: Sorry, forgot to propagate the change to the actual `AbsorptionModel`.

src/celeritas/optical/ImportedModelAdapter.cc
hhollenb: Ah that makes sense. I should probably also check that all the MFP vectors in the table are valid as well then.

amandalund (approved):
Sweet, thanks @hhollenb! I'll let @sethrj take a final look before merging.

hhollenb:
Ack sorry I clicked merge before I read you comment. I'll setup the Rayleigh scattering PR which will be mostly the same and can be used for some further review.

sethrj:
I didn't even realize you had permissions to merge 🤔

PR #1441, opened 2024-10-03
Add optical "locate vacancies" and pre-step actions

amandalund:
This adds an optical stepping loop action to locate the vacant track slots (which can eventually be expanded to create initializers from secondaries) and a pre-step action (which for now just clears the step limit and sets the track status to alive).


sethrj (approved):
LGTM, thanks!

PR #1443, opened 2024-10-08
Add safety calculator for visualization

sethrj:
**DRAFT** because this is an unpolished work in progress: it should be cleaned up and added to celer-geo .

This adds a helper function that lets us evaluate the safety distance on a grid of points (for now, exporting to a JSON lines file) to compare across geometries, mainly to illustrate the current shortcoming but also for testing in the future.

For the following hex pin:
xy-mid

Geant4 is correct:
g4-pincell-xy-mid

And ORANGE adds internal boundaries:
org-pincell-xy-mid


sethrj:
@elliottbiondo if your dives into g4/vg safety code show any weird approximations/inaccuracies, this would be a great way to test them. I don't think I have the bandwidth to fully generalize and integrate this at the moment though.

sethrj:
Closing since this was a one-off experiment.

PR #1444, opened 2024-10-08
Implement spline interpolation for cross-section calculation

lebuller:
Adds new file in `grid`, `SplineXsCalculator.hh` that contains a new class, `SplineXsCalculator`. This class contains similar functions and members to the existing `XsCalculator` with the addition of a member variable for the order of the interpolation to be performed.

This class performs an interpolation on cross section grid data similar to `XsCalculator` except it allows an arbitrary order of interpolation (limited by the number of cross section data points). When interpolation near the min or max grid data point are requested, the order of the interpolation may be clipped to prevent indexing out-of-bounds.

A unit test for the Spline cross-section calculation was added that performs the same tests as the existing `XsCalculator` tests with interpolation orders ranging from 1-3


sethrj:
Thanks @lebuller ! For development I strongly recommend building with `-Wall -Wextra -pedantic -Werror`; you should consider creating `scripts/cmake-presets/${HOSTNAME}.json` with your build options to configure repeatably. Check out the other presets in that directory for examples.

```
/home/runner/work/celeritas/celeritas/src/celeritas/grid/SplineXsCalculator.hh:156:14: error: variable 'clipped' set but not used [-Werror,-Wunused-but-set-variable]
        bool clipped = false;
             ^
```

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+  private:
+    XsGridData const& data_;
+    Values const& reals_;
+    UniformGrid loge_grid_;
+    size_type order_;
+
+    CELER_FORCEINLINE_FUNCTION real_type get(size_type index) const;
+    CELER_FORCEINLINE_FUNCTION real_type interpolate(real_type energy,
```
sethrj, line 80: This function is too big to forceinline, I think. Since "forceinline" is about the definition rather than the declaration, we've started using `CELER_FUNCTION inline` in the declaration and `CELER_FORCEINLINE_FUNCTION` in the declaration (if appropriate, i.e. one or two lines).

test/celeritas/CMakeLists.txt
```diff
@@ -280,6 +280,7 @@ celeritas_add_test(grid/RangeCalculator.test.cc)
 celeritas_add_test(grid/ValueGridBuilder.test.cc)
 celeritas_add_test(grid/ValueGridInserter.test.cc)
 celeritas_add_test(grid/XsCalculator.test.cc)
+celeritas_add_test(grid/SplineXsCalculator.test.cc)
```
sethrj: Move this up a couple of lines to keep the tests sorted

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,268 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the raw cross section data at a particular index.
+ */
+CELER_FUNCTION real_type SplineXsCalculator::interpolate(
+    real_type energy, size_type low_idx, size_type high_idx) const
+{
+    CELER_EXPECT(high_idx < loge_grid_.size());
```
sethrj: Instead of looping over `[a, b]` and using `b+1` everywhere, use `[a, b)` which is more like how C++ works anyway.

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,265 @@
...
+    }
+    else if (loge >= loge_grid_.back())
+    {
+        lower_idx = loge_grid_.size() - 1;
+        result = this->get(lower_idx);
+        if (lower_idx >= data_.prime_index)
+        {
+            result /= energy.value();
+        }
+    }
```
sethrj: I wrote the original XS calculator toward the beginning of the project, and some of the things I wanted to try (having only one `return` per function) I think aren't doing their job (making it easier to understand by simplifying the flow).

We also want to avoid uninitialized objects: see http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es20-always-initialize-an-object .

I think a better solution (and I'll push this actually) is:
```
    auto calc_extrapolated = [this, &energy](size_type idx)
    {
        real_type result = this->get(idx);
        if (idx >= data_.prime_index)
        {
            result /= energy.value();
        }
        return result;
    };

    if (loge <= loge_grid_.front())
    {
        return calc_extrapolated(0);
    }
    if (loge >= loge_grid_.back())
    {
        return calc_extrapolated(loge_grid_.size() - 1);
    }
```
and then unindent the rest of the functoin.

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,265 @@
...
+            true_low_idx = 0;
+        }
+        if (lower_idx + order_steps < loge_grid_.size())
+        {
+            true_high_idx = lower_idx + order_steps;
+        }
+        else
+        {
+            true_high_idx = loge_grid_.size() - 1;
+        }
```
sethrj:
```
size_type true_high_idx = min(lower_idx + order_steps, loge_grid_.size() - 1);
```

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,265 @@
...
+CELER_FUNCTION real_type SplineXsCalculator::interpolate(
+    real_type energy, size_type low_idx, size_type high_idx) const
+{
+    CELER_EXPECT(high_idx < loge_grid_.size());
+    real_type result = 0.0;
+
+    // Outer loop over indices for contributing to the result
+    for (size_type outer_idx = low_idx; outer_idx < high_idx + 1; ++outer_idx)
+    {
+        real_type outer_e = std::exp(loge_grid_[outer_idx]);
```
sethrj, line 226: I think we want the spline interpolation to happen *before* applying the exponential scaling. The idea of the exponential uniform grid and 1/E scaling after `E'` is that the transformed function is smoother/more linear so that linear interpolation works better, although I think that's not exactly how Geant4 (or we) implement it... that's probably worth revisiting.

sethrj (changes requested):
Thanks @lebuller , this is an excellent first PR! I've got a couple of comments, plus one bigger request: the original unit tests were for linear interpolation which is trivial to verify; but I think we need to extend testing for functions that are polynomial as well.

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Good point, `interpolate` probably is too big to inline. That make sense, using the native c++ `inline` in declaration and `CELER_FORCEINLINE_FUNCTION` in the definition is the current strategy, I assume for `interpolate` that is probably unnecessary due to the function size and presence of loops so just using `CELER_FUNCTION` in both the declaration and definition is the right move with no inline options?

sethrj: The declaration still has to have the `inline` annotation but other than that, exactly! Almost all of our GPU-compatible code does this: member declarations tagged with `inline CELER_FUNCTION`, definitions just have `CELER_FUNCTION`. (Functions that are defined inside the class definition skip the `inline`.)

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Do you mean performing the interpolation in the log(E) space entirely or am I misunderstanding? Maybe an example on some actual xs functions/data would be helpful to me but I think this would just lead to an inaccurate interpolation. In testing it locally I found doing so on a linear function or even quadratic or exponential functions gives less accurate values than the version in this MR

sethrj:
Hi @lebuller , I'm sorry for not giving this more attention over the last couple of weeks: they've been extremely busy. I'll give this another look this afternoon!

src/celeritas/grid/SplineXsCalculator.hh
sethrj: You're right, after #1450 I see that we should be interpolating in `E` to be consistent. We definitely *should* generate some cross sections...

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,255 @@
...
+            }
+        }
+        real_type weight = num / denom;
+        real_type value = this->get(outer_idx);
+        if (outer_idx >= data_.prime_index)
+        {
+            value /= outer_e;
+        }
+
+        result += weight * value;
```
sethrj: Since registry pressure is so important for GPU, and I don't think the compiler can optimize this without `fast-math`, it might be better to combine weight and value into a single `weighted_value` scalar

test/celeritas/grid/SplineXsCalculator.test.cc
```diff
@@ -0,0 +1,446 @@
...
+{
+    auto reference_xs = [](real_type energy) {
+        auto result = 0.1 * energy * energy;
+        return result;
+    };
+
+    this->build({1.0e-3, 1e2}, 6, reference_xs);
+    this->convert_to_prime(0);
+    for (size_type order = 2; order < 5; ++order)
+    {
```
sethrj, line 228: To see where exactly the single-precision failure is occurring, add
```suggestion
    {
        SCOPED_TRACE("order=" + std::to_string(order));
```

It's probably nothing to worry about, but if there's some surprise numeric instability the single-precision test will usually find it :)

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,255 @@
...
+CELER_FUNCTION
+SplineXsCalculator::SplineXsCalculator(XsGridData const& grid,
+                                       Values const& values,
+                                       size_type const order)
+    : data_(grid), reals_(values), loge_grid_(data_.log_energy), order_(order)
+{
+    CELER_EXPECT(data_);
+    CELER_ASSERT(grid.value.size() == data_.log_energy.size);
+
+    // order of interpolation must be smaller than the grid size for effective
```
sethrj: For consistency we try to start comments with a capital letter.

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,255 @@
...
+    {
+        true_low_idx = 0;
+    }
+    size_type true_high_idx
+        = min(lower_idx + order_steps + 1, loge_grid_.size());
+
+    // if the requested interpolation order is even, a direction must be
+    // selected to interpolate to
+    if (order_ % 2 == 0)
+    {
```
sethrj: It often improves readability to put the comment inside the condition it applies to:
```suggestion
    if (order_ % 2 == 0)
    {
        // The requested interpolation order is even: select a direction to interpolate toward
```

src/celeritas/grid/SplineXsCalculator.hh
```diff
@@ -0,0 +1,255 @@
...
+        real_type low_dist = std::fabs(loge - loge_grid_[lower_idx]);
+        real_type high_dist = std::fabs(loge_grid_[lower_idx + 1] - loge);
+
+        // if we already clipped based on the bounding indices, don't clip
+        // again
+        if (true_high_idx - true_low_idx > order_ + 1)
+        {
```
sethrj: Same here

sethrj (approved):
Looks great @lebuller ! Just a few small comments to address. Thanks for the great first PR!

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Makes sense, I have combined those values, let me know if you intended it to be done differently

test/celeritas/grid/SplineXsCalculator.test.cc
lebuller: Added this to the quadratic tests. I'd guess that the failures are just caused by interpolating values with too large of a magnitude difference for single-precision to not lose accuracy

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Fixed here and in other comments in this file

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Fixed

src/celeritas/grid/SplineXsCalculator.hh
lebuller: Fixed

lebuller:
@sethrj I think all your comments should have been addressed with my recent push. If the float build still fails, I can try to decrease the energy grid size in the quadratic tests so that the values for neighboring bins are fewer orders of magnitude different. That should help the single-precision build

test/celeritas/grid/SplineXsCalculator.test.cc
sethrj: @lebuller I took a more careful look and this was because of accidental promotion to double-precision arithmetic: `0.1` is a double literal, so the reference value was doing the computation at a different precision level leading to slightly different results. (I.e., the single-precision was imprecise by 1e-7 but the reference should've had the same imprecise value.)

PR #1446, opened 2024-10-11
Add distance-to-bbox function

elliottbiondo:
This MR adds the ability to calculate the distance to the inside of a bounding box for a given point and direction. This small enhancement should be all that is necessary gearing up to a larger MR for BIH background intersections.


src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,68 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+
+            if (dist < 0)
+            {
+                continue;
+            }
+
+            // Find the intersection point
+            Array<U, 3> intersect = pos + dist * dir;
+
+            // Check that the intersection point occurs within region bounded
```
elliottbiondo: I could just use `is_inside` here, but then I would have to bump the calculated intersection distance by some tolerance. This approach side steps the issue.

elliottbiondo:
@sethrj this small MR is ready for you review

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,70 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+ */
+template<class T, class U>
+inline U calc_dist_to_inside(BoundingBox<T> const& bbox,
+                             Array<U, 3> const& pos,
+                             Array<U, 3> const& dir)
+{
+    if (is_inside(bbox, pos))
+    {
+        // Starting from inside the bbox so the distance to inside is zero
+        return 0.;
```
sethrj: Better to upcast I think:
```suggestion
        return 0;
```

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,70 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+            U dist = (bbox.point(Bound{bound})[ax] - pos[ax]) / dir[ax];
+
+            if (dist < 0)
+            {
+                // Short circut if the plane is behind us
+                continue;
+            }
+
+            // Calculate the actual intersection point
+            Array<U, 3> intersect = pos + dist * dir;
```
sethrj: Use `axpy` instead of temporaries

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,70 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+            for (auto other_ax : range(to_int(Axis::size_)))
+            {
+                if (other_ax != ax
+                    && (intersect[other_ax] < bbox.lower()[other_ax]
+                        || intersect[other_ax] > bbox.upper()[other_ax]))
+                {
+                    in_bounds = false;
+                    break;
+                }
+            }
```
sethrj: I'd say to write this as an algorithm but using `range` makes it hard... using an IFFE allows you to express the logic more clearly (or you could write this as an anonymous function and pass the local arguments)
```suggestion
            bool in_bounds = [&] {
                for (auto other_ax : range(to_int(Axis::size_)))
                {
                    if (other_ax == ax)
                        continue;

                    if (intersect[other_ax] < bbox.lower()[other_ax]
                        || intersect[other_ax] > bbox.upper()[other_ax])
                    {
                        return false;
                    }
                }
                return true;
            }();
```
If we wanted to be robust against NaNs, we'd have to write:
```c++
                    if (!(intersect[other_ax] >= bbox.lower()[other_ax]
                        && intersect[other_ax] <= bbox.upper()[other_ax]))
```

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,70 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+ * Calculate the distance to the inside of the bbox from a pos and dir.
+ *
+ * If there is no intersection, the result will be inf.
+ */
+template<class T, class U>
+inline U calc_dist_to_inside(BoundingBox<T> const& bbox,
+                             Array<U, 3> const& pos,
+                             Array<U, 3> const& dir)
+{
+    if (is_inside(bbox, pos))
```
sethrj: I think we might be better served to put this outside the function: I could imagine situations where we know *a priori* whether the point is inside/outside the box.

src/orange/BoundingBoxUtils.hh
sethrj: And actually you could probably move this function outside of the loop so that the intersect is an argument. And you should also consider evaluating `pos + dist * dir` for one element at a time, since you never need them all simultaneously (and in fact you *always* exclude one of them), to reduce register pressure.

src/geocel/Types.hh
```diff
@@ -66,10 +66,11 @@ enum class Axis
  * Here, lo/hi correspond to left/right, back/front, bottom/top. It's used for
  * the two points in a bounding box.
  */
-enum class Bound : bool
+enum class Bound
```
pcanal:
```suggestion
enum class Bound : unsigned char
```
(assuming that the reason for the previous `: bool` was to save space)

src/orange/BoundingBoxUtils.hh
pcanal: Then I assume we would a variation of:
```suggestion
    CELER_ASSERT(is_inside(bbox, pos);
```

elliottbiondo:
@sethrj @pcanal this is ready for another look

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,72 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+ */
+template<class T, class U>
+inline U calc_dist_to_inside(BoundingBox<T> const& bbox,
+                             Array<U, 3> const& pos,
+                             Array<U, 3> const& dir)
+{
+    CELER_EXPECT(!is_inside(bbox, pos));
+
+    // Test if an intersection is outside the bbox for a given axis
+    auto out_of_bounds = [&](U intersect, int ax) {
```
sethrj: Closures are a bit clearer when you explicitly specify the things they capture:
```suggestion
    auto out_of_bounds = [&bbox](U intersect, int ax) {
```

src/orange/BoundingBoxUtils.hh
```diff
@@ -221,6 +221,72 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
...
+
+            if (dist < 0)
+            {
+                // Short circut if the plane is behind us
+                continue;
+            }
+
+            // Check that the intersection point occurs within the region
+            // bounded by the planes of the other two axes
+            bool in_bounds = [&] {
```
sethrj: I think you could move this closure next to the `out_of_bounds` function and write it as
```suggestion
            bool in_bounds = [&] (int ax, U dist) {
```

sethrj (approved):
Couple of tiny little last comments, good to go once there! Thanks!

Issue #1144, opened 2024-03-08
ATLAS integration


sethrj:
This is the primary tracking issue for integrating Celeritas with the simulation workflow for the [ATLAS experiment](https://atlas.cern/). The primary Celeritas contact points are the "assignees" to the right.

# Athena integration

- [x] Demonstrate celeritas a part of FullSimLight (https://github.com/celeritas-project/GeoModel)
- [x] [Prototype celeritas external](https://gitlab.cern.ch/bmorgan/atlasexternals/-/tree/atlassim-6635?ref_type=heads)
- [x] [Prototype CPU integration](https://gitlab.cern.ch/bmorgan/athena/-/tree/atlassim-6635?ref_type=heads)
- [x] Protype GPU integration blocked by: 
  - [x]  https://github.com/celeritas-project/celeritas/issues/1462
  - [x] https://github.com/celeritas-project/celeritas/issues/1494
- [ ] Merge celeritas external https://gitlab.cern.ch/atlas/atlasexternals/-/merge_requests/1001
- [ ] Merge CPU integration
- [ ] Merge GPU integration

[ATLAS JIRA issue](https://its.cern.ch/jira/browse/ATLASSIM-6635) is accessible to ATLAS collaboration members

# Core capabilities

- [ ]  R,Z,θ magnetic field and reader: see [MagField::AtlasFieldSvc::readMap](https://gitlab.cern.ch/atlas/geomodelatlas/ATLASExtensions/-/blob/master/ATLASMagneticFieldMapPlugin/src/AtlasFieldSvc.cxx#L709) inside [atlas extensions](https://geomodel.web.cern.ch/home/fullsimlight/atlas-extensions/)
- [x] Use nav history to reconstruct G4 steps for callback due to bajillion EMEC boundaries: https://github.com/celeritas-project/celeritas/issues/1248
- [ ] Hit sequence for muon detectors: https://github.com/celeritas-project/celeritas/issues/1503
- [ ] Reconstruction properties for SDs #1505 
- [ ] Region support #983 

# Extra capabilities

- [ ] Woodcock tracking improved performance 17.5% overall in ATLAS due to highly segmented detectors
- [ ] Scoring observables #1594 
- [ ] Region- and particle-dependent energy loss enable/disable ([UR-105](https://jira-geant4.kek.jp/browse/UR-105))

drbenmorgan:
[...]

PR #1447, opened 2024-10-14
Define a "unique event ID" different from event counter

sethrj:
ATLAS integration is currently failing because the framework is [now setting the Geant4 event ID](https://gitlab.cern.ch/atlas/athena/-/merge_requests/74551) based on the global event context, whose event number is a globally unique identifier. In contrast, Celeritas expected the event number to be a *counter* (and a thread-local one at that). Since there are far more than 4 billion total events simulated across all runs and all simulations, and each event should have a unique random sequence, we need to define a special long integer for seeding.

Since the unique event ID (aka the event ID from geant4) doesn't necessarily correlate to the event *counter*, I'm forcing the *EventId* that we use for the primary (used for track counting) to zero. This may hurt "debuggability" but means that we can eliminate the "max events" from Geant4 integration. A follow-on PR will add support a special mode for resetting the track counter at the beginning of each event (and having a single "global" event ID) rather than using a unique event-ID-based slot.

A follow-on pull request should map our *internal* "event ID" (which is consecutive from zero and local to a thread) to the *external* "event ID" (which is some arbitrary integer used to reference an event, whether from ATLAS or a HepMC3 file etc) for user output.

See #1144 , #1233

### DIFF
--- a/src/accel/LocalTransporter.cc
+++ b/src/accel/LocalTransporter.cc
@@ -146,7 +146,7 @@ void LocalTransporter::InitializeEvent(int id)
     CELER_EXPECT(*this);
     CELER_EXPECT(id >= 0);
 
-    event_id_ = EventId(id);
+    event_id_ = UniqueEventId(id);
     track_counter_ = 0;
 
     if (!(G4Threading::IsMultithreadedApplication()
@@ -166,7 +166,6 @@ void LocalTransporter::InitializeEvent(int id)
 void LocalTransporter::Push(G4Track const& g4track)
 {
     CELER_EXPECT(*this);
-    CELER_EXPECT(event_id_);
 
     Primary track;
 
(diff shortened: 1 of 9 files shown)


pcanal:
>A follow-on pull request should map our internal "event ID" 

We may want to rename our internal "event ID" to avoid confusion with the external ones.

pcanal (approved):

src/accel/LocalTransporter.cc
```diff
@@ -199,7 +198,7 @@ void LocalTransporter::Push(G4Track const& g4track)
      * back to Geant4.
      */
     track.track_id = TrackId{track_counter_++};
-    track.event_id = event_id_;
+    track.event_id = EventId{0};
```
amandalund, line 201: So this is just temporary, and the plan is to set the Celeritas event ID from a global counter and to reset the (single, local) sim data track counter at the start of a new event? And I guess until we do that our track IDs won't be meaningful...

sethrj: Yep, the next PR will include a function to reset these at every event. Track IDs aren't currently very meaningful anyway since they're nondeterministic.

PR #1448, opened 2024-10-15
Remove trackid counter from primary

sethrj:
Following up to #1477, it's bad for us to manually assign track IDs since these should be *internal counters*: assigning them outside of the internal initialization will almost certainly result in duplicate track IDs. This removes track IDs from the primaries.

### DIFF
--- a/app/celer-g4/ActionInitialization.cc
+++ b/app/celer-g4/ActionInitialization.cc
@@ -36,7 +36,16 @@ ActionInitialization::ActionInitialization(SPParams params)
     // Create Geant4 diagnostics to be shared across worker threads
     diagnostics_ = std::make_shared<GeantDiagnostics>();
 
-    num_events_ = GlobalSetup::Instance()->setup_options().max_num_events;
+    if (auto const& hepmc_gen = GlobalSetup::Instance()->hepmc_gen())
+    {
+        num_events_ = hepmc_gen->NumEvents();
+    }
+    else
+    {
+        num_events_
+            = GlobalSetup::Instance()->input().primary_options.num_events;
+    }
+
     CELER_ENSURE(num_events_ > 0);
 }
 
@@ -73,7 +82,7 @@ void ActionInitialization::Build() const
     CELER_LOG_LOCAL(status) << "Constructing user action";
 
     // Primary generator emits source particles
-    if (auto hepmc_gen = GlobalSetup::Instance()->hepmc_gen())
+    if (auto const& hepmc_gen = GlobalSetup::Instance()->hepmc_gen())
     {
         this->SetUserAction(new HepMC3PrimaryGeneratorAction(hepmc_gen));
     }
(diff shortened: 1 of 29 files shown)


app/celer-g4/ActionInitialization.cc
```diff
@@ -73,7 +82,7 @@ void ActionInitialization::Build() const
     CELER_LOG_LOCAL(status) << "Constructing user action";
 
     // Primary generator emits source particles
-    if (auto hepmc_gen = GlobalSetup::Instance()->hepmc_gen())
+    if (auto const& hepmc_gen = GlobalSetup::Instance()->hepmc_gen())
```
pcanal, line 85: That seems that it would be a superfluous change (i.e. redundant with the change in the signature).

pcanal (approved):

pcanal (approved):
Look good except for at least one missed change (see CI failures)

app/celer-g4/ActionInitialization.cc
sethrj: `auto` copies the shared pointer, `const auto&` makes a reference to it.

pcanal: The '&' is necessary, the `const` is redundant.

src/accel/LocalTransporter.cc
```diff
@@ -192,12 +192,8 @@ void LocalTransporter::Push(G4Track const& g4track)
     }
 
     /*!
-     * \todo Celeritas track IDs are independent from Geant4 track IDs, since
-     * they must be sequential from zero for a given event. We may need to save
-     * (and share with sensitive detectors!) a map of track IDs for calling
-     * back to Geant4.
+     * \todo Eliminate event ID from primary.
      */
-    track.track_id = TrackId{track_counter_++};
```
amandalund, line 200: Can we remove this `track_counter_` from the local transporter now?

test/celeritas/track/TrackInit.test.cc
```diff
@@ -417,19 +416,38 @@ TYPED_TEST(TrackInitTest, primaries)
...
+        16, 17, 18, 19, 20, 21, 22, 23, 32, 33, 34, 35,
+        36, 37, 38, 39, 48, 49, 50, 51, 52, 53, 54, 55,
+    };
     auto result = RunResult::from_state(this->state());
     EXPECT_VEC_EQ(expected_track_ids, result.track_ids);
+    PRINT_EXPECTED(result.track_ids);
     EXPECT_VEC_EQ(expected_parent_ids, result.parent_ids);
     EXPECT_VEC_EQ(expected_vacancies, result.vacancies);
     EXPECT_VEC_EQ(expected_init_ids, result.init_ids);
+    PRINT_EXPECTED(result.init_ids);
```
amandalund:
```suggestion
    EXPECT_VEC_EQ(expected_parent_ids, result.parent_ids);
    EXPECT_VEC_EQ(expected_vacancies, result.vacancies);
    EXPECT_VEC_EQ(expected_init_ids, result.init_ids);
```

amandalund (approved):

PR #1449, opened 2024-10-15
Fix optical import when unused materials are defined

sethrj:
This reproduces and fixes a crash seen during the initial ATLAS run inside of the `GeantOpticalModelImporter` constructor. In #1439, I didn't notice that the "physical material IDs" (aka "material cuts couple") were used instead of the "geometry material IDs" (aka "materials"), so there ended up being an out-of-bounds error because there were fewer physics materials compared to geometry materials. It was also improper to do the remapping during setup for problems that *don't* use optical processes, so this fixes that as well.

### DIFF
--- a/src/celeritas/ext/GeantImporter.cc
+++ b/src/celeritas/ext/GeantImporter.cc
@@ -481,47 +481,54 @@ std::vector<ImportElement> import_elements()
 /*!
  * Store material-dependent optical properties.
  *
- * This returns:
- * - A vector of optical materials corresponding to an "optical material ID",
- *   and
- * - A "map" (indirection vector) of the optical material ID for each "geometry
- *   material"
+ * This returns a vector of optical materials corresponding to an "optical
+ * material ID".
  */
-std::pair<std::vector<ImportOpticalMaterial>, std::vector<OpticalMaterialId>>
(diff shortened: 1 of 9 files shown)


src/celeritas/ext/GeantImporter.cc
```diff
@@ -481,47 +481,54 @@ std::vector<ImportElement> import_elements()
...
                                        "RINDEX",
                                        ImportUnits::unitless);
-        if (!has_rindex)
-        {
-            // Refractive index is required: assume the material isn't optical
-            continue;
-        }
+        // Existence of RINDEX should correspond to GeoOpticalIdMap
+        // construction
+        CELER_ASSERT(has_rindex);
```
pcanal, line 531: Shouldn't this be a on-all-the-time exception (or at least a warning) to warn the user the wrong type of information is being loaded?

sethrj: No, this truly is just an internal consistency check between the `import_optical` function and the `GeoOpticalIdMap` constructor. The always-on warning at the top of the function is what tells the user if their request is inconsistent with what's there.

pcanal (approved):

PR #1451, opened 2024-10-15
Extend debug utilities for stuck/errored tracks

sethrj:
While debugging a hang in ATLAS, we tried the `debug_print` method [documented in our manual](https://celeritas-project.github.io/celeritas/user/development/testing.html) to see track information, but the symbol wasn't available due to use of LTO and/or `-Wl,--exclude-libs,ALL` and/or static libs. By adding a secret environment-based call, which we've placed in the most sensible location of `KernelContextException` to ensure that symbol stays in the library.

This also adds a `kill_active` method to the `Stepper` that can be called in a debugger or in between steps. It marks all tracks as "errored" and will apply a tracking cut at the next step, which (on CPU) also prints diagnostic information about the track.

A new `SetupOptions::geometry_output_file` can be used to export the in-memory geometry alongside the physics and primaries to aid in reproducing errors.

Finally, this greatly enhances the diagnostic output from errored/cut tracks, now printing out a JSON dump of the track *with* names substituted via core params metadata. This is accomplished by hiding a host-only "observer pointer" to CoreParams inside the core params data. Because that pointer is only usable on the host and only needed inside kernels, it shouldn't result in people abusing it.
```
Killing track {"geo":{"dir":[1.0,0.0,0.0],"is_on_boundary":true,"is_outside":false,"pos":[[-5.0,0.0,0.0],"cm"],"volume_id":"inner@0x0"},"particle":{"energy":[[0.0,"MeV"],"MeV"],"particle_id":"gamma"},"sim":{"event_id":0,"num_steps":1,"parent_id":-1,"post_step_action":"tracking-cut","status":"errored","step_length":[17.0,"cm"],"time":[0.25,"s"],"track_id":1},"thread_id":7,"track_slot_id":7}: lost 100 MeV
```

#1144


src/celeritas/global/KernelContextException.cc
```diff
@@ -133,6 +135,16 @@ void KernelContextException::initialize(CoreTrackView const& core)
             }
             surface_ = geo.surface_id();
         }
+
+        if (!celeritas::getenv("CELERITAS_EXCEPTION_DEBUG").empty())
```
pcanal, line 139: When is `void KernelContextException::initialize(CoreTrackView const& core)` called?

sethrj: During the constructor of the KernelContextException... which happens if there's an assertion/runtime failure inside one of the `launch_action` loops.

drbenmorgan:
This looks o.k. to me modulo the tests passing!

test/celeritas/global/Stepper.test.cc
```diff
@@ -229,6 +226,75 @@ TEST_F(SimpleComptonTest, TEST_IF_CELER_DEVICE(device))
...
+    step.kill_active();
+    counters = step();
+    EXPECT_EQ(0, counters.alive);
+    if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE
+        && CELERITAS_UNITS == CELERITAS_UNITS_CGS)
+    {
+        static char const* const expected_log_messages[] = {
+            "Killing 2 active tracks",
+            R"(Killing track 0 of event 0 (in track slot 6) at {-5, 0, 0} cm along {1, 0, 0}: depositing 100 MeV in volume 1)",
+            R"(Killing track 1 of event 0 (in track slot 7) at {-5, 0, 0} cm along {1, 0, 0}: depositing 100 MeV in volume 1)"};
```
pcanal:
```suggestion
            R"(Killing track 0 of event 0 (in track slot 6) at {-5, 0, 0} cm along {1, 0, 0}: depositing 100 MeV in volume 0)",
            R"(Killing track 1 of event 0 (in track slot 7) at {-5, 0, 0} cm along {1, 0, 0}: depositing 100 MeV in volume 0)"};
```

pcanal (approved):

test/celeritas/global/Stepper.test.cc
sethrj: Unfortunately the reverse happens on my machine... hence the update to actually print the volume *names* since the IDs might be nondeterminstic...

PR #1452, opened 2024-10-15
Improve single-event mode in `accel`

sethrj:
This follow-on to discussions in #1447 adds a non-synchronizing "fill" call to the track slot IDs (and also removes the synchronization in the RNG reseed). It also marks as "deprecated" the `max_num_events` setting in Celeritas. Finally, it also tests reproducibility using the full stepper.


src/corecel/data/Filler.hh
```diff
@@ -3,7 +3,7 @@
 // See the top-level COPYRIGHT file for details.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file corecel/data/detail/Filler.hh
+//! \file corecel/data/Filler.hh
```
amandalund, line 6: Maybe we should add a little `Filler` test?

src/celeritas/global/Stepper.cc
```diff
@@ -165,7 +165,7 @@ auto Stepper<M>::operator()(SpanConstPrimary primaries) -> result_type
 
 //---------------------------------------------------------------------------//
 /*!
- * Reseed the RNGs at the start of an event for "strong" reproducibility.
+ * Reseed RNGs and counters the start of an event for "strong" reproducibility.
```
pcanal:
```suggestion
 * Reseed RNGs and counters at the start of an event for "strong" reproducibility.
```

amandalund (approved):

test/celeritas/global/Stepper.test.cc
```diff
@@ -290,6 +291,46 @@ TEST_F(StepperOrderTest, step)
...
+    EXPECT_EQ(TrackId{0}, sim.track_id());
+    EXPECT_NE(orig_next_random, engine());
+    sim.status(TrackStatus::inactive);
+    step();
+
+    // Original event should have the same RNG sequence
+    step.reseed(UniqueEventId{123});
+    step(make_span(primaries));
+    EXPECT_EQ(TrackStatus::alive, sim.status());
+    EXPECT_EQ(TrackId{0}, sim.track_id());
```
pcanal, line 330: Should we extent the test to other ids than 0?

sethrj: I don't think it's necessary, since there's nothing special about track ID 0. We've got other tests to make sure that the code treats all track slots fairly, etc.

PR #1455, opened 2024-10-17
Add CELER_NONFATAL_FLUSH to cleanly kill stuck tracks

sethrj:
This adds a new environment variable that can be set to change the behavior of "max steps exceeded" and "interrupted". Normally, both of those call G4Exception which (when run through frameworks) kills the code. With the flag enabled, it will kill all active tracks (follow-on to #1451).

### DIFF
--- a/app/celer-g4/CMakeLists.txt
+++ b/app/celer-g4/CMakeLists.txt
@@ -108,7 +108,7 @@ function(celer_app_test ext)
     if(CELER_DISABLE_DEVICE)
       list(APPEND _extra_props DISABLED true)
     endif()
-  elseif(ext STREQUAL "cpu")
+  elseif(ext MATCHES "^cpu")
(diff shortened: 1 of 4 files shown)


app/celer-g4/CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ endfunction()
 
 celer_app_test("gpu")
 celer_app_test("cpu")
+celer_app_test("cpu-nonfatal")
```
pcanal, line 133: Is the testing cpu specific or is the feature cpu specific?

pcanal (approved):

app/celer-g4/CMakeLists.txt
sethrj: Neither :) The test adds 100% coverage to the new code.

PR #1456, opened 2024-10-17
Add muon decay interactor

stognini:
This PR adds the muon decay interactor. Muons can only decay via one channel, either
```math
\mu^- \longrightarrow e^- \overline{\nu}_e \nu_\mu
```
or
```math
\mu^+ \longrightarrow e^+ \nu_e \overline{\nu}_\mu .
```

An important caveat is that the Geant4 physics manual defines the maximum possible sampled energy for the electron to be $$E_\text{max} = m_\mu / 2$$, while the source code defines $$E_\text{max} = m_\mu / 2 - m_e$$. This difference leads to a small missing energy, as the total energy of secondaries on a muon decay at rest is not equal to the muon rest energy in the case of the source code definition.


stognini:
# Sampling loop
The `MuDecayInteractor` uses two `for` loops with a hardcoded limit of `1000` iterations, instead of usual `do {} while`'s, to sample the secondary energy. A test with 10k iterations leads to only up to 50 samplings, as the plot shows, being far below the `1000` threshold. I kept the sampling equivalent to Geant4, but we can move to a `while` loop if that is preferred.

sample-counts

# $$E_\text{max}$$ definition difference

Using $$E_\text{max} = m_\mu / 2$$ (the physics manual one) I do recover the muon mass on a decay at rest (as shown in the `MuDecayInteractor.test`). Nevertheless, since our sampling limit changes, our sampled energies are slightly shifted to higher values for all 3 secondaries. Plots below show the shift for the electron. Top plot uses the definition used in the source code, while the bottom uses the one defined in the physics manual.

code-energy_e

manual-energy_e

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    muon_nu_dir = rotate(muon_nu_dir);
+
+    // Boost secondaries to the lab frame
+    auto charged_lep_4vec = this->to_lab_frame(
+        charged_lep_dir, charged_lep_energy, shared_.electron_mass);
+    auto electron_nu_4vec
+        = this->to_lab_frame(electron_nu_dir, electron_nu_energy, 0);
+    auto muon_nu_4vec = this->to_lab_frame(muon_nu_dir, muon_nu_energy, 0);
+
+    // \todo Return electron only?
```
amandalund: Looking great @stognini! Still have to go over most of this, but since we were talking about this point on the call, I guess my preference would be to just create the electron secondary until we have a use case for the neutrinos. The extra bit of computation won't be a big deal, but the two extra secondaries per decay (that we can't track anyway) could increase the secondary/initializer memory requirements a bit. And since you've already implemented it and we'll have it in the history, we can easily just revert the change when we need it. (I'll let @sethrj and @whokion weigh in, though.)

whokion: I do agree with @amandalund for now (unless we want to add neutrino-nucleus processes in a near-term future or want to send these neutrinos back to CPU so that Geant4 can handle them in the case that users add neutrino processes in their physics list, of which data transfer needs to be supported in a similar way when we add  lepto-/photo-nuclear processes).  Of course, please share any other users cases that you are considering. 

Regarding to the maximum energy of the daughter electron from the muon decay, I think that the formulas in the P.M. is an approximation (i.e, assuming that secondaries are massless in the V-A calculation except the primary muon).  The implementation in Geant4 may be more practically/experically correct.  You may still recover the energy conservation in the rest frame of the muon following correct kinematics between (e + n_e) + n_muon and muon as necessary - I am surprised that Geant4 does not calculate them correctly.  Anyway, I will also review and give more detailed comments later.  Thanks for a good addition!

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    template<class Engine>
+    inline CELER_FUNCTION Interaction operator()(Engine& rng);
+
+  private:
+    //// DATA ////
+
+    // Constant data
+    MuDecayData shared_;
+    // Incident muon energy
+    units::MevEnergy const inc_energy_;
```
whokion:
```suggestion
    Energy const inc_energy_;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    // Incident muon direction
+    Real3 const inc_direction_;
+    // Allocate space for a secondary particles
```
whokion:
```suggestion
    // Allocate space for secondary particles 
```
or a secondary particle (electron) if we agree to store only electron.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    StackAllocator<Secondary>& allocate_;
+    // Define decay channel based on muon or anti-muon primary
+    ParticleId secondary_id_[3];
```
whokion: Do we need this temporary particle id?  Can we use the data ids directly when fill the particle_id of secondaries?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    // Define decay channel
+    auto const& ids = shared_.ids;
+    bool muon = (particle.particle_id() == ids.mu_minus) ? true : false;
+    secondary_id_[0] = muon ? ids.electron : ids.positron;
+    secondary_id_[1] = muon ? ids.anti_electron_neutrino
+                            : ids.electron_neutrino;
+    secondary_id_[2] = muon ? ids.muon_neutrino : ids.anti_muon_neutrino;
+
+    // Set up muon four vector to boost decay to the lab-frame
+    Real3 inc_momentum = inc_direction_ * inc_energy_.value();
```
whokion:
```suggestion
    Real3 inc_momentum = inc_direction_ * particle.momentum();
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    inc_fourvec_ = {
+        inc_momentum,
+        std::sqrt(ipow<2>(norm(inc_momentum)) + ipow<2>(shared_.muon_mass))};
```
whokion:
```suggestion
    inc_fourvec_ = {inc_direction_ * particle.momentum(), particle_total_energy()};
```
and drop the line above?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+
+    // Sampling constants
+    sample_max_ = real_type{1}
+                  + ipow<2>(shared_.electron_mass) / ipow<2>(shared_.muon_mass);
```
whokion:
```suggestion
    sample_max_ = real_type{1} + ipow<2>(shared_.electron_mass / shared_.muon_mass);
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+
+    // Geant4 physics manual defines E_{max} = m_\mu / 2, while the source code
+    // (since v10.2.0 at least) defines E_{max} = m_\mu / 2 - m_e . The source
+    // code implementation leads to a total CM energy of ~104.6 MeV instead of
+    // the expected 105.7 MeV (muon mass), which is achieved by using the
+    // physics manual definition
```
whokion: As I shortly mentioned earlier, the energy conservation at the rest frame of the muon may be nothing to do with this E_{max}. Can you double check whether this difference is due to the wrong initialization of the muon four vector?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    max_energy_ = 0.5 * shared_.muon_mass;
```
whokion:
```suggestion
    max_energy_ = real_type{0.5} * shared_.muon_mass;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+/*!
+ * Sample a muon decay via the
+ * \f[
+ * \mu^- \longrightarrow e^- \overline{\nu}_e \nu_\mu
+ * \f]
+ * or
+ * \f[
+ * \mu^+ \longrightarrow e^+ \nu_e \overline{\nu}_\mu
+ * \f]
+ * channel, with a branching ratio of 100%.
```
whokion: Can you consolidate this with a similar block at the top?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    real_type electron_energy_frac;
+    real_type electron_nu_energy_frac;
+    size_type const max_iter{1000};  // \todo Why?
+
+    for ([[maybe_unused]] auto i : range(max_iter))
+    {
+        for ([[maybe_unused]] auto j : range(max_iter))
+        {
+            electron_nu_energy_frac = sample_max_ * generate_canonical(rng);
+            if (generate_canonical(rng)
```
whokion: Use BernoulliDistribution(...) (rng)

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+                <= electron_nu_energy_frac
+                       * (real_type{1} - electron_nu_energy_frac))
+            {
+                break;
+            }
+            electron_nu_energy_frac = sample_max_;
+        }
+
+        electron_energy_frac = generate_canonical(rng);
+        if (electron_nu_energy_frac >= real_type{1} - electron_energy_frac)
```
whokion: Same here.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+        {
+            break;
+        }
+        electron_nu_energy_frac = real_type{1} - electron_energy_frac;
+    }
```
whokion: Probably `do ... while();` may be better to avoid the max_iter ?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    real_type muon_nu_energy_frac = real_type{2} - electron_energy_frac
+                                    - electron_nu_energy_frac;
+
+    // Momentum of secondaries at rest frame; neutrino masses are ignored
+    auto charged_lep_energy
+        = this->momentum(electron_energy_frac, shared_.electron_mass);
```
whokion:
```suggestion
        = this->calc_momentum(electron_energy_frac, shared_.electron_mass);
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+
+    //// HELPER FUNCTIONS ////
+
+    // Boost four vector from the rest frame to the lab frame
+    inline CELER_FUNCTION FourVector to_lab_frame(Real3 const& dir,
+                                                  Energy const& energy,
+                                                  real_type const& mass);
+
+    // Calculate particle momentum (or kinetic energy) in the center of mass
+    inline CELER_FUNCTION Energy momentum(real_type const& energy_frac,
```
whokion:
```suggestion
    inline CELER_FUNCTION Energy calc_momentum(real_type const& energy_frac,
```

whokion: And Energy --> MevMomentum?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate final particle momentum (or kinetic energy) from its sampled
+ * fractional energy.
+ */
+CELER_FUNCTION units::MevEnergy
+MuDecayInteractor::momentum(real_type const& energy_frac, real_type const& mass)
+{
+    return Energy{std::sqrt(ipow<2>(energy_frac) * ipow<2>(max_energy_)
```
whokion:
```suggestion
    return Energy{std::sqrt(ipow<2>(energy_frac * max_energy_)
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    Real3 muon_nu_dir
+        = {-(electron_nu_energy_frac / muon_nu_energy_frac) * sin_theta,
+           0,
+           -electron_energy_frac / muon_nu_energy_frac
+               - (electron_nu_energy_frac / muon_nu_energy_frac) * cos_theta};
+
+    // Apply a spherically uniform rotation to the decay
+    real_type euler_phi
+        = UniformRealDistribution<real_type>(0, 2 * constants::pi)(rng);
+    real_type euler_theta = std::acos(2 * generate_canonical(rng) - 1);
```
whokion:
```suggestion
    real_type euler_theta = std::acos(UniformRealDistribution<real_type>(-1, 1));
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+    real_type euler_psi
+        = UniformRealDistribution<real_type>(0, 2 * constants::pi)(rng);
```
whokion: Probably, define
`using UniformRealDist = UniformRealDistribution<real_type>; `
in the class type alias and use UniformRealDist in other all places too.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,299 @@
...
+ */
+CELER_FUNCTION FourVector MuDecayInteractor::to_lab_frame(Real3 const& dir,
+                                                          Energy const& energy,
+                                                          real_type const& mass)
+{
+    CELER_EXPECT(norm(dir) > 0);
+    CELER_EXPECT(energy > zero_quantity());
+    CELER_EXPECT(mass >= 0);
+
+    Real3 p = dir * energy.value();  // Momentum [MeV]
```
whokion: ??

src/celeritas/em/xs/MuBremsDiffXsCalculator.hh
```diff
@@ -39,7 +39,7 @@ namespace celeritas
  * The contribution to the cross section from the nucleus is given by
  * \f[
    \Phi_n = \ln \frac{B Z^{-1/3} (\mu + \delta(D'_n \sqrt{e} - 2))}{D'_n (m +
-   \delta \sqrt{e} B Z^{-1/3})} \ ,
+   \delta \sqrt{e} B Z^{-1/3})} ,
```
whokion: Okay, but this is nothing to do with this MR.

whokion (commented):
@stognini Please address the first round of review comments.  Thanks.

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: Or if it's a rejection function, use that helper instead

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: I initially had a `is_muon_` boolean assigned via the `ParticlTrackView`, but ended up with that array so that I could set up the particle decay channel during construction instead of operator().

Since we are agreeing on allocating the electron only, I'll revert to the boolean, which is enough to define the secondary id. This will also clean up the neutrino ids in `MuDecayData` and `PDGNumber`...

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: Unlikely. I set up a test problem in G4 with a vacuum box and a muon with zero energy decaying in it. When I add up the total energy of secondaries I still can't recover the muon mass.

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: By top you mean above the class? Because I do have that there also.

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: Yep. If you see the top plot on my comment just after the pull request, it shows that the number of iterations on these two `for` loops are way below that `max_iter` defined in G4. I am happy to convert this to `do while` if everyone is fine with it. My worry was that G4 used the for loops because of some very rare cases that the sampling got stuck for too long, so they just sample the max electron energy on those cases.

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: Equivalent to `inc_direction_ * particle.momentum().value()`, since [E] = [p].

src/celeritas/em/xs/MuBremsDiffXsCalculator.hh
stognini: I can revert it. Seemed an overkill to fix it on a separate one.

amandalund: Yeah, too small unless it were bundled with some other doc fixes... I think you can keep it here.

sethrj: Isn't this ok as is though? `\ ` in latex is how to add a space between an equation and a punctuation.

stognini: Yeah, that would be just an extra space when rendered. I can revert my nitpicking.

src/celeritas/decay/data/MuDecayData.hh
```diff
@@ -0,0 +1,61 @@
...
+/*!
+ * Device data for creating a muon decay interactor.
+ */
+struct MuDecayData
+{
+    //! Particle identifiers
+    MuDecayIds ids;
+
+    //! Muon/anti-muon mass [MeV]
+    real_type muon_mass;
```
whokion:
```suggestion
    units::MevMass muon_mass;
```

src/celeritas/decay/data/MuDecayData.hh
```diff
@@ -0,0 +1,61 @@
...
+
+    //! Electron/positron mass [MeV]
+    real_type electron_mass;
```
whokion: same here.  Maybe `Mass` with `using Mass = units::MevMass;`

src/celeritas/decay/data/MuDecayData.hh
```diff
@@ -0,0 +1,61 @@
...
+
+    //! Check whether the data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return ids && muon_mass > 0 && electron_mass > 0;
```
whokion:
```suggestion
        return ids && muon_mass > zero_quantity() && electron_mass >  zero_quantity();
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Perform muon decay.
+ *
+ * Only one decay channel is implemented, with muons decaying to
+ * \f[
+ * \mu^- \longrightarrow e^- \overline{\nu}_e \nu_\mu
```
whokion:
```suggestion
 * \mu^- \longrightarrow e^- \bar{\nu}_e \nu_\mu
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+ * \f]
+ * or
+ * \f[
+ * \mu^+ \longrightarrow e^+ \nu_e \overline{\nu}_\mu
```
whokion: same here.  Probably overline may be okay - even though I always use bar for anti-neutrinos.  You may know better as you came from the neutrino community.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+ * \f].
+ *
+ * This interactor follows \c G4MuonDecayChannel and the Physics Reference
+ * Manual, Release 11.2, section 4.2.3.
+ *
+ * \warning Neutrinos are currently not returned in this interactor to minimize
+ * secondary memory requirements.
```
whokion:
```suggestion
 * \warning Neutrinos are currently not returned in this interactor as they 
 * are not tracked down or transported.
```
As it sounds like that we do a short cut for performance^^.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+
+    // Sample an interaction with the given RNG
+    template<class Engine>
+    inline CELER_FUNCTION Interaction operator()(Engine& rng);
+
+  private:
+    //// DATA ////
+
+    // Constant data
+    MuDecayData shared_;
```
whokion:
```suggestion
    MuDecayData const& shared_;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    // Incident muon energy
+    Energy const inc_energy_;
+    // Incident muon direction
+    Real3 const inc_direction_;
```
whokion:
```suggestion
    Real3 const& inc_direction_;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    real_type sample_max_;
+    // Maximum electron energy
+    real_type max_energy_;
+
+    //// HELPER FUNCTIONS ////
+
+    // Boost four vector from the rest frame to the lab frame
+    inline CELER_FUNCTION FourVector to_lab_frame(Real3 const& dir,
+                                                  MevMomentum const& momentum,
+                                                  real_type const& mass);
```
whokion:
```suggestion
                                                  real_type const& mass) const;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+
+    // Calculate particle momentum (or kinetic energy) in the center of mass
+    inline CELER_FUNCTION MevMomentum calc_momentum(real_type const& energy_frac,
+                                                    real_type const& mass);
```
whokion:
```suggestion
                                                    real_type const& mass) const;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
whokion: I would keep one of them as they are redundant.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+ * \f[
+ * \mu^- \longrightarrow e^- \overline{\nu}_e \nu_\mu
+ * \f]
+ * or
+ * \f[
+ * \mu^+ \longrightarrow e^+ \nu_e \overline{\nu}_\mu
+ * \f]
+ * channel, with a branching ratio of 100%.
+ *
+ * The decay is sampled at the muon's reference frame (based on
```
whokion:
```suggestion
 * The decay is sampled at the muon rest frame (based on
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    // Allocate secondaries
+    Secondary* secondaries = allocate_(1);
+    if (secondaries == nullptr)
+    {
+        // Failed to allocate secondaries
+        return Interaction::from_failure();
+    }
+
+    real_type electron_energy_frac;
+    real_type electron_nu_energy_frac;
```
whokion:
```suggestion
    real_type electron_energy_frac{0};
    real_type electron_nu_energy_frac{0};
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    {
+        do
+        {
+            electron_nu_energy_frac = UniformRealDist(0, sample_max_)(rng);
+        } while (generate_canonical(rng)
+                 > electron_nu_energy_frac
+                       * (real_type{1} - electron_nu_energy_frac));
+
+        electron_energy_frac = generate_canonical(rng);
+    } while (electron_nu_energy_frac < real_type{1} - electron_energy_frac);
```
whokion: This may be safe without a loop check as the sampling is well-defined.  Nevertheless, have you thought about combining two do-while loops into one (i.e., 2D sampling with a composition and rejection technique as their energies are correlated)?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+
+    // Apply a spherically uniform rotation to the decay
+    auto sample_twopi = UniformRealDist(0, 2 * constants::pi);
+    real_type euler_phi = sample_twopi(rng);
+    real_type euler_theta = std::acos(UniformRealDist(-1, 1)(rng));
+    real_type euler_psi = sample_twopi(rng);
+
+    EulerRotation rotate(euler_phi, euler_theta, euler_psi);
+    Real3 charged_lep_dir = {0, 0, 1};
+    charged_lep_dir = rotate(charged_lep_dir);
```
whokion: As we are not creating neutrinos, we may need a comment for not calculating` cos_theta` between the decay electron and its neutrinos (i.e., electron neutrino direction w.r.t electron).   As discussed, I guess that the electron direction can be simply sampled by 
`charged_lep_dir = from_spherical(UniformRealDist(-1, 1)(rng), UniformRealDist(0, 2 * constants::pi));`
without `EulerRotation`.

whokion (commented):
@stognini Thanks for the update.  Please see the second set of comments (which are mostly minors).

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: I actually did not really think when writing this, but the `\overline` and `\bar` are more of a stylistic preference. I see people using both, but for whatever reason `overline` usually renders better outside latex documents, so I just got used to it: $$\overline{\nu}$$ (overline) vs. $$\bar{\nu}$$ (bar). If we want to standardize and use `\bar`, I am okay with it.

src/celeritas/decay/data/MuDecayData.hh
```diff
@@ -0,0 +1,67 @@
...
+    //! Check whether the data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return ids && muon_mass > zero_quantity()
+               && electron_mass > zero_quantity();
+    }
+};
+
+using MuDecayDeviceRef = MuDecayData;
+using MuDecayHostRef = MuDecayData;
```
amandalund:
```suggestion
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    //// HELPER FUNCTIONS ////
+
+    // Boost four vector from the rest frame to the lab frame
+    inline CELER_FUNCTION FourVector to_lab_frame(Real3 const& dir,
+                                                  MevMomentum const& momentum,
+                                                  Mass const& mass) const;
+
+    // Calculate particle momentum (or kinetic energy) in the center of mass
+    inline CELER_FUNCTION MevMomentum calc_momentum(real_type const& energy_frac,
+                                                    Mass const& mass) const;
```
amandalund: No reason not to pass these by value:
```suggestion
    inline CELER_FUNCTION FourVector to_lab_frame(Real3 const& dir,
                                                  MevMomentum momentum,
                                                  Mass mass) const;

    // Calculate particle momentum (or kinetic energy) in the center of mass
    inline CELER_FUNCTION MevMomentum calc_momentum(real_type energy_frac,
                                                    Mass mass) const;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,249 @@
...
+    : shared_(shared)
+    , inc_energy_(particle.energy())
+    , inc_direction_(inc_direction)
+    , allocate_(allocate)
+{
+    CELER_EXPECT(particle.particle_id() == shared_.ids.mu_minus
+                 || particle.particle_id() == shared_.ids.mu_plus);
+
+    // Define decay channel
+    is_muon_ = (particle.particle_id() == shared_.ids.mu_minus) ? true : false;
```
amandalund:
```suggestion
    is_muon_ = (particle.particle_id() == shared_.ids.mu_minus);
```
This (and the variables below) could also go directly in the member initializer list.

test/celeritas/phys/InteractorHostTestBase.cc
```diff
@@ -32,8 +32,10 @@ InteractorHostTestBase::InteractorHostTestBase()
 
     constexpr MevMass emass{0.5109989461};
     constexpr MevMass mumass{105.6583745};
+    constexpr double mu_decay_constant{1 / 2.1969811e-6};  // [s^-1]
 
     // Default particle params
+    // clang-format off
```
amandalund:
```suggestion
```

src/corecel/math/EulerRotation.hh
```diff
@@ -0,0 +1,101 @@
...
+ * The Euler angles follow the x-convention, where:
+ * - \em phi is the angle of the first rotation (D), on the \em z-axis
+ * - \em theta is the angle of the second rotation (C), on the \em x'-axis
+ * (which is the rotated x-axis)
+ * - \em psi is the angle of the third rotation (B), on the \em z'-axis (which
+ * is the rotate z-axis)
+ *
+ * \note For details see https://mathworld.wolfram.com/EulerAngles.html
+ */
+class EulerRotation
```
amandalund: This is no longer used now that we're only storing the electron secondary, right? Should we postpone adding it until it is, or do you think we'll need it for something else in the near term?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,246 @@
...
+    real_type electron_energy_frac{};
+    real_type electron_nu_energy_frac{};
+    do
+    {
+        do
+        {
+            electron_nu_energy_frac = UniformRealDist(0, sample_max_)(rng);
+        } while (generate_canonical(rng)
+                 > electron_nu_energy_frac
+                       * (real_type{1} - electron_nu_energy_frac));
```
amandalund: The `RejectionSampler` could be used here (if the sampling is not refactored according to Soon's suggestion).

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,246 @@
...
+    // Incident muon direction
+    Real3 const& inc_direction_;
+    // Allocate space for secondary particles (currently, electron only)
+    StackAllocator<Secondary>& allocate_;
+    // Define decay channel based on muon or anti-muon primary
+    bool is_muon_;
+    // Incident muon four vector
+    FourVector inc_fourvec_;
+    // Max sampled fractional energy for the electron neutrino
+    real_type sample_max_;
```
amandalund: Alternatively you could just store the distribution for sampling:
```suggestion
    // Sample the fractional energy for the electron neutrino
    UniformRealDist sample_energy_fraction_;
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,246 @@
...
+/*!
+ * Boost secondary to the lab frame.
+ *
+ * \note This assumes the primary to be at rest and, thus, there is no need
+ * to perform an inverse boost of the primary at the CM frame.
+ */
+CELER_FUNCTION FourVector MuDecayInteractor::to_lab_frame(
+    Real3 const& dir, MevMomentum const& momentum, Mass const& mass) const
+{
+    CELER_EXPECT(norm(dir) > 0);
```
amandalund:
```suggestion
    CELER_EXPECT(is_soft_unit_vector(dir));
```
?

amandalund (commented):
Looking good @stognini! A few small suggestions:

src/corecel/math/EulerRotation.hh
stognini: We should remove it if we want to be strict about following rules... but for using it in the near future, my guess it's that it will likely be needed select random directions for the outgoing 3 secondaries produced by the muCF process. Thoughts?

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: So this is a weird detail that happens in the sampler that doesn't allow me to use any of the standard helper classes (that have reasonable assumptions, such as asserting that the acceptance value is non-negative). Here is the problem:

The `electron_nu_energy_frac` range is `[0, sample_max_)`, which is `[0, 1.000023)`. Now, the `f_` value set in the `RejectionSampler` is `f_ = electron_nu_energy_frac * (1 - electron_nu_energy_frac)`. So that means that for the upper limit (`sample_max_ = 1.000023`), `f_ < 0`, which leads to an assertion error.

src/corecel/math/EulerRotation.hh
sethrj: If it's not used in this PR, please open a separate one for it. I think something like this already exists in `orange/MatrixUtils.hh` so let's open a discussion separately.

stognini: Sounds good. Removing until further notice.

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: Overline is usually for multi-character bits, and bar is used for single character modifications right?

sethrj:
@stognini can you add a decay section to `doc/implementation` and include the new decay class? I guess we'd want it to have its own section at the same level as EM/optical as opposed to a subsection of EM or living inside "core"?

src/celeritas/decay/data/MuDecayData.hh
```diff
@@ -0,0 +1,64 @@
...
+ */
+struct MuDecayData
+{
+    //!@{
+    //! \name Type aliases
+    using Mass = units::MevMass;
+    //!@}
+
+    //! Particle identifiers
+    MuDecayIds ids;
```
sethrj: We don't really need separate IDs for these any more: that used to be a limitation of how we were instantiating the interactors. It might improve readability a bit through: `data.ids.positron` instead of `data.positron_id`?

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,242 @@
...
+ * \f]
+ * or
+ * \f[
+ * \mu^+ \longrightarrow e^+ \nu_e \overline{\nu}_\mu
+ * \f].
+ *
+ * This interactor follows \c G4MuonDecayChannel and the Physics Reference
+ * Manual, Release 11.2, section 4.2.3.
+ *
+ * \warning Neutrinos are currently not returned in this interactor as they
```
sethrj: I think this should be a `note` ... no one is going to be doing EM physics and neutrino physics inside the same calculation since the cross sections are so absurdly different.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,242 @@
...
+    // Constant data
+    MuDecayData const& shared_;
+    // Incident muon energy
+    Energy const inc_energy_;
+    // Incident muon direction
+    Real3 const& inc_direction_;
+    // Allocate space for secondary particles (electron only)
+    StackAllocator<Secondary>& allocate_;
+    // Define decay channel based on muon or anti-muon primary
+    bool is_muon_;
```
sethrj: This is confusing... maybe `is_antiparticle_` since if true it creates a positron rather than electron?

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: Correct, I think you cannot do `\bar{xyz}`. So, always use `\bar` for anti-particles then? It reads better on the code (since many people actually say pbar instead of anti-proton), and makes more sense, and we accept the less nice rendering.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,242 @@
...
+                                     ParticleTrackView const& particle,
+                                     Real3 const& inc_direction,
+                                     StackAllocator<Secondary>& allocate)
+    : shared_(shared)
+    , inc_energy_(particle.energy())
+    , inc_direction_(inc_direction)
+    , allocate_(allocate)
+    , is_muon_(particle.particle_id() == shared_.ids.mu_minus)
+    , inc_fourvec_({inc_direction_ * particle.momentum().value(),
+                    particle.total_energy().value()})
```
sethrj: should be
```suggestion
    , inc_fourvec_{inc_direction_ * value_as<MevMomentum>(particle.momentum()),
                    value_as<MevEnergy>(particle.total_energy())}
```
to avoid a copy and to guarantee the consistency between the units

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,242 @@
...
+ * \c G4MuonDecayChannel::DecayIt ) and the resulting four vector is boosted to
+ * the lab frame using the muon's incident energy and direction.
+ *
+ * Since only the charged lepton is being returned, these steps are ommitted:
+ * - Calculate the angle between charged lepton and electron neutrino.
+ * - Calculate the energies of the neutrinos.
+ * - Calculate the final directions of the neutrinos using the aforementioned
+ * angle and conservation of momentum.
+ * - Sample a spherically uniform direction and rotate all three secondaries.
+ * - Boost all particles to the lab frame
```
sethrj: Move this documentation to the main class body docs, so that it shows up in the user manual.

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: I agree we should try to refactor this. Do you know the underlying PDFs to be sampled? It would be really good to explicitly document those as well.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,242 @@
...
+/*!
+ * Calculate final particle momentum (or kinetic energy) from its sampled
+ * fractional energy.
+ */
+CELER_FUNCTION units::MevMomentum
+MuDecayInteractor::calc_momentum(real_type energy_frac, Mass mass) const
+{
+    return MevMomentum{
+        std::sqrt(ipow<2>(energy_frac * max_energy_)
+                  + 2 * energy_frac * max_energy_ * mass.value())};
```
sethrj, line 228: Maybe this function should just be electron momentum? Not passing in the mass (just used `shared_.electron_mass`) and instead passing in `Energy{energy_frac * max_energy_}`.

src/celeritas/phys/Interaction.hh
```diff
@@ -35,6 +35,7 @@ struct Interaction
     {
         scattered,  //!< Still alive, state has changed
         absorbed,  //!< Absorbed or transformed to another particle type
+        decay,  //!< N-body decay
```
sethrj: If we're going to shoehorn the decay process into the regular "interactor", this is exactly equivalent to absorption, right? So let's just delete this and the `from_decay`.

And in fact, you *need* to do this, because the `InteractionApplier` tries to change the direction if
`(result.action != Interaction::Action::absorbed)`, and it only kills the particle if it's absorbed.

So we should probably just delete this and use "from_absorption" . If you (or anyone else) think it's too confusing to mark it as `absorbed`, then just use
```suggestion
        decay = absorbed,  //!< N-body decay
```
and replace your function below with
```c++
CELER_FORCEINLINE_FUNCTION Interaction from_decay() { return from_absorption(); }

File: test/celeritas/decay/MuDecay.test.cc [hunk 50]
```diff
@@ -0,0 +1,139 @@
...
+
+        for (auto i : range(sec.size()))
+        {
+            auto const& part = sec[i];
+            avg_sec_energies[i] += part.energy.value();
+            for (auto j : range(3))
+            {
+                avg_total_momentum[j] += part.direction[j]
+                                         * part.energy.value();
+            }
```
sethrj:
```suggestion
            axpy(part.direction, part.energy.value(), &avg_total_momentum);
```

test/celeritas/decay/MuDecay.test.cc
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    for (auto j : range(num_secondaries))
+    {
+        avg_sec_energies[j] /= num_samples;
```
sethrj: If you include `ArrayOperators` you can just do `avg_sec_energies /= num_samples`

test/celeritas/phys/InteractorHostTestBase.cc
```diff
@@ -32,31 +32,31 @@ InteractorHostTestBase::InteractorHostTestBase()
...
+           {"mu_minus",
+            pdg::mu_minus(),
+            mumass,
+            ElementaryCharge{-1},
+            mu_decay_constant},
+           {"mu_plus",
+            pdg::mu_plus(),
+            mumass,
+            ElementaryCharge{1},
+            mu_decay_constant}};
```
sethrj: Keep the trailing comma to help the formatting

sethrj (changes requested):

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: They won't be doing neutrino physics per se, but vertex quantities won't be conserved, and reconstruction/filtering for analyses might depend directly on verification of conservation of E and p to reconstruct a mother particle from an observed signal. So I thought it was reasonable to have a warning.

src/celeritas/decay/interactor/MuDecayInteractor.hh
amandalund: Or since there's only one secondary now, maybe we should just store the secondary ID instead?

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: Does Geant4 actually do this? Would the user have to have custom physics/stepper code to pull out the neutrinos and save their momentum and kill them?

stognini: Geant4 just returns all three particles. So if the user doesn't want the neutrinos to be transported and/or use their initial state, then yes, the user would likely have that coded in their `G4VUserTrackingAction` implementation.

src/celeritas/decay/interactor/MuDecayInteractor.hh
stognini: This was originally as is because it was the same function for all three secondaries. I kept it isolated for future refactoring if we need to add the neutrinos, but it could technically just be in the main interactor body, since it is just used once.

src/celeritas/phys/Interaction.hh
stognini: But we still end up with `result.action == absorbed` in this case, right? Because having it tagged as decay in our actions is important. For example: in a perfect vacuum, a particle will never undergo absorption, but it will always decay if not stable.

sethrj: Currently we don't store the interaction result in any way or provide a way for the user to interrogate it. It's only used to update the post-step data in the InteractionApplier. I think that's OK, since we should be able to use the post-step action ID for users to interrogate that sort of thing.

sethrj: Maybe we should change the name from `absorbed` to `killed`? But then that might be ambiguous with a particle that had an error and had to be killed.

gxbert: So, what is then the downside or the necessity of introducing this additional Action, decay or from_decay()?  Indeed, it is very confusing that `decay = absorbed` as the decay process is a material independent (in a sense that the particle never interacts with a material and nerved be absorbed, but decay) even though minimizing the number of actions may be better if it simplifies underlying logic without loss of generality.

whokion: Sorry that I used a wrong account for the earlier comment by gxbert (now deleted).  Recovering/rewording it here (but shorten):  It is confusing that `decay = absorbed `as the decay process is a material independent (in a sense that the particle never interacts with a material and never be physically absorbed).  The action decay = absorbed will be ambiguous if the muon capture process is added later which can be considered as absorbed and physically totally different process even though multiple processes can be categorized to the same action.  Nevertheless, a question (to @stognini ) is whether there is any specific treatment for decay products from others (for InteractionApplier) which may justify the addition.

src/celeritas/decay/interactor/MuDecayInteractor.hh
amandalund: I also don't think sampling the energy fraction in [0, `sample_max`) makes sense... won't that be equivalent to sampling it in [0, 1), just slightly less efficient?

amandalund: Some thoughts on how this might be refactored: it looks like the inner rejection loop samples the $\nu_e$ energy fraction from the PDF $f(x) = 6 x (1 - x)$, $x \in [0, 1)$. This has the CDF $F(x) = 3x^2 - 2x^3$, so to use the inverse transform method we'd have to invert that cubic function.

This PDF is also the beta distribution $\text{B}(2, 2)$. To [sample from a beta distribution](https://en.wikipedia.org/wiki/Beta_distribution#Random_variate_generation) $\text{B}(\alpha, \beta)$, we can draw two independent samples from a gamma distribution, $X \sim \Gamma(\alpha, \theta)$ and $Y \sim \Gamma(\beta, \theta)$: then $\frac{X}{X + Y} \sim \text{B}(\alpha, \beta)$.

Our `GammaDistribution` implements the algorithm described in [Marsaglia and Tsang](https://doi.org/10.1145/358407.358414) which uses a rejection method with a very high (>95%) acceptance rate, but also has nested loops and draw samples from a normal distribution. For this simple case where we'd just need samples from $\Gamma(2, 1)$, we could use [this technique](https://en.wikipedia.org/wiki/Gamma_distribution#Random_variate_generation) which would just require generating two exponentially distributed RVs: $-\sum_{k=1}^n \ln U_k \sim \Gamma(n, 1)$.

I think this would at least be more efficient than the current rejection method.

src/celeritas/phys/Interaction.hh
stognini: My initial enum addition was a starting point to expand the `InteractionApplier` when the decay process class comes in, as well as its inclusion in the stepping loop on future PRs. So we would have a particle killed if it is absorbed or underwent decay. For now, our actions are internal, but I assume that at some point we will be able to provide access to actions to the end user, similar to querying processes in G4, for analysis purposes. So one could, for example, store only data from a given decay, if their focus is, say, understanding signals from a given channel.

sethrj: We can always query processes by looking at the post-step action ID, since each of those correlates to a model (decay or interaction) that the track undergoes. I don't think we'll want to expand that further. I think instead we may want to make the "interaction result" less about the physics and more about the programming implementation: whether to update the direction and/or kill the particle, e.g.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,243 @@
...
+ * angle and conservation of momentum.
+ * - Sample a spherically uniform direction and rotate all three secondaries.
+ * - Boost all particles to the lab frame
+ *
+ * \warning Neutrinos are currently not returned by this interactor as they
+ * are not tracked down or transported and would significantly increase
+ * secondary memory allocation usage.
+ *
+ * \note The full three-body decay can be recovered from PR #1456, commit
+ * `ecc4326`.
```
sethrj:
```suggestion
 * \note Neutrinos are currently not returned by this interactor as they
 * are not tracked down or transported and would significantly increase
 * secondary memory allocation usage. See discussion and commit history
 * at https://github.com/celeritas-project/celeritas/pull/1456 .
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,243 @@
...
+                                     StackAllocator<Secondary>& allocate)
+    : shared_(shared)
+    , inc_energy_(particle.energy())
+    , inc_direction_(inc_direction)
+    , allocate_(allocate)
+    , sec_id_((particle.particle_id() == shared_.mu_minus_id)
+                  ? shared_.electron_id
+                  : shared_.positron_id)
+    , inc_fourvec_({inc_direction_ * value_as<MevMomentum>(particle.momentum()),
+                    value_as<Energy>(particle.total_energy())})
```
sethrj: Does this not work?
```suggestion
    , inc_fourvec_{inc_direction_ * value_as<MevMomentum>(particle.momentum()),
                    value_as<Energy>(particle.total_energy())}
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: Is it efficient enough to reject `f(x)` using `g(x)=1/24`? That would give us about a 2/3 acceptance rate and take two random samples?

src/celeritas/phys/Interaction.hh
stognini: Fair point. Should we rename `absorbed` to a catch-all word, such as `stopped`? This would cover all cases: scatters, stops (absorption, decay), remains unchanged or fails during sampling. Hopefully this is not confusing, such as you mentioned with using `killed`.

sethrj: `stopped` means "lost all energy but still exists" in the geant4 parlance right? We don't want that. Let's discuss at today's meeting.

test/celeritas/phys/InteractorHostTestBase.cc
```diff
@@ -32,6 +32,7 @@ InteractorHostTestBase::InteractorHostTestBase()
 
     constexpr MevMass emass{0.5109989461};
     constexpr MevMass mumass{105.6583745};
+    constexpr double mu_decay_constant{1 / 2.1969811e-6};  // [s^-1]
```
sethrj: This should include programmatic units for the builds that aren't CGS.

src/celeritas/phys/Interaction.hh
```diff
@@ -35,6 +35,7 @@ struct Interaction
     {
         scattered,  //!< Still alive, state has changed
         absorbed,  //!< Absorbed or transformed to another particle type
+        decay = absorbed,  //!< N-body decay
```
sethrj: For consistency should be `decayed`

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,243 @@
...
+    auto charged_lep_energy
+        = this->calc_momentum(electron_energy_frac, shared_.electron_mass);
+
+    // Apply a spherically uniform rotation to the decay
+    IsotropicDistribution rotate;
+    Real3 charged_lep_dir = rotate(rng);
+
+    // Boost secondaries to the lab frame
+    auto charged_lep_4vec = this->to_lab_frame(
+        charged_lep_dir, charged_lep_energy, shared_.electron_mass);
```
sethrj: I'm sure this was originally written like this to keep the neutrinos, but...

```suggestion
    // Decay isotropically in rest frame, and boost secondaries to the lab frame
    auto charged_lep_4vec = this->to_lab_frame(
        IsotropicDistribution{}(rng), charged_lep_energy, shared_.electron_mass);
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
amandalund: Sorry, just fixed a typo (the PDF normalization is 6,  not 1/6) but yeah g(x) = 3/2 would also be good enough (3 random samples on average).

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,234 @@
...
+ * muon's rest frame, with the result being uniformly rotated and finaly
+ * boosted to the lab frame.
+ *
+ * As it is a three-body decay, the energy sampling happens for \f$ e^\pm \f$
+ * and \f$ \nu_e(\bar{\nu}_e) \f$, with the \f$ \nu_\mu(\bar{\nu}_\mu) \f$
+ * final energy directly calculated from energy conservation. The sampling loop
+ * selects fractional energies \f$[0,1)\f$ for the first two particles. For the
+ * electron neutrino, its energy fraction \f$ f_{E_{\nu_e}} \f$ is sampled from
+ * the unnormalized PDF \f$ f(x) = x(1-x), \ x \in [0,1) \f$, with a maximum
+ * acceptance probability of \f$ g(x) = 0.25 \f$. The charged lepton's
```
amandalund:
```suggestion
 * selects fractional energies \f$[0,1)\f$ for the first two particles. The
 * electron neutrino's energy fraction \f$ f_{E_{\nu_e}} \f$ is sampled from
 * the PDF \f$ f(x) = 6 x(1-x), \ x \in [0,1) \f$ using the rejection method,
 * with proposal distribution PDF \f$ g(x) = U(0,1) \f$ and bounding
 * constant \f$ M = 1.5 \f$. The charged lepton's
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,234 @@
...
+ */
+class MuDecayInteractor
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Energy = units::MevEnergy;
+    using MevMomentum = units::MevMomentum;
+    using Mass = units::MevMass;
+    using UniformRealDist = UniformRealDistribution<real_type>;
```
amandalund:
```suggestion
```

amandalund (approved):
Looks good on my end, thanks @stognini!

test/celeritas/phys/InteractorHostTestBase.cc
```diff
@@ -50,12 +52,12 @@ InteractorHostTestBase::InteractorHostTestBase()
          pdg::mu_minus(),
          mumass,
          ElementaryCharge{-1},
-         stable_decay_constant},
+         mu_decay_constant.value()},
```
sethrj: Wrong: this should be `native_value_from(mu_decay_constant )`. If the "CLHEP" unit test is passing, this probably means you've got some implicit `seconds` unit elsewhere...

Also looking at `ParticleParams` the documentation there should be 
```c++
        real_type decay_constant;  //!< Decay constant [1/time]
```

This might not matter at this point since we're not actually calculating the "effective cross section" (aka decay constant / speed)?

sethrj (approved):
Looks good, thanks! Just the last comment about units.

test/celeritas/phys/InteractorHostTestBase.cc
stognini: Sorry I missed the last bit. I think this class is only used by interactors, so the decay constant should not affect any test. I added it for consistency, so if any future interactor needs to use the decay constant of the muon we have a reasonable value.

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,235 @@
...
+ * \mu^- \longrightarrow e^- \bar{\nu}_e \nu_\mu
+ * \f]
+ * or
+ * \f[
+ * \mu^+ \longrightarrow e^+ \nu_e \bar{\nu}_\mu
+ * \f].
+ *
+ * This interactor follows \c G4MuonDecayChannel::DecayIt and the Physics
+ * Reference Manual, Release 11.2, section 4.2.3. The sampling happens at the
+ * muon's rest frame, with the result being uniformly rotated and finaly
```
whokion:
```suggestion
 * muon's rest frame, with the result being uniformly rotated and finally
```

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,235 @@
...
+            electron_nu_energy_frac = generate_canonical(rng);
+        } while (RejectionSampler(
+            electron_nu_energy_frac * (real_type{1} - electron_nu_energy_frac),
+            real_type{0.25})(rng));
+
+        electron_energy_frac = generate_canonical(rng);
+    } while (electron_nu_energy_frac + electron_energy_frac < real_type{1});
+
+    // Momentum of secondaries at rest frame
+    auto charged_lep_energy
```
whokion:
```suggestion
    auto charged_lep_momentum
```
just to be consistent

src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -0,0 +1,235 @@
...
+                                                          MevMomentum momentum,
+                                                          Mass mass) const
+{
+    CELER_EXPECT(is_soft_unit_vector(dir));
+    CELER_EXPECT(momentum > zero_quantity());
+    CELER_EXPECT(mass >= zero_quantity());
+
+    Real3 p = dir * momentum.value();
+    FourVector lepton_4vec{
+        p, std::sqrt(ipow<2>(norm(p)) + ipow<2>(mass.value()))};
```
whokion:
```suggestion
        p, std::sqrt(dot_product(p, p) + ipow<2>(mass.value()))};
```

test/celeritas/decay/MuDecay.test.cc
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    avg_sec_energies /= num_samples;
+    avg_total_momentum /= num_samples;
+
+    // With only one secondary being returned, there is no expectation of
+    // energy or momentum conservation
+    static double const expected_avg_sec_energies[] = {384.834176348064};
+    static double const expected_avg_total_momentum[]
+        = {0.369721983532691, 0.109494401494642, 383.064043487629};
```
whokion: Are these values consistent with Geant4 or your expectation?  It may be good to add a test with a low energy muon (e.g., 1 MeV).

whokion (approved):
@stognini Good work!  I left a couple of minor comments/suggestions which you may consider before merging.  Thanks.

test/celeritas/decay/MuDecay.test.cc
stognini: I'll add the low energy test.

About the expectation: when the 3 neutrinos were being returned the numbers made sense, as they added up to the total muon energy in the lab frame.
It is trickier to compare this directly to G4 since G4 will propagate the muon until it decays, and it needs ionization for that to happen. So I can only have an interactor-like behavior in G4 if I create muons at rest.

Thus, I am trusting that since at rest the electron energy distribution follows G4 (and experiment, at least visually -- I have to collect a table with real data to make a final plot), the resulting boost to lab frame is also correct.

whokion: Thank you for adding a low energy muon test.

Actually, G4 has both `in-flight` and `at-rest`  muon decays.  Of course, the in-flight muon decay should compete with regular processes.  As the muon half lifetime is relatively long, a test detector with a large volume (order of 10 meters) is required, in which ionization will contribute to the final energy of muon before decays, but should be relatively minimal (MIP) for high energy muons with a sparse material (so that we may roughly assume that the total muon energy at the decay vertex is close to the input for GeV) - to be exact, universe or air may be the best material to test decay without loosing any enrgy.  No problem for leaving this test for later if it is difficult to isolate in-flight decay only as it will only check the boost effect for different input energies.  I do not think that we need to validate MC to experimental data (see for an example [here](https://www.researchgate.net/figure/Energy-spectrum-of-the-positron-in-Michel-decay-Points-data-Solid-histogram-Monte_fig3_237471229)) as the V-A calculation is a pure QED and should be (theoretically) good enough by itself (so, verification to Geant4 may be sufficient as Geant4 or other MC implementations are validated).

src/celeritas/decay/interactor/MuDecayInteractor.hh
sethrj: I think:
```suggestion
        p, norm(p) + ipow<2>(mass.value()))};
```

stognini: But this leads to $$E = p + m^2$$...?

sethrj: Sorry I misread the parens there! Silly me. But instead of `dot_product` and such, use `momentum.value()`!

PR #1457, opened 2024-10-18
Give priority to existing environment variables for G4 data location

pcanal:
This is necessary for proper operation when building and testing Celeritas
in a ATLAS build environment.  In the ATLAS build environment the Geant4
cmake fragment contains incorrect information about the location of the
data file (it points to the long gone build directory) so we need
to trust the environment variables instead.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -88,9 +88,11 @@ if(CELERITAS_USE_Geant4)
   # Optional dependence on low-energy EM data and neutron data
   set(CELERITASTEST_G4ENV)
   foreach(_ds G4EMLOW G4ENSDFSTATE G4PARTICLEXS)
-    list(APPEND CELERITASTEST_G4ENV
-      "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
-    )
+    if(NOT DEFINED ENV{${Geant4_DATASET_${_ds}_ENVVAR}})
+      list(APPEND CELERITASTEST_G4ENV
+         "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
+       )
+    endif()
   endforeach()
 endif()
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -93,7 +93,7 @@ if(CELERITAS_USE_Geant4)
          "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
        )
     else()
-      message(STATUS "Value for ${Geant4_DATASET_${_ds}_ENVVAR} taken from the enviroment rather than the Geant4 cmake fragment")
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -88,9 +88,13 @@ if(CELERITAS_USE_Geant4)
...
   foreach(_ds G4EMLOW G4ENSDFSTATE G4PARTICLEXS)
-    list(APPEND CELERITASTEST_G4ENV
-      "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
-    )
+    if(NOT DEFINED ENV{${Geant4_DATASET_${_ds}_ENVVAR}})
+      list(APPEND CELERITASTEST_G4ENV
+         "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
+       )
+    else()
+      message(STATUS "Value for ${Geant4_DATASET_${_ds}_ENVVAR} taken from the enviroment rather than the Geant4 cmake fragment")
```
pcanal: This prints
```
-- Value for G4LEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4ENSDFSTATEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4PARTICLEXSDATA taken from the environment rather than the Geant4 cmake fragment
```
at every (re)invocation of cmake.

test/CMakeLists.txt
```diff
@@ -88,9 +88,13 @@ if(CELERITAS_USE_Geant4)
...
   foreach(_ds G4EMLOW G4ENSDFSTATE G4PARTICLEXS)
-    list(APPEND CELERITASTEST_G4ENV
-      "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
-    )
+    if(NOT DEFINED ENV{${Geant4_DATASET_${_ds}_ENVVAR}})
+      list(APPEND CELERITASTEST_G4ENV
+         "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
+       )
+    else()
+      message(STATUS "Value for ${Geant4_DATASET_${_ds}_ENVVAR} taken from the environment rather than the Geant4 cmake fragment")
```
pcanal, line 96: This prints
```
-- Value for G4LEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4ENSDFSTATEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4PARTICLEXSDATA taken from the environment rather than the Geant4 cmake fragment
```
at every (re)invocation of cmake.

sethrj: I think that's fine. It's dependent only on the configure time environment rather than stored in cache: so it's appropriate to warn every time it might change, and it's a pretty exceptional circumstance. If it gets annoying we can turn it into a cmake cache and warn only the first time.

sethrj (approved):

sethrj (approved):

PR #1459, opened 2024-10-20
Define `id_cast` helper function

sethrj:
This adds a helper function that validates the argument against allowed values for an opaque ID. It particularly helps with conversions to/from signed integer from Geant4.


amandalund (approved):
Looks good @sethrj!

Issue #1248, opened 2024-05-22
Use logical rather than physical navigation state to reconstruct G4NavigationHistory


sethrj:
Related to https://github.com/celeritas-project/celeritas/issues/891: we need a higher accuracy way of reconstructing the Geant4 volume hierarchy from the Celeritas navigation state.

Here is how [AdePT reconstructs the navigation index](https://github.com/apt-sim/AdePT/blob/1b1370e62a8474365fd8a13b0b63de132475797b/src/AdePTGeant4Integration.cpp#L309-L352).
- Construct (recursively) a map of G4 physical volumes to VecGeom physical volumes. (This assumes the daughter list has the same ordering.) We should do this in g4vg.
- Copies the full nav state (with current depth) to CPU
- Loops over vecgeom nav state levels
- At each level, update a `G4NavigationHistory` to explicitly match the VecGeom PV IDs to the Geant4 PV IDs (use `BackLevel`, `NewLevel`, `SetFirstEntry`
- Call `touchable->UpdateYourself` with nav history

I added a touch of code in that direction: https://github.com/sethrj/celeritas/tree/navindex

- Change 'locate_touchable' to be part of the `selection.points` logic since we may need to do it both before and after. Similarly add `level_volume_instances` to `StepSelection`
- Add maximum level to geometry classes
- Add TrackView function to copy `VolumeInstanceId` at every level to a vector
- Add multi-volume level to step state data
- Implement `TouchableUpdater` to update based on `Span<VolumeInstanceId>`

PR #1460, opened 2024-10-20
Simplify geometry "volume" interface

sethrj:
In preparation for #1248 we will be adding `VolumeInstanceId` to universes. Rather than adding another set of multimap interfaces, let's just expose the `MultiMap` as the interface for volumes and surfaces. The volume instances will just be another multimap accessor.

### DIFF
--- a/src/corecel/cont/LabelIdMultiMap.hh
+++ b/src/corecel/cont/LabelIdMultiMap.hh
@@ -56,25 +56,39 @@ class LabelIdMultiMap
     // Empty constructor for delayed build
     LabelIdMultiMap() = default;
 
-    // Construct from a vector of label+sublabel pairs
-    explicit LabelIdMultiMap(VecLabel keys);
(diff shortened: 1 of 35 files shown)


src/corecel/cont/LabelIdMultiMap.hh
```diff
@@ -27,18 +27,21 @@ namespace celeritas
...
  * Duplicate labels are allowed but will be added to a list of duplicate IDs
  * that can be warned about downstream. Empty labels will be ignored.
  *
- * If no sublabels or labels are available, an empty span or "false" OpaqueId
- * will be returned.
+ * If no sublabels or labels are available for a \c find_X call, an empty span
+ * or "false" OpaqueId will be returned.
+ *
+ * The three kinds of \c find methods are named differently to avoid ambiguity:
+ * - \c find_all returns
```
amandalund: 🐿️

amandalund (approved):
Nice simplification, thanks @sethrj!

PR #1461, opened 2024-10-22
Add multilevel "volume instance" 

### LINKED ISSUE #1353: Define "volume instance" IDs and expose physical volumes for Geant4/VecGeom
I think for the optical physics (#1351 ) we may need to reconsider volume IDs in ORANGE because they're derived from the SCALE implementation/behavior, where every object is defined as a distinct "media" entry. Right now, volume IDs are somewhat like the "physical"/"placed" volumes in Geant4/VecGeom: copying a shape/solid results in two distinct IDs. However, for optical physics (and indeed for mapping in general to sensitive detectors with the new conversion layer), we may need to assign multiple volume IDs to the same "logical"/"unplaced" volume.

- Do we rename `VolumeId` to `VolumeInstanceId`?
- Do we add a new `UniqueVolumeId`/`ReusableVolumeId`/...?
- Do we change the name of `LocalVolumeId` based on whatever new names we come up with? (I'd say not...)
- Can logical IDs be arbitrarily set? (I think so?) and if so then we can define a one-to-one mapping with geant4.

Then implementing:
- The volume ID in ORANGE is still useful as a "placed" volume and we still will keep everything the same for indexing into the data.
- We'll need to add another mapping of instance ID -> unique volume ID to ORANGE
- We'll need to expose the physical volumes for Geant4/VecGeom for optical physics

CC @elliottbiondo @mrguilima

sethrj:
For #1248 we need to reconstruct the *hierarchy of physical volumes* instead of merely the current logical volume. This adds an abstraction, the "volume instance", corresponding to a physical volume ID, as well as the "level" (an index into the array of volume instance IDs).

Closes #1353 except for ORANGE geometry, which requires further lower-level implementation.


src/geocel/g4/GeantGeoTrackView.hh
```diff
@@ -85,9 +85,14 @@ class GeantGeoTrackView
     CELER_FORCEINLINE Real3 const& dir() const { return dir_; }
     //!@}
 
-    // Get the volume ID in the current cell.
-    CELER_FORCEINLINE VolumeId volume_id() const;
-    CELER_FORCEINLINE int volume_physid() const;
+    // Get the volume ID in the lowest level volume.
+    inline VolumeId volume_id() const;
```
mrguilima, line 89: Why was `CELER_FORCEINLINE` dropped for `inline` only?

src/geocel/vg/VecgeomParams.cc
```diff
@@ -468,34 +479,68 @@ void VecgeomParams::build_data()
...
+                // Get label
+                vecgeom::VPlacedVolume const* vol
+                    = vg_manager.FindPlacedVolume(vol_idx);
+                CELER_ASSERT(vol);
+
+                auto label = [vol] {
+                    std::string const& label = vol->GetLabel();
+                    if (starts_with(label, "[TEMP]"))
+                    {
+                        // Temporary logical volume not directly used in
```
mrguilima:
```suggestion
                        // Temporary placed volume not directly used in
```

src/geocel/Types.hh
```diff
@@ -36,15 +36,21 @@ using SquareMatrixReal3 = SquareMatrix<real_type, 3>;
...
 using GeoMaterialId = OpaqueId<struct GeoMaterial_>;
 
 //! Identifier for a surface (for surface-based geometries)
 using SurfaceId = OpaqueId<struct Surface_>;
 
-//! Identifier for a geometry volume
+//! Identifier for a geometry volume that may be repeated
 using VolumeId = OpaqueId<struct Volume_>;
 
+//! Identifier for an instance of a geometry volume (aka Phys/Placed)
```
mrguilima:
```suggestion
//! Identifier for an instance of a geometry volume (aka Physical/Placed)
```

src/geocel/g4/VisitGeantVolumes.hh
```diff
@@ -18,35 +19,80 @@ namespace celeritas
...
 /*!
- * Do a recursive depth-first listing of Geant4 logical volumes.
+ * Perform a breadth-first traversal of physical volumes.
  *
- * This will visit each volume exactly once based on when it's encountered in
- * the hierarchy. The visitor function F should have the signature
- * \code void(*)(G4LogicalVolume const&) \endcode .
+ * The function must have the signature
+ * <code>bool(*)(G4VPhysicalVolume const&, int)</code>
+ * where the return value indicates whether the volume's daughters should be
```
mrguilima:
```suggestion
 * where the return value indicates whether the volume's children should be
```

test/geocel/g4/VisitGeantVolumes.test.cc
```diff
@@ -0,0 +1,129 @@
...
+}
+
+//---------------------------------------------------------------------------//
+
+TEST_F(FourLevelsTest, physical)
+{
+    PhysicalVisitor visit;
+    visit_geant_volume_instances(visit, *this->geometry()->world());
+
+    static char const* const expected_names[] = {
```
mrguilima, line 103: Is there any requirement for breadth-first search in the geometry algorithms?  
For a simple geometry like fhis one (four-levels), there is a one-to-one relationship between `1:envX`, `2:Shape1` and `3:Shape2`, but that's not the case in most of the realistic detectors, and the depth-first approach is much easier to understand, I think.

mrguilima (approved):
This looks very nice!
Thanks, @sethrj

src/geocel/g4/GeantGeoTrackView.hh
sethrj: Ah, so the `CELER_FORCEINLINE` is most clear as a tag on the definition, not the declaration, since its use really depends on how small the implementation is. I'm working to make this more consistent in the code: the one-line (1-2 operation) functions should probably be forced inline, and larger functions might not need to be. This is most relevant for debug/reldeb builds since the full optimization mode should take care of that mostly.

src/geocel/vg/VecgeomParams.cc
sethrj: Actualy we don't make temporary "placed" volumes so this copy-pasted code shouldn't be needed at all. I've removed it.

test/geocel/g4/VisitGeantVolumes.test.cc
sethrj: The reason I did breadth-first was so that we can correctly count the maximum depth... but I think that's no longer necessary. I've added another example and converted to DFS.

sethrj:
@mrguilima Thanks for the review! Per the new guidelines, please resolve the conversations once you're satisfied.

test/geocel/g4/VisitGeantVolumes.test.cc
mrguilima: Thanks

mrguilima (approved):

PR #1463, opened 2024-10-23
Delete obsolete demonstration apps

sethrj:
Delete the old "geo check" and "KN interactor" apps. The former is made obsolete by `celer-geo` and other unit tests, and the latter hasn't been tested or run in a long time.


amandalund (approved):
Makes sense, thanks!

PR #1464, opened 2024-10-23
Improve error checking and diagnostic output for managing GPU

sethrj:
Following on to https://github.com/celeritas-project/celeritas/issues/1462 which seems in part to be caused by incompatible CUDA architectures affecting Thrust.

---

- Warn at configure time about missing `CMAKE_{CUDA|HIP}_ARCHITECTURES` *or* differences between the CMAKE and VecGeom_CMAKE architecture
- Warn at runtime if the device's capabilities aren't found in the compiled architectures
- A new `celer-sim --device` utility prints the device diagnostics.
- Add additional diagnostic output for the build configuration including the CUDA and Thrust version, as well as whether it's a static or shared build.

This adds Thrust as an explicit requirement to CUDA, and prints out the version as part of the build config. The ROCm 5.7 stack includes a "rocthrust" that doesn't directly translate to Thrust, so I have *not* included that. Including `rocthrust` also spits out a bunch of warnings and includes the `hip` targets which I think are obsolete anyway.

When configuring on `lxplus-gpu` without frameworks we now see:
```
CMake Warning at CMakeLists.txt:291 (message):
  The CMAKE_CUDA_ARCHITECTURES variable, necessary to specify the GPU target
  device, is not defined, and the default guess may be wrong.  Verify that
  this value matches your CUDA device capability.
```

And when running the executable:
```
$ ./bin/celer-sim --device
/tmp/seth/tmp.Owtw35TDvx/celeritas/src/corecel/sys/Device.cc:332: warning: Device 'Tesla T4' has CUDA compute capability of 75, but Celeritas was compiled with CMAKE_CUDA_ARCHITECTURES="52": code may mysteriously die at runtime
{
 "can_map_host_memory": true,
 "capability": 75,
 "capability_major": 7,
 "capability_minor": 5,
 "clock_rate": 1590000,
 "debug": false,
 "device_id": 0,
 "eu_per_cu": 1,
 "max_blocks_per_grid": 2147483647,
 "max_blocks_per_multiprocessor": 16,
 "max_cache_size": 4194304,
 "max_threads_per_block": 1024,
 "max_threads_per_cu": 1024,
 "memory_clock_rate": 5001000,
 "multiprocessor_count": 40,
 "name": "Tesla T4",
 "num_devices": 1,
 "platform": "cuda",
 "regs_per_block": 65536,
 "regs_per_multiprocessor": 65536,
 "shared_mem_per_block": 49152,
 "stream_async": true,
 "threads_per_warp": 32,
 "total_const_mem": 65536,
 "total_global_mem": 15642329088
}
```

On lxplus-gpu without any framework loads, running `./bin/celer-sim --config | jq .`:
```json
{
  "_category": "system",
  "_label": "build",
  "config": {
    "CELERITAS_BUILD_TYPE": "release,shared",
    "CELERITAS_CORE_GEO": "ORANGE",
    "CELERITAS_CORE_RNG": "xorwow",
    "CELERITAS_DEBUG": false,
    "CELERITAS_HOSTNAME": "lxplus900",
    "CELERITAS_REAL_TYPE": "double",
    "CELERITAS_UNITS": "CGS",
    "CELERITAS_USE_CUDA": true,
    "CELERITAS_USE_GEANT4": false,
    "CELERITAS_USE_HEPMC3": false,
    "CELERITAS_USE_HIP": false,
    "CELERITAS_USE_MPI": false,
    "CELERITAS_USE_OPENMP": true,
    "CELERITAS_USE_ROOT": false,
    "CELERITAS_USE_VECGEOM": false,
    "CUDA_VERSION": "12.4.131",
    "Thrust_VERSION": "2.3.2.0"
  },
  "version": "0.5.0-19+bedf4db8"
}
```

On frontier:
```json
{
  "_category": "system",
  "_label": "build",
  "config": {
    "CELERITAS_BUILD_TYPE": "release,static",
    "CELERITAS_CORE_GEO": "ORANGE",
    "CELERITAS_CORE_RNG": "xorwow",
    "CELERITAS_DEBUG": false,
    "CELERITAS_HOSTNAME": "frontier",
    "CELERITAS_REAL_TYPE": "double",
    "CELERITAS_UNITS": "CGS",
    "CELERITAS_USE_CUDA": false,
    "CELERITAS_USE_GEANT4": true,
    "CELERITAS_USE_HEPMC3": true,
    "CELERITAS_USE_HIP": true,
    "CELERITAS_USE_MPI": false,
    "CELERITAS_USE_OPENMP": true,
    "CELERITAS_USE_ROOT": false,
    "CELERITAS_USE_VECGEOM": false,
    "CLHEP_VERSION": "2.4.7.1",
    "Geant4_VERSION": "11.0.4",
    "HIP_VERSION": "5.7.23384"
  },
  "version": "0.5.0-19+bedf4db8e"
}
```


src/corecel/sys/Device.cc
```diff
@@ -317,6 +320,23 @@ void activate_device(Device&& device)
     CELER_DEVICE_CALL_PREFIX(SetDevice(device.device_id()));
     d = std::move(device);
 
+    // Check capability version against cmake variable; rough but better than
+    // nothing! CMake format: "70-real 72-virtual" or "35;50;72" or "
```
sethrj: ... "or gfx90a"

cmake/FindThrust.cmake
```diff
@@ -0,0 +1,18 @@
+#----------------------------------*-CMake-*----------------------------------#
+# Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+# See the top-level COPYRIGHT file for details.
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+#[=======================================================================[.rst:
+
+FindThrust
+----------
+
+Find the Thrust algorithm library for HIP or CUDA.
```
sethrj:
```suggestion
Find the Thrust algorithm library for CUDA.
```

pcanal:
> The CMAKE_CUDA_ARCHITECTURES variable, necessary to specify the GPU target
  device
  
An option might (or might not) be to use a default `CMAKE_CUDA_ARCHITECTURES=native` (available from CMake 3.24)

sethrj:
@pcanal Do you know what "native" means?

src/corecel/sys/Device.cc
pcanal: and currently we would issue a spurious message for the `gfx` case, right?

src/corecel/sys/Device.cc
```diff
@@ -317,6 +320,23 @@ void activate_device(Device&& device)
...
+    if (std::string(celeritas_gpu_architectures)
```
pcanal:
```suggestion
    if (std::string(celeritas_gpu_architectures).find("native") == std::string::npos
         && std::string(celeritas_gpu_architectures)
```

src/corecel/sys/Device.cc
pcanal:
```suggestion
    // nothing! CMake format: "native" or "70-real 72-virtual" or "35;50;72" or "
```

src/corecel/sys/Device.cc
```diff
@@ -317,6 +320,23 @@ void activate_device(Device&& device)
...
+            .find(std::to_string(d.capability()))
+        == std::string::npos)
```
esseivaju: Just as a note, compute capabilities are forward compatible within major versions, e.g. code for 8.0 runs on device with sm_8.6

pcanal:
>  Do you know what "native" means?

Yes: "Compile for the architecture(s) of the host's GPU(s)."  See https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html#prop_tgt:CUDA_ARCHITECTURES.  (And I checked that it works for me).  No clue what it does if there is no GPU though :)

src/corecel/sys/Device.cc
sethrj: No, we search for "90" in "gfx90a" so it works. Same with other mixed architectures

sethrj:
But does it something different than the default which always seems to be wrong?

src/corecel/sys/Device.cc
sethrj: "Runs on" 😒

pcanal:
> But does it something different than the default which always seems to be wrong?

Yes, or so it seems.  In my build, the code refuses to load if the PTX is not compiled for `80` and `native` runs fine.

drbenmorgan:
Looks fine from my CMake-hat perspective, but defer to @pcanal and @esseivaju as they're actually testing the details.

sethrj:
Thanks @drbenmorgan ! Please let me know if any of these warnings show up in the athena external builds.

sethrj:
Ping @pcanal

src/corecel/sys/Device.cc
pcanal: What happens if the user choose `native`?

pcanal: Nevermind, github confused me :)

CMakeLists.txt
```diff
@@ -283,9 +287,13 @@ if(BUILD_TESTING)
...
 
 if(CELERITAS_USE_CUDA)
-  # Use host compiler by default to ensure ABI consistency
-  set(CMAKE_CUDA_HOST_COMPILER "${CMAKE_CXX_COMPILER}" CACHE STRING
-    "Set to CMAKE_CXX_COMPILER by Celeritas CMakeLists")
+  if(NOT CMAKE_CUDA_ARCHITECTURES)
+    message(WARNING "The CMAKE_CUDA_ARCHITECTURES variable, necessary "
+      "to specify the GPU target device, is not defined, and the default "
+      "guess may be wrong.  Verify that this value matches your CUDA "
+      "device capability."
```
pcanal: For completeness, note that the default is actually the `NVCC` default (as confirmed by log examination and by the documentation at https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/).  As of `CUDA 12.6` the default is still `sm52`) [So technically `CMake` is guessing "*right*" but the `NVCC` default is inappropriate for any modern GPU]

CMakeLists.txt
```diff
@@ -383,6 +399,13 @@ if(CELERITAS_USE_VecGeom)
       CELERITAS_USE_CUDA
       "${VecGeom_CUDA_FOUND}"
     )
+  elseif(CELERITAS_USE_CUDA
+      AND NOT (VecGeom_CUDA_ARCHITECTURES STREQUAL CMAKE_CUDA_ARCHITECTURES))
```
pcanal, line 403: This is in practice too strict.  We only need to have a match.  I.e. `VecGeom` with '70;75' and `Celeritas` with `75` is okay. Technically we need to have `CMAKE_CUDA_ARCHITECTURES` being a subset of `VecGeom_CUDA_ARCHITECTURES`
something like
```
set(problem False)
foreach(_arch IN LISTS CMAKE_CUDA_ARCHITECTURES)
    string(REGEX MATCH "${CMAKE_CUDA_ARCHITECTURES}(;|$)" _arch_found "${VecGeom_CUDA_ARCHITECTURES}")
    if(NOT _arch_found) 
       set(problem True)
    endif()
endforeach()
```

sethrj: It might be, but who knows how RDC behaves? It would be better to warn if it's at all inconsistent for now.

pcanal (approved):

CMakeLists.txt
pcanal: Apriori if the `Celeritas' arch` is a subset of the other architectures then at link time it will bring with only the code for the matching architecture ... (Since to be true on the one example with the finicky `Thurst` kernel) ... but the warning message is lenient enough.

PR #1465, opened 2024-10-23
Refactor step storage to use aux state

sethrj:
This removes the "stream store" from the state data in preparation for refactoring the hit manager to use multi-level volume instance IDs.


src/celeritas/user/detail/StepParams.hh
```diff
@@ -0,0 +1,99 @@
...
+               GeoParams const& geo,
+               VecInterface const& interfaces);
+
+    //!@{
+    //! \name Aux interface
+    //! Short name for the aux data
+    std::string_view label() const final { return "detector-step"; }
+    //! Index of this class instance in its registry
+    AuxId aux_id() const final { return aux_id_; }
+    // Build optical core state data for a stream
```
amandalund:
```suggestion
    // Build state data for a stream
```

amandalund (approved):
LGTM, thanks @sethrj!

PR #1467, opened 2024-10-24
Add "cut" for individual track steps

sethrj:
This adds a step count cutoff (off by default) for individual tracks to try to assist with the ATLAS "stuck track" in the tile calorimeter.


src/celeritas/io/ImportParameters.hh
```diff
@@ -114,6 +114,8 @@ struct ImportTransParameters
     ImportLoopingMap looping;
     //! Maximum number of substeps in the field propagator
     int max_substeps{1000};
+    //! Optional global "step count" cutoff (use with care!): ignored if zero
+    int max_steps{0};
```
stognini: This step count option is not a Geant4 transportation parameter, right? Since we're not importing it, seems like it makes sense to only have it in our own `SetupOptions`. Or is this an option used in the ATLAS framework that we load in at run time?

sethrj: Yeahhhhh so the reason I did this is a little wacky, and probably should wait for future disccussion... whether we want to make the "import data" more Celeritas-centric and less geant-centric, making it the main way to provide input to our classes. Otherwise we get this:
```c++
    static std::shared_ptr<SimParams> from_import(ImportData const&,
                                                  SPConstParticles,
                                                  size_type max_steps,
                                                  size_type max_field_substeps);
```
which is error prone.

So maybe we need `SimParams::Input::from_import` and supplement it with extra data?

sethrj: For now I've implemented `SimParams::Input::from_import`.

src/accel/LocalTransporter.hh
```diff
@@ -102,7 +102,8 @@ class LocalTransporter
     UniqueEventId event_id_;
 
     size_type auto_flush_{};
-    size_type max_steps_{};
+    size_type max_step_iters_{};
+    double buffer_energy_{0};
```
stognini, line 106: Was this `real_type` or `double`? Still makes sense, store it as double in the cpu thread and cast to `real_type` on device, but wanted to make sure.

src/celeritas/io/ImportParameters.hh
stognini: Oh I see your point. I like the latest implementation. And agree about the `Import` discussion. Not sure what makes more sense. Maybe a celeritas-centric will allow for expanding the importing bit to have tailored parameters for different experiments.

stognini (approved):
Thanks! Looks good to me.

src/celeritas/track/SimTrackView.hh
```diff
@@ -120,6 +118,9 @@ class SimTrackView
     inline CELER_FUNCTION LoopingThreshold const&
         looping_threshold(ParticleId) const;
 
+    // Maximum number of tracks before killing the track
```
amandalund:
```suggestion
    // Maximum number of steps before killing the track
```

src/celeritas/track/SimTrackView.hh
```diff
@@ -412,5 +413,14 @@ SimTrackView::looping_threshold(ParticleId pid) const
     return params_.looping[pid];
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Maximum number of tracks before killing the track.
```
amandalund:
```suggestion
 * Maximum number of steps before killing the track.
```

src/celeritas/track/SimData.hh
```diff
@@ -58,7 +58,14 @@ struct LoopingThreshold
...
 /*!
- * Persistent simulation data.
+ * Shared simulation data.
+ *
+ * These are cutoff parameters based on the number of steps a track has taken.
+ * Currently these are global or per particle type (with a single energy cut);
+ * we should make them [energy, particle, region] for full extensibility.
+ *
+ * \note These params are used both by the main tracking loop \em and the
+ * SimTrackView in optical physics.
```
amandalund, line 68: Actually since #1438 these are no longer used for optical physics.

src/accel/LocalTransporter.cc
```diff
@@ -88,7 +88,7 @@ LocalTransporter::LocalTransporter(SetupOptions const& options,
                                    SharedParams& params)
     : auto_flush_(options.auto_flush ? options.auto_flush
                                      : options.max_num_tracks)
-    , max_steps_(options.max_steps)
+    , max_step_iters_(options.max_steps)
```
amandalund: Should this be:
```suggestion
    , max_step_iters_(options.max_step_iters)
```
?

amandalund (approved):
Looks good @sethrj! Were you going to add this as an option in our apps too?

src/accel/LocalTransporter.hh
sethrj: It's never used on the device, just for a message, so it probably doesn't matter...

src/celeritas/track/SimData.hh
sethrj: Ah yes. I was fooled by the optical tests failing after I introduced this change, but it's just because the cerenkov test uses the sim state.

PR #1468, opened 2024-10-24
Refactor pinned allocator to reduce need for templates

sethrj:
When I tried to add a new DetectorSteps (see #1248) I got a missing template instantation. I don't think we actually need to template the free/malloc calls itself? The only thing I can think of is whether operator delete is overloaded for a type, and we won't be able to pin such types anyway.


esseivaju (approved):
LGTM! It definitely makes sense not to have the malloc calls templated

PR #1469, opened 2024-10-24
Implement optical Rayleigh model

hhollenb:
Implements the optical Rayleigh model, similar to the optical absorption model.

Whereas absorption was relatively trivial, Rayleigh has material dependent data that needs to be accounted for. This model helps serve as a case for how more complicated models need to be imported and built by the rest of the optical physics framework.

Some design choices to note:
 - Expect all `ImportOpticalModel::mfp_table`s to be initialized with the same size as the number of optical materials. The values may be empty physics vectors, indicating the model is responsible for building its own MFP vector from material parameters. Having all tables be the same size allow intermixing imported MFP vectors and ones built from material parameters, with it explicit which ones should be built.
 - Vectors of `ImportOpticalProperties` and `ImportOpticalMaterials` are copied to the input of `RayleighModel`. This is similar to how I've seen other `Input` struct get initialized, but it seems like a pattern where we transpose the `std::vector<ImportOpticalMaterial>` into vectors of its members (and potentially a lot of copying?). Similar to `ImportedModels` it might make sense to have an `ImportedMaterials` with some adapter to iterate over its members.


src/celeritas/ext/GeantImporter.cc
```diff
@@ -564,6 +564,7 @@ import_optical(detail::GeoOpticalIdMap const& geo_to_opt)
         get_property(&optical.rayleigh.compressibility,
                      "ISOTHERMAL_COMPRESSIBILITY",
                      ImportUnits::len_time_sq_per_mass);
+        optical.rayleigh.temperature = material->GetTemperature();
```
sethrj: Instead of this, we can pull it from the MaterialView when we construct our optical materials.

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        build(calculated_mfp);
+    }
+    else
+    {
+        CELER_LOG(warning)
+            << "Could not construct optical Rayleigh MFP table for "
+               "optical material "
+            << mat.get() << " since its imported data was invalid";
```
sethrj: We discussed at one point how to treat missing data, but I think we should error out if there are optical materials with missing data. The user should either turn off the model or explicitly set the cross sections to zero for that material.

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -0,0 +1,116 @@
...
+    else if (auto const& rayl = rayleigh_[mat.get()])
+    {
+        ImportPhysicsVector const& r_index
+            = properties_[mat.get()].refractive_index;
+
+        constexpr real_type hbarc = hbar_planck * c_light;
+
+        // Einstein-Smoluchowski formula
+        real_type scale_factor = rayl.scale_factor;
+        real_type beta_t = rayl.compressibility;
```
sethrj: The constructor needs to validate these,

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -0,0 +1,116 @@
...
+
+            CELER_ASSERT(n > 1);
+
+            real_type c2 = ipow<4>(energy / hbarc);
+            real_type c3 = ipow<2>((ipow<2>(n) - 1) * (ipow<2>(n) + 2) / 3);
+
+            CELER_ASSERT(c2 > 0);
+            CELER_ASSERT(c3 > 0);
+
+            calculated_mfp.y[i] = 1 / (c1 * c2 * c3);
```
sethrj: This should be done in a separate `RayleighMfpCalculator` class, using `optical::MaterialView::make_refractive_index_calculator()`. I think the class declaration should probably be:
```c++
class RayleighMfpCalculator
{
  public:
    //!@{
    //! \name Type aliases
    using Energy = units::MevEnergy;
    //!@}

  public:
    inline CELER_FUNCTION RayleighMfpCalculator(MaterialView const& mat,
                                                real_type scale_factor,
                                                real_type compressibility);

    inline CELER_FUNCTION real_type operator()(Energy) const;

  private:
    GenericCalculator calc_rindex_;
    real_type c1_;
};
```

There shouldn't be a hard requirement to calculate the rayleigh cross sections on the same grid as the rindex (though that is likely the sensible way to start).

src/celeritas/optical/model/RayleighModel.hh
```diff
@@ -0,0 +1,59 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using SPConstImported = std::shared_ptr<ImportedModels const>;
+    //!@}
+
+    struct Input
+    {
+        std::vector<ImportOpticalProperty> properties;
+        std::vector<ImportOpticalRayleigh> rayleigh;
```
sethrj: These are redundant with the imported data right? We should also be using the already-processed `optical::MaterialParams` rather than the raw import: so we ought to provide the optical material shared pointer.

src/celeritas/ext/GeantImporter.cc
hhollenb: Removed temperature from `ImportOpticalRayleigh` and made a separated `OpticalRayleighMaterial` struct to store all the parameters for calculating the MFP. Since this data isn't used by the interactor, I stored it in `MaterialParams` instead of adding it to `MaterialData`, but can change that easily if need be.

src/celeritas/optical/ImportedModelAdapter.cc
```diff
@@ -49,12 +49,10 @@ ImportedModels::ImportedModels(std::vector<ImportOpticalModel> models)
...
                        << model_id << "'");
-        CELER_VALIDATE(
-            std::all_of(model.mfp_table.begin(),
-                        model.mfp_table.end(),
-                        [](auto const& v) { return static_cast<bool>(v); }),
-            << "Import optical model id '" << model_id << "' ("
-            << to_cstring(model.model_class) << ") has invalid MFP vectors");
+
+        // Model MFP vectors may be empty, indicating the model should attempt
+        // to build from optical material data.
```
amandalund, line 54: So for a given model it's possible that for some materials the MFP vector will be imported, and for other materials it will need to be built, correct?

src/celeritas/optical/MaterialParams.hh
```diff
@@ -79,8 +81,32 @@ class MaterialParams final : public ParamsDataInterface<MaterialParamsData>
 
   private:
     CollectionMirror<MaterialParamsData> data_;
+
+    std::vector<OpticalRayleighMaterial> rayleigh_materials_;
```
amandalund: These properties are only used by the Rayleigh model and only needed on the host for builiding the MFPs (not used after the model construction), right? Instead of storing them in the material params, maybe we should pass them directly as input to the model?

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,133 @@
...
+ */
+RayleighMfpCalculator::RayleighMfpCalculator(
+    MaterialView material, OpticalRayleighMaterial const& rayleigh_material)
+    : calc_rindex_(material.make_refractive_index_calculator())
+    , density_fluctuation_(
+          rayleigh_material.scale_factor * rayleigh_material.compressibility
+          * rayleigh_material.temperature * celeritas::constants::k_boltzmann
+          / (6 * celeritas::constants::pi))
+{
+    CELER_EXPECT(density_fluctuation_ > 0);
```
amandalund, line 112:
```suggestion
    CELER_EXPECT(density_fluctuation_ > 0);
    CELER_EXPECT(rayleigh_material);
```

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,133 @@
...
+    //!@{
+    //! \name Type aliases
+    using Energy = units::MevEnergy;
+    using Grid = typename GenericCalculator::Grid;
+    //!@}
+
+  public:
+    // Construct from material and Rayleigh properties
+    inline CELER_FUNCTION
+    RayleighMfpCalculator(MaterialView material,
```
amandalund:
```suggestion
    RayleighMfpCalculator(MaterialView const& material,
```

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,133 @@
...
+    CELER_EXPECT(density_fluctuation_ > 0);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the optical Rayleigh mean free path at the given energy.
+ */
+CELER_FUNCTION real_type RayleighMfpCalculator::operator()(Energy energy) const
+{
+    CELER_EXPECT(value_as<Energy>(energy) > 0);
```
amandalund:
```suggestion
    CELER_EXPECT(energy > zero_quantity());
```

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,133 @@
...
+
+    constexpr real_type hbarc = celeritas::constants::hbar_planck
+                                * celeritas::constants::c_light;
+
+    real_type rindex = calc_rindex_(native_value_from(energy));
+    CELER_ASSERT(rindex > 1);
+
+    real_type wave_number = native_value_from(energy) / hbarc;
```
amandalund: We could use the `energy_to_wavelength()` helper function in OpticalUtils.hh for this (multiplied by the 2 pi factor).

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,133 @@
...
+ * Calculate the optical Rayleigh mean free path at the given energy.
+ */
+CELER_FUNCTION real_type RayleighMfpCalculator::operator()(Energy energy) const
+{
+    CELER_EXPECT(value_as<Energy>(energy) > 0);
+
+    constexpr real_type hbarc = celeritas::constants::hbar_planck
+                                * celeritas::constants::c_light;
+
+    real_type rindex = calc_rindex_(native_value_from(energy));
```
amandalund: is `native_value_from` correct here, or should this be `value_as<Energy>(energy)` (the grid is in MeV, right)?

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -0,0 +1,100 @@
...
+                                   std::vector<double>(energy_grid.size())};
+
+        for (auto i : range(energy_grid.size()))
+        {
+            result.x[i] = energy_grid[i];
+            result.y[i]
+                = calc_mfp(native_value_to<units::MevEnergy>(result.x[i]));
+        }
+
+        build(result);
```
amandalund: Perhaps instead of creating an `ImportPhysicsVector` here we could use the `GenericGridInserter::operator()` that takes two spans, and just create a vector of the calculated mfps? And if we added a method to get a span of all the grid values from the `NonuniformGrid`, something like:
```c++
template<class T>
CELER_FUNCTION auto NonuniformGrid<T>::values() const -> SpanT
{
    return storage_[offset_];
}
```
then we wouldn't have to copy the grid values either.

amandalund (commented):
Great work @hhollenb!

src/celeritas/optical/ImportedModelAdapter.cc
hhollenb: Yes - they can be mixed. I'm not familiar with how actual experiments provide data for Rayleigh scattering, but Geant4 will attempt to use `kRAYLEIGH` mean free paths if provided, and will then attempt to build them from material parameters if they're not present, so this will be consistent with that behavior.

My logic is that all imported model data should have a MFP vector for every optical material, where empty ones indicate that no data has been imported for that material and the model should attempt to build it. Because MFP vectors are present for every material, I think it makes explicit which are imported vs which need to be built. Individual models can then validate if they can import or build MFPs for every optical material.

src/celeritas/optical/model/RayleighModel.hh
```diff
@@ -0,0 +1,57 @@
...
+    // Execute the model with host data
+    void step(CoreParams const&, CoreStateHost&) const final;
+
+    // Execute the model with device data
+    void step(CoreParams const&, CoreStateDevice&) const final;
+
+  private:
+    ImportedModelAdapter imported_;
+    SPConstMaterials materials_;
+    std::vector<OpticalRayleighMaterial> rayleigh_materials_;
```
amandalund: I don't think we need to store the `rayleigh_materials_` since they're only used during construction, right?

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -0,0 +1,100 @@
...
+    , materials_(std::move(materials))
+    , rayleigh_materials_(std::move(rayleigh_materials))
+{
+    CELER_EXPECT(materials_);
+    CELER_EXPECT(materials_->num_materials() == imported_.num_materials());
+    CELER_EXPECT(materials_->num_materials() == rayleigh_materials_.size());
+
+    for (auto mat : range(OpticalMaterialId(materials_->num_materials())))
+    {
+        CELER_EXPECT(imported_.mfp(mat) || rayleigh_materials_[mat.get()]);
```
amandalund: We should `VALIDATE` this input instead.

src/celeritas/optical/model/RayleighMfpCalculator.cc
```diff
@@ -0,0 +1,58 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct Rayleigh material data from imported data and core Material
+ * parameters.
+ */
+std::vector<OpticalRayleighMaterial> OpticalRayleighMaterial::from_import(
+    std::vector<ImportOpticalRayleigh> const& imported,
+    ::celeritas::MaterialParams const& mat)
```
amandalund: It might make more sense to construct the `RayleighModel` from the imported data and core material params and get rid of the `OpticalRayleighMaterial` altogether.  The mfp calculator could also be constructed with the core `MaterialView` (and the optical properties); then we could just get the temperature from the view and wouldn't have duplicate the other data.

src/celeritas/optical/model/RayleighModel.hh
hhollenb: Currently they're used by `Model::build_mfp` in the `build_mfp` step in `PhysicsParams` which is after construction of the models. Since optical physics has only one particle type we might be able to merge constructing the model and building the MFPs into a single call in the `PhysicsParams` constructor, but I'd have to take a look at rewriting that.

src/celeritas/optical/model/RayleighModel.cc
hhollenb: Is it ok to do asserts in a loop like this, or should I have the whole check in something like
```
CELER_VALIDATE(std::all_of(...), << "...");
```
so the predicate is a single statement? I assume loops are fine I just haven't seen them for simple assertions in the code I've touched.

src/celeritas/optical/model/RayleighModel.hh
amandalund: Ah that's right, we build them later.

src/celeritas/optical/model/RayleighMfpCalculator.cc
hhollenb: My one concern about this was making optical and core physics too interdependent, in case we want to be able to run the optical loop completely independently of the core loop. For example, in running tests we'd have to also pass some mocked version of the core `MaterialParams` to `RayleighModel`.

That being said, what might work best is to instead have 2 constructors for `RayleighModel`, one of which can take the core material params to optionally build missing MFPs, and the other to just require imported MFPs and can run independently of any core data. Thoughts?

hhollenb: It does mean if someone wanted to run the optical loop and construct the Rayleigh MFP tables from material data, then they'd also have to initialize the core material params with just the temperatures for each corresponding optical material. Not sure how much of a concern this would be in practice.

src/celeritas/optical/model/RayleighModel.cc
amandalund: `CELER_VALIDATE` is an always-on check, so this is fine. For assertions that are enabled with `CELERITAS_DEBUG` we can always do something like [this](https://github.com/celeritas-project/celeritas/blob/3269beff167334e28890838e7d9d9638341f015d/src/celeritas/em/model/detail/LivermoreXsInserter.hh#L67-L76).

src/celeritas/optical/model/RayleighMfpCalculator.cc
sethrj: ~~Yes the material temperature should be part of the core material params—if we don't care about the other materials, and assuming we're not doing doppler broadening for neutrons, we can just set their temperatures to zero.~~
I see what your'e saying now, but I think we're always going to build the core geometry and core materials. We need to map the geometry's material IDs to the optical material IDs, after all.

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,156 @@
...
+    real_type temperature{0};  //!< Material temperature
+
+    operator bool() const
+    {
+        return scale_factor > 0 && compressibility > 0 && temperature > 0;
+    }
+
+    static std::vector<OpticalRayleighMaterial>
+    from_import(std::vector<ImportOpticalRayleigh> const& imported,
+                ::celeritas::MaterialParams const& mat);
```
sethrj: This is unusual to have in a "runtime calculation" file. Since we don't have a `RayleighData.hh` file and that might be overkill, I think it's fine to keep the `OpticalRayleighMaterial` struct here. But could you make the `from_import` part of the model, and then add a convenience function to get the rayleigh data from the model for a given optical material ID?

src/celeritas/optical/model/RayleighMfpCalculator.hh
```diff
@@ -0,0 +1,156 @@
...
+ * \f]
+ * The final equation for the MFP in terms of energy:
+ * \f[
+    l^{-1} = \frac{k_B T \beta_T}{6\pi} \left(\frac{E}{\hbar c}\right)^4 \left[
+ \frac{(\varepsilon - 1)(\varepsilon + 2)}{3} \right]^2.
+ * \f]
+ *
+ * The scale factor is a unitless user customizable factor that's multiplied
+ * to the inverse MFP.
+ */
```
sethrj, line 60: Great documentation! Can you add
```
.. doxygenclass:: celeritas::optical::RayleighMfpCalculator
```
under the volumetric processes of `doc/implementation/optical-physics.rst`?

src/celeritas/optical/model/RayleighModel.cc
sethrj: Yep, `CELER_VALIDATE` inside the loop is great for checking input data.

sethrj (approved):
@hhollenb This looks great, thanks! I'll wait for the other approvers to weigh in. Can you also `git checkout upstream/develop -- test/celeritas/data` to remove the incidental changes to the ROOT files?

src/celeritas/optical/model/RayleighMfpCalculator.hh
amandalund: To clarify, by "runtime calculation" do you mean not setup time? I think this calculator/struct is only ever used during setup for the mfp construction and only inside the Rayleigh model, right? What would the convenience function be for?

sethrj (approved):
Looks great! Thanks for the excellent work.

src/celeritas/optical/model/RayleighMfpCalculator.hh
sethrj: I was thinking ahead toward if we wanted to generate cross sections (mfps) on the fly, we could use this calculator. I assumed the `OpticalRayleighMaterial` might be used for testing? (If not, no worries)

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -19,27 +20,73 @@ namespace celeritas
...
     SPConstMaterials materials,
-    std::vector<OpticalRayleighMaterial> rayleigh_materials)
+    SPConstCoreMaterials core_materials,
+    std::vector<ImportOpticalRayleigh> import_rayleigh_materials)
     : Model(id, "optical-rayleigh", "interact by optical Rayleigh")
     , imported_(ImportModelClass::rayleigh, imported)
     , materials_(std::move(materials))
-    , rayleigh_materials_(std::move(rayleigh_materials))
+    , core_materials_(std::move(core_materials))
+    , import_rayleigh_materials_(std::move(import_rayleigh_materials))
```
amandalund: Since we'll be getting the `ImportOpticalRayleigh` vector directly from the `ImportData` now, I'm not sure we should be moving it... maybe store a shared pointer to it?

amandalund (approved):
Excellent, thanks @hhollenb!

src/celeritas/optical/model/RayleighModel.cc
hhollenb: `ImportData` has a vector of `ImportOpticalMaterial`s which have `ImportOpticalRayleigh` as members. So a vector of `ImportOpticalRayliegh` would have to be made separately, like in `optical::MaterialParams` for `ImportOpticalProperty`.

I've started drafting an `ImportedMaterials` class, similar to `ImportedModels`, to pass around as a shared pointer since it seems like many classes want vectors of a specific `ImportOptical...`.

amandalund: Makes sense!

PR #1470, opened 2024-10-25
Tweak Perfetto integration

sethrj:
Follow-on to #1238. I was trying to find the perfetto counter and didn't expect `trace_counter` to be defined in a `Counter.hh` (which sounds like it should be like `itertools.count`) so I moved the name to `TraceCounter.hh`. I also removed the *definition* of the perfetto class when the library isn't available; instead I use an extern deleter.


esseivaju (approved):
Looks good!

PR #1471, opened 2024-10-26
Add multi-level state to detector step result

sethrj:
This is the next step toward #1248 : it allows us to gather and return the volume instance ID at every level, on GPU and CPU. Currently the test only works as a demonstration for Geant4, since I haven't yet added the VecGeom->G4PV mapping, nor ORANGE->VolumeInstanceId.


src/celeritas/user/detail/StepGatherExecutor.hh
```diff
@@ -132,12 +145,44 @@ StepGatherExecutor<P>::operator()(celeritas::CoreTrackView const& track)
...
+        {
+            auto dst = [this,
+                        &vid = this->state.data.points[P].volume_instance_ids,
+                        tid = track.track_slot_id()] {
+                // Destination size
+                size_type const size = this->params.volume_instance_depth;
+                size_type offset = tid.unchecked_get() * size;
+                auto all_ids = vid[AllItems<VolumeInstanceId>{}];
+                return all_ids.subspan(offset,
+                                       this->params.volume_instance_depth);
```
amandalund:
```suggestion
                return all_ids.subspan(offset, size);
```

src/celeritas/user/DetectorSteps.cu
```diff
@@ -123,6 +70,26 @@ void copy_field(DetectorStepOutput::vector<T>* dst,
...
+    copy(MemSpace::device, {src.data().get(), num_valid});
+}
+
+//---------------------------------------------------------------------------//
+template<class T>
+void copy_field(DetectorStepOutput::vector<T>* dst,
+                ItemRef<T> const& src,
+                size_type num_valid,
+                size_type per_thread,
+                StreamId stream)
```
amandalund: This should be resizing/copying `num_valid * per_thread`, right?

src/celeritas/user/detail/StepGatherExecutor.hh
```diff
@@ -132,12 +146,44 @@ StepGatherExecutor<P>::operator()(celeritas::CoreTrackView const& track)
...
+            }();
+
+            // Fill every level from the geometry
+            size_type depth = geo.level().unchecked_get() + 1;
+            CELER_ASSERT(depth <= dst.size());
+            geo.volume_instance_id(dst.first(depth));
+            if constexpr (CELERITAS_DEBUG)
```
amandalund, line 166: Now that we have a separate configure option for device debug assertions, should we be checking `CELERITAS_DEVICE_DEBUG` in addition to `CELERITAS_DEBUG` in case only the former is enabled? Is there any reason we wouldn't want debug to always be on when device debug is? (Wondering about this more generally since there are several other places we do this in host/device code, e.g. [killing stuck tracks](https://github.com/celeritas-project/celeritas/blob/5a0178e78d8a4b6b68db1ab5890653a3bdea7f12/src/celeritas/global/alongstep/detail/PropagationApplier.hh#L113-L142).)

amandalund (approved):
Nice, thanks @sethrj! All looks good to me.

src/celeritas/user/detail/StepGatherExecutor.hh
sethrj: Oh that's a really good point... should we require `CELERITAS_DEBUG=ON` to enable `CELERITAS_DEVICE_DEBUG`? Or define `CELER_DEBUG = (CELERITAS_DEBUG && !CELER_DEVICE_COMPILE)  || (CELERITAS_DEVICE_DEBUG && CELER_DEVICE_COMPILE)`? I'll update this use case and perhaps we can address the broader issue in a follow-on?

amandalund: I think requiring `CELERITAS_DEBUG` to enable device debug makes sense... I can't think of a good reason for having only the latter enabled.

PR #1472, opened 2024-10-28
Escape semicolons in exported CMake variables

drbenmorgan:
Noted in ATLAS builds that the export of `CMAKE_CUDA_ARCHITECTURES` to the `Config.hh` and `CeleritasConfig.cmake` files resulted in newlines replacing the semicolons when this variable is a proper CMake list. For example, running CMake as

```
$ cmake -DCMAKE_CUDA_ARCHITECTURES="50;61;72;80" ...
```

as in ATLAS builds created a `Config.hh` with:

```c++
inline constexpr char celeritas_gpu_architectures[] = "50
61
75
80";
```

and thus a compile-time error. Similarly, `CeleritasConfig.cmake` got

```cmake
set(CELERITAS_CMAKE_CUDA_ARCHITECTURES "50
61
75
80")
```

Assuming that we want to retain these as `"X;Y;Z"`, the changes here explicitly escape any semicolons in variables exported to CMake files to ensure their retention, and also the CUDA/HIP architecture variable exported to `Config.hh` to retain them and prevent compilation error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -631,7 +631,9 @@ elseif(CELERITAS_USE_HIP)
   set(_celer_other_cmake_variables CMAKE_HIP_ARCHITECTURES)
 endif()
 foreach(_key IN LISTS CELERITAS_DEFAULT_VARIABLES _celer_other_cmake_variables)
-  list(APPEND CELERITAS_EXPORT_VARIABLES "set(CELERITAS_${_key} \"${${_key}}\")")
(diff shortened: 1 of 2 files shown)


pcanal (approved):

sethrj (approved):
Ugh, good catch @drbenmorgan . As a guard against invalid C++ strings we could also wrap them with `R"({var})"` in case they have quotes or whatever else as well.

See https://gitlab.kitware.com/cmake/cmake/-/issues/23315 .

PR #1475, opened 2024-10-29
Override and save Geant4 dataset variables at configure time

pcanal:
This is necessary for proper operation when building and testing Celeritas in a ATLAS build environment.  In the ATLAS build environment the Geant4 cmake fragment contains incorrect information about the location of the data file (it points to the long gone build directory) so we need to trust the environment variables instead.

This is essential the same as fe52da7ebaadce8f6fd3199a95490d048094c18a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,18 +323,18 @@ if(CELERITAS_USE_Geant4 AND NOT Geant4_FOUND)
       "consider setting CELERITAS_OPENMP to \"event\" or disabling OpenMP"
     )
   endif()
-   # Set environment variables from Geant4-exported configuration
+  # Set environment variables from Geant4-exported configuration
   if(NOT DEFINED CELER_G4ENV)
     set(_CELER_G4ENV)
     foreach(_ds IN LISTS Geant4_DATASETS)
       if(NOT DEFINED ENV{${Geant4_DATASET_${_ds}_ENVVAR}})
         list(APPEND _CELER_G4ENV
-           "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
+          "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
         )
       else()
         message(STATUS "Set ${Geant4_DATASET_${_ds}_ENVVAR}=$ENV{${Geant4_DATASET_${_ds}_ENVVAR}} from the environment")
         list(APPEND _CELER_G4ENV
-           "${Geant4_DATASET_${_ds}_ENVVAR}=$ENV{${Geant4_DATASET_${_ds}_ENVVAR}}}"
+          "${Geant4_DATASET_${_ds}_ENVVAR}=$ENV{${Geant4_DATASET_${_ds}_ENVVAR}}}"
         )
       endif()
     endforeach()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,7 +332,7 @@ if(CELERITAS_USE_Geant4 AND NOT Geant4_FOUND)
            "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
         )
       else()
-        message(STATUS "Value for ${Geant4_DATASET_${_ds}_ENVVAR} taken from the environment rather than the Geant4 cmake fragment")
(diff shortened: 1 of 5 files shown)


sethrj:
Should we do this at a global level in our Findgeant4 module?

pcanal:
It might make sense.  I noted that the 2 set of variable is slightly different:
```
-- Value for G4LEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4ENSDFSTATEDATA taken from the environment rather than the Geant4 cmake fragment
-- Value for G4PARTICLEXSDATA taken from the environment rather than the Geant4 cmake fragment
```
vs
```
-- For apps value for G4NEUTRONHPDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4LEDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4LEVELGAMMADATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4RADIOACTIVEDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4PARTICLEXSDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4PIIDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4REALSURFACEDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4SAIDXSDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4ABLADATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4INCLDATA taken from the environment rather than the Geant4 cmake fragment
-- For apps value for G4ENSDFSTATEDATA taken from the environment rather than the Geant4 cmake fragment
```

sethrj:
Yeah the unit tests only use a subset of the data. The app runs coupled hadronic so it needs all of it. Since this is a fix up to the Geant4 cmake data would you add code to update it there and save in cache?

pcanal:
`to update it there and save in cache?`

Do you mean to then use the same set of environment variables (i.e. for all the data sets) for both the `test` and `app` cases?

sethrj:
I mean overwriting and saving the geant4 dataset variables that are loaded through Geant config

sethrj:
I'd like to have our unit test code look like:
```cmake
  foreach(_ds G4EMLOW G4ENSDFSTATE G4PARTICLEXS)
    list(APPEND CELERITASTEST_G4ENV
      "${Celeritas_DATASET_${_ds}_ENVVAR}=${Celeritas_DATASET_${_ds}_PATH}"
     )
  endforeach()
```
i.e. replacing Geant4->Celeritas. In the helper macro to be called, we can just use `set(${Celeritas_DATASET_${_ds}_ENVVAR} ${Geant4_DATASET_${_ds}_ENVVAR}})`, and then during the first configure we can save `Celeritas_DATASET_${_ds}_PATH` based on the environment variable, or empty to indicate that it'll be pulled from Geant4. We probably want the latter behavior (rather than copying and caching) because the user could reconfigure again with a different g4 version if they got it wrong the first time, for example.

CMakeLists.txt
```diff
@@ -323,6 +323,23 @@ if(CELERITAS_USE_Geant4 AND NOT Geant4_FOUND)
...
+   # Set environment variables from Geant4-exported configuration
+  if(NOT DEFINED CELER_G4ENV)
+    set(_CELER_G4ENV)
+    foreach(_ds IN LISTS Geant4_DATASETS)
+      if(NOT DEFINED ENV{${Geant4_DATASET_${_ds}_ENVVAR}})
+        list(APPEND _CELER_G4ENV
+           "${Geant4_DATASET_${_ds}_ENVVAR}=${Geant4_DATASET_${_ds}_PATH}"
+        )
+      else()
+        message(STATUS "Value for ${Geant4_DATASET_${_ds}_ENVVAR} taken from the environment rather than the Geant4 cmake fragment")
```
sethrj:
```suggestion
        message(STATUS "Set ${Geant4_DATASET_${_ds}_ENVVAR}=$ENV{${Geant4_DATASET_${_ds}_ENVVAR}} from the environment")
```

CMakeLists.txt
```diff
@@ -323,6 +323,23 @@ if(CELERITAS_USE_Geant4 AND NOT Geant4_FOUND)
       "consider setting CELERITAS_OPENMP to \"event\" or disabling OpenMP"
     )
   endif()
+   # Set environment variables from Geant4-exported configuration
```
sethrj: This and a few other places have some funny spacing 🙃

sethrj (approved):
Couple of superficial comments after we got through the discussion of how to handle the override. Thanks!

PR #1477, opened 2024-11-01
Use kernel launchers instead of macros

sethrj:
This is a minor refactor to try to see if the behavior of #1462 changes: since the failure is currently the first execution of the CUDA "launch kernel" wrapper, let's see if it happens earlier during the RNG reseeding. I've updated a couple of the two remaining kernels to use the wrapper, and also added a stream to the CuHip RNG stream initialization.


src/celeritas/random/detail/RngReseedExecutor.hh
```diff
@@ -0,0 +1,82 @@
...
+#include "celeritas/random/RngData.hh"
+#include "celeritas/random/RngEngine.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Brief class description.
```
amandalund: Missing description.

amandalund (approved):
Thanks @sethrj!

PR #1478, opened 2024-11-01
Store edge bounding boxes on BIH and refactor for upcoming intersection capabilities

elliottbiondo:
This MR is the first work toward using BIH acceleration for background_intersect. First, "edge" bounding boxes are stored during construction. An edge bounding box bounds all child nodes of an edge. It is formed by clipping a infinite bounding box with every bounding plane between the root and the edge (inclusive). During intersect operations these bounding boxes are used to determine if traversal should continue down an edge.

Originally, we planned on having a single BIHTraversal class for both point-in-vol and intersection operations, with functors used to test edges and leaves for enclosure/intersection. I started doing it this way, but things got messy for several reasons. With point-in-vol, traversal stops as soon as a suitable volume is found. With intersection operations, traversal continues to find the minimum over the full tree, which also requires storing the current minimum. In addition, point-in-vol requires only a pos, whereas intersection requires a pos and dir. The abstractions required to shoe-horn this all into a single class are really not worth it. 

The approach I am pursuing involves separate `BIHEnclosingVolFinder` and `BIHIntersectingVolFinder` classes, which share a common `BIHTraversalHelper` class, employed via composition. This MR refactors the current point-in-vol code into this format, with another MR in process for the intersection version. I might be able to roll a few more methods into `BIHTraversalHelper` with functors, but that will be in the next MR.


elliottbiondo:
@sethrj this is RTR

src/orange/detail/BIHData.hh
```diff
@@ -24,37 +24,43 @@ namespace detail
...
 struct BIHInnerNode
 {
     using real_type = fast_real_type;
 
-    struct BoundingPlane
+    struct Edge
     {
-        real_type position;
+        //! The position of the bounding plane along the partition axis
+        real_type bounding_plane_pos;
```
sethrj: As a habit, default initialize (I missed this when it was first put in)
```suggestion
        real_type bounding_plane_pos{};
```

src/orange/detail/BIHData.hh
```diff
@@ -24,37 +24,43 @@ namespace detail
...
+        //! The child node connected to this edge
         BIHNodeId child;
+        //! Bbox created by clipping an inf bbox with the bounding planes
+        //! between this edge (inclusive) and the root.
+        FastBBox bbox;
     };
 
-    enum class Edge
+    enum class Side
```
sethrj, line 50: You could also use `enum class Bound` (geocel/Types) lo/hi instead of this enum... might be more generic than left/right which imply x axis?

src/orange/detail/BIHBuilder.cc
```diff
@@ -131,19 +133,31 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
...
+        node.edges[Side::left].bounding_plane_pos = left_pos;
+        node.edges[Side::left].bbox = get_shrunk_bbox(Bound::hi, left_pos);
+
+        auto right_pos = p.bboxes[Side::right].lower()[to_int(p.axis)];
+        node.edges[Side::right].bounding_plane_pos = right_pos;
+        node.edges[Side::right].bbox = get_shrunk_bbox(Bound::lo, right_pos);
 
         // Recursively construct the left and right branches
-        for (auto edge : range(Edge::size_))
+        for (auto edge : range(Side::size_))
```
sethrj:
```suggestion
        for (auto side : range(Side::size_))
```
maybe? Since you might have `Edge& edge = node.edges[side]`?

src/orange/detail/BIHEnclosingVolFinder.hh
```diff
@@ -3,26 +3,30 @@
...
 namespace celeritas
 {
 namespace detail
 {
 //---------------------------------------------------------------------------//
 /*!
- * Traverse BIH tree using a depth-first search.
+ * Traverse the BIH tree to find a volume that contains a point.
  *
- * \todo move to top-level orange directory out of detail namespace
```
sethrj, line 23: This to-do still applies... I think the traverser should be in `univ` or the top level (or maybe we should try to reorganized based on runtime/configure time). It shouldn't be in `orange/detail` since it's used by ../univ/SimpleUnitTracker`

src/orange/detail/BIHEnclosingVolFinder.hh
```diff
@@ -89,31 +81,31 @@ class BIHTraverser
  * Construct from vector of bounding boxes and storage.
  */
 CELER_FUNCTION
-BIHTraverser::BIHTraverser(BIHTree const& tree,
-                           BIHTraverser::Storage const& storage)
-    : tree_(tree), storage_(storage), leaf_offset_(tree.inner_nodes.size())
+BIHEnclosingVolFinder::BIHEnclosingVolFinder(
+    BIHTree const& tree, BIHEnclosingVolFinder::Storage const& storage)
```
sethrj:
```suggestion
    BIHTree const& tree, Storage const& storage)
```

src/orange/detail/BIHData.hh
```diff
@@ -24,37 +24,43 @@ namespace detail
...
- * boundary of the right side of the tree. Since the halfspaces of created by
- * the bounding planes may overlap, the LEFT bounding plane position could be
- * either left or right of the RIGHT bounding plane position.
+ * As a convention, a node's LEFT edge corresponds to the half space that is
+ * less than the partition value. In other words, the LEFT bounding plane
+ * position is the far right boundary of the left side of the tree, and the
+ * RIGHT bounding plane position is the far left boundary of the right side of
+ * the tree. Since the halfspaces created by the bounding planes may overlap,
+ * the LEFT bounding plane position could be either left or right of the RIGHT
+ * bounding plane position.
```
sethrj, line 33: Another (less ambiguous) way to think about this is that the *center* of the left/lower region is always less than the *center* of the right/upper half-space?

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+    // Construct from vector of bounding boxes and storage for LocalVolumeIds
+    inline CELER_FUNCTION
+    BIHTraversalHelper(BIHTree const& tree, Storage const& storage);
+
+    // Determine if a node is inner, i.e., not a leaf
+    inline CELER_FUNCTION bool is_inner(BIHNodeId id) const;
+
+    // Get an inner node for a given BIHNodeId
+    inline CELER_FUNCTION BIHInnerNode const&
+    get_inner_node(BIHNodeId id) const;
```
sethrj: We avoid `get_` for accessors: just use `inner_node` and update the similar values in this file.

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Storage = NativeCRef<BIHTreeData>;
+    //!@}
+
+    // Construct from vector of bounding boxes and storage for LocalVolumeIds
+    inline CELER_FUNCTION
+    BIHTraversalHelper(BIHTree const& tree, Storage const& storage);
```
sethrj: Consider `BihView` as an alternative name for this class?

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+    inline CELER_FUNCTION size_type get_num_inf_volids() const;
+
+    // Get the ith vol_id in inf_volids
+    inline CELER_FUNCTION LocalVolumeId get_inf_volid(size_type i) const;
+
+  private:
+    //// DATA ////
+    BIHTree const& tree_;
+    Storage const& storage_;
+    size_type leaf_offset_;
```
sethrj: I don't think it's necessary to cache this (using another register to do so): just have a helper function `size_type leaf_offset() const`.

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+BIHLeafNode const& BIHTraversalHelper::get_leaf_node(BIHNodeId id) const
+{
+    CELER_EXPECT(!this->is_inner(id));
+    return storage_
+        .leaf_nodes[tree_.leaf_nodes[id.unchecked_get() - leaf_offset_]];
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ *  Get a leaf node for a given BIHNodeId.
```
sethrj:
```suggestion
 * Get the bounding box of a local volume.
```

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    // Get the vol_id of the ith volume on a given leaf node
+    inline CELER_FUNCTION LocalVolumeId get_leaf_volid(BIHLeafNode leaf,
+                                                       size_type i) const;
+
+    // Get the number of inf_volids on the tree
+    inline CELER_FUNCTION size_type get_num_inf_volids() const;
+
+    // Get the ith vol_id in inf_volids
+    inline CELER_FUNCTION LocalVolumeId get_inf_volid(size_type i) const;
```
sethrj: I think `Span<LocalVolumeId> inf_volids() const` would be more straightforward: these seem to be used only in a loop.

src/orange/detail/BIHTraversalHelper.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    // Get a leaf node for a given BIHNodeId
+    inline CELER_FUNCTION BIHLeafNode const& get_leaf_node(BIHNodeId id) const;
+
+    // Get a bbox for a given vol_id
+    inline CELER_FUNCTION FastBBox const& get_bbox(LocalVolumeId vol_id) const;
+
+    // Get the vol_id of the ith volume on a given leaf node
+    inline CELER_FUNCTION LocalVolumeId get_leaf_volid(BIHLeafNode leaf,
+                                                       size_type i) const;
```
sethrj: Maybe here also you can return a span to the leaf volumes.
```suggestion
    inline CELER_FUNCTION Span<LocalVolumeId const> leaf_volumes(BIHLeafNode const& leaf) const;
```
(note also you should be using const references to the leaf nodes since they're larger than a pointer)

src/orange/detail/BIHData.hh
elliottbiondo: These are in reference to the left/right edges of the tree itself, not what side of the bounding plane the point is on. Using `Bound` still makes sense semantically as the left side of the tree is the side corresponding to the "low" side of the bounding plane. However, since all of the comments in this class refer to the left/right edges (consistent with tree terminology used in the literature), I think there is value in keeping this enum for readability.

src/orange/detail/BIHData.hh
elliottbiondo: I don't think the term "center" is ideal here since these are infinite half spaces. I think what I have is fairly explicit, and I think it's important to state it in these terms even if it is a bit wordy. In what way do you think it is ambiguous?

src/orange/detail/BIHEnclosingVolFinder.hh
elliottbiondo: Good point on SimpleUnitTracker. I will add the to-do back and we can remove things around as a separate MR.

src/orange/detail/BIHTraversalHelper.hh
elliottbiondo: Good point; I was thinking these where not contiguous in memory, but they actually are.

elliottbiondo:
@sethrj this is ready for another look.

Also, I kind of want to put in another MR that goes through ORANGE and replaces all `vol_id` with `volid` or vice versa, because the inconsistency is getting annoying. Which do you prefer?

src/orange/detail/BIHData.hh
sethrj: They're not really infinite since they're the union of all the bounding boxes of all the daughter nodes, right? (All the infinite bboxes are excluded.) I wasn't suggesting you take out any of the text, which does make sense, but thought it could be another way of thinking of the tree layout.

sethrj (approved):
Looks great! nice refactor. I think a consistency refactor would be great: I'm happy with the shortened form `volid`, `vol_id`, or `vol`.

sethrj:
@elliottbiondo Once you "resolve" https://github.com/celeritas-project/celeritas/pull/1478#discussion_r1829255615 this should merge! 😄

src/orange/detail/BIHData.hh
elliottbiondo: @sethrj there are two separate concepts here: bounding planes, and the edge bounding boxes. Both of which are found on every edge.

When traversing the tree for point-in-vol,  the decision to traverse an edge is only contingent on the bounding plane of the edge. Bounding planes always create infinite half spaces.

When traversing the tree for intersections, the decision to traverse an edge is only contingent on the pre-computed edge bounding boxes. Edge bounding boxes are not the union on the bounding boxes of the daughter nodes --- that would create something more like a BVH. Edge bounding boxes can be thought of as "what we know so far" as the tree is traversed. They are the intersection of half spaces created by all bounding planes between the root and the current edge. Many of the edge bounding boxes are infinite in a least one direction. In fact, the leftmost and rightmost edges at every depth, i.e. LL...L and RR...R, are always infinite on the + or - side of every direction, because the represent the case that is always on the + or - side of every bounding plane.

PR #1479, opened 2024-11-02
Extend BIH capabilities for intersection operations

elliottbiondo:
This MR creates the `BIHIntersectingVolFinder` class which is used to find the surface that a ray intersects with first, and the corresponding distance to this surface. This is done using a traversal strategy that is distinct from `BIHEnclosingVolFinder`, as is documented in their respective top-level doc strings. In a subsequent MR, this will be "hooked up" for use in background volumes.


elliottbiondo:
@sethrj this is ready. Probably better to review this one first, then I will do the vol_id + others one.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using Storage = NativeCRef<BIHTreeData>;
+
+    struct Ray
+    {
+        Real3 pos;
+        Real3 dir;
+    };
```
sethrj, line 53: I don't think this is a good idea; it might force a copy of the data and increases register pressure. If anything, make the data members `Real3 const& ` and pass `Ray` by value.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+        if (!view_.is_inner(current_node))
+        {
+            intersection = this->visit_leaf(
+                view_.leaf_node(current_node), ray, intersection, visit_vol);
+        }
+
+        previous_node = exchange(
+            current_node,
+            this->next_node(
+                current_node, previous_node, ray, intersection.distance));
```
sethrj, line 154: Nice use of `exchange`! 😄

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+ * current edge (inclusive). If a ray's intersection with the edge bbox is
+ * found to be nearer than the current minimum intersection, traversal procedes
+ * down that edge. Likewise, when a root node is reacted, intersections with
+ * volume bboxes are first tested against the minimum intersection prior to
+ * testing the the volume itself. The minimum intersection is only modified
+ * when a nearer minimumium intersection with a actual volume if found, NOT a
+ * nearer intersection with an edge bbox or volume bbox. This is because is is
+ * possible to have a ray that interects with a volume's bbox, but not the
+ * volume itself.
+ *
```
sethrj, line 39: Can you add a brief synopsis of the expected form of the `F` template function? (i.e. `real_type(*)(VolumeId)`, it should return the distance to the intersection with the volume.)

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+ * Calculate the current min intersection, which may/may not be in inf_vols.
+ */
+template<class F>
+CELER_FUNCTION auto
+BIHIntersectingVolFinder::visit_inf_vols(Intersection min_intersection,
+                                         F&& visit_vol) const -> Intersection
+{
+    for (auto id : view_.inf_volids())
+    {
+        auto intersection = visit_vol(id);
```
sethrj: Since we can often reduce the cost of an intersection by passing a "maximum search distance", (i.e. if we know the closest volume is nearby we don't need to evaluate logic if the nearest surface is further than that) I think the visit function should include a "minimum" as an argument, just like the SimpleUnitTracker intersect does.

sethrj (commented):
Still working on this, but gotta take the day off and here's what I have so far.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+    else
+    {
+        // Leaf node; return to parent
+        CELER_EXPECT(previous_id == view_.leaf_node(current_id).parent);
+        next_id = previous_id;
+    }
+
+    return next_id;
```
sethrj: I've waffled over the years about the clarity of having a single "return" versus a longer scope and if/else, but I think this function would read better without the local `next_id` and just:

```c++
    if (!view_.is_inner(current_id))
    {
        CELER_EXPECT(previous_id == view_.leaf_node(current_id).parent);
        return previous_id;
    }

    auto const& current_node = view_.inner_node(current_id);
    auto const& l_edge = current_node.edges[Side::left];
    auto const& r_edge = current_node.edges[Side::right];

    if (previous_id == current_node.parent)
    {
        // Visiting this inner node for the first time; go down either left
        // or right edge
        return this->visit_bbox(l_edge.bbox, ray, min_dist) ? l_edge.child
                                                            : r_edge.child;
    }

    if (previous_id == current_node.edges[Side::left].child)
    {
        // Visiting this inner node for the second time; go down right edge
        // or return to parent
        return this->visit_bbox(r_edge.bbox, ray, min_dist)
                   ? r_edge.child
                   : current_node.parent;
    }

    // Visiting this inner node for the third time; return to parent
    CELER_ASSERT(previous_id == current_node.edges[Side::right].child);
    next_id = current_node.parent;
```

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,260 @@
...
+
+  private:
+    //// DATA ////
+    BIHView view_;
+
+    //// HELPER FUNCTIONS ////
+
+    // Get the ID of the next node in the traversal sequence
+    inline CELER_FUNCTION BIHNodeId next_node(BIHNodeId const& current_id,
+                                              BIHNodeId const& previous_id,
```
sethrj: Since NodeID are POD , it's better to pass by value (and that way the compiler can save some work trying to figure out if the IDs are aliased or not).

sethrj (commented):
One final comment is to consider whether it would be better to split this class into `class BihIntersectFinder` and `template<class F> class BihIntersectingVolFinder`, which I think would have a more sensible interface. (If we do that we should also add a note in `BIHEnclosingVolFinder` that we could update its interface in the future since I know that's what we based this on.)

src/orange/detail/BIHIntersectingVolFinder.hh
elliottbiondo: I believe you are the progenitor of this line via a comment within one of the original BIH merge requests 😄

src/orange/detail/BIHIntersectingVolFinder.hh
elliottbiondo: I added this to the `operator()` doc string

elliottbiondo:
@sethrj this is ready for another look.

I don't understand what you mean by splitting this into  `class BihIntersectFinder` and `template<class F> class BihIntersectingVolFinder`. I don't think there will ever be any other form of `BihIntersectFinder` other than the volume version.

sethrj:
> I don't understand what you mean by splitting this into `class BihIntersectFinder` and `template<class F> class BihIntersectingVolFinder`. I don't think there will ever be any other form of `BihIntersectFinder` other than the volume version.

I just mean this as an implementation detail: half the functions are templated on F (which is probably better as a class template parameter) and the other half aren't (so they probably shouldn't be part of a templated class). One solution would be to move the non-templated to an "implementation base class" and then template the class on F.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,264 @@
...
+    };
+    //!@}
+
+    // Construct from a vector of bounding boxes and storage for LocalVolumeIds
+    inline CELER_FUNCTION
+    BIHIntersectingVolFinder(BIHTree const& tree, Storage const& storage);
+
+    // Calculate the minimum intersection
+    template<class F>
+    inline CELER_FUNCTION Intersection operator()(Ray ray, F&& visit_vol) const;
```
sethrj, line 69: I think this should have a `min_dist` as well.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,264 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Determine if the intersection with an edge/vol bbox is less than min_dist.
+ */
+CELER_FUNCTION
+bool BIHIntersectingVolFinder::visit_bbox(FastBBox const& bbox,
+                                          Ray ray,
+                                          double min_dist) const
```
sethrj:
```suggestion
                                          real_type min_dist) const
```

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -0,0 +1,264 @@
...
+    // Get the ID of the next node in the traversal sequence
+    inline CELER_FUNCTION BIHNodeId next_node(BIHNodeId current_id,
+                                              BIHNodeId previous_id,
+                                              Ray ray,
+                                              double min_dist) const;
+
+    // Determine if the intersection with an edge/vol bbox is less than
+    // min_dist
+    inline CELER_FUNCTION bool
+    visit_bbox(FastBBox const& bbox, Ray ray, double min_dist) const;
```
sethrj: Missed a couple of `double`s here, they should be `real_type`.

sethrj (approved):
Looks good after you add the last `min_dist` and change `double` to `real_type`. Thanks!

sethrj (approved):

PR #1482, opened 2024-11-04
Improve logger when using either CUDA or MPI

sethrj:
- Add a "null logger" so that the logger can accept (but ignore) anything passed to it. I intend to port this to VecGeom to suppress the "invalid use of `__host__` function in HD code" warnings.
- Remove dependence of logger on MPI. This should fix a link error @stognini saw and a build error I saw.


src/corecel/io/Logger.hh
```diff
@@ -44,17 +49,24 @@
...
  * Like \c CELER_LOG but for code paths that may only happen on a single
- * process. Use sparingly.
+ * process or thread. Use sparingly.
  */
 #define CELER_LOG_LOCAL(LEVEL)                        \
     ::celeritas::self_logger()(CELER_CODE_PROVENANCE, \
                                ::celeritas::LogLevel::LEVEL)
 
+// Allow CELER_LOG to be present (but ignored) in device code
+#if CELER_DEVICE_COMPILE
```
pcanal, line 63: This is following a different pattern than the other uses of `CELER_DEVICE_COMPILE` where it is more often, in place, `#if CELER_DEVICE_COMPILE ... do something ... #else .... do something else ... #endif`.  You may want to augment the commit log to give a slight justification of why the code in the PR is the better choice here.  (i.e. the alternative would be the self_logger thingy to have embedded the `nop` `device` functions)

sethrj: Yeah I wrote it like this because the use case here is "well, in this weird case let's suppress the logger so that we can at least compile".

> the self_logger thingy to have embedded the nop device functions

Not sure what that means; but if you're talking about using the self/world functions, they would have to be host/device and return a mutable reference, I don't think that's the right approach.

pcanal: I was unprecise.  What I meant is that the pattern we have elsewhere is to have the alternative inside `MessageLogger` itself and to have for example in `LoggerMessage.hh`
```
#if CELER_DEVICE_COMPILE
    //! Do not set any state
    CELER_FORCEINLINE_FUNCTION void setstate(std::ostream::iostate) {}
#else
    // Update the stream state
    inline void setstate(std::ostream::iostate state);
#endif
```
Both solution `works` but since the code in this PR deviates a bit from the norm, I think we should sprinkle the commit log with a vague explanation for that deviation :)

sethrj: We don't have that pattern elsewhere, I'm not sure it works for HIP, and it definitely won't work via the host-only global `Logger& world_logger()` function, for example...

sethrj:
@pcanal note that I extended the scope of the PR to refactor the logger internals not to have any MPI usage: instead the default setters are what determines whether the local process has a handler or a "null" handler. I think the previous iteration wasn't right, and it was hard to tell because of the structure.

src/corecel/io/Logger.hh
pcanal: > We don't have that pattern elsewhere,

Humm ... right ... instead we have the `#if CELER_DEVICE_COMPILE` inside the functions body.  So the difference here could be "Replace whole class" instead of "modifying all functions" to adapt to device compilation.

>  and it definitely won't work via the host-only global Logger& world_logger() function, for example...

Fair enough (i.e. this is an additional note for the commit log :) )

test/corecel/io/Logger.test.cc
```diff
@@ -20,8 +21,22 @@
 
 namespace celeritas
 {
+namespace detail
+{
 namespace test
 {
+TEST(NullLoggerMessage, all)
+{
+    NullLoggerMessage() << "This should not print anything " << range(10)
```
pcanal, line 30: Does this actually test that nothing is printed? (I.e. do we check the standard output against a references?)

src/corecel/io/Logger.hh
```diff
@@ -44,33 +49,44 @@
...
  *
  * This object \em is assignable, so to replace the default log handler with a
  * different one, you can call \code
-   world_logger = Logger(celeritas::comm_world(), my_handler);
+   world_logger = Logger(my_handler);
  * \endcode
+ *
+ * When using with MPI, the \c world_logger global objects are different on
+ * each process: rank 0 will have a handler that outputs to screen, and the
+ * other ranks will have a "null" handler that suppresses all log output.
```
pcanal, line 89: Is that temporary or a restriction we have to live with?  I.e. with additional work in future PRs can we record the output for the other ranks or is there already another mechanism for us to see error messages that have happened in the other ranks?

pcanal (approved):

src/corecel/io/Logger.hh
sethrj: No, this is just the default behavior for `world_logger`. Anyone can assign a new `Logger` object to it on any process, so you could have it (e.g.) log error messages of a certain level on non-rank-0 nodes.

test/corecel/io/Logger.test.cc
sethrj: All the tests in this file are manually verified. This one in particular, though, shouldn't even compile if the null logger is working correctly: the `Range` object doesn't have an `operator<<`.

pcanal: > shouldn't even compile if the null logger is working correctly

That sounds bad/inverted ... it read that the build will fail to compile when things are right and compile and pass the test when things are wrong ... I must be missing something :)

sethrj: Right sorry, commenting at the middle of the night. It should fail to compile if it's not working :)

pcanal: Does the test runs on the device? (if not we would need to add such a test to check the main purpose of the `NullLoggerMessage`)

sethrj: I tested manually; it wouldn't be much of a test if we had one. The "no failure when streaming an unstreamable object" should be test enough...

pcanal: Don't we also want the automatic testing of 'nvcc not complaining about device code calling host code' not appearing?

PR #1483, opened 2024-11-04
Refactor "touchable updater" that uses a navigator

sethrj:
This is the penultimate PR for #1248. It refactors the touchable updater into a generic interface that works for both ORANGE and VecGeom. A new updater (the next pull request) will use the volume instance ID to reconstruct the touchable, rather than a navigator.

This also unconditionally enables updating of some pre-step LV properties which should be independent from the touchable state.

### DIFF
--- a/src/accel/detail/HitManager.cc
+++ b/src/accel/detail/HitManager.cc
@@ -96,7 +96,8 @@ HitManager::HitManager(GeoParams const& geo,
  * Create local hit processor.
  *
  * Due to Geant4 multithread semantics, this \b must be done on the same CPU
- * thread on which the resulting processor used.
+ * thread on which the resulting processor used. It must be done once per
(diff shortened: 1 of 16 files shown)


src/accel/detail/HitManager.cc
```diff
@@ -96,25 +96,18 @@ HitManager::HitManager(GeoParams const& geo,
  * Create local hit processor.
  *
  * Due to Geant4 multithread semantics, this \b must be done on the same CPU
- * thread on which the resulting processor used.
+ * thread on which the resulting processor used. It must be done once per
+ * thread and can be done separtely.
```
amandalund:
```suggestion
 * thread and can be done separately.
```

amandalund (approved):
Looks good @sethrj!

PR #1486, opened 2024-11-05
Replace "volid" with "vol_id" throughout ORANGE

elliottbiondo:
This MR updates all files within ORANGE to use the form "vol_id" for variable names instead of "volid", for consistency with other opaque IDs such as `universe_id`, `transform_id`, `level_id`, etc.

### DIFF
--- a/src/orange/OrangeData.hh
+++ b/src/orange/OrangeData.hh
@@ -152,12 +152,12 @@ struct OrientedBoundingZoneRecord
 
     // Transformation from the OBZ coordinate system to the unit coordinate
     // system
-    TransformId transform_id;
+    TransformId trans_id;
 
     //! True if assigned
     explicit CELER_FUNCTION operator bool() const
     {
-        return offset_ids[0] && offset_ids[1] && transform_id;
+        return offset_ids[0] && offset_ids[1] && trans_id;
     }
 };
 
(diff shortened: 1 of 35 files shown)


elliottbiondo:
@sethrj RTR

sethrj (commented):
Looks great. Do you want to do the same for univ_id, trans_id, ...? I think it would be a nice touch. Also like we have `uid` in there, there are a couple of `vid` for `vol_id`.

elliottbiondo:
Yeah I can do that too. I wasn't sure how far too go. I am thinking:

surf_id,
trans_id,
univ_id,

but only for local variable names. If I start changing APIs (geo.volume_id() to geo.vol_id()) the changes become more invasive. I also don't want to mess but the .jsons.

sethrj:
Yes, agreed. I think it's ok (and often better?) for API names to be spelled out. In fact, I read/saw recently a guideline that said the more global an identifier's scope, the longer/more descriptive the name should be; and conversely very local variables can be very short (think `i` in an inner loop, or `v` for volume id in a 3-line loop).

sethrj:
@elliottbiondo Were you planning to update the other IDs before merging this, or should we defer that?

elliottbiondo:
@sethrj yes, I will update all of these. It will probably be easier to start from fresh main branch at this point. You can close this if you'd prefer for me to start a new MR rather than force-pushing here.

sethrj:
I think we can make an exception and force-push 😉 if you wanted to be really fancy you could make a new temporary branch with `git fetch upstream; git checkout -b temp upstream/develop`; make your global upates there, and then return to the original branch with `git checkout vol_id` and make a "pretend" conflict-free merge by
```sh
git reset --hard $(git commit-tree -p HEAD -p upstream/develop -m "Merge upstream" temp^{tree})
```
Then you'd be more easily able to see if you accidentally reverted something.

elliottbiondo:
@sethrj I think this is ready to review, but I am not sure if the Docker build failure is real?

sethrj:
The "failure" is just because when a PR is draft, it skips docker builds to save compute time. There *is* a "ready to review" button at the bottom when it's in draft form. I'll update to develop and that'll trigger the full CI.

sethrj (approved):
Thanks!

PR #1487, opened 2024-11-06
Fix unnecessary linking when using VecGeom+CUDA

pcanal:
Due to missing initialization and missing an intermediary variable, dependencies found by cuda_rdc_cuda_gather_dependencies where bleeding from one target to the next.


sethrj (approved):
Before:
opaqueid-test-orig

After:
opaqueid-test-new

Thanks @pcanal !

PR #1490, opened 2024-11-07
Fix dBVH memcpy size

esseivaju:
Introduced in #1481, copy the entire pointer instead of 1 byte.


pcanal (approved):
Thanks for finding this!

esseivaju:
Missed that in the review, better late than never 😅

pcanal:
Does it fix the crash related to `dBVH`?

sethrj:
This was only for the diagnostic. Sorry I didn't notice it!

esseivaju:
right, it's only for the diagnostic, and that line is still showing up in a full static build of AthSimulation+AthExternals

PR #1491, opened 2024-11-07
Allow core and optical loops to have different numbers of track slots

amandalund:
This lets us set the number of track slots in the optical stepping loop to something different from the core loop.

### DIFF
--- a/app/celer-sim/Runner.cc
+++ b/app/celer-sim/Runner.cc
@@ -598,15 +598,19 @@ void Runner::build_optical_collector(RunnerInput const& inp,
     }
     CELER_ASSERT(inp.optical);
 
+    size_type num_streams = core_params_->max_streams();
+
     OpticalCollector::Input oc_inp;
     oc_inp.material = MaterialParams::from_import(
         imported, *core_params_->geomaterial(), *core_params_->material());
     oc_inp.cerenkov = std::make_shared<CerenkovParams>(oc_inp.material);
     oc_inp.scintillation
         = ScintillationParams::from_import(imported, core_params_->particle());
-    oc_inp.buffer_capacity = inp.optical.buffer_capacity;
-    oc_inp.primary_capacity = inp.optical.primary_capacity;
-    oc_inp.auto_flush = inp.optical.auto_flush;
+    oc_inp.num_track_slots = ceil_div(inp.optical.num_track_slots, num_streams);
+    oc_inp.buffer_capacity = ceil_div(inp.optical.buffer_capacity, num_streams);
+    oc_inp.initializer_capacity
+        = ceil_div(inp.optical.initializer_capacity, num_streams);
+    oc_inp.auto_flush = ceil_div(inp.optical.auto_flush, num_streams);
 
     CELER_ASSERT(oc_inp);
     optical_collector_
--- a/src/celeritas/optical/detail/CerenkovGeneratorAction.cc
+++ b/src/celeritas/optical/detail/CerenkovGeneratorAction.cc
@@ -107,7 +107,8 @@ void CerenkovGeneratorAction::step_impl(CoreParams const& core_params,
                    << "insufficient capacity (" << initializers_size
                    << ") for optical photon initializers (total capacity "
                       "requirement of "
-                   << num_photons + num_new_photons << ")");
+                   << num_photons + num_new_photons << " and current size "
+                   << num_photons << ")");
 
     auto& offload = offload_state.store.ref();
     auto& buffer_size = offload_state.buffer_size.cerenkov;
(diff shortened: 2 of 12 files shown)


app/celer-sim/Runner.cc
```diff
@@ -598,15 +598,19 @@ void Runner::build_optical_collector(RunnerInput const& inp,
...
     OpticalCollector::Input oc_inp;
     oc_inp.material = MaterialParams::from_import(
         imported, *core_params_->geomaterial(), *core_params_->material());
     oc_inp.cerenkov = std::make_shared<CerenkovParams>(oc_inp.material);
     oc_inp.scintillation
         = ScintillationParams::from_import(imported, core_params_->particle());
-    oc_inp.buffer_capacity = inp.optical.buffer_capacity;
-    oc_inp.primary_capacity = inp.optical.primary_capacity;
-    oc_inp.auto_flush = inp.optical.auto_flush;
+    oc_inp.num_track_slots = ceil_div(inp.optical.num_track_slots, num_streams);
```
amandalund, line 609: I suppose we should divide the optical track/buffer sizes by the number of streams too?

src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -107,19 +113,18 @@ std::string_view OpticalLaunchAction::description() const
...
-                                       StreamId sid,
-                                       size_type size) const -> UPState
+auto OpticalLaunchAction::create_state(MemSpace m, StreamId sid, size_type) const
+    -> UPState
 {
     if (m == MemSpace::host)
     {
         return std::make_unique<optical::CoreState<MemSpace::host>>(
-            *optical_params_, sid, size);
+            *optical_params_, sid, num_track_slots_);
```
amandalund: This may not be the best way to set the state size... another option might be to store the size in each of the aux params classes (although this should be the only case where the size is different?).

app/celer-sim/Runner.cc
sethrj: Well, unless we go to a "weak scaling" model, but we should do that across all applications and parameters (so now's not the time to change one thing).

src/celeritas/optical/detail/OpticalLaunchAction.cc
sethrj: I think it'd be fine to store in the launcher or in the `optical_params_`.

src/celeritas/optical/detail/OpticalLaunchAction.hh
```diff
@@ -60,15 +60,17 @@ class OpticalLaunchAction : public AuxParamsInterface,
...
 
     // Construct with IDs, core for copying params, offload gen data
     OpticalLaunchAction(ActionId id,
                         AuxId data_id,
                         CoreParams const& core,
                         SPConstMaterial material,
                         SPOffloadParams offload,
-                        size_type primary_capacity);
+                        size_type num_track_slots,
+                        size_type iniitializer_capacity);
```
sethrj: To control for the growing number of parameters, perhaps the material through `initializer_capacity` (note there's a double `i` typo in the commented line) should be in an `Input` helper struct. Then you could reuse it with `make_and_insert` and just pass it into the constructor.

sethrj (commented):
Still working on this, but gotta take the day off and here's what I have so far.

app/celer-sim/Runner.cc
amandalund: We're already doing it for the core parameters in celer-sim, so should I keep it for these optical parameters too, or no? At some point we had decided to do this in celer-sim and not celer-g4, though I can't remember the reasoning now...

sethrj: Let's keep it as consistent as possible for now? 😅 though it sounds like it might be too late...

src/celeritas/optical/detail/CerenkovGeneratorAction.cc
```diff
@@ -123,6 +124,9 @@ void CerenkovGeneratorAction::step_impl(CoreParams const& core_params,
     // Generate the optical photon initializers from the distribution data
     this->generate(core_params, core_state);
 
+    CELER_LOG(debug) << "Generated " << count << " Cerenkov photons from "
```
sethrj:
```suggestion
    CELER_LOG_LOCAL(debug) << "Generated " << count << " Cerenkov photons from "
```

src/celeritas/optical/detail/ScintGeneratorAction.cc
```diff
@@ -121,6 +122,9 @@ void ScintGeneratorAction::step_impl(CoreParams const& core_params,
     // Generate the optical photon initializers from the distribution data
     this->generate(core_params, core_state);
 
+    CELER_LOG(debug) << "Generated " << count << " Scintillation photons from "
```
sethrj:
```suggestion
    CELER_LOG_LOCAL(debug) << "Generated " << count << " Scintillation photons from "
```

sethrj (approved):
Looks great, thanks @amandalund !

PR #1492, opened 2024-11-09
Use canonical loop form with OpenMP

amandalund:
OpenMP requires the for loop following a directive to be in [canonical loop form](https://www.openmp.org/spec-html/5.2/openmpsu28.html#x58-600004.4.1), which does not support multiple variables in the initialization. Versions before 5.0 also do not allow `!=` in the termination condition.

### DIFF
--- a/src/celeritas/global/ActionLauncher.hh
+++ b/src/celeritas/global/ActionLauncher.hh
@@ -43,10 +43,11 @@ void launch_core(std::string_view label,
                  F&& execute_thread)
 {
     MultiExceptionHandler capture_exception;
+    size_type const size = state.size();
 #if defined(_OPENMP) && CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
 #    pragma omp parallel for
 #endif
-    for (size_type i = 0; i < state.size(); ++i)
+    for (size_type i = 0; i < size; ++i)
     {
         CELER_TRY_HANDLE_CONTEXT(
             execute_thread(ThreadId{i}),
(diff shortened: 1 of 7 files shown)


src/celeritas/global/ActionLauncher.hh
```diff
@@ -46,7 +46,7 @@ void launch_core(std::string_view label,
 #if defined(_OPENMP) && CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
 #    pragma omp parallel for
 #endif
-    for (size_type i = 0, size = state.size(); i != size; ++i)
+    for (size_type i = 0; i < state.size(); ++i)
```
sethrj: Should we still try to precalculate and cache the size outside the loop? Regular C++ can't tell that `primaries` won't change size over the course of the loop; can OpenMP?? Does it actually allow for the parallelism to change inside the loop (hard), or does it require that the loop bounds don't change (easy), or something else?

sethrj (approved):
Good catch. OpenMP is weird...

src/celeritas/global/ActionLauncher.hh
amandalund: It requires the loop bounds don't change:
> The iteration count for each associated loop is computed before entry to the outermost loop. If execution of any associated loop changes any of the values used to compute any of the iteration counts, then the behavior is unspecified.

so it expects that neither the iteration variable nor the loop bound is modified in the loop body. We shouldn't _need_ to cache the size, but it wouldn't hurt/might be a good idea.

sethrj (approved):
Thanks @amandalund !

PR #1493, opened 2024-11-10
Fix/standardize multi-level.gdml file and improve debug logging

sethrj:
The `multi-level.gdml` file worked fine through Geant4's reader but deleted most volumes when read through VecGeom's GDML reader, because the 'empty' name field was being saved as a unique and global identifier.

I added additional testing for the multilevel geometry for VecGeom and Geant4, and also added a `CELER_LOG_SCOPED` environment variable for unit testing that allows "scoped" log messages through to the user.


amandalund (approved):
Looks good, thanks @sethrj!

PR #1495, opened 2024-11-11
Reconstruct "touchable" using navigation hierarchy

### LINKED ISSUE #1248: Use logical rather than physical navigation state to reconstruct G4NavigationHistory

sethrj:
This should solve, once and for all, the difficulty reconstructing "touchable" states at volume boundaries. It defines a "level"-based touchable updater (rather than navigation-based) and hooks it up to the hit manager so that it's created for vecgeom/geant4 "core" geometry. ORANGE does not yet support volume instance ID mapping so it can't yet use this feature.

Closes #1248.


amandalund (approved):
Looks great, thanks for all the work getting this in!

PR #1496, opened 2024-11-12
Add option to use spline-interpolated energy loss

lebuller:
Adds an input option in the runner, `spline_eloss_order`, which is passed to `PhysicsParams::Input` as an option. This option is used to define the order of interpolation to be preformed in `calc_mean_energy_loss` in `PhysicsStepUtils.hh`. If this order is 1 the existing linear cross-section calculator is used and if it is greater than 1 a spline calculator of the given order is used.


test/celeritas/phys/MockProcess.cc
```diff
@@ -78,11 +78,12 @@ auto MockProcess::step_limits(Applicability applic) const -> StepLimitBuilders
     {
         auto eloss_rate = native_value_to<units::MevEnergy>(
             native_value_from(data_.energy_loss) * numdens);
+
         builders[ValueGridType::energy_loss]
             = std::make_unique<ValueGridLogBuilder>(
                 applic.lower.value(),
                 applic.upper.value(),
-                VecDbl{eloss_rate.value(), eloss_rate.value()});
+                VecDbl{eloss_rate.value(), eloss_rate.value(), eloss_rate.value()});
```
sethrj: At this point it might be clearer to use the size/fill constructor
```suggestion
                VecDbl(3, eloss_rate.value()));
```

app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -88,6 +88,7 @@ void from_json(nlohmann::json const& j, RunnerInput& v)
     LDIO_LOAD_OPTION(max_steps);
     LDIO_LOAD_REQUIRED(initializer_capacity);
     LDIO_LOAD_REQUIRED(secondary_stack_factor);
+    LDIO_LOAD_OPTION(spline_eloss_order);
```
sethrj, line 91: If you scroll down in the file, you'll see there's also a "save" that you should add as well.

src/celeritas/phys/PhysicsData.hh
```diff
@@ -265,7 +268,8 @@ struct PhysicsParamsScalars
                    == static_cast<bool>(fixed_step_action))
                && lambda_limit > 0 && range_factor > 0 && range_factor < 1
                && safety_factor >= 0.1
-               && step_limit_algorithm != MscStepLimitAlgorithm::size_;
+               && step_limit_algorithm != MscStepLimitAlgorithm::size_
+               && spline_eloss_order > 0;
```
sethrj: Move this check to right before `lambda_limit > 0` so that the ordering is the same as the variables.

src/celeritas/phys/PhysicsParams.hh
```diff
@@ -69,6 +69,10 @@ class ParticleParams;
  *   processes use MC integration to sample the discrete interaction length
  *   with the correct probability. Disable this integral approach for all
  *   processes.
+ *   \c spline_eloss_order: the order of interpolation to be used for the
+ *   spline interpolation. If it is 1, then the existing linear interpolation
+ * is used. If it is 2+, the spline interpolation is used for energy loss using
+ *   the specified order. Default value is 1
```
sethrj: Reindent please (guessing this is VSCode's fault...)

src/celeritas/phys/PhysicsStepUtils.hh
```diff
@@ -194,9 +195,22 @@ calc_mean_energy_loss(ParticleTrackView const& particle,
     {
         auto grid_id = physics.value_grid(VGT::energy_loss, ppid);
         CELER_ASSERT(grid_id);
-        auto calc_eloss_rate
-            = physics.make_calculator<EnergyLossCalculator>(grid_id);
-        eloss = Energy{step * calc_eloss_rate(pre_step_energy)};
+
+        size_type order = physics.scalars().spline_eloss_order;
+        CELER_ASSERT(order > 0);
```
sethrj: Since we already asserted by construction that the physics is "true", let's skip this check.

src/celeritas/phys/PhysicsStepUtils.hh
```diff
@@ -194,9 +195,22 @@ calc_mean_energy_loss(ParticleTrackView const& particle,
...
+
+        if (order == 1)
+        {
+            auto calc_eloss_rate
+                = physics.make_calculator<XsCalculator>(grid_id);
+            eloss = Energy{step * calc_eloss_rate(pre_step_energy)};
+        }
+        else if (order > 1)
+        {
```
sethrj: This should probably be:
```suggestion
        else
        {
```
unless you want to add an `else { CELER_ASSERT_UNREACHABLE(); }` after the `else`.

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -153,6 +153,11 @@ class PhysicsTrackView
     template<class T>
     inline CELER_FUNCTION T make_calculator(ValueGridId) const;
 
+    // Construct a grid calculator from a physics table
+    template<class T>
+    inline CELER_FUNCTION T make_calculator(ValueGridId,
+                                            const size_type order) const;
```
sethrj: Generally we don't `const` local function arguments, particularly because this can affect Windows ABI
```suggestion
                                            size_type order) const;
```

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -672,6 +677,21 @@ CELER_FUNCTION T PhysicsTrackView::make_calculator(ValueGridId id) const
...
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a Spline grid calculator of the given type
+ *
+ * The calculator must take three arguments: a reference to XsGridRef, a
+ * reference to the Values data structure, and an interpolation order.
+ */
+template<class T>
+CELER_FUNCTION T PhysicsTrackView::make_calculator(ValueGridId id,
+                                                   const size_type order) const
```
sethrj:
```suggestion
                                                   size_type order) const
```

sethrj:
Great work @lebuller ! There are just a few things to look at.

sethrj (approved):
Awesome! nice work.

PR #1489, opened 2024-11-06
Fix linking Athena with VecGeom+CUDA

pcanal:
`libvecgeomcuda.[a|so]` is the final library and it includes the result
of nvlink.  It is not appropriate to use as the intermediary link library.

VecGeom also never produce a shared 'middle' library, so we just always
need to use`libvecgeomcuda.[a|so]` is the final library and it includes the result
of nvlink.  It is not appropriate to use as the intermediary link library.

VecGeom also never produce a shared 'middle' library, so we just always
need to use `libvecgeomcuda_static.a`

Other fixes includes:

* removing spurrious dependencies
* adding support for private linking
* fix handling of some internals when seeing static libraries.

PR #1497, opened 2024-11-12
Attempt to fix mixed builds

pcanal:
This PR is opened (and will be immediately closed) to record the current (non working) state of work for trying to support mixed build.

mixed builds: VecGeom and Celeritas with static library vs shared library and vice et versa.

Those 3 commits (on top of #1489) seems to go in the right direction ... but are not sufficient (and are not needed for the full shared not the full static builds) so are should not be merged.

The static build of Celeritas using as shared build of VecGeom fails with things like:
```
Command: "/afs/cern.ch/user/p/pcanal/atlas-on-eos/celeritas/gcc13-local/build/celeritas-static-shared/test/orange/RaytraceImager"
[ RUN      ] RaytraceImagerTest/two_volume.device
unknown file: Failure
C++ exception with description "celeritas: CUDA error: invalid device function
.../celeritas/src/corecel/sys/KernelAttributes.hh:75: 'cudaFuncGetAttributes(&attr, reinterpret_cast<void const*>(func))' failed" thrown in the test body.
```

The shared build of Celeritas using as static build of VecGeom fails due an internal error in the RDC infrastructure:
```
==2747505== Invalid write of size 8
==2747505==    at 0x487716: __cudaRegisterLinkedBinary(__fatBinC_Wrapper_t const*, void (*)(void**), void*) (link.stub:153)
==2747505==    by 0x48649C: __cudaRegisterLinkedBinary_4565b2eb_23_PlacedEllipticalTube_cu_706e5800_4124581 (tmpxft_0025573d_00000000-3_cmake_device_link.reg.c:27)
==2747505==    by 0x400507D: call_init (dl-init.c:70)
==2747505==    by 0x400507D: call_init (dl-init.c:26)
==2747505==    by 0x400516B: _dl_init (dl-init.c:117)
==2747505==    by 0x401CCB9: ??? (in /usr/lib64/ld-linux-x86-64.so.2)
==2747505==  Address 0x2c1f7d48 is not stack'd, malloc'd or (recently) free'd
```
likely due to some misery with under counting (or double adding) the number of `LinkedBinary` to register.


sethrj:
😢

PR #1498, opened 2024-11-13
Work around RDC-related link errors when +cuda+vecgeom~shared

esseivaju:
Since #1489, a static build of Celeritas and VecGeom with `CELERITAS_USE_PNG=OFF` the linker probably figure out that geocel is not used by `ImageWriter.test.cc`, but it results in undefined symbols because we link to the VecGeom middle library. Move the inline dummy implementation if `ImageWriter` to its own TU to force the correct behavior.


sethrj (commented):
The problem isn't that imagewriter needs a .cc file, it's that the linker is removing `geocel` from the link libraries because it's "not needed", and that's breaking something with the RDC implementation. For the CI we have to add `-Wl,--no-as-needed`  for ubunutu systems, maybe we could try adding that to the link flags if vecgeom+static+cuda is enabled?

I think the proposed solution is a little clumsy and is working around the wrong issue, and later we might be confused why we need a separate `.cc` file when we don't do that for any of the other "not implemented" cases. Since this is an edge case it might be better to document by not building the test under this circumstance?

esseivaju:
Sure, the problem is the linker, and this is just a way to force the linker's hand since AFAIK, Cmake doesn't have good support for passing `--no-as-needed` just for one library. And it'd be a different argument for static libs, `--whole-archive`? That's probably not worth adding some CMake complexity, so yeah, let's just disable that test in that specific scenario.

pcanal:
> a static build of Celeritas and VecGeom  ... the linker probably figure out that geocel is not used... but it results in undefined symbols because we link to the VecGeom middle library. 

This sounds odd.  Presumably the link line has both `geocel_final.a` and the "VecGeom middle library" why would it ignore/drop one but not the other?

esseivaju:
It is pretty odd indeed, I don't know why the linker is acting that way. Unfortunately, Perlmutter is down today so I can't check the link line but I'm pretty sure it was indeed linking against `libvecgeomcuda_static`.

sethrj (approved):

PR #1499, opened 2024-11-14
Move `alongstep` directory to top level

sethrj:
This moves `alongstep` out from under `global` which makes less sense now than when we first added it. See #1263


amandalund (approved):
Makes sense!

PR #1500, opened 2024-11-14
Fix redundant move warning

esseivaju:
This is showing up as a warning with GCC 13.1.


sethrj (approved):
Interesting... I thought that NRVO only applied if it's a named variable in the function. It looks like it's automatically elided because it's an [`xvalue`](https://en.cppreference.com/w/cpp/language/value_category#xvalue)?

esseivaju:
I think, this specific case, URVO would apply, which is mandatory per the semantics of xvalues, and not under the  " mandatory copy elision" rules

PR #1501, opened 2024-11-14
Default SD integration to "true"

sethrj:
Now that our SD integration is more mature than it was for v0.2, and since "eat all our particles and don't send back hits" is surprising for adding Celeritas, I think we should turn full integration *on* by default and have users disable components that they don't need to further accelerate their code.

### DIFF
--- a/src/accel/SetupOptions.hh
+++ b/src/accel/SetupOptions.hh
@@ -49,22 +49,22 @@ struct SDSetupOptions
 {
     struct StepPoint
     {
-        bool global_time{false};
-        bool position{false};
-        bool direction{false};  //!< AKA momentum direction
-        bool kinetic_energy{false};
+        bool global_time{true};
+        bool position{true};
+        bool direction{true};  //!< AKA momentum direction
+        bool kinetic_energy{true};
     };
 
     //! Call back to Geant4 sensitive detectors
-    bool enabled{false};
+    bool enabled{true};
     //! Skip steps that do not deposit energy locally
     bool ignore_zero_deposition{true};
     //! Save energy deposition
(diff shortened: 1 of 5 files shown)


amandalund (approved):
Makes sense to me, thanks!

src/accel/SetupOptions.hh
```diff
@@ -26,45 +26,67 @@ struct AlongStepFactoryInput;
...
  *   KineticEnergy, and \c MomentumDirection will be copied to the \c Track
- *   when the combination of options is enabled
- * - Track and Parent IDs will \em never be a valid value since Celeritas track
- *   counters are independent from Geant4 track counters.
- * - Some within-step properties (material, material cuts couple, and
+ *   when the combination of options is enabled.
+ * - Some pre-step properties (\c Material and \c MaterialCutsCouple, and
  *   sensitive detector) are always updated. Post-step values for those are not
  *   set.
+ * - Track and Parent IDs will \em never be a valid value since Celeritas track
```
pcanal, line 54:
```suggestion
 * - Track and Parent IDs will \em never be a valid value since Celeritas track
```
humm .. sounds like they are random value and thus not even needed to kept?  Do we mean just that the value there are internal to Celeritas and completely de-corelated from any similar sounding Geant4 values?

pcanal (approved):

src/accel/SetupOptions.hh
sethrj: The track/parent IDs in the reconstructed track aren't valid since we don't set them. We don't set them because yes they're not directly correlated to Geant4 tracks.

PR #1502, opened 2024-11-14
Fix documentation

sethrj:
There were a few small things in the docs not being generated correctly or needing updating.

### DIFF
--- a/src/celeritas/decay/interactor/MuDecayInteractor.hh
+++ b/src/celeritas/decay/interactor/MuDecayInteractor.hh
@@ -117,12 +117,12 @@ class MuDecayInteractor
 /*!
  * Construct with shared and state data.
  *
- * \note Geant4 physics manual defines \f$ E_{max} = m_\mu / 2 \f$, while the
- * source code (since v10.2.0 at least) defines \f$ E_{max} = m_\mu / 2 - m_e
- * \f$ . The source
- * code implementation leads to a total CM energy of ~104.6 MeV instead of the
- * expected 105.7 MeV (muon mass), which is achieved by using the physics
- * manual definition.
+ * \note Geant4 physics manual defines \f$ E_{max} = m_\mu / 2 \f$,
+ * while the source code (since v10.2.0 at least) defines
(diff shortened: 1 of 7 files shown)


src/celeritas/decay/interactor/MuDecayInteractor.hh
```diff
@@ -117,8 +117,9 @@ class MuDecayInteractor
 /*!
  * Construct with shared and state data.
  *
- * \note Geant4 physics manual defines E_{max} = m_\mu / 2, while the source
- * code (since v10.2.0 at least) defines E_{max} = m_\mu / 2 - m_e . The source
+ * \note Geant4 physics manual defines \f$ E_{max} = m_\mu / 2 \f$, while the
+ * source code (since v10.2.0 at least) defines \f$ E_{max} = m_\mu / 2 - m_e
+ * \f$ . The source
```
stognini: Nitpick, but maybe select and rewrap this comment block?

elliottbiondo (approved):

stognini:
Also it seems the decay documentation has a couple mistakes I missed:

- The `(:math:\Gamma_i/\Gamma)` in the html table is not rendering correctly. Probably has to be `\Gamma_i / \Gamma`
- The `doxygenclass` calling `MuDecayInteractor` is not being found

Edit: just noticed the addition of `"${PROJECT_SOURCE_DIR}/src/celeritas/decay/interactor/*.hh"`. So that should fix the second bullet point.

amandalund (approved):

stognini (approved):

sethrj:
@stognini From the "checks" tab you can see the "artifacts" list which includes [doc-user](https://github.com/celeritas-project/celeritas/actions/runs/11843230966/artifacts/2189286673) that you can download and use to preview the user docs.  It's not a bad idea to use this trick to preview major chunks of documentation that you add to the manual :)

stognini:
Ohh I never noticed that download link. Nicer than generating it locally during build. Thanks for the tip.

PR #1485, opened 2024-11-04
Add a cheat-sheet to run app outside ctest

pcanal:
To make it easier to rerun outside of `ctest` the app runs using `test-harness.py`

PR #1504, opened 2024-11-15
Add debugging information for app tests

sethrj:
This prints additional information and input files when running `app` tests to improve the ability to debug if needed. Supersedes #1485 which only does this for celer-g4.


pcanal (approved):
LGTM.  Thanks :)

PR #1507, opened 2024-11-15
Add an Optical WLS

whokion:
Add an optical warelength shift (WLS) interactor and an associated test:
 - Add data and params for WLS
 - Implement an interactor and tests for WLS
 - Add secondary photon memory allocation in the optical Interact

This is an initial implementation of an optical wavelength interactor based on the G4OpWLS process with a couple of differences: 
 - Uses an inverse cumulative distribution for sampling the re-emitted photon spectrum replacing the rejection technique used in Geant4. 
 - Supports only an exponential time structure for remitted photons for now, which will be extended for a general time structure later (depending on whether we support two WLS processes similar to Geant4 and use configurable (run-time) options for the re-emitted time structure".


src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+    for (size_type i : range(num_photons))
+    {
+        // Sample the emitted energy from the inverse cumulative distribution
+        real_type energy = calc_energy_(generate_canonical(rng));
+        if (CELER_UNLIKELY(energy > inc_energy_.value()))
+        {
+            // Sample a restricted energy below the incident photon energy
+            real_type cdf_max = calc_cdf_(inc_energy_.value());
+            UniformRealDistribution<real_type> sample_cdf(0, cdf_max);
+            energy = calc_energy_(sample_cdf(rng));
```
sethrj, line 154: I think we need a CDF sampler data type + class, rather than constructing an inverted xs/cdf data, then creating an xs calculator, and having these two helper clasess.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+#include "celeritas/random/distribution/UniformRealDistribution.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Apply the wavelength shift (WLS) to optical photons.
+ *
```
sethrj, line 32: Can you add a few sentences about the physical process of WLS and the underlying sampling methodology?

src/celeritas/optical/WavelengthShiftData.hh
```diff
@@ -0,0 +1,80 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Material dependent scalar property of wavelength shift (WLS).
+ */
+struct ScalarPropertyWLS
+{
+    real_type mean_num_photons{};  //!< Mean number of reemitted photons
+    real_type time_constant{};  //!< Time delay of WLS // units::nanosecond
```
sethrj: This (and anything else that's not internal to a class or wrapped in a `Quantity`) should be in native units:
```suggestion
    real_type time_constant{};  //!< Time delay of WLS [time]
```
From the test code and setup it looks like the units are correctly native, it's just the doc here that needs fixigin.

src/celeritas/optical/WavelengthShiftData.hh
```diff
@@ -0,0 +1,80 @@
...
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Material dependent scalar property of wavelength shift (WLS).
+ */
+struct ScalarPropertyWLS
```
sethrj: Maybe rename `WlsMaterialRecord`? Our style is to try to avoid consecutive blocks of upper-case letters, and this is all the data needed for a single optical material.

The `GenericGridRecord` could be a scalar data member of this class as well, since it's tied to the same optical material.

src/celeritas/optical/WavelengthShiftParams.cc
```diff
@@ -0,0 +1,112 @@
...
+        std::vector<real_type> cdf(comp_vec.x.size());
+
+        CELER_ASSERT(comp_vec.y[0] > 0);
+        // The value of cdf at the low edge is zero by default
+        cdf[0] = 0;
+        for (size_type i = 1; i < comp_vec.x.size(); ++i)
+        {
+            cdf[i] = cdf[i - 1]
+                     + real_type(0.5) * (comp_vec.x[i] - comp_vec.x[i - 1])
+                           * (comp_vec.y[i] + comp_vec.y[i - 1]);
```
sethrj: A little function `real_type integrate_trapezoid(Real2 lo, Real2 hi)` would improve clarity here.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+ * \note This performs the same sampling routine as in the G4OpWLS class of
+ * the Geant4 release 11.2.
+ */
+class WavelengthShiftInteractor
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Energy = units::MevEnergy;
+    using WLSDataCRef = NativeCRef<WavelengthShiftData>;
```
sethrj: Could just be `ParamsRef`?

sethrj (commented):
Really nice work! I've got a couple of suggestions; if you want I can write the CDF sampler class.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: Thanks, it would be great if you can help to write a CDF sampler class.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: No problem.  A short description of WLS is added.

src/celeritas/optical/WavelengthShiftData.hh
whokion: Updated.

src/celeritas/optical/WavelengthShiftData.hh
whokion: Renamed as suggested.

src/celeritas/optical/WavelengthShiftParams.cc
whokion: Can you help this when you add the CDF sampler class too?  Thanks.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: Renamed.

whokion:
@sethrj Thanks a lot for your review and comments.  I addressed most of them, but left a couple of suggested improvements to you (a CDF sampler class and the integrate_trapezoid utility function).  Sorry for asking additional help - I will be away for a week, but should catch up any other suggestion later.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
amandalund: For muon pair production we also need to sample from a CDF, but on a 2D grid like we do in the [cascade collider](https://github.com/celeritas-project/celeritas/blob/e390d55956c3f59124f97bf7cddb8d1d259ecd93/src/celeritas/neutron/interactor/detail/CascadeCollider.hh#L146-L161): I was also thinking it would be nice to have a helper class for that.

sethrj: Good thought @amandalund . I think the 2d grid has (or at least should have!) a "slice" functionality that lets it act like 1D grid data. So maybe we template the CDF sampler so we can use it for both.

src/celeritas/optical/WavelengthShiftParams.cc
```diff
@@ -0,0 +1,112 @@
...
+    GenericGridInserter insert_energy_cdf(&data.reals, &data.energy_cdf);
+    for (auto const& wls : input.data)
+    {
+        // Store WLS component tabulated as a function of photon energy
+        auto const& comp_vec = wls.component;
+
+        if (comp_vec.x.empty())
+        {
+            // No WLS data for this material
+            continue;
```
hhollenb, line 66: Should the material be skipped if it doesn't have WLS components? My understanding is this would require all materials to specify WLS data.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -0,0 +1,174 @@
...
+template<class Engine>
+CELER_FUNCTION Interaction WavelengthShiftInteractor::operator()(Engine& rng)
+{
+    // Sample the number of photons generated from WLS
+    unsigned int num_photons = sample_num_photons_(rng);
+
+    if (num_photons == 0)
+    {
+        // Return unchanged photon without reemitted optical photons
+        return Interaction::from_unchanged();
```
hhollenb: If 0 photons are re-emitted, is the incident photon not still absorbed? I'm unsure about how absorption via WLS is accounted for in the usual absorption data.

test/celeritas/optical/WavelengthShift.test.cc
```diff
@@ -0,0 +1,251 @@
...
+{
+  protected:
+    using HostDataCRef = HostCRef<WavelengthShiftData>;
+
+    void SetUp() override
+    {
+        // Build wavelength shift (WLS) property data from a test input
+        ImportWavelengthShift wls;
+        wls.mean_num_photons = 2;
+        wls.time_constant = native_value_from(TimeSecond(1e-9));
```
hhollenb, line 59: Rather than the single use of `TimeSecond` here, can probably use `1 * units::nanosecond` as you do in the test below.

test/celeritas/optical/WavelengthShift.test.cc
```diff
@@ -0,0 +1,251 @@
...
+                                                 2.46151256760185e-06};
+
+    static double const expected_avg_costheta[] = {0.0157611129315312,
+                                                   0.0325629374415683,
+                                                   0.0082191738981211,
+                                                   0.0128202506207409};
+
+    static double const expected_avg_orthogonality[] = {0, 0, 0, 0};
+
+    static double const expected_avg_time[] = {1.08250310854364e-09,
```
hhollenb, line 230: Expressing the expected values in units should fix the test errors, and make it easier to someone to come back and check in case of a regression.

src/celeritas/optical/WavelengthShiftParams.hh
```diff
@@ -0,0 +1,58 @@
...
+{
+struct ImportData;
+
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Build and manage wavelength shift (WLS) data.
+ */
+class WavelengthShiftParams final
```
amandalund, line 27: Instead of a `Params` class, should this be a `Model` (like Rayleigh and absorption) which also builds the mfp from the imported data?

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -0,0 +1,174 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Sampling the wavelength shift (WLS) photons as in G4OpWLS of the Geant4
+ * release 11.2 with a modification in sampling the reemission spectrum.
+ */
+template<class Engine>
+CELER_FUNCTION Interaction WavelengthShiftInteractor::operator()(Engine& rng)
+{
+    // Sample the number of photons generated from WLS
+    unsigned int num_photons = sample_num_photons_(rng);
```
amandalund, line 120: Sampling the number of photons before allocating space for secondaries means we won't be able to reproducibly recover in the case of an out-of-memory error. I don't think we need to worry about it now since we don't even have that capability yet, but maybe add a note or todo here?

amandalund (commented):
Great work @whokion!

src/celeritas/optical/WavelengthShiftParams.hh
whokion: Yes, this can be consolidated into a WLS model when it is added later. For now, I just want to add an interactor as the first step until we feature out how to support multiple WLS models/processes in a organized way.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: Network access is so slow here - almost impossible to do anything. Can you help to add the note? Otherwise, I will do it when I get back  (this weekend). Thanks.

amandalund: No problem!

src/celeritas/optical/WavelengthShiftParams.hh
hhollenb: I might misunderstand, but these `Params` refer to data needed by the interactor rather than the MFP grid, similar to something like `em/KleinNishinaData.hh`, correct? I think WLS is unique in that it's the only one of the interactors that needs on-device data, whereas Rayleigh's data wasn't used by the interactor and purely for building MFPs.

amandalund: Right, this is just the data needed by the interactor. We typically build and store that data directly in the model class, and for EM physics get the cross sections from the process. Since we have a combined model/process for optical, I think the model will build the mfps and the data for the interactor.

test/celeritas/optical/WavelengthShift.test.cc
sethrj: `native_value_from(TimeSecond(..))` is more readable: I've frequently encountered errors when `*` should be `/` or vice versa.

src/celeritas/optical/WavelengthShiftParams.hh
hhollenb: Ah, gotcha. That makes sense.

src/celeritas/optical/WavelengthShiftParams.cc
whokion: I think that this is still okay as this constructor is called through`from_import` which loads imported data only for the optical material.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: Yes, it won't be actually absorbed (i.e., no wavelength shifting is applied) even though the optical photon should be moved for the proposed length - in principle, the zero optical photon may be emitted from the excited WLS material (something like the internal conversion which may not be allowed in the selection rule, but can happen quantum mechanically).

test/celeritas/optical/WavelengthShift.test.cc
whokion: Okay, thanks.

whokion:
@sethrj can you help how to fix the compilation error from [build-spack / orange-float-g4@11.0](https://github.com/celeritas-project/celeritas/actions/runs/12035855167/job/33556318541?pr=1507#logs)?  Thanks.

src/celeritas/optical/WavelengthShiftParams.hh
amandalund: @whokion what were your thoughts on how to support multiple WLS models? Could we e.g. as a start just have multiple instances of a single `WavelengthShiftModel` with different WLS properties as inputs (and I suppose some way to determine which MFPs went with which model)?

whokion: @amandalund One tricky part is how to support multiple models/cases for different ranges of WLS absorption lengths (ABS) and output spectra within a single process.  There are several different cases: 1)  input optical photon ranges (either in energy or wavelength) of ABS and output spectra are completely exclusive each other, which is the simpliest case that can be handled by one model and one interactor with a combined data, 2)  part of ABS ranges are overlapped, but with completely exclusive output spectra  which is the case that multiple models can be compete for the WLS process (based on their relative absorption probabilities) and produce output spectrum with the corresponding data and in the target range if the selected WLS model wins over other discrete processes 3) both of ABS ranges and spectra are overlapped (i,e., output of WLS may be input of WLS2, which may be treated as a similar way as 2) with some variations, 4) one ABS input range, but have two different output spectra with different time structures (two transition levels from one WLS material).  Apparently, Geant4 tries to support these different cases with the combination of two separated/independent WLS and WLS2 with different input data (so, users must know exactly what they want to simulate and how to configure different cases correctly).   At this moment, I do not have a clear picture that is different from the Geant4 approach, so add this initial interactor with a minimal data structure as a starting point.  Of course, other pieces of good ideas are welcomed for the better design.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
sethrj: This still sounds wrong: even though we've sampled a photon interaction at this point, no interaction takes place? The original photon continues with its energy, direction, polarization?

src/celeritas/optical/WavelengthShiftParams.cc
sethrj: I'll leave a note as a to-do.

src/celeritas/optical/WavelengthShiftParams.hh
sethrj: So since we *do* have an optical model class, I think it would be good to update that before merging.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
sethrj: I've annotated the to-do.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
sethrj: I added a to-do here; we don't *have* to have this as part of the PR.

src/celeritas/optical/WavelengthShiftParams.hh
sethrj: It seems likely we should be able to use a single WLS model, so perhaps we should go ahead and make this class a Model?

src/celeritas/optical/WavelengthShiftParams.cc
sethrj: Good catch @hhollenb . @whokion I've pushed what I think is a fix: if there's no WLS data (`!wls`) then we push back an empty vector for both.

sethrj:
@whokion I've fixed the floating point error and added a couple of to-dos. Could you please take a look at the remaining discussion?

src/celeritas/optical/WavelengthShiftParams.cc
whokion: Thanks for the update!

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
whokion: Okay, I simply changed the action from `from_unchanged` to `from_absorption` (i.e, kill the primary optical photon) so that the outer loop treats WLS as same as the absorption process - this is equivalent to what Geant4 does.

src/celeritas/optical/WavelengthShiftParams.hh
whokion: Actually we need to add a separate WLS2 if we want to follow what exactly Geant4 does.  Otherwise, we need to combine different cases into one convolved model.  So, if you do not mind, let's leave implementation options for the next MR.

hhollenb (approved):
Pending Seth's and Amanda's suggestions, I think it looks good!

src/celeritas/optical/WavelengthShiftParams.hh
amandalund: I don't see why we can't do exactly what Geant4 does with a single WLS model class.  Couldn't we do something like:
```c++
auto wls1 = std::make_shared<WavelengthShiftModel>(action_id++, imported, ImportModelClass::wls1);
auto wls2 = std::make_shared<WavelengthShiftModel>(action_id++, imported, ImportModelClass::wls2);
```
where the last argument determines which imported WLS optical properties and MFPs to use? The two models will compete for the discrete interaction like in Geant4.

whokion: Yes, this is equivalent to what Geant4 does (assuming that there is an additional mechnism to call a common interactor with different wls data to generate WLS photon with a different/corresponding spectrum depending on which one is selected).  The main point is whether we can combine these two identical models in the step competition into single one with a combined data (which is more general, I think) - it is straightforward if ranges of ABS are exclusive each other as I mentioned earlier, but not clear how to do it cleanly for other cases even though it is a purely technical question.

whokion:
@sethrj Thanks a lot for fixing the float issue and adding other notes.  Besides the discussion for adding a model or models which can be revisited in the next MR, it would be great that we can merge this as it is.  Thanks again!

src/celeritas/optical/WavelengthShiftParams.hh
amandalund: Yes, launching the interactors should work fine without any doing anything special since the models use exactly the same interactor and data classes. I think this would be the simplest way to do things for now, and we could think about having a "combined" model in the future.

whokion: Yes, I agree.  Chatting with @hhollenb, he also has a decent idea how to combine them multiple WLS processes for the future improvement.

sethrj:
@amandalund If you're satisfied then please merge!

sethrj (approved):
Thanks for the new model! I've added the little "todo" tasks to https://github.com/celeritas-project/celeritas/issues/1263 .

amandalund (approved):
Excellent, thanks @whokion!

PR #1509, opened 2024-11-16
Refactor muon bremsstrahlung angular distribution

amandalund:
This refactors the muon bremsstrahlung polar angle sampling into a separate class (which will also be used for pair production) and uses the `TwodGridBuilder` to build the SB DCS tables.

### DIFF
--- a/src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
+++ b/src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
@@ -1,0 +1,103 @@
+//----------------------------------*-C++-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/em/distribution/MuBremsPPAngularDistribution.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "corecel/Macros.hh"
+#include "corecel/Types.hh"
+#include "corecel/math/Algorithms.hh"
+#include "celeritas/Constants.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/random//distribution/GenerateCanonical.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the polar angle for muon bremsstrahlung and pair production.
+ *
+ * The polar angle is sampled according to a simplified PDF
+ * \f[
+   f(r) \sim \frac{r}{(1 + r^2)^2}, \ r = frac{E\theta}{m}
+ * \f]
+ * by sampling
+ * \f[
+   \theta = r\frac{m}{E}
+ * \f]
+ * with
+ * \f[
+   r = \sqrt{frac{a}{1 - a}},
+   a = \xi \frac{r^2_{\text{max}}}{1 + r^2_{\text{max}}},
+   r_{\text{max}} = frac{\pi}{2} E' / m \min(1, E' / \epsilon),
+ * \f]
+ * and where \f$ m \f$ is the incident muon mass, \f$ E \f$ is incident energy,
+ * \$ \epsilon \f$ is the emitted energy, \f$ E' = E - \epsilon \f$, and \f$
+ * \xi \sim U(0,1) \f$.
+ *
+ * \note This performs the same sampling routine as in Geant4's \c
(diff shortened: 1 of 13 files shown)


src/celeritas/em/model/SeltzerBergerModel.cc
```diff
@@ -154,30 +155,16 @@ void SeltzerBergerModel::step(CoreParams const&, CoreStateDevice&) const
...
-    // Incident charged particle log energy grid
-    table.grid.x = reals.insert_back(imported.x.begin(), imported.x.end());
-
-    // Photon reduced energy grid
-    table.grid.y = reals.insert_back(imported.y.begin(), imported.y.end());
-
-    // 2D scaled DCS grid
-    table.grid.values
-        = reals.insert_back(imported.value.begin(), imported.value.end());
+    table.grid = TwodGridBuilder{&tables->reals}(imported);
```
sethrj: I think we (I) still need to go back and refactor this as a builder helper class that can keep a persistent 2d grid builder across multiple elements.

src/celeritas/io/ImportPhysicsVector.hh
```diff
@@ -60,7 +60,7 @@ struct ImportPhysics2DVector
 
     explicit operator bool() const
     {
-        return !x.empty() && !y.empty() && value.size() == x.size() * y.size();
+        return !value.empty() && value.size() == x.size() * y.size();
```
sethrj, line 63: So an empty vector is "true" because it's consistently defined, even though it's empty? The semantics of `operator bool` may be a little inconsistent across all our import data, we should at least mention in the documentation block that it's true if it's consistent regardless of being empty.

src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
```diff
@@ -0,0 +1,103 @@
...
+#include "corecel/math/Algorithms.hh"
+#include "celeritas/Constants.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/random//distribution/GenerateCanonical.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the polar angle for muon bremsstrahlung and pair production.
```
sethrj, line 21: Nice docs, can you include the class in the `Bremsstrahlung` section of the `em-physics` section?

src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
```diff
@@ -0,0 +1,103 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the cosine of the polar angle of the secondary.
+ */
+template<class Engine>
+CELER_FUNCTION real_type MuBremsPPAngularDistribution::operator()(Engine& rng)
+{
+    real_type a = generate_canonical(rng) * a_over_xi_;
+    return std::cos(std::sqrt(a / (1 - a)) / gamma_);
```
sethrj, line 99: I wonder if there's a small numerical stability issue here that's causing the windows test to fail?

sethrj (approved):
Great refactor! Now to just track down the windows failure...

src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
amandalund: Weird, I don't know what I've changed that would have affected the `BetheBloch` or `MuBetheBloch` tests...

src/celeritas/io/ImportPhysicsVector.hh
amandalund: This should still only be true for non-empty, right?

sethrj: Oh silly me, didn't read it thoroughly enough 🤦‍♂️

src/celeritas/em/distribution/MuBremsPPAngularDistribution.hh
sethrj: Yeah I didn't even notice it was the BetheBloch ionization as opposed to brems... it still uses the same runner version that Soon's recent (successful) PR did... if it fails on the next run let's just loosen the tolerance I guess?

amandalund: Still failing, so I loosened the tolerance.

src/celeritas/em/model/SeltzerBergerModel.cc
amandalund: I went ahead and added this.

PR #1510, opened 2024-11-18
Add FourVector helper functions

sethrj:
While looking at adding CDF utilities for #1507 I noticed some common (and suboptimal) patterns for FourVector that could be simplified and clarified with helper functions.
- Add `from_particle` and `from_mass_momentum` constructors for FourVector
- Make "internal" typedefs private
- Use helper typedefs for particle class

### DIFF
--- a/src/celeritas/phys/FourVector.hh
+++ b/src/celeritas/phys/FourVector.hh
@@ -67,7 +67,8 @@ CELER_FUNCTION FourVector FourVector::from_mass_momentum(Mass m,
                                                          Momentum p,
                                                          Real3 const& direction)
 {
-    return {p.value() * direction, std::hypot(p.value(), m.value())};
+    return {p.value() * direction,
+            std::sqrt(ipow<2>(p.value()) + ipow<2>(m.value()))};
 }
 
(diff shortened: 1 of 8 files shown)


amandalund (approved):
Nice, thanks @sethrj!

src/celeritas/phys/FourVector.hh
```diff
@@ -37,6 +56,33 @@ struct FourVector
 
 //---------------------------------------------------------------------------//
 // INLINE UTILITY FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from rest mass, momentum, direction.
+ *
+ * Note that this uses \c std::hypot which yields more accurate answers if the
```
stognini: Where is `std::hypot()` being called?

stognini (approved):
Looks great. Thanks @sethrj !

src/celeritas/phys/FourVector.hh
sethrj: It's not any more since it changed the results on some of the CI platforms... we should consider a platform-portable reimplementation for reproducibility.

PR #1511, opened 2024-11-19
Add a CI job for c++20

esseivaju:
Even though we don't use c++20 yet, some clients might. Add fast CI jobs to check that we are compatible.

### DIFF
--- a/.github/workflows/build-fast.yml
+++ b/.github/workflows/build-fast.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   linux:
-    name: "${{matrix.runner}}-${{matrix.compiler}}-${{matrix.version}}"
+    name: "${{matrix.runner}}-${{matrix.compiler}}-${{matrix.version}}${{ (matrix.cxx_std && matrix.cxx_std != '17') && format('-c++{0}', matrix.cxx_std) || ''}}"
     strategy:
       matrix:
         include:
@@ -23,7 +23,11 @@ jobs:
           - runner: noble
             compiler: gcc
             version: 14
-            cxx_std: [17, 20]
+            cxx_std: 17
+          - runner: noble
+            compiler: gcc
+            version: 14
+            cxx_std: 20
           - runner: focal
             compiler: clang
             version: 10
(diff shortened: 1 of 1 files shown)


.github/workflows/build-fast.yml
```diff
@@ -23,6 +23,10 @@ jobs:
           - runner: noble
             compiler: gcc
             version: 14
+          - runner: noble
+            compiler: gcc
+            version: 14
+            cxx_std: 20
```
sethrj: Let's drop the `_` since spack's standard variant is `cxxstd` and it makes the matrix optinos look more consistent :)

.github/workflows/build-fast.yml
```diff
@@ -72,7 +80,7 @@ jobs:
       - name: Configure Celeritas
         run: |
           ln -fs scripts/cmake-presets/ci-ubuntu-github.json CMakeUserPresets.json
-          cmake --preset=${CMAKE_PRESET} \
+          cmake ${{matrix.cxx_std && format('-DCMAKE_CXX_STANDARD={0} ', matrix.cxx_std) || ''}}--preset=${CMAKE_PRESET} \
             -DCeleritas_GIT_DESCRIBE="${{github.event.pull_request
               && format(';-pr.{0};', github.event.pull_request.number)
               || format(';-{0};', github.ref_name)}}"
```
sethrj:
```suggestion
          cmake --preset=${CMAKE_PRESET} \
            ${{matrix.cxx_std && format('-DCMAKE_CXX_STANDARD={0}', matrix.cxx_std) || ''}} \
            -DCeleritas_GIT_DESCRIBE="${{github.event.pull_request
              && format(';-pr.{0};', github.event.pull_request.number)
              || format(';-{0};', github.ref_name)}}"
```

.github/workflows/build-fast.yml
```diff
@@ -111,7 +119,7 @@ jobs:
                 || steps.unittest.outcome == 'failure')
           }}
         with:
-          name: test-results-fast-${{matrix.compiler}}-${{matrix.version}}
+          name: test-results-fast-${{env.GHA_JOB_NAME}}
```
sethrj, line 123: Could you ensure the other `test-results-` cache names in the other build scripts are updated equivalently?

sethrj (approved):
Thanks @esseivaju ! I thought you said that the `range` test failed under c++20; did you not see that in this case? Or was that only on CUDA?

.github/workflows/build-fast.yml
esseivaju: do you mean in `build-spack`, `build-docker` and `build-ultralite`? All of them use a different matrix

esseivaju:
> Thanks @esseivaju ! I thought you said that the `range` test failed under c++20; did you not see that in this case? Or was that only on CUDA?

Yes, it only happens when CUDA is enabled. The [step function](https://github.com/celeritas-project/celeritas/blob/0b7a41182f1000269dd834daa7af1dac42c40800/test/corecel/cont/Range.test.cu#L27-L28) is causing the issue on device.

.github/workflows/build-fast.yml
sethrj: Right, I forgot that we're the one who defines GHA_JOB_NAME.

PR #1513, opened 2024-11-19
Define hypotenuse function using fma

sethrj:
Following on to #1510 this adds a `hypot` function to allow us to add a platform-independent and potentially higher accuracy definition. The current implementation uses std::fma rather than ipow to reduce the number of floating point operations and registers.


amandalund (approved):
Nice addition!

PR #1514, opened 2024-11-19
Split CeleritasUtils cmake and use VecGeom RDC if available

sethrj:
This splits most components of `CeleritasUtils.cmake` into different scripts:
- CeleritasOptionUtils has all the code needed to set up dependencies, options, etc.
- CeleritasLibraryUtils forwards arguments to cmake or CudaRdc and defines the helper functions for adding more complicated library macros

Instead of forwarding arguments to cmake inside of the library macros, it now defines the macros based on whether forwarding is needed.

We also now avoid installing CudaRdcUtils if VecGeom provides it.

Finally, I updated the `celeritas_add_library` macro for consistency with the other RDC forwarding ones, and added `celeritas_add_src_library` for libraries inside `src/` that are to be installed.


sethrj:
@pcanal This is working and ready for review.

pcanal (approved):
LGTM.  Thanks.

PR #1515, opened 2024-11-19
Fix celer-sim MC truth ROOT output file generation during tests

stognini:
This PR fixes a few minor bugs:
- Assign volume instance ids even when there are no detector map entries in `StepParams`.
- Disable volume instance ids in `RootStepWriter`, as these are not implemented by the class.
- Fix a typo in `app/celer-sim/CMakeLists.txt` that prevented generating mctruth output files during `celer-sim` tests.


sethrj (approved):
Thanks!

PR #1516, opened 2024-11-20
Stop CMake from cleaning root test data

hhollenb:
# Stop CMake from cleaning root test data

Since the files `test/celeritas/data/*.root` are outputs of `add_custom_command`, they acquire the `GENERATED` property which forces their inclusion in `clean` targets. The `GENERATED` property cannot be unset, so we can't just un-tag these files.

This is a modification to the `update-root-test-data` target. Instead of generating the data files directly, a brief CMake script is generated in the build directory. This script checks if the files exists and builds them if missing, and is sourced in `add_custom_target`s. The `clean` target will remove the script but won't touch the root files, and `update-root-test-data` will not regenerate them if they exist. 

I added the `force-update-root-test-data` target to force a regeneration of all the test data files. There's also targets for each individual root data file, so they can be generated / regenerated as needed.


test/celeritas/CMakeLists.txt
```diff
@@ -428,34 +428,72 @@ celeritas_add_test(user/StepCollector.test.cc
...
 endif()
 
 add_custom_target(update-root-test-data
-  COMMENT "Regenerate ROOT test data"
+  COMMENT "Regenerate ROOT test data if it doesn't exist"
   DEPENDS ${_root_data}
 )
 
+add_custom_target(force-update-root-test-data
+  COMMENT "Force regeneration of ROOT test data"
```
sethrj: This was the intent of `update-root-test-data`; to forcibly regenerate it on command. It's also what we need to call from the CI workflow to test that the ROOT I/O is in use.

test/celeritas/CMakeLists.txt
```diff
@@ -428,34 +428,72 @@ celeritas_add_test(user/StepCollector.test.cc
...
     )
-    list(APPEND _root_data "${_output}")
+
+    list(APPEND _force_update_root_data "force-update-${_target_name}")
   endforeach()
 endif()
 
 add_custom_target(update-root-test-data
-  COMMENT "Regenerate ROOT test data"
+  COMMENT "Regenerate ROOT test data if it doesn't exist"
```
sethrj: Maybe instead this target should be `root-test-data` (or `generate-`...).

test/celeritas/CMakeLists.txt
```diff
@@ -428,34 +428,72 @@ celeritas_add_test(user/StepCollector.test.cc
...
+      DEPENDS ${_update_deps}
+      COMMAND "${CMAKE_COMMAND}" "-P" "${_update_script}"
+      VERBATIM
+    )
+
+    list(APPEND _root_data "${_target_name}")
+
+    # Can't order targets in CMake: use commands + script instead
+
+    add_custom_target("force-update-${_target_name}"
```
sethrj: This ends up adding a lot of targets and helper scripts. Is this really the best solution?

hhollenb: It's definitely doable to have a single script & target that handles all the data files. The benefit of targets per file (might not be a benefit really) is that tests could specify the specific root file they depend on.

test/celeritas/CMakeLists.txt
hhollenb: If we only ever want to overwrite, that simplifies the logic I think.

test/celeritas/CMakeLists.txt
sethrj: I think if they *do* specify the specific dependencies, then `ninja` after `ninja clean` should indeed regenerate those files... which is the point I was trying to make on slack

sethrj (approved):
Nice solution!

PR #1517, opened 2024-11-20
Regenerate ROOT data for celeritas tests if missing

sethrj:
I propose this as a simpler alternative to #1516. As the reviewers previously described on slack, running
```console
$ ninja clean
$ ninja
$ ctest
```
fails because the ROOT files are not regenerated.

With this, ROOT data is regenerated automatically as part of `celeritas/test/all` (or specifically, the `testcel_celeritas` library) if are absent when building. Importantly, running `ninja` doesn't renew the files if they already exist.

### DIFF
--- a/test/celeritas/CMakeLists.txt
+++ b/test/celeritas/CMakeLists.txt
@@ -458,4 +458,11 @@ add_custom_target(update-root-test-data
   DEPENDS ${_root_data}
 )
 
+if(CELERITAS_USE_Geant4 AND CELERITAS_USE_ROOT)
+  # MscTestBase and other tests require ROOT data to be generated. If they're
+  # deleted by `ninja clean`, then this dependency reconstructs them if any
+  # test in this directory is built.
+  add_dependencies(testcel_celeritas update-root-test-data)
+endif()
+
 #-----------------------------------------------------------------------------#


stognini:
Tried a fully clean build here. The files existed, but were overridden during the fresh build, leading to
```
❯ git st
 M ../test/celeritas/data/four-steel-slabs.root
 M ../test/celeritas/data/lar-sphere.root
 M ../test/celeritas/data/simple-cms.root
```

sethrj:
Ooh that's not what we want either...

sethrj:
This appears to be happening (and on the CI) because of the many dependencies on the "custom command". Rebuilding the `celer-export-geant` binary forces the file to become out of date, and cloning/updating some of the other files does the same. This is technically correct behavior, but in practice we don't want to regenerate the files except rarely. I'm not sure what the best thing to do is; maybe #1516 really is the way to go.

hhollenb:
If it's safe to assume the files are persistent in `test/celeritas/data/`, then changing the commands to custom targets without any outputs will not tag the root files as `GENERATED` and therefore won't be cleaned. As far as I can tell, our use case just isn't well handled by cmake (it's not quite generated and not quite a source; it's a dependency that can't be determined at the configure stage, etc etc). Treating them like "source files" with a custom target to regenerate them manually might be the most idiomatic cmake way?

PR #1518, opened 2024-11-24
Add muon pair production

amandalund:
This adds the energy distribution, interactor, model, process, and tests for electron-positron pair production by muons.


test/celeritas/data/four-steel-slabs.root
sethrj, line 1: This file is becoming a chonker!

src/celeritas/ext/detail/GeantProcessImporter.cc
```diff
@@ -428,6 +428,13 @@ import_physics_vector(G4PhysicsVector const& g4v, Array<ImportUnits, 2> units)
 //---------------------------------------------------------------------------//
 /*!
  * Import a 2D physics vector.
+ *
+ * \note In Geant4 the values are stored as a vector of vectors indexed as
+ * [y][x]. Because the Celeritas \c TwodGridCalculator and \c
+ * TwodSubgridCalculator expect the y grid values to be on the inner dimension,
+ * the table is inverted during import so that the x and y grids are swapped.
+ *
+ * TODO: should we swap the grids after they're imported instead of here?
```
sethrj: No, let's keep it as you've done it. I think we should move toward import data being more like "Celeritas problem definition", so reflecting our code's capabilities/layout (or perhaps a format convenient for input) rather than trying to mirror what Geant4 has (which is how the IO stuff started out).

src/celeritas/ext/detail/GeantProcessImporter.cc
```diff
@@ -438,17 +445,17 @@ ImportPhysics2DVector import_physics_2dvector(G4Physics2DVector const& g4pv,
...
     for (auto i : range(pv.x.size()))
     {
-        pv.x[i] = g4pv.GetX(i) * x_scaling;
+        pv.x[i] = g4pv.GetY(i) * y_scaling;
         for (auto j : range(pv.y.size()))
         {
-            pv.y[j] = g4pv.GetY(j) * y_scaling;
-            pv.value[pv.y.size() * i + j] = g4pv.GetValue(i, j) * v_scaling;
+            pv.y[j] = g4pv.GetX(j) * x_scaling;
+            pv.value[pv.y.size() * i + j] = g4pv.GetValue(j, i) * v_scaling;
```
sethrj: Remind me later to improve the `corecel/data/HyperslabIndexer.hh` class so we can use it for consistent ordering for multi-D data in Celeritas...
```suggestion
    HyperslabIndexer index{{pv.x.size(), pv.y.size()}};
    pv.value[index(i, j)] = g4pv.GetValue(j, i) * v_scaling;
```

doc/implementation/em-physics.rst
```diff
@@ -120,6 +124,8 @@ The following table summarizes the EM processes and models in Celeritas.
                                     &                                     & Mu Bethe--Bloch      & \texttt{\scriptsize celeritas::MuHadIonizationInteractor}   & 200 keV -- 100 TeV \\
                                     \cline{2-5}
                                     & Bremsstrahlung                      & Mu bremsstrahlung    & \texttt{\scriptsize celeritas::MuBremsstrahlungInteractor}  & 0--100 TeV \\
+                                    \cline{2-5}
+                                    & Pair production                     & Mu pair production   & \texttt{\scriptsize celeritas::MuPairProductionInteractor}  & 0--100 TeV \\
```
whokion: Is the low energy limit right?

src/celeritas/em/model/MuPairProductionModel.hh
```diff
@@ -0,0 +1,74 @@
...
+#pragma once
+
+#include <functional>
+#include <memory>
+
+#include "corecel/data/CollectionMirror.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/em/data/MuPairProductionData.hh"
+#include "celeritas/io/ImportMuPairProductionTable.hh"
+#include "celeritas/phys/AtomicNumber.hh"
```
whokion:
```suggestion
```

src/celeritas/em/interactor/MuPairProductionInteractor.hh
```diff
@@ -0,0 +1,177 @@
...
+        // Failed to allocate space for a secondary
+        return Interaction::from_failure();
+    }
+
+    // Sample the electron and positron energies
+    auto energy = sample_energy_(rng);
+    Energy pair_energy = energy.electron + energy.positron;
+
+    // Sample the secondary directions
+    MuBremsPPAngularDistribution sample_costheta(
```
whokion: This class name is somewhat confusing here as the e+e- production is nothing to do with brems even though the distribution is re-used for sampling the same angular distribution used in the muon brems interactor.  Okay as it is, but the name may be changed to the more general one later.

whokion (commented):
@amandalund Excellent!  Just for your information, there is Geant4 TestEm17 example (under extended/electromagnetic) which tests different models of muon processes and can be used for PR verification plots with respect to Geant4 later - of course, our physics validate suite can be used for the model test.  Additional observables that can be added are the scattered (post) muon energy (or energy loss) and the angular deviation, which essentially carry equivalent kinematical information for the final state (i.e., distributions of secondaries).  Anyway, very nice work!

whokion (approved):

src/celeritas/em/interactor/MuPairProductionInteractor.hh
amandalund: I don't like the name either, I just wasn't sure what to call it (happy for any suggestions!). In Geant4 it's `G4ModifiedMephi`, which sounds like something to do with the azimuthal angle but I think actually stands for Moscow Engineering Physics Institute (MEPhI), where some of the authors are from. :)

doc/implementation/em-physics.rst
amandalund: Good point, it should be higher. In the Geant4 model it looks like the `lowestKinEnergy`/`MinPrimaryEnergy` is approximately 8 times the muon mass (0.85 GeV). The default physics list constructor says:
```console
Pair production
muPairProd:   SubType=4
    dE/dx and range tables from 100 eV  to 100 TeV in 84 bins
    Lambda tables from threshold to 100 TeV, 7 bins/decade, spline: 1
    Sampling table 21x1001; from 1 GeV to 100 TeV
    ===== EM models for the G4Region  DefaultRegionForTheWorld ======
        muPairProd : Emin=    0 eV  Emax=  100 TeV
```
I'm not sure what we want to use here.

amandalund:
Thanks for pointing me to the TestEm17 example @whokion: we still need to validate all of these muon models, and it looks like this will be helpful.

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+    , inc_energy_(value_as<Energy>(particle.energy()))
+    , total_energy_(value_as<Energy>(particle.total_energy()))
+    , inc_mass_sq_(ipow<2>(value_as<Mass>(particle.mass())))
+    , electron_mass_(value_as<Mass>(shared.electron_mass))
+    , min_pair_energy_(4 * value_as<Mass>(shared.electron_mass))
+    , max_pair_energy_(inc_energy_
+                       + value_as<Mass>(particle.mass())
+                             * (1
+                                - real_type(0.75) * std::sqrt(constants::euler)
+                                      * element.cbrt_z()))
```
sethrj, line 159: How is this calculated? (Could be good to add to the class documentation.)

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+    real_type pair_energy
+        = inc_energy_ * std::exp(coeff_ * this->sample_scaled_energy(rng));
+    CELER_ASSERT(pair_energy >= min_energy_ && pair_energy <= max_pair_energy_);
+
+    // Sample the energy partition between the electron and positron
+    real_type r_max = (1
+                       - 6 * inc_mass_sq_
+                             / (total_energy_ * (total_energy_ - pair_energy)))
+                      * std::sqrt(1 - min_pair_energy_ / pair_energy);
+    real_type r = UniformRealDistribution(-r_max, r_max)(rng);
```
sethrj, line 202: Also would be good to add to the documentation :)

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+    CELER_ASSERT(iter != indices.end());
+    size_type idx = iter - indices.begin();
+    CELER_ASSERT(idx > 0);
+
+    // Calculate the y value corresponding to the sampled CDF value
+    real_type y_upper = y_grid[idx];
+    real_type y_lower = y_grid[idx - 1];
+    real_type cdf_lower = calc_cdf(y_lower);
+    real_type frac = (cdf - cdf_lower) / (calc_cdf(y_upper) - cdf_lower);
+    return fma<real_type>(frac, y_upper - y_lower, y_lower);
```
sethrj: If this has an ambiguous overload, you can use `std::fma`. The `fma` in `math/Algorithms` is meant for generic use inside array/matrix operations.

Better yet, though:
```c++
    return LinearInterpolator<real_type>{
        {calc_cdf(y_grid[idx - 1]), y_grid[idx - 1]},
        {calc_cdf(y_grid[idx]), y_grid[idx]}}(cdf);
```

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the scaled energy for a given Z grid and sampled CDF value.
+ */
+CELER_FUNCTION real_type
+MuPPEnergyDistribution::calc_scaled_energy(size_type z_idx, real_type u) const
+{
+    CELER_EXPECT(z_idx < table_.grids.size());
```
sethrj, line 237:
```suggestion
    CELER_EXPECT(z_idx < table_.grids.size());
    CELER_EXPECT(u >= 0 && u < 1);
```

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+
+    TwodGridData const& cdf_grid = table_.grids[ItemId<TwodGridData>(z_idx)];
+    auto calc_cdf
+        = TwodGridCalculator(cdf_grid, table_.reals)(std::log(inc_energy_));
+
+    // Get the sampled CDF value between the y bounds
+    real_type cdf_min = calc_cdf(y_min_);
+    real_type cdf = std::fma(calc_cdf(y_max_) - cdf_min, u, cdf_min);
```
sethrj: using `corecel/grid/Interpolator.hh`:
```suggestion
    real_type cdf = LinearInterpolator<real_type>{{0, calc_cdf(y_min_)}, {1, calc_cdf(y_max_)}}(u);
```
This compiles down to the same code and is more expressive.

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+ * Sample the scaled energy and interpolate in log Z.
+ */
+template<class Engine>
+CELER_FUNCTION real_type
+MuPPEnergyDistribution::sample_scaled_energy(Engine& rng) const
+{
+    real_type u = generate_canonical(rng);
+    real_type y_lower = this->calc_scaled_energy(logz_interp_.index, u);
+    real_type y_upper = this->calc_scaled_energy(logz_interp_.index + 1, u);
+    return y_lower + (y_upper - y_lower) * logz_interp_.fraction;
```
sethrj: Or:
```suggestion
    LinearInterpolator<real_type> interp_energy{
        {0, this->calc_scaled_energy(logz_interp_.index, u)},
        {1, this->calc_scaled_energy(logz_interp_.index + 1, u)}};
    return interp_energy(logz_interp_.fraction);
```

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
```diff
@@ -0,0 +1,249 @@
...
+    TwodGridData const& cdf_grid = table_.grids[ItemId<TwodGridData>(z_idx)];
+    auto calc_cdf
+        = TwodGridCalculator(cdf_grid, table_.reals)(std::log(inc_energy_));
+
+    // Get the sampled CDF value between the y bounds
+    real_type cdf_min = calc_cdf(y_min_);
+    real_type cdf = std::fma(calc_cdf(y_max_) - cdf_min, u, cdf_min);
+
+    // Find the grid index of the sampled CDF value
+    NonuniformGrid y_grid(cdf_grid.y, table_.reals);
```
sethrj: I wonder if there's a nice way to abstract the rest of the function into a helper that takes `<class GridT, class CalcCdf>`... could make the code more testable and reusable?

src/celeritas/em/interactor/MuPairProductionInteractor.hh
sethrj: Looks like the key assumptions in this distribution are:
- small angle scatter
- ultrarelativistic
- small/no recoil from muon?

If there's a muon brems reference where this is derived, we could just name it after that person(s)...

src/celeritas/em/model/MuPairProductionModel.cc
```diff
@@ -0,0 +1,177 @@
...
+ */
+void MuPairProductionModel::build_table(
+    ImportMuPairProductionTable const& imported, HostTable* table) const
+{
+    CELER_EXPECT(imported);
+    CELER_EXPECT(table);
+
+    // Build 2D sampling table
+    TwodGridBuilder build_grid{&table->reals};
+    auto grids = make_builder(&table->grids);
```
sethrj: Since C++17 we've been able to do
```suggestion
    CollectionBuilder grids{&table->grids};
```

src/celeritas/em/model/MuPairProductionModel.cc
```diff
@@ -0,0 +1,177 @@
...
+        CELER_VALIDATE(pvec,
+                       << "invalid grid in sampling table for '"
+                       << this->description() << "'");
+
+        // Normalize the CDF
+        std::vector<double> cdf(pvec.value.size());
+        size_type num_y = pvec.y.size();
+        for (size_type i : range(pvec.x.size()))
+        {
+            double norm = 1 / pvec.value[(i + 1) * num_y - 1];
```
sethrj: Another good place for an indexer?

sethrj (commented):
Looks good! Just a few uses of linterp perhaps, and some documentation updates?

doc/implementation/em-physics.rst
whokion: This threshold seems arbitrary anyway (anything greater than 2 x m_e) as the restricted cross section for e+e- pair production for the low energy muon (say below 1 GeV) is very small anyway.  So, to be consistent with Geant4, we may just use 0.85 GeV or 1 GeV (the low limit of the sampling table) for the process/model limit.

src/celeritas/em/interactor/MuPairProductionInteractor.hh
whokion: As I look Geant4, this distribution is only used for muon brems and the muon pair-production model (in reality, it is hard to imagine that the pair production angular distribution follows the same distribution of brems).  Anyway, taking into account the comment of @sethrj , we may name it to MuAngularDistributionVI (or MuAngularDistributionBogdanovVI) as the Geant4 author of the class is our own VI - even though the new name may be no better than the current one (so, we may also simply leave the name as it is).

amandalund: The reference they point to is [this paper](https://cds.cern.ch/record/1020037/files/cer-002679298.pdf) on muon interactions in Geant4. They mention the brems angular distribution briefly (Eq. 5) but just say it's a simplified probability function. I kind of do like just `MuAngularDistribution` better.

test/celeritas/data/four-steel-slabs.root
amandalund: Yeah the muppet is pretty hefty... 21 log E x 1001 log e / E x 5 Z.

sethrj: For testing purposes we don't need that high a table density. Maybe we could add an option to trim imported grid data for testing, then we could write that out instead?

amandalund: That could be a nice option to have... would it trim all the physics vectors? Just this table?

sethrj: Maybe all to the extent that we can: we don't need high-grain resolution for the unit tests anyway. I've got a draft that I based on the `RootJsonDumperTest` so maybe I'll get that in over the long weekend :)

amandalund: Sweet, thanks!

src/celeritas/em/distribution/MuPPEnergyDistribution.hh
amandalund: Done in #1525.

amandalund:
Thanks for the feedback @sethrj! I think I've addressed everything now.

sethrj:
@amandalund I think if you merge upstream and re-export the data, we should now have a less preposterous MuPPET size.

sethrj (approved):
Nice, thanks!

sethrj:
@amandalund Feel free to merge if you're not waiting on @stognini for feedback.

stognini:
Sorry for going slowly through the PR, but it looks pretty good to me.

amandalund:
No prob @stognini, I'll wait until you're through.

src/celeritas/em/process/MuPairProductionProcess.cc
```diff
@@ -0,0 +1,71 @@
...
+    return imported_.step_limits(std::move(applic));
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Name of the process.
+ */
+std::string_view MuPairProductionProcess::label() const
+{
+    return "Muon pair production";
```
stognini: Maybe "Muon electron-positron pair production"?

stognini (approved):
Looks great to me. Thanks @amandalund !

PR #1519, opened 2024-11-25
Clean up optical mock test data

hhollenb:
# Clean up optical mock test data

This PR cleans up the mock test data classes that I wrote previously.

 * The `MockImportedData` class is replaced with `OpticalMockTestBase` which inherits from `GlobalTestBase`, matching mock data for core physics tests.
 * Checks that didn't rely on mock data were moved to `ValidationUtils.hh` so they can be reused for later `optical::PhysicsData` tests.
 * Added more explicit unit annotations to test data
 * Simplified the hard-coded mock data, and made it accessible as `ImportData`.


test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+{
+using namespace ::celeritas::test;
+//---------------------------------------------------------------------------//
+/*!
+ * Helper function to annotate units of a hard-coded test data array.
+ *
+ * Converts the arguments supplied in units \c UnitType to native units.
+ */
+template<class UnitType, class... Args>
+std::array<real_type, sizeof...(Args)> constexpr expressed_as(Args const&... args)
```
sethrj: It might be more consistent to use `celeritas::Array`? Also I'd change the name to be something closer to the existing `native_value_from` so that if you know how the Quantity class works you know how this function works. What about `native_array_from`?

test/celeritas/optical/RayleighMfpCalculator.test.cc
```diff
@@ -68,40 +34,44 @@ class RayleighMfpCalculatorTest : public MockImportedData
...
+                                        12005096.767467,
+                                        6888377.4406869,
+                                        3466623.2384762,
+                                        1189584.7068151,
+                                        277.60444893823,
+                                        11510.805603078,
+                                        322.70360179716,
+                                        4.230373664558,
+                                        12005096.767467,
+                                        2801.539271218);
```
sethrj: Adding a trailing comma will make clang-format indent this more reliably...
```suggestion
                                        2801.539271218,);
```

test/celeritas/optical/ValidationUtils.cc
```diff
@@ -0,0 +1,111 @@
...
+using namespace ::celeritas::test;
+//---------------------------------------------------------------------------//
+/*!
+ */
+void check_physics_vector(ImportPhysicsVector const& expected,
+                          ImportPhysicsVector const& actual)
+{
+    EXPECT_EQ(expected.vector_type, actual.vector_type);
+    EXPECT_VEC_EQ(expected.x, actual.x);
+    EXPECT_VEC_EQ(expected.y, actual.y);
```
sethrj: The problem with this is that errors now look like they're coming from this file instead of the source where the test occurs. Another issue is that you're using exact equality for floating point values. If you take a look at `test/TestMacros.hh` you can see how to correctly do this with googletest (and the implementation is in `test/testdetail/TestMacrosImpl.hh`). To make life easier, I think you should just overload the `IsVecSoftEquiv` function,  declaring it inside ValidationUtils.hh
```c++
::testing::AssertionResult IsVecSoftEquiv(char const* expected_expr,
                                          char const* actual_expr,
                                          ImportPhysicsVector const& expected,
                                          ImportPhysicsVector const& actual)
```

test/celeritas/optical/OpticalMockTestBase.hh
```diff
@@ -0,0 +1,77 @@
...
+    SPConstMaterial build_material() override;
+
+    // Access mock imported data
+    ImportData const& imported_data() const;
+
+    // Retrieve imported optical model data by class
+    ImportOpticalModel const& import_model_by_class(ImportModelClass) const;
+
+    //! Number of mock optical materials
+    inline OpticalMaterialId::size_type num_optical_materials() const
```
sethrj: No `inline` needed here

test/celeritas/optical/OpticalMockTestBase.cc
```diff
@@ -0,0 +1,281 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Create mock imported data in-place.
+ */
+void OpticalMockTestBase::build_import_data(ImportData& data) const
+{
+    // Build mock imported optical materials
+    {
+        data.optical_materials = std::vector<ImportOpticalMaterial>(5);
```
sethrj:
```suggestion
        data.optical_materials.resize(5);
```

test/celeritas/optical/OpticalMockTestBase.cc
```diff
@@ -0,0 +1,281 @@
...
+            = physics_vector_expressed_as<ElectronVolt, units::Native>(
+                {1.098177, 6.812319}, {1.3235601610672, 1.4679465862259});
+        data.optical_materials[4].rayleigh.scale_factor = 1.7;
+        data.optical_materials[4].rayleigh.compressibility
+            = native_value_from(Quantity<MeterCubedPerMeV>{4.213e-24});
+    }
+
+    // Build mock imported optical models
+    {
+        data.optical_models = std::vector<ImportOpticalModel>(3);
```
sethrj:
```suggestion
        data.optical_models.resize(3);
```

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Helper function to annotate units of hard-coded test data.
+ *
+ * Same as \c expressed_as, except returns a vector.
+ */
+template<class UnitType, class... Args>
+std::vector<real_type> vector_expressed_as(Args const&... args)
```
sethrj: `native_vector_from`?

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+    void check_span(Span<real_type const> const& t,
+                    ItemRange<real_type> const& real_ids,
+                    Softness soft);
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * A \c GridValidator that stores its own collections.
+ */
+class GridStorage : public GridValidator
```
sethrj: This should be something like `OwningGridValidator`. Its purpose isn't to store; its purpose is validating with local storage.

test/celeritas/optical/OpticalMockTestBase.cc
```diff
@@ -0,0 +1,281 @@
...
+// UNITS
+//---------------------------------------------------------------------------//
+struct ElectronVolt
+{
+    static CELER_CONSTEXPR_FUNCTION real_type value()
+    {
+        return units::Mev::value() / 1e6;
+    }
+    static char const* label() { return "eV"; }
+};
```
sethrj: I think we should put this in `celeritas/UnitTypes.hh` before MeV

```
//! Atom-scale energy
struct ElectronVolt
{
    static CELER_CONSTEXPR_FUNCTION real_type value()
    {
#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
        return units::megaelectronvolt / real_type{1e6};
#else
        return constants::e_electron * units::volt;
#endif
    }
    static char const* label() { return "eV"; }
};

//! Nucleus-scale energy
struct Mev
{
    static CELER_CONSTEXPR_FUNCTION real_type value()
    {
#if CELERITAS_UNITS == CELERITAS_UNITS_CLHEP
        return units::megaelectronvolt;
#else
        return real_type(1e6) * ElectronVolt::value;
#endif
    }
    static char const* label() { return "MeV"; }
};

```

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+ * Convert a floating point type collection into a \c real_type span.
+ *
+ * Used to convert data which may be expressed as double or single precision
+ * into the same size as \c real_type. If the supplied data is the same
+ * precision, then just a span to the original data is returned. Otherwise,
+ * it is copied into a local buffer and casted to the correct precision. The
+ * \c RealSpanGenerator object should live as long as its spans are being
+ * used in the tests.
+ */
+class RealSpanGenerator
```
sethrj: Is this class used anywhere? I think we may be spending too much effort supporting single precision. I'm actually fine if the physics tests pass only with double; the single precision mainly helps elucidate edge cases in low-level math functions (rounding error) and geometry. Eventually we may want to switch to separate "setup time" and "runtime" precision levels so that most interfaces are doubles.

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+    using Items = Collection<T, Ownership::value, MemSpace::host>;
+    //!@}
+
+    //! Whether to use soft or exact equivalence
+    enum Softness
+    {
+        Soft,
+        Exact
+    };
+
```
sethrj: I definitely don't like this. If anything, make it a templated function with a comparator that could use SoftEqual or std::equal .

sethrj (changes requested):
Thanks for the refactor! I see that a lot of the code modified here is from #1440, which was merged before we resolved some of the comments. (And because I was rushing through the review on travel I didn't even get to look at the testing code there.) I think a lot of these changes are a step in the right direction, but there was a lot of testing code in that PR that needed further revision. Here are some initial comments for now but I think some of the helper classes/function need more work.

test/celeritas/optical/ValidationUtils.cc
hhollenb: I was using exact equality here just to check that imported physics vectors get forwarded to adapters / models correctly. Since they're only copies and references of doubles to doubles, I believe exact equality should be the desired behavior.

test/celeritas/optical/RayleighMfpCalculator.test.cc
hhollenb: Because it's a function argument it doesn't support trailing commas. I added an option for mapping `Array`s so that a brace-initializer can be passed as an argument, which has better clang-format indenting.

test/celeritas/optical/ValidationUtils.hh
hhollenb: After peeking at some of the test implementations, I think I managed to refactor this out entirely now.

test/celeritas/optical/ValidationUtils.hh
hhollenb: I ended up not using soft equality, so I've removed this enum.

hhollenb:
Did some refactoring to use custom googletest `EXPECT` implementations. I added `EXPECT_GRID_EQ` and `EXPECT_TABLE_EQ` to do checks on data built in collections, with a `PhysicsGridTraits` to allow comparing `ImportPhysicsVector`s, tuples of `Span`s, etc. 

The `GridValidator` has been changed to just be a `GridAccessor` to have consistent access of built grid data, and then tests are responsible for how they want to check that data.

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,136 @@
...
+{
+using namespace ::celeritas::test;
+//---------------------------------------------------------------------------//
+/*!
+ * Helper function to annotate units of a hard-coded test data array.
+ *
+ * Converts the arguments supplied in units \c UnitType to native units.
+ */
+template<class UnitType, class... Args>
+Array<real_type, sizeof...(Args)> constexpr native_array_from(Args const&... args)
```
sethrj, line 48: These are actually really cool. At some point (not this PR) we should use this in our other celeritas tests as well.

test/celeritas/optical/ValidationUtils.hh
```diff
@@ -0,0 +1,136 @@
...
+    return Array<real_type, sizeof...(Args)>{
+        native_value_from(UnitType(args))...};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Helper function to annotate units of a hard-coded test data array.
+ *
+ * This overload accepts fixed size \c Arrays which clang-format handles
+ * better.
```
sethrj: This (and the detail) are also neat, but I think it should be removed even if the format is ugly (which we can always suppress with the `// clang-format: off`) for two reasons:
1. The manually specified `N` will silently make the array bigger and fill it with zeros if it's greater than the number of actual entries given
2. You have to go through several extra hoops (forward declaring the function in the impl file) to get it to work.

So I'd say just kill this function and replace its single use in RayleighMfp with the one above.

sethrj (approved):
Awesome! Very sleek, thanks for the cleanup.

test/celeritas/optical/ValidationUtils.hh
sethrj: P.S. I know I was the one who recommended the trailing comma to you (but at the time I didn't see that it was a function call rather than brace initialization). So I'm sorry about the misdirection!

hhollenb: All good - when I wrote this it had a bit of code smell to me, and just manually formatting is definitely more appropriate.

PR #1522, opened 2024-11-26
Udate upload-artifact and download-artifact Github Actions

esseivaju:
This was probably missed during an update, as everywhere else already uses `upload-artifact@v4`. [V3 will be deprecated soon](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/).

### DIFF
--- a/.github/workflows/pull_request_completed.yml
+++ b/.github/workflows/pull_request_completed.yml
@@ -18,11 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Download and Extract Artifacts
-      uses: dawidd6/action-download-artifact@e7466d1a7587ed14867642c2ca74b5bcc1e19a2d # v6
+      uses: actions/download-artifact@v4
       with:
-        name: "(event-file|test-results-.*)"
-        name_is_regexp: true
-        run_id: ${{github.event.workflow_run.id}}
+        pattern: "(event-file|test-results-.*)"
+        merge-multiple: true
+        run-id: ${{github.event.workflow_run.id}}
         path: artifacts
     - name: Publish PR comment
       uses: EnricoMi/publish-unit-test-result-action@82082dac68ad6a19d980f8ce817e108b9f496c2a # v2.17.1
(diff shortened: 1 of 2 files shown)


sethrj:
@esseivaju ~~it doesn't work on the docker images because their OS is too old for the Node.js version.~~ I guess v4 didn't work on ROCm in particular and now I've had to disable it. As I said I was going to update those images this week and the docker CI script that corresponds to it.

.github/workflows/pull_request_completed.yml
```diff
@@ -18,11 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Download and Extract Artifacts
-      uses: dawidd6/action-download-artifact@e7466d1a7587ed14867642c2ca74b5bcc1e19a2d # v6
+      uses: actions/download-artifact@v4
```
pcanal, line 21: The 2 are not quite the same. The former allows to `download this artifact in another workflow that is run after the first one.`.  Do we need that feature?

sethrj: Yes we do.

esseivaju: `download-artifact@v4`  should now [support](https://github.com/actions/download-artifact/tree/v4?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories) this as well

esseivaju:
> @esseivaju ~it doesn't work on the docker images because their OS is too old for the Node.js version.~ I guess v4 didn't work on ROCm in particular and now I've had to disable it. As I said I was going to update those images this week and the docker CI script that corresponds to it.

Sure, I guess you can update everything in one go with the new images then.

.github/workflows/pull_request_completed.yml
sethrj: ooh good to know!

sethrj:
By the way @esseivaju , all changes to `pull_request_completed` have to be tested on `develop` on a personal repo before deploying to celeritas: those changes won't be tested as part of the PR (since it's the PR action on the *develop* branch that gets triggered), so I'd be happy to have those separate from the CI update.

esseivaju:
@sethrj Sounds good; I'll test the update to `pull_request_completed` and re-open a new PR later.

PR #1524, opened 2024-11-28
Require clang-tidy in CI

esseivaju:
We've had a clang-tidy job for a while, but it wasn't required to succeed for the CI pipeline to pass. This PR makes it required: if we ever want to require it, the sooner the easier it will be. 

I silenced most of the remaining warnings as they were inconsequential, except for a few suspicious cases where I fixed the warning, if they were fine they'd warrant a comment. In particular, fix a potential memory leak in `CelerOpticalPhysics` if a process is inactive.


.github/workflows/build-spack.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           - geometry: "geant4"
             special: null
             geant: "11.0"
-          - geometry: "orange"
+          - geometry: "vecgeom"
```
sethrj, line 44: Does ORANGE not pass?

.github/workflows/build-spack.yml
```diff
@@ -139,9 +138,12 @@ jobs:
...
         id: build
         working-directory: build
-        continue-on-error: ${{matrix.special == 'clang-tidy'}}
         run: |
-          ninja -v -k0
+          if [ "${{matrix.special}}" == "clang-tidy" ]; then
+            ninja -k0
+          else
+            ninja -v -k0
+          fi
```
sethrj, line 146: Why don't we want the verbose line for clang-tidy? (Or should we turn it off for the other builds too?)

.github/workflows/build-spack.yml
```diff
@@ -184,18 +186,19 @@ jobs:
...
       - name: Install
         id: install
-        if: ${{!cancelled() && steps.build.outcome == 'success'}}
+        if: ${{!cancelled() && steps.build.outcome == 'success' && matrix.special != 'clang-tidy'}}
         working-directory: build
         run: |
           ninja install
       - name: Check installation
-        if: ${{steps.install.outcome == 'success'}}
+        if: ${{steps.install.outcome == 'success' && matrix.special != 'clang-tidy'}}
```
sethrj: Since the install step gets skipped I don't think this is necessary

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -54,6 +54,46 @@ enum class OpticalProcessType
     size_,
 };
 
+/*!
+ * Wrapper around a unique pointer to accomodate keeping track whether we
+ * deleguated ownership to Geant4. We have to assume that Geant4 won't free the
```
sethrj:
```suggestion
 * delegated ownership to Geant4. We have to assume that Geant4 won't free the
```

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -239,13 +279,13 @@ void CelerOpticalPhysics::ConstructProcess()
 
     // NB: boundary is also used later on in loop over particles,
     // though it's only ever applicable to G4OpticalPhotons
-    auto boundary = std::make_unique<G4OpBoundaryProcess>();
+    auto boundary = UniquePtrOwner{std::make_unique<G4OpBoundaryProcess>()};
```
sethrj: Instead of the `UniquePtrOwner` which has unusual semantics, what about:
```suggestion
    auto boundary_storage = std::make_unique<G4OpBoundaryProcess>();
    auto* boundary = boundary_storage.get();
```
and then `boundary.get()` becomes `boundary.release()`?

Or use my updated class suggestion of the class which is somewhere between our two suggestions.

src/celeritas/ext/detail/CelerOpticalPhysics.cc
```diff
@@ -54,6 +54,46 @@ enum class OpticalProcessType
...
+        if (!is_owner_)
+        {
+            // NOLINTNEXTLINE(bugprone-unused-return-value)
+            ptr_.release();
+        }
+    };
+
+  private:
+    std::unique_ptr<T> ptr_;
+    bool is_owner_{true};
```
sethrj:
```suggestion
template<class T>
class ObservingUniquePtr
{
  public:
    explicit ObservingUniquePtr(std::unique_ptr<T> ptr)
        : uptr_(std::move(ptr)), ptr_{uptr_.get()}
    {}

    CELER_DELETE_COPY_DEFAULT_MOVE(ObservingUniquePtr);

    T* release() noexcept { return uptr_.release(); }
    operator T*() const noexcept { return ptr_; }
    T* operator->() const noexcept { return ptr_; }

  private:
    std::unique_ptr<T> uptr_;
    T* ptr_;
```

I'm guessing by the nomenclature and trailing `;` after the function definitions that you copied this from ATLAS? 😉

src/celeritas/geo/GeoMaterialParams.cc
```diff
@@ -40,9 +40,11 @@ using MapLabelMatId = std::unordered_map<Label, MaterialId>;
  * The input is effectively an "unzipped" unordered list of (volume label,
  * material id) pairs.
  */
-MapLabelMatId build_label_map(MaterialParams const& mat_params,
-                              std::vector<Label>&& labels,
-                              std::vector<MaterialId> const& materials)
+MapLabelMatId build_label_map(
+    MaterialParams const& mat_params,
+    std::vector<Label>&&
+        labels,  // NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
```
sethrj: Maybe just change this to const ref since it's a local function and we don't actally need the labels. (And we can still `std::move` into a const reference)

src/celeritas/optical/OpticalCollector.cc
```diff
@@ -34,6 +34,7 @@ namespace celeritas
  *
  * This adds several actions and auxiliary data to the registry.
  */
+// NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved)
```
sethrj: If we round out the function with `inp = {}` does that silence the warning?

src/celeritas/user/SlotDiagnostic.cc
```diff
@@ -32,7 +32,16 @@ struct SlotDiagnostic::State final : AuxStateInterface
     std::ofstream outfile;
     std::vector<int> buffer;
 
-    ~State() { CELER_LOG_LOCAL(debug) << "Closing slot diagnostic file"; }
+    State() = default;
+    State(State const&) = delete;
+    State& operator=(State const&) = delete;
+    State(State&&) noexcept = default;
+    State& operator=(State&&) noexcept = default;
```
sethrj:
```suggestion
    CELER_DEFAULT_MOVE_DELETE_COPY(State);
```

src/corecel/io/ColorUtils.cc
```diff
@@ -86,6 +86,7 @@ char const* color_code(char abbrev)
     if (!use_color())
         return "";
 
+    // NOLINTNEXTLINE(bugprone-switch-missing-default-case)
```
sethrj: Maybe better would be moving the terminal `return ""` into a `default`?

src/corecel/io/LoggerTypes.cc
```diff
@@ -40,7 +40,7 @@ char const* to_color_code(LogLevel lev)
     char c = ' ';
     switch (lev)
     {
-        case LogLevel::debug:      c = 'x'; break;
+        case LogLevel::debug:
```
sethrj: Will this need to be annotated with `[[fallthrough]]` to stop some compilers complaining?

src/corecel/sys/ScopedMem.cc
```diff
@@ -116,7 +118,7 @@ ScopedMem::ScopedMem(std::string_view label, MemRegistry* registry)
 /*!
  * Register data on destruction.
  */
-ScopedMem::~ScopedMem()
+ScopedMem::~ScopedMem()  // NOLINT(bugprone-exception-escape)
```
sethrj: Does this just need a `noexcept`?

src/corecel/sys/TracingSession.cc
```diff
@@ -48,7 +48,6 @@ initialize_session(TracingMode mode) noexcept
             case TracingMode::in_process:
                 return perfetto::kInProcessBackend;
             case TracingMode::system:
-                return perfetto::kSystemBackend;
```
sethrj, line 51: Same question about fallthrough

src/geocel/GeoParamsInterface.hh
```diff
@@ -127,7 +127,7 @@ class GeoParamsSurfaceInterface : public GeoParamsInterface
 
   public:
     // Default destructor
-    virtual ~GeoParamsSurfaceInterface() = 0;
+    ~GeoParamsSurfaceInterface() override = 0;
```
sethrj, line 130: Could we just delete this since it inherits?

src/geocel/GeoVolumeFinder.cc
```diff
@@ -16,6 +16,7 @@ namespace celeritas
 /*!
  * Perform the search.
  */
+// NOLINTNEXTLINE(bugprone-exception-escape)
```
sethrj: What is the problem here?

src/orange/g4org/SolidConverter.cc
```diff
@@ -186,14 +186,15 @@ auto make_shape(G4VSolid const& solid, Args&&... args)
...
                 std::optional<CR>&& excluded,
                 SolidEnclosedAngle&& enclosed)
 {
-    return Solid<CR>::or_shape(std::string{solid.GetName()},
-                               std::move(interior),
-                               std::move(excluded),
-                               std::move(enclosed));
+    return Solid<CR>::or_shape(
+        std::string{solid.GetName()},
+        std::move(interior),  // NOLINT(bugprone-move-forwarding-reference)
```
sethrj: Maybe this actually should be
```suggestion
        std::forward(interior),
```
and then the other NOLINT goes away too?

src/orange/surf/GeneralQuadric.cc
```diff
@@ -44,6 +44,7 @@ GeneralQuadric::GeneralQuadric(Real3 const& abc,
 /*!
  * Promote from a simple quadric.
  */
+// NOLINTNEXTLINE(bugprone-exception-escape)
```
sethrj, line 48: Took me a second to figure this out; but since the GQ constructor it dispatches to *can* throw an exception then perhaps it's better to just drop the `noexcept` from this constructor.

src/orange/surf/SimpleQuadric.cc
```diff
@@ -48,6 +48,7 @@ SimpleQuadric::SimpleQuadric(Real3 const& abc, Real3 const& def, real_type g)
  * whereas the simple quadric has a different sign for the constant:
  * \f$ dx + ey + fz + g = 0 \f$ .
  */
+// NOLINTNEXTLINE(bugprone-exception-escape)
```
sethrj, line 52: Same here and below.

app/celer-dump-data.cc
```diff
@@ -968,7 +968,7 @@ void print_optical_materials(std::vector<ImportOpticalModel> const& io_models,
 /*!
  * Execute and run.
  */
-int main(int argc, char* argv[])
+int main(int argc, char* argv[])  // NOLINT(bugprone-exception-escape)
```
sethrj, line 971: Better would be to move the main body to a `run` function and wrap with try/catch I think... this is fine for now though.

sethrj (approved):
Thanks @esseivaju ! It's nice to finally have this in place 😄 

I think given the number of "false" positives for `cppcoreguidelines-rvalue-reference-param-not-moved` and `cppcoreguidelines-missing-std-forward` we should just disable those.

.github/workflows/build-spack.yml
esseivaju: It passes but VecGeom compiles more files, and I think is a superset of files compiled with ORANGE, so we might as well switch.

.github/workflows/build-spack.yml
esseivaju: Since it'd be required to fix the clang-tidy checks, verbose makes the build logs hard to parse since it adds a lot of extra noise. Removing it makes it much easier to find which file triggered the warning. It's probably fine to keep it for the other jobs.

src/celeritas/optical/OpticalCollector.cc
esseivaju: No, it doesn't seem to silence it.

src/corecel/io/LoggerTypes.cc
esseivaju: Mmh, I guess it could help to add it.

src/geocel/GeoParamsInterface.hh
esseivaju: I think this was here to anchor the vtable

src/corecel/sys/ScopedMem.cc
esseivaju: I think destructor should be `noexcept` by default. It needs:
```c++
 //NOLINTNEXTLINE(cppcoreguidelines-noexcept-destructor)
ScopedMem::~ScopedMem() noexcept(!CELERITAS_DEBUG)
```

src/orange/surf/GeneralQuadric.cc
esseivaju: It'd only throw in debug mode because of `CELER_EXPECT`? So maybe `noexcept(!CELER_DEBUG)`?

src/geocel/GeoVolumeFinder.cc
esseivaju: I guess it can throw some debug asserts?

esseivaju:
> I think given the number of "false" positives for `cppcoreguidelines-rvalue-reference-param-not-moved` and `cppcoreguidelines-missing-std-forward` we should just disable those.

I enabled `AllowPartialMove` which makes the check more lenient. There's only a handful left now. Since this is an unusual way of passing parameters (per [F.15](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-conventional)), I think these checks could be useful, it makes you think twice about it, and future readers know that it was deliberate and not an oversight.

src/corecel/sys/ScopedMem.cc
esseivaju: `clang-tidy-15` doesn't respect the `noexcept(false)`, this work on my local build.

app/celer-dump-data.cc
```diff
@@ -624,7 +624,7 @@ void print_trans_params(ImportTransParameters const& trans_params,
     for (auto const& kv : trans_params.looping)
     {
         auto pid = particles.find(PDGNumber{kv.first});
-        auto par = particles.id_to_label(pid);
+        auto const& par = particles.id_to_label(pid);
```
sethrj, line 627: Aha thanks, this one is quite useful; I noticed a couple places in the data dumping that needed attention.

src/corecel/sys/Stream.cc
```diff
@@ -209,8 +208,7 @@ Stream::Stream(Stream&& other) noexcept
  */
 Stream& Stream::operator=(Stream&& other) noexcept
 {
-    Stream temp(std::move(other));
-    this->swap(temp);
+    this->swap(other);
```
sethrj: This isn't quite  the same thing; before, the incoming `other` gets reset to a blank state, and now it gets the original state of the LHS. I know `move` is supposed to produce some sort of unspecificed state but I feel like clearing the RHS of the move (the original behavior) is better. Did this trigger a warning?

sethrj (approved):

src/orange/surf/GeneralQuadric.cc
sethrj: Yeah exactly.

src/geocel/GeoVolumeFinder.cc
sethrj: Yep righto.

src/geocel/GeoParamsInterface.hh
sethrj: Ah indeed, forgot I added that. Should've commented in the header file too...

src/corecel/sys/Stream.cc
esseivaju: Right, blank state is better. This didn't trigger a warning; the other constructor did, and I noticed that extra temporary in the move assignment. I figured since move semantics are supposed to be an optimization of copy, I wanted to remove it but the compiler might be able to do it, I'll revert this.

PR #1525, opened 2024-11-28
Add a helper class for inverse transform sampling

amandalund:
Pulling this out from #1518: this adds a little class to find the interpolated grid value corresponding to a sampled CDF value to help with inverse transform sampling from a tabulated CDF.


src/celeritas/grid/InverseCdfFinder.hh
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Both the input grid and the CDF must be monotonically increasing. The
+ * sampled CDF value must be in range.
+ */
+template<class G, class C>
+class InverseCdfFinder
+{
+  public:
+    // Construct from grid and CDF calculator
+    inline CELER_FUNCTION InverseCdfFinder(G const& grid, C&& calc_cdf);
```
sethrj: Since the grid is always a lightweight object maybe this should be `G&&` as well? That way in tests we can do `InverseCdfFinder{UniformGrid{blah}, calc_cdf};`

test/celeritas/grid/InverseCdfFinder.test.cc
```diff
@@ -0,0 +1,56 @@
...
+    std::vector<real_type> cdf{0.0, 0.1, 0.2, 0.8, 0.9, 1.0};
+    InverseCdfFinder find(grid, [&cdf](size_type i) { return cdf[i]; });
+
+    EXPECT_SOFT_EQ(0, find(0));
+    EXPECT_SOFT_EQ(0.1, find(0.05));
+    EXPECT_SOFT_EQ(0.2, find(0.1));
+    EXPECT_SOFT_EQ(1, find(0.5));
+    EXPECT_SOFT_EQ(6.0, find(0.9));
+    EXPECT_SOFT_EQ(8.0, find(0.95));
+    EXPECT_SOFT_EQ(9.99996, find(0.999999));
```
sethrj, line 54:
```suggestion
    EXPECT_SOFT_EQ(9.99996, find(0.999999));
    if (CELERITAS_DEBUG)
    {
        EXPECT_THROW(find(1.0), DebugError);
    }
```

src/celeritas/grid/InverseCdfFinder.hh
```diff
@@ -0,0 +1,84 @@
...
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from grid and CDF calculator.
+ */
+template<class G, class C>
+CELER_FUNCTION
+InverseCdfFinder<G, C>::InverseCdfFinder(G const& grid, C&& calc_cdf)
+    : grid_(grid), calc_cdf_(celeritas::forward<C>(calc_cdf))
+{
```
sethrj, line 53: Should we check that the grid size is >= 2? That the front is zero and the back is soft equal to 1?

src/celeritas/grid/InverseCdfFinder.hh
```diff
@@ -0,0 +1,84 @@
...
+ */
+template<class G, class C>
+CELER_FUNCTION real_type InverseCdfFinder<G, C>::operator()(real_type cdf) const
+{
+    CELER_EXPECT(cdf >= 0 && cdf < 1);
+
+    // Find the grid index of the sampled CDF value
+    Range indices(grid_.size());
+    auto iter = celeritas::lower_bound(
+        indices.begin(), indices.end(), cdf, [this](size_type i, real_type c) {
```
sethrj, line 70: We should be able loop only through `grid_.size() - 1` right? And that way if we *do* sample a `1.0` in release mode we don't go out of range.

sethrj (approved):
Nifty, thanks!

src/celeritas/grid/InverseCdfFinder.hh
amandalund: ~We can check all except that the last CDF value is 1 since the `TwodSubgridCalculator` asserts that `y` is strictly less than `y_grid.back()`~ I added `operator[]` to the `TwodSubgridCalculator` which will let us check the max cdf value and simplify things elsewhere.

src/celeritas/grid/InverseCdfFinder.hh
amandalund: Yes, good point.

PR #1527, opened 2024-11-30
Trim export data for unit tests

sethrj:
Following on to [a discussion](https://github.com/celeritas-project/celeritas/pull/1518#discussion_r1856602784) on #1518, this reduces the size of the MuPP table and has options for trimming additional elements. I don't have it enabled by default since it messes with the spacing of the physics vectors (and we don't yet support nonuniform sizes). However I use it to trim the ROOT JSON export test. We could extend it to simplify the Geant export test as well.

### DIFF
--- a/app/celer-export-geant.cc
+++ b/app/celer-export-geant.cc
@@ -28,6 +28,7 @@
 #include "celeritas/ext/RootExporter.hh"
 #include "celeritas/ext/RootJsonDumper.hh"
 #include "celeritas/ext/ScopedRootErrorHandler.hh"
+#include "celeritas/io/ImportDataTrimmer.hh"
 
 namespace celeritas
 {
@@ -82,8 +83,15 @@ GeantPhysicsOptions load_options(std::string const& option_filename)
 void run(std::string const& gdml_filename,
          std::string const& opts_filename,
          std::string const& out_filename,
-         GeantImporter::DataSelection selection)
+         bool gen_test)
 {
+    // TODO: expose data selection to JSON users?
+    GeantImporter::DataSelection selection;
+    selection.particles = GeantImporter::DataSelection::em
+                          | GeantImporter::DataSelection::optical;
+    selection.processes = selection.particles;
+    selection.reader_data = !gen_test;
+
     // Construct options, set up Geant4, read data
     auto imported = [&] {
         GeantImporter import(
@@ -91,7 +99,21 @@ void run(std::string const& gdml_filename,
         return import(selection);
     }();
 
+    // TODO: expose trim data rather than bool 'gen_test'
+    if (gen_test)
+    {
+        ImportDataTrimmer::Input options;
+        options.max_size = 3;
+        ImportDataTrimmer trim(options);
+        trim(&imported);
+    }
+
     ScopedRootErrorHandler scoped_root_error;
+
+    if (gen_test)
+    {
+        CELER_LOG(info) << "Trimming data for testing";
+    }
 
     if (ends_with(out_filename, ".root"))
     {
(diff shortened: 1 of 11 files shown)


src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -0,0 +1,382 @@
...
+        {
+            result.insert(m->extract(cur_iter));
+        }
+    }
+    *m = std::move(result);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Whether to keep the data at this index.
```
amandalund: Update description?

src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -0,0 +1,382 @@
...
+    (*this)(&data->livermore_pe_data);
+
+    if (this->options_.physics)
+    {
+        for (auto&& kv : data->neutron_elastic_data)
+        {
+            (*this)(&kv.second);
+        }
+        // TODO: trim relaxation shells
+    }
```
amandalund: Can this go in the `if (options_.physics)` block above?

src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -0,0 +1,382 @@
...
+{
+    if (options_.materials)
+    {
+        (*this)(&data->isotopes);
+        (*this)(&data->elements);
+        (*this)(&data->geo_materials);
+        (*this)(&data->phys_materials);
+        (*this)(&data->optical_materials);
+        (*this)(&data->neutron_elastic_data);
+        (*this)(&data->atomic_relaxation_data);
```
amandalund: Should SB and Livermore PE be in here too?

src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -0,0 +1,382 @@
...
+
+    // TODO: remap IDs?
+}
+
+//---------------------------------------------------------------------------//
+void ImportDataTrimmer::operator()(ImportOpticalMaterial* data)
+{
+    CELER_EXPECT(data);
+
+    // TODO: trim WLS components?
```
amandalund: Or refractive index? I guess for now all the imported optical properties in the tests are defined by us in the GDML and are already trimmed or somewhat invented.

amandalund (approved):
Thanks @sethrj!

app/celer-export-geant.cc
```diff
@@ -105,7 +105,7 @@ void run(std::string const& gdml_filename,
         ImportDataTrimmer::Input options;
         options.mupp = true;
         ImportDataTrimmer trim(options);
-        trim(&imported);
+        trim(imported);
```
amandalund, line 108: Why the change from pointers to references (since our style is typically to pass pointers instead of mutable references when a value is going to be modified)?

sethrj: OK I should've talked with the group about this first... I think the *idea* of annotating arguments to be modified is good when it could be ambiguous, but I increasingly think pointers are the wrong way to do it. Google somewhat recently changed their style guide to use references instead of pointers (except for optional in/out or out-only values), and the standard library and friends don't generally use raw pointers like that. So I think we should remove that guideline from our code style and start think about using references instead... perhaps using some clang libtooling to help refactor our code?

amandalund: Sure, that seems sensible to me. Maybe we can add that as a todo in one of our issues?

src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -0,0 +1,376 @@
...
+        // Reduce the resolution of the muon pair production table
+        (*this)(data.mu_pair_production_data);
+    }
+
+    this->for_each(data.elements);
+    this->for_each(data.geo_materials);
+    this->for_each(data.phys_materials);
+
+    this->for_each(data.processes);
+    this->for_each(data.msc_models);
```
amandalund, line 78: @sethrj these will cause the `ProcessBuilder` test to fail and change the results in the MSC tests. What do you want to do about those?

sethrj: Uh oh, I clearly didn't think that through. Maybe all the `for_each` should live inside the `if (options_.physics)`? Or we need an option to trim "number of processes" vs "contents of each process"? I think the former should let you trim the mupp table without breaking anything else. Could you try that on your branch? Sorry about that.

PR #1530, opened 2024-12-02
Support building infix logic representation

esseivaju:
Refactor the `PostfixLogicBuilder` to support other types of logic notations. In particular, support for infix notation as part of #1282. `PostfixLogicBuilder` is now a free function called `build_logic` that is templated on a policy object dictating the notation to build. Any state is now held in the policy object instead of the builder object and `build_logic` provides common behavior (faces mapping). 

This simple implementation inserts parentheses around every single `Joined` operator even if they wouldn't be necessary. This also doesn't check that the tree is in a valid infix-compatible form (no `Negated {Joined{}}` node)


src/orange/orangeinp/detail/LogicBuilderPolicies.hh
```diff
@@ -69,21 +72,69 @@ class PostfixLogicBuilderImpl
...
+ LogicBuilder::operator() . The
+ * user invokes this class with a node ID (usually representing a cell), and
+ * then this class recurses into the daughters using a tree visitor.
+ *
+ * Example: \verbatim
+    all(1, 3, 5) -> {{1, 3, 5}, "(0 & 1 & 2)"}
+    all(1, 3, any(~(2), ~(4))) -> {{1, 2, 3, 4}, "(0 & 2 & (~1 | ~3))"}
+ * \endverbatim
+ */
+class InfixLogicBuilderPolicy : public PostfixLogicBuilderPolicy
```
esseivaju: I am not sure about inheritance vs independent implementation. We can avoid duplication by inheriting, but it can be error-prone as methods are not virtual, derived classes must make sure to reimplement base methods calling `operator()`, i.e. the NodeId, Negated, Joined, and Aliased overloads.

esseivaju (commented):
@sethrj I think this is ready now.

src/orange/orangeinp/UnitProto.cc
```diff
@@ -182,7 +185,9 @@ void UnitProto::build(ProtoBuilder& input) const
         VolumeInput vi;
 
         // Construct logic and faces with remapped surfaces
-        auto&& [faces, logic] = build_logic(node_id);
+        auto&& [faces, logic]
+            = build_logic.operator()<detail::PostfixLogicBuilderPolicy>(
```
sethrj: Yuck! Usually policy functions are called with an object so maybe we should have it in instantiate one here. (Maybe the policy instance rather than class can take the construction arguments, and maybe even the node ID since it's only called once.) If that won't work, then let's at least rename the function so we don't have to manually type `operator()`.

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
```diff
@@ -69,21 +72,69 @@ class PostfixLogicBuilderImpl
     inline void operator()(Joined const&);
     //!@}
 
+  protected:
+    VecLogic& logic() { return *logic_; }
+    ContainerVisitor<CsgTree const&, NodeId>& visite_node()
```
sethrj: Visit?

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
sethrj: What about private inheritance, CRTP, or composition?

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -103,10 +109,13 @@ TEST_F(CsgTreeUtilsTest, postfix_simplify)
 
     // Test postfix and internal surface flagger
     InternalSurfaceFlagger has_internal_surfaces(tree_);
-    PostfixLogicBuilder build_postfix(tree_);
+    LogicBuilder builder(tree_);
+    auto postfix_policy
+        = &LogicBuilder::operator()<detail::PostfixLogicBuilderPolicy>;
```
sethrj: I think it would be clear with just a lambda

src/orange/orangeinp/detail/LogicBuilder.cc
```diff
@@ -0,0 +1,89 @@
...
+    static_assert(std::is_constructible_v<LogicBuilderPolicy,
+                                          CsgTree const&,
+                                          VecSurface const*,
+                                          VecLogic*>);
+    static_assert(std::is_invocable_v<LogicBuilderPolicy, NodeId>);
+
+    // Construct logic vector as local surface IDs
+    VecLogic lgc;
+    LogicBuilderPolicy build_impl{tree_, mapping_, &lgc};
+    build_impl(n);
```
sethrj: This is where I wonder if we could take a `Policy` instance as an argument and call `policy(tree, mapping, &lgc, n)`...

src/orange/orangeinp/detail/LogicBuilder.cc
```diff
@@ -0,0 +1,89 @@
...
+
+namespace celeritas
+{
+namespace orangeinp
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a single node to a given notation.
```
elliottbiondo: Convert a single node to logic in the format dictated by the LogicBuilderPolicy.

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
```diff
@@ -3,35 +3,38 @@
...
 {
-namespace
-{
 //---------------------------------------------------------------------------//
 /*!
  * Recursively construct a logic vector from a node with postfix operation.
  *
- * This is an implementation detail of the \c PostfixLogicBuilder class. The
+ * This is a policy used as template parameter of the \c
+ LogicBuilder::operator() . The
```
elliottbiondo: Remove space before period and fix line wrapping

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
```diff
@@ -3,35 +3,38 @@
...
  * user invokes this class with a node ID (usually representing a cell), and
  * then this class recurses into the daughters using a tree visitor.
+ *
+ * Example: \verbatim
+    all(1, 3, 5) -> {{1, 3, 5}, "0 1 & 2 & &"}
+    all(1, 3, !all(2, 4)) -> {{1, 2, 3, 4}, "0 2 & 1 3 & ~ &"}
```
elliottbiondo: Missing \endverbatim

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
elliottbiondo: You could have a base class `BuilderPolicyBase` that implements the shared methods and declares `operator()` and others to be pure virtual.

test/orange/orangeinp/CsgTreeUtils.test.cc
```diff
@@ -231,6 +240,167 @@ TEST_F(CsgTreeUtilsTest, postfix_simplify)
     }
 }
 
+TEST_F(CsgTreeUtilsTest, infix_simplify)
+{
+    // NOTE: mz = below = "false"
+    auto mz = this->insert_surface(PlaneZ{-1.0});
```
elliottbiondo, line 247: Would it be possible to add a ASCII diagram or description of what this cell looks like (i.e. concentric cylinders)?

sethrj (commented):
Sorry I've had these queued up for a couple days but forgot to send

src/orange/orangeinp/UnitProto.cc
esseivaju: If we move the instantiation out of the `LogicBuilder` class, we'd still have to be templated on the `Policy` type so the call would still be very verbose. Given that the builder holds all the objects necessary to construct the policy rather than asking the user to create a new policy object every time we call the builder.

src/orange/orangeinp/detail/LogicBuilder.cc
esseivaju: we could have `LogicBuilderPolicy build_impl{tree_, mapping_, &lgc}(n);`?

src/orange/orangeinp/detail/LogicBuilderPolicies.hh
esseivaju: I was trying to avoid using virtual functions, though this is only called on the host so this would be fine but CRTP can be a good alternative there.

src/orange/orangeinp/detail/LogicBuilder.cc
sethrj: Policies in the standard library (and thrust e.g.) are passed as instances, as template parameters... that's what I'm trying to avoid here.

test/orange/orangeinp/CsgTreeUtils.test.cc
esseivaju: I took this from the existing `postfix_simplify` and adjusted it to the infix notation. I'm not sure what this cell exactly looks like...

src/orange/orangeinp/UnitProto.cc
sethrj: Wouldn't it look like `build_logic(PostfixLogicBuilderPolicy{}, node_id)`? Maybe `LogicBuilder` should actually be a `class PostfixLogicBuilder : public LogicBuilder<PostfixLogicBuilder>`?

esseivaju: I went with the suggestion of having the policy take the construction argument, the `LogicBuilder` could be a free function now since it no longer has a state and a single call operator.

src/orange/orangeinp/UnitProto.cc
```diff
@@ -185,11 +182,11 @@ void UnitProto::build(ProtoBuilder& input) const
         VolumeInput vi;
 
         // Construct logic and faces with remapped surfaces
-        auto&& [faces, logic]
-            = build_logic.operator()<detail::PostfixLogicBuilderPolicy>(
-                node_id);
+        auto faces = build_logic(
+            detail::PostfixLogicBuilderPolicy{
+                csg_unit.tree, &sorted_local_surfaces, &vi.logic},
+            node_id);
```
sethrj: I think this still needs another iteration... getting one return result by reference and another by return value is inconsistent/surprising.

esseivaju: 42bd6c0 and 32a9fdd are the two alternatives. I think that returning by value and moving from lgc is nicer to use.

sethrj: Definitely agree: returning by value (32a9) is better.

esseivaju:
@sethrj @elliottbiondo Ready for a 2nd look.

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,343 @@
...
+    //!@}
+
+    static_assert(std::is_same_v<LocalSurfaceId::size_type, logic_int>,
+                  "unsupported: add enum logic conversion for different-sized "
+                  "face and surface ints");
+
+  public:
+    // Construct with optional mapping
+    inline BaseBuildLogicPolicy(CsgTree const& tree, VecSurface const* vs);
+
```
sethrj: Instead of passing a pointer (i.e. an optional const refernce), how about this class store a pointer internally and just overload the constructor:
```
explicit inline BaseBuildLogicPolicy(CsgTree const& tree);
inline BaseBuildLogicPolicy(CsgTree const& tree, VecSurface const& vs);
```

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,342 @@
...
+    VecLogic logic;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a logic representation of a node.
+ *
+ * The result is a pair of vectors: the sorted surface IDs comprising the faces
+ * of this volume, and the logical representation using \em face IDs, i.e. with
+ * the surfaces remapped to index of the surface in the face vector.
```
elliottbiondo: to the index

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,342 @@
...
+                  "Will move from policy: rvalue ref expected");
+
+    // Construct logic vector as local surface IDs
+    auto& lgc = policy.logic();
+    CELER_EXPECT(lgc.empty());
+    policy(n);
+
+    // Construct sorted vector of faces
+    BuildLogicResult::VecSurface faces;
+    for (auto const& v : lgc)
```
elliottbiondo, line 47: Optionally use `copy_if`

elliottbiondo (approved):

src/orange/orangeinp/detail/BuildLogicUtils.hh
esseivaju:
```c++
    std::copy_if(lgc.begin(), lgc.end(), std::back_inserter(faces),
                 [](logic_int v) { return !logic::is_operator_token(v); });
```
doesn't work because there is no implicit conversion from `logic_int`to. `LocalSurfaceId`. Without temporary storage, I don't see a way to compose `std::copy_if` and `std::transform`.

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,342 @@
...
+inline BuildLogicResult build_logic(BuildLogicPolicy&& policy, NodeId n)
+{
+    static_assert(std::is_invocable_v<BuildLogicPolicy, NodeId>);
+    static_assert(std::is_rvalue_reference_v<BuildLogicPolicy&&>,
+                  "Will move from policy: rvalue ref expected");
+
+    // Construct logic vector as local surface IDs
+    auto& lgc = policy.logic();
+    CELER_EXPECT(lgc.empty());
+    policy(n);
```
sethrj: I hate to keep going with this, but the fact that `lgc` is "moved out" as a reference at the end the function, but required to be in place for `policy(n)` to work, is not intuitive and could easily cause a bug.

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,342 @@
...
+    for (auto& v : lgc)
+    {
+        if (!logic::is_operator_token(v))
+        {
+            auto iter
+                = find_sorted(faces.begin(), faces.end(), LocalSurfaceId{v});
+            CELER_ASSUME(iter != faces.end());
+            v = iter - faces.begin();
+        }
+    }
```
sethrj, line 69: I think this code should be a standalone function since it doesn't depend on the policy.

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,342 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Base class for logic builder policies following CRTP pattern.
+ *
+ * The call operator for Negated and Joined are not implemented in the base
+ * policy and must be provided by the derived class.
+ */
+template<class BuilderPolicy>
```
sethrj, line 111: Man I really hate how many suggestions I've made that don't seem write when implemented... I don't think the complexity (and potentially code size, due to having to duplicate the `std::visit` dispatch) of CRTP is justified here. I see one big function that doesn't require a template (`operator()(Surface const&)`), and a couple of trivial functions that could easily be copied.

Maybe it's good as is, but perhaps a small standalone helper class that manages the `mapping_` would be a nice final touch?

src/orange/orangeinp/detail/BuildLogicUtils.hh
esseivaju: the problem is that now that the client constructs the policy, and the policy needs the logic vector to store the results, we would have to pass a partially constructed policy, construct the result vector, and then assign it to the policy, which I'm not sure is better... We check at compile time that the policy is a rvalue reference and the vector is initialized in the policy' ctor, so it seems easy to use.

src/orange/orangeinp/detail/BuildLogicUtils.hh
esseivaju: As you suggested previously, we could replace the `BaseBuildLogicPolicy` inheritance by using composition. It would be simpler but result in more boilerplate as we have to wrap and forward the multiple overloads. 

Regarding code size, I don't think the different options are any different, in any case, we need an implementation of `std::variant` dispatch for `PostfixBuildLogicPolicy` and one for `InfixBuildLogicPolicy`. The alternative would be virtual calls and dynamic dispatch but that has a runtime cost since I assume this is called many times when building the geometry.

src/orange/orangeinp/detail/BuildLogicUtils.hh
```diff
@@ -0,0 +1,365 @@
...
+    static_assert(std::is_invocable_v<BuildLogicPolicy, NodeId>);
+    static_assert(std::is_rvalue_reference_v<BuildLogicPolicy&&>,
+                  "Will move from policy: rvalue ref expected");
+
+    // Construct logic vector as local surface IDs
+    auto& lgc = policy.logic();
+    CELER_EXPECT(lgc.empty());
+    policy(n);
+
+    return {remap_faces(lgc), std::move(lgc)};
```
esseivaju, line 101: IIRC order of evaluation is well-defined for list initialization

sethrj: Correct, it's just for function calls that they're unspecified. This is quite subtle though...

sethrj: But I guess `std::move` doesn't really do anything anyway, it's after those two guys are evaluated that `BuildLogicResult` is constructed and `lgc`'s state changes (i.e. vector(vector&&) is called).

sethrj (approved):
I think this is good enough, we have enough to do besides going in circles about the details of the details of the details here. Thanks!

PR #1531, opened 2024-12-03
Run clang-tidy on modified files only

esseivaju:
Clang-tidy is quite slow; we can only run it on files updated w.r.t. the target branch during PR to speed it up significantly. During the weekly cron, we still run on the entire codebase.


.github/workflows/build-spack.yml
```diff
@@ -137,10 +137,22 @@ jobs:
...
+              ninja -k0
+              exit $?
+            fi
+            FILES=$(git diff --name-only --diff-filter=ACM "$BASE_REF" HEAD)
+            if [ -z "$FILES" ]; then
+              echo "No files to run clang-tidy on."
+              exit 0
+            fi
+            grep -E '^(src|app|example)/.*\.cc$' - <<< "$FILES" | xargs ${{env.CLANG_TIDY}} -p build
+            exit $?
```
sethrj, line 147: Instead of the special casing in here, how about we do a different `Run incremental clang-tidy` step and add  `if` blocks here and for that?  (We're not "building all"!)

.github/workflows/build-spack.yml
```diff
@@ -137,10 +137,22 @@ jobs:
...
+            if [ "${{github.event_name}}" == "schedule" ]; then
+              ninja -k0
+              exit $?
+            fi
+            FILES=$(git diff --name-only --diff-filter=ACM "$BASE_REF" HEAD)
+            if [ -z "$FILES" ]; then
+              echo "No files to run clang-tidy on."
+              exit 0
+            fi
+            grep -E '^(src|app|example)/.*\.cc$' - <<< "$FILES" | xargs ${{env.CLANG_TIDY}} -p build
```
sethrj: Do we not want to run on tests? If only documentation has been added does what's the exit code?

sethrj (approved):
Thanks for that... I knew there was a way to do incremental tidying 😄 

One more question though... does this tidy code *downstream* of modified headers? It seems like with incremental but not recursive, you could end up introducing a tidy error that wouldn't get caught until the next cron job.

esseivaju:
> does this tidy code _downstream_ of modified headers?

Good point. With the current configuration, we don't output warnings that would be produced in a header file, though modifying a declaration in a header could introduce a warning in an (un)related source file. That error would only be caught in the cron job.  Maybe it's just easier to not do incremental check then?

.github/workflows/build-spack.yml
esseivaju: Eventually, it'd be good, but the current setup doesn't run on test, so there might still be warnings there.

sethrj:
> Good point. With the current configuration, we don't output warnings that would be produced in a header file, though modifying a declaration in a header could introduce a warning in an (un)related source file. That error would only be caught in the cron job. Maybe it's just easier to not do incremental check then?

I know I've seen at least one GHA that claimed to do an incremental but thorough check... since it seems like this could really be a source of issues (a couple of headers in #1524 did require updating) perhaps we do need it. Incidentally, do you know whether clang-tidy does check header files, or if we need to do something special to make sure inline code in those is clean?

esseivaju:
> I know I've seen at least one GHA that claimed to do an incremental but thorough check... since it seems like this could really be a source of issues (a couple of headers in #1524 did require updating) perhaps we do need it. 

Yes, seems like it'd be needed. You'd have to potentially check every cc file transitively including the modified headers..

> Incidentally, do you know whether clang-tidy does check header files, or if we need to do something special to make sure inline code in those is clean?

We can configure it to do so by setting `HeaderFilterRegex` in `.clang-tidy`. I can look at that if there aren't too many warnings there.

sethrj:
This is a tricky one... it looks like *basically* all of the fixes you had in the clang-tidy branch were on `.cc` files so I *think* that means that an incremental test would catch most of them. Let's merge it and see what happens?

esseivaju:
> This is a tricky one... it looks like _basically_ all of the fixes you had in the clang-tidy branch were on `.cc` files so I _think_ that means that an incremental test would catch most of them. Let's merge it and see what happens?

Yes, I think it would catch most of them, and for those that aren't, it wouldn't interfere with the PR workflow until the file containing the missed warning is modified again, which could be updated then if not caught before in the weekly cron.

sethrj (approved):
LGTM, should we wait until #1534 or merge now? Merge if you want :)

esseivaju:
Actually, running git diff against develop isn't the best thing because some PR might have been merged in the meantime, we should find the first commit of that PR and run a git diff against that to find the correct set of modified files.

.github/workflows/build-spack.yml
```diff
@@ -135,15 +135,38 @@ jobs:
...
+        env:
+          BASE_REF: "${{format('{0}', github.base_ref || 'develop')}}"
+        run: |
+          if [ "${{github.event_name}}" == "schedule" ]; then
+            echo "Full clang-tidy check on scheduled run."
+            ninja -k0
+            exit $?
+          fi
+          set -x
+          FIRST_PR_HASH=$(git log --reverse --oneline --format="%H" origin/${BASE_REF}..HEAD | head -1)
```
sethrj: Does this work if it has a fetch depth of zero?

pcanal: And related question, isn't there a github action variable that already has this information?

esseivaju: it should, fetch-depth of 0 means fetch all the history. I reproduced the commands that the checkout action uses to clone the repo locally and it works, but not on the runner...

sethrj: Oh, it's not the "shallowest". Actually I think you may want to use the "destination branch" sha; the GHA always runs on the result of merging the PR into the target. So usually `develop` *would* be correct—sorry I forgot that earlier.

esseivaju: > And related question, isn't there a github action variable that already has this information?

There are way to get the source/destination branch or commit that triggered the workflow but I haven't seen a variable with the base commit of a PR.




> Oh, it's not the "shallowest". Actually I think you may want to use the "destination branch" sha; the GHA always runs on the result of merging the PR into the target. So usually `develop` _would_ be correct—sorry I forgot that earlier.

But that would include any change now in develop that not present in that PR, from the [action logs](https://github.com/celeritas-project/celeritas/actions/runs/12169546804/job/33942867739?pr=1531), I get the correct list of modified files, it seems that grep is failing. 🤔

esseivaju:
LGTM now

sethrj (approved):

pcanal (approved):

PR #1533, opened 2024-12-04
Rename Cerenkov to Cherenkov

amandalund:
This renames Cerenkov to the more common spelling Cherenkov.


sethrj (approved):
Thanks!  I remembered the docs might need updating so I fixed that; and `git grep -i` looks good otherwise.

PR #1534, opened 2024-12-04
Tidy headers

esseivaju:
@sethrj That's what it takes to tidy the headers, 90% of it is just the rule of 5, missing std::forward on function objects passed as universal reference, and override/final annotations.


src/accel/HepMC3PrimaryGenerator.hh
```diff
@@ -51,6 +51,7 @@ class HepMC3PrimaryGenerator final : public G4VPrimaryGenerator
     explicit HepMC3PrimaryGenerator(std::string const& filename);
 
     CELER_DELETE_COPY_MOVE(HepMC3PrimaryGenerator);
+    ~HepMC3PrimaryGenerator() override = default;
```
sethrj: Should these be final instead of override? (The class is final and the rest of the functions are too.) Or does it matter?

src/celeritas/Types.hh
```diff
@@ -43,7 +43,7 @@ using IsotopeId = OpaqueId<struct IsotopeRecord>;
 using MaterialId = OpaqueId<class Material_>;
 
 //! Opaque index of model in the list of physics processes
-using ModelId = OpaqueId<class Model>;
+using ModelId = OpaqueId<class Model_>;
```
sethrj: Why was this needed? There exists a `class Model`... but maybe we should make all of the opaque IDs generic `<struct Foo_>` if don't directly index into an array with element type `T`.

src/celeritas/em/data/AtomicRelaxationData.hh
```diff
@@ -133,6 +133,7 @@ struct AtomicRelaxStateData
 
     //! Assign from another set of states
     template<Ownership W2, MemSpace M2>
+    // NOLINTNEXTLINE(cppcoreguidelines-c-copy-assignment-signature)
```
sethrj: All of our data classes will need this I think... right?

src/celeritas/ext/detail/AllElementReader.hh
```diff
@@ -42,6 +42,7 @@ class AllElementReader
 
     //! Load a map of data for all stored elements
     template<class ReadOneElement>
+    // NOLINTNEXTLINE(cppcoreguidelines-missing-std-forward)
```
sethrj: This thing is obnoxious.  LLVM algorithm implementations don't follow it:
https://github.com/llvm/llvm-project/blob/bb9bb686741feeb2c65097ef750126477f96ff30/libcxx/include/__algorithm/make_heap.h#L31

but should we instead omit the `&&`? (Would that copy objects rather than referencing them?)
https://github.com/llvm/llvm-project/blob/bb9bb686741feeb2c65097ef750126477f96ff30/libcxx/include/__algorithm/for_each_n.h#L26

src/celeritas/field/MakeMagFieldPropagator.hh
```diff
@@ -34,6 +34,7 @@ namespace celeritas
  */
 template<template<class EquationT> class StepperT, class FieldT>
 CELER_FUNCTION decltype(auto)
+// NOLINTNEXTLINE(cppcoreguidelines-missing-std-forward)
```
sethrj: OK this is really absurd... since we're forwarding with our own forward...

src/celeritas/grid/GridIdFinder.hh
```diff
@@ -86,14 +86,10 @@ GridIdFinder<K, V>::operator()(argument_type quant) const -> result_type
...
+                             // Below first point
+        || (iter == grid_.begin() && quant.value() != *iter))
     {
-        // Higher than end point
-        return {};
-    }
-    else if (iter == grid_.begin() && quant.value() != *iter)
-    {
-        // Below first point
         return {};
```
sethrj:
```c++
if (iter == grid_.end() 
        || (iter == grid_.begin() && quant.value() != *iter))
{
 // Higher than end point or below first point
}
```

sethrj:
@esseivaju This is great, but before you go any further can you turn off the "operator= signature" and "missing std::forward" since those are always going to generate new warnings for correct code?

esseivaju:
> @esseivaju This is great, but before you go any further can you turn off the "operator= signature" and "missing std::forward" since those are always going to generate new warnings for correct code?

I'll turn these off, out of curiosity, what's the advantage of passing a function object by forwarding reference instead of constructing referenced if we need to call it multiple times?

sethrj:
@esseivaju The problem is that if you've got a functor that has a mutable `operator()` then you have to pass a mutable reference; but if you're passing a const function object then you have to use a const reference. Passing by value or rvalue reference is I think the only way to do it.

src/celeritas/Types.hh
esseivaju: It's a forward declaration with a class definition in another namespace: 
```
$ clang-tidy src/celeritas/optical/model/RayleighModel.cc
/home/esseivaj.linux/devel/celeritas/src/celeritas/Types.hh:46:32: error: no definition found for 'Model', but a definition with the same name 'Model' found in another namespace 'celeritas::optical' [bugprone-forward-declaration-namespace,-warnings-as-errors]
   46 | using ModelId = OpaqueId<class Model>;
      |                                ^
/home/esseivaj.linux/devel/celeritas/src/celeritas/optical/model/../Model.hh:29:7: note: a definition of 'Model' is found here
   29 | class Model : public OpticalStepActionInterface, public ConcreteAction
```

src/celeritas/ext/detail/AllElementReader.hh
esseivaju: Yeah it would make a copy, though in the typical use case, you pass in a temporary. So, the universal reference is actually deduced to a value type, and directly constructs the argument in place. I guess we can keep it as is for uniformity

src/celeritas/field/MakeMagFieldPropagator.hh
esseivaju: yeah, it's specifically looking for a `std::forward` 😅

src/accel/HepMC3PrimaryGenerator.hh
esseivaju: When updating these, I tried to use final for final classes, and override instead. I missed that one

sethrj (approved):
Thanks for all that (hopefully not too tedious) work!

PR #1535, opened 2024-12-05
Fix hepmc3/macro initialization in celer-g4

sethrj:
@wdconinc encountered a few issues recently when trying to run `celer-g4` from scratch.
- Mark `.mac` file setup as deprecated since we're trying to transition `celer-g4` to JSON-only input
- Default secondary stack factor to 2; we're not enabling atomic relaxation by default we only need room for annihilation
- Wrap HepMC3 construction in exception conversion to catch errors while loading
- No longer set `num_events` since we're using the single-event model


amandalund (approved):
Thanks for the updates!

PR #1536, opened 2024-12-05
Store a single process-integrated energy loss and range table per particle

amandalund:
As a step toward #253 and #907, this stores a single energy loss and range table summed over processes for each particle in the physics data (so the only per-process value grids are the macro xs), removing the hacky `eloss_ppid`.


src/celeritas/phys/PhysicsData.hh
```diff
@@ -118,22 +117,20 @@ struct IntegralXsProcess
...
- * a fixed-size number of ItemRange references to ValueTables. The first index
- * of the table (hard-coded) corresponds to ValueGridType; the second index is
- * a ParticleProcessId. So the cross sections for ParticleProcessId{2} would
- * be \code tables[ValueGridType::macro_xs][2] \endcode. This
- * awkward access is encapsulated by the PhysicsTrackView. \c integral_xs will
- * only be assigned if the integral approach is used and the particle has
- * continuous-discrete processes.
+ * Each index should be accessed with type ParticleProcessId. \c macro_xs
+ * stores the cross section tables for each process, while \c energy_loss and
+ * \c range are the process-integrated dE/dx and range for the particle.  c
```
sethrj:
```suggestion
 * \c range are the process-integrated dE/dx and range for the particle.  \c
```

sethrj (approved):
Awesome! Thanks for picking this back up again.

PR #1539, opened 2024-12-07
Support sense caching for future lazy evaluation

esseivaju:
This is a prototype implementation of lazily evaluating senses. It will only be useful after switching to the `InfixEvaluator`, which can short-circuit senses evaluation. For now, expand the `LogicEvaluator` interface to support lazy evaluation. `complex_intersect` still uses eager evaluation as some senses need to be flipped; it's not clear how that should be supported with lazy evaluation yet.


esseivaju:
I have a prototype working with the `complex_intersect` scenario, but it adds quite a bit of complexity. We have to associate metadata with each surface to tell if it needs to be flipped. That metadata scratch space must be cleared when evaluating a new position.

src/orange/OrangeTypes.hh
```diff
@@ -101,6 +101,18 @@ enum class Sense : bool
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Transformations to apply to senses when using lazy sense evaluation.
+ */
+enum class SenseMod : unsigned char
+{
+    normal = 0,
+    flipped = 1 << 0,
+    cached = 1 << 1,
```
sethrj: Similar enums (surface flags, Geant import flags)  use `0x1`, `0x2`, ... to indicate that the bit sequence is important, and I think the `normal` value should be removed since you can't treat it as a bit position. I think this enum should be named  `SenseFlags`...

Also I wonder if, instead of storing separate senses and sense modifier flags—since each is 1 byte (8 bits) we could combine them into a `FlaggedSense`, and then the bits would correspond to:
```c++
enum class SenseBits
{
    sense,
    flipped,
    cached
};
```

src/orange/OrangeTypes.hh
```diff
@@ -378,6 +390,46 @@ CELER_CONSTEXPR_FUNCTION Sense to_sense(bool s)
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Flip a sense modifier.
+ */
+[[nodiscard]] CELER_CONSTEXPR_FUNCTION SenseModFlags
+flip_sense_mod(SenseMod mod, SenseModFlags flags)
+{
+    return flags ^ static_cast<SenseModFlags>(mod);
+}
```
sethrj: At this point we should just have a device compatible bitset class... I think these helper functions have a slight code smell.

src/orange/univ/detail/LogicEvaluator.hh
```diff
@@ -59,6 +65,17 @@ CELER_FUNCTION LogicEvaluator::LogicEvaluator(SpanConstLogic logic)
...
+{
+    auto calc_sense = [&](FaceId face_id) { return values[face_id.get()]; };
+    return (*this)(calc_sense);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Evaluate a logical expression, with on-the-fly sense evaluation.
+ */
+template<class F, std::enable_if_t<std::is_invocable_v<F, FaceId>, bool>>
```
sethrj, line 78: Neat!

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -202,26 +206,29 @@ CELER_FUNCTION auto
...
 
         VolumeView vol = this->make_local_volume(id);
-        auto logic_state
-            = calc_senses(vol, detail::find_face(vol, state.surface));
-
-        if (detail::LogicEvaluator(vol.logic())(logic_state.senses))
+        auto on_face = detail::find_face(vol, state.surface);
+        auto bind_calc_sense = [&](FaceId face_id) {
+            return calc_senses(vol, face_id, on_face);
+        };
```
sethrj: Why not have the LazySenseCalculator just take a `face_id` since it's constructed locally where you already have `vol` and `on_face`? Alternatively why not construct the lazy calculator outside `cross_boundary`?

src/orange/univ/detail/LazySenseCalculator.hh
```diff
@@ -0,0 +1,163 @@
...
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate senses with a fixed particle position.
+ *
+ * This is an implementation detail used in initialization *and* complex
+ * intersection.
```
sethrj: Add more description here. Note also you're going from `senses` to a single sense on one face.

src/orange/univ/detail/LazySenseCalculator.hh
```diff
@@ -0,0 +1,163 @@
...
+                                              Span<SenseModFlags> sense_flags);
+
+    // Calculate senses for a single face of the given volume, possibly on a
+    // face
+    inline CELER_FUNCTION Sense operator()(VolumeView const& vol,
+                                           FaceId face_id,
+                                           OnFace face = {});
+
+    //! Clear the cached sense values
+    CELER_FUNCTION void invalidate_cache()
```
sethrj: Would it be OK (no less efficient) to invalidate every time you instantiate the LSC?

src/orange/univ/detail/LazySenseCalculator.hh
```diff
@@ -0,0 +1,163 @@
...
+    {
+        for (auto& flags : sense_flags_)
+        {
+            flags = unset_sense_mod(SenseMod::cached, flags);
+        }
+    }
+
+    //! The first face encountered that we are "on"
+    CELER_FUNCTION OnFace& on_face() { return face_; }
```
sethrj: Maybe instead the class should take `OnFace&` as a construction argument? The initial value should be left unchanged.

sethrj (commented):
Interesting work... any performance testing yet (gpu/cpu)?

src/orange/univ/detail/LazySenseCalculator.hh
esseivaju: We already kinda do that, the ctor clears all the bit flags, this only clears the cached flag

src/orange/univ/detail/LogicEvaluator.hh
esseivaju: this is actually necessary otherwise this overload is a better match than `SpanConstSense` in some situations.

src/orange/univ/SimpleUnitTracker.hh
esseivaju: We could change the API and pass `vol` and `on_face` when constructing the LSC.  We have to build the LSC inside the `is_inside` lambda because of the handling logic for `on_face`, which must be reset every time. 

This would make the LSC instance usable only for a specific volume, however, given that we need to insert an ´invalidate_cache` if we're reusing it for another volume ( as is done inside `initialize`),  this might be better as this is quite easy to misuse as it is now.

src/orange/OrangeTypes.hh
esseivaju: I was also thinking of combining the two, however, many classes rely on the `OrangeTrackView` API returning a `Span<Sense>`, so this change would require updating many users of that interface.  Maybe we could make an on-the-fly conversion of `Span<FlaggedSense>` to `Span<Sense>`

src/orange/OrangeTypes.hh
esseivaju: I agree these don't make much sense to be specific to that enum type and could be more generic. I can have a look at implementing a simple device compatible bitset with these primitives

src/orange/univ/detail/LazySenseCalculator.hh
esseivaju: >  The initial value should be left unchanged.

Mmh, what do you mean by this?

esseivaju (commented):
I haven't done any performance testing yet. Given that we're still using the same postfix logic evaluator we can't take advantage of short-circuiting sense evaluations yet so I wouldn't expect any improvement, and parity at best.

sethrj:
The caching though should reduce the number of calculations to be done though, right?

src/orange/univ/detail/LazySenseCalculator.hh
sethrj: Instead of the function-like class storing and providing an accessor for the face, have it take a pointer to one from the parent scope. It doesn't even have to initialize it on construction: the original "sense calculator" respects the original value (so that it will only override a "false" value if a new face is found).

src/orange/univ/SimpleUnitTracker.hh
sethrj: Yeah, I think that's a better approach. The original sense calculator had track/universe as "constant" data as a convenience (the constructor does nothing) and transformed one volume to one list of results; there's no reason we can't change the semantics for this use case.

src/orange/OrangeTypes.hh
sethrj: #1547

src/orange/OrangeTypes.hh
sethrj: Is it legal to reinterpret cast `FlaggedSense*` to `Sense*`?  Maybe we could have a function that checks that the reinterpreted values are equivalent and performs the span cast? And I see now that `Sense` has an underlying type of `bool`, but since that still has sizeof(1) should it be ok?

esseivaju: I don't think `reinterpret_cast` would work here, any non-zero value, i.e., having any flag set would be interpreted as true. We'd have to check the sense bit.

esseivaju:
> The caching though should reduce the number of calculations to be done though, right?

Yes, though I don't know how common to evaluate the sense of a surface multiple times and whether that is offset by having to  memset(0) of the sense flags when we create a new LSC.

src/orange/OrangeTypes.hh
esseivaju: maybe we could have something like: 
```c++
enum class Sense : bool
{
    inside,  //!< Quadric expression is less than zero
    outside,  //!< Expression is greater than zero
};

enum SenseFlags : unsigned char
{
    cached = 0x1,  //!< Sense is cached
    sense = 0x2,  //!< Sense
};

struct SenseMeta
{
bitset<2> sense_;
operator bool() const
{
return sense_[SenseFlags::sense];
}

operator Sense() const
{
    return static_cast<Sense>(bool(*this));
}
};
```

src/orange/OrangeTypes.hh
```diff
@@ -436,6 +424,67 @@ CELER_CONSTEXPR_FUNCTION Sense to_sense(SignedSense s)
...
+    //! Check wether there is a cached sense value
+    CELER_CONSTEXPR_FUNCTION bool cached() const
+    {
+        return sense_[SenseFlags::cache];
+    }
+
+    //! Invalidate the cached sense value
+    CELER_CONSTEXPR_FUNCTION void invalidate()
+    {
+        sense_[SenseFlags::cache] = false;
```
sethrj: It might be cheaper to clear the value entirely since the state of the other bits doesn't matter at that point.

src/orange/OrangeTypes.hh
```diff
@@ -436,6 +424,67 @@ CELER_CONSTEXPR_FUNCTION Sense to_sense(SignedSense s)
...
+    }
+
+  private:
+    Bitset<2> sense_;
```
sethrj: Hm, doesn't this increase `sizeof(Sense) == 1` to `sizeof(Sense) == unsigned int`? Should our bitset have an additional "underlying type" parameter, or based on the size of `N` increase the word size up to a certain point?
```c++
using word_type = std::conditional<(N <= 8), std::uint_least8_t,
                  std::conditional<(N <= 16), std::uint_least16_t,
                  unsigned int>>;
```

src/orange/OrangeTypes.hh
```diff
@@ -436,6 +424,67 @@ CELER_CONSTEXPR_FUNCTION Sense to_sense(SignedSense s)
...
+ */
+class SenseMeta
+{
+    /*!
+     * Bitset indices for the sense metadata.
+     */
+    enum SenseFlags : unsigned char
+    {
+        cache,  //!< Sense is cached
+        sense,  //!< Sense value
```
sethrj: Maybe it doesn't matter, but wouldn't we want `sense` to be the first bit? That way we could interpret a vector of senses as a vector of SenseMeta.

src/orange/OrangeTypes.hh
```diff
@@ -436,6 +424,67 @@ CELER_CONSTEXPR_FUNCTION Sense to_sense(SignedSense s)
     return Sense(static_cast<int>(s) >= 0);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Wrapper for a sense value which associated metadata.
+ */
+class SenseMeta
```
sethrj: At this point we should put SenseMeta (and related Sense helper functions) into its own header 😄

src/orange/OrangeTypes.hh
esseivaju: that would be possible, but a bit tricky, we have to be careful because `std::uint8_t` is promoted to a int for `operator ~` which causes issue, maybe do so in a separate PR?

esseivaju:
We can wait before merging this until I get performance number to make sure it's not a performance regression with postfix evaluation. I'll also split off the Bitset changes in a separate PR. And I'm not convinced about the `SenseMeta` name, if you have a better suggestion (`SenseStorage`?)

esseivaju:
@sethrj It seems that the lazy sense evaluation is slightly slower, see the event throughput tables below, running ORANGE regression tests on 19517d6:

| Problem                                                        | Geometry |    GPU [1/s] |    CPU [1/s] | GPU+G4 [1/s] |
| -------------------------------------------------------------- | -------- | ------------ | ------------ | ------------ |
| atlas-tilecal                            | orange   | 1.12 (±0.00) | 0.10 (±0.00) | 1.03 (±0.01) |
| cms-hgcal                  | orange   | 0.63 (±0.00) | 0.06 (±0.00) | 0.57 (±0.00) |
| testem15                                 | orange   | 8.23 (±0.04) | 0.43 (±0.00) | 6.36 (±0.03) |
| testem15+field             | orange   | 7.78 (±0.04) | 0.40 (±0.00) | 5.95 (±0.18) |
| testem15+field+msc            | orange   | 6.51 (±0.02) | 0.35 (±0.00) | 5.22 (±0.02) |
| testem3-composite+field+msc  | orange   | 1.04 (±0.00) | 0.11 (±0.00) | 1.16 (±0.01) |
| testem3-composite+msc                  | orange   | 2.10 (±0.01) | 0.13 (±0.00) | 1.86 (±0.00) |
| testem3-expanded+field+msc | orange   | 1.04 (±0.01) | 0.12 (±0.00) | 1.16 (±0.01) |
| testem3-flat                       | orange   | 2.56 (±0.01) | 0.19 (±0.00) | 2.46 (±0.00) |
| testem3-flat+field      | orange   | 1.92 (±0.01) | 0.17 (±0.00) | 1.93 (±0.01) |
| testem3-flat+field+msc     | orange   | 1.12 (±0.01) | 0.11 (±0.00) | 1.30 (±0.01) |
| testem3-flat+msc                         | orange   | 1.94 (±0.01) | 0.12 (±0.00) | 1.82 (±0.01) |

and here's develop (6ec7f84):
| Problem                                                        | Geometry |    GPU [1/s] |    CPU [1/s] | GPU+G4 [1/s] |
| -------------------------------------------------------------- | -------- | ------------ | ------------ | ------------ |
| atlas-tilecal                          | orange   | 1.12 (±0.00) | 0.09 (±0.00) | 1.04 (±0.01) |
| cms-hgcal                               | orange   | 0.64 (±0.00) | 0.06 (±0.00) | 0.59 (±0.00) |
| testem15                            | orange   | 8.25 (±0.05) | 0.43 (±0.00) | 6.29 (±0.07) |
| testem15+field       | orange   | 7.80 (±0.04) | 0.39 (±0.00) | 5.85 (±0.04) |
| testem15+field+msc        | orange   | 6.55 (±0.02) | 0.34 (±0.00) | 5.18 (±0.06) |
| testem3-composite+field+msc | orange   | 1.05 (±0.00) | 0.11 (±0.00) | 1.18 (±0.01) |
| testem3-composite+msc                   | orange   | 2.14 (±0.00) | 0.12 (±0.00) | 1.88 (±0.00) |
| testem3-expanded+field+msc | orange   | 1.05 (±0.01) | 0.11 (±0.00) | 1.18 (±0.01) |
| testem3-flat                          | orange   | 2.98 (±0.00) | 0.19 (±0.00) | 2.72 (±0.02) |
| testem3-flat+field        | orange   | 2.15 (±0.01) | 0.18 (±0.00) | 2.13 (±0.00) |
| testem3-flat+field+msc     | orange   | 1.21 (±0.00) | 0.12 (±0.00) | 1.39 (±0.00) |
| testem3-flat+msc                     | orange   | 2.20 (±0.00) | 0.13 (±0.00) | 1.98 (±0.00) |

having to memset(0) the sense storage every time is expensive. Before 19517d6 the atlas-tilecal and cms-hgcal were significantly slower as we were clearing `max_faces` senses every time (305 and 403, respectively):

| Problem                                                        | Geometry |    GPU [1/s] |    CPU [1/s] | GPU+G4 [1/s] |
| -------------------------------------------------------------- | -------- | ------------ | ------------ | ------------ |
| atlas-tilecal                               | orange   | 0.39 (±0.00) | 0.09 (±0.00) | 0.39 (±0.00) |
| cms-hgcal                                | orange   | 0.21 (±0.00) | 0.06 (±0.00) | 0.21 (±0.00) |

Though with lazy sense evaluation, I don't see a way around avoiding cleaning the temporary sense storage, we'll have to see if the shortcutting when using infix evaluation saves time. See commits in [regression/lazy-sense-eval](https://github.com/celeritas-project/regression/tree/lazy-sense-eval)  for more details.

For this PR, should we revert to using the default sense calculator, and only use the lazy sense calculator with infix logic evaluation?

sethrj:
@esseivaju Those are nice results. One of the biggest differences between the "real" detectors (which don't suffer much) versus testem3 (which loses 20% performance) is that the latter doesn't have repeated surfaces in the same tree. So I think we're seeing a benefit from the caching when it's used, but in most cases it's wiped out by the other overhead.

I think we should hold off on this PR until we do infix notation. Maybe we could commit some of the classes?

esseivaju:
I would revert the changes in `SimpleUnitTracker` and keep the rest; then, once we merge #1530 we'll have everything in place to use infix notation.

esseivaju:
>  So I think we're seeing a benefit from the caching when it's used, but in most cases it's wiped out by the other overhead.

I was thinking about whether we could bypass needing to clear up the scratch space but I couldn't find a way around that unless we recompute the sense every time at which point the storage isn't needed. Once we have the infix notation we'll see hopefully see benefits.

elliottbiondo:
Should I wait for the infix MR to go in, or review now?

esseivaju:
This is ready and can be merged independently even though potential benefits will come once we have infix.

elliottbiondo (approved):
Looks good! Interested in seeing the timing results with short circuiting.

sethrj (approved):
The new implementation and evaluation looks great! However, I'm not super happy with the name `SenseMeta` (should be something like SenseFlags, SenseBits, SenseBitset ... I don't think it conveys "sense plus metadata" which it seems like it's meant to? "Meta" on its own makes me think metaclass or somehow it's a sense about a sense.) or moving the `SignedSense` code to SenseUtils since it serves a different internal purpose (but honestly that should probably be in the `univ/detail`).

Any thoughts on a name change? If you disagree let's merge since I've sat on this way too long. Sorry!

src/orange/SenseUtils.hh
```diff
@@ -0,0 +1,170 @@
...
+/*!
+ * Wrapper for a sense value with associated metadata.
+ */
+class SenseMeta
+{
+  private:
+    enum : size_type
+    {
+        sense_bit,
+        cached_bit,
```
pcanal:
```suggestion
        is_cached_bit,
```

src/orange/SenseUtils.hh
```diff
@@ -0,0 +1,170 @@
...
+            return "outside";
+    }
+    return "<invalid>";
+}
+
+//---------------------------------------------------------------------------//
+// CLASSES
+//---------------------------------------------------------------------------//
+/*!
+ * Wrapper for a sense value with associated metadata.
```
pcanal: Is the 'metadata' really 'just' whether or not the value is cached !?  If not, we should document here what is the meaning of meta.

esseivaju:
yeah, I mentioned that in an earlier comment as well, I don't like the `SenseMeta` name but I'm not sure what is best. Since we only store whether the sense is cached as extra information, I was thinking `CachedSense`, but this is not really future-proof, maybe `SenseStorage`, `SenseWrapper`, or `SenseFlags` is good too..

src/orange/SenseUtils.hh
esseivaju: for now yes, I used meta to be generic for potential future data associated with sense but I don't like that name that much as mentioned in the main comment

src/orange/SenseUtils.hh
```diff
@@ -0,0 +1,170 @@
...
+    //! Assign a sense value
+    CELER_CONSTEXPR_FUNCTION SenseMeta& operator=(Sense sense)
+    {
+        sense_[sense_bit] = static_cast<bool>(sense);
+        sense_[cached_bit] = true;
+        return *this;
+    }
+
+    //! Check wether there is a cached sense value
+    CELER_CONSTEXPR_FUNCTION bool cached() const { return sense_[cached_bit]; }
```
pcanal:
```suggestion
    CELER_CONSTEXPR_FUNCTION bool is_valid() const { return sense_[cached_bit]; }
```
or `is_assigned`.  The 'thing' might be used as a cache but know nothing on how to calculate/acquire the needed to fill the 'cache'.

src/orange/SenseUtils.hh
```diff
@@ -0,0 +1,170 @@
...
+    return "<invalid>";
+}
+
+//---------------------------------------------------------------------------//
+// CLASSES
+//---------------------------------------------------------------------------//
+/*!
+ * Wrapper for a sense value with associated metadata.
+ */
+class SenseMeta
```
pcanal:
```suggestion
class SenseValue
```
it has 3 states:
- unassigned
- assigned and false
- assigned and true

Semantically, this sounds very similar to `std::optional<bool>`

src/orange/SenseUtils.hh
```diff
@@ -0,0 +1,170 @@
...
+{
+  private:
+    enum : size_type
```
pcanal: Why `size_type`? Why not
```suggestion
    enum : char
```

pcanal:
> but this is not really future-proof, 

We have been very agile in updating names, so unless there is already a clear 'will add this feature(s) in the next (few) rounds' (in which cases we should bring those into the conversation), I would stay simple/straightforward.

src/orange/SenseUtils.hh
esseivaju: The bitset is indexed with `size_type` so there will be an implicit conversion. Since it's all at compile time I don't think it would change anything either way

src/orange/SenseUtils.hh
esseivaju: I think `SenseValue` is a good name. Right, it is essentially a `std::optional<bool>` but this needs to be compatible with device code, which `std::optional` isn't.

pcanal (approved):

src/orange/SenseUtils.hh
pcanal: What do mean by 'indexed' in this context?

src/orange/SenseUtils.hh
pcanal: obsolete due to other changes.

src/orange/SenseUtils.hh
pcanal: Resolved by using the name `SenseValue`

src/orange/SenseUtils.hh
esseivaju: Calling `Bitset::operator[](size_type)` by using the enum values as argument

sethrj (approved):
Nice!

PR #1544, opened 2024-12-10
Add a calculator to integrate range from energy loss

amandalund:
As we refactor the physics table construction we'll probably move toward replacing the `Process::step_limits` with `macro_xs` and `energy_loss` (initially the process-integrated energy loss, but eventually letting each process provide its own and summing them) and then integrating the range from the energy loss ourselves. This adds a calculator for integrating the range from a dedx grid and uses it in the physics grid construction.

I'm leaving this as a draft for now because it changes a lot of test results. I think this could be because of differences in the spline interpolation, but it would be good to understand and discuss what we want to do before I go too far with it. I added a failing test comparing range values calculated from the imported dedx to the imported range to get a rough idea of the magnitude of the differences.


src/celeritas/phys/PhysicsParams.cc
```diff
@@ -533,7 +534,16 @@ void PhysicsParams::build_xs(Options const& opts,
                 xs_grid_ids[mat_idx] = build_grid(builders[VGT::macro_xs]);
                 eloss_grid_ids[mat_idx]
                     = build_grid(builders[VGT::energy_loss]);
-                range_grid_ids[mat_idx] = build_grid(builders[VGT::range]);
+                if (auto grid_id = eloss_grid_ids[mat_idx])
+                {
+                    //! \todo Use \c spline_eloss_order when enabled by default
+                    size_type constexpr spline_order = 2;
```
sethrj: @amandalund Did you say Geant4 uses cubic splines? Maybe that's the discrepancy here? Perhaps also the boundary treatment (reduced order, extrapolation, ...?) is different

amandalund: Yes, I suspect the differences might be from the interpolation method... this produced the closest results to Geant4. Do you know what our `SplineXsCalculator` uses? It looks like it might be Lagrange polynomial interpolation?

amandalund:
I've updated this to use spline interpolation, and the calculated range values match the imported ones. The only tests where the results changed are the mock tests (since we're calculating the range in the physics params instead of constructing it in the mock process) and a few float tests.

One thing I noticed is spline interpolation does not work well if there are a small number of grid points and the scale of the x grid is large. Hopefully Geant4 always uses a dense enough grid, and at least this should give us a little more control of the calculation.

sethrj:
I guess since the grid is logarithmic in energy, the `delta_x` for a point before and after are very different, which (since they're the off-diagonal elements) I expect could lead to an ill-conditioned matrix. I wonder if we did the spline interpolation in log-E space we'd be ok? (But then maybe we'd introduce an inconsistency between the interpolated functional representations of dE/dx and the range?) Not something to worry about at the moment: consistency with Geant4 is fine. 

Maybe investigating other representations of dE/dx and consistent representations of range would be good future work for the "student projects" section.

src/celeritas/grid/RangeGridCalculator.hh
```diff
@@ -0,0 +1,74 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the range from the energy loss.
+ *
+ * The range of a particle with energy \f$ E_0 \f$ is calculated by integrating
+ * the reciprocal of the stopping power over the energy:
+ * \f[
+   R(E_0) = \int_0^{E_0} - \frac{\dif x}{\dif E} \dif E.
```
sethrj:
```suggestion
   R(E_0) = \int_0^{E_0} - \difd{x}{E} \dif E.
```

test/celeritas/phys/Physics.test.cc
```diff
@@ -141,7 +141,7 @@ TEST_F(PhysicsParamsTest, output)
...
+        R"json({"_category":"internal","_label":"physics","models":{"label":["mock-model-1","mock-model-2","mock-model-3","mock-model-4","mock-model-5","mock-model-6","mock-model-7","mock-model-8","mock-model-9","mock-model-10","mock-model-11"],"process_id":[0,0,1,2,2,2,3,3,4,4,5]},"options":{"fixed_step_limiter":0.0,"heavy.lowest_energy":[0.001,"MeV"],"heavy.max_step_over_range":0.2,"heavy.min_range":0.010000000000000002,"light.lowest_energy":[0.001,"MeV"],"light.max_step_over_range":0.2,"light.min_range":0.1,"linear_loss_limit":0.01,"min_eprime_over_e":0.8,"spline_eloss_order":1},"processes":{"label":["scattering","absorption","purrs","hisses","meows","barks"]},"sizes":{"integral_xs":8,"model_groups":8,"model_ids":11,"process_groups":5,"process_ids":8,"reals":269,"value_grid_ids":89,"value_grids":89,"value_tables":29}})json",
```
sethrj, line 144: The increased number of reals is that we're now *always* storing second derivatives and using spline interpolation for the range calculation, correct? I think that's a significant enough change that we should either break this into two commits (one to add the capability, one to activate it) or just rename the PR something like "Always use spline interpolation for range calculation"

sethrj (approved):

test/celeritas/phys/Physics.test.cc
amandalund: This is because for the mock processes we previously had three values in the energy loss grids and only two values in the range grids, but now that the range is calculated in the physics params we have three in each. We're currently only using spline interpolation during initialization to calculate the range grid; the range calculation during the simulation is still done using linear interpolation.

sethrj: Got it. Thanks!

PR #1547, opened 2024-12-18
Implement device bitset

esseivaju:
Following the discussion in #1539, this adds a `bitset` that can be used by device functions. This is based on the `std::biteset`[libstdc++ implementation](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/std/bitset). For now, I only implemented a subset of the API that should cover our use case.  I'll add a bit of extra documentation, but the code is ready for review.


src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+    static constexpr size_t bits_per_word = CHAR_BIT * sizeof(word_type);
+    static constexpr size_t num_words = (N / bits_per_word)
+                                        + (N % bits_per_word == 0 ? 0 : 1);
+
+  public:
+    /*!
+     * Reference to a single bit in the bitset.
+     * This is used to implement the mutable operator[].
+     */
+    class reference
```
sethrj: Can you declare the class here and define it out-of-line? That would help readability.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+                }
+            }
+            return *this;
+        }
+
+        // Flips the bit
+        CELER_CONSTEXPR_FUNCTION
+        bool operator~() const noexcept
+        {
+            return (*(word_pointer_)&Bitset::mask(bit_pos_)) == 0;
```
sethrj:
```suggestion
            return !static_cast<bool>(*this);
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        // For x = b[i];
+        CELER_CONSTEXPR_FUNCTION
+        operator bool() const noexcept
+        {
+            return (*(word_pointer_)&Bitset::mask(bit_pos_)) != 0;
```
sethrj: Not sure if this is formatted weirdly or incorrect with order of operations... should it be
```suggestion
            return ((*word_pointer_) & Bitset::mask(bit_pos_)) != 0;
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Device-compatible bitset implementation.
+ *
+ * This implementation is based on libstdc++'s std::bitset implementation.
+ * It it a subset of the C++ standard, but it should be sufficient
+ * for our current use case. Given that GPU typically use 32-bit words, this
+ * uses unsigned int as the word type instead of the unsigned long used by the
+ * standard library.
```
sethrj: I'd add a note that this is not "thread-safe" in that multiple threads can't modify bits of the same container (unlike if it were chars or ints).

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+        }
+        // only compare the last word up to the last bit of the bitset
+        return last_word()
+               == (~static_cast<word_type>(0)
+                   >> (num_words * bits_per_word - N));
+    }
+
+    CELER_CONSTEXPR_FUNCTION bool any() const noexcept
+    {
+        for (size_t i = 0; i < num_words; ++i)
```
sethrj:
```suggestion
        for (size_type i = 0; i < num_words; ++i)
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+ * - hash support
+ * - construct from string, from_ulong, from_ullong
+ */
+template<size_t N>
+class Bitset
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using word_type = unsigned int;
```
sethrj, line 47: Or would this be better as size_type?

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+ * standard library.
+ *
+ * The following methods are not implemented:
+ * - conversions to string, to_ulong, to_ullong
+ * - set operations with other bitsets
+ * - stream operators
+ * - hash support
+ * - construct from string, from_ulong, from_ullong
+ */
+template<size_t N>
```
sethrj: Probably better as size_type? If not should be `std::size_t`.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+        size_t count = 0;
+        for (size_t i = 0; i < num_words; ++i)
+        {
+#if CELER_DEVICE_COMPILE
+            count += __popc(words_[i]);
+#elif defined(_MSC_VER)
+            count += __popcnt(words_[i]);
+#else
+            count += __builtin_popcount(words_[i]);
+#endif
```
sethrj: This architecture-dependent guy might be better as a function in math/algorithms?

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+            if (words_[i] != static_cast<word_type>(0))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    CELER_CONSTEXPR_FUNCTION bool none() const noexcept { return !any(); }
```
sethrj: Check that the member functions are called with `this->`
```suggestion
    CELER_CONSTEXPR_FUNCTION bool none() const noexcept { return !this->any(); }
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+template<size_t N>
+class Bitset
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using word_type = unsigned int;
+    //!@}
+
+    static constexpr size_t bits_per_word = CHAR_BIT * sizeof(word_type);
```
sethrj: Since the original class was in `namespace std` this was fine but now we probably want either `std::size_t` or typedef a `size_type`.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,310 @@
...
+    {
+        return words_[num_words - 1];
+    }
+
+    CELER_CONSTEXPR_FUNCTION word_type last_word() const noexcept
+    {
+        return words_[num_words - 1];
+    }
+
+    word_type words_[num_words] = {};
```
sethrj: This class is no longer trivially constructable so I'm not sure we can have it as device data: should our `bitset` be an "operation class" that takes a Span or a pointer so that it can operate on preallocated data?

sethrj (commented):
Neato! Needs a little documentation, and maybe a decision about whether this should be a data class that can be in an uninitialized state (in which case I think `operator==` would have to be modified to account for the last non-full word) or if it should be a "view" class that takes a pointer to someone else's words that represent a bitset.

src/corecel/cont/Bitset.hh
esseivaju: weird formatting indeed, I removed superfluous parentheses.

src/corecel/cont/Bitset.hh
esseivaju: I'd keep this as is, or maybe `storage_type`.

src/corecel/cont/Bitset.hh
esseivaju: Also, the libstdc++ uses a base class to specialize in case a single word is needed, we might want to do that as well since it seems like this would be the most common use case?

esseivaju:
> Neato! Needs a little documentation, and maybe a decision about whether this should be a data class that can be in an uninitialized state (in which case I think `operator==` would have to be modified to account for the last non-full word) or if it should be a "view" class that takes a pointer to someone else's words that represent a bitset.

I think it'd make sense for it to be a data class, it's hard to imagine a case where you have a word sometimes viewed as a bitset, the reverse seems more likely.  maybe we should provide a conversion to `unsigned int` operator in case a user wants to view the bitset as a normal word. The standard only allows converting a bitset to a [single word](https://en.cppreference.com/w/cpp/utility/bitset/to_ulong) but we could provide a `Span` view of it.

sethrj:
I totally agree @esseivaju. The only reason I asked about a view is that I *thought* our collection required trivially constructible objects, but looking again it's just trivially *copyable* and *destructible*. So I think this is fine!

src/corecel/cont/Bitset.hh
sethrj: I'm fine with `word_type`.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using word_type = unsigned int;
+    //!@}
+
+    static constexpr size_type bits_per_word = CHAR_BIT * sizeof(word_type);
+    static constexpr size_type num_words = (N / bits_per_word)
+                                           + (N % bits_per_word == 0 ? 0 : 1);
```
sethrj: from math algorithms:
```suggestion
    static constexpr size_type num_words = ceil_div(N, bits_per_word);
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+
+  public:
+    class reference;
```
sethrj: Move this up below the type aliases since it's a type definition.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+
```
sethrj, line 54: Would a `Bitset() = default;` be a problem? It took me quite some time to find out how to construct this 😅 (i.e. realizing it uses the default constructor).

src/corecel/math/Algorithms.hh
```diff
@@ -742,5 +742,26 @@ CELER_FORCEINLINE_FUNCTION void sincospi(double a, double* s, double* c)
...
+//---------------------------------------------------------------------------//
+// Portable utilities functions
+//---------------------------------------------------------------------------//
+/*!
+ * Count the number of set bits in an integer.
+ */
+#if defined(_MSC_VER)
+inline int popcount(unsigned int x) noexcept
+#else
+inline constexpr int popcount(unsigned int x) noexcept
```
sethrj, line 754: And add a tiny test for this too please! :)

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+};
+
+/*!
+ * Reference to a single bit in the bitset.
+ * This is used to implement the mutable operator[].
+ */
+template<size_type N>
+class Bitset<N>::reference
+{
+    friend class Bitset;
```
sethrj: Is it necessary to add it as a friend? Since it's a nested class, I think it should have access to everything private.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+/*!
+ * Device-compatible bitset implementation.
+ *
+ * This implementation is based on libstdc++'s std::bitset implementation.
+ * It it a subset of the C++ standard, but it should be sufficient
+ * for our current use case. Given that GPU typically use 32-bit words, this
+ * uses unsigned int as the word type instead of the unsigned long used by the
+ * standard library. This container is not thread-safe, multiple threads are
+ * likely to manipulate the same word and access is not synchronized.
+ *
```
sethrj, line 31: Since all the methods are defined inline the class definition, the class is hard to read. Can you either un-inline those classes or provide a synopsis in the class documentation here?

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,395 @@
...
+    set(size_type pos, bool value = true) noexcept
+    {
+        if (value)
+        {
+            this->get_word(pos) |= Bitset::mask(pos);
+        }
+        else
+        {
+            this->get_word(pos) &= ~Bitset::mask(pos);
+        }
```
sethrj: Is this any more efficient than `(*this)[pos] = value;`?

sethrj (approved):

src/corecel/cont/Bitset.hh
sethrj: Before we go through the trouble, maybe we could compile our "unoptimized" and the libstdc++ "optimized" on godbolt and see if there's any actual differences. Since the size is known at compile time and everything's constexpr, I *hope* the compiler could make them equivalent?

src/corecel/cont/Bitset.hh
esseivaju: make sense to have this explicitly declared

src/corecel/cont/Bitset.hh
esseivaju: it _should_ be the same if the compiler can inline everything which seems very easy to do.

src/corecel/cont/Bitset.hh
esseivaju: Actually not necessary for the current implementation

src/corecel/cont/Bitset.hh
esseivaju: gcc and clang seem to optimize that: https://godbolt.org/z/Eqjvj17PK

sethrj: Awesome! No specialization needed then (I assume).

test/corecel/math/Algorithms.test.cc
```diff
@@ -571,6 +571,21 @@ TEST(MathTest, signum)
...
+
+TEST(PortabilityTest, popcount)
+{
+    unsigned int x = 0xAA000000;
+    EXPECT_EQ(4, popcount(x));
+
+    x &= 0xF0000000;
+    EXPECT_EQ(2, popcount(x));
+    x <<= 2;
+    EXPECT_EQ(1, popcount(x));
```
sethrj: This would fail if `sizeof(unsigned int) != 4` right? Maybe it'd be safer in this test to right-shift or use `std::uint32_t`.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,553 @@
...
+}
+
+//---------------------------------------------------------------------------//
+//! Get the value of the bit at position pos. Equivalent to operator[] with
+//! a precondition check in debug mode.
+template<size_type N>
+CELER_CONSTEXPR_FUNCTION bool Bitset<N>::test(size_type pos) const
+    noexcept(!CELERITAS_DEBUG)
+{
+    CELER_EXPECT(pos < N);
```
sethrj: Are you sure this compiles (constexpr can't throw in C++17)? I don't see a test for this function. Maybe should be declared `inline CELER_FUNCTION`.

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,553 @@
...
+//! Flip all bits
+template<size_type N>
+CELER_CONSTEXPR_FUNCTION Bitset<N>& Bitset<N>::flip() noexcept
+{
+    for (size_type i = 0; i < num_words; ++i)
+    {
+        words_[i] = ~words_[i];
+    }
+    // sanitize the last word
+    detail::Sanitize<N % bits_per_word>::sanitize(last_word());
```
sethrj: Now that we don't support N=0 I think it would be cleaner to make this a member function. (Also note that "sanitize" means "zero unused bits")

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,553 @@
...
+    CELER_CONSTEXPR_FUNCTION word_type& get_word(size_type pos) noexcept;
+
+    // Get the last word of the bitset
+    CELER_CONSTEXPR_FUNCTION word_type& last_word() noexcept;
+
+    // Get the last word of the bitset
+    CELER_CONSTEXPR_FUNCTION word_type last_word() const noexcept;
+
+    //! storage
+    word_type words_[num_words]{};
```
sethrj: We group these
```c++
    //// DATA ////
 
    words

    //// HELPER FUNCTIONS ////
```

src/corecel/cont/Bitset.hh
```diff
@@ -0,0 +1,553 @@
...
+    //!@{
+    //! \name Type aliases
+    using word_type = unsigned int;
+    //!@}
+
+    class reference;
+
+    static constexpr size_type bits_per_word = CHAR_BIT * sizeof(word_type);
+    static constexpr size_type num_words = ceil_div(N, bits_per_word);
+
```
sethrj: These don't need to be public, so perhaps they'd be better as private static data with a `_` suffix?

src/corecel/cont/Bitset.hh
esseivaju: Yep, that should be good for now!

src/corecel/cont/Bitset.hh
esseivaju: We still the no-op specialization when `N % bits_per_word=0`, and we can't specialize a member of an unspecialized template, so I think it still needs to be a separate function (struct)

src/corecel/cont/Bitset.hh
esseivaju: This compiles, `constexpr` functions can throw at runtime. I updated `test` to match the standard specification and always check bounds.

src/corecel/cont/Bitset.hh
```diff
@@ -282,12 +286,16 @@ Bitset<N>::operator[](size_type pos) noexcept -> reference
...
-//! a precondition check in debug mode.
+//! a precondition check.
 template<size_type N>
 CELER_CONSTEXPR_FUNCTION bool Bitset<N>::test(size_type pos) const
-    noexcept(!CELERITAS_DEBUG)
 {
-    CELER_EXPECT(pos < N);
+    if (CELER_UNLIKELY(pos >= N))
+    {
+        throw std::out_of_range("Bitset out of bounds: pos="
```
sethrj: Is this device compatible? I don't need to strictly hold to the standard behavior wrt exceptions... (And I guess it wouldn't be constexpr on device because that could use non-constexpr functions `assert` and `__trap`)?

esseivaju: It probably doesn't make sense to call this on device in the first place; maybe we can remove it or make it host-only?

sethrj: Maybe we should just remove it then? I still don't get why we'd treat the bounds checking differently in this case than for our other quasi-standard-library classes Span and Array.

esseivaju: Because the only purpose of that `test` function is to have always-on bound checks, otherwise, there is an equivalent `bool operator[]` where we can use the usual `CELER_EXPECT` bound check. I have updated it to make it device-compatible by using `assert` but am fine removing it if you think it's better.

sethrj: I gotcha. Let's try first to be consistent with our other utilities: our `array` doesn't have an `at` function, and generally we always bound check if debug, never if not. So perhaps just delete this function?

esseivaju: Okay, I'll also add a `CELER_EXPECT` where needed

sethrj (approved):

PR #1549, opened 2024-12-20
Define precision-agnostic Constant class

sethrj:
The constants in Celeritas should avoid double-precision arithmetic in GPU kernels if using single-precision mode, but we also sometimes want to use constants in double-precision operations, particularly working with external codes on the host side that only use doubles. This indeed caused an error message in Geant4 because of `double msc_theta_limit{constants::pi}` since `double(float(pi)) > double(pi)`.

My solution is a `Constant` class that tries to conform to the precision being used with it. Multiplied by a double-precision value, it's a double; multiplied by float it's a single-precision operation; and multiplied by an int preserves it as a Constant.

A couple of positive effects of this PR:
- Eliminated `m_pi` since our constant is now double-precision even when building `real_type = float`.
- Prevented accidental runtime call of `sqrt(euler)`; instead we now precalculate it.
- Fixed roundoff error in pi described above
- Improves precision when translating geant4 objects

The main negatives are:
- More places require explicit casting to real_type
- Using RealQuantity instead of Quantity (whose type is unclear when based on a Constant unit type)


test/celeritas/Units.test.cc
```diff
@@ -17,6 +17,13 @@ namespace units
 {
 namespace test
 {
+//---------------------------------------------------------------------------//
+// Locally replace the Celeritas "real" expectatoin for one that forces
```
amandalund:
```suggestion
// Locally replace the Celeritas "real" expectation for one that forces
```

test/orange/surf/detail/InvoluteSolver.test.cc
```diff
@@ -628,7 +628,7 @@ TEST(Components, line_angle_param)
     // beta = -pi/3
     {
         real_type u = 0.5;
-        real_type v = std::sin(pi / 3);
+        real_type v = std::sin(pi / real_type{3});
```
pcanal: This (type of) change gives me pause.  The intent of 'real_type' being set to `float` is to switch calculation to using the single floating point processors.  In order for this to happen properly and consistently, with the current PR code, we must never remember to add the explicit casting to real_type of all integer that are use with constants (and then passed to a function); as in this change.

It is likely that we might let in the code the 'obvious' spelling, i.e.: `real_type v = std::sin(pi / 3);` (which currently using the double version of `std::sin`).

The question is whether this (potential) cost is worth the advantage (i.e. if I understand correctly, being able to write `Constant x = y * 2;` rather than `Constant y = y * 2.0` to get the constant calculation in double).

We "can" catch both type of errors during PR review.  It feels that the `Constant` creation is going to be 'rarer' (and more obvious) than the usage inside formula, so maybe we should switch the operation with integer to default to `real_type` rather than `double`.

sethrj: Actually the way I rewrote `pi` at the start of this file, `constexpr real_type pi{constants::pi};`, means this change is unnecessary. I think my initial attempt in that testing file was to reference the constant instead of define a local `real_type` to preserve the same results.

Below is what I wrote under the misapprehension that `pi` was a constant:

---

> This (type of) change gives me pause. The intent of 'real_type' being set to float is to switch calculation to using the single floating point processors. In order for this to happen properly and consistently, with the current PR code, we must never remember to add the explicit casting to real_type of all integer that are use with constants (and then passed to a function); as in this change.

Not here: since the standard library doesn't take Constant as operators, this would be a compiler error without the cast, and that forces us to be specific about whether this is a "setup time"/host calculation or "runtime" calculation.

> It is likely that we might let in the code the 'obvious' spelling, i.e.: `real_type v = std::sin(pi / 3)`; (which currently using the double version of std::sin).

Even if `std::sin` was a function that takes (and returns) a Constant, we can't implicitly cast to a `real_type`.

> so maybe we should switch the operation with integer to default to real_type rather than double

This seems odd too, because `2` is higher precision than `2.0` which is higher precision than `2.0f` (aka possibly `real_type{2}`). I think demoting to runtime precision is the wrong move there.

Since it's ambiguous, what we *could* do is completely eliminate operations with integers, forcing the explicit use of `x = y * Constant{2}`, `x < Constant{0}`, etc.: but I judged that caused more code changes and was a little awkward.

src/celeritas/phys/InteractionUtils.hh
```diff
@@ -58,7 +58,8 @@ struct ExitingDirectionSampler
     template<class Engine>
     inline CELER_FUNCTION Real3 operator()(Engine& rng)
     {
-        UniformRealDistribution<real_type> sample_phi(0, 2 * constants::pi);
+        UniformRealDistribution<real_type> sample_phi(
+            0, real_type(2 * constants::pi));
```
amandalund, line 62: Would there be anything wrong with using `2.0 * constants::pi` instead of `real_type(2 * constants::pi)` in these cases?

amandalund (approved):

src/celeritas/phys/InteractionUtils.hh
sethrj: I think that would work with `()` but not `{}`, because the `2.0 * pi` is a double expression, which can't be implicitly converted to a single precision under C++11 rules.

What I was actually thinking, since we have so many `sample_phi` in our code, is a free function `sample_twopi(RNG&)` (or `sample_azi` or something) that we could just call.

test/orange/surf/detail/InvoluteSolver.test.cc
pcanal: > . I think demoting to runtime precision is the wrong move there.

But with `pi` being a real_type`, both `pi / 3` and `pi / real_type{3}` do the same thing (demote to float).

> rewrote pi at the start of this file, `constexpr real_type pi{constants::pi};`

I missed that.  I think it actually make the code much harder to understand/read.  If one must reduce the code length, I would do at most: `constexpr auto pi = constant::pi;`, i.e. I would ***not*** change the type (and If you must change the type, I would change the name too `constexpr real_type real_type_pi {constants::pi};`

Anyway, there is 2 types of usages I had in mind:
```
(a) Constant 3pi = constant::pi * three;
(b) real_type v = overloaded_numerical_function( constant_pi * three );
```
The question is how to spell `three` (or anything else) in both case.  As you pointed out `(a)` is nice with the current PR:
```
 Constant 3pi = constant::pi * 3;
```
but `(b)` fails to compile because there is no implicit conversion.  So one possible spelling with the current PR is
```
real_type v = overloaded_numerical_function( real_type{constant::pi * three} );
```
This looks fine albeit a bit annoying.  Did you consider making the conversion to `real_type` implicit? (it probably would require so `#ifdef` :(  in the `Constant` definition)

sethrj: Since it was a unit test I wanted to minimize changes over the file, especially since evaluating at a different precision level would change the answers.

For (b) I think explicitly casting is the better approach because a constant is "full precision" (we could even make them long doubles or use some `constexpr` extended precision class to get more accurate combined "constant" expressions) but we need to evaluate it in a floating point context: so telling explicitly which level of floating point precision actually improves matters IMHO.

pcanal: > Since it was a unit test I wanted to minimize changes over the file

I think that is counter-productive here.  The tests should follow the same patterns that we would want in the code otherwise they pollute various code investigation (including but not limiting to 'grepping for stuff' but also learn how things should work in Celeritas).

I.e. I think the PR is good as is except for needing to remove any 're-declaration' of constants to not be `Constant`

sethrj: That's a very good point. I'll evaluate the changes...

src/celeritas/em/params/WentzelOKVIParams.cc
```diff
@@ -43,7 +43,7 @@ WentzelOKVIParams::from_import(ImportData const& data,
         if (!coulomb)
         {
             // Set the maximum scattering angle for Wentzel VI MSC
-            return constants::pi;
+            return real_type(constants::pi);
```
pcanal, line 46: Just a slight note that this PR uses both `real_type(...)` and `real_type{...}` which is 'fine' but vaguely inconsistent.

pcanal: and a 3rd spelling: `static_cast<real_type>(...)`

test/celeritas/field/FieldPropagator.test.cc
```diff
@@ -42,8 +42,8 @@ namespace celeritas
 namespace test
 {
 //---------------------------------------------------------------------------//
-using constants::pi;
-using constants::sqrt_three;
+constexpr real_type pi{constants::pi};
```
pcanal, line 45: This might have been missed as part of the commit "Use 'typical' usage of constant in unit tests "

test/corecel/math/BisectionRootFinder.test.cc
```diff
@@ -57,6 +58,7 @@ TEST(Bisection, trigometric)
 
     BisectionRootFinder find_root{f, tol};
 
+    constexpr auto pi = static_cast<real_type>(constants::pi);
```
pcanal, line 61:
```
constexpr auto real_type_pi = static_cast<real_type>(constants::pi);
```

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -87,7 +86,8 @@ TEST(PolygonUtilsTest, convex)
     VecReal2 oct{8};
     for (size_type i = 0; i < 8; ++i)
     {
-        oct[i] = {std::cos(2 * pi * i / 8), std::sin(2 * pi * i / 8)};
+        oct[i] = {std::cos(static_cast<real_type>(2 * constants::pi * i / 8)),
```
pcanal, line 89: This is an interesting pattern (since it is a test the actual usage here does not matter).  In this case do we want:
```
std::cos(static_cast<real_type>(2 * constants::pi * i / 8));
```
Which has a run-time double multiplication (by `i`)
or
```
std::cos(static_cast<real_type>(2 * constants::pi) * i / 8);
```
which the `2*` in double precision at compile time but both `* i / 8` in (potentially) single precision at run-time
or
```
std::cos(static_cast<real_type>(2 * constants::pi / 8) * i);
```
with only the `* i` in (potentially) single precision at run-time but a different ordering of the operations?

test/orange/surf/detail/InvolutePoint.test.cc
```diff
@@ -17,7 +17,9 @@ namespace detail
 {
 namespace test
 {
-using constants::pi;
+constexpr real_type pi{constants::pi};
```
pcanal, line 20: This might have been missed as part of the commit "Use 'typical' usage of constant in unit tests "

src/orange/g4org/SolidConverter.cc
```diff
@@ -511,7 +511,7 @@ auto SolidConverter::polyhedra(arg_type solid_base) -> result_type
     auto const& params = *solid.GetOriginalParameters();
 
     // Convert from circumradius to apothem
-    double const radius_factor = std::cos(m_pi / params.numSide);
+    double const radius_factor = cospi(1 / static_cast<double>(params.numSide));
```
pcanal, line 514: Is the `double` intentional or should it be `real_type`?

pcanal: Given the code below, it seems intentional.

PR #1550, opened 2024-12-20
Reduce `celer-sim` memory usage and improve its `CELER_LOG` transport output

stognini:
For large simulations the `SimulationResult::events` (which is a `std::vector<TransporterResult>`) continuously grows in-memory, being flushed to disk only at the end of `celer-sim`. This can lead to huge memory usage---in one test, `celer-sim` reached >25 GB. Another effect that slows down the run and clutters the I/O is the `CELER_LOG` message issued at every new event. To address these, this PR adds two options to `RunnerInput`:
- New `bool transporter_result`, which only accumulates event data on request.
- New `size_type print_progress`, which if set to `N` (`N>0`), will only issue the `CELER_LOG` message after `N` events.


app/celer-sim/Transporter.hh
```diff
@@ -45,6 +45,7 @@ struct TransporterInput
     size_type max_steps{};
     bool store_track_counts{};  //!< Store track counts at each step
     bool store_step_times{};  //!< Store time elapsed for each step
+    size_type print_progress{};  //!< Print progress every N events
```
sethrj: Just default this to 1 so that it prints every event, and validate that it's greater than zero

app/celer-sim/Transporter.cc
```diff
@@ -203,6 +204,33 @@ void Transporter<M>::accum_action_times(MapStrDouble* result) const
...
+template<MemSpace M>
+void Transporter<M>::progress(EventId const id,
+                              size_type const num_primaries) const
+{
+    CELER_EXPECT(num_primaries > 0);
+
+    auto const id_val = id.unchecked_get();
+    std::string prim = (num_primaries == 1) ? " primary" : " primaries";
+    std::string msg = "Event " + std::to_string(id_val) + ": transporting "
+                      + std::to_string(num_primaries) + prim;
```
sethrj: You're doing a lot of work that may not be output. This is also very different from how the output message is constructed: `operator<<` does nothing if the log level is suppressed, but here you're doing a lot of conversion and object construction.

app/celer-sim/Transporter.cc
```diff
@@ -203,6 +204,33 @@ void Transporter<M>::accum_action_times(MapStrDouble* result) const
...
+    {
+        if (id_val % print_progress_ == 0)
+        {
+            CELER_LOG_LOCAL(status) << msg;
+        }
+    }
+    else
+    {
+        CELER_LOG_LOCAL(status) << msg;
+    }
```
sethrj: Just do
```suggestion
    if (id_val % print_progress_ == 0)
    {
        CELER_LOG_LOCAL(status) << msg;
    }
```
and replace `msg` with all the string stuff you did above. Use `(num_primaries == 1 ? " primary" : " primaries")` to avoid making a temporary string.

I'd also put the message writing up in the anonymous namespace (there's no need for templating) and put the `if (id_val % print_progress_ == 0) print_progress();` in the transporter function itself.

app/celer-sim/RunnerInputIO.json.cc
```diff
@@ -178,6 +180,8 @@ void to_json(nlohmann::json& j, RunnerInput const& v)
     LDIO_SAVE_OPTION(slot_diagnostic_prefix);
     LDIO_SAVE(write_track_counts);
     LDIO_SAVE(write_step_times);
+    LDIO_SAVE_OPTION(transporter_result);
+    LDIO_SAVE_OPTION(print_progress);
```
sethrj: Just use `LDIO_SAVE` since their value is meaningful even when not explicitly set.

app/celer-sim/celer-sim.cc
```diff
@@ -131,16 +134,23 @@ void run(std::istream* is, std::shared_ptr<OutputRegistry> output)
...
             // Run a single event on a single thread
-            CELER_TRY_HANDLE(result.events[event] = run_stream(
-                                 StreamId(get_openmp_thread()), EventId(event)),
-                             capture_exception);
+            CELER_TRY_HANDLE(
+                (run_input->transporter_result)
+                    ? result.events[event] = run_stream(
+                          StreamId(get_openmp_thread()), EventId(event))
+                    : run_stream(StreamId(get_openmp_thread()), EventId(event)),
+                capture_exception);
```
sethrj:
```suggestion
            TransporterResult event_result;
            CELER_TRY_HANDLE(event_result = run_stream(
                          StreamId(get_openmp_thread()), EventId(event)),
                capture_exception);
            if (run_input->transporter_result)
            {
                result.events[event] = event_result;
            }
```

sethrj (commented):
Good thought: we definitely don't want people with 20GB memory usage by accident 😬

app/celer-sim/Transporter.hh
stognini: Oh, It works with either 0 or 1. So as it is, it already prints all events. Thought that this way would make its use a bit easier...

sethrj: There's extra work that goes into checking zero as a special case. But  maybe it's still fine to do `if (print_progress == 0 || (events % print_progress == 0)) ...` but then you have two different selections mapping to the same behavior, and "print_progress = 0" enabling printing at every event is weird.

app/celer-sim/Transporter.cc
stognini: I see here why you suggested `print_progress` to be always positive above...

app/celer-sim/celer-sim.cc
stognini: Do you think this copy could be slow if the event is large? I tried to avoid a copy.

app/celer-sim/Transporter.cc
stognini: Good point. The suggested refactoring with `if (id_val % print_progress_ ==0) { print_progress(...); }` also looks cleaner.

app/celer-sim/celer-sim.cc
sethrj: You're right, I should have std::move. The readability trumps potential (small) copy cost and increased object size

app/celer-sim/Transporter.hh
stognini: Agreed. Initialized to `1` and made sure value is positive.

pcanal:
>  being flushed to disk only at the end of celer-sim.

Why do we not flush it every events? (I.e. store in the result in a `TTree`/`RNTuple`, this seems to be the typical use case for that)?

stognini:
> Why do we not flush it every events? (I.e. store in the result in a TTree/RNTuple, this seems to be the typical use case for that)?

Oh, the ROOT output is fine. This is dumped at the end via `JsonPimpl`, at `celer-sim.cc::274`
```cpp
output->output(&cout);
```
and is necessary if `CELERITAS_USE_ROOT=OFF`.

app/celer-sim/celer-sim.cc
```diff
@@ -113,7 +116,7 @@ void run(std::istream* is, std::shared_ptr<OutputRegistry> output)
 
     // Start profiling *after* initialization and warmup are complete
     Stopwatch get_transport_time;
-    if (run_input->merge_events)
+    if (run_input->transporter_result && run_input->merge_events)
     {
         // Run all events simultaneously on a single stream
         result.events.front() = run_stream();
```
amandalund: This means if `transporter_result` is false we use the multi-stream logic in the `else` block even if we're running in combined event mode.

app/celer-sim/Transporter.cc
```diff
@@ -34,6 +34,22 @@ namespace celeritas
...
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Print progress after N events when requested.
+ */
+void progress(EventId const id, size_type const num_primaries)
+{
+    CELER_EXPECT(num_primaries > 0);
```
amandalund, line 45:
```suggestion
    CELER_EXPECT(id);
    CELER_EXPECT(num_primaries > 0);
```

app/celer-sim/Transporter.cc
```diff
@@ -34,6 +34,22 @@ namespace celeritas
...
+    CELER_LOG_LOCAL(status)
+        << "Event " + std::to_string(id.unchecked_get()) + ": transporting "
+               + std::to_string(num_primaries)
+               + (num_primaries == 1 ? " primary" : " primaries");
```
amandalund: If you use the stream insertion operator instead of concatenating strings with `+` you don't need the `to_string()` calls and avoid creating temporary strings:
```suggestion
        << "Event " << id.unchecked_get() << ": transporting " << num_primaries
        << (num_primaries == 1 ? " primary" : " primaries");
```

app/celer-sim/Transporter.cc
```diff
@@ -34,6 +34,22 @@ namespace celeritas
 {
 namespace app
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Print progress after N events when requested.
+ */
+void progress(EventId const id, size_type const num_primaries)
```
sethrj:
```suggestion
void log_progress(EventId const id, size_type const num_primaries)
```

Otherwise one might think `progress` is a verb (pro-GRESS) rather than noun (PRO-gress) .. isn't English fun?!

app/celer-sim/Transporter.cc
sethrj: Make sure to stream into the logger: otherwise you're *still* doing a bunch of useless string creations.
```suggestion
        << "Event " << id.unchecked_get() << ": transporting "
        << num_primaries << (num_primaries == 1 ? " primary" : " primaries");
```

app/celer-sim/Transporter.cc
```diff
@@ -127,8 +145,12 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries) -> TransporterResult
...
     ScopedSignalHandler interrupted{SIGINT};
 #endif
-    CELER_LOG_LOCAL(status)
-        << "Transporting " << primaries.size() << " primaries";
+
+    auto const evt_id = primaries.front().event_id;
+    if (evt_id.get() % print_progress_ == 0)
+    {
+        progress(evt_id, primaries.size());
+    }
```
sethrj: Since we're planning to get rid of the "event ID" part of the primaries at some point (a group of primaries should always be from the same event, and it's just going to be input to the track initializers) we might even want to just have an internal counter.
But anyway, since `evt_id` is just used inside the conditional you can use this C++17 syntax (and add a check for the event ID being set I suppose)
```suggestion
    if (auto const evt_id = primaries.front().event_id;
        evt_id && evt_id.get() % print_progress_ == 0)
    {
        progress(evt_id, primaries.size());
    }
```

app/celer-sim/celer-sim.cc
```diff
@@ -116,7 +119,8 @@ void run(std::istream* is, std::shared_ptr<OutputRegistry> output)
     if (run_input->merge_events)
     {
         // Run all events simultaneously on a single stream
-        result.events.front() = run_stream();
+        (run_input->transporter_result) ? result.events.front() = run_stream()
+                                        : run_stream();
```
sethrj: I think you should do the same thing here: `auto event_result = run_stream()`, and `if (transporter_result) result.events.front() = std::move(event_result);`

sethrj (approved):
One last change I think :)

app/celer-sim/celer-sim.cc
```diff
@@ -131,16 +135,25 @@ void run(std::istream* is, std::shared_ptr<OutputRegistry> output)
...
     }
+
     result.action_times = run_stream.get_action_times();
     result.total_time = get_transport_time();
     record_mem = {};
-    output->insert(std::make_shared<RunnerOutput>(std::move(result)));
+    if (run_input->transporter_result)
+    {
+        output->insert(std::make_shared<RunnerOutput>(std::move(result)));
+    }
```
amandalund: We should still write the simulation result (timing info, number of CPU threads) either way.
```suggestion
    output->insert(std::make_shared<RunnerOutput>(std::move(result)));
```

amandalund (approved):
One more from me too :) thanks @stognini!

pcanal:
> Oh, the ROOT output is fine. This is dumped at the end via JsonPimpl, at celer-sim.cc::274

So turning on ROOT 'solves' the problem.  The issue is only if the output is JSON.  correct?  If it is correct, do we really store 25GB in a text file !? (humm okay that was the total memory, was it the file size then?)

stognini:
> So turning on ROOT 'solves' the problem. The issue is only if the output is JSON. correct? If it is correct, do we really store 25GB in a text file !?

Using ROOT wouldn't solve it, since we'd end up having both ROOT and json outputs. So I proposed the option to minimize the json info, which is also useful if you just want to time the execution.

I did not check how large a final file would have been since I did not worry about directing the `cout` to a file. But when the 25-ish GB run finished, it took probably 10+ minutes to flush the result to the terminal...

pcanal:
> , it took probably 10+ minutes to flush the result to the terminal...

Yikes :(

> Using ROOT wouldn't solve it, ...So I proposed the option to minimize the json info

Indeed text based information should be minimized.  What kind of information is in the json file?  What information is being dropped?  What information is being kept?  How much really needs to be in a file that is directly human?  What is the maximum sizes where the JSON is still actually readable? (few MB?).   How much of the information that is not strictly needed in the human readable file can be stored in (yet another) ROOT file?	
(sorry if there is already answers to some of those questions in other places)

sethrj:
@pcanal This is the "diagnostic" JSON output which wasn't meant to have much data at all. It's primarily meant to help us analyze performance data and do simple postprocessing. Because it's performance related, we want it to be easy/automatic to stand up, which rules ROOT out. For output that scales with increasing problem difficulty (e.g. the transport results) maybe we should hardcode/use a backdoor environment variable for the maximum number of iterations before we stop saving and writing.

amandalund:
@stognini out of curiosity, how many events were you running and was this with both `write_track_counts` and `write_step_times` off?

One final modification you could make in the `Transporter` is to only allocate memory for [these guys](https://github.com/celeritas-project/celeritas/blob/b9c98bcd6ae4d9aa0042aef987f09952452599c0/app/celer-sim/Transporter.cc#L115-L118) if `store_track_counts_` is true.

stognini:
> @stognini out of curiosity, how many events were you running and was this with both write_track_counts and write_step_times off?

If I recall correctly that was a 10M events, 1 primary each. And yes both of those options were off. I turned off all json options I could, related to writing info.

pcanal:
> Because it's performance related, we want it to be easy/automatic to stand up, which rules ROOT out.

humm ... 'easy/automatic' is clearly doable in ROOT (delta the learning curve).  Things that a human readable (json) format would give you are ability to grep through random-ish text, quick visual inspection (vi/less) [albeit if was remembers the structure RDataFrame::Display might work as well). and web display (github actions).  Most of which disappear if the file is too large.

Example of 'quick' textual display of a ROOT file content:
```
root.exe -b -l tutorials/hsimple.root -e 'ROOT::RDataFrame df(*ntuple); df.Range(100,200).Display({"px","py"},100)->Print();' -q | less
```
See also `RDataFrame`'s [user manual](https://root.cern/manual/data_frame/) and [documentation](https://root.cern/doc/v634/classROOT_1_1RDataFrame.html).  (This is a bit more typing but much clearer/flexible that the previous `TTree::Scan` version).

>  maybe we should hardcode/use a backdoor environment variable for the maximum number of iterations before we stop saving and writing.

This makes some sense. (Although the next request is also a start (i.e. capture from the 100th events to the 150th events instead of the first 50 events).

sethrj:
I mean easy for other people to install, or (like our external) automatic.

sethrj:
@stognini @pcanal I think we got all the major kinks worked out, please take a look at the discussions and resolve as needed.

pcanal (approved):

stognini:
Thought I answered this thread, just to notice that I probably never hit the `comment` button during the holidays:

> Example of 'quick' textual display of a ROOT file content

For the `celer-sim` I ruled out any textual output that we get from ROOT, despite it having some nice features (such as `TBufferJSON::ConvertToJSON`) to avoid making it a necessary dependency for I/O, since we want to be able to test it on any system (e.g. summit's powerpc was a limitation).

sethrj (approved):

amandalund:
@stognini one last update that would also slightly improve the memory usage:

> One final modification you could make in the `Transporter` is to only allocate memory for [these guys](https://github.com/celeritas-project/celeritas/blob/b9c98bcd6ae4d9aa0042aef987f09952452599c0/app/celer-sim/Transporter.cc#L115-L118) if `store_track_counts_` is true.

PR #1552, opened 2024-12-23
Update expected safeties in Vecgeom tests and CMake version range

mrguilima:
Update CMake version range to 3.31.
Update expected values of safeties of a few Vecgeom test instances.


sethrj:
Thanks for this @mrguilima ! It's going to have to wait until I update our CI, but I'm going to try to make headway on that during some downtime today...

CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ set(CMAKE_USER_MAKE_RULES_OVERRIDE
   "${CMAKE_CURRENT_LIST_DIR}/cmake/CeleritasMakeRulesOverride.cmake")
 
 project(Celeritas VERSION "${Celeritas_VERSION}" LANGUAGES CXX)
-cmake_policy(VERSION 3.18...3.28)
+cmake_policy(VERSION 3.18...3.31)
+cmake_policy(SET CMP0175 OLD)
```
sethrj: Please revert this line; instead, we should fix whatever is broken in the CMake policy. https://cmake.org/cmake/help/latest/policy/CMP0175.html#cmp0175

CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ set(CMAKE_USER_MAKE_RULES_OVERRIDE
   "${CMAKE_CURRENT_LIST_DIR}/cmake/CeleritasMakeRulesOverride.cmake")
 
 project(Celeritas VERSION "${Celeritas_VERSION}" LANGUAGES CXX)
-cmake_policy(VERSION 3.18...3.28)
+cmake_policy(VERSION 3.18...3.31)
```
sethrj: It seems like this is redundant with the `cmake_minimum_required` above... maybe delete the cmake policy here? (I somehow thought it had to be after `project`)

sethrj:
@mrguilima I think once you fix the remaining tests this should be good to go! https://github.com/celeritas-project/celeritas/actions/runs/12635001598/job/35204556112

CMakeLists.txt
mrguilima: The build succeeds with CMP0175 set to OLD, but fails with NEW.  I will try to fix it...

sethrj: @mrguilima Don't worry about it. Can you build without these changes? It would be better for you to revert this file and  make this PR *just* about the VG update.

sethrj: Moved changes to #1563 .

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -1237,7 +1210,7 @@ TEST_F(FourLevelsGeantTest, tracking)
     }
     {
         SCOPED_TRACE("From outside edge");
-        auto result = this->track({-24, 10., 10.}, {1, 0, 0});
+        auto result = this->track({-24.1, 10., 10.}, {1, 0, 0});
```
sethrj: This is no longer the outside edge as the test description shows. I assume you did this because it's failing with the surface geometry? It's important to keep this edge case—if we need to, we can skip this one for the surface geometry.

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -1278,7 +1251,7 @@ TEST_F(FourLevelsGeantTest, tracking)
     {
         // Formerly in linear propagator test, used to fail
         SCOPED_TRACE("From just outside world");
-        auto result = this->track({-24, 10, 10}, {1, 0, 0});
+        auto result = this->track({-24.1, 10, 10}, {1, 0, 0});
```
sethrj: Looks like this test is actually a duplicate of the one above: I'm just going to delete this one.

sethrj (approved):

test/geocel/vg/Vecgeom.test.cc
mrguilima: The reason I moved the point was because the behavior is drastically different between v1.* (points on the border of the world are [OUTSIDE]) and v2.* (those points are *in* the World volume).
Hence all the arrays had a different number of elements.
I guess your suggestion is best handled in a test without propagation.

test/geocel/vg/Vecgeom.test.cc
mrguilima: The difference is the Geant vs. no-Geant, right?  Converted geometry vs. VGDML-based?
I thought you wanted to keep because of that.  I am fine with removing one of them.

sethrj: I meant it's a duplicate of `From outside edge` a few lines above 😅

test/geocel/vg/Vecgeom.test.cc
sethrj: I get it, but a very important purpose for unit testing is to test and document edge cases and how they change from one version to another. What we *might* want to do is add code for the volume-based geometry so that, when initialized outside or on a surface, it moves to the nearest volume.

PR #1554, opened 2024-12-27
Adapt bitset word size

esseivaju:
I extracted bite set-related changes from #1539, this uses the smallest possible word size to store the requested bits.


src/corecel/cont/Bitset.hh
```diff
@@ -44,7 +46,10 @@ class Bitset
   public:
     //!@{
     //! \name Type aliases
-    using word_type = unsigned int;
+    using word_type = std::conditional_t<
+        (N <= 8),
+        std::uint8_t,
+        std::conditional_t<(N <= 16), std::uint16_t, std::uint32_t>>;
```
pcanal: ~~Technically~~ Maybe:
```suggestion
        std::conditional_t<(N <= 16), std::uint16_t, 
            std::conditional_t<(N <= 32), std::uint32_t, std::uint64_t>>>;
```
~~and even
suggestion
        std::conditional_t<(N <= 16), std::uint16_t, 
            std::conditional_t<(N <= 32), std::uint32_t,
                std::conditional_t<(N <= 64),  std::uint64_t, void /* or anything that makes it fails */ >>>>;~~

~~unless we already have somewhere else a static_assert< sizeof(word_type)*8 >= N);~~

sethrj: It automatically becomes a multi element array

sethrj: But we don't want 64-bit words on GPU (where 32 is native I think)... if anything it should be `celeritas::size_type` which is `unsigned int` when GPU is enabled or `std::size_t` when not.

sethrj (approved):

src/corecel/cont/Bitset.hh
esseivaju: Using `celeritas::size_type` now, which will be 32bits on GPU and 64bits (unsigned long) otherwise.

sethrj (approved):
Nifty!

PR #1555, opened 2024-12-29
Update docker and Spack CI

sethrj:
This finally upgrades the CUDA and ROCm CI images to latest versions.

- [x] CUDA docker image
- [x] ROCm docker image
- [x] Update VecGeom in spack image
- [x] ~~Incorporate VecGeom updates from~~ Disable failing VecGeom tests, to be reverted in  #1552
- [x] Ensure "gold" tests are all working


.github/workflows/build-spack.yml
```diff
@@ -66,28 +73,35 @@ jobs:
...
           fetch-tags: true # to get version information
       - name: Setup Spack
-        uses: spack/setup-spack@0ce61628ed8a32a5664cf01a0e0b5a4834a3b413 # 2024/03
+        uses: spack/setup-spack@5792f7c7055c3707819380b5e3831d2be6e64b6c # 2024/12
         with:
           ref: ${{env.SPACK_REF}}
           buildcache: true
           color: true
           path: spack-src
+      - name: Patch spack
```
pcanal: Should we add an explanation of why these patches are needed (and for 'how long')?

README.md
```diff
@@ -108,22 +108,26 @@ $ make && ctest
...
-    - GCC 8.4, 12.3
-    - Clang 10.0, 15.0
-    - GCC 11.3 + NVCC 11.8
-    - HIP-Clang 15.0
+- Compilers and standard:
+    - GCC 8, 11, 12, 14
+    - Clang 10, 15, 18
+    - GCC 11.5 + NVCC 12.6
+    - ROCm Clang 18
+    - C++17 and C++20
```
pcanal, line 117: The standard coverage might depend on the version of Geant4, is it worth mentioning here?

README.md
```diff
@@ -108,22 +108,26 @@ $ make && ctest
...
 
 Partial compatibility and correctness is available for an extended range of
 Geant4:
 - 10.5-10.7: no support for tracking manager offload
 - 11.0: no support for fast simulation offload
-- 11.1-11.2: [no support for default Rayleigh scattering cross section](see
+- 11.1-11.3: [no support for default Rayleigh scattering cross section](see
   https://github.com/celeritas-project/celeritas/issues/1091)
 
+Note also that navigation bugs in older versions of Geant4 and VecGeom can
```
pcanal, line 129: Do we know the upper bound version number of those bugs (i.e. version number after which those bugs are all gone)?

scripts/cmake-presets/ci-ubuntu-github.json
```diff
@@ -52,7 +52,7 @@
       "cacheVariables": {
         "CELERITAS_USE_Geant4":  {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_HepMC3":  {"type": "BOOL", "value": "ON"},
-        "CELERITAS_USE_ROOT":    {"type": "BOOL", "value": "ON"},
+        "CELERITAS_USE_ROOT":    null,
```
pcanal, line 55: Does this means 'use the default value'? (i.e. do not set it?)

scripts/cmake-presets/ci-ubuntu-rocm.json
```diff
@@ -1,11 +1,11 @@
 {
-  "version": 3,
-  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0},
+  "version": 6,
+  "cmakeMinimumRequired": {"major": 3, "minor": 23, "patch": 0},
   "configurePresets": [
     {
       "name": "base",
-      "displayName": "CentOS7/rocm-docker default options",
```
pcanal, line 7: Do we have test on ~~CentOS8~~ AlmaLinux and/or EL9?

pcanal (approved):
A few questions but LGTM.

esseivaju (approved):

README.md
sethrj: I think "bug-free" is an unbounded function for any code 😄 but 11.3 is the first one with no known bugs (twisted object navigation being some of the more recent issues).

scripts/cmake-presets/ci-ubuntu-github.json
sethrj: Correct; it forces the variable to be "unset". This allows us to use some configurations with ROOT and some without, needed because I didn't want to do two separate ROOT builds (one c++17, one c++20).

README.md
sethrj: All these apply to the 11.0.4 version being tested, which is compatible with both 17 and 20.

.github/workflows/build-spack.yml
sethrj: Sure, I've added a note (and added a bit more doc to the top of the file about the spack updating). Hopefully in the next week or so the patches will be merged and I can rebuild the stack with a clean version of Spack.

scripts/cmake-presets/ci-ubuntu-rocm.json
sethrj: Rocky9 (the new CUDA docker image) is a very close relative to both of those right? I think it's the spiritual successor to CentOS.

sethrj:
Thanks @pcanal and @esseivaju ! Please resolve the issues if you feel they've been addressed.

scripts/cmake-presets/ci-ubuntu-rocm.json
pcanal: > Rocky9 (the new CUDA docker image) is a very close relative to both of those right?

it is a detail but why did you go with Rocky9 then? (AlmaLinux/EL9 are used in HEP, I am not sure Rocky is)

README.md
pcanal:
```suggestion
Note also that known navigation bugs in versions of Geant4 older than v11.3 and VecGeom v1.x will cause test failures in Celeritas.
```

scripts/cmake-presets/ci-ubuntu-rocm.json
sethrj: Availability of CUDA docker images I think? I looked at the Linux family tree to try to make a determination

pcanal: Fair enough.  Thanks.

Issue #1556, opened 2024-12-30
Restructure stepper/transporter/state/params


sethrj:
The Stepper should be refactored since it's grown from a single-purpose "do one step" to a general interface:
1. I don't think it's right that it owns the state, since we have to extract data from it occasionally. We also set `params.set_state(stream_id.get(), step_->sp_state())` in `accel` so we can do data reductions. (In total there's three accessors to get the state.)
2. It's no longer a function-like object since we have so many other functions attached to it.
3. The `step(primaries)` is a little odd.
4. The stepper owns an action sequence. Although the action times currently are "local", we ought to have one action sequence per Params.
5. We should have an "end run" called over all states, and "begin run" should be over all states as well. This is necessary for global reductions.

Current construction order relating to actions:
1. Core params input. Some of the components (e.g. physics) add actions.
2. Core params, which also adds actions.
3. During stepper construction, action sequence
4. Then state, which initializes auxiliary data
6. Then `begin_run` is called on each state

Use cases:
1. Geant4 offloading: LocalTransporter registers a pointer to the state with the global parameters
2. Celer-sim: Runner creates multiple transporters which are used to step in a parallel openmp loop

sethrj:
Inputs:
- Imported data from Geant4
- Callback functions to generate input functions: physics processes
- Options to construct built-in Celeritas actions (e.g. SD callback, calo, diagnostics)
- Callback to generate additional actions, including along-step (if we do an along-step action manager, it should include a mapping of {region, particle} -> along_step; that stuff shouldn't be done in the CoreParams)

Parallelism/multithread requirements:
- Be able to do parallel (all track slots, streams/CPU threads, MPI processes) operations at beginning, event boundaries (or "batches" of histories in parallel), and end
- Match Geant4 allocation requirements: state creation, use, and destruction must be on same thread
- (??) In thread-independent model, allow different memory management models so that one shared params can be used by different states

# During construction

On the **[M]ain** thread or in **[P]arallel**:

## BeginOfRunAction (main)

- **[M]** Load import data
- **[M]** Create global objects that don't have any associated actions (geometry, material, particle)
- **[M]** Create core params and incorporate any additional user actions
- **[M]** Create action sequence from core params, which should "finalize" the number of actions and anything action-related
- **[M]** Create state collection(s) with unallocated per-stream states, based on maximum local threads; potentially have CPU and GPU state collections side by side, incorporate knowledge of parallel MPI processes

## BeginOfRunAction (worker)

[...]

PR #1557, opened 2024-12-30
Fix accel examples and related CI issues

sethrj:
A dumb typo (using a variable name instead of dereferencing the variable name...) meant we weren't testing the `accel` examples in CI. The examples have been failing since at least #1501 . I'm also enabling debug logging so we can figure out initialization ordering for #1556 .

An additional error is that "error on doxygen warning" was disabled for PR documentation builds since I changed the name of the workflow in  `#1243`. This meant a minor documentation error in #1518 was causing issues.


scripts/ci/test-examples.sh
```diff
@@ -32,12 +32,14 @@ build_local() {
...
 
+export CELER_LOG=debug CELER_LOG_LOCAL=debug
+
 cd "${CELER_SOURCE_DIR}/example/minimal"
 build_local
 ./minimal
 
 
-if [ -z "CELER_DISABLE_ACCEL_EXAMPLES" ]; then
+if [ -z "${CELER_DISABLE_ACCEL_EXAMPLES}" ]; then
```
pcanal, line 43: Maybe the original writer meant
```suggestion
if [ -v CELER_DISABLE_ACCEL_EXAMPLES ]; then
```
(true is the variable is set/exist)

sethrj: I definitely just forgot to dereference, but I did't know that `-v` existed. Is that a standard POSIX `test` function? I use `-z` so that you can `export FOO=` to unset a variable, since it's easier to `export FOO=` rather than `env -u FOO` (and more intuitive to understand a false/empty variable has the same behavior as "not set"... but then again this means `FOO=0` is "true").

If we use `-v` then how should I change
https://github.com/celeritas-project/celeritas/blob/6ec7f84e9a407aae378624937c37850a603ca161/.github/workflows/build-spack.yml#L232-L237
?

pcanal: > Is that a standard POSIX test function?

It appears that it is not available everywhere (eg. MacOS).

> If we use -v then how should I change

This is a good reason not to use `-v`

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -237,14 +247,13 @@ CELER_FUNCTION real_type WentzelHelper::calc_screening_coefficient(
    C_{TF} = \frac{1}{2} \left(\frac{3\pi}{4}\right)^{2/3}.
  * \f]
  */
-CELER_CONSTEXPR_FUNCTION real_type WentzelHelper::screen_r_sq_elec() const
+CELER_CONSTEXPR_FUNCTION auto WentzelHelper::screen_r_sq_elec() -> MomentumSq
```
pcanal, line 250: Is the removal of the `const` intentional?

pcanal (approved):
LGTM

src/celeritas/em/xs/WentzelHelper.hh
sethrj: Yep, it's a static function now since it doesn't use any member data: just a constexpr helper function.

PR #1558, opened 2025-01-01
Add support for Geant4 navigator through offload interface

sethrj:
Since #1403, the per-thread states are deallocated on the main thread rather than on the workers. This meant that we couldn't use Geant4 navigation because it has thread-local allocators.

This small patch re-enables that capability.


amandalund (approved):
Thanks @sethrj!

PR #1561, opened 2025-01-02
Enable more dependencies in clang-tidy checks

esseivaju:
Turn on more dependencies in the clang-tidy preset. Files not compiled in the preset would trigger an error in an incremental run as they are not present in the `compile_commands.json` database, `clang-tidy` wouldn't know how to find included headers.


.github/workflows/pull_request.yml
```diff
@@ -50,7 +50,6 @@ jobs:
     needs: [all-prechecks]
     uses: ./.github/workflows/build-docker.yml
   build-spack:
-    needs: [all-prechecks]
```
sethrj: Revert please

.clang-tidy
```diff
@@ -19,6 +19,7 @@ Checks: |
   -cppcoreguidelines-init-variables
   -cppcoreguidelines-pro-type-member-init
   -cppcoreguidelines-avoid-do-while
+  -cppcoreguidelines-avoid-non-const-global-variables
```
sethrj, line 22: With this we might be able to restore `examples` to the list of files we check...

sethrj (approved):

.clang-tidy
esseivaju: I wanted to see if `ImplementationFileExtensions` would filter out that root file since it uses a different extension, but it seems that CMake is still checking it..

esseivaju:
@sethrj since you are updating the CI to `clang-tidy-18` in #1555 I think we should wait until that is merged, running the same setup locally I do not get any warnings but [this](https://github.com/celeritas-project/celeritas/actions/runs/12588649748/job/35086975368?pr=1561) seems to be checking disabled checks...

sethrj:
Would this also be a good time to fix the "needs newer clang-tidy" NOLINT markers?

esseivaju:
yes, planning on doing so

sethrj:
(Blocking #1550)

Issue #1204, opened 2024-04-25
Generate input documentation from app structs


sethrj:
- [ ] Rename `app` namespace to `app_g4` or `app_sim` or just `g4`/`sim`
- [ ] Rename `Run(ner)?Input.hh` to `Input.hh` and also struct
- [ ] Decorate `Input` structs with more doxygen
- [ ] Add `*/app/Input.hh` to `file(GLOB _DOXYGEN_SOURCE` in `doc/CMakeLists.txt`
- [ ] Add `doc/api/app.rst` and include the input structs from the applications
- [ ] Link to the app "API" from the usage section

sethrj:
Or potentially generate C++ structs and conversion code from Python data models... 🤔

sethrj:
Closing in favor of #2029

PR #1562, opened 2025-01-06
Add global Celeritas input definition

sethrj:
**UPDATE**: this is ready to go. For a preview of the documentation, see https://sethrj.github.io/celeritas/user/usage/input.html  and https://sethrj.github.io/celeritas/user/implementation/core-physics/problem-def.html#setting-up-problems . For an example of how the input will be used, see  769fb8bfbcf13d5e3eafc052aed1f2a62bfca981 .

To cleanly reuse the same input across all front ends, I've refactored the input into:
- System parameters (GPU, MPI, logger) that *have* to be set up at the very beginning
- Code to set up Geant4 (for standalone apps only (celer-sim and celer-g4),   not applicable to user app/framework integration)
- Problem definition
- Code to load data from Geant4, i.e. setting up what part of Geant4 to pull into Celeritas
- Event files (for standalone apps only)
- Custom hook for adding user-defined processes, user-defined diagnostics, etc. (for framework only).

There's a lot of code changes, but most of it is moving stuff around and adding "adapter" code. There is also a *lot* of documentation.

Currently the code is "orphaned" i.e. it doesn't do anything useful; the next stage will be to merge the shared params setup code from the three front ends into a single place in `celeritas/setup/`.

---

Related to #1556 ,  I'd like to specify a unified mechanism for creating core params, states, etc. before continuing work on that front. Simultaneously (in light of the Optical work) we've been discussing a "refactored" IO where we specify Celeritas inputs to classes separately from the classes themselves. This ties in with #1204 and #1263 and the unification of `RunInput`/`RunnerInput`/`SetupOptions`.

As a first step, I've defined a new directory and namespace `inp` where we can move *all* the "input"/"option" classes into. For now we can hand-roll these in C++ (and adapt them from existing input classes), but I'd like to keep working on https://github.com/celeritas-project/celerpy/issues/4 . We should get some working C++ structs before writing the python code and adapters, but ChatGPT seems able to help here, since it was able to generate this from the new `inp::Tuning` C++ class:
```py
class Tuning(BaseModel):
    """Set up system/tuning parameters that don't affect physics.

    Defaults:
        - `track_order`: `init_charge` on GPU, `none` on CPU.
    """

    capacity: StateCapacity
    "Per-process state sizes."

    optical_capacity: Optional[StateCapacity] = None
    "Per-process state sizes for *optical* tracking loop."

    num_streams: PositiveInt = 0
    "REMOVE: number of streams."

    device: Optional[Device] = None
    "Optional: activate GPU."

    track_order: Optional[str] = None
    "Track sorting and initialization."

    warm_up: bool = False
    "Perform a no-op step at the beginning to improve timing measurements."

    seed: PositiveInt
    "Random number generator seed."

    environ: Dict[str, str]
    "Environment variables used for program setup/diagnostic."
```

I used ChatGPT to help with the initial translation of SetupOptions, RunInput, and RunnerInput to the new unified `inp::Input` class.

Note that for now we will need to restrict the along-step field choices to our built-in ones. I'd like to extend so that we can use more type-safe attributes (e.g. `ParticleId`) along @hhollenb 's line of thinking.

If this sounds good, I'll add documentation to the manual about how I expect this will be used. Follow-on PRs will:
- create core params from `inp` data
- support JSON->inp conversion (and output?? maybe instead we should have output of the lower-level data)
- move `FooParams::Input`, `FooParams::Option`, `FooOptions`, etc. into `inp` structures
- and then continue with refactoring.

Talking this through with @hhollenb I think we'll want to separate "standalone" driver functionality (primary generator definition) from the rest of it.

## Goals

There are three categories of input types:
- Exclusive Celeritas inputs such as state size and other diagnostic/tuning  parameters. Also some parameters we *cannot* deduce from Geant4: e.g., sensitive detector attributes.
- Parameters that we want to use to drive Geant4 for celer-g4, or pull from Geant4 if we're not. This would include geometry definition (GDML), EM parameters, active processes, scoring meshes.
- Problem setup options that we cannot directly understand from Geant4 but must be provided directly to lower-level Celeritas objects. In particular, we need to be able to allow users to add custom processes, magnetic fields, etc.

### Reduce duplication of CoreParams construction and front end input

Currently celer-sim, celer-g4, and direct framework coupling all have slightly different interfaces. They also all independently construct CoreParams. I want to be able to use all the built-in functionality of Celeritas from any front end.

### Reduce duplication of input parameters

Currently for celer-g4, we have (using `linear_loss_limit` as an example):
> JSON/macro → GeantPhysicsOptions → CelerEmPhysicsList → Geant4 → GeantImporter → ImportEmParameters → PhysicsParamsOptions::linear_loss_limit → PhysicsParamsScalars

This is a crazy amount of duplication and hard to keep track of where parameters come from and go. The duplication makes it harder to extend: to add support for region/particle-dependent limits we'd have to change a lot of code.

### Unify and refactor input to classes

A data-oriented approach will let us restructure e.g. Sim/TrackInit params, as well as Physics, since the input can be pulled from multiple input structs. It also provides a single location where input structs are defined, rather than scattered throughout the codebase. That will ease the transition to a JSON front end and remove a lot of the assignment of `foo.x = bar.x` between input classes.

### Enable well-defined extension points

The current generic `SetupOptions` only gives one way to extend the actions: `make_along_step`. Instead we want to be able to extend actions, along-step, physics processes, etc.


src/celeritas/inp/Physics.hh
```diff
@@ -0,0 +1,99 @@
...
+    real_type step_limit{};
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Set up physics options.
+ *
+ * TODO: refactor ignore_processes so it ties in the with IO classes.
+ */
+struct Physics
```
hhollenb, line 83: This struct feels a little weird to me. Is it meant to be deciding which IO method to use for importing physics, or is it meant to be what is directly passed into `PhysicsParams`? It feels like it's in charge of handling different importing routines that require different parameters (from Geant4, from files, etc) while also managing further levels of logical filtering and modification of the imported data (ignoring processes, selecting particles / processes, etc).

src/celeritas/inp/Field.hh
```diff
@@ -0,0 +1,57 @@
...
+    FieldDriverOptions driver_options;
+};
+
+//---------------------------------------------------------------------------//
+// TODO: move field input here, make variant input/filename?
+using RZMapField = ::celeritas::RZMapFieldInput;
+
+//---------------------------------------------------------------------------//
+//! Field type
+using Field = std::variant<NoField, UniformField, RZMapField>;
```
hhollenb, line 57: Having everything be structs + variants which will have their respective functions to dispatch to does make me think of just a bog-standard OOP approach. Not necessarily the best idea but might be good to keep in mind why OOP wouldn't fit and what the final goal of this input format is.

hhollenb (commented):
I'm a little confused about how this input fits into building celeritas. It seems like there's quite a bit of logic between the `inp` structs and what would get passed directly to `Params`.

My personal imagining is that there's a layer of purely input / user configuration that gets canonicalized into a single format that is used to initialize celeritas. Inputs from prebuilt binaries like `celer-sim` and `celer-g4` would have default and easy ways to build the canonical input. If users link celeritas as a library and want to hook in to alter the defaults and force their desired configurations, then they can do so in between the "build input" level and the "initialize celeritas" level. This could very much be me misunderstanding the purpose / existing state of the initialization as well.

src/celeritas/inp/Physics.hh
sethrj: I'm thinking some classes will get passed in the whole `Input` struct, others (e.g. sim params) might get one or two. Maybe this structure will be shared by the geant "setup" (if we're driving Geant4), and the geant "exporter", and the physics.

Maybe `physics_file` should be a variant switching between "import physics from file" and "import physics from Geant4"? Although that ends up pulling in a bunch of EM options as well 🤔 maybe I should sketch this part out a bit more.

src/celeritas/inp/Field.hh
sethrj: I don't understand your comment. I'm thinking of this as a DOP approach, so we can serialize all the options as JSON and communicate them back and forth; and also allow code restructuring by passing all the data into the Params constructors if we need to.

I primarily envision the structs as a driver for built-in Celeritas objects. I also will be adding callback functions for parts that we want to be arbitrarily extensible (i.e., adding actions). I think that such a compartmentalized approach is more flexible and would work better than the Geant4/dd4hep "override a class to one thing or another" method, if that's what you mean by OOP.

sethrj:
I've been thinking about this a bit more too. Let's say the goal is to have *all* the input data, aside from the in-memory geometry definition (which we do have an issue to create a standalone 'model' from). We have I think three kinds of inputs:
- Exclusive Celeritas inputs: state size, other diagnostic and tuning parameters
- Parameters that we want to use to drive Geant4 for celer-g4, or pull from Geant4 if we're not. This would include magnetic field, EM parameters, active processes, and maybe someday the scoring setup.
- Problem setup options that we *cannot* directly understand but must be provided directly to lower-level Celeritas objects. In particular, we need to be able to allow users to add custom processes, magnetic fields, etc.

I'll be thinking more about this before our meeting this afternoon...

sethrj:
@hhollenb Based on our discussions: thoughts?

options

sethrj:
options

Updated. I don't think it's feasible to define an "updater" that merges two structs (that would be super tedious, potentially confusing, etc.). So what I think we want is that the input is defined once (or in the case of framework/user application, not at all), and then updated sequentially by different helper functions:
- Geant4 importer: fulfills the same role as the current I/O, but the input (and updated input) will pass through Celeritas-specific options too.
- Data file reader: given G4 data file paths and the materials in the input, import SB tables and such
- JSON reader: can be used to update tuning/diagnostic options from an existing export file
- Application-defined input adjuster: will set up field, user process callbacks, etc. by directly modifying the input structs.
- (from Soon) User-friendly physics list constructors

Then the input gets passed into the params construction and Celeritas takes it from there (using callback functions to add user actions, processes, etc.).

The input should ideally *only* contain nonderivative data: e.g., microscopic cross sections, not macroscopic. We could eventually have the inputs contain other currently hardcoded data, like the element parameters in `RayleighModel`, so that in the future they could be migrated to different formats.

hhollenb:
I think it looks good! Addresses my primary concern about having a separate interface for input and initializing, and looks extensible enough for different application types. I'm sure actually implementing and refactoring will have its own slew of issues, but I think this is a solid starting point.

hhollenb:
A brief thought I had while looking at it during the meeting: the driving Geant4 flow looks pretty much identical to the framework flow. If we treat driving Geant4 as a kind of "trivial framework", it would reduce our use cases and act as an example framework for us to develop around or a starting point for users.

sethrj:
The reason I broke out the "driving geant4" was to ensure there was a good place to put the "standalone input" code for setting up Geant4. In practice, yes the Geant4 importer should be doing most of the work.

sethrj:
One gotcha is that we want some "system" stuff (device, MPI, environment, logger) to be configured before running *any* celeritas code, including the importers (which use profiling and logger).

I think I'll make `System` a separate configuration that has to be set first... 🤔

doc/index.rst
```diff
@@ -62,18 +62,159 @@ Celeritas [#celeritas_vers]_ is a new MC particle transport code designed for
...
 
 .. [#celeritas_vers] This documentation is generated from Celeritas |release|.
 
+Background
+==========
+
+.. note:: This background section is largely a quote of the Celeritas R&D
+   report :cite:`celer-rd-2024`.
+
+The first investigation of using GPU to accelerate Geant4 computing was a
```
amandalund:
```suggestion
The first investigation of using GPUs to accelerate Geant4 computing was a
```

doc/index.rst
```diff
@@ -62,18 +62,159 @@ Celeritas [#celeritas_vers]_ is a new MC particle transport code designed for
...
+
+Celeritas implements physics processes and models for transporting electron,
+positron, and gamma particles. Initial support is being added for muon EM
+physics.  Implementation details of these models and their corresponding Geant4
+classes are documented in :ref:`api_em_physics`.
+
+Optical Physics
+---------------
+
+Optical physics are being added to Celeritas to support various high energy
```
amandalund:
```suggestion
Optical physics is being added to Celeritas to support various astroparticle, high energy
```

src/celeritas/inp/Events.hh
```diff
@@ -0,0 +1,115 @@
...
+#include "geocel/Types.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/phys/PDGNumber.hh"
+
+namespace celeritas
+{
+namespace inp
+{
+//---------------------------------------------------------------------------//
+//! Sample a single point
```
amandalund:
```suggestion
//! Generate at a single point
```

src/celeritas/phys/PhysicsParams.hh
```diff
@@ -73,9 +73,8 @@ class ParticleParams;
  *   is used. If it is 2+, the spline interpolation is used for energy loss
  *   using the specified order. Default value is 1
  *
- * NOTE: min_range/max_step_over_range are not accessible through Geant4, and
- * they can also be set to be different for electrons, mu/hadrons, and ions
- * (they are set in Geant4 with \c G4EmParameters::SetStepFunction()).
+ * \todo min_range/max_step_over_range should be extended to be per particle
+ * (or particle "grouping", e.g. electron/muon+hadron/light ion/ion).
```
amandalund: Done at least for electron/muon? It might not hurt to keep the note somewhere that while we could set these parameters in Geant4 there are no accessors for them.

src/celeritas/phys/PrimaryGeneratorOptions.cc
```diff
@@ -47,6 +48,68 @@ void check_params_size(char const* sampler,
                    << "' distribution needs exactly " << required_params);
 }
 
+//---------------------------------------------------------------------------//
+// Helper: Convert energy distribution to inp::EnergyDistribution
+inp::EnergyDistribution inp_from_energy(DistributionOptions const& options)
```
amandalund, line 53: Why do we need to convert the distribution/primary generator options to the inputs (rather than read the inputs directly from json)? Is this for backward compatibility?

src/celeritas/inp/Import.hh
```diff
@@ -0,0 +1,87 @@
...
+ * SeltzerBergerReader, AtomicRelaxationReader to fill cross section data.
+ * Since Geant4 data structures don't provide access to these, we must read
+ * them ourselves.
+ */
+struct GeantDataImport
+{
+    //! Livermore photoelectric data directory (G4LEDATA)
+    std::string livermore_dir;
+    //! Particle cross section data directory (G4PARTICLEXS)
+    std::string particle_dir;
```
amandalund, line 68: Why do we specify the paths rather than using the environment variables?

src/celeritas/inp/PhysicsProcess.hh
```diff
@@ -0,0 +1,57 @@
...
+ * Construct a physics process for bremsstrahlung.
+ */
+struct BremsProcess
+{
+    std::optional<SBBremsModel> sb{std::in_place};
+    std::optional<RelBremsModel> rel{std::in_place};
+    std::optional<MuBremsModel> mu;
+
+    //! Use a unified relativistic/SB interactor
+    bool combined_model{true};
```
amandalund:
```suggestion
    bool combined_model{false};
```

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -0,0 +1,44 @@
...
+{
+};
+
+//---------------------------------------------------------------------------//
+//!@{
+//! \name Model aliases
+//! \todo rename `em/model` to match
+
+using SeltzerBergerBremsModel = SBBremsModel;
+using RelativisticBremModel = RelBremsModel;
```
amandalund: What are the model/process aliases for?

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -0,0 +1,44 @@
...
+struct RelBremsModel
+{
+    //! Account for LPM effect at very high energies
+    bool enable_lpm{true};
+};
+
+//! Muon bremsstrahlung model
+struct MuBremsModel
+{
+};
```
amandalund, line 39: Is this where imported data like micro xs will eventually go?

src/celeritas/inp/Physics.hh
```diff
@@ -0,0 +1,98 @@
...
+    std::optional<CoulombScatProcess> coulombscat{std::in_place};
+    std::optional<IoniProcess> ioni{std::in_place};
+    std::optional<AnniProcess> anni{std::in_place};
+    std::optional<ConversionProcess> conversion{std::in_place};
+    std::optional<PhotoelectricProcess> photoelectric{std::in_place};
+    std::optional<RayleighScatProcess> rayleighscat{std::in_place};
+#endif
+
+    //!@{
+    //! \name Energy loss and slowing down
```
amandalund, line 35: So this will replace the `GeantPhysicsOptions`, as well as the `ImportEmParameters` and `PhysicsParamsOptions`? And will be used to set the options in Geant4 and then updated from the imported data?

src/celeritas/inp/Model.hh
```diff
@@ -0,0 +1,33 @@
...
+
+namespace celeritas
+{
+namespace inp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Set up geometry/material model.
+ */
+struct Model
```
stognini, line 22: `GeometryModel` or `GeoModel`?

src/celeritas/inp/Physics.hh
```diff
@@ -0,0 +1,98 @@
...
+struct Physics
+{
+    //! Enable electromagnetic physics
+    std::optional<EmPhysics> em{std::in_place};
+
+    //! Enable optical photon physics
+    std::optional<OpticalPhysics> optical;
+
+    //! Enable hadronic physics
+    std::optional<HadronicPhysics> hadronic;
```
stognini, line 92: Probably a new `struct DecayPhysics` also makes sense here

src/celeritas/inp/Model.hh
sethrj: It has material and region properties too, so I'd say it's more than just the geometry

stognini: Thought that too, but also was feeling odd to call the geometry (and materials/regions) `model`, even though I don't have a better name yet...

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,109 @@
...
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Set up system/tuning parameters that don't affect physics.
+ *
+ * Defaults:
+ * - \c track_order: \c init_charge on GPU, \c none on CPU
+ *
+ * \todo \c seed doesn't really belong here; not sure where to put it though
```
stognini: How about `Problem` directly? If you need, say, 10 runs with the same source (e.g. just reusing the same `hepmc3` input), you want to control the seed for each new problem definition, while keeping everything else constant.

src/celeritas/inp/Model.hh
sethrj: That's what we ended up calling it in Omnibus 🤷‍♂️

src/celeritas/inp/Events.hh
```diff
@@ -0,0 +1,119 @@
...
+    size_type num_events{};
+    //! Number of primaries per event
+    size_type primaries_per_event{};
+
+    //! Distribution for sampling source position
+    ShapeDistribution shape;
+    //! Distribution for sampling source direction
+    AngleDistribution angle;
+    //! Distribution for sampling source energy
+    EnergyDistribution energy;
```
amandalund, line 87: By replacing the `std::function`s in the `PrimaryGenerator::Input` with these `variant`s we'll lose some generality: before, we could programmatically construct a primary generator that could sample from arbitrary spatial/angular/energy distributions, while now we'll be limited to the few that are predefined in the input. That might be fine though if a user is never going to construct a generator themselves, i.e. they'll only ever define it through the input?

src/celeritas/inp/Physics.hh
```diff
@@ -0,0 +1,98 @@
...
+    std::optional<AnniProcess> anni{std::in_place};
+    std::optional<ConversionProcess> conversion{std::in_place};
+    std::optional<PhotoelectricProcess> photoelectric{std::in_place};
+    std::optional<RayleighScatProcess> rayleighscat{std::in_place};
+#endif
+
+    //!@{
+    //! \name Energy loss and slowing down
+
+    //! Use quadratic spline interpolation for energy loss
```
amandalund: I think Geant4 uses cubic spline interpolation by default (which it looks like is also different from what we do)?

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,109 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Set up per-process state capacities.
+ *
+ * Capacities are defined as the number per application process: this means
+ * that in a multithreaded context it implies "strong scaling" (i.e., the
+ * allocations are divided among threads), and in a multiprocess context it
+ * implies "weak scaling" (the problem size grows with the number of
+ * processes).
```
amandalund: This is only true for our standalone celeb-sim, right? I.e. in Geant4-integrated apps we currently don't divide the track/initializer capacities by the number of threads.

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,109 @@
...
+    //! Maximum number of queued primaries+secondaries
+    size_type initializers{};
+    //! Maximum number of secondaries (automatic if zero)
+    size_type secondaries{};
+
+    //! Maximum number of simultaneous events (zero for Geant4 integration)
+    size_type events{0};
+
+    //! Minimum number of primaries before generating and advancing a step
+    size_type primaries{};
```
amandalund: I had to read the note about this previously being called `auto_flush` to understand what it was (although I don't love that name either)... not sure if there's a better name that's still simple while making it more obvious? `offload_threshold`, `buffered_primaries`...?

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,109 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * When using GPU, change execution options that make it easier to debug.
+ */
+struct DeviceDebug
+{
+    //! Launch all kernels on the default stream
+    bool default_stream{false};
+    //! Synchronize the stream after every kernel launch
+    bool sync_stream{false};
```
amandalund: What if `DeviceDebug::sync_stream` is false and `Timers::action` is true? Do we validate, does one take precedence and we warn? Is only one a user option and the other just updated based on that value?

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,109 @@
...
+struct Tuning
+{
+    //! Per-process state sizes
+    StateCapacity capacity;
+
+    //! Per-process state sizes for *optical* tracking loop
+    std::optional<StateCapacity> optical_capacity;
+
+    //! Number of streams
+    size_type num_streams{};
```
amandalund: This will be calculated from the number of threads rather than set by the user, right?

src/celeritas/track/TrackInitParams.hh
```diff
@@ -18,6 +18,9 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 /*!
  * Manage persistent track initializer data.
+ *
+ * \todo \c max_events could potentially be per thread, not global? And we
+ * should differentiate between user events and events in flight.
```
amandalund, line 23: Maybe? I think it's only used for the track counters, so having a max events per thread would make mapping to the event ID more complicated (unless I'm misunderstanding the comment).

amandalund (commented):
I think this is looking great @sethrj! It will definitely be nice to have a unified input and reduce a lot of the duplication. Just a few minor suggestions and clarifications, but I think most of this will become clearer as it’s fleshed out more.

src/celeritas/inp/Tuning.hh
sethrj: That seems like too important a place to put something that usually won't even need changing... 🤔

src/celeritas/phys/PhysicsParams.hh
sethrj: Yeah I added this before you opened your other PR 😄

src/celeritas/phys/PrimaryGeneratorOptions.cc
sethrj: Yes, exactly. This lets us (for now) use the existing problem inputs. I wanted to stabilize the internals before replacing the JSON I/O.

src/celeritas/inp/Import.hh
sethrj: The readers have the option to use it; we'll potentially set these values in unit tests for example. I've updated the docs to be more clear.

src/celeritas/inp/PhysicsProcess.hh
sethrj: I took this from the default value in `BremsstrahlungProcess::Options` but you're right, the `RunnerInput` and `ProcessBuilderOptions` default is false so let's default to that (and I'll change the other value to match).

src/celeritas/inp/PhysicsModel.hh
sethrj: The todo here is because we've been inconsistent about our model/process names; I think we should unify them somewhat as above (formalizing the names in the input), then later rename the physics classes to match.

src/celeritas/inp/PhysicsModel.hh
sethrj: Yep. I'm going to add some more example uses.

src/celeritas/inp/Physics.hh
sethrj: That is indeed the plan!

src/celeritas/inp/Events.hh
sethrj: That's correct. We could add a `std::function` overload, but honestly we overengineered the PG in the first place since we never even used those functionalities. I think we can reduce the functionality to what's required, then expand again as needed.

src/celeritas/inp/Physics.hh
sethrj: Hmmm good to know; should we generalize this since Lance's code supports it, or just give linear/cubic options?

src/celeritas/inp/Tuning.hh
sethrj: Good point: I'll check this and update the conversion function to match. We should take a consistent approach for all our front ends, though. The next move after this will be to create a "multi-state" (one object linking states for all threads for reductions) which will let us understand the total scaling numbers.

src/celeritas/inp/Tuning.hh
sethrj: Yeah, it's "kind of" a capacity but also kind of not? And is it slightly different from photons (where we coudl wait for "generators" or "primaries")

src/celeritas/inp/Tuning.hh
sethrj: Maybe this should be a std::optional so we can default it? Ideally our python front end will take care of most of the "automatic" stuff but that won't apply to the `FrameworkInput` mode. I think if not optional, we'd respect both values and then warn that the meaning of the times isn't necessarily GPU work (but it could be interesting to see where the CPU/syncs are happening?)

src/celeritas/inp/Tuning.hh
sethrj: Ideally... right now some of our diagnostics and such require this parameter, but we'll set it automatically for the time being.

src/celeritas/track/TrackInitParams.hh
sethrj: My thought is that it's nonsensical to use `max_events` in a multi-stream context in the likely case that different streams have different event ids.

src/celeritas/inp/PhysicsProcess.hh
amandalund: Yep, combined model still only works for single-element materials, we probably missed that we needed to update the brems options... a hazard of all the duplication :)

src/celeritas/inp/PhysicsModel.hh
amandalund: Gotcha, sounds good.

src/celeritas/inp/Physics.hh
amandalund: Since we're unifying all those options, we might want to clarify in the documentation which parameters
 - Can be used to set up Geant4
 - Are imported from Geant4
 - Are only options for Celeritas

src/celeritas/inp/Physics.hh
amandalund: This will only set the interpolation method in Celeritas, not Geant4, right? If we have that capability we might as well add an option to use it.

Also related to #1544 : do we plan to add support for cubic spline interpolation? Or do we think our polynomial interpolation is "close enough"?

src/celeritas/inp/PhysicsProcess.hh
```diff
@@ -29,25 +29,37 @@ struct BremsProcess
...
 using EPlusAnnihilationProcess = AnniProcess;
-using GammaConversionProcess = ConversionProcess;
 using PhotoelectricProcess = PhotoelectricProcess;
 using RayleighProcess = RayleighScatProcess;
 
-using MuBremsstrahlungProcess = MuBremsProcess;
-using MuIonizationProcess = MuIoniProcess;
-using MuPairProductionProcess = MuConversionProcess;
+using MuBremsstrahlungProcess = BremsProcess;
+using MuIonizationProcess = IoniProcess;
```
amandalund: For the processes I sort of prefer the full names (mainly don't love `Anni` or `Scat`):

`ComptonScatteringProcess` or `ComptonProcess`
`CoulombScatteringProcess` or `CoulombProcess`
`RayleighScatteringProcess` or `RayleighProcess`
`AnnihilationProcess`
`BremsstrahlungProcess`
`IonizationProcess`
`PairProductionProcess`
`PhotoelectricProcess`

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -11,31 +11,79 @@ namespace celeritas
...
+using MollerBhabhaModel     = MollerBhabhaModel;
+using MuBetheBlochModel     = MuBetheBlochModel;
+using RayleighModel         = RayleighModel;
+#endif
+
+using BetheHeitlerModel = BHPairProdModel;
 using MuBremsstrahlungModel = MuBremsModel;
+using MuPairProductionModel = MuPairProdModel;
+using RelativisticBremModel = RelBremsModel;
+using SeltzerBergerModel = SBBremsModel;
```
amandalund: More naming discussion:

- I slightly prefer the more descriptive names to the abbreviations for the models (`BetheBloch` to `BB`; `BetheHeitler` to `BH`; `SeltzerBerger` to `SB`), although I'm guilty of abbreviating the distributions and other helpers
- Maybe keep as `BetheBlochModel` and `BraggModel`, or to be consistent change to `MuBetheBlochIoniModel` and `MollerBhabhaIoniModel`
- `BHPairProdModel` could alternatively be `RelPairProdModel` (which it now technically is)

src/celeritas/inp/PhysicsProcess.hh
sethrj: I suspect Geant4 abbreviates `Ioni` because no one was sure whether to use British or American spelling 🙃 And yeah I'm not thrilled about Anni/Scat either but I think the full names are a bit long (and inconsistent if we abbreviate Brems but not the others)

src/celeritas/inp/PhysicsModel.hh
sethrj: I remember @pcanal way back in the day weighing in on our first model naming scheme and preferring the longer spelling. My biggest concern is consistency, then I suppose readability should be second, then writability. I think the model names start to blur a bit as we start changing them slightly to apply different corrections... RelPair is BetheHeitler with LPM so maybe we should call it LPMBHPairProd 🤪

src/celeritas/inp/PhysicsProcess.hh
amandalund: Yeah, `WhateverScatteringProcess` probably is too long, and we should be consistent. Though we're not abbreviating photoelectric...

Lengthwise these are pretty similar:

ComptonScatProcess
CoulombScatProcess
RayleighScatProcess
IoniProcess
AnniProcess
BremsProcess
PairProdProcess
PhotoelectricProcess

ComptonProcess
CoulombProcess
RayleighProcess
IonizationProcess
AnnihilationProcess
BremsstrahlungProcess
PairProductionProcess
PhotoelectricProcess

but I think the second is a bit clearer/nicer sounding...

src/celeritas/inp/PhysicsModel.hh
amandalund: I prefer LPMBHPP

src/celeritas/track/TrackInitParams.hh
amandalund: Trying to recall what our plan was for the event IDs/counters:
- In Geant4-integrated apps, `max_events` is always set to one since #1447 
  - Currently the internal event ID is always set to zero and tracks IDs are sequential from zero up to the total number of tracks transported on that _stream_, not in each event
  - The goal is to have a single global counter for the internal event IDs (which will map to the user event IDs) and a single stream-local track counter that will be reset at the beginning of each event
  - With a single event in flight at a time we should be able to make the track IDs reproducible
- celer-sim event-parallel mode still sets `max_events` equal to the total number of events
- celer-sim combined-event mode will sill need `max_events` track counters

src/celeritas/inp/PhysicsModel.hh
pcanal: 🤮! as far as I know that acronym might just be VFNSFW 😹

pcanal: I still lean for the longer names, they make it a bit easier for newcomers to get into the code without having to search through for the acronyms ...

src/celeritas/inp/PhysicsProcess.hh
pcanal: If there is only one type of 'SomeName' process, I agree with Amanda that the 2nd list is sufficient.  But for the rare rare (non-existent?) case where there  2 kinds (i.e. both a 'SomeName scattering process' and 'SomeName ionization process' then those process could always a 3rd noun. 

Monkey wrench. What about have fun with namespace?:
```
namespace Process {
   struct Compton { ... };
   struct Photoelectric { ... };
}
```
and hence the name are 
```
Process::Compton
Process::PhotoElectric
```
Probably not as readable but allow (if one wants to) to reduce the number of characters used in some places.

pcanal: One could even have (!?):
```
namespace Process {
namespace Muon {
  struct Bremsstrahlung { ... };
  struct Ionization { ... };
}
}
```

sethrj: Process::Muon vs Muon::Process? 🙃

src/celeritas/inp/PhysicsModel.hh
sethrj: Yeah true... but since a lot of these models are effectively just `"<process> optimized for <energy range> with <assumptions> by being a combination of <foo>, modified by <bar>, with <jimmy's> form factors, and <vladimir's> arbitrary tweaks"` (edit: html ate my comment 😅)

pcanal: It is still "in the family of ..." or at least "inspired by ...." ... but more accurately it is "best numerical approximation of the XYZ model/process that we could come up with given the information we have so far (and to be fair, also with a bit of duck tape, chewing gum, self annulling pair of bugs and prayers and somehow it looks like it was passing some sort of test someday, I am sure ... maybe )"  :)

sethrj: All right, so to get this moving I'll delete these aliases for now and we can resume in a later commit :)

src/celeritas/track/TrackInitParams.hh
sethrj: I think that's all exactly right. I've updated https://github.com/celeritas-project/celeritas/issues/1233 to try to better explain the problems and potential avenues of research. For `celer-sim` we don't (yet) care about actual event IDs so we can just use it to measure performance of multiple simultaneous streams.

src/celeritas/inp/Tuning.hh
sethrj: @amandalund Any more thoughts on this? I think it's reasonable to reframe as "Maximum number of primaries that can be buffered before stepping" : we probably want to buffer until we've got as many tracks as possible, *unless* we're worried about trying to overlap Geant4/Celeritas work.

src/celeritas/inp/Tuning.hh
sethrj: I updated the conversion function for the different front ends but will *definitely* check again given our recent "uh oh" moment with performance. It's regrettable that we can't (yet) decouple resource consumption with performance (and sometimes more track slots means slower performance!) so hopefully we can do that someday too.

src/celeritas/inp/Tuning.hh
pcanal: So would't it be `size_t max_in_flight_primaries{};` or `size_t max_buffered_primaries{};`?

sethrj: Everything is a buffer, effectively. The "in flight" aspect of the tracks is that we launch kernels based on the size (or the number of active tracks, depending on track slot options); also for the primaries we *do* launch a kernel to convert them to track initializers. So since everything is a max, and a num, and a buffer, I think `primaries` rather than `max_num_buffered_primaries` 😉

amandalund: Yeah, I think we usually want to buffer as many as we can (e.g. for CMS run3/4 I was setting this to half the initializer capacity). I think `tracks` and `initializers` are clearly capacities, but the others are a bit fuzzier: `secondaries` is really a factor that's multiplied by the number of tracks to get the capacity, and `primaries` is a threshold value that determines when the Celeritas stepping loop starts. I'm not necessarily opposed to those names (vs. something like `max_num_buffered_primaries` ;)), but to a new user they might be confusing. Maybe just adding a little more documentation on what each one means would help?

app/celer-g4/RunInput.cc
```diff
@@ -6,13 +6,173 @@
...
+inp::System load_system(RunInput const& ri)
+{
+    inp::System s;
+
+    if (celeritas::Device::num_devices())
+    {
+        inp::Device d;
+        d.stack_size = ri.cuda_stack_size;
+        d.heap_size = ri.cuda_heap_size;
+        s.device = std::move(d);
```
hhollenb, line 36: Since all struct fields are being initialized, can just write
```
s.device = {
    .stack_size = ri.cuda_stack_size,
    .heap_size = ri.cuda_heap_size
};
```
or if you want the abbreviation, maybe
```
inp::Device& d = s.device;
```
?

app/celer-g4/RunInput.cc
```diff
@@ -6,13 +6,173 @@
...
+
+    // Tuning
+    {
+        inp::StateCapacity capacity;
+        capacity.tracks = ri.num_track_slots;
+        capacity.initializers = ri.initializer_capacity;
+        capacity.secondaries = static_cast<size_type>(ri.secondary_stack_factor
+                                                      * ri.num_track_slots);
+        capacity.primaries = ri.auto_flush;
+        p.tuning.capacity = std::move(capacity);
```
hhollenb: Ditto my previous comment. It's easy to omit a move at the end of an initialization section and relies on warning unused flags to catch.

src/celeritas/inp/Tuning.hh
sethrj: In the *updated* parameters, `secondaries` is actually a size, not a factor. We definitely want most of them to be automatic and I will definitely add documentation.

amandalund: Ah I missed that, so in the future the user will specify a size for the secondary stack rather than a factor? Like the initializers buffer, is this something we eventually hope to be able to reallocate automatically if we run out of memory?

sethrj: Yes exactly, and for now I think we can make it optional by setting it to twice the number of tracks. (For cases like this, I have size_type=0 as a special case meaning "default me", but I guess it would be more clear to have it as an `optional<size_type>` and disallow the `=0` case?)

src/accel/SetupOptions.cc
```diff
@@ -6,12 +6,181 @@
...
+        p.model.geometry = so.geometry_file;
+    }
+    p.diagnostics.output_file = so.output_file;
+
+    p.tuning.num_streams = so.get_num_streams();
+
+    {
+        inp::StateCapacity c;
+        c.tracks = so.max_num_tracks * p.tuning.num_streams;
+        c.initializers = so.initializer_capacity * p.tuning.num_streams;
```
amandalund: Why are we multiplying by `num_streams` here?

src/celeritas/inp/Tuning.hh
amandalund: ~Having trouble finding them, where did you make the updates?~ I think this was clarified in another comment

src/celeritas/inp/Tuning.hh
```diff
@@ -0,0 +1,129 @@
...
+ * - \c device_debug: absent unless device is enabled
+ * - \c optical_capacity: absent unless optical physics is enabled
+ * - \c track_order: \c init_charge on GPU, \c none on CPU
+ *
+ * \todo \c seed doesn't really belong here; not sure where to put it though
+ * \todo \c num_streams is used to set up per-stream state sizes:
+ * perhaps the number of sterams (and processes and events?) should be passed
+ * in to create the multi-thread states.
+ */
+struct Tuning
```
stognini: How about `Control`?

src/accel/SetupOptions.cc
sethrj: The previous number of initializers was per stream, and the new options are *per process*. In the draft code for building the core params, we then have:
```c++
oc_inp.initializer_capacity = ceil_div(optical_capacity.initializers, num_streams);
```

amandalund: I see, so for the time being we'll keep the `RunnerInput` capacities as per-process and `RunInput`/`SetupProcess` capacities as per-stream?

hhollenb (commented):
Looking good so far - just some small comments!

app/celer-g4/RunInput.cc
sethrj: Neither will work because `device` is `std::optional`. The first usage also requires C++20 and we still want to support C++17.

app/celer-g4/RunInput.cc
sethrj: Yep, could definitely construct this as a reference instead of `move`ing.

sethrj: Actually trying to go back through and remove the `move`... I find myself questioning which one of the parameters are optionals and which aren't. From a user perspective, constructing options and doing an rvalue assignment (whether with `std::move` or not) is definitely easier.

An alternative that would be safe, efficient, *and* readable would be to use IIFE:
```c++
    p.tuning.capacity = [&ri]{
        inp::StateCapacity capacity;
        capacity.tracks = ri.num_track_slots;
        capacity.initializers = ri.initializer_capacity;
        capacity.secondaries = static_cast<size_type>(ri.secondary_stack_factor
                                                      * ri.num_track_slots);
        capacity.primaries = ri.auto_flush;
        return capacity;
    }();
```

src/accel/SetupOptions.cc
sethrj: Yep exactly. And see #1591.

app/celer-g4/RunInput.cc
sethrj: I think in c++20 we could preferably do
```c++
s.device = inp::Device{
    .stack_size = ri.cuda_stack_size,
    .heap_size = ri.cuda_heap_size
};
```

amandalund (approved):
Thanks @sethrj! I think there still might be one or two tiny comments that slipped through the cracks? But looks great on my end!

src/celeritas/inp/Physics.hh
sethrj: Currently yes, but when we start building Geant4 using these insted of "geant setup" options then we should use it there too.  I think we should probably use Geant4's spline settings (one or the other) to begin with?

amandalund: I'm not sure whether we actually _can_ control the spline method in Geant4? And if we could I think what we're doing In the `SplineXsCalculator` is still different (it looks like polynomial interpolation, not spline)? Geant4 also removed the spline option from the EM parameters.

app/celer-g4/RunInput.cc
pcanal: In the meantime you can also use:
```suggestion
        s.device = inp::Device{};
        s.device->stack_size = ri.cuda_stack_size;
        s.device->heap_size = ri.cuda_heap_size;
```
or even (less descriptive I guess)
```
        s.device = Device{ri.cuda_stack_size, ri.cuda_heap_size};
```

src/celeritas/inp/Import.hh
```diff
@@ -0,0 +1,91 @@
...
+{
+    //! Livermore photoelectric data directory (optional)
+    std::string livermore_dir;
+    //! Particle cross section data directory (optional)
+    std::string particle_dir;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Update tuning and diagnostic options from an external input file.
```
stognini: Nitpick: update control/tuning? Same for a couple lines below.

stognini (approved):
Looks great @sethrj . Thanks for all the reorganization work.

src/celeritas/inp/Physics.hh
sethrj: I guess we can no longer  control the geant4 spline, so we will need to update our "geant4 import to `inp`" logic to either warn or change it back.

src/celeritas/inp/PhysicsProcess.hh
sethrj: Also with C++ namespacing rules, nested namespaces would get very confusing very quickly...

PR #1563, opened 2025-01-07
Update CMake policy

sethrj:
This updates the CMake support to bleeding-edge 3.31, and fixes an unwitting misuse of the cmake's [add_custom_command]( https://cmake.org/cmake/help/latest/command/add_custom_command.html#generating-files) which is probably why #1516 was so confusing.

### DIFF
--- a/test/celeritas/CMakeLists.txt
+++ b/test/celeritas/CMakeLists.txt
@@ -434,11 +434,6 @@ if(CELERITAS_USE_Geant4 AND CELERITAS_USE_ROOT)
     add_custom_command(
       TARGET update-root-test-data
       POST_BUILD
-      DEPENDS
-        celer-export-geant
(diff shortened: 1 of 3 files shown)


pcanal (approved):
LGTM

test/celeritas/CMakeLists.txt
```diff
@@ -434,11 +434,6 @@ if(CELERITAS_USE_Geant4 AND CELERITAS_USE_ROOT)
     add_custom_command(
       TARGET update-root-test-data
       POST_BUILD
-      DEPENDS
```
pcanal, line 437: Consider adding to the commit logs that `DEPENDS` is not used in case of `Build Event/POST_BUILD add_custom_command` to point out the reason for the invalidity.

PR #1564, opened 2025-01-07
Add 'LogicalTrue' functor

sethrj:
We often have to define `[](Foo const& foo) { return static_cast<bool>(foo); }` for checking things in a `std::all_of`. This adds a helper along the lines of the std library's `logical_not` to do the cast for us.


amandalund (approved):
Nice idea!

PR #1566, opened 2025-01-07
Use enums for indexing bitset

sethrj:
This refactors the Bitset class to index on enumerations rather than using integers. All the bitsets we care about have a fixed meaning for each one of the bits, and we will want to convert existing enums into bitsets: for example,
```c++
enum class MscModelSelection
{
    none,
    urban,  //!< Urban for all energies
    wentzelvi,  //!< Wentzel VI for all energies
    urban_wentzelvi,  //!< Urban below 100 MeV, Wentzel VI above
    size_
};
```
will become 

```c++
enum class MscModelSelection
{
    none,
    urban,
    wentzelvi,
    size_
};
```

and
```c++
        if (options_.msc == MMS::urban || options_.msc == MMS::urban_wentzelvi)
```
will become
```c++
        if (options_.msc[MMS::urban])
```

This PR also flips the `EnumArray` template parameters to mirror the `Array` parameters (value type, then size).


pcanal:
Can we extend the intro (and then commit) with some additional information on the advantages of the change?

src/orange/SenseUtils.hh
```diff
@@ -120,13 +120,16 @@ inline char const* to_cstring(SignedSense s)
...
+    enum class Bits : char
     {
-        sense_bit,
-        is_assigned_bit,
+        sense,
+        is_assigned,
+        size_
     };
 
+    static inline constexpr auto sense_bit = Bits::sense;
```
esseivaju: is inline needed?

src/corecel/cont/EnumBitset.hh
```diff
@@ -37,10 +37,11 @@ namespace celeritas
  * - hash support
  * - construct from string, from_ullong
  */
-template<size_type N>
-class Bitset
+template<class E>
+class EnumBitset
 {
-    static_assert(N > 0, "Zero-sized bitset is not supported");
+    static_assert(std::is_enum<E>::value, "Template parameter must be an enum");
```
esseivaju:
```suggestion
    static_assert(std::is_enum_v<E>, "Template parameter must be an enum");
    static_assert(sizeof(std::underlying_type_t<E>) <= sizeof(size_type), "Size of enum's underlying type must be smaller than size_type");
```

esseivaju (approved):
LGTM! Just a few minor comments

src/orange/SenseUtils.hh
sethrj: Apparently not... from cppreference:
> A constexpr specifier used in the first declaration of a function or static data member(since C++17) implies inline

Guess I was just too used to using C++14!

src/corecel/cont/EnumBitset.hh
sethrj: Since the `size_` member has to fit inside the type, and our enum requires 2^size_ bits, I can't imagine we'd ever be able to define an enum with 2B values 😂

src/corecel/cont/EnumBitset.hh
```diff
@@ -237,28 +240,37 @@ class Bitset<N>::reference
...
         return *this;
     }
 
   private:
     word_type* word_pointer_{nullptr};
     size_type bit_pos_{0};
+
+    CELER_CONSTEXPR_FUNCTION word_type mask() const noexcept
+    {
+        return word_type(1) << bit_pos_;
```
pcanal, line 267: Performance-wise, this is the crux of this functionality (setting might be rare but testing might not be).
Thus there is a shift operation plus a binary or operation for each test.
An alternative implementation is to use (not as easily enforceable from a C++ stand point )
```
enum class MscModelSelection
{
    none,
    urban = 1, 
    wentzelvi = 2, /* really 1  << 1 */
    somethingelse = 4, /* really 1 << 2 */
    size_
};
```
The `Bitset` associated with that now wouldn't have to do the shift operation at run-time (as long as there is less than 64 enum values).

src/corecel/cont/EnumBitset.hh
esseivaju: haha I'd be surprised 😄  but if someone defines
 ```c++
enum class MyEnum : std::uint64_t
{
my_value = 64,
size_
}
```
~~they might run into issues with narrowing conversion when accessing that 64th bit since we `static_cast<size_type>`~~

Nevermind, my suggested check is wrong and won't help with that...

Comment (pcanal) [reply src/corecel/cont/EnumBitset.hh]:
```suggestion
    static_assert(std::is_enum_v<E>, "Template parameter must be an enum");
    static_assert(static_cast<size_t>(E::size_) <= sizeof(size_type), "Largest enum value must be smaller than size_type");
```

esseivaju: Sorry, I got confused 😅 , what we actually want is to avoid truncating the enum value when casting to `size_t`, so my original assertion is right. As Seth said, that would and should never happen, we can just leave it like that.

src/corecel/cont/EnumBitset.hh
sethrj: @pcanal Being constexpr, this should be optimized away. Since we define `CELER_CONSTEXPR_FUNCTION` as `constexpr inline __attribute__((always_inline))` it even gets replaced by a scalar value in completely "unoptimized" code:
https://godbolt.org/z/eYehqebqa

So there's no penalty to indexing with `i` versus `2^i`.

pcanal: `constexpr` is not a guarantee, it is a possibility that depends on all inputs to also be `constexpr`.  Unlike the example you used, `bit_pos_` is a data member and thus not guaranteed to be `constexpr`.  `bit_pos_` is in turn set from the parameter to `operator[]`  and so `options_.msc[MMS::urban]` would indeed have the possibility of executing the shift at compile time but `MscModelSelection m = ...; .... then somewhere else/indirect .... options_.msc[m]` would not be able to.   If only the former  is intended to be used, we could/should/might add code to make the later fail to compile.

sethrj: Oh I see what you mean. Yes we do have some manual bit flags already (ORANGE `VolumeRecord::Flags` for instance). Unlike your enum, the Bitset can be extended to an arbitrary number of bits (though now that I think about it, perhaps we've overengineered it since I can't think of any case where we want more than 32 separate flags...)

Lol when compiling a test code that gets a bit from the Bitset, what's 1 line with `-O1` becomes 58 lines with `-O0` (on ARM with Apple Clang anyway), compared to 8 instructions with the manual mask approach.

```c++
enum class Foo
{
    a,
    b,
    c,
    size_
};

using Bitset = celeritas::EnumBitset<Foo>;

bool get_b(Bitset bs)
{
    return bs[Foo::b];
}

bool get_manual(short int const s)
{
    return s & 0x2;
}
```

So, the question is which do we value more: clarity and ease of use, or performance in debug mode?

sethrj: Actually I have a potentially easy fix for at least part of your problem... just store `mask_` instead of `bit_pos_`.

pcanal: > since I can't think of any case where we want more than 32 separate flags...

I agree that a solution that support 32 flags (or 64 with a `uint64_t`) will be sufficient.

pcanal: > what's 1 line with -O1 

For `constexpr` inputs or run-time inputs?

sethrj: Storing `mask_` saved like one instruction in `-O0`, looking closer at the assembly the main overhead is stack pointer manipulation 🤷 

This is for a constexpr enum access but run-time value being accessed: `get_b` above compiles down to `get_manual`.

Maybe we've approached this wrong: perhaps we just need an "enum bitset" with a single word... @esseivaju what circumstance would require a bitset with compile-time length > 32?

esseivaju: I can't foresee many situations where a length larger than 32bits would be needed. We briefly touched on that in the PR implementing the bitset, since `N` is a constexpr, `N=1` reduces to a single word. So, I'm not sure what would be the benefit of supporting a single word only as it doesn't add that much complexity.

pcanal: > This is for a constexpr enum access but run-time value being accessed:

The part that was surprising was `what's 1 line with -O1`.   What is the one instructions that does both shift and test?  (More likely I miss assumed what that was covering :) ).

sethrj:
I'm going to just close this ... we probably want a "dynamic bitset" like Julien has implemented and a "static bitset" that uses enums.

PR #1567, opened 2025-01-08
Fix Doxygen member group formatting

amandalund:
This fixes a number of instances in the documentation where the description of a member following the `\name` command was being interpreted as the group description, e.g.
before
is now
after


sethrj (approved):
Good catch, thank you!

pcanal (approved):
Thanks.

PR #1568, opened 2025-01-08
Fix unnecessary use of inline constexpr

sethrj:
As I learned from @esseivaju today, `inline` is implied for `static constexpr` class data.


esseivaju (approved):
LGTM!

PR #1572, opened 2025-01-11
Hook up BIH for background_intersection calls

elliottbiondo:
This MR hooks up the `BIHIntersectingVolume` capability from #1479 to `SimpleUnitTracker::background_intersect`. The BIH is used to determine the volume that lies on the other side of a candidate intersecting surface, avoiding a potentially expensive search through a neighbor list. Originally I thought there would be more sophisticated approached that would not involve first calculating all intersections, but I don't actually think there is a way of doing that, mainly because background volumes themselves could be complex (i.e. with internal surfaces).

### DIFF
--- a/src/orange/detail/BIHIntersectingVolFinder.hh
+++ b/src/orange/detail/BIHIntersectingVolFinder.hh
@@ -22,16 +22,16 @@ namespace detail
  * Traverse the BIH to the find the volume that the ray intersects with first.
  *
  * Traversal is carried out using a depth first search. During traversal, the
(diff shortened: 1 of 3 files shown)


elliottbiondo:
@sethrj this is ready

elliottbiondo:
Reverting to draft because CPU ray tracing performance w/background volumes is not looking good w.r.t. `HEAD`. Will update with numbers.

elliottbiondo:
This was the wrong approach. Will reopen another MR once I have something better.

PR #1573, opened 2025-01-13
Add example offloading template

stognini:
This PR adds a new example that acts more as a template structure for users to learn how to either add Celeritas capabilities to their existing framework or build a new Geant4 application (with offloading capabilities) from scratch.

It is technically very similar to the existing `simple-offload.cc`, but it provides more directions to the user, as it is much more similar to the well known examples found in Geant4. The main reasons to add it, in my opinion, is that it provides:
- A short, independent README with code directions and explanations; 
- A very simple and clear `CMakeLists.txt` that teaches the user how to link a Celeritas' install against their project; and
- An equivalent and well known code structure organization, in terms of `.hh/.cc` files, as any Geant4 example, which should make direct comparisons to existing Geant4 codes more intuitive.


example/accel/offload-template/src/Celeritas.cc
```diff
@@ -0,0 +1,90 @@
...
+        so.make_along_step = UniformAlongStepFactory();
+
+        so.max_num_tracks = 1024 * 16;
+        so.initializer_capacity = 1024 * 128 * 4;
+        so.secondary_stack_factor = 3.0;
+        so.ignore_processes = {"CoulombScat"};  // Ignored processes list
+
+        // Use Celeritas "hit processor" to call back to Geant4 SDs.
+        so.sd.enabled = true;
+
```
sethrj: This is no longer necessary

example/accel/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+project(offload-template)
+
+# Export/set Celeritas_ROOT environment variable pointing to its install dir
+# Defaults to celeritas/install
+set(ENV{Celeritas_ROOT} "../../../../install")
+
+#-----------------------------------------------------------------------------#
+# Find packages
+find_package(Geant4 REQUIRED)
+find_package(Celeritas 0.5 REQUIRED)
```
sethrj: Since Celeritas points to Geant4 but not vice versa, switch these (otherwise the user will have to set both paths).

example/accel/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# Minimal Geant4 application with Celeritas offloading capabilities
+#-----------------------------------------------------------------------------#
+cmake_minimum_required(VERSION 3.10)
+cmake_policy(VERSION 3.12...3.30)
+project(offload-template)
+
+# Export/set Celeritas_ROOT environment variable pointing to its install dir
+# Defaults to celeritas/install
+set(ENV{Celeritas_ROOT} "../../../../install")
+
```
sethrj: It might be good to delete this custom bit of code (and this probably shouldn't be set through the environment anyway). If you want to make it user friendly, you could drop the REQUIRED on 17/18 and then print a helpful message (pointing to `Celeritas_ROOT` or `CMAKE_INSTALL_PREFIX`).

example/accel/offload-template/README.md
```diff
@@ -0,0 +1,48 @@
+Minimal Geant4-Celeritas offloading template
+============================================
```
sethrj: Can you make this file an RST and include it using
```rst
.. include:: ../../example/offload-template/README.rst
```
in the `doc/example/geant4.rst` ?

example/accel/offload-template/src/ActionInitialization.cc
```diff
@@ -0,0 +1,34 @@
...
+ */
+void ActionInitialization::Build() const
+{
+    CelerSimpleOffload().Build(
+        &CelerSetupOptions(), &CelerSharedParams(), &CelerLocalTransporter());
+
+    this->SetUserAction(new RunAction());
+    this->SetUserAction(new EventAction());
+    this->SetUserAction(new TrackingAction());
+    this->SetUserAction(new PrimaryGeneratorAction());
```
sethrj: I think instead of a user tracking action we should focus on the 11+ use case and use the tracking manager, since that's what the Geant4 folks prefer (and 10.7 is quite old now).

sethrj (commented):
Let's move the example up a directory, make sure it shows up in the manual, *and* add it to `scripts/ci/test-examples.sh`

whokion:
A couple of general comments:

1. Use G4RunManagerFactory instead of G4RunManager (also  use G4PhysListFactory if there is no custom physics list)
2. Use G4VTrackingManager (i.e, add some like OffloadingTrackingManager to implement an example of those pure/virtual methods - see an example at ​[CMSEmStandardPhysicsTrackingManager](https://cmssdt.cern.ch/lxr/source/SimG4Core/PhysicsLists/src/CMSEmStandardPhysicsTrackingManager.cc))

example/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+  src/EventAction.cc
+  src/PrimaryGeneratorAction.cc
+  src/RunAction.cc
+  src/SensitiveDetector.cc
+)
+
+set(_includes
+  ${CMAKE_CURRENT_SOURCE_DIR}/src
+  ${Geant4_INCLUDE_DIR}
+  ${Celeritas_ROOT}/include
```
whokion:
```suggestion
file(GLOB sources ${PROJECT_SOURCE_DIR}/src/*.cc)
file(GLOB headers ${PROJECT_SOURCE_DIR}/include/*.hh)
```
so that they do not need to add an additional list of codes in the cmake file.

example/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+#-----------------------------------------------------------------------------#
+# Minimal Geant4 application with Celeritas offloading capabilities
+#-----------------------------------------------------------------------------#
+cmake_minimum_required(VERSION 3.10)
+cmake_policy(VERSION 3.12...3.30)
+project(offload-template)
+
+#-----------------------------------------------------------------------------#
+# Find packages
+find_package(Celeritas 0.5)
```
whokion, line 18: Is the version 0.5 the minimum requirement?

example/offload-template/README.rst
```diff
@@ -0,0 +1,65 @@
...
+  - :code:`CELERITAS_USE_Geant4=ON`
+
+Build and run
+-------------
+
+.. code-block:: sh
+
+   $ mkdir build
+   $ cd build
+   $ cmake ..
```
whokion, line 22: This assumes that Geant4_DIR is known.  Is it always the case?

example/offload-template/main.cc
```diff
@@ -0,0 +1,43 @@
...
+{
+    if (argc != 1)
+    {
+        // Print help message
+        std::cout << "Usage: " << argv[0] << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    // Construct run manager
+    G4RunManager run_manager;
```
whokion:
```suggestion
      auto runManager = G4RunManagerFactory::CreateRunManager(G4RunManagerType::Default);
```

example/offload-template/src/ActionInitialization.cc
```diff
@@ -0,0 +1,46 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Set up all user actions and Celeritas offload interface.
+ */
+void ActionInitialization::Build() const
+{
+    // Construct Celeritas offloading interface
+    CelerSimpleOffload().Build(
+        &CelerSetupOptions(), &CelerSharedParams(), &CelerLocalTransporter());
```
whokion: Should this be in BuildForMaster  (assuming that the default run manager (tasking run manager) is used)?

example/offload-template/src/DetectorConstruction.cc
```diff
@@ -0,0 +1,58 @@
...
+    auto const nist = G4NistManager::Instance();
+    auto const world_material = nist->FindOrBuildMaterial("G4_Pb");
+    double const world_size = 1 * m;
+    auto world_box = new G4Box("world_box", world_size, world_size, world_size);
+    auto world_lv = new G4LogicalVolume(world_box, world_material, "world_lv");
+    auto world_pv = new G4PVPlacement(
+        nullptr, G4ThreeVector(), world_lv, "world_pv", nullptr, false, 0);
+
+    return world_pv;
+}
```
whokion: This may be too specific as this example serves a template - I would rather add 1) a hook to pick up a default gdml with an option to set a user provided gdml or 2) a simple, but generally configurable detector setup (size, materials, etc. despite that this requires additional UI input if users want to modify the geometry setup, so less straight forward).

whokion (commented):
A good start, but may need some refinement to make the example more general.

example/offload-template/CMakeLists.txt
sethrj: No, globbing source is bad practice. The includes here are just directories, and in fact the `Geant4_INCLUDE_DIR` and `Celeritas_ROOT` lines should be removed since they're propagated by the CMake library targets.

example/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+set(_libs
+  ${Geant4_LIBRARIES}
+  Celeritas::celeritas
+  Celeritas::accel
+)
+
+#-----------------------------------------------------------------------------#
+# Add executable
+add_executable(run-offload main.cc ${_sources})
+target_include_directories(run-offload PRIVATE $<BUILD_INTERFACE:${_includes}>)
```
sethrj: Make it easier on your users:
```suggestion
add_executable(run-offload
  main.cc
  src/ActionInitialization.cc
  src/Celeritas.cc
  src/DetectorConstruction.cc
  src/EventAction.cc
  src/PrimaryGeneratorAction.cc
  src/RunAction.cc
  src/SensitiveDetector.cc
)
target_include_directories(run-offload PRIVATE $<BUILD_INTERFACE:target_include_directories(run-offload PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
```

example/offload-template/CMakeLists.txt
stognini: Yes. I avoided globbing. Good catch on the include dirs, remnants from a poor copy-paste, since the `find_package` should already provide those.

example/offload-template/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+# Use Celeritas CUDA RDC utility function to correctly link libraries
+include(CudaRdcUtils)
+cuda_rdc_target_link_libraries(run-offload ${_libs})
```
sethrj: Likewise I'd also just put the libs in here directly rather than defining an extra variable.

example/offload-template/src/ActionInitialization.cc
stognini: Not really, because it is not an MT application. But good point, I may have to either make it an MT example, or add a few notes throughout it about the MT case.

example/offload-template/src/DetectorConstruction.cc
stognini: I'd prefer to keep it as the simplest geometry possible, just so the code compiles and shows that Celeritas offload happens. It is a skeleton just to highlight where the Celeritas' offload code should be placed. If you depend on GDML, then you depend on G4 compiled with GDML=ON, you need a geometry file from somewhere, and everything becomes more convoluted than just `cmake; make; ./run` and see the CELER_LOG messages.

example/offload-template/README.rst
stognini: I could add an extra note. But yes, I did assume that the reader already had an installed Geant4. Either through a package manager, or a sourced local install, so their environment is correctly set.

whokion:
>     2. Use G4VTrackingManager (i.e, add some like OffloadingTrackingManager to implement an example of those pure/virtual methods - see an example at ​[CMSEmStandardPhysicsTrackingManager](https://cmssdt.cern.ch/lxr/source/SimG4Core/PhysicsLists/src/CMSEmStandardPhysicsTrackingManager.cc))

Sorry that I did not know that [TrackingManagerOffload](https://github.com/celeritas-project/celeritas/blob/develop/src/accel/TrackingManagerOffload.cc) already implemented and you already used the interface.  Your approach is good.

example/offload-template/CMakeLists.txt
stognini: I initially added the extra variables for clarity, but this actually looks better/cleaner...

example/accel/offload-template/CMakeLists.txt
stognini: Thanks. I wasn't sure how to make this easy/automatic, and the message solution is much better indeed.

example/offload-template/CMakeLists.txt
stognini: I think the current offload interface became available on 0.5, but that requirement is from my memory. I can double check it.

example/offload-template/main.cc
sethrj: Would this make it MT if Geant4 is built with threading?

stognini: I actually made this a `G4RunManagerType::Serial` to run a single-thread execution, and just added a comment on the MT case.

Edit: My reasoning to keep it single-threaded is so that we don't have to also require an install with `GEANT4_BUILD_MULTITHREADED=ON` as a condition just to test a tiny offload example.

example/offload-template/CMakeLists.txt
sethrj: There's no reason not to fix it at the current "minimum available" version: I think this is fine.

example/offload-template/src/ActionInitialization.cc
sethrj: It probably should support MT? Most new Geant4 apps should.

stognini: Yeah, I thought of that. But then we have to require a Geant4 install with `GEANT4_BUILD_MULTITHREADED=ON` for a tiny example. Fairly, that would be standard for most cases, but seemed simpler to run a single-threaded one, so any compilation works, and explain how MT should be handled in a comment.

example/offload-template/main.cc
whokion: GEANT4_BUILD_MULTITHREADED=ON is now the default (so, you do not need explicitly to set it), which then is able to run both sequential and tasking (default tasking option is the traditional old MT).  So it will support all cases.

example/offload-template/main.cc
```diff
@@ -0,0 +1,45 @@
...
+    if (argc != 1)
+    {
+        // Print help message
+        std::cout << "Usage: " << argv[0] << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    std::unique_ptr<G4RunManager> run_manager;
+    run_manager.reset(
+        G4RunManagerFactory::CreateRunManager(G4RunManagerType::Serial));
```
drbenmorgan:
```suggestion
    run_manager.reset(
        G4RunManagerFactory::CreateRunManager(G4RunManagerType::Default));
```

This will make things work irrespective of the Geant4 install having MT support or not.

example/offload-template/src/ActionInitialization.cc
drbenmorgan: The `BuildForMaster` member function can be supplied always. If Geant4 is _not_ MT enabled, it'll never be called, but by being there both cases can be supported.

example/offload-template/src/ActionInitialization.cc
```diff
@@ -0,0 +1,54 @@
...
+    // Add Celeritas tracking manager to electrons, positrons, and gammas.
+    // celeritas::TrackingManagerOffload automatically assigns available
+    // physics processes to the selected particles and offloads them to
+    // Celeritas by updating their G4TrackStatus to fStopAndKill in Geant4 and
+    // creating a new track in Celeritas.
+    auto* celer_tracking = new celeritas::TrackingManagerOffload(
+        &CelerSharedParams(), &CelerLocalTransporter());
+    G4Electron::Definition()->SetTrackingManager(celer_tracking);
+    G4Positron::Definition()->SetTrackingManager(celer_tracking);
+    G4Gamma::Definition()->SetTrackingManager(celer_tracking);
```
drbenmorgan, line 59: This should _probably_ go into a specialised PhysicsList/Constructor as part of the `ConstructProcess` member function, for example as in:

https://github.com/celeritas-project/celeritas/blob/dafd27f4bbbe862b22f8646dc409e7a7f7446cda/example/accel/trackingmanager-offload.cc#L134-L151

I'm not sure it's possible to do the setup in the action initialization, but maybe worth asking Mihaly (@mnovak42) on this. This might be more something to do in core Celeritas so one could wrap any EM constructor, maybe using CRTP.

example/offload-template/main.cc
stognini: Yes. I just changed the code to be MT, so it is `Default` now.

example/offload-template/src/ActionInitialization.cc
stognini: Yes. I internally debated if I should have added a custom physics list or keep it as simple as possible, like now, with only a `new FTFP_BERT()` call in `main`. The current tracking manager assignment in action initialization class works nicely, and my only reasoning until now to keep it that way is because I kept focusing on "how little is needed to show an offload case", so I reduced the number of classes as much as possible.

Happy to add a custom physics list if people think that this will add value while not adding confusion to new users.

drbenmorgan: If it works o.k., that's fine!

example/offload-template/src/DetectorConstruction.cc
```diff
@@ -0,0 +1,58 @@
...
+G4VPhysicalVolume* DetectorConstruction::Construct()
+{
+    // Construct single material world volume box
+    auto const nist = G4NistManager::Instance();
+    auto const world_material = nist->FindOrBuildMaterial("G4_Pb");
+    double const world_size = 1 * m;
+    auto world_box = new G4Box("world_box", world_size, world_size, world_size);
+    auto world_lv = new G4LogicalVolume(world_box, world_material, "world_lv");
+    auto world_pv = new G4PVPlacement(
+        nullptr, G4ThreeVector(), world_lv, "world_pv", nullptr, false, 0);
```
sethrj: using `auto const*` or `auto*` makes the types more understandable.

example/offload-template/src/DetectorConstruction.cc
```diff
@@ -0,0 +1,58 @@
...
+ * Initialize sensitive detectors.
+ *
+ * Every volume that needs to collect data from Celeritas *must* be defined as
+ * a sensitive detector.
+ *
+ * \sa \c SensitiveDetector::ProcessHits from this example.
+ */
+void DetectorConstruction::ConstructSDandField()
+{
+    auto world_sd = new SensitiveDetector("world_sd");
```
sethrj: same here, but since it's the output `new` of course it's a pointer.

scripts/ci/test-examples.sh
```diff
@@ -39,8 +39,19 @@ cd "${CELER_SOURCE_DIR}/example/minimal"
...
 
-# Run Geant4 app example(s)
+# Run Geant4 app examples
 if [ -z "${CELER_DISABLE_ACCEL_EXAMPLES}" ]; then
+  if [ "$(echo $CELER_GEANT4_VERSION | awk -F. '{print $1*100 + $2}')" \
+      -ge 1100 ]; then
+    # Run offload-template only on Geant4 v11
+    cd "${CELER_SOURCE_DIR}/example/offload-template"
+    build_local
+    export CELER_DISABLE_PARALLEL=1
```
sethrj: The CI doesn't use MPI so this shouldn't be needed. If you build with MPI then just export it in your terminal before running this.

sethrj (approved):
Looks good, but please take a look at the couple remaining suggestions.

example/offload-template/src/ActionInitialization.cc
mnovak42: As we introduced `G4VTrackingManager` for allowing to implement specialised, external stepping/tracking per particle types (originally for `G4HepEm`), they can be attached to a `G4ParticleDefinition`. As all our particle definitions are implemented as singletons, unfortunately one can get them and attach a tracking manager to them anywhere. However, in general that should be done before initialising the physics (i.e. before calling `PreparePhysicsTable` and `BuildPhysicsTable`) so it's a good practice to attach the tracking manager inside a physics list. Everything might work well even if you do that at other places (especially that the physics tables are built rather late), i.e. not in the physics list, but this will guarantee.  

Moreover, when a tracking manager is attached to a particle then any processes attached to that particle will be ignored (unless you use them in your tracking manager): as the tracking manager will take over the track with that particle immediately when popped from the stack. So it might be confusing having several processes attached to a given particle in the physics list while they will be just all ignored when a tracking manager is also attached. So usually we make a switch in the physics list such that either we attach the processes or the tracking manager. However, having both won't hurt anything (beyond clarity) and there are some special cases when you actually really what to have both of them. 

Last but not least, you do not need to make a complete new physics list. We usually use modular physics list in `Geant4` so you can make only one custom e.g. `CeleritasEmPhysics` EM physics constructor in which you implement that switch of attaching either `G4VProcesses` or a `G4VTrackingManager` to a particle definition. Then after making any of the `Geant4` reference physics lists like `FTFP_BERT` their `G4VModularPhysicsList::ReplacePhysics(G4VPhysicsConstructor*)` can be used to replace the original EM physics module by your custom one.   

So after this long story (that I added only for clarity), everything is possible if that works and fine for you :smile:

sethrj:
Hey @drbenmorgan , I've been trying to understand some weird behavior in the multiple cmake package scripts we use and thought you might know... https://github.com/celeritas-project/celeritas/actions/runs/12812027794 shows @stognini 's  example failing for Geant4 11.1 and 11.2 . 

I finally realized we'd already "solved" this issue before: https://github.com/celeritas-project/celeritas/blob/eb9b0d721354503f12d3fbf18be0171225dcfdb8/cmake/CeleritasConfig.cmake.in#L70--L72 is due to https://github.com/celeritas-project/celeritas/pull/1152 , which notes "issues with the new geant4 cmake variable cache break[ing] xercesc importing for 11.1–11.2". For some reason though our previous fix fails when Celeritas isn't marked `REQUIRED`: see https://github.com/celeritas-project/celeritas/pull/1573/commits/e4eed68522b47991067427f7cbcb7f3737919ddb. (Maybe it invalidates the `find_dependency` cache?)

Do you have any ideas how to better fix this? Thanks!

drbenmorgan:
Argh, this _looks_ like the same issue as in https://github.com/apt-sim/AdePT/pull/267, the ~~hack~~ fix there was to explicitly find Xerces before anything else. I'll have to go back to that and the reproducer to try and properly track this down upstream...

sethrj:
Yeah, we added that and it worked as long as celeritas and Geant were both required 😅

drbenmorgan:
O.k., no better ideas from my side at the moment - will try and take a look on the VecGeom etc side next week...

sethrj:
Aha, I  see in the vecgeom config file:
```
if(NOT DEFINED XercesC_INCLUDE_DIR)
  set(XercesC_INCLUDE_DIR "/opt/spack/opt/spack/sequoia/xerces-c/3.2.5/3zdtn2m/include")
endif()
```
... which I introduced in VecGeom 52f767d29f, which maybe is conflicting with something in Geant4 and/or XercesC...

pcanal:
The discussion at https://github.com/root-project/root/issues/16374#issue-2506427559 might or might not be relevant here (if find_package is used somewhere that is not the top level CMakeLists.txt).

sethrj:
Sorry @stognini i forgot to unmask as draft!

example/offload-template/src/ActionInitialization.cc
```diff
@@ -0,0 +1,65 @@
...
+    CelerSimpleOffload().Build(
+        &CelerSetupOptions(), &CelerSharedParams(), &CelerLocalTransporter());
+
+    // Add Celeritas tracking manager to electrons, positrons, and gammas.
+    // celeritas::TrackingManagerOffload automatically assigns available
+    // physics processes to the selected particles and offloads them to
+    // Celeritas by updating their G4TrackStatus to fStopAndKill in Geant4 and
+    // creating a new track in Celeritas.
+    auto* celer_tracking = new celeritas::TrackingManagerOffload(
+        &CelerSharedParams(), &CelerLocalTransporter());
```
sethrj, line 56: @drbenmorgan , I'm just now noticing (and this is copied from your original example) that we're `new`ing a pointer, passing the raw pointer to several classes, and moving on without explicitly managing ownership. I assume this is a leak, and if so, how do we fix it?

PR #1574, opened 2025-01-13
Extend Coulomb scattering to support additional particle types

amandalund:
This extends the single and multiple Coulomb scattering models to support muons and hadrons. I hope I've identified all the places where the treatment is different for these particles, although if we made simplifications anywhere assuming the particle was an electron without documenting it (like omitting abs(charge)) I won't have caught it. The changes to the Coulomb and Wentzel test results are just from using the imported electron mass instead of the constant to calculate the `kin_factor` in the `WentzelHelper`.


src/celeritas/phys/PhysicsData.hh
```diff
@@ -242,9 +242,14 @@ struct PhysicsParamsScalars
...
     real_type lambda_limit{};  //!< lambda limit
-    real_type range_factor{};  //!< range factor for e-/e+ (0.2 for muon/h)
+    real_type range_factor{};  //!< range factor for e-/e+
+    real_type muhad_range_factor{};  //!< range factor for muon/h
     real_type safety_factor{};  //!< safety factor
+    bool displaced{};  //!< lateral displacement for e-/e+
+    bool muhad_displaced{};  //!< lateral displacement for muon/h
     MscStepLimitAlgorithm step_limit_algorithm{MscStepLimitAlgorithm::size_};
+    MscStepLimitAlgorithm muhad_step_limit_algorithm{
+        MscStepLimitAlgorithm::size_};
```
amandalund: I might refactor this a bit and add a separate struct for the particle-dependent scalars (since we'll also need muon/hadron `min_range`, `max_step_over_range`, `lowest_electron_energy`) and an accessor in `PhysicsTrackView` to return either the e-/e+ or mu/had scalars depending on the particle ID.

src/celeritas/em/data/UrbanMscData.hh
```diff
@@ -38,6 +38,12 @@ struct UrbanMscParameters
     Energy low_energy_limit{0};
     Energy high_energy_limit{0};
 
+    //! Special value indicating MSC does not apply to a particle type
+    static CELER_CONSTEXPR_FUNCTION size_type inapplicable()
+    {
+        return size_type(-1);
+    }
+
```
sethrj: Instead of a `size_type` and magic value, define (under `UrbanMscParMatData`) a new type alias `UrbanParMatId = OpaqueId<UrbanMscParMatData>;`.

src/celeritas/em/data/UrbanMscData.hh
```diff
@@ -143,6 +150,10 @@ struct UrbanMscData
     UrbanMscParameters params;
     //! Material-dependent data
     MaterialItems<UrbanMscMaterialData> material_data;
+    //! Map from particle ID to index in particle and material-dependent data
+    ParticleItems<size_type> pid_to_pmdata;
```
sethrj: Use the new opaque ID here.

src/celeritas/em/params/UrbanMscParams.cc
```diff
@@ -66,21 +66,50 @@ UrbanMscParams::UrbanMscParams(ParticleParams const& particles,
...
-    static Array<double, 2> const b_coeff{{2.0 / 3, 1.0 / 2}};
+    // Number of different particle categories in the particle and
+    // material-dependent parameter data: electrons and positrons always, and
+    // muons/hadrons when present
+    CELER_ASSERT(helper.particle_ids().size() >= 2);
+    size_type num_particles = min<size_type>(helper.particle_ids().size(), 3);
+
+    // Map from particle ID to index in particle and material-dependent data
+    std::vector<size_type> pid_to_pmdata(particles.size(),
+                                         UrbanMscParameters::inapplicable());
```
sethrj:
```suggestion
    std::vector< UrbanParMatId > pid_to_pmdata(particles.size());
```

src/celeritas/em/params/UrbanMscParams.cc
```diff
@@ -66,21 +66,50 @@ UrbanMscParams::UrbanMscParams(ParticleParams const& particles,
...
+            pid_to_pmdata[par_id.unchecked_get()] = 0;
+        }
+        else if (par_id == host_data.ids.positron)
+        {
+            pid_to_pmdata[par_id.unchecked_get()] = 1;
+        }
+        else
+        {
+            // Muons and hadrons
+            pid_to_pmdata[par_id.unchecked_get()] = 2;
```
sethrj: These might be good to have as an enum?

src/celeritas/em/params/UrbanMscParams.cc
```diff
@@ -102,8 +129,7 @@ UrbanMscParams::UrbanMscParams(ParticleParams const& particles,
 
             // Compute the maximum distance that particles can travel
             // (different for electrons, hadrons)
-            if (par_ids[p] == host_data.ids.electron
-                || par_ids[p] == host_data.ids.positron)
+            if (p < 2)
```
sethrj:
```suggestion
            if (p != UrbanParticleType::mu_had)
```

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -214,16 +213,29 @@ CELER_FUNCTION real_type WentzelHelper::calc_screening_coefficient(
...
+            // Electrons and positrons
+            real_type tau = value_as<Energy>(particle.energy())
+                            / value_as<Mass>(particle.mass());
+            factor = std::sqrt(tau / (tau + sq_cbrt_z));
+        }
+        else
+        {
+            // Muons and hadrons
+            factor = ipow<2>(value_as<Charge>(particle.charge()));
+            z_factor += std::exp(-ipow<2>(target_z_.get()) * real_type(0.001));
```
sethrj, line 235: Any idea where `0.001` comes from? Looks more like a "tolerance" or a rejection sampling window rather than an actual cross section correction...

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -267,16 +279,12 @@ CELER_CONSTEXPR_FUNCTION auto WentzelHelper::screen_r_sq_elec() -> MomentumSq
...
-    real_type constexpr twopi_mrsq
-        = 2 * constants::pi
-          * ipow<2>(native_value_to<Mass>(constants::electron_mass).value()
-                    * constants::r_electron);
-
-    return twopi_mrsq * target_z_.get()
-           * ipow<2>(value_as<Charge>(particle.charge()))
+    return 2 * constants::pi
+           * ipow<2>(value_as<Mass>(electron_mass) * constants::r_electron)
+           * target_z_.get() * ipow<2>(value_as<Charge>(particle.charge()))
```
sethrj: or (saving a couple runtime operations):
```suggestion
    constexpr Constant twopirsq = 2 * constants::pi
                                  * ipow<2>(constants::r_electron);
    return twopirsq * ipow<2>(value_as<Mass>(electron_mass)
                              * value_as<Charge>(particle.charge()))
           * target_z_.get();
```

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -288,27 +296,45 @@ WentzelHelper::calc_kin_factor(ParticleTrackView const& particle) const
  * This calculates the cosine of the maximum polar angle that the incident
  * particle can scatter off of the target's electrons.
  */
-CELER_FUNCTION real_type WentzelHelper::calc_cos_thetamax_electron(
-    ParticleTrackView const& particle, CoulombIds const& ids, Energy cutoff) const
+CELER_FUNCTION real_type
+WentzelHelper::calc_cos_thetamax_electron(ParticleTrackView const& particle,
```
sethrj, line 307: I think this function can be static.

src/celeritas/em/xs/WentzelHelper.hh
```diff
@@ -288,27 +296,45 @@ WentzelHelper::calc_kin_factor(ParticleTrackView const& particle) const
...
+        // Muons and hadrons
+        real_type mass_ratio = value_as<Mass>(electron_mass) / mass;
+        real_type tau = inc_energy / mass;
+        real_type max_energy
+            = 2 * value_as<Mass>(electron_mass) * tau * (tau + 2)
+              / (1 + 2 * mass_ratio * (tau + 1) + ipow<2>(mass_ratio));
+        return 1
+               - min(value_as<Energy>(cutoff), max_energy)
+                     * value_as<Mass>(electron_mass)
+                     / value_as<MomentumSq>(particle.momentum_sq());
```
sethrj: I guess since the other particles are heavier than electrons, we know this last quantity will be small so the result is in `[0,1]`.

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -60,14 +60,14 @@ struct GeantMuonPhysicsOptions
     bool bremsstrahlung{false};
     //! Enable muon single Coulomb scattering
     bool coulomb{false};
-    //! Enable muon multiple Coulomb scattering
-    bool msc{false};
+    //! Enable multiple coulomb scattering and select a model
+    MscModelSelection msc{MscModelSelection::none};
```
sethrj, line 64: Can you add backward compatibility code for the JSON I/O to handle the old true/false?

sethrj (commented):
Very nice! Quite a lot of work.

src/celeritas/em/xs/WentzelHelper.hh
hhollenb: This is in the calculation of the screening parameter $A_s$, so it is in an actual cross section correction. In Geant4, it appears in `G4WentzelOKandVIxSection::InitialiseA()` when building tables of `ScreenRSquare` for different atomic numbers. The screening parameter depends upon the target's charge, and `z_factor` appears to be a correction for the target's charge. My naive guess is that it's a finite mass correction, where electrons / positrons are taken to be light and don't penetrate the orbital electron shell deeply, while muons and hadrons are heavy and see more of the target's charge?

hhollenb: The closest I could find in the physics reference manual is that the screening length $a_I$ is different for incident particles with $z = 1$ and $z \geq 2$. This doesn't remotely follow the exponential form and I can't find where they use the different screening lengths in the code, but at least there's some precedence for having a correction to the screening length that's different for $Z = 1$ and $Z \geq 2$?

sethrj: Yeah, it's *because* it's an actual correction that I'm so confused: we're giving a physical quantity a value that has no physical meaning (the only time 0.001 shows up in physics is when converting units 🙃 )

whokion (commented):
This is a very nice extension and an archeological work to add MSC support for muon and other charged hadron!  As long as Urban MSC concerns, it all looks good.  On the other hand, I do not really see any application case for Urban MSC (i.e., below 100 MeV) for muon and charged hadron at least for HEP applications - probably same for @stognini's $\mu$-catalyzed fusion simulation, i.e., WentzelVI and Coulomb scattering may be sufficient for these additions.  Nevertheless, It may be still good to have this general extension for Urban too as the new code introduces a minimal number of extra branches and additional code.  Anyway, I will do some high level verification for muon and other charged hadrons (i.e., comparisons with respect to Geant4) once this MR is merged.  Thanks for the nice extension!

amandalund:
Thanks @whokion! Yes, lots of digging :). I agree about the use cases for Urban MSC... at least though it will let us support MSC for muons until the Wentzel VI model is completed and validated. Verification for muons and charged hadrons would be greatly appreciated!

src/celeritas/em/xs/WentzelHelper.hh
amandalund: Good point, I'll add an assertion.

src/celeritas/em/xs/WentzelHelper.hh
amandalund: 🪄

amandalund:
Thanks for the feedback @sethrj! I think I've addressed all of it now.

src/celeritas/em/msc/detail/UrbanMscHelper.hh
```diff
@@ -181,6 +171,39 @@ UrbanMscHelper::calc_end_energy(real_type step) const -> Energy
     }
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Scaled cross section data for this particle+material.
+ */
+CELER_FUNCTION XsGridData const& UrbanMscHelper::xs() const
+{
+    size_type num_particles = shared_.xs.size() / shared_.material_data.size();
```
sethrj: It might be good to store `ParticleId::size_type num_particles` explicitly for error checking and to avoid slow integer division.

src/celeritas/em/xs/WentzelHelper.hh
sethrj: Maybe it's rounding $`2/\alpha^2 \sim 0.00106`$ 🙄

sethrj (approved):
Looks great! It would be nice to understand better the source of some of these corrections 😅 I spent a few minutes looking at some of the relevant literature but nothing jumped out at me 😕

src/celeritas/em/xs/WentzelHelper.hh
amandalund: I didn't see anything in the literature either, but the Geant4 forum helped: https://geant4-forum.web.cern.ch/t/screening-parameter-for-coulomb-scattering-in-g4wentzel-model/6025

It's an empirical correction to better match the data from https://doi.org/10.1016/j.nimb.2006.05.006. I'll add a note in the documentation.

PR #1575, opened 2025-01-13
Propagate code citations into user manual

sethrj:
This (re)defines doxygen commands `\cite`, `\citet`, and `\citep` for adding citations as a combination of Zotero keys and DOI URLs (or just other URLs). The procedure to add references is documented in the `coding` guidelines.


doc/development/coding.rst
```diff
@@ -28,7 +28,14 @@ semi-automatically into this user manual via the Breathe tool integrated
 into the Celeritas build system (see :ref:`dependencies`). High-level classes
 should describe the functionality of the class in a way understandable to both
 power users and developers, and such classes should be included in the
-:ref:`api` section.
+:ref:`api` section. Citations to the Celeritas Zotero database can be
+referenced inline code using the special ``\citep{whatever-2025,
```
amandalund: Should we include the instructions on how to add references to the zotero group and update the bib file?
```suggestion
referenced inline code using the special ``\citep{author-keyword-year,
```

src/celeritas/em/data/WentzelOKVIData.hh
```diff
@@ -50,12 +50,12 @@ struct CoulombParameters
...
  *
  * The matrix of coefficients used to approximate the ratio of the Mott to
- * Rutherford cross sections was developed in T. Lijian, H. Quing and L.
- * Zhengming, Radiat. Phys. Chem. 45 (1995), 235-245. Using the same procedure
- * as in Lijian, the coefficients were extended in M.J. Boschini et al, Radiat.
- * Phys. Chem. 90 (2013), 39-66 (doi.org/10.1016/j.radphyschem.2013.04.020) to
- * include positrons and the interaction of electrons and positrons with higher
- * Z nuclei (1 <= Z <= 118).
+ * Rutherford cross sections was developed in \citet{lijian-mott-1995,
+ * https://doi.org/10.1016/0969-806X(94)00063-8}. Using the same procedure as
```
amandalund, line 54: Are most of these not in the bib file yet? I get some warnings about bibtex keys not found.

doc/backmatter/references.rst
```diff
@@ -14,6 +14,5 @@ References
 
    .. _sphinxbib: https://pypi.org/project/sphinxcontrib-bibtex/
 
-.. bibliography:: ../_static/references.bib
-   :all:
```
amandalund, line 18: Now I'm only seeing a single entry in the references (the PRM), I guess because it's the only one that's currently both in the bib file and being cited? Will we need to go through and update the other citations in the docs if we want them to appear here?

doc/development/coding.rst
sethrj: Absolutely; and it turns out I neglected to push that commit 🙃

src/celeritas/em/data/WentzelOKVIData.hh
sethrj: Yeah, this is blocked on some changes in the main docs that I put into #1562 from which this PR was split out.

doc/backmatter/references.rst
sethrj: Yes, or we could manually use the [:cite:empty:](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#role-cite-empty) directive to force inclusion of some.

What I'd *like* to do is export a bibliography set if and only if the zotero entries have a "citation key": i.e., we want to cite it in a work. And then to automate that export process.

sethrj: Added a feature request, https://github.com/retorquere/zotero-better-bibtex/issues/3137 . I suppose we could also add a "collection" in zotero that could serve a similar purpose.

amandalund (approved):
Nice, looks good to me!

PR #1569, opened 2025-01-08
Compose caching functionality of LazySenseCalculator

esseivaju:
As discussed in #1565, have a simple `LazySenseCalculator` and a front-end with caching functionality. We could probably rework the original sense calculator as an `EagerSenseCalculator`.

PR #1576, opened 2025-01-14
Use lazy sense calculator

esseivaju:
The lazy sense calculators require potentially recalculating the same sense multiple times. However, they do not use global memory so they have the potential to be faster, especially once we move to infix evaluation and short-circuiting. 

I am splitting this from #1565 to test whether we see any performance improvement without short-circuiting.


esseivaju:
~~Not sure why only that one unit test would fail~~ 🤔

esseivaju:
Timings are essentially the same as in https://github.com/celeritas-project/celeritas/pull/1569#issuecomment-2581632960, `complex_intersect` doesn't make a difference.

esseivaju:
@sethrj Do you want to merge this, or should we do everything as part of the infix eval PR?

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -546,22 +543,30 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
...
+        real_type const distance = state.temp_next.distance[isect];
         // Face being crossed in this ordered intersection
         FaceId face = state.temp_next.face[isect];
-        // Flip the sense of the face being crossed
-        calc_senses.flip_sense(face);
+        Real3 pos = state.pos;
+
+        // cross the internal surface, effectively flipping
+        // internal senses
+        axpy(distance + bump_dist, state.dir, &pos);
```
sethrj: We *do not* want to bump the surface crossings if ever we can (and I don't think we need to here). Since at each intersection we're testing "on that face" (on the other side), I think we should calculate the sense from the previous point (using the previous on_face if we're jumping across the inside of a two-intersection quadric), then flip that sense as we cross it.

esseivaju: I think we need to update the position, we're testing that intersection, but senses for previous intersections still need to be flipped. Without dynamically allocated memory, we have to update to position so that previous senses are calculated correctly

sethrj: What about this:

At a point `p`, potentially "on a face" with a face ID `f` and sense `s`:
- Let `s'` be sense wrt the next face `f'` from `p` on `f` (no dynamic storage needed, just calling the lazy sense evaluator)
- Update `p`: `axpy` with the distance to the next face
- Update the current face and flip the sense: `f←f'` and `s←~s'`
- Now the new logic state can be evaluated with the lazy evaluator using `p` and on `f` with `s`.

esseivaju: mmh right, if we need to avoid bumping then a combination of updating on_face and the position should work. Is the reason you avoid bumping numerical instability?

sethrj: Inaccuracy, really: you can jump across corners or nearby surfaces very easily. It turns an "exact" surface crossing into an inexact one.

esseivaju: I see, current implementation should be better

sethrj: I made an additional change that correctly implements the algorithm I described above. The version that flips the sense from the *original* position fails if the track moves through a surface twice before exiting.

Maybe it would be good to add little unit test where we have a union of a sphere and a box so that we can test tracking through the double interior crossing without exiting the geometry.

esseivaju: the distance is from the original position though? So we'd need to substract the previous intersect distance to get the delta. Also, taking references to the VolumView has caused some issues before where the sense calculator outlives the view.

esseivaju: [The failing test](https://github.com/celeritas-project/celeritas/actions/runs/12798537330/job/35682794658) is why taking references is error-prone. I think the lifetime of the temporary is only extended until the end of the statement, so we end up with a dangling reference as the member. From [cppreference](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary):

> a temporary bound to a reference parameter in a function call exists until the end of the full expression containing that function call: if the function returns a reference, which outlives the full expression, it becomes a dangling reference.

sethrj: Yeah I should've checked the unit test. But on GPU you don't want to be glib about copying data locally since that can increase local memory and register usage.

sethrj: Yes sorry, you're quite right I should've updated the distance from the original position

test/orange/univ/detail/SenseCalculator.test.cc
```diff
@@ -109,26 +109,22 @@ using TestTypes = ::testing::
 
 TYPED_TEST_SUITE(SenseCalculatorTest, TestTypes, );
 
+#if CELERITAS_DEBUG
```
sethrj: Why does `if (CELERITAS_DEBUG` no longer work? Or were you doing this for testing?

test/orange/univ/detail/SenseCalculator.test.cc
```diff
@@ -109,26 +109,22 @@ using TestTypes = ::testing::
...
-        this->make_volume_view(LocalVolumeId{0}),
-        Real3{123, 345, 567},
-        face);
-    if (CELERITAS_DEBUG)
-    {
-        EXPECT_THROW(calc_senses(FaceId{0}), DebugError);
-    }
+    EXPECT_THROW(this->construct_sense_calculator(
+                     this->make_surf_visitor(),
+                     this->make_volume_view(LocalVolumeId{0}),
```
sethrj: Oh yeah if we're taking these as const references perhaps it's better to construct them locally and pass it in as an lvalue?

sethrj (approved):
Thanks @esseivaju !

test/orange/univ/detail/SenseCalculator.test.cc
esseivaju: It still works but that test doesn't do anything in non-debug build so that's why I commented the whole thing out

esseivaju: Using `TEST_IF_CELERITAS_DEBUG` instead

test/orange/univ/detail/SenseCalculator.test.cc
esseivaju: This should be legal since the lifetime of these references is extended to the end of the statement maybe clearer to pass lvalues

PR #1577, opened 2025-01-14
Release v0.5.1

sethrj:
## Pre-merge checklist

- [x] Ensure all CI jobs on the branch to be released (develop or backports/vX.Y) pass. https://github.com/celeritas-project/celeritas/actions/runs/12888835166
- [x] Run performance regression tests on Perlmutter, Frontier, and an additional machine with debug assertions enabled (e.g., Wildstyle). https://github.com/celeritas-project/regression/commit/ca92064c60508ba241959921598d04956b2049f0
- [x] Update documentation with release notes from all pull requests newly included in the release.
- [x] Ensure PDF documentation builds without error.

## Post-merge checklist

- [x] If releasing a backported version branch, cherry-pick this documentation commit into the backport branch.
- [x] Use the [GitHub interface](https://github.com/celeritas-project/celeritas/releases/new) to create a new release with the documentation update that was just added.

## Post-release checklist

- [x] Save the ``tar.gz`` and attach to the release, because the hash changes if the git "describe" function returns a different result for the release tag's hash (e.g., if a collaborative branch on the main repository points to that commit).
- [x] Pull locally (make sure to use the ``--tags`` option) and build PDF user documentation for the release. Ensure breathe is activated (so the API is listed) and that the version is embedded correctly.
- [x] Update the Spack recipe for Celeritas with the new version and sha256 value (either manually or using ``spack checksum``) and submit a [pull request to the Spack project](https://github.com/spack/spack/pull).
- [x] Mark the GitHub [release milestone](https://github.com/celeritas-project/celeritas/milestones) as completed.


sethrj:
Waiting on #1586 and #1578.

sethrj:
Looks like ORANGE is slightly slower on CPU on both platforms? But there shouldn't be any real changes...

esseivaju:
do we have a range of commits to run git bisect on?

amandalund (approved):
> Looks like ORANGE is slightly slower on CPU on both platforms?

How much is slightly?

sethrj:
Figured out the discrepancy; some tweaks Julien and I in the runner script meant celeritas on CPU and through Geant4 got much more track slots, which boosted g4+GPU performance and hit CPU performance. @esseivaju @amandalund  We need to rerun those scaling studies 😅

sethrj:
@amandalund @esseivaju Here's the relative throughput on Perlmutter compared to the v0.5.0 results:
rel-throughput
so that the previous throughput:
throughput-old
now looks like:
throughput-new

... note how the suspicious performance jumps in Geant4 performance correspond to it matching the Celeritas standalone results 😬 This corresponds to me trying to make the track slots consistent between sim/g4 but obviously failing:
Screenshot 2025-01-22 at 04 57 16

sethrj:
@amandalund @esseivaju Fixed comparison wrt v0.5.0 for perlmutter results are here.

Absolute performance:
throughput-with-geant

Releative from 0.5.1 to 0.5.0 with updated slot counts:
rel-throughput

The number of total (integrated across streams) track slots, giving total resource requirements, is:

Arch  | New | Old
--- | --- | ---
cpu     |   1048576.0 | 1048576.0
cpu+g4    |    65536.0 | 1048576.0
gpu    |     1048576.0 | 1048576.0
gpu+g4     |  262144.0 | 65536.0

And for per-stream (related to the performance characteristics) is:

Arch  | New | Old
--- | --- | ---
cpu     |   65536.0 | 65536.0
cpu+g4    |    4096.0 | 65536.0
gpu    |     1048576.0 | 1048576.0
gpu+g4     |  262144.0 | 65536.0

These are compared to the non-screwed-up v0.5.0 results. I think with the CPU runs "merge_events: False" means we have 16 different set of states.

amandalund:
Thanks! So the gist is we're now using more GPU and fewer CPU track slots with celer-g4 because it improved performance? And celer-sim performance and capacities have stayed the same?

With celer-sim CPU (multi-stream), are the capacities listed above the values that are used as problem input (i.e., the values before they are divided by the number of streams in the app) so new cpu and cpu+g4 are using the same total resources?

sethrj:
@amandalund The reduced CPU was more or less accidental because of the different definitions of streams and state sizes and such 🤪 I think we should wait on your and @esseivaju 's upcoming parameter search work to update these.

The updated problem parameters are in https://github.com/celeritas-project/regression/blob/1318c27074c758c1315491fb12980414befba2d2/run-problems.py leading to these inputs:
```
g4: 4096 (which isn't actually used since geant4 is managing the tracks),
cpu+g4: 4096,
cpu: 65536,
gpu+g4: 262144,
gpu: 1048576,
```

I think the updated inputs are still inconsistent. I really want to refactor the input/output so we get total state size, and number of streams.

amandalund:
Yeah, I was trying to figure it out from the script, all the different run modes and state size adjustments there and in the apps make it very confusing 😅... looks like CPU multithreaded celer-sim and celer-g4 are using the same capacities, GPU single stream celer-sim is 4x smaller (integrated) than multi-stream celer-g4?

sethrj:
Maybe I should redo the problem script from scratch 😆

PR #1578, opened 2025-01-14
Require less ancient CUDA architecture and OpenMP implementation

sethrj:
This adds a CMake-based check that OpenMP support conforms to 3.0 and above (spec released in 2008). It also removes support for ancient CUDA cards (older than 2016) and simplifies some of the atomic macros. As part of the CMake check, the "invalid option" now *updates* the CMake Cache rather than suggesting a variable change, so it's likely a reconfigure will pass.


src/corecel/math/Atomics.hh
```diff
@@ -13,18 +20,22 @@
...
 
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ < 600)
+#    error "Celeritas requires CUDA arch 6.0 (P100) or greater"
+#endif
+
+#if defined(_OPENMP) && CELERITAS_OPENMP == CELERITAS_OPENMP_TRACK
+//! Capture the subsequent expression as an OpenMP atomic
+#    define CELER_CAPTURE_IF_OPENMP_TRACK _Pragma("omp atomic capture")
+#else
+#    define CELER_CAPTURE_IF_OPENMP_TRACK if (true)
```
pcanal:
```suggestion
#    define CELER_CAPTURE_IF_OPENMP_TRACK
```
should work too.

scripts/cmake-presets/ci-windows-github.json
```diff
@@ -39,7 +39,8 @@
       "displayName": "fast build with testing and JSON",
       "cacheVariables": {
         "CELERITAS_BUILD_TESTS": {"type": "BOOL", "value": "ON"},
-        "CELERITAS_USE_JSON": {"type": "BOOL", "value": "ON"}
```
pcanal, line 42: Is JSON support intentional removed?

CMakeLists.txt
```diff
@@ -343,7 +343,23 @@ if(CELERITAS_USE_MPI)
 endif()
 
 if(CELERITAS_USE_OpenMP)
-  find_package(OpenMP REQUIRED)
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    # MSVC requires LLVM OpenMP support to provide 3.0 or higher
```
amandalund: I think the `llvm` option is still experimental and only supports a subset of features from newer specs.

scripts/cmake-presets/ci-windows-github.json
sethrj: The configure option is obsolete; it's now always on and this option is ignored.

CMakeLists.txt
sethrj: Yeah. It supports the `atomic capture` keyword in theory but breaks in practice and self-reports as 2.0 anyway. I'll revert this.

amandalund (approved):
Looks good, thanks!

pcanal (approved):

PR #1581, opened 2025-01-15
Import and use muon+hadron EM physics parameters

amandalund:
This pulls out and extends and refactors a bit the import and use of the muon/hadron EM physics parameters from  #1574.
  - Imports mu/had `lowest_energy`, `range_factor`, `step_limit_algorithm`, and e-/e+/mu/had `displaced` flag for MSC
  - Adds mu/had `min_range` and `max_range_over_step` to the physics options (these could be set in Geant4 but can't be imported since there is no way to access them from the `EmExtraParameters` or energy loss processes)
  - Adds a new accessor to `PhysicsTrackView` for particle-dependent scalars
  - Sets the MSC physics options from the imported parameters


src/celeritas/phys/PhysicsParams.hh
```diff
@@ -71,42 +108,67 @@ class ParticleParams;
...
 
-    real_type secondary_stack_factor = 3;
-    bool disable_integral_xs = false;
+    real_type secondary_stack_factor;
+    bool disable_integral_xs;
+    size_type spline_eloss_order;
 
-    size_type spline_eloss_order = 1;
+    //! Default constructor
+    PhysicsParamsOptions()
```
sethrj: Instead of a default constructor for these, what about static helper functions `ParticleOptions::default_muhad()` and `ParticleOptions::default_em()`?

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -643,6 +650,32 @@ PhysicsTrackView::scalars() const
...
+        return params_.scalars.em;
+    }
+    return params_.scalars.muhad;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Whether the particle is an electron or positron.
+ */
+CELER_FORCEINLINE_FUNCTION bool PhysicsTrackView::is_electron() const
```
sethrj: Is there possibly a better name for this? Unless the reader knows it's contrasted with mu-hadron, it looks like it's just for e-.

Since we're using `em` as our physics option input for e+/-, what about `is_em`?

Also, this doesn't account for the case where we have made-up particles like celeriton/celerino. Perhaps `is_em` (`is_light_em`?) should be a particle property flag rather than a hardcoded particle type switch?

sethrj (commented):
Looking good! I hope we can get #1562 in before too many more params/input modifications are made 😅

src/celeritas/phys/PhysicsTrackView.hh
amandalund: I could also keep it as `is_muhad`? Yeah, a flag on the particle data might be better... though I'm not sure how best to flag our mock particles as "em". Would we need to add a field `is_em` to `ParticleParams::ParticleInput`? How bad would it be to just add a method to `ParticleView` that checks that the mass is > 1 MeV? ;)

amandalund:
> I hope we can get https://github.com/celeritas-project/celeritas/pull/1562 in before too many more params/input modifications are made 😅

For sure, this is definitely driving home how much these parameters are duplicated :)

src/celeritas/phys/PhysicsTrackView.hh
sethrj: yes, `is_muhad` also works. And I guess adding that explicitly would be idealistic 🤔 why 1 MeV and not 600 KeV 🤣

src/celeritas/phys/PhysicsParams.hh
```diff
@@ -71,42 +134,36 @@ class ParticleParams;
...
+    real_type lambda_limit{real_type{1} * units::millimeter};
+    real_type safety_factor{0.6};
     //!@}
 
-    real_type secondary_stack_factor = 3;
-    bool disable_integral_xs = false;
+    //!@{
+    //! \name Particle-dependent parameters
+    ParticleOptions em{ParticleOptions::default_em()};
+    ParticleOptions muhad{ParticleOptions::default_muhad()};
```
sethrj: Not trying to rock the boat, but since we now have `is_heavy` (which I like, since that's really what characterizes the different slowing down behaviors) should we make these light/heavy rather than em/muhad? I'm fine with keeping them as is to avoid you spending your whole life renaming things 😅

sethrj (approved):

src/celeritas/phys/PhysicsParams.hh
amandalund: Nah, I think light/heavy is good.

stognini (approved):
Looks great. Thanks @amandalund

src/celeritas/phys/PhysicsParams.hh
```diff
@@ -34,14 +34,83 @@ class ParticleParams;
 
 //---------------------------------------------------------------------------//
 /*!
- * Physics configuration options.
+ * Particle-dependent physics configuration options.
+ *
+ * These parameters have different values for electrons/positrons and
```
stognini: Maybe also explicit here the convention:
electrons/positrons (light) and muons/hadrons (heavy). Or the other way around: heavy particles (e-/e+)...

Since it is kind of arbitrary, repeating in the documentation helps tracking the definition while navigating the code.

PR #1582, opened 2025-01-16
Add postfix to infix conversion utility

esseivaju:
SCALE exports logic expressions in the postfix notation which is parsed by Celeritas to a sequence of `logic_int`. This add an utility to convert a postfix sequence to infix.


src/orange/orangeinp/detail/LogicUtils.hh
```diff
@@ -23,6 +23,106 @@ namespace orangeinp
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a postfix logic expression to an infix expression.
+ *
+ * The \c InfixEvaluator will short-circuit evaluation of operands based
+ * on parenthesis depth. Minimizing that depth in the expression
+ * will allow to short-circuit more efficiently.
+ */
+inline std::vector<logic_int> convert_to_infix(Span<logic_int> postfix)
```
sethrj: Input should be span to const data; also this is a pretty big function for inlining. I think it could also be more readable as a little private helper class?

test/orange/orangeinp/detail/LogicUtils.test.cc
```diff
@@ -0,0 +1,97 @@
...
+//! \file orange/orangeinp/detail/LogicUtils.test.cc
+//---------------------------------------------------------------------------//
+#include "orange/orangeinp/detail/LogicUtils.hh"
+
+#include <string_view>
+#include <vector>
+
+#include "corecel/cont/Span.hh"
+#include "orange/OrangeTypes.hh"
+#include "orange/detail/OrangeInputIOImpl.json.hh"
```
sethrj: Instead of this detail include and the JSON requirement, what about moving `string_to_logic` and `logic_to_string` to `orangeinp/LogicUtils.{hh,cc}` as well as putting the infix conversion there?

src/orange/orangeinp/detail/LogicUtils.hh
```diff
@@ -23,6 +25,45 @@ namespace orangeinp
 {
 namespace detail
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a logic token to a string.
+ */
+inline void logic_to_stream(std::ostream& os, logic_int val)
```
sethrj: When I said `orangeinp/LogicUtils` I was particularly thinking of dropping the detail namespace since this file is now going to be used outside `orangeinp`. I think we'd also call `convert_to_infix` from the input as well, right?

src/orange/orangeinp/detail/LogicUtils.cc
```diff
@@ -0,0 +1,220 @@
...
+        // hijack ltrue as the token to represent a primitive
+        infix.push_back({logic::ltrue, {elem}});
+    }
+
+    //! Get the infix expression.
+    std::vector<logic_int> const& get_infix() const
+    {
+        CELER_EXPECT(infix.size() == 1);
+        return infix.front().expr;
+    }
```
sethrj: Either add or change:
```suggestion
    std::vector<logic_int> get_infix() &&
    {
        CELER_EXPECT(infix.size() == 1);
        return std::move(infix.front().expr);
    }
```

src/orange/orangeinp/detail/LogicUtils.hh
sethrj: And actually, since this helper function only requires a span of logic ints (it doesn't touch any of the CSG tree stuff), maybe it should go into orange/detail and not be in orangeinp at all? 🤔 I'm not sure it belongs here.

sethrj (approved):
Much more readable 😄 

Now just to figure out the place to put these guys...

src/orange/orangeinp/detail/LogicUtils.hh
esseivaju: Ok, it makes sense to move it outside of the detail namespace. Right that it doesn't use the CSG tree but it still deals with transforming input to orange, that why I placed it there. The other logic-related stuff is inside `orange/univ/detail` so maybe there?

sethrj (approved):
Looks great. Thanks!

src/orange/orangeinp/detail/LogicUtils.hh
sethrj: This is only about input (and `univ` should be mostly runtime?) so I like the place you have it now.

PR #1584, opened 2025-01-16
Fix spurious vecgeom warning

sethrj:
This fixes a warning that occurs when boolean volumes are present in a converged Geant4 geometry:
```
src/geocel/vg/VecgeomParams.cc:287: warning: Geant4 logical volume {null G4LogicalVolume} maps to multiple volume IDs 
```

We can safely ignore volumes that don't map to Geant4 volumes.


drbenmorgan (approved):
Thanks for the quick fix @sethrj, all good to go!

PR #1585, opened 2025-01-16
Add ARM runner to CI

sethrj:
This increases testing by adding support for GitHub's [newly released](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) ARM runner preview.


esseivaju (approved):
LGTM! Maybe just mention somewhere in the doc/readme that arm is tested?

PR #1586, opened 2025-01-17
Fix weird edge cases when trying to `find_package(Celeritas)`

sethrj:
https://github.com/celeritas-project/celeritas/pull/1573 saw some weird failures against Geant4 11.1 and 11.2 but not 11.0 nor 11.3. The failures occurred in the second invocation of `find_dependency(XercesC)` (missing `XercesC_INCLUDE_DIR`) but went away if we added a `REQUIRED` argument to `find_package(Celeritas)`.

Ultimately there were several factors at play:
1. `find_dependency` caching suppressed the failing XercesC only when `REQUIRED` was set
2. VecGeom and Celeritas both set `XercesC_INCLUDE_DIR` as non-cache variables, and Geant4 set it as a cache variable
3. Old CMake policy [CMP0102](https://cmake.org/cmake/help/latest/policy/CMP0102.html) causes the `mark_as_advanced` in `FindXercesC.cmake` to write an empty cache variable if a normal variable is set
4. Old CMake policy [CMP0126](https://cmake.org/cmake/help/latest/policy/CMP0126.html) then causes the `find_path` command to overwrite the normal variable with the empty cache value (on the next call to FindXercesC)
5. Maybe policy [CMP0011](https://cmake.org/cmake/help/latest/policy/CMP0011.html) which protects policy scopes wasn't working??
6. The *real* underlying cause: Geant4 11.1 included PTL, which had a `cmake_minimum_required(3.8)` that reset all the policies to that ancient value. This was fixed in https://github.com/jrmadsen/PTL/pull/51 and the fix was merged into Geant4 11.3.

I've bypassed the root cause of normal/cache mismatch by propagating the CACHE status to the CeleritasConfig.cmake macro. I also added policy scoping to try to limit the damage the other scripts can do.


drbenmorgan (approved):
This looks o.k. to me. The only _possible_ concern is whether there could be issues from the use of `CACHE` and the order in which packages are found and thus how the export variables are set/unset. However, that's not worth worrying about at this stage.

pcanal:
As a side reference, see https://github.com/root-project/root/issues/16374 related to the scope of items introduced by `find_package`.

PR #1587, opened 2025-01-18
Update Perlmutter env

esseivaju:
`libsci` was being linked twice, from OpenMP and the libsci module. This caused a warning  at runtime which which was making the regression test fail to parse the JSON output... Disabling that module results in another error: a few CUDA test fail to link with the following error: 
`test/celeritas/CMakeFiles/celeritas_phys_Particle.dir/phys/Particle.test.cu.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'`. I looked at the symbol `Particle.test.cu.o` and there is indeed `dlclose`, it is not clear to me who is bringing that symbol there.

The failure only happens with VecGeom build, the ORANGE build has `-ldl` in the linker invocation, I don't know why it's different for VecGeom.  Just work around it by setting `CMAKE_EXE_LINKER_FLAGS`...

### DIFF
--- a/scripts/cmake-presets/perlmutter.json
+++ b/scripts/cmake-presets/perlmutter.json
@@ -19,6 +19,7 @@
         "CELERITAS_USE_MPI":    {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_ROOT":    {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_VecGeom": {"type": "BOOL", "value": "ON"},
+        "CMAKE_EXE_LINKER_FLAGS": {"type": "STRING", "value": "-ldl"},
         "CMAKE_CXX_FLAGS": "-Wall -Wextra -Wno-psabi -pedantic -pedantic-errors",
         "CMAKE_CUDA_FLAGS": "-lineinfo -Xptxas=-v -Xcompiler -Wno-psabi",
         "CMAKE_CUDA_ARCHITECTURES": {"type": "STRING", "value": "80"},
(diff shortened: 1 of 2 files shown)


sethrj (approved):
Good to get that debugged; and cray libs are the worst. Thanks @esseivaju !

@pcanal the missing `dl` might be a RDC link magic issue 😬

esseivaju:
I checked a verbose [CI](https://github.com/esseivaju/celeritas/actions/runs/12838732166/job/35804860775) run, and you can see that orange links to `libdl` here too, whereas VecGeom does not. It doesn't fail to link though, so maybe something in my environment is injecting that dependency in my VecGeom build 🤔

pcanal:
> the missing dl might be a RDC link magic issue 😬

I would not be surprised ;( ... sigh ...

scripts/cmake-presets/perlmutter.json
```diff
@@ -19,6 +19,7 @@
         "CELERITAS_USE_MPI":    {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_ROOT":    {"type": "BOOL", "value": "ON"},
         "CELERITAS_USE_VecGeom": {"type": "BOOL", "value": "ON"},
+        "CMAKE_EXE_LINKER_FLAGS": {"type": "STRING", "value": "-ldl"},
```
pcanal, line 22: There is also `${CMAKE_DL_LIBS}` that can be used instead of literal `-ldl`

PR #1591, opened 2025-01-23
Output diagnostic information about core sizes

sethrj:
Based on the mess in https://github.com/celeritas-project/celeritas/pull/1577 and discussions in https://github.com/celeritas-project/celeritas/pull/1562 , we really need a definitive diagnostic about the state sizes.

The gotcha is that the number of states isn't previously known to the core params (it's passed on constructing the stepper or core state). With #1556 we want to transition to more tightly integrating the params and state, so for now I'm adding an optional "state size" to the params that can be used to create the core state. For now this state size is set by the apps (where we care about the diagnostic) but not the unit tests (which often reuse the same core params with different stepper sizes).

### DIFF
--- a/app/celer-sim/Runner.cc
+++ b/app/celer-sim/Runner.cc
@@ -448,6 +448,12 @@ void Runner::build_core_params(RunnerInput const& inp,
                       "streams ("
                    << params.max_streams << " requested)");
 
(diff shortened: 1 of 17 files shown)


amandalund (approved):
Good idea, thanks @sethrj!

Issue #1375, opened 2024-08-19
Basic muon decay physics


sethrj:
Add decay for muons.

Possible design:
- new `DecayProcess` class (maybe eventually we should rename `Process` to `InteractionProcess`?) that applies to one or more particle IDs
- Decay process vector added to `PhysicsParams`
- decay constants from particle types get turned into effective cross sections inside `calc_physics_step_limit`
- select decay process as part of `select_discrete_interaction`?
- Do we have multiple "models" for the process? Or separate decay channels that should be sampled? Do we ever need to have different models per physics region?
- Otherwise like optical physics we could have single model per process, and if so `DecayProcess` should probably inherit from `CoreStepActionInterface`

- [ ] `G4VDecayChannel`: `G4MuonDecayChannel`

See https://github.com/celeritas-project/celeritas/pull/1592/files/75c451b9264031390b19fa23c24d2817bbfc1bc9#diff-6d35409924ea422b7d39eda626cacd61c9c093bee132d578e797ee8abe5f9d64

PR #1592, opened 2025-01-24
Always select at-rest process for stopped tracks

amandalund:
This updates `select_discrete_interaction` to check first whether the particle is stopped and has an at-rest process and to choose that process if so. This is similar to what Geant4 does at the beginning of a step: check at-rest processes that apply to the particle and choose the one with the smallest time to interaction. This should prevent us from sampling a process that doesn't apply to a stopped particle (e.g. bremsstrahlung for a stopped positron) and then rejecting it through the integral method and taking an extra step (see discussion in #907). This is also a first step toward #1375.


src/celeritas/phys/PhysicsData.hh
```diff
@@ -121,6 +121,12 @@ struct IntegralXsProcess
  * \c range are the process-integrated dE/dx and range for the particle.  \c
  * integral_xs will only be assigned if the integral approach is used and the
  * particle has continuous-discrete processes.
+ *
+ * \todo If it's possible for a particle to have multiple at-rest processes, \c
+ * at_rest should be the process with the smallest lifetime. This is used in \c
+ * select_discrete_interaction to choose the at-rest process with the smallest
+ * time to interaction if the particle is stopped and has a process that
+ * applies at rest.
```
sethrj, line 129: Depending what processes and their probabilities are in play, we might need to sample from a combination of exponential distributions. Are we going to have a single decay process per particle type, or can multiple compete? How do we determine the time to absorption of an "at rest" positron: is it deterministic or probabilistic, and with what distribution?

src/celeritas/phys/PhysicsStepUtils.hh
```diff
@@ -269,28 +269,40 @@ select_discrete_interaction(MaterialView const& material,
     CELER_EXPECT(physics.interaction_mfp() <= 0);
     CELER_EXPECT(pstep.macro_xs() > 0);
 
-    // Sample ParticleProcessId from physics.per_process_xs()
-    ParticleProcessId ppid = celeritas::make_selector(
-        [&pstep](ParticleProcessId ppid) { return pstep.per_process_xs(ppid); },
-        ParticleProcessId{physics.num_particle_processes()},
-        pstep.macro_xs())(rng);
```
sethrj, line 276: Could you break `find_ppid` into a separate helper function, returning `{}` for rejection? Also I noticed that the integral rejection should probably use the rejection sampler for error checking + clarity.

test/celeritas/global/AlongStep.test.cc
```diff
@@ -228,7 +228,7 @@ TEST_F(MockAlongStepTest, basic)
         EXPECT_SOFT_EQ(1, result.angle);
         EXPECT_SOFT_NEAR(1.2431209185653e-12, result.time, 1e-11);
         EXPECT_SOFT_EQ(5.2704627669473e-05, result.step);
-        EXPECT_EQ("physics-discrete-select", result.action);
+        EXPECT_EQ("eloss-range", result.action);
```
sethrj, line 231: It looks like this changed because `celeriton` doesn't have an equivalent at-rest process. Can you add one so that we can check that we correctly test?

test/celeritas/phys/Physics.test.cc
```diff
@@ -376,7 +376,7 @@ TEST_F(PhysicsTrackViewHostTest, processes)
         ParticleProcessId const abs_ppid{1};
         EXPECT_EQ(ProcessId{0}, phys.process(scat_ppid));
         EXPECT_EQ(ProcessId{1}, phys.process(abs_ppid));
-        EXPECT_TRUE(phys.has_at_rest());
+        EXPECT_EQ(ParticleProcessId{}, phys.at_rest_process());
```
sethrj, line 379: It looks like this change is related; we want `celeriton` to have an at-rest process for testing purposes.

amandalund: Yeah, previously we were determining if a particle has an at-rest process by checking if the cross section at zero energy was positive. This wasn't really the right way to do it, e.g. some processes like photoelectric have a non-zero cross section but aren't really "at-rest" processes. I'm not sure what's the best way to identify processes that apply at rest... we might need to do something like add an `applies_at_rest()` method to the `Process` class?

sethrj: Yeah, I guess this ties into the applicability mess. If it has a nonzero cross section at zero energy, it really *should* be at rest. But we don't have a good way of creating a cross section grid that's valid down to "low-ish" energies and zero below that. Related to #907 I guess...

src/celeritas/phys/PhysicsData.hh
amandalund: My current thinking was a single decay process, with the handful of concrete implementations of `G4VDecayChannel` (`G4MuonDecayChannel`, `G4PhaseSpaceDecayChannel`, etc.) as the "models". If the decay process is selected, the decay channel will be sampled according to the branching ratios and will map to one of those models. Still just sketching it out though, so I don't know yet if that's the right approach.

At-rest positrons (in Geant4, at least) always have zero time to interaction.

src/celeritas/phys/PhysicsStepUtils.hh
amandalund: I think we weren't using the rejection sampler there because it's possible for the post-step xs to be larger than the estimate of the max xs over the step.

src/celeritas/phys/PhysicsData.hh
sethrj: A couple thoughts:
- Whatever we should do ought to allow decay to compete with in-flight processes and not just apply to at-rest.
- Models are selected based on the particle energy/material/local physics options; they don't compete like decay channels. So we probably need another abstraction beside "model" if decay is a single "process". If, in the future, we need to sample from potential models at a given energy (e.g. for smoothing out energy-dependent hadronic processes which I believe Soon mentioned) we could also potentially use a "channel" to sample from multiple applicable models.
- I know that positron annihilation happens (sometimes at least?) with positronium as an intermediate, which has a finite lifetime, and the time to annihilation is used in [certain applications](https://en.wikipedia.org/wiki/Positron_annihilation_spectroscopy). So there's other models where annihilation have nonzero decay times.
- @stognini Does muonic fusion have an exponential distribution?

amandalund: Yep, decay will compete with in-flight processes. And yeah, the channels aren't really analogous to models...  I was just thinking like our EM process/models we might have a single decay process and an action/interactor for each channel type. I'll try to sketch out something a little more concrete that we can bounce around...

Positrons in flight will have a nonzero time to annihilation, but as far as I can tell, Geant4 for whatever reason uses zero for the [time to interaction at rest](https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/processes/electromagnetic/standard/src/G4eplusAnnihilation.cc#L101-L106) (at rest GPIL returns time rather than length). Hopefully that's good enough for us too for now? 🤷

stognini: > Does muonic fusion have an exponential distribution?

I don't think so. Both the muonic atom formation and fusion are `at rest` processes. The atomic capture has a physical interaction length (GPIL) set to `0`, while the fusion itself (dd, dt, or tt) depends on a set of hardcoded tables that are interpolated. These tables are seconds (x) and temperature (y), and they are used to define the GPIL. These curves are not exponential, from what I recall.

sethrj: Ok that's right. So does it *override* muonic decay or *compete* with it? I guess it overrides and then the atom formation incorporates decay?

stognini: There should still be at least some competition. Since the muon should be able to: just decay; form a muonic atom; then the muonic atom undergoes molecule formation, then fusion, and during all that the muon is still physically able to decay.

Now, the fusion itself is ~ $10^{-12}$ s, while the muon lifetime is $2.2 \times 10^{-6}$ so I guess codewise when you sample the exponential for the decay, it will hardly win after the molecule formation is called... So in the current G4 implementation done by Ara, it looks like if the molecule formation is called, it will fuse no matter what, and that part overrides the decay itself.

test/celeritas/phys/Physics.test.cc
amandalund: I pushed a [commit](https://github.com/celeritas-project/celeritas/pull/1592/commits/cd4312e48a6b2b41451b7457a4d295d3cd7e4df6) with some ideas (flagging at-rest processes in the imported data; adding a `applies_at_rest()` method to `Process`) which I can partially or totally revert. Using imported data to determine whether the process applies at rest might be silly and we could just hardcode it, or we could try to pass that imported data directly into the `PhysicsParams` constructor somehow and avoid modifying `Process`, or do something totally different...

src/celeritas/phys/PhysicsStepUtils.hh
```diff
@@ -28,6 +28,56 @@
...
 namespace celeritas
 {
+namespace
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample the process for the discrete interaction.
+ */
+template<class Engine>
+CELER_FUNCTION ParticleProcessId find_ppid(MaterialView const& material,
```
sethrj:
```suggestion
inline CELER_FUNCTION ParticleProcessId find_ppid(MaterialView const& material,
```

src/celeritas/phys/PhysicsStepUtils.hh
sethrj: If our estimate of the maximum is wrong, doesn't that mean we're not doing rejection sampling correctly? (Or is it just better than not doing rejection even though it's incorrect?) Or am I being obtuse?

test/celeritas/global/AlongStep.test.cc
```diff
@@ -279,8 +279,8 @@ TEST_F(MockAlongStepFieldTest, TEST_IF_CELERITAS_DOUBLE(basic))
         EXPECT_SOFT_EQ(6.9431339225049422e-10, result.time);
         EXPECT_SOFT_EQ(0.930177246841563, result.step);
         EXPECT_SOFT_EQ(0, result.mfp);
-        EXPECT_SOFT_EQ(1, result.alive);
-        EXPECT_EQ("physics-discrete-select", result.action);
+        EXPECT_SOFT_EQ(0, result.alive);
+        EXPECT_EQ("eloss-range", result.action);
```
sethrj, line 283: Good to have the unit test document the key change 😄

test/celeritas/phys/MockProcess.hh
```diff
@@ -74,6 +74,7 @@ class MockProcess : public Process
         SPConstMaterials materials;
         std::string label;
         bool use_integral_xs;
+        bool applies_at_rest;
```
sethrj: We should default initialize these (true for integral xs,false for at rest?)

sethrj (approved):
Those unit tests look like a pain to have updated, but this is a nice result. Thanks @amandalund !

src/celeritas/phys/PhysicsStepUtils.hh
amandalund: No you're right, if the estimate of the maximum is wrong the sampling won't be correct. I think the only time this should happen though if the cross section has more than one peak and the second local maximum is within the pre- and post-step energy. We could mitigate this by storing multiple maximum xs values if there are multiple peaks (it looks like Geant4 now has a [`G4CrossSectionType`](https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/processes/electromagnetic/utils/include/G4EmTableType.hh#L54) that let's you specify up to two peaks). My understanding though is that the cross sections for these energy loss processes tend to be unimodal.

PR #1593, opened 2025-01-24
Change PrimaryGeneratorAction to be a wrapper 

sethrj:
This prerequisite for #1583  simplifies the generator action so that it simply wraps the other Celeritas class. In the future we could make some minor changes to make this a generalized wrapper for any Celeritas primary generator.


amandalund (approved):
Looks good!

PR #1595, opened 2025-01-26
Set step length and other data in reconstructed tracks

sethrj:
It looks like the failures in https://github.com/celeritas-project/celeritas/issues/1144#issuecomment-2614035073 were due to https://gitlab.cern.ch/atlas/athena/-/blob/main/LArCalorimeter/LArG4/LArG4Barrel/src/LArBarrelCalculator.cxx#L192 , where the hit is discarded if the step length is zero: and prior to now we didn't set the step length.

I looked at what else is set in the track/step/steppoint and set a couple of other attributes. Some attributes that we *don't* set I've set to nonsensical values that will hopefully cause something to blow up rather than silently discarding the hit.


esseivaju (approved):

amandalund (approved):
Good find!

PR #1596, opened 2025-01-26
Add hit manager output diagnostic

sethrj:
It was super useful to have the SD information from Athena, since that helped point to the location of the relevant code in a *very* large code base. Instead of having to turn on debug logging and grep through the results, this now lets us programmatically pull the details of the mapped SDs: Celeritas volume ID (from which we can get a volume via the geo diagnostic), SD name given by the Geant4 app, SD C++ type, and the Geant4 volume it's connected to.
```json
{
  "_category": "internal",
  "_label": "hit-manager",
  "locate_touchable": true,
  "lv_name": [
    "em_calorimeter",
    "had_calorimeter"
  ],
  "sd_name": [
    "em_calorimeter",
    "had_calorimeter"
  ],
  "sd_type": [
    "celeritas::test::SimpleSensitiveDetector",
    "celeritas::test::SimpleSensitiveDetector"
  ],
  "vol_id": [
    3,
    4
  ]
}
```

### DIFF
--- a/src/accel/CMakeLists.txt
+++ b/src/accel/CMakeLists.txt
@@ -24,6 +24,7 @@ list(APPEND SOURCES
   SimpleOffload.cc
   detail/GeantSimpleCaloSD.cc
   detail/HitManager.cc
+  detail/HitManagerOutput.cc
   detail/HitProcessor.cc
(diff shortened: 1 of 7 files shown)


amandalund (approved):
Nice!

PR #1597, opened 2025-01-27
Write out optical size diagnostic

sethrj:
This is a minor follow-on to #1591. It writes out the optical sizes just like we do the core sizes. I also changed the output key for consistency with the other output diagnostics.


amandalund (approved):
Looks good, thanks!

PR #1599, opened 2025-01-28
Use BIH to accelerate background_intersect

elliottbiondo:
This PR hooks up `BIHIntersectingVolumeFinder` to `background_intersect` in order to improve performance for LZ-like geometries, in which many of small volumes are present within a background volume. A brief scaling study was conducted with such a model, in which 2D grids of cylindrical wedges were created over a sweep of grid sizes [10, 100]. The 100x100 grid is shown here:

simple-lz-geom

Ray tracing was done via `celerpy` using this branch and develop. Results show significant performance gains:

scaling


elliottbiondo:
Draft until I work out these DBC failures.

sethrj:
Nice @elliottbiondo ! Can you plot on a log-log graph as well? Is this just single-thread CPU performance?

elliottbiondo:
Yes this is single-thread CPU performance. We can discuss how we want to demonstrate GPU performance, possibly with optical photons.

Here is the log-log plot:
scaling

W.r.t. the DBC errors here: they are cropping up because I made it so background volumes have null bounding boxes. On further consideration, I think I am going to change it back so that background volumes have infinite bboxes, but the BIHBuilder constructor will take an `std::set` of `vol_ids` denoting the implicit vols (to be omitted from the tree).

elliottbiondo:
@sethrj pending CI this is ready for review. Two things:

- I can cleanup the `BIHIntersectingVolFinder` `visit_vol` functor to not require the `[[maybe used]]` args which can always be handled by lambda capture.
- Right now the BIH stores, but does use, the infinite bounding boxes associated with background volumes. This keeps the indexing simple. If we anticipate a very large number of these infinite bounding boxes, I can do something more sophisticated.

src/orange/detail/UnitInserter.cc
```diff
@@ -303,7 +303,7 @@ UniverseId UnitInserter::operator()(UnitInput&& inp)
         // Create a set of background volume ids for BIH construction
         if (inp.volumes[i].flags & VolumeRecord::Flags::implicit_vol)
         {
-            background_vol_ids.insert(LocalVolumeId{i});
+            background_vol_ids.insert(LocalVolumeId(i));
```
sethrj: There's a newish helper class for constructing IDs from possibly different integer types:
```suggestion
            background_vol_ids.insert(id_cast<LocalVolumeId>(i));
```

src/orange/detail/BIHBuilder.cc
```diff
@@ -32,7 +32,9 @@ BIHBuilder::BIHBuilder(Storage* storage)
 /*!
  * Create BIH Nodes.
  */
-BIHTree BIHBuilder::operator()(VecBBox&& bboxes)
+BIHTree
+BIHBuilder::operator()(VecBBox&& bboxes,
+                       BIHBuilder::SetLocalVolId const& background_vol_ids)
```
sethrj: What clang-format version do you have, by the way? I've seen some back-and-forth changes with odd formatting across the different versions.

src/orange/BoundingBoxUtils.hh
```diff
@@ -228,9 +228,9 @@ inline bool encloses(BoundingBox<T> const& big, BoundingBox<T> const& small)
  * intersection, the result will be inf.
  */
 template<class T, class U>
-inline U calc_dist_to_inside(BoundingBox<T> const& bbox,
-                             Array<U, 3> const& pos,
-                             Array<U, 3> const& dir)
+inline CELER_FUNCTION U calc_dist_to_inside(BoundingBox<T> const& bbox,
+                                            Array<U, 3> const& pos,
+                                            Array<U, 3> const& dir)
```
sethrj, line 233: Now that we're doing this on GPU:
- The loop over bounds could be a bit better with `range(Bound::size_)` (avoid the static cast)
- `auto intersect = `  could use an `axpy` (aka, reliably using std::fma)
- Should we be doing all these operations in the "fast" precision `T`? I think we can assume that our bbox has been expanded outward (bump plus float-to-double rounding) but we should document that and do some testing as well. So perhaps everything except for pos/dir should be in `T` precision.
- On 273, do we want to allow a zero distance to intersection? We forbid that in most cases.
- Also on 273, we could save half of the (fp conversion + division + comparison) opertions if you wait to divide by `dir[ax]` until after the `dist < 0`.

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -589,78 +597,55 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
...
 {
-    // Calculate bump distance
-    real_type const bump_dist
-        = detail::BumpCalculator{params_.scalars.tol}(state.pos);
+    auto is_intersecting
+        = [this, &state, &is_valid](LocalVolumeId vol_id) -> Intersection {
+        VolumeView vol = this->make_local_volume(vol_id);
+
+        detail::CalcIntersections calc_intersections{
+            celeritas::forward<F>(is_valid),
```
sethrj: This isn't safe to forward because you're calling this multiple times per `background_intersect`. You should just pass by reference.

src/orange/detail/BIHIntersectingVolFinder.hh
```diff
@@ -117,15 +116,13 @@ BIHIntersectingVolFinder::BIHIntersectingVolFinder(
...
- * detail::Intersection(*)(LocalVolumeId id, Ray ray, real_type
- * max_search_dist)
+ * detail::Intersection(*)(LocalVolumeId id)
  *
- * In other words, for a given LocalVolumeId and Ray, it should provide an
- * Intersection object denoting the next surface and the corresponding
- * distance. If no intersection is found within max_search_dist, an empty
- * Intersection object is returned.
+ * Other information required by the functor should be handled through
+ * lambda capture.
```
sethrj, line 125: I don't think we want this. The lambda can't capture the decreasing distance to nearest volume, so it's doing extra work (testing surfaces) it doesn't need to. Right?

src/orange/detail/UnitInserter.cc
```diff
@@ -299,6 +300,12 @@ UniverseId UnitInserter::operator()(UnitInput&& inp)
             bboxes.push_back(BoundingBox<fast_real_type>::from_infinite());
         }
 
+        // Create a set of background volume ids for BIH construction
+        if (inp.volumes[i].flags & VolumeRecord::Flags::implicit_vol)
```
sethrj, line 304: I think this will also pick up "implicit exterior" (which can never intersect) volumes. Could you check?

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -541,7 +547,7 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
     // Calculate local senses, taking current face into account
     // Current senses should put us inside the volume
     detail::LogicEvaluator is_inside(vol.logic());
-    CELER_ASSERT(is_inside(calc_sense));
+    CELER_ASSERT(!(is_inside(calc_sense) ^ static_cast<bool>(target_sense)));
```
sethrj: bool != bitwise, and I *think* this can be more explicit with a comparison:
```suggestion
    CELER_ASSERT(to_sense(is_inside(calc_sense)) == flip_sense(target_sense));
```
or perhaps more readably:
```suggestion
    CELER_ASSERT(is_inside(calc_sense) != (target_sense == Sense::inside));
```

We definitely don't want to static cast the target sense to `bool` because we might someday flip inside/outside sense so that  "inside" is true.

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -565,7 +571,9 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
         }();
         axpy(distance - previous_distance, state.dir, &pos);
 
-        if (!is_inside(calc_sense))
+        // Intersection is found if is_inside is true and the target sense
+        // is inside (n.b., static_cast<inside> is false), or vice-versa
+        if (is_inside(calc_sense) ^ static_cast<bool>(target_sense))
```
sethrj:
```suggestion
        if (is_inside(calc_sense) == (target_sense == Sense::inside))
```

sethrj (commented):
Great speed-up results on CPU; hope the added complexity doesn't kill the GPU. I still have to dive into the updated `background_intersect`.

src/orange/detail/BIHBuilder.cc
elliottbiondo: Version 19.1.5. Just updated via Homebrew to 19.1.7. Where do you get it?

src/orange/detail/BIHIntersectingVolFinder.hh
elliottbiondo: You mean that the `complex_intersect` loop should break early, before testing intersections at distances that would be greater than the current minimum. Yes that would involve passing a `max_search_dist` to the functor. I can add all this in.

src/orange/detail/UnitInserter.cc
elliottbiondo: Yes, in `orangeinp/UnitProto.cc`, `implicit_vol` get set for exterior volumes:

Lines 236-240:
```
    // Save attributes for exterior volume
    if (input.next_id() != orange_global_universe)
    {
        vol_iter->zorder = ZOrder::implicit_exterior;
        vol_iter->flags |= VolumeRecord::implicit_vol;
    }
  ```
  
 I will change `background_vol_ids` to `implicit_vol_ids` for clarity.

File: src/orange/BoundingBoxUtils.hh (hunk 3 above)
Comment (elliottbiondo) [reply src/orange/BoundingBoxUtils.hh:233]: When switching to fast precision, the use of `axpy` becomes awkward. Its it worth creating copies of `pos` and `dir` just to get the performance benefits of `axpy`?

File: src/orange/detail/BIHIntersectingVolFinder.hh (hunk 5 above)
Comment (elliottbiondo) [reply src/orange/detail/BIHIntersectingVolFinder.hh:125]: (done)

**Comment by elliottbiondo** [comment]

File: src/orange/BoundingBoxUtils.hh (hunk 3 above)
Comment (elliottbiondo) [reply src/orange/BoundingBoxUtils.hh:233]: (all other changes implemented)

Comment (elliottbiondo) [reply src/orange/BoundingBoxUtils.hh:233]: > but we should document that and do some testing as well

I forgot to mention: I added a line to the BIHBuilder stating that it expects bumped bounding boxes. `UnitInserter` bumps every bounding box, so there isn't a mechanism for non-bumped bounding boxes to reach the BIH. I don't think we need to test anything specific in this respect, as any tracking tests with background volumes test this implicitly.

Comment (sethrj) [reply src/orange/BoundingBoxUtils.hh:233]: I'm sorry, I meant `std::fma`:
```c++
auto intersect = celeritas::fma<T>(dist, dir[other_ax], pos[other_ax]);
```

src/orange/BoundingBoxUtils.hh
```diff
@@ -267,14 +267,14 @@ inline U calc_dist_to_inside(BoundingBox<T> const& bbox,
                 continue;
             }
 
-            U dist = (bbox.point(static_cast<Bound>(bound))[ax] - pos[ax])
-                     / dir[ax];
+            T temp = (bbox.point(static_cast<Bound>(bound))[ax] - pos[ax]);
 
-            if (dist < 0)
+            if (temp * signum(dir[ax]) <= 0)
```
sethrj: Oh yeah I forgot that the sign of the direction mattered 🤦 it's more readable and possibly faster the original way you wrote it, sorry!

src/orange/detail/BIHBuilder.cc
sethrj: 18 I think? and same but that's good

sethrj (approved):

PR #1601, opened 2025-01-29
Construct CoreParams from new problem input in celer-sim

sethrj:
This follow-on to #1562 for #1556 moves the core construction from `app/celer-sim` to `celeritas/setup`.  I removed a couple of small capabilities related to the deprecated celer-sim json input for now:
- We no longer echo the JSONified input to the root output file
- If something goes wrong during problem setup, we no longer echo the "processed" JSON input

Changed in `inp`:
- Split `inp::StateCapacity`  into core/optical
- Restored full eloss spline order option
- Made a few "ignored if zero" options `std::optional` rather than sentinel values


amandalund:
Currently I'm hitting an error trying to run celer-sim on the GPU:
```console
/home/alund/celeritas_project/celeritas/app/celer-sim/celer-sim.cc:273: critical: While running input at simple-cms-gpu.inp.json: celeritas: runtime error: celeritas::activate_device may be called only once per application
/home/alund/celeritas_project/celeritas/src/corecel/sys/Device.cc:308: '!d' failed
```

sethrj:
Glad someone here is paying attention 😅 I'll check it out on wildstyle and fix.

sethrj:
@amandalund all tests pass on a debug+openmp+cuda build on wildstyle!

amandalund:
Awesome, thanks! I'll get back to this today.

app/celer-sim/celer-sim.cc
```diff
@@ -255,26 +260,40 @@ int main(int argc, char* argv[])
...
+    CELER_ASSERT(output);
     output->output(&cout);
     cout << endl;
 
+    // Delete streams before end of program (TODO: this is because of a static
+    // initialization order issue; CUDA can be deactivated before the global
+    // celeritas::device is reset)
+    if (celeritas::device())
+    {
+        celeritas::device().create_streams(0);
```
amandalund: We also do this in `SharedParams`, maybe we should have a `Device::destroy_streams()`?

app/celer-sim/RunnerInput.cc
```diff
@@ -207,6 +265,9 @@ inp::StandaloneInput to_input(RunnerInput const& ri)
     si.problem = load_problem(ri);
     if (!ri.physics_file.empty())
     {
+        CELER_VALIDATE(
+            ends_with(ri.physics_file, ".root"),
+            << R"(physics_file must be a ROOT input: use GDML for geometry_file and if forcing an ORANGE geometry, use the `ORANGE_FORCE_INPUT` environment variable)");
```
amandalund, line 274: Should GDML physics/org.json ORANGE be deprecated input formats? Or are we not worrying about backward compatibility here.

app/celer-sim/RunnerInput.cc
```diff
@@ -219,6 +280,23 @@ inp::StandaloneInput to_input(RunnerInput const& ri)
...
+            [](inp::ReadFileEvents const& rfe) {
+                if (ends_with(rfe.event_file, ".root"))
+                {
+                    return RootEventReader{rfe.event_file, nullptr}.num_events();
+                }
+
+                return EventReader{rfe.event_file, nullptr}.num_events();
+            },
+        },
+        si.events);
```
amandalund, line 302: Since we're getting the number of events here, could we also update the number of streams to not be more than the number of events?

src/celeritas/setup/StandaloneInput.hh
```diff
@@ -0,0 +1,40 @@
...
+class CoreParams;
+
+namespace inp
+{
+struct StandaloneInput;
+}
+namespace setup
+{
+//---------------------------------------------------------------------------//
+struct StandaloneLoaded
```
amandalund, line 26: How will this be used?

src/celeritas/setup/Problem.cc
```diff
@@ -0,0 +1,519 @@
...
+        return std::make_shared<GeoParams>(filename);
+    }
+
+    //! Build from Geant4
+    result_type operator()(G4VPhysicalVolume const* world)
+    {
+        if constexpr (CELERITAS_CORE_GEO == CELERITAS_CORE_GEO_ORANGE)
+        {
+            static char const fi_hack_envname[] = "ORANGE_FORCE_INPUT";
+            auto const& filename = celeritas::getenv(fi_hack_envname);
```
amandalund, line 103: Why is this hack necessary?

src/celeritas/setup/Problem.cc
```diff
@@ -0,0 +1,519 @@
...
+                -> std::shared_ptr<CoreStepActionInterface> {
+                CELER_NOT_IMPLEMENTED("building RZ map field through input");
+            },
+        }),
+        var_field);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Costruct magnetic field from variant input.
```
amandalund: Update description.

amandalund (commented):
This looks good @sethrj! One kind of odd thing is the results on this branch were different from develop (but still statistically the same) when I tried running the cms2018 problem below. Is there any reason the results might have changed?

</pre></details>

 <details><summary>cms2018 json input</summary><pre>

```json
{
 "geometry_file": "/home/alund/celeritas_project/regression/input/cms2018.gdml",
 "physics_options": {
  "coulomb_scattering": false,
  "eloss_fluctuation": true,
  "em_bins_per_decade": 56,
  "lpm": true,
  "msc": "urban",
  "physics": "em_basic",
  "rayleigh_scattering": true
 },
 "brem_combined": false,
 "primary_options": {
  "seed": 12345,
  "direction": {"distribution": "isotropic"},
  "energy": 10000,
  "num_events": 8,
  "pdg": 11,
  "position": [0, 0, 0],
  "primaries_per_event": 130
 },
 "initializer_capacity": 8388608,
 "num_track_slots": 65536,
 "field": [0.0, 0.0, 1.0],
 "step_diagnostic": true,
 "step_diagnostic_bins": 1000,
 "max_events": 1024,
 "secondary_stack_factor": 3.0,
 "seed": 20220905,
 "merge_events": false,
 "action_times": false,
 "use_device": false
}
```

</pre></details>

app/celer-sim/RunnerInput.cc
sethrj: Yes, `org.json` should be deprecated, but there's two reasons we keep it for now:
- I didn't fix the single-precision case for constructing ORANGE using in-memory Geant4, so our testing relies on that combination for now.
- If you recall we saw a huge performance hit for the manually-generated versus automatic TestEm3 representation: keeping this capability allows us to choose when to flip the switch on that change.

app/celer-sim/RunnerInput.cc
sethrj: Oh yeah! Good idea, I'll do that.

src/celeritas/setup/StandaloneInput.hh
sethrj: In celer-sim:
```c++
    inp::StandaloneInput si = to_input(old_inp);
    auto loaded = setup::standalone_input(si);
    core_params_ = std::move(loaded.core_params);
    CELER_ASSERT(core_params_);
    events_ = std::move(loaded.events);
```
but I'll add documentation here :)

src/celeritas/setup/Problem.cc
sethrj: This is to support the "flat" TestEM3 case in the regression tests. Previously it lived in celer-sim/Runner.cc; currently it still lives in `app/celer-g4/GlobalSetup.cc`, but it will be removed from that front end on the next PR.

amandalund: Ah right, I missed that it was in the code already.

app/celer-sim/RunnerInput.cc
amandalund: Makes sense, thanks. And we're dropping support for using GDML for the physics file here?

sethrj: As an explicit option, yes: it previously defaulted to using the geometry file. Now "GDML for physics" is automatic. I'll write out a couple of examples in the docs about the combinations of physics/geometry input.

sethrj:
I've gotten the difference to show up in this reduced input:
```json
{
    "geometry_file": "/home/s3j/celeritas-regression/input/cms2018.gdml",
    "physics_options": {
        "coulomb_scattering": false,
        "eloss_fluctuation": true,
        "em_bins_per_decade": 56,
        "lpm": true,
        "msc": "urban",
        "physics": "em_basic",
        "rayleigh_scattering": true
    },
    "brem_combined": false,
    "primary_options": {
        "seed": 12345,
        "direction": {
            "distribution": "isotropic"
        },
        "energy": 1000,
        "num_events": 8,
        "pdg": 11,
        "position": [
            0,
            0,
            0
        ],
        "primaries_per_event": 1
    },
    "initializer_capacity": 8388608,
    "num_track_slots": 65536,
    "field": [0.0, 0.0, 1.0],
    "step_diagnostic": true,
    "step_diagnostic_bins": 1000,
    "max_events": 1024,
    "max_steps": 256,
    "secondary_stack_factor": 3.0,
    "seed": 20220905,
    "merge_events": false,
    "action_times": false,
    "use_device": false
}
```
with `OMP_NUM_THREADS=8 CELER_DISABLE_DEVICE=1`. On the branch I see:
```
PropagationApplier.hh:163: debug: Track (pid=1, E=0.954094 MeV) is looping after 10 steps
TrackingCutExecutor.hh:68: debug: Killing track {"geo":{"dir":[-0.17697263856129397,-0.5682311795472658,0.8036130983196932],"is_on_boundary":false,"is_outside":false,"pos":[[65.9453854632571,-11.344627517630743,4.700573907719613],"cm"],"volume_id":"TOBModule0"},"particle":{"energy":[0.9394601819435245,"MeV"],"particle_id":"e-"},"sim":{"along_step_action":"along-step-uniform-msc","event_id":2,"num_looping_steps":10,"num_steps":16,"parent_id":0,"post_step_action":"tracking-cut","status":"alive","step_length":[14.73748896958745,"cm"],"time":[2.7713283081828317e-09,"s"],"track_id":12},"thread_id":8189,"track_slot_id":8189}: lost 0.93946 MeV
```
but not on develop... not sure if this is the ultimate cause or just a side effect of a different rng

EDIT: indeed, looks like I wasn't setting `p.tracking.limits.field_substeps`; develop uses `inp.field_options.max_substeps`.

sethrj:
@amandalund Results for your input are identical... the error was the looping counter not being set correctly.

sethrj:
@amandalund I confirmed the outputs are identical for your input. I think this is finally g2g.

amandalund (approved):
Excellent, thanks for debugging that!

PR #1603, opened 2025-01-30
Move physics lists out of details and rename tracking offload

sethrj:
Split from #1602, this renames some "detail" classes that we want to be available to users and other parts of Celeritas. It also drops the potentially confusing `Offload` part of the Celeritas tracking manager.

There's an inconsistency in our naming/paths now, that the `celeritas::FooPhysicsList` classes don't have a `g4` or `Geant`-qualified filename or directory, but you can't include them without having Geant4 enabled. However I think it's better to have `celeritas::Bar` or `celeritas::FooBar` (where `Bar` is a Geant4 virtual base class and/or `Foo` is the celeritas implementation) rather than `celeritas::GeantFooBar`. This is consistent with the new `celeritas::TrackingManager` rename. We may want to rename (or refactor to move) things for v1.0 (see https://github.com/celeritas-project/celeritas/issues/1263).

### DIFF
--- a/src/celeritas/ext/detail/PhysicsListUtils.hh
+++ b/src/celeritas/ext/detail/PhysicsListUtils.hh
@@ -6,11 +6,25 @@
 //---------------------------------------------------------------------------//
 #pragma once
 
+#include <G4VModularPhysicsList.hh>
+
 namespace celeritas
 {
 namespace detail
 {
 //---------------------------------------------------------------------------//
+/*!
+ * Create a suitably owned physics class and add it to the physics list.
+ */
+template<class PL, class... T>
+void emplace_physics(G4VModularPhysicsList& list, T&&... args)
+{
+    // Construct physics
+    auto phys = std::make_unique<PL>(std::forward<T>(args)...);
+
+    // Register, passing ownership to the list
+    list.RegisterPhysics(phys.release());
+}
 
 //---------------------------------------------------------------------------//
 }  // namespace detail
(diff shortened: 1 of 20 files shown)


src/celeritas/ext/detail/PhysicsListUtils.hh
```diff
@@ -0,0 +1,17 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/ext/detail/PhysicsListUtils.hh
```
amandalund, line 5: What's this guy for?

amandalund (approved):
Thanks @sethrj!

src/celeritas/ext/detail/PhysicsListUtils.hh
sethrj: Oops, forgot to finish one additional minor change 😳 will do that before merging.

sethrj: Take a look at https://github.com/celeritas-project/celeritas/pull/1603/commits/2478d253c007b2119cd9790f1461ab62774601ad and resolve this conversation if it looks good 😄 then this will merge cleanly.

amandalund: Looks great!

PR #1604, opened 2025-01-30
Implement discrete optical physics

hhollenb:
Implementation of `PhysicsData` and associated views necessary to do volumetric optical physics.

`PhysicsParams` closely matches the core physics version, but slimmed down just for discrete interactions. The `DiscreteSelectAction`, `PhysicsTrackView`, and `PhysicsStepView` were implemented similarly. I've updated the hooks into the core optical classes so executors should have access to physics data.

Currently only absorption and rayleigh models are implemented, and I've limited the importing of models in tests to just those two. 

I've left the executor implementation empty for the moment to reduce the overall complexity of the merge. I can start working on them and adding them to this PR. I can also add some extra device tests and I/O tests for optical data.


src/celeritas/optical/PhysicsData.hh
```diff
@@ -0,0 +1,175 @@
...
+    //! Offset to create an ActionId from a ModelId
+    ActionId::size_type model_to_action{};
+
+    //! Whether data is assigned and valid
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return num_models > 0 && model_to_action >= 1;
+    }
+
+    //! Indicate a discrete interaction was rejected by integral method
```
sethrj:
```suggestion
    //! Undergo a discrete interaction
```

src/celeritas/optical/PhysicsStepView.hh
```diff
@@ -0,0 +1,143 @@
...
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Access step-local (non-persistent) optical physics track data.
+ */
+class PhysicsStepView
```
sethrj: Since optical physics is so much simpler than full EM, I'm not sure we need a separate "step" view. Let's discuss at today's meeting?

src/celeritas/optical/PhysicsData.hh
```diff
@@ -0,0 +1,175 @@
...
+        this->grids = other.grids;
+        this->mfp_tables = other.mfp_tables;
+        this->reals = other.reals;
+        return *this;
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Physics state data for a single track.
```
sethrj: Nowadays we're preferring SOA rather than AOS, and I think it would be better not to blend the "persistent" (across steps) with the "temporary" (within a step) in the same data structure: can you replace this struct with two `StateItems<real_type`?.

src/celeritas/optical/PhysicsData.hh
```diff
@@ -0,0 +1,175 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+// TYPE ALIASES
+//---------------------------------------------------------------------------//
+
+using ValueGrid = GenericGridRecord;
+using ValueGridId = OpaqueId<ValueGrid>;
+using ValueTable = ItemRange<ValueGrid>;
+using ValueTableId = OpaqueId<ValueTable>;
```
sethrj: We should sync up with @amandalund about these. The requirements for optical will be much simpler (we'll just be using nonuniform grids I think?), so I don't think we'll even need them.

src/celeritas/optical/action/DiscreteSelectAction.hh
```diff
@@ -0,0 +1,41 @@
...
+#include "corecel/Macros.hh"
+
+#include "ActionInterface.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
```
sethrj, line 20: Need to add a line of documentation here and elsewhere...

src/celeritas/optical/model/AbsorptionExecutor.hh
```diff
@@ -0,0 +1,32 @@
...
+#include "../CoreTrackView.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ */
+struct AbsorptionExecutor
```
sethrj, line 18: Maybe we should move these to a subsequent PR?

test/celeritas/optical/Physics.test.cc
```diff
@@ -0,0 +1,534 @@
...
+    void step(CoreParams const&, CoreStateDevice&) const final {}
+};
+
+struct MockModelBuilder
+{
+    std::shared_ptr<Model> operator()(ActionId id) const
+    {
+        return std::make_shared<MockModel>(id);
+    }
+};
```
sethrj: It would make the file a bit easier to read if you move these classes to a separate file.

test/celeritas/optical/Physics.test.cc
```diff
@@ -0,0 +1,534 @@
...
+//---------------------------------------------------------------------------//
+// TEST HARNESS
+//---------------------------------------------------------------------------//
+
+class OpticalPhysicsTest : public OpticalMockTestBase
+{
+  protected:
+    using RngEngine = ::celeritas::test::DiagnosticRngEngine<std::mt19937>;
+
+    ModelId::size_type const num_models = 4;
```
sethrj: maybe `static constexpr`?

sethrj (commented):
Great work and very nice unit test! Sorry it took me so long to dive into it...

src/celeritas/optical/PhysicsData.hh
amandalund: Agreed, and yeah AFAIK only nonuniform grids.

src/celeritas/optical/PhysicsData.hh
```diff
@@ -0,0 +1,175 @@
...
+    template<class T>
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using StateItems = StateCollection<T, W, M>;
+    //!@}
+
+    //// Data ////
+
+    StateItems<PhysicsTrackState> states;
+    Items<real_type> per_model_xs;  //!< XS [track][model]
```
amandalund: Since there are only ever two or three models (processes) and the track doesn't lose energy along the step, we may want to consider the tradeoff of storing this extra state data vs. recalculating the cross section on the fly in `select_discrete_interaction`.

src/celeritas/optical/PhysicsStepView.hh
hhollenb: Along with Amanda's suggestion of not storing per model cross sections, the step view was even smaller so I just merged it into the track view. This does mean the track view needs to be a mutable ref in more places, but that's not an issue.

src/celeritas/optical/PhysicsData.hh
hhollenb: Took out the material tables, which works so long as we assume all the models have the same number of materials. For non-WLS materials we'll probably need to insert some dummy grids with infinite MFPs, which we'd likely need to do anyways unless the WLS materials are sorted somehow.

src/celeritas/optical/PhysicsData.hh
hhollenb: I think recalculating is a good initial implementation, as it's a bit simpler and has less state to manage.

hhollenb:
Managed to figure out the Geant versioning issue: in v10.6 and before, the optical photon PDG is 0 (updated to -22 in v10.7), so I added a macro to lookup optical photons by name in the particle table for older Geant versions.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -905,7 +904,42 @@ auto import_processes(GeantImporter::DataSelection::Flags process_flags,
             append_process(*g4_particle_def, process);
         }
     }
+
+// Optical photon PDG in Geant4 is 0 before version 10.7
+#if G4VERSION_NUMBER < 1070
```
sethrj: Does any of the code below fails to compile for older versions? If not, it would be better to include this in the C++ code (rather than the preprocessor):
```c++
if (G4VERS < 1070 && ...)
```

src/celeritas/optical/model/RayleighExecutor.hh
```diff
@@ -0,0 +1,45 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample optical Rayleigh interaction from the current track.
+ */
+CELER_FUNCTION Interaction
+RayleighExecutor::operator()(CoreTrackView const& track)
+{
+    auto particle = track.particle();
+    auto direction = track.geometry().dir();
```
amandalund:
```suggestion
    auto const& direction = track.geometry().dir();
```

src/celeritas/optical/InteractionApplier.hh
```diff
@@ -0,0 +1,120 @@
...
+#include "ParticleTrackView.hh"
+#include "SimTrackView.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Wrap an interation executor and apply it to a track.
```
amandalund:
```suggestion
 * Wrap an interaction executor and apply it to a track.
```

src/celeritas/optical/InteractionApplier.hh
```diff
@@ -0,0 +1,120 @@
...
+
+    CELER_FUNCTION void operator()(CoreTrackView const&);
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ *
+ * This class is partially specialized with a second template argument to
+ * extract any launch bounds from the functor class. TODO: we could probably
+ * inherit from a helper class to pull in those constants (if available).
```
amandalund: Are we going to want these specializations for optical (@sethrj)?

src/celeritas/optical/InteractionApplier.hh
```diff
@@ -0,0 +1,120 @@
...
+template<class F>
+CELER_FUNCTION void
+InteractionApplierBaseImpl<F>::operator()(CoreTrackView const& track)
+{
+    Interaction result = this->sample_interaction(track);
+
+    if (!result.changed())
+    {
+        // Currently we have no optical actions capable of failing. Can add a
+        // failure case in the future as necessary.
```
amandalund: Could this be an assertion?

src/celeritas/optical/PhysicsStepUtils.hh
```diff
@@ -0,0 +1,77 @@
...
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the discrete physics step limit for the given track.
+ *
+ * Cross sections for each model are calculated and cached in the
+ * \c PhysicsStepView scratch space. The total cross section is also
+ * calculated and cached.
```
amandalund:
```suggestion
 * The total cross section is cached in the \c PhysicsTrackView.
```

amandalund (approved):
Excellent work on this @hhollenb! Sorry for the delayed review, and thanks for continuing to drive the optical work forward!

src/celeritas/optical/InteractionApplier.hh
sethrj: We might as well. I think this forwarding function should probably be more generic though... refactoring it can be a follow-on change (@esseivaju )

src/celeritas/optical/PhysicsTrackView.hh
```diff
@@ -0,0 +1,273 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Get the MFP grid ID for the given model.
+ *
+ * The grid corresponds to the optical material this track view was
+ * constructed with.
+ */
+CELER_FUNCTION ValueGridId PhysicsTrackView::mfp_grid(ModelId model) const
+{
+    size_type num_materials = params_.grids.size() / this->num_models();
```
sethrj: I think we should store this as a separate integer to make our lives easier (and avoid integer division and/or accidental truncation)

sethrj (approved):
Fantastic! Apologies for the delay in review.

src/celeritas/optical/PhysicsTrackView.hh
hhollenb: I figured the hefty assertions on model count, grid size, and material count would probably make it safe, but I also am unsure of the relative speed of integer division vs looking up a single integer on the GPU.

sethrj: Agreed, it's just that integer division makes me wary. There's tradeoffs either way; `grids.size()` is looking up an integer too. I think it's better to have a little redundancy (and possibly an extra integer read) for clarity. Thanks for doing that!

PR #1605, opened 2025-01-31
Use external G4VG

sethrj:
Following https://github.com/celeritas-project/g4vg/pull/12 we can start using G4VG as an external package. I'll add it to spack and build it into the CI.

Waiting on https://github.com/celeritas-project/g4vg/pull/14 which we need for the tests to pass.


drbenmorgan (approved):
Looks good to me, so ready to merge once CI is happy.

PR #1607, opened 2025-02-03
Implement cubic spline interpolation

amandalund:
This implements cubic spline interpolation, which is the default spline interpolation in Geant4. This was motivated by wanting to avoid updating a bunch of test results when we construct our own physics tables (#1544) and because it could eliminate the interpolation method as the reason for any discrepancies when doing physics comparisons.

There are options for using natural or not-a-knot boundary conditions: the latter is supposedly what's used by Geant4, but the second derivatives calculated using Geant4's code don't seem to match the expected values, so for now there's also a hack to optionally calculate the derivatives by exactly matching what Geant4 does.

Currently this is just hooked up in the `XsCalculator`, which is all that's needed for the range grid calculation, but it would be easy to support this in the non-uniform calculator and to optionally calculate the derivatives for the physics grids (so we can do some studies comparing all our interpolation methods). With this we can construct a range table that exactly matches the imported values.


src/corecel/math/TridiagonalSolver.hh
```diff
@@ -0,0 +1,48 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Solve a tridiagonal system of equations using the Thomas algorithm.
+ *
+ * This is a simplified form of Gaussian elimination that can solve a
+ * tridiagonal system \f$ a_i x_{i - 1} + b_i x_i + c_i x_{i + 1} = d_i \f$
+ * with \f$ n \f$ unknowns where \f$ a_1 = 0 \f$ and \f$ c_n = 0 \f$ in O(n)
+ * time.
+ */
```
sethrj, line 24: Even thought it might be slightly more efficient to smash the LHS and RHS together, I think this class would be better written as "apply T^{-1}"; where T is the tridiagonal. Since I guess you want to "return" into the middle of an existing array I think we can make this class behave like `x = A \ b` in matlab, but returning by reference: `void  operator()(Span<real_type const> rhs, Span<real_type> x);` In theory if we need to do this to multiple `b` simultaneously we could overload for better efficiency.

src/celeritas/em/xs/MottRatioCalculator.hh
```diff
@@ -9,8 +9,8 @@
 #include "corecel/Macros.hh"
 #include "corecel/Types.hh"
 #include "corecel/math/ArrayUtils.hh"
+#include "corecel/math/PolyEvaluator.hh"
 #include "celeritas/em/data/WentzelOKVIData.hh"
-#include "celeritas/grid/PolyEvaluator.hh"
```
sethrj: Given the number of files this changes, would it be easy enough for you to turn this rename into a trivial PR?

src/celeritas/grid/SplineDerivCalculator.hh
```diff
@@ -0,0 +1,122 @@
...
+    using Values = detail::XsGridAccessor::Values;
+    using VecReal = std::vector<real_type>;
+    //!@}
+
+    //! Cubic spline interpolation boundary conditions
+    enum class BoundaryCondition
+    {
+        natural = 0,
+        not_a_knot,
+        not_not_a_knot,  //!< Geant4's "not-a-knot"
```
sethrj: 🤣

sethrj (commented):
Still working on this, sorry... and I regret not doing more background search before handing off this task to @lebuller  😞

src/celeritas/em/xs/MottRatioCalculator.hh
amandalund: No problem! If you want I could also split the tridiagonal solver into separate small PR.

sethrj: Only if you want more lines devoted to yourself in the release notes 😉

amandalund: Yes gimme more lines 🤑

sethrj: More lines, more 💰

src/celeritas/grid/SplineDerivCalculator.hh
sethrj: Seriously though "geant" might be a better description? (I might see if geant3 does the same thing, because this seems like the kind of implementation that would propagate from an ancient code)

amandalund: haha yeah just goofing around, I'll rename it.

src/corecel/grid/Interpolator.hh
```diff
@@ -84,6 +86,53 @@ class Interpolator
...
+ * \f{align}{
+   a_0 &= y_i \\
+   a_1 &= \frac{\Delta y_i}{\Delta x_i} - \frac{\Delta x_i}{6} \left[ S''_{i +
+          1} + 2 S''_{i} \right] \\
+   a_2 &= \frac{S''_i}{2} \\
+   a_3 &= \frac{1}{6 \Delta x_i} \left[ S''_{i + 1} - S''_i \right]
+ * \f}
+ */
+template<typename T = ::celeritas::real_type>
+class SplineInterpolator
```
sethrj: I think this class should be in its own file.

src/celeritas/grid/SplineDerivCalculator.hh
```diff
@@ -0,0 +1,122 @@
...
+
+  public:
+    // Construct with x and y grids and boundary type
+    SplineDerivCalculator(SpanConstReal, SpanConstReal, BoundaryCondition);
+
+    // Construct with grid data and boundary type
+    SplineDerivCalculator(XsGridData const&, Values const&, BoundaryCondition);
+
+    // Calculate the second derivatives
+    VecReal operator()() const;
```
sethrj: I feel like the interface here isn't very functor-y since it can only usefully be called once. What about keeping only the BC as "class data", and having the `operator()` do all the work (and it could even be templated rather than having a janky unique pointer)?

If you that's too much work then forget about it: I think we've already sunk enough time into the physics interpolation 😅

src/celeritas/grid/SplineDerivCalculator.cc
```diff
@@ -0,0 +1,257 @@
...
+        rhs[i - 1] = 6 * grid_->delta_slope(i);
+    }
+
+    // Calculate the last row coefficients using the boundary conditions
+    this->calc_final_coeffs(tridiag[num_knots - 3], rhs[num_knots - 3]);
+
+    // Solve the tridiagonal system
+    VecReal result(num_knots);
+    TridiagonalSolver(std::move(tridiag))(make_span(rhs),
+                                          {result.data() + 1, num_knots - 2});
```
sethrj:
```suggestion
                                          make_span(result).subspan(1, num_knots - 1));
```

src/celeritas/grid/SplineDerivCalculator.cc
```diff
@@ -0,0 +1,257 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the coefficients for the first row using the boundary conditions.
+ */
+void SplineDerivCalculator::calc_initial_coeffs(Real3& tridiag,
+                                                real_type& rhs) const
+{
+    CELER_EXPECT(grid_->x(1) > grid_->x(0));
```
sethrj: It might be better to validate in the grid constructors that it's monotonic increasing? (Or validate in `delta_x`).

src/celeritas/grid/XsCalculator.hh
```diff
@@ -129,24 +128,41 @@ CELER_FUNCTION real_type XsCalculator::operator()(Energy energy) const
...
     CELER_ASSERT(lower_idx + 1 < loge_grid_.size());
 
-    real_type const upper_energy = std::exp(loge_grid_[lower_idx + 1]);
-    real_type upper_xs = this->get(lower_idx + 1);
-    if (lower_idx + 1 == data_.prime_index)
+    real_type lower_energy = std::exp(loge_grid_[lower_idx]);
+    real_type upper_energy = std::exp(loge_grid_[lower_idx + 1]);
+
+    real_type result;
+    if (data_.derivative.empty())
```
sethrj, line 136: This class is getting increasingly complicated 😅 In practice do we use all four combinations of {1/E scaling, spline}? I would expect only xs uses the 1/E scaling and only dE/dx and/or range uses spline.

src/corecel/grid/Interpolator.hh
```diff
@@ -84,6 +86,53 @@ class Interpolator
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using real_type = T;
+    using Point = Array<T, 3>;
```
sethrj: Since this isn't R3, perhaps this should be a small `struct Spline { real_type x; real_type y; real_type y_dd; };`? (That lets you eliminate the local enum in `operator()` as well.)

src/corecel/grid/Interpolator.hh
```diff
@@ -135,5 +209,16 @@ Interpolator<XI, YI, T>::operator()(real_type x) const -> real_type
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Interpolate using a cubic spline.
+ */
+template<class T>
+CELER_FUNCTION auto SplineInterpolator<T>::operator()(real_type x) const
+    -> real_type
+{
+    return PolyEvaluator{a_}(x - x_lower_);
```
sethrj: Nice :)

src/corecel/math/TridiagonalSolver.cc
```diff
@@ -0,0 +1,52 @@
...
+    for (size_type i = 1; i < x.size(); ++i)
+    {
+        auto const& a = tridiag_[i];
+        real_type factor = 1 / (a[1] - a[0] * c_prime[i - 1]);
+        c_prime[i] = a[2] * factor;
+        x[i] = (rhs[i] - a[0] * x[i - 1]) * factor;
+    }
+
+    // Back substitution
+    for (size_type i = x.size() - 2; i != size_type(-1); --i)
```
sethrj: Oh this is unpleasant (relies on unsigned integer underflow)... what about letting `ip = i + 1`, so then your loop is `x[ip-1] -= c_prime[ip-1] * x[ip]` and you loop with:
```suggestion
    for (size_type ip = x.size() - 1; i > 0; --i)
```

test/celeritas/grid/CalculatorTestBase.cc
```diff
@@ -50,8 +51,13 @@ auto CalculatorTestBase::mutable_values() -> SpanReal
 /*!
  * Construct from an arbitrary function.
  */
-void CalculatorTestBase::build(Real2 bounds, size_type count, XsFunc calc_xs)
+void CalculatorTestBase::build(Real2 bounds,
+                               size_type count,
+                               XsFunc calc_xs,
+                               bool spline)
```
sethrj: Instead of `bool spline` (very difficult to tell what's `true` in the caller) how about making this an implementation function

```suggestion
void CalculatorTestBase::build_impl(Real2 bounds,
                               size_type count,
                               XsFunc calc_xs,
                               SplineDerivCalculator::BoundaryCondition bc)
```
and build the spline data if `bc != SplineDerivCalculator::BoundaryCondition::size_`, then adding two inline functions in the header:

```c++
void build(Real2 bounds,
                               size_type count,
                               XsFunc calc_xs)
{
return this->build(bounds, count, std::move(calc_xs), BC::size_);
}
void build_spline(Real2 bounds,
                               size_type count,
                               XsFunc calc_xs,
BC bc)
{
CELER_EXPECT(bc != BC::size_);
return this->build(bounds, count, std::move(calc_xs), bc);
}
```

test/celeritas/grid/CubicSpline.test.cc
```diff
@@ -0,0 +1,370 @@
...
+            = SplineDerivCalculator(make_span(x), make_span(y), BC::geant)();
+        EXPECT_VEC_SOFT_EQ(expected_result, result);
+
+        SplineInterpolator interpolate({x[0], y[0], result[0]},
+                                       {x[1], y[1], result[1]});
+        EXPECT_SOFT_EQ(834.59530552174078, interpolate(1.1e-4));
+        EXPECT_SOFT_EQ(646.15145372907716, interpolate(5e-4));
+    }
+}
+
```
sethrj: I'd break the test in half here: before this is just `SplineDerivCalculator` (excellent tests by the way); and after you're testing the XS calculator. You could move those tests into the XsCalculator test itself.

src/celeritas/grid/XsCalculator.hh
amandalund: I think in practice we'll have xs with all combinations of with/without scaling/spline (only a few have scaling, most have spline) and then range and energy loss without scaling and with spline. What I have here currently isn't correct for the xs with scaling and spline case because it both calculates the derivatives and interpolates using the unscaled values. Maybe for this PR I'll just assert that there's no 1/E scaling instead when calculating the derivatives.

I've been wanting to do some refactoring/renaming of the grids and calculators. I think instead of forcing the scaled xs into a single grid we could use separate grids for the scaled and unscaled values, which would fix the scaling+spline issue as well as the inconsistency in the Rayleigh grid.

src/celeritas/grid/SplineDerivCalculator.hh
amandalund: Good idea, that's no problem... the sunk time here has really just been trying to figure out what Geant4 is doing 😔

src/celeritas/grid/XsCalculator.hh
sethrj: Yes, indeed that would fix it. I think my decision to force them onto the same grid was premature optimization 😅

src/celeritas/grid/SplineDerivCalculator.hh
```diff
@@ -0,0 +1,123 @@
...
+
+    //// HELPER FUNCTIONS ////
+
+    template<class GridAccessor>
+    VecReal operator()(GridAccessor&&) const;
+
+    template<class GridAccessor>
+    void calc_initial_coeffs(GridAccessor const&, Real3&, real_type&) const;
+    template<class GridAccessor>
+    void calc_final_coeffs(GridAccessor const&, Real3&, real_type&) const;
```
amandalund, line 115: I could also just inline these...

sethrj (approved):
Thanks @amandalund ! Brutal work trying to understand what Geant4's spline does. Incidentally I wonder how yours and Geant4's compare to [G4HepEm's](https://github.com/mnovak42/g4hepem/blob/master/G4HepEm/G4HepEmInit/src/G4HepEmInitUtils.cc) ...

src/celeritas/grid/SplineDerivCalculator.hh
sethrj: Eh let's just call it a day😅

amandalund:
Looks like G4HepEm's is pretty much a copy-paste of [Geant4](https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/global/management/src/G4PhysicsVector.cc#L292)'s. Hard to say if they're doing something different intentionally or by accident...

PR #1606, opened 2025-02-01
Add TrackingManagerConstructor and SharedParams::GetMode

sethrj:
This reworks two behind-the-scenes elements for constructing Celeritas. The first is the addition of `TrackingManagerConstructor`, which is a `G4VPhysicsConstructor` that sets up the `TrackingManager` on each thread based on a hardcoded set of particle types.

Additionally this refactors the `SharedParams` code so that we always construct SharedParams when integrating with Celeritas, even if `CELER_DISABLE` or `CELER_KILL_OFFLOAD` are set.  This allows us to process the setup options and environment variables during construction rather than not initializing the shared params, and in the future will let us specify execution/config options in the `inp` data structures rather than environment. It also lets us output configuration/environment data to the `.out.json` file even when Celeritas offloading is disabled.

I added testing to the examples so that we ensure all configurations of cpu/gpu/g4/kill-offload works. Previous to this PR the tracking manager test failed because we'd build the tracking manager even when doing a G4-only mode, so I had to wrap the tracking offload construction inside `if (!shared_params.CeleritasDisabled())`.

However, this disagrees with the new implementation because we don't process the "disabled" status until we build the SharedParams (during run initialization), and we ought to query the constructed shared params to know what particles to kill/offload (based on user input), using `SharedParams::OffloadParticles` rather than a hard-coded list. The only solution I have is to build the tracking manager during `BeginningOfRunAction` rather than the physics initialization. This mirrors the SimpleOffload behavior but isn't quite right...

The difference in log messages for a 1-worker MT run show that the physics table prep is no longer performed:
```diff
--- orig.err	2025-02-01 12:28:49
+++ new.err	2025-02-01 12:30:16
@@ -1,22 +1,8 @@
-trackingmanager-offload.cc:226: status: Constructing user actions
+trackingmanager-offload.cc:222: status: Constructing user actions
-trackingmanager-offload.cc:110: status: Setting up detector
+trackingmanager-offload.cc:109: status: Setting up detector
-trackingmanager-offload.cc:141: status: Setting up tracking manager offload
-TrackingManager.cc:95: debug: Preparing physics table for e+
-TrackingManager.cc:95: debug: Preparing physics table for e-
-TrackingManager.cc:95: debug: Preparing physics table for gamma
-TrackingManager.cc:54: debug: Building physics table for gamma
-TrackingManager.cc:54: debug: Building physics table for e-
-TrackingManager.cc:54: debug: Building physics table for e+
[...]

Issue #1600, opened 2025-01-28
Geant4 benchmark integration


sethrj:
Based on discussions today with @stognini @amandalund @whokion and @pcanal 

- Split  SimpleOffload into ManagerIntegration/ActionIntegration classes: the first (preferred, requires G4 11.1) uses the G4TrackingManager and doesn't need begin/end events; the second is for older versions that uses a tracking action.
- Updated Integration classes (only one can be used at a time) will own the global and thread-local objects that previously were boilerplate; and it will provide safe access to the setup options during setup.
- Eventually need to add region/volume-based field description (map to `AlongStepId`)
- Individual integration: see https://github.com/celeritas-project/gpu-delta-review-2025/issues (private repo for now)

PR #1609, opened 2025-02-03
Add ultra-simple tracking manager integration

sethrj:
Follow-on to #1606; see #1600. This defines static singleton classes used for integrating with external applications. It carefully checks usage patterns for single and MT use cases. Currently only the "tracking manager" version is implemented; a "user action" one will follow for older versions of Geant4.


example/offload-template/src/DetectorConstruction.cc
```diff
@@ -14,14 +14,25 @@
...
 #include <G4VPhysicalVolume.hh>
+#include <accel/AlongStepFactory.hh>
+#include <accel/SetupOptions.hh>
+#include <accel/TrackingManagerIntegration.hh>
 
 #include "SensitiveDetector.hh"
 
 //---------------------------------------------------------------------------//
 /*!
  * Construct empty.
```
stognini: Construct and setup Celeritas field option.

example/offload-template/src/ActionInitialization.cc
```diff
@@ -6,21 +6,30 @@
...
-#include "EventAction.hh"
-#include "G4Electron.hh"
-#include "G4Gamma.hh"
-#include "G4Positron.hh"
 #include "PrimaryGeneratorAction.hh"
 #include "RunAction.hh"
 
 //---------------------------------------------------------------------------//
 /*!
  * Construct empty.
```
stognini: Setup Celeritas simulation, physics, and diagnostics.

example/offload-template/README.rst
```diff
@@ -28,29 +28,14 @@ Build and run
...
 :code:`G4UserRunAction`
   :code:`BeginOfRunAction` initializes Celeritas global shared data on master
-  and worker threads. :code:`EndOfRunAction` clears data and finalizes
-  Celeritas data.
-
-:code:`G4UserEventAction`
-  :code:`BeginOfEventAction` initializes an event in Celeritas, and
-  :code:`EndOfEventAction` flushes remaining particles.
+  and worker threads, setting up a tracking manager under the hood.
+  :code:`EndOfRunAction` clears data and finalizes Celeritas data.
```
stognini, line 38: Maybe add a short description for :code:`DetectorConstruction()`, since we need it too now.

src/accel/detail/IntegrationSingleton.hh
```diff
@@ -0,0 +1,89 @@
...
+    //! Static global setup options before or after constructing params
+    SetupOptions const& setup_options() const { return options_; }
+
+    //! Whether Celeritas is enabled
+    Mode mode() const { return mode_; }
+
+    //!@{
+    //! Static global Celeritas problem data
+    SharedParams& shared_params() { return params_; }
+    SharedParams const& shared_params() const { return params_; }
```
stognini, line 56: Out of curiosity, what happens here if one calls `auto IntegrationSingleton::instance().shared_params()`? Will the compiler complain since you did not explicitly define its constness?

example/offload-template/src/RunAction.cc
```diff
@@ -6,8 +6,9 @@
 //---------------------------------------------------------------------------//
 #include "RunAction.hh"
 
-#include "Celeritas.hh"
-#include "G4Threading.hh"
+#include <G4ParticleDefinition.hh>
```
stognini: Trailing include?

src/accel/detail/IntegrationSingleton.hh
sethrj: No, `instance()` is mutable so the mutable version will be returned.

example/offload-template/README.rst
sethrj: I'm going to move the whole celeritas setup to the `main` part.

src/accel/LocalTransporter.cc
```diff
@@ -242,6 +243,27 @@ void LocalTransporter::Flush()
...
+            event_manager_ = G4EventManager::GetEventManager();
+            CELER_ASSERT(event_manager_);
+        }
+
+        G4Event const* event = event_manager_->GetConstCurrentEvent();
+        CELER_ASSERT(event);
+        if (event_id_ != id_cast<UniqueEventId>(event->GetEventID()))
+        {
+            // The event ID has changed: reseed it
+            this->InitializeEvent(event->GetEventID());
```
stognini, line 262: Maybe we should move `InitializeEvent` to `private`.

sethrj: No, we still need it for user action integration including celer-g4.

src/accel/detail/IntegrationSingleton.cc
```diff
@@ -0,0 +1,231 @@
...
+ *   \c G4Threading::IsWorkerThread()==true. For MT mode, the master thread
+ *   does not track any particles. For single-thread mode, the master thread
+ *   \em does do work.
+ */
+void IntegrationSingleton::initialize_shared_params()
+{
+    ExceptionConverter call_g4exception{"celer.init.global"};
+
+    if (!G4Threading::IsMultithreadedApplication()
+        || G4Threading::IsMasterThread())
```
amandalund: Based on the note above, is `if (!G4Threading::IsMultithreadedApplication())` needed here, or can we just check if it's the master thread?

src/accel/TrackingManagerIntegration.cc
```diff
@@ -0,0 +1,136 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize during ActionInitialization on non-worker thread in MT mode.
+ */
+void TrackingManagerIntegration::BuildForMaster()
+{
+    CELER_VALIDATE(
+        G4Threading::IsMasterThread()
+            || G4Threading::IsMultithreadedApplication(),
```
amandalund: Should this be
```suggestion
        G4Threading::IsMasterThread()
            && G4Threading::IsMultithreadedApplication(),
```
?

src/accel/TrackingManagerIntegration.cc
```diff
@@ -0,0 +1,136 @@
...
+/*!
+ * Initialize during ActionInitialization.
+ */
+void TrackingManagerIntegration::Build()
+{
+    if (G4Threading::IsMasterThread())
+    {
+        CELER_VALIDATE(!G4Threading::IsMultithreadedApplication(),
+                       << "cannot call Integration::Build from worker thread "
+                          "in a multithreaded application");
```
amandalund, line 171: Is this to protect against a user calling this from `ActionInitialization::BuildForMaster()`?

Maybe we could add a comment on why this is only called in serial mode (logger should only be initialized by the master thread?).

src/accel/detail/IntegrationSingleton.cc
```diff
@@ -0,0 +1,231 @@
...
+        CELER_LOG_LOCAL(debug)
+            << R"(Skipping state construction since Celeritas is completely disabled)";
+        return false;
+    }
+
+    if (params_.mode() == celeritas::SharedParams::Mode::kill_offload)
+    {
+        // When "kill offload", we still need to intercept tracks
+        CELER_LOG_LOCAL(debug)
+            << R"(Skipping state construction with offload enabled)";
```
amandalund: Maybe make it clear in the log message that nothing is being offloaded in this case?

src/accel/TrackingManagerConstructor.cc
```diff
@@ -49,6 +52,24 @@ TrackingManagerConstructor::TrackingManagerConstructor(
     this->SetPhysicsType(G4BuilderType::bUnknown);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from tracking manager integration.
+ *
+ * Since there's only ever one trackign manager integration, we can just use
```
amandalund:
```suggestion
 * Since there's only ever one tracking manager integration, we can just use
```

amandalund (approved):
Thanks for putting this together @sethrj!

src/accel/detail/IntegrationSingleton.cc
sethrj: Good catch!

src/accel/TrackingManagerIntegration.cc
sethrj: Yes indeed. I mess up the logic for these so much because in serial, `Threading::IsMaster` is true but `BuildForMaster` is never called 🤪  they really fucked up the naming there; it should've been `BuildForNotWorker`.

doc/implementation/geant4-interface.rst
```diff
@@ -17,11 +17,21 @@ advanced implementation can be inspected in the :ref:`celer-g4` app.
 High-level interface
 --------------------
 
-The :cpp:class:`TrackingManagerIntegration` provides easy-to-use methods for
-using Celeritas as a tracking manager to offload electrons, photons, and gammas
-to Celeritas. See :ref:`example_template` for the boilerplate code
+Using Celeritas to "offload" all electrons, photons, and gammas from Geant4 can
+be done using the new-ish Geant4 interface :cpp:class:`G4VTrackingManager`
+implemented by :cpp:class:`celeritas::TrackingManager`. To set up the tracking
+manager correctly, we recommend using this helper class class:
```
amandalund:
```suggestion
manager correctly, we recommend using this helper class:
```

sethrj:
From @whokion at https://github.com/celeritas-project/celeritas/pull/1615#issuecomment-2654379240:
> One more minor comment on TrackingManagerConstructor: this class name is confusing as it sounds like constructing a TrackingManager rather than a PhysicsConstructor. A concrete class of G4VPhysicsConstructor is meant to be added to a modular physics list, so its name may be better to contain "Physics" (this is also a common practice in all Geant4 physics constructors). So, suggested name may be OffloadEmPhysics or something similar. Also, the OffloadParticles still be included in ConstructParticle() rather than empty even though they are overloaded somewhere (where?) without any side effect, which clarifies what particles are specifically involved for this physics constructor. On the other hand, this is still purely a semantic argument.

PR #1610, opened 2025-02-04
Move `PolyEvaluator` to corecel/math

amandalund:
Pulled out of #1607: this moves `PolyEvaluator` from celeritas/grid to corecel/math.


sethrj (approved):
Thanks!

PR #1615, opened 2025-02-07
Add high-level user/fast sim integration helpers

sethrj:
This completes the first bullet on #1600 . It deprecates `SimpleOffload` and adds new integration classes for UserAction and FastSimulation . I'm using inheritance for implementation rather than interface, but it substantially cuts down the boilerplate at the expense of making the class harder to read (you need to open two classes to get the full interface).


whokion:
@sethrj A quick test with `fastsim-offload `of this branch (built with Geant4 11.3) has an exception with

```
info: Activating Celeritas version 0.5.0-153+f2c0e17d on CPU

*** G4Exception : celer.init.global
      issued by : accel/SharedParams.cc:471
Celeritas runtime error: along-step action factory 'make_along_step' was not defined in the celeritas::SetupOptions

```
even though it seems that `make_along_step` is set at [this line](https://github.com/sethrj/celeritas/blob/user-action-integration/example/accel/fastsim-offload.cc#L176).
I also had to change the version of Celeritas from 0.6 to 0.5, 
`find_package(Celeritas 0.6 REQUIRED)`
in CMakeLists.txt (as it has not been tagged to v.06 yet, I guess).  Other two examples under `accel` run fine.  Anyway, excellent works to support all uses cases!

sethrj:
Thanks @whokion , I'll see if I can reproduce that odd failure 🤔 

Regarding the version, perhaps you need to run `git fetch upstream --tags`?

whokion:
Okay, the version problem is resolved by `git fetch upstream --tags`.  Regarding to the crash, it seems that MakeOptions() is never set to  FastSimulationIntegration::Instance().SetOptions(MakeOptions()) anywhere (?).

sethrj:
Great catch @whokion , I had accidentally disabled execution of the `fastsim` example in our CI. I'm adding additional checking to ensure that the options are set before beginning.

amandalund (approved):
Looks like fastsim might be trying to flush without Celeritas enabled? Looks great on my end, thanks @sethrj!

whokion:
>  I'm adding additional checking to ensure that the options are set before beginning.

Thanks for quick updating, which fixes the crash!

whokion:
@sethrj One more minor comment on `TrackingManagerConstructor`: this class name is confusing as it sounds like constructing a TrackingManager rather than a PhysicsConstructor.  A concrete class of G4VPhysicsConstructor is meant to be added to a modular physics list, so its name may be better to contain "Physics" (this is also a common practice in all Geant4 physics constructors).  So, suggested name may be `OffloadEmPhysics` or something similar.  Also, the OffloadParticles still be included in `ConstructParticle()` rather than empty even though they are overloaded somewhere (where?) without any side effect, which clarifies what particles are specifically involved for this physics constructor. On the other hand, this is still purely a semantic argument.

whokion:
One more minor comment (sorry that I have not followed this MR from the beginning, so do not see all updated changes in "Files Changed"):  In the main of three examples under example/acc, the default particle of PrimaryGeneratorAction is set to 100 MeV neutron, which will not produce any meaningful EM or hadronic secondaries as 100 MeV is below the pion production threshold, and the neutron mainly undergoes elastic scatterings and only produces very soft gammas during the deexcitation process even if there is (rare) an inelastic inter-nuclear interaction or a neutron capture process.  So, it would be better to set to something that can generate secondary hadrons and then produce EM particles from either inelastic or the ${\pi}_o$ production. Anyway, 1 GeV (or higher energy) pion-/+ may be a good default.

sethrj:
@whokion Good point. I dialed down the energy to reduce the compute time since it's just a silly example, but maybe I'll just make the world smaller to let some of the energy escape.

PR #1616, opened 2025-02-09
Rename `Generic` grid to `Nonuniform`

amandalund:
This updates the name of the nonuniform grid data and associated classes.


sethrj (approved):
Thanks!

PR #1617, opened 2025-02-11
Add G4Ellipsoid without bottom/top cuts

elliottbiondo:
This small MR adds supports for `G4Ellipsoids` in the case where there is no top/bottom cut. `G4Ellipsoids` and `G4EllipticalCones` can be truncated by top/bottom z-planes, instead of being truncated by an azimuthal wedge like other solids. I can certainly add support for these cuts, but I figured I'd punt until we can formulate a plan that will work for other solids with additional arguments as well. For example `G4CutTubs` can be truncated both with an azimuthal wedge and non-axis-aligned top/bottom planes.


elliottbiondo:
@sethrj RTR

sethrj (approved):
Nice!

PR #1618, opened 2025-02-11
Fix CUDA+VecGeom linking with Ubuntu

pcanal:
This ensures that the final library is not dropped by a 'as-needed' linker config, which will solve the problem seen on Ubuntu at https://github.com/spack/spack/pull/48528#issuecomment-2624468780

### DIFF
--- a/cmake/CudaRdcUtils.cmake
+++ b/cmake/CudaRdcUtils.cmake
@@ -1,5 +1,6 @@
-#------------------------------- -*- cmake -*- -------------------------------#
-# Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+#----------------------------------*-CMake-*----------------------------------#
+# Copyright 2020-2024 UT-Battelle, LLC, and other Celeritas developers.
+# See the top-level COPYRIGHT file for details.
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 #[=======================================================================[.rst:
 
@@ -11,8 +12,8 @@ relocatable device code and most importantly linking against those libraries.
 
 .. command:: cuda_rdc_add_library
 
-  Add a library to the project using the specified source files *with* special handling
-  for the case where the library contains CUDA relocatable device code.
+  Add a library to the project using the specified source files *with* special
+  handling for the case where the library contains CUDA relocatable device code.
 
   ::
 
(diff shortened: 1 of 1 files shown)


cmake/CudaRdcUtils.cmake
```diff
@@ -141,6 +147,21 @@ define_property(TARGET PROPERTY CUDA_RDC_OBJECT_LIBRARY
...
+  # (we are looking at you Ubuntu) but also in case user as explicitly using it.
+  set(CMAKE_CXX_LINK_LIBRARY_USING_rdc_no_as_needed_SUPPORTED TRUE CACHE INTERNAL "")
+  set(CMAKE_CXX_LINK_LIBRARY_USING_rdc_no_as_needed
+    "LINKER:--push-state,--no-as-needed"
+    "<LINK_ITEM>"
+    "LINKER:--pop-state"
+    CACHE INTERNAL ""
+  )
+else()
+	 message(ERROR "Got ${CMAKE_C_COMPILER_ID} and ${CMAKE_SYSTEM_NAME}")
```
wdconinc:
```suggestion
  message(ERROR "Got ${CMAKE_C_COMPILER_ID} and ${CMAKE_SYSTEM_NAME}")
```

pcanal: sigh this was not meant to be up there :(  Sorry.

cmake/CudaRdcUtils.cmake
```diff
@@ -1,5 +1,6 @@
-#------------------------------- -*- cmake -*- -------------------------------#
-# Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+#----------------------------------*-CMake-*----------------------------------#
+# Copyright 2020-2024 UT-Battelle, LLC, and other Celeritas developers.
+# See the top-level COPYRIGHT file for details.
```
sethrj: Please update the "source" RdcUtils with the updated template and revert this

cmake/CudaRdcUtils.cmake
```diff
@@ -141,6 +147,20 @@ define_property(TARGET PROPERTY CUDA_RDC_OBJECT_LIBRARY
   FULL_DOCS "Name of the object (without nvlink step) library corresponding to this cuda library"
 )
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  # This is necessary on platform that default to calling ld with --as-needed
+  # (we are looking at you Ubuntu) but also in case user as explicitly using it.
+  set(CMAKE_CXX_LINK_LIBRARY_USING_rdc_no_as_needed_SUPPORTED TRUE CACHE INTERNAL "")
+  set(CMAKE_CXX_LINK_LIBRARY_USING_rdc_no_as_needed
```
sethrj, line 170: Can you add documentation linking to https://cmake.org/cmake/help/latest/variable/CMAKE_LINK_LIBRARY_USING_FEATURE.html ?This also bumps the minimum CMake version for linux+cuda+vecgeom to 3.24.

sethrj (approved):

cmake/CudaRdcUtils.cmake
pcanal: Somehow the version in VecGeom was uploaded with white space changes that were not propagated to Celeritas nor the 'source' (CudaRDCExercise) and in the diff length I missed that actual change :(

cmake/CudaRdcUtils.cmake
pcanal: Fair.  Note that we had already broken with CMake v3.18 with:
```
CMake Error at cmake/CudaRdcUtils.cmake:951 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "CUDA_RDC_CACHED_LIB_DEPENDENCIES" is not allowed.
```
I think if useful, I can make it go with:
* Use only if the new code for CMake >= v3.24
* Fail explicitly on Ubuntu if CMake < v3.24

Should I do that or increase the CMake requirement (I will already increase it to v3.19)

sethrj: Um honestly I think the best thing would be a "cmake minimum" at the top of the file. Since:
> CMake also manages a new entry for scripts loaded by [include()](https://cmake.org/cmake/help/latest/command/include.html#command:include) and [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) commands except when invoked with the NO_POLICY_SCOPE option (see also policy [CMP0011](https://cmake.org/cmake/help/latest/policy/CMP0011.html#policy:CMP0011))

then we could probably just put `cmake_policy(VERSION 3.24...3.31)` at the top of the file?

cmake/CudaRdcUtils.cmake
pcanal: Reverting would work for the version in Celeritas but not for the other 2 copies which might not have have the right copyright nor license files ...

pcanal: I uploaded an update proposal for the copyright lines.

cmake/CudaRdcUtils.cmake
```diff
@@ -1,7 +1,7 @@
 #------------------------------- -*- cmake -*- -------------------------------#
-# Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+# Copyright Celeritas contributors, for more details see:
+#   https://github.com/celeritas-project/celeritas/blob/develop/COPYRIGHT
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
-#[=======================================================================[.rst:
```
sethrj, line 4: This line needs to be reintroduced; it makes the following text a comment.

cmake/CudaRdcUtils.cmake
sethrj: Updated copyright looks good to me, thanks.

cmake/CudaRdcUtils.cmake
pcanal: > then we could probably just put cmake_policy(VERSION 3.24...3.31) at the top of the file?

I will add it and it will fail with:

```
CMake Error at cmake/CudaRdcUtils.cmake:127 (cmake_policy):
  An attempt was made to set the policy version of CMake to "3.24" which is
  greater than this version of CMake.  This is not allowed because the
  greater version may have new policies not known to this CMake.  You may
  need a newer CMake version to build this project.
```

Also of notes:
* On Ubuntu 22, the default CMake version is 3.22.1
* On Ubuntu 24 it is 3.28.3
* On the original Spack/Ubuntu 22 failing case, the CMake version is 3.31.4

sethrj: I think that's a decent failure mode.

PR #1620, opened 2025-02-11
Remove pointers from Geant4 labels

sethrj:
This eliminates the pointer-based uniquification of Geant4 volumes and other objects, instead constructing reproducible unique integers for duplicate-named volumes by their traversal order in the volume hierarchy. This is reproducible so long as daughters are inserted in a consistent order. Updating this is a prereq to fixing https://github.com/celeritas-project/g4vg/issues/16 .


src/orange/g4org/Volume.hh
```diff
@@ -55,8 +56,8 @@ struct LogicalVolume
     //! Associated Geant4 logical volume
     G4LogicalVolume const* g4lv{nullptr};
 
-    //! Logical volume name
-    std::string name;
+    //! Logical volume name and optional uniqifying extension
```
mrguilima:
```suggestion
    //! Logical volume name and optional uniquifying extension
```

mrguilima (approved):
Looks good to me.
Thanks @sethrj

src/orange/g4org/PhysicalVolumeConverter.cc
```diff
@@ -117,7 +123,21 @@ PhysicalVolumeConverter::Builder::make_pv(int depth,
...
+        }
+        if (result.empty())
+        {
+            result.name = g4pv.GetName();
+            CELER_LOG(warning) << "Physical volume '" << result.name
+                               << "' is not in the world geometry";
+        }
+        return result;
+    }();
+    result.label = g4pv.GetName();
```
stognini: Why do we have 2 `result.label =`?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -1272,26 +1270,22 @@ import_volumes(G4VPhysicalVolume const& world, bool unique_volumes)
             {
                 volume.phys_material_id = cuts->GetIndex();
             }
-            volume.name = lv.GetName();
+            // TODO: when changing to celeritas::inp, just make this a label
+            // instead of converting to and from a std::string
+            volume.name = to_string(labels[lv.GetInstanceID()]);
```
stognini, line 1275: Should we just move from `volume.name` to `volume.label` at this point?

src/orange/g4org/PhysicalVolumeConverter.cc
sethrj: Oops! Good catch.

src/celeritas/ext/GeantImporter.cc
sethrj: We don't wrap `Label` in our ROOT interface, so we can't, right?

stognini: Correct, we'd end up adding `celeritas::Label` to the root dictionary and updating things accordingly. This ties up with a discussion a while ago about the input structure and less indirections, such as directly storing opaque ids instead of ints, for example. I also see this quickly becoming a rabbit hole, so happy to defer this to a future discussion and PR whose focus is to improve our root I/O library.

sethrj: Yeah let's postpone that until the refactor.

PR #1621, opened 2025-02-11
Rename `EXTERNAL` CMake variable

sethrj:
It makes sense from a repository standpoint to have an `external` directory for packages "external" to Celeritas. However, the variable `CELERITAS_EXTERNAL_Foo` makes it sound like setting it to "true" will look for it external to the build system, but in fact the opposite is happening. I'm renaming it from `EXTERNAL` to `BUILTIN`, so `CELERITAS_BUILTIN_G4VG` clearly indicates it's using a Celeritas-vendored copy rather than a system/toolchain/user installation outside the build system.

The `external` top-level directory *could* one day include Celeritas-provided "external" packages/modules but currently only has a CMakeLists: do we want to change it?


pcanal (approved):
LGTM.

Issue #588, opened 2023-01-04
Support nonuniform energy grid present in Geant4 11.1


sethrj:
In `geant4@10.7:11.0`, EM cross sections are stored on two separate physics tables that share the same log-energy spacing, but the second table stores $E_i \sigma_i$ rather than $\sigma_i$. Unfortunately `geant4@11.1` adjusts the lower energy grid (I think based on the energy cutoffs), so that the grid spacing is no longer the same for the two tables. Our assertion does catch this in debug mode:
```
/home/s3j/.local/src/celeritas/app/demo-loop/demo-loop.cc:244: critical: While running input at <stdin>: /home/s3j/.local/src/celeritas/src/celeritas/grid/ValueGridBuilder.cc:111:
celeritas: precondition failed: has_same_log_spacing(lambda_energy, lambda_prim_energy)
```

We will have to either change the xs grid to support two different spacings (which should probably be rolled into the refactor from #253) or completely recalculate the cross sections ourselves (which is what G4HepEm does).

sethrj:
This is due to Geant4 change set `phys-ctor-em-V11-00-26: fixed gamma general at low energy` (V Ivanchenko):

> ## 2022-07-06 V. Ivanchenko (emlowen-V11-00-08)
> - G4Rayleigh scattering moved boundary between low and high energy
>     table from 100 keV to 150 keV to be coherent with G4GammaGeneralProcess

And `G4GammaGeneralProcess` has a "low photoelectric" boundary of 150 keV.

Here are the cross sections compared from Geant4@11.0 and @11.1:
![nonuniform-xs](https://github.com/celeritas-project/celeritas/assets/741229/2298e074-561f-4181-859c-ccc7fa5745f7)

So we should be able to work around this at least by turning off Rayleigh scattering?

sethrj:
Honestly I'm surprised this is using a uniform log energy grid at all; that bump that looks like a possible resonance is indeed a resonance:
![nonuniform-xs](https://github.com/celeritas-project/celeritas/assets/741229/3de90011-8681-4762-8276-dfad330ed441)

PR #1622, opened 2025-02-12
Support different grid spacings for cross section data

### LINKED ISSUE #1091: Support nonuniform Rayleigh scattering cross section grid (Geant4 11.2)
An error message occurs related to energy inconsistencies/format in Celeritas when I run atlas-tilecal-integration, it seems that the max/min energies of what is used in ValuedGridBuilder are not equal and breaks the simulation. Here is the excerpt of the error message:
```
Some /vis commands (optionally) take a string to specify colour.
"/vis/list" to see available colours.
### Adding tracking cuts for neutron  TimeCut(ns)= 10000  KinEnergyCut(MeV)= 0
G4WT1 > Using
G4WT0 > Using
G4VisManager: Using G4TrajectoryDrawByCharge as fallback trajectory model.
See commands in /vis/modeling/trajectories/ for other options.
### Run 0 starts.
SharedParams.cc:212: status: Initializing Celeritas shared data
status: Loading external elemental data
/jet/home/amirreza/celeritas-GPU/celeritas/src/celeritas/phys/ProcessBuilder.cc:234: warning: Omitting coulomb_scat from physics process list
ExceptionConverter.cc:152: critical: The following error is from: Particle ID=0, process 'rayleigh, material ID=0

*** G4Exception : celer0001
      issued by : /jet/home/amirreza/celeritas-GPU/celeritas/src/celeritas/grid/ValueGridBuilder.cc:118
Lower and upper energy grids have inconsistent spacing: log delta E for lower grid is 1.39434 log(MeV) per bin but upper is 1.38778

*** G4Exception: Aborting execution ***
srun: error: v023: task 0: Aborted (core dumped)
```

I tried different Geant4 11.2 and 11.0.3 with physics list FTFP_BERT. FYI, I can not install Celeritas properly with the Geant4.10.7.P03 as it is suggested in the atlas-tilecal-integration. The problem seems to be simple to fix as it seems to be format inconsistency between Geant4 codes, I appreciate if you take a look and keep me updated.
Thanks!

amandalund:
This refactors the grid classes to support two different grid spacings for cross section data, which fixes the error from the nonuniform Rayleigh grid (#588, Fixes #1091) and addresses https://github.com/celeritas-project/celeritas/pull/1607#discussion_r1945562139. Scaled cross sections are now stored on a separate grid instead of using a single grid with a `prime_index`.


src/celeritas/grid/UniformGridInserter.hh
```diff
@@ -23,53 +23,34 @@ namespace celeritas
...
     // Add a grid of uniform log-grid data
-    XsIndex operator()(UniformGridData const& log_grid, SpanConstDbl values);
-    XsIndex operator()(UniformGridData const& log_grid, SpanConstFlt values);
+    GridId operator()(UniformGridData const& grid, SpanConstDbl values);
+    GridId operator()(UniformGridData const& grid, SpanConstFlt values);
 
   private:
-    CollectionBuilder<real_type, MemSpace::host, ItemId<real_type>> values_;
-    CollectionBuilder<XsGridData, MemSpace::host, ItemId<XsGridData>> xs_grids_;
+    CollectionBuilder<real_type, MemSpace::host, ItemId<real_type>> reals_;
```
sethrj: The default template arguments should work here:
```suggestion
    CollectionBuilder<real_type> reals_;
```
and we should consider changing it to a DedupeCollectionBuilder since a lot of the energy grids will be the same.

src/celeritas/grid/XsCalculator.hh
```diff
@@ -90,107 +84,67 @@ class XsCalculator
...
- * Get the raw cross section data at a particular index.
+ * Whether to use the scaled cross section grid.
  */
-CELER_FUNCTION real_type XsCalculator::get(size_type index) const
+CELER_FUNCTION bool XsCalculator::use_scaled(Energy energy) const
 {
-    CELER_EXPECT(index < data_.value.size());
-    return reals_[data_.value[index]];
+    real_type loge = std::log(value_as<Energy>(energy));
+    return !data_.lower || (data_.upper && loge >= data_.upper.grid.front);
```
sethrj: It shouldn't be possible to have neither `lower` nor `upper`, and it might be good to defer the call to std::log.
```suggestion
    return data_.upper && std::log(value_as<Energy>(energy)) >= data_.upper.grid.front;
```

src/celeritas/grid/XsGridInserter.hh
```diff
@@ -0,0 +1,70 @@
...
+                      SpanConstFlt lower_values,
+                      UniformGridData const& upper_grid,
+                      SpanConstFlt upper_values);
+
+    // Add a grid of uniform log-grid data
+    GridId operator()(UniformGridData const& grid, SpanConstDbl values);
+    GridId operator()(UniformGridData const& grid, SpanConstFlt values);
+
+  private:
+    CollectionBuilder<real_type, MemSpace::host, ItemId<real_type>> reals_;
```
sethrj: Same comment as above about template arguments and possible use of dedupe.

src/corecel/grid/VectorUtils.hh
```diff
@@ -48,5 +51,31 @@ inline bool is_monotonic_increasing(Span<T> grid)
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the grid cell width.
+ */
+template<class T>
+T calc_log_delta(Span<T const> grid)
+{
+    return std::pow(grid.back() / grid.front(), T(1) / (grid.size() - 1));
```
sethrj: Could use `fastpow` here, and also should assert that grid.size() > 1.

src/corecel/grid/VectorUtils.hh
```diff
@@ -48,5 +51,31 @@ inline bool is_monotonic_increasing(Span<T> grid)
     });
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the grid cell width.
```
sethrj: Maybe:
```suggestion
 * Calculate the (geometric) ratio of successive grid points in a uniform logarithmic grid.
```

sethrj (approved):
Props for the massive effort to complete this fix/feature/refactor! It'll be good to finally have that done.

src/celeritas/grid/XsCalculator.hh
amandalund: True, thanks!

amandalund: Actually I think it was ok before, otherwise we're not  correctly handling the case where there are only scaled values and the xs is below the grid.

PR #1623, opened 2025-02-13
Add generic geometry test interface to reduce duplication

sethrj:
This defines an interface for writing a GDML-based test once (doing ray tracing, getting volume names, and getting physical volumes) and reusing the results across the various geometry implementations. I currently have applied it to several test cases and use it to show differences between geometry implementations.

```c++

using FourLevelsTest
    = GenericGeoParameterizedTest<VecgeomGeantTestBase, FourLevelsGeoTest>;

TEST_F(FourLevelsTest, accessors)
{
    this->impl().test_accessors();
}

TEST_F(FourLevelsTest, trace)
{
    this->impl().test_trace();
}

```


src/geocel/GeantGdmlLoader.cc
```diff
@@ -0,0 +1,79 @@
...
+    ScopedMem record_mem("load_geant_geometry");
+    ScopedTimeLog scoped_time;
+
+    ScopedGeantLogger scoped_logger;
+    ScopedGeantExceptionHandler scoped_exceptions;
+
+    G4GDMLParser gdml_parser;
+    gdml_parser.SetStripFlag(opts_.clean);
+
+    gdml_parser.Read(filename, /* validate_gdml_schema = */ false);
```
mrguilima: Maybe add this schema validation flag as an option?
Actually, it is probably better to add a testing-time utility to validate all our GDML files once and for all.

test/geocel/GenericGeoTestBase.hh
```diff
@@ -109,23 +80,17 @@ class GenericGeoTestBase : virtual public Test, private LazyGeoManager
...
-    }
+    //// GenericGeoTestInterface ////
+
+    // Get the label for this geometry: Geant4, VecGeom, ORANGE
+    std::string_view geometry_type() const final;
+    // Access the geometry interface, building if needed
+    SPConstGeoInterface geometry_interface() const final;
+    // Find linear segments until outside
+    TrackingResult track(Real3 const& pos_cm, Real3 const& dir) final;
+    // Find linear segments until outside (maximum count
```
mrguilima, line 91:
```suggestion
    // Find linear segments until outside or max_step count is reached
```

src/corecel/io/Label.hh
```diff
@@ -50,14 +51,13 @@ struct Label
     //! Create an empty label
     Label() = default;
 
-    //! Create *implicitly* from a C string (mostly for testing)
-    Label(char const* cstr) : name{cstr} {}
-
-    //! Create *implicitly* from just a string name (capture)
+    //! Create *implicitly* from a captured string name (capture)
```
pcanal: Is the different between this constructor documentation and the next constructor intentional?  The other are specifically calling out that the input name is without extension.  Isn't it the case here too?

sethrj: No; I could consolidate them all.

sethrj: Actually the difference is that this one captures whereas the next few copy.

pcanal: That's fair and deserve a callout but the rest of the text should probably be identical.

pcanal:
```suggestion
    //! Create *implicitly* from a captured string name without extension (capture)
```

test/geocel/g4/GeantGeo.test.cc
```diff
@@ -446,222 +317,18 @@ TEST_F(SolidsTest, output)
...
-
 //---------------------------------------------------------------------------//
 
-TEST_F(SolidsTest, trace)
+TEST_F(SolidsTest, accessors)
 {
-    if (CELERITAS_UNITS != CELERITAS_UNITS_SI)
-    {
-        /*
-         * Note: with SI, this trace skips over the sphere! (During
```
amandalund, line 457: Was this resolved or removed from testing?

amandalund (approved):
Nice! Looks good to me.

test/geocel/g4/GeantGeo.test.cc
sethrj: Resolved; the difference was because this was along the edge of a surface.

mrguilima (commented):
just minor things.  LGTM (a bit too late)

src/geocel/GeantGdmlLoader.cc
sethrj: The only thing I've ever seen schema validation do is cause 20 second lags from trying to connect to the internet 🥲

test/geocel/GenericGeoTestBase.hh
mrguilima: At some point, I've run schema validation on GDML files (maybe in VecGeom repository), and I found and fixed a bunch GDML schema violations.  It was useful to run, but I agree with you, that it is just a waste of time in the usual set of tests.

src/geocel/GeantGdmlLoader.cc
mrguilima: At some point, I've run schema validation on GDML files (maybe in VecGeom repository), and I found and fixed a bunch GDML schema violations. It was useful to run, but I agree with you, that it is just a waste of time in the usual set of tests.

test/geocel/GenericGeoTestBase.hh
sethrj: I'll add this to the next PR

src/geocel/GeantGdmlLoader.cc
sethrj: Good to know!

PR #1624, opened 2025-02-13
Prune primaries that start outside the world volume

sethrj:
This does a quick comparison on a "primary" (offloaded track)'s position to the world solid's extents. If outside, it accumulates the lost energy for printing, and skips adding it to the buffer. When it flushes and at the end of the run, the amount of lost energy and number of primaries lost are printed. The detailed information about the primary can be viewed by  setting the local logging level to debug.

### DIFF
--- a/src/accel/LocalTransporter.cc
+++ b/src/accel/LocalTransporter.cc
@@ -171,7 +171,7 @@ void LocalTransporter::InitializeEvent(int id)
     CELER_EXPECT(id >= 0);
 
     event_id_ = id_cast<UniqueEventId>(id);
-    ++accum_num_events_;
+    ++run_accum_.events;
 
     if (!(G4Threading::IsMultithreadedApplication()
           && G4MTRunManager::SeedOncePerCommunication()))
@@ -221,7 +221,7 @@ void LocalTransporter::Push(G4Track const& g4track)
     track.event_id = EventId{0};
 
     buffer_.push_back(track);
-    buffer_energy_ += track.energy.value();
+    buffer_accum_.energy += track.energy.value();
     if (buffer_.size() >= auto_flush_)
     {
         /*!
@@ -268,7 +268,8 @@ void LocalTransporter::Flush()
     {
         CELER_LOG_LOCAL(debug)
             << "Transporting " << buffer_.size() << " tracks ("
-            << buffer_energy_ << " MeV cumulative kinetic energy) from event "
+            << buffer_accum_.energy
+            << " MeV cumulative kinetic energy) from event "
             << event_id_.unchecked_get() << " with Celeritas";
     }
 
@@ -289,11 +290,11 @@ void LocalTransporter::Flush()
 
     // Copy buffered tracks to device and transport the first step
     auto track_counts = (*step_)(make_span(buffer_));
-    accum_num_steps_ += track_counts.active;
-    accum_num_primaries_ += buffer_.size();
+    run_accum_.steps += track_counts.active;
+    run_accum_.primaries += buffer_.size();
 
     buffer_.clear();
-    buffer_energy_ = 0;
+    buffer_accum_ = {};
 
     size_type step_iters = 1;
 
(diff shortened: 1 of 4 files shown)


src/accel/SharedParams.cc
```diff
@@ -320,6 +322,24 @@ SharedParams::SharedParams(SetupOptions const& options)
...
+        CELER_ASSERT(solid);
+        G4VisExtent const& extent = solid->GetExtent();
+
+        BBox result{{extent.GetXmin(), extent.GetYmin(), extent.GetZmin()},
+                    {extent.GetXmax(), extent.GetYmax(), extent.GetZmax()}};
+        CELER_VALIDATE(result,
+                       << "world bounding box {" << result.lower() << ", "
+                       << result.upper() << "} is invalid");
+        return result;
+    }();
```
esseivaju, line 341: I was actually working on that in parallel but didn't add diagnostic 😅 

Can we retrieve the bounding box from `params.geant_geo_params()->bbox()` instead of constructing it in place. Also, doesn't it make more sense semantically for the `LocalTransporter` to hold the Bounding box(or keep a reference to `SharedParams`?

sethrj: The geo bbox could be slightly different: it is expanded by VecGeom and may have round off error. And yes the semantics make more sense to reference, but we can't store a safe reference to the shared params.

src/accel/LocalTransporter.cc
```diff
@@ -191,6 +193,23 @@ void LocalTransporter::Push(G4Track const& g4track)
 {
     CELER_EXPECT(*this);
 
+    if (Real3 pos = convert_from_geant(g4track.GetPosition(), 1);
+        !is_inside(bbox_, pos))
```
amandalund, line 197: I wonder if (for our `UserTrackingAction` integration at least) we could prevent this by checking that the `G4TrackStatus` is valid before offloading in the pre tracking action? Looking at how Geant4 handles tracks outside the geometry, in the stepping manager `SetInitialStep` [if a track is outside](https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/tracking/src/G4SteppingManager.cc#L348-L362) an error is thrown if it's a primary and the track status is set to `fStopAndKill` if it's a secondary. The `PreUserTrackingAction` is called just after this.

Not sure about the tracking manager integration though, I don't think the volume is checked or status changed before handing over the track...

src/accel/SharedParams.cc
esseivaju: > we can't store a safe reference to the shared params

Why is that?  The documentation says that `LocalTransporter` should have shared ownership of the sharedparams?

src/accel/LocalTransporter.cc
esseivaju: > Not sure about the tracking manager integration though, I don't think the volume is checked or status changed before handing over the track...

Indeed, it looks like the status stays the same with the tracking manager, a printout for outside tracks shows that they are marked as alive:

```
LocalTransporter.cc:202: debug: Discarding track outside world: 0.00119591 MeV from e- at {1990.41,-1141.8,-6566.06} along {-0.181364,-0.833315,-0.5222}
LocalTransporter.cc:216: debug: Track status: fAlive
```

src/accel/SharedParams.cc
sethrj: That probably should be updated to match the realities of the Geant4 allocators...  we have to allow `LocalTransporter` to take a `SharedParams*` because we don't want to enforce users having shared pointers. (Indeed, our singletons and most of our existing implementations use `G4ThreadLocal`.)

Anyway, it probably doesn't matter too much about which bbox we use; I figured making this part of code more Geant4-y would be ok.

sethrj:
Sorry for the duplicate effort @esseivaju 😅 at least @amandalund wasn't also trying to build the same thing simultaneously? Hopefully?

esseivaju (approved):

PR #1625, opened 2025-02-14
Support spline interpolation on a nonuniform grid

amandalund:
This is a little update to support cubic spline interpolation on a nonuniform grid, which could be used e.g. for the Livermore photoelectric cross sections.

### DIFF
--- a/src/celeritas/grid/NonuniformGridCalculator.hh
+++ b/src/celeritas/grid/NonuniformGridCalculator.hh
@@ -88,6 +88,7 @@ class NonuniformGridCalculator
 CELER_FUNCTION NonuniformGridCalculator NonuniformGridCalculator::from_inverse(
     NonuniformGridRecord const& grid, Values const& reals)
 {
+    CELER_EXPECT(grid.derivative.empty());
     return NonuniformGridCalculator{reals, grid.value, grid.grid, {}};
 }
 
@@ -175,6 +176,7 @@ NonuniformGridCalculator::grid() const
 CELER_FUNCTION NonuniformGridCalculator
 NonuniformGridCalculator::make_inverse() const
 {
+    CELER_EXPECT(deriv_offset_.empty());
     return NonuniformGridCalculator{reals_, y_offset_, x_grid_.offset(), {}};
 }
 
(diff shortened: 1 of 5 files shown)


src/celeritas/grid/NonuniformGridCalculator.hh
```diff
@@ -152,11 +169,13 @@ NonuniformGridCalculator::grid() const
...
  *
  * \pre The y values must be monotonic increasing.
+ *
+ * \note Spline interpolation will not be used on an inverse grid.
  */
 CELER_FUNCTION NonuniformGridCalculator
 NonuniformGridCalculator::make_inverse() const
 {
-    return NonuniformGridCalculator{reals_, y_offset_, x_grid_.offset()};
+    return NonuniformGridCalculator{reals_, y_offset_, x_grid_.offset(), {}};
```
sethrj, line 180: Maybe add an assertion for not having derivatives? Do we use spline interpolation for the range calculation? (Is it possible for us to be consistent with the interpolated functional representations of dE/dx, range, and inverse range?)

sethrj (approved):
LGTM!

src/celeritas/grid/NonuniformGridCalculator.hh
amandalund: Good question: it looks like all three use spline interpolation by default, so with spline enabled we'll probably need to build the inverse range table.

sethrj: Since the range is the integral of the inverse of a piecewise cubic (spline) function, we're not going to be able to exactly represent it with a spline function, right? Maybe it's good enough to use linear interpolation for range and inverse range.

And I would think that since the "inverse range" is really just transposing the x and y axes, it should be straightforward to have an "inverse spline" interpolation?

amandalund: Linear interpolation might be good enough? I guess I don't have a great feel yet for what will be good enough (or, close enough to Geant4).

Hmm I'm not sure how we could do inverse spline interpolation without recalculating the derivatives?

PR #1626, opened 2025-02-15
Fix loading of GDML files with reflection

sethrj:
Material mapping in geometry with reflection has been failing since #1620 due to inconsistent names between Geant4 and VecGeom, breaking the cms2018 test problem.

The Geant4 GDML loader has a few subtleties/bugs when reflected volumes (i.e. a negative `scale`) are involved. Most notably:
- If the input volume has a pointer and "strip pointers" is set to true, the `_refl` suffix will be stripped from the reflected volume.
- If an SD is attached to the reflected volume, it won't show up in the list of aux data.

This solves the first problem by removing pointers from volume names *after* loading by using a regular expression. It solves the second by looking through the reflection factory after loading, and inserting reflected SDs as needed.

To assist with testing, the SD loading code was moved from `celer-g4` to a `GdmlLoader` helper class.

Finally, by instantiating the UI pointers before the GDML parser, I've worked around a longstanding issue of the GDML parser resetting the Geant4 logger streams.


test/geocel/GeantGdmlLoader.test.cc
```diff
@@ -0,0 +1,147 @@
...
+    opts.detectors = true;
+    opts.pointers = PT::excise;
+
+    GeantGdmlLoader load_gdml(opts);
+    auto loaded = load_gdml(this->gdml_path("cms-eebackdee"));
+
+    EXPECT_EQ(2, loaded.detectors.count("ee_back_plate"));
+    EXPECT_EQ(2, loaded.detectors.count("ee_s_ring"));
+
+    // Reflected volume name is deleted by G4 GDML parser
```
amandalund:
```suggestion
    // Pointer is removed but reflected volume suffix is retained
```

src/geocel/GeantGdmlLoader.hh
```diff
@@ -0,0 +1,123 @@
...
+    //! \name Type aliases
+    using MapDetectors = std::multimap<std::string, G4LogicalVolume*>;
+    //!@}
+
+    //! How to handle pointers in volume names
+    enum class PointerTreatment
+    {
+        ignore,  //!< Pointers will remain in the volume name
+        amputate,  //!< All text after '0x' is removed
+        excise,  //!< Only pointers are carefully removed
```
amandalund: Haha, nice surgical theme... but maybe `strip` instead of `amputate`?

test/geocel/GeantGdmlLoader.test.cc
```diff
@@ -0,0 +1,147 @@
...
+    EXPECT_EQ(2, found.size());
+
+    static char const* const expected_log_messages[]
+        = {"Multiple Geant4 volumes are mapped to name 'EEBackQuad'"};
+    EXPECT_VEC_EQ(expected_log_messages, scoped_log_.messages());
+    static char const* const expected_log_levels[] = {"warning"};
+    EXPECT_VEC_EQ(expected_log_levels, scoped_log_.levels());
+}
+
+TEST_F(CmsEeTest, excise)
```
amandalund: Do we have to worry about round-tripping the GDML load/write?

amandalund (approved):
Oh no, not the refl suffixes again... thanks @sethrj!

src/geocel/GeantGdmlLoader.hh
sethrj: ChatGPT recommends `truncate` and `remove`, does that work?

test/geocel/GeantGdmlLoader.test.cc
sethrj: It seems the under-the-hood antics of the reflection factory mean that 'remove' and 'truncate' produce the same file (modulo pointer addresses) as the input. The `ignore` option results in two pointers in the output.

amandalund: Hmm, when I try to load this saved file I don't see any sensitive detectors... but they should be saved, right?

sethrj: If we attached the SDs inside the code, we would; Geant4's GDML loader doesn't actually do that (and for the unit test we don't either). If we did this as part of a celer-g4 run I think it would round trip correctly.

PR #1627, opened 2025-02-17
Fix crash in multithreaded celer-g4

amandalund:
Multithreaded celer-g4 using the primary generator was crashing because the `PGPrimaryGeneratorAction` created a `ParticleParams` (which uses the not-thread-safe `ScopedMem` in the constructor) on each thread. Rather than use the `ParticleParams` in the Geant4 generator, this adds a constructor to the Celeritas `PrimaryGenerator` that takes particle IDs.


sethrj (approved):
Good catch!

PR #1628, opened 2025-02-17
Require VecGeom 1.2.10 for CUDA RDC support

pcanal:
With VecGeom 1.2.9, we get the following error:

```
CMake Error at /home/pcanal/geant/gcc13/build/celeritas/_deps/g4vg-src/test/CMakeLists.txt:17 (add_executable):
  Impossible to link target 'g4vg_test' because the link item
  'VecGeom::vecgeomcuda', specified with the feature 'rdc_no_as_needed', has
  already occurred without any feature or 'DEFAULT' feature, which is not
  allowed.
```

In that version (and older) the RDC ‘final’ library and the ‘object’ library are combined under one name and thus the final library is used without the link feature (when handled as the ‘object’ library) and with the link feature (when handled as the ’final library) leading to the error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,7 +392,15 @@ if(CELERITAS_USE_ROOT)
 endif()
 
 if(CELERITAS_USE_VecGeom)
-  set(_min_vecgeom_version 1.2.10)
+  if(CELERITAS_USE_CUDA)
+    # 1.2.10 is needed due to the `no-as-needed` link feature
+    # requiring VecGeom to be build with a distinct object
+    # and final libraries
+    set(_min_vecgeom_version 1.2.10)
+  else()
+    # 1.2.8 is needed due to G4VG, which needs the logger.
+    set(_min_vecgeom_version 1.2.8)
+  endif()
   if(NOT VecGeom_FOUND)
     find_package(VecGeom ${_min_vecgeom_version} REQUIRED)
   elseif(VecGeom_VERSION VERSION_LESS _min_vecgeom_version)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,7 +392,7 @@ if(CELERITAS_USE_ROOT)
 endif()
 
 if(CELERITAS_USE_VecGeom)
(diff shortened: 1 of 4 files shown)


esseivaju:
we should update `scripts/spack.yaml` too

CMakeLists.txt
```diff
@@ -392,7 +392,7 @@ if(CELERITAS_USE_ROOT)
 endif()
 
 if(CELERITAS_USE_VecGeom)
-  set(_min_vecgeom_version 1.2.8)
+  set(_min_vecgeom_version 1.2.10)
```
sethrj: Let's add a comment about the feature/fixes needed for each version. 1.2.8 is needed due to G4VG (which needs the logger), and we could also use `CELERITAS_USE_CUDA` to leave the older compatibility in place.

sethrj (approved):
As Julien says, also update the spack recipes:
```console
$ git ls-files '*.yaml'
.readthedocs.yaml
scripts/ci/spack.yaml
scripts/docker/dev/rocky-cuda12.yaml
scripts/docker/dev/ubuntu-rocm6.yaml
scripts/docker/interactive/spack.yaml
scripts/spack.yaml
```

mrguilima:
I fully agree on dropping support to 1.2.9, and also on dropping support to 1.2.10 as soon as 1.2.11 tag is created.

sethrj:
@mrguilima I disagree especially with the thought of dropping 1.2.10 soon. Best practice is usually to support as wide a range of versions as possible to prevent users from having to upgrade all their dependencies at once. It's bad to require one specific patch version: what often happens is that some *other* dependency requires some other patch version, so the downstream user can't compile the conflicting versions.

I think a good policy is to support software versions *at least* a year old (and more like three years) unless it's a bleeding-edge package. "Support" can be kind of a loose term: until recently we "supported" down to 1.2.4 even though tests failed and some problems might not run. With <1.2.10, "support" could mean without CUDA. But generally we should at least try to get the code to compile with something older (and 1 year ago was actually 1.2.7!)

PR #1629, opened 2025-02-18
Fix navigation and mapping of reflecting volumes

sethrj:
This fixes the representation and mapping of sensitive detectors in reflecting volumes using a new capability/fix in G4VG. The  only real change in this code is `opts.reflection_factory = false;`, but there are a lot of new tests including a multi-level one that now agrees in a raytrace and matches geant4 when reconstructing:
multi-level-g4
multi-level-vg-latest


mrguilima (approved):
LGTM.  Thanks!

mrguilima:
I see that there are several tests failing in this branch, related to reflected volumes.

mrguilima:
> I see that there are several tests failing in this branch, related to reflected volumes.

... and after fetching your latest updates, all tests are now passing!

mrguilima (approved):

PR #1631, opened 2025-02-20
Move Geant4 SD integration into celeritas/ext

sethrj:
If we want to define the SD configuration options in `celeritas/inp` then we must also incorporate SD integration into `celeritas`. Since we have other low-level geant4 integration classes (e.g., physics list) in celeritas I think this is OK. `accel` remains the high-level external user integration code.


amandalund (approved):
Works for me!

PR #1632, opened 2025-02-20
Use `inp` to build `accel` core params

sethrj:
Continuing on the path for #1556 this builds core params using `inp`. Chained on #1631 .


sethrj:
@amandalund Finally passing 🎉 but low priority while we work on the magnetic field and user actions.

amandalund (approved):
Nice! I think once the conflicts are resolved this is good to go.

sethrj:
Also going to wait on #1664 .

sethrj:
One final thing I will add (in a separate PR) to this is direct `inp` construction for the integration singleton and shared params. Then we can mark the SetupOptions as deprecated before v0.6 (#1688)

PR #1633, opened 2025-02-20
Expand support for spline interpolation

### LINKED ISSUE #213: Implement spline interpolation
Spline interpolation (or some other nonlinear interpolation) is used in Geant4 by :
- Energy loss rate calculation (see `calc_energy_loss`)
- LivermorePE cross sections above K-shell energy (see LivermoreElement)
- Cross section (mean free path) for multiple scattering models (Urban and WenzelVI)

Research why it's necessary (the energy loss one provides an explanation of the consequences of linear interpolation, although it seems to me an alternate interpolation scheme such as log-log might have the same effect with less overheard) and compare spline interpolation with alternatives.

amandalund:
This expands the support for using our two different spline interpolation methods:
- Import from Geant4 which physics vectors use spline interpolation
- Add an input option for the interpolation method
- If spline is enabled, build all physics grids that should use spline interpolation with the second derivatives or order

A couple notes:
 - Cubic spline interpolation is used for the high energy Livermore PE cross sections, inverse range, and range, but the poly spline interpolation is not used for these since it's not yet supported on a nonuniform grid
 - I ran into a case where the poly spline interpolation produced a negative value, which I haven't investigated: I reverted to using linear interpolation in the app tests
 - Cubic spline interpolation is not perfectly invertible (for range and inverse range calculations), so we can occasionally run into cases where the calculated energy loss is negative
 - I haven't updated the root test data because I'm using Geant4 11.3 and I think we were using an older version for the data?
 
I reran testem3 with `simple_calo` (10 GeV primaries, 7 bins per decade for all cases), but the interpolation didn't really affect the energy deposition for this problem, and we're still seeing a small difference compared to Geant4:
testem3-edep-10GeV


src/celeritas/alongstep/detail/FluctELoss.hh
```diff
@@ -114,9 +114,8 @@ CELER_FUNCTION auto FluctELoss::calc_eloss(CoreTrackView const& track,
 
     // Calculate mean energy loss
     auto eloss = calc_mean_energy_loss(particle, phys, step);
-    CELER_EXPECT(eloss > zero_quantity());
 
-    if (eloss < particle.energy())
+    if (eloss < particle.energy() && eloss > zero_quantity())
```
sethrj, line 118: It would be good to add the note about non-invertible range leading to this error. That's an assumption that we're intentionally breaking and leads to an erroneous (but hopefully "good enough") nonphysical result.

src/geocel/GeantGdmlLoader.hh
```diff
@@ -40,7 +40,8 @@ class GeantGdmlLoader
   public:
     //!@{
     //! \name Type aliases
-    using MapDetectors = std::multimap<std::string, G4LogicalVolume*>;
+    using MapDetectors
+        = std::map<std::pair<std::string, int>, G4LogicalVolume*>;
```
sethrj: I don't understand this changeset (or why it's part of this PR, I guess cause of the testem3?): a multimap (SD name -> multiple logical volumes, from which you can get instance ID) seems like a reasonable way of storing the detector maps: it's a cleaner version of `map<string, vector<G4LV*>>`. The only thing your change should do differently is prohibit duplicate LVs, but duplicates shouldn't be possible when loading from G4.

src/corecel/grid/detail/GridAccessor.hh
```diff
@@ -95,20 +95,37 @@ class UniformGridAccessor : public GridAccessor
                                Values const& values);
 
     //! Get the x grid value at the given index
-    inline real_type x(size_type index) const final;
+    inline real_type x(size_type index) const;
 
     //! Get the y grid value at the given index
-    inline real_type y(size_type index) const final;
+    inline real_type y(size_type index) const;
```
sethrj: `override` and same above

sethrj (commented):
Still digging through this a bit. Some thoughts
- The entire point of using inverse range is improved accuracy when integrating dE/dx over a step, so it's concerning to use a range function that's not invertible.
- There's a *lot* of `interpolation` in the EM process input now... maybe there's a cleaner way to apply and propagate "common" options such as that and `integral_xs` we can move toward.

src/celeritas/alongstep/detail/FluctELoss.hh
amandalund: I thought this one would be ok since this was actually inconsistent with `MeanELoss` and `ElossApplier`, where we do allow zero loss... I did add a note in the `PhysicsStepUtils` though.

I agree about the inverse range, and I'm skeptical that spline is the best choice for some of these tables (especially given the log energy grid). For now I'm just trying to be as consistent as possible with Geant4, but we may reconsider once we've done some studies comparing the interpolation methods and have a better understanding of the impact on accuracy and performance.

src/geocel/GeantGdmlLoader.hh
amandalund: Ah, this can be reverted: this was for convenience when using the simple calo with testem3. The LVs were scrambled in the result, and this change ensured they were sorted by ID.

amandalund:
Yeah, I didn't have any better ideas on how to pass in that input (happy for any suggestions). I think also if we do eventually have the processes build the tables we'll need it there anyway?

amandalund:
ROOT test data needs to be updated with the correct Geant4 version.

sethrj (commented):
Sorry I still haven't gone through this in depth... it's good functionality to have, for sure, but the spline-ness has really infected a lot of the code. And as we will discuss today, maybe  we should abandon "exactly" reproducing geant4 with our code and go for something simpler, allowing g4hepem to manage the physics for experiments.

Have you also by chance taken a look at the GPU perforamnce/kernel occupancy from adding splines? I noticed in @esseivaju 's last results that the pre-step kernel occupancy dropped from 50% to 25% in his tests for v0.6... and I suspect that's from the spline work.

sethrj:
@amandalund Can we plan:
- fix conflicts from #1734 
- merge this
- tag as rc.3 and re-run tests?

amandalund:
@sethrj I've resolved the conflicts but still need to self-review/clean up some things.

amandalund:
Ok @sethrj hopefully this is a little cleaner after the grid/process simplifications.

src/celeritas/grid/detail/GridUtils.hh
```diff
@@ -0,0 +1,77 @@
...
+            CELER_LOG(warning)
+                << to_cstring(interpolation.type)
+                << " interpolation with order " << interpolation.order
+                << " is not supported on a grid with size "
+                << data.value.size() << ": defaulting to linear";
+            return;
+        }
+        CELER_ASSERT(interpolation.order > 1);
+        data.spline_order = interpolation.order;
+    }
```
sethrj, line 71:
```suggestion
    }
    else
    {
        /* No spline data added */
    }
```

test/celeritas/data/four-steel-slabs.root
sethrj, line 1: Did anything change here?

sethrj (approved):
Great! Much cleaner and more natural. Thanks!

test/celeritas/data/four-steel-slabs.root
amandalund: I don't think so...

PR #1634, opened 2025-02-21
Add perfetto track counters to accel LocalTransporter

esseivaju:
We were only using Perfetto track counter events in celery-sim. This moves it to access `LocalTransporter` so that geant4 applications integrating Celeritas can also access these counters.


src/accel/LocalTransporter.cc
```diff
@@ -57,6 +62,47 @@ bool nonfatal_flush()
...
+        {
+            std::string stream_id = std::to_string(get_geant_thread_id());
+            active_counter_ = std::string("active-" + stream_id);
+            alive_counter_ = std::string("alive-" + stream_id);
+            dead_counter_ = std::string("dead-" + stream_id);
+            queued_counter_ = std::string("queued-" + stream_id);
+        }
+    };
+
+    void operator()(StepperResult const& track_counts)
```
sethrj: This and the `static` variable could be const, right?

src/accel/LocalTransporter.cc
```diff
@@ -57,6 +62,47 @@ bool nonfatal_flush()
...
+};
+
+void trace(StepperResult const& track_counts)
+{
+    static thread_local TrackCounters trace_;
+    trace_(track_counts);
+}
+#else
+void trace(StepperResult const&) {}
+#endif
```
sethrj: Since we don't have any perfetto-enabled CI builds, and `trace_counter` is a null-op, I'd rather have this class be "always-on" so that it's guaranteed to compile. Hopefully the compiler will optimize it (aside from the negligible four static std::string storage) out if perfetto is disabled.

sethrj (approved):

src/accel/LocalTransporter.cc
esseivaju: Since `trace_counter` depends on the runtime environment, it's not always possible to optimize it to null-op, but the cost is very small, one `if(false)`

PR #1636, opened 2025-02-24
Refactor log handlers for better reuse

sethrj:
This moves the log handlers into a separate file for better reuse downstream.


sethrj:
Ping @pcanal

src/corecel/io/Logger.cc
```diff
@@ -144,8 +124,9 @@ LogLevel log_level_from_env(std::string const& level_env, LogLevel default_lev)
  */
 Logger make_default_world_logger()
 {
-    Logger log{celeritas::comm_world().rank() == 0 ? &default_global_handler
-                                                   : nullptr};
+    Logger log{celeritas::comm_world().rank() == 0
+                   ? LogHandler{StreamLogHandler{std::clog}}
```
pcanal, line 132: This is seems to be the one functional change and may deserve a note in the commit and/or a comment here (why the lock is not needed).

pcanal (approved):
LGTM.

PR #1639, opened 2025-02-25
Improve testing and documentation for VecGeom

sethrj:
There's a few minor things I pulled from #1422 that are needed for surface geometry but are useful outside of that

Infrastructure:
- Make transitive `geocel` geometry dependencies public

Testing:
- Loosen hepmc3 tolerance
- Add geometry 'inside' check to status executor
- Unify safety tolerance
- Do not gather step info for errored tracks

Documentation:
- Use bib reference for dormand prince

Github workflow:
- Always save cache if build succeeds
- Use expressions instead of shell


test/accel/HepMC3PrimaryGenerator.test.cc
```diff
@@ -148,7 +148,7 @@ TEST_F(HepMC3PrimaryGeneratorTest, no_vertex)
         999.99998921041, 1000, 999.99998921041, 1000.0000107896,
         999.99998921041, 999.99998921041, 1000, 1000, 999.99998921041,
         1000.0000107896, 1000.0000107896, 999.99998474121,};
-    EXPECT_VEC_SOFT_EQ(expected_energy, result.energy);
+    EXPECT_VEC_NEAR(expected_energy, result.energy, coarse_eps);
```
amandalund, line 151: What caused this change?

amandalund (approved):
Thanks for improving :shipit:

test/accel/HepMC3PrimaryGenerator.test.cc
sethrj: The `develop` branch fails on my local machine (M2 ARM): I guess this calculation not only introduces 7 digits of roundoff error, but the roundoff is quite sensitive to the architecture.

PR #1640, opened 2025-02-25
Fix rendering of README

dalg24:
The end of the code block was incorrectly annotated and messed up
the rendering on GitHub.


sethrj (approved):
Thanks!

PR #1641, opened 2025-02-25
Update externals and enforce G4VG version

sethrj:
This updates the minimum required G4VG version to 1.0.3 (you may want to wait till https://github.com/spack/spack/pull/49195 is merged to update). It also sets the `CMAKE_CXX_SCAN_FOR_MODULES` variable to `OFF` to prevent extra build time since Celeritas does not consume or produce C++20 modules.


stognini (approved):
Thanks for the fix @sethrj !

PR #1642, opened 2025-02-26
Add support for elliptical cylinders and elliptical cones in orange and g4org

elliottbiondo:
This MR adds `IntersectRegion` entries for elliptical cylinders and elliptical cones, both simple quadratics, and corresponding `SolidConverter` functions within g4org.


elliottbiondo:
@sethrj RTR

src/orange/g4org/SolidConverter.cc
```diff
@@ -382,21 +382,46 @@ auto SolidConverter::ellipsoid(arg_type solid_base) -> result_type
 }
 
 //---------------------------------------------------------------------------//
-//! Convert an elliptical cone
+/*!
+ * Convert an elliptical cone
+ *
+ * Expressions for lower/upper radii were found by solving the system of
+ * equations given by [1]:
```
sethrj: Instead of this doxygen reference (it would probably be better to use the github mirror or the official doxygen docs) we usually just reference `\c G4EllipticalCone`.

src/orange/g4org/SolidConverter.cc
```diff
@@ -382,21 +382,46 @@ auto SolidConverter::ellipsoid(arg_type solid_base) -> result_type
...
+    auto r_x = solid.GetSemiAxisX();
+    auto r_y = solid.GetSemiAxisY();
+    auto v = solid.GetZMax();
+    auto hh = solid.GetZTopCut();
+
+    auto lower_radii = scale_.to<Real2>(r_x * (v + hh), r_y * (v + hh));
+    auto upper_radii = scale_.to<Real2>(r_x * (v - hh), r_y * (v - hh));
+
+    return make_shape<EllipticalCone>(
+        solid, lower_radii, upper_radii, scale_(hh));
```
sethrj: It might be less error-prone (and consistent with what you do below) to immediately scale the values from the solid, then manipulate them in place. You could potentially use vector arithmetic from `corecel/math/ArrayOperators.hh` to simplify further.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+ * restrictions on these radii:
+ *
+ * 1) Either the lower or upper radii may be (0, 0); this is the only permitted
+ *    way for the elliptical cone to include the vertex.
+ * 2) The aspect ratio of the elliptical cross sections is constant. Thus, the
+ *    aspect ratio at z = -hh must equal the aspect ratio at z = hh.
+ * 3) Degenerate elliptical cones with lower_radii == upper_radii (i.e.,
+ *    elliptical cylinders) are not permitted.
+ * 4) Degenerate elliptical cones where lower or upper radii are equal to
+ *    (0, x) or (x, 0), where x is non-zero, are not permitted.
```
sethrj: Since we generally don't show the documentation from constructors in the user doc, but this is good to have available on the ORANGE doc page, could you move this to the class documentation?

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+    auto has_negative
+        = [&](Real2 const& radii) { return radii[X] < 0 || radii[Y] < 0; };
+
+    // True if radii is (0, 0)
+    auto is_vertex = [&](Real2 const& radii) {
+        return soft_zero(radii[X]) && soft_zero(radii[Y]);
+    };
+
+    // True if radii is (0, x) || (x, 0), where x != 0
+    auto is_partial_zero = [&](Real2 const& radii) {
```
sethrj: These lambdas don't need captures (if you use the built-in function `soft_zero` from SoftEqual.hh insted of the local functor.)

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+        return (soft_zero(radii[X]) != soft_zero(radii[Y]));
+    };
+
+    // Check for negatives
+    CELER_VALIDATE(!has_negative(lower_radii_),
+                   << "Lower radii cannot have negative values: "
```
sethrj: Minor formatting issue: this should be formatted consistently with the [docs](https://celeritas-project.github.io/celeritas/user/implementation/corecel.html#c.CELER_VALIDATE) and elsewhere in the file
```suggestion
                   << "negative lower radii: "
```

and with the VALIDATE below too

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+    return hh_ >= other.hh_;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build surfaces.
+ *
+ * The elliptical surface can be expressed as [1]:
+ *
+ * (x/r_x)^2 + (y/r_y)^2 = (v-z)^2,
```
sethrj: Use doxygen `\f[` markers here and probably a `\verbatim` block below. Same comment about the elliptical cone source (and use doxygen markups for that too.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -26,19 +26,20 @@ class IntersectSurfaceBuilder;
...
+ * the exterior bounding box of a intersect region should be <em>centered
+ * on the origin</em>, and objects should be aligned along the \em z axis.
  *
  * When implementing this class, prefer to build simpler surfaces (planes)
  * before complex ones (cones) in case we implement short-circuiting logic,
  * since expressions are currently sorted.
  *
- * \note Additional methods such as volume calculation may be added here later.
+ * \note Additional methods such as volume calculation may be added here
+ * later.
```
sethrj: Any idea why this got rewrapped? The first line changed is only at 76 cols.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -254,7 +342,8 @@ class GenPrism final : public IntersectRegionInterface
                               TrapFace const& lo,
                               TrapFace const& hi);
 
-    // Construct from half Z height and 4 vertices for top and bottom planes
+    // Construct from half Z height and 4 vertices for top and bottom
+    // planes
```
sethrj: Maybe your editor somehow has comment lengths set to less than 80? Can you try to revert these changes?

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -429,6 +429,7 @@ TEST_F(EllipsoidTest, errors)
 TEST_F(EllipsoidTest, encloses)
 {
     Ellipsoid ellisoid({1, 2, 3});
+    EXPECT_TRUE(ellisoid.encloses(Ellipsoid({1, 2, 3})));
```
sethrj: Whoops just noticed that the variable name is misspelled...

sethrj (approved):
Looks great! Only a few (mostly formatting) comments.

src/orange/g4org/SolidConverter.cc
elliottbiondo: I did it this way because while `r_x` and `r_x` have physical interpretations, they are used as a slope within the lower/upper radii calcs and are therefore unitless. So I figured it was best to do everything in native units and then convert it. I can do it the other way and add a comment though.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: They need captures for `X` and `Y`

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: Sure. I am using off-the-shelf clang-format, so it should be respecting whatever is in `.clang-format`. I am really not sure what is going on and why it only affects these lines (if it was a 76-character problem there would probably be a lot more changes). Frustratingly, this is exactly the kind of tedious work clang-format is supposed to be preventing.

src/orange/g4org/SolidConverter.cc
sethrj: Oh I didn't realize they are unitless. I think converting first (and noting that x/y are ratios) would make it clearer.

src/orange/orangeinp/IntersectRegion.cc
sethrj: X and Y should be `constexpr`, in which case they don't need to be captured because they're effectively globals.

src/orange/orangeinp/IntersectRegion.hh
sethrj: This isn't happening from inside vscode?

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+ *
+ * [1] apc.u-paris.fr/~franco/g4doxy/html/G4EllipticalCone_8hh-source.html
+ */
+void EllipticalCone::build(IntersectSurfaceBuilder& insert_surface) const
+{
+    insert_surface(Sense::outside, PlaneZ{-hh_});
+    insert_surface(Sense::inside, PlaneZ{hh_});
+
+    auto X = to_int(Axis::x);
+    auto Y = to_int(Axis::y);
```
sethrj: These should be constexpr

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -334,6 +334,233 @@ void Ellipsoid::output(JsonPimpl* j) const
...
+EllipticalCone::EllipticalCone(Real2 const& lower_radii,
+                               Real2 const& upper_radii,
+                               real_type halfheight)
+    : lower_radii_{lower_radii}, upper_radii_{upper_radii}, hh_{halfheight}
+{
+    SoftZero soft_zero;
+    SoftEqual soft_equal;
+
+    auto X = to_int(Axis::x);
+    auto Y = to_int(Axis::y);
```
sethrj: Also constexpr

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: This MR was done completely in `vim`. My `vimrc` is nearly empty.

sethrj: Weird... no idea why it would reformat those then, especially if you didn't touch them. Do you use the git-clang-format commit hook? (Installed via scripts/dev/install-commit-hooks.sh)

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Two also need `soft_zero`, so I am now capturing only that.

sethrj: I think the local `soft_zero`, because it's using the default tolerance, is no different than the `celeritas::soft_zero` free function , which doesn't need to be bound.  Not super important, it's fine to capture blindly, but it's good knowledge to have.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Both of your comments are on elliptical cone. I have moved the all the info from the .cc file in the class-level doc string.

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: Nope

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -424,7 +538,7 @@ class Involute final : public IntersectRegionInterface
  *   - \c theta polar angle of the shape's main axis, e.g. the segment defined
  *     by the centers of the Z faces. Validity range is `[0, 1/4)`;
  *   - \c phi azimuthal angle of the shape's main axis (as explained above).
- *     Validity range is `[0, 1)`.
```
elliottbiondo, line 427: I have not explanation for this change, which is seemingly a no-op.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -424,7 +538,7 @@ class Involute final : public IntersectRegionInterface
...
+ *     Validity range is `[0, 1)`.
```
elliottbiondo, line 541: I have no explanation for this whitespace change, which should be a no-op.

elliottbiondo:

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: (done)

src/orange/orangeinp/IntersectRegion.hh
sethrj: weeeird

sethrj: Maybe Vidor committed some weird non-space space :\?

sethrj: Yeah there's some weird garbage in there, `0xC2A0`... no worries!

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Oops, looks like this breaks Windows. Shall I change it back?

```
D:\a\celeritas\celeritas\src\orange\orangeinp\IntersectRegion.cc(427): error C3493: 'X' cannot be implicitly captured because no default capture mode has been specified
```

sethrj: Ech I was afraid of that, I've seen MSVC be buggy about constexpr. Yeah go ahead and change to universal capture... the compiler should be able to optimize it out. Alternatively do like our `rotate` function and define a local `enum {X = 0, Y = 1, Z=2};`.

sethrj (approved):

PR #1643, opened 2025-02-27
Update cmake-git-version

sethrj:
This updates the cmake git version script so that branch names are included in the version metadata, and the version is automatically regenerated when switching branches. Adding new commits on the same branch does *not* update the hash. This seems like a good balance for keeping the version in sync.

```console
$ ./bin/celer-sim --version
0.6.0-dev.135+cmake-git-version-update.c39ec01bd
$ git co -b derpyderp
Switched to a new branch 'derpyderp'
$ ninja celer-sim
[0/1] Re-running CMake...
-- Configuring done (0.3s)
-- Generating done (0.2s)
-- Build files have been written to: /Users/seth/Code/celeritas-temp/build
[5/6] Linking CXX executable bin/celer-sim
$ ./bin/celer-sim --version
0.6.0-dev.136+derpyderp.69d33bf6a
$ git commit --allow-empty -m empty
[derpyderp 6a6997b05] empty
$ ninja celer-sim
ninja: no work to do.
```

### DIFF
--- a/.github/workflows/build-spack.yml
+++ b/.github/workflows/build-spack.yml
@@ -51,7 +51,7 @@ jobs:
             geant: "11.2"
     env:
       CCACHE_DIR: "${{github.workspace}}/.ccache"
-      CCACHE_MAXSIZE: "${{matrix.special == 'asanlite' && '300' || '100'}}Mi"
+      CCACHE_MAXSIZE: "100Mi"
       CMAKE_PRESET: >-
         ${{format('reldeb-{0}{1}{2}',
                   matrix.geometry,
--- a/.github/workflows/build-spack.yml
+++ b/.github/workflows/build-spack.yml
@@ -51,7 +51,7 @@ jobs:
             geant: "11.2"
     env:
       CCACHE_DIR: "${{github.workspace}}/.ccache"
-      CCACHE_MAXSIZE: "100Mi"
+      CCACHE_MAXSIZE: "${{matrix.special == 'asanlite' && '300' || '100'}}Mi"
       CMAKE_PRESET: >-
         ${{format('reldeb-{0}{1}{2}',
                   matrix.geometry,
--- a/cmake/CgvFindVersion.cmake
+++ b/cmake/CgvFindVersion.cmake
@@ -102,12 +102,16 @@ macro(_cgv_timestamp tsfile tsvar)
 endmacro()
 
 #-----------------------------------------------------------------------------#
-# Execute a command, logging to the screen
-macro(_cgv_execute_process)
-  message(VERBOSE "Executing ${ARGV} from ${CGV_SOURCE_DIR}")
-  execute_process(${ARGV}
+# Execute a command, logging verbosely, saving output
+macro(_cgv_git_call_output output_var)
+  message(VERBOSE "Executing ${GIT_EXECUTABLE} from ${CGV_SOURCE_DIR}: ${ARGV}")
+  execute_process(
+    COMMAND "${GIT_EXECUTABLE}" ${ARGN}
     WORKING_DIRECTORY "${CGV_SOURCE_DIR}"
     OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_VARIABLE GIT_ERR
+    RESULT_VARIABLE GIT_RESULT
+    OUTPUT_VARIABLE ${output_var}
   )
 endmacro()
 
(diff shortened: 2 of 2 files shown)


.github/workflows/build-spack.yml
```diff
@@ -51,7 +51,7 @@ jobs:
             geant: "11.2"
     env:
       CCACHE_DIR: "${{github.workspace}}/.ccache"
-      CCACHE_MAXSIZE: "100Mi"
+      CCACHE_MAXSIZE: "${{matrix.special == 'asanlite' && '300' || '100'}}Mi"
```
pcanal: this seems unrelated ...

cmake/CgvFindVersion.cmake
```diff
@@ -180,25 +271,46 @@ function(_cgv_try_git_describe)
...
+  # Get git branch: may fail if detached
+  execute_process(
+    COMMAND "${GIT_EXECUTABLE}" "symbolic-ref" "--short" "HEAD"
+    WORKING_DIRECTORY "${CGV_SOURCE_DIR}"
+    ERROR_VARIABLE _GIT_ERR
+    OUTPUT_VARIABLE _BRANCH_STRING
+    RESULT_VARIABLE _GIT_RESULT
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
```
pcanal, line 307: Shouldn't we handle the case there the `git symbolic-ref` failed?

.github/workflows/build-spack.yml
sethrj: It's an infrastructure change I had queued up, the cache size is too small to fit a whole build. Too small for its own PR, close enough to this one to be merged alongside 😀

cmake/CgvFindVersion.cmake
sethrj: It is automatically handled, the output is blank which results in no branch. I did test this case on a detached head

pcanal: Fair enough.  Consider adding that information as a code. comment.

.github/workflows/build-spack.yml
pcanal: I disagree on both point :).  It is (in my humble opinion) okay to have the odd/rare very small commit in the repository and we don't have enough churn in the CI that having one extra PR slows up down.  (And even though it is infrastructure it is both unrelated to CMake configuration and version number handling).

cmake/CgvFindVersion.cmake
```diff
@@ -56,39 +56,110 @@ CgvFindVersion
...
+    repository) sets this property.
+
+This script can also be run from the command line to determine git repository
+data::
+
+  cmake -P cmake/CgvFindVersion.cmake
+
+To print only the version string (to stderr), and from a custom directory::
+
+  cmake -DSOURCE_DIR=. -DONLY=VERSION -P cmake/CgvFindVersion.cmake
```
pcanal, line 84: I am a bit confused by the example.  What is 'custom' about the 2nd case?  It seems in both case, we need have the current directory be the top level of the Celeritas source directory or maybe it is meant to describe:
```
cd where_standalone_cgv_find_version_is
cmake -DSOURCE_DIR=path_to_some_git_repository -DONLY=VERSION -P cmake/CgvFindVersion.cmake
```
but in this case, I failed to make it work:
```
$ cmake -DSOURCE_DIR=../VecGeom -DONLY2=VERSION -P cmake/CgvFindVersion.cmake
VERSION="0.6.0"
VERSION_STRING="0.6.0-dev.133+2bd1777cc"
$ (cd ../VecGeom; git lg1 -1 )
* d801efb1e - (8 days ago) Fix build failure in cuda 11.5 with veccore 0.8.2 — Seth Johnson (HEAD -> v1-patches, origin/v1-patches)
$ cd ../VecGeom
$ cmake -P ../celeritas/cmake/CgvFindVersion.cmake
VERSION="0.6.0"
VERSION_STRING="0.6.0-dev.133+2bd1777cc"
```

.github/workflows/build-spack.yml
sethrj: We've disagreed on this countless times I feel, and each time we end up spending more time commenting and ranting than we would just on merging 😞 which means you're probably right and it should just be a separate trivial PR

cmake/CgvFindVersion.cmake
sethrj: Are you sure you have this branch checked out?
```cmake
$ cmake -DSOURCE_DIR=../vecgeom -P cmake/CgvFindVersion.cmake
VERSION="1.2.10"
VERSION_STRING="1.2.10-6+v1-patches.d801efb1e"
```
PS your `DONLY2=VERSION` is a typo

.github/workflows/build-spack.yml
pcanal: Thanks. (This challenges is also in part due to the squash on merge, i.e. those 2 things deserve separate commits, not necessarily separate PRs).

cmake/CgvFindVersion.cmake
pcanal: > Are you sure you have this branch checked out?

Sigh .. of course not :( .. Sorry for the notice.

> PS your DONLY2=VERSION is a typo

Sort of (it was intentional to see the full output) but is indeed completely irrelevant here.

pcanal: > Are you sure you have this branch checked out?

Now that I did, everything works as expected.

pcanal (approved):
LTGM.  Thanks.

PR #1644, opened 2025-02-27
Remove G4EmExtraPhysics from FtfpBertPhysicsList

whokion:
# Remove G4EmExtraPhysics
- Remove G4EmExtraPhysics from the celeritas FtfpBertPhysicsList until Gamma- and Lepto-Nuclear processes are added.


src/celeritas/ext/FtfpBertPhysicsList.cc
```diff
@@ -51,8 +50,7 @@ FtfpBertPhysicsList::FtfpBertPhysicsList(Options const& options)
     // Muon and hadrom EM standard physics not supported in Celeritas
     detail::emplace_physics<detail::MuHadEmStandardPhysics>(*this, verbosity);
 
-    // Synchroton radiation & GN physics
-    detail::emplace_physics<G4EmExtraPhysics>(*this, verbosity);
```
sethrj, line 55: Would it be possible to make this an additional flag in GeantPhysicsOptions rather than turning it off wholesale? (And should we also add the same option to the EmPhysics since that's part of Geant4's EmStandard?)

whokion: As the physics list is constructed by `particles` and `processes`, it would be better not mangle physics processes between offloaded-particles and the rest - otherwise, it will always lead a bias in physics (i.e, not apple-to-apple) between Geant4 and Geant4+Celeritas. In other words, why do we want to have unnecessary Geant4 physics constructor option if associated particles (i.e., gamma and leptons in this case) never exist on the host.  This is exactly the same logic behind why there is no optional switch between  (celeritas) EmStandardPhysics and G4EmStandardPhysics in this physics list.  I think that the option for the em extra physics  should be reserved for Celeritas EmExtraPhysics when it is available, not for G4EmExtraPhysics here.

amandalund: On a slightly different note: for EM standard physics we moved the muon and hadron physics to a separate constructor (`MuHadEmStandardPhysics `) so that Geant4 will still simulate those when we use our Celeritas EM physics. It looks like there are also muon processes in EM extra physics: do we need to do the same here (until muon physics is enabled in Celeritas)?

whokion: Okay, as long as the muon is not offloaded, we can re-arrange physics processes as needed - so, in this case, your are right that the muon-nuclear processes should be added to `MuHadEmStandardPhysics`.  I thought that   MuHadEmStandardPhysics is for a celeritas muon physics constructor of which processes have been added by you and @stognini.  Anyway, it is a good point that we re-organize physics constructors explicitly for an offloading particle type or a group of offloading particles, and should not mix them together with non-offloading particles.

amandalund: We added muon processes to the Celeritas EM standard physics, but they are disabled for now. Once we enable them we should remove the muon processes from `MuHadEmStandardPhysics` (and rename it to reflect that it's only hadronic).

sethrj (changes requested):
TODO: configure extra physics via setup

whokion:
@sethrj You are asking to add an option that we do not support (at this moment), which leads a possibility of biases in physics results.

sethrj:
Sorry @whokion , I missed that takeaway from the conversation. It makes sense to me now. @amandalund are you good with this?

sethrj (approved):

amandalund (approved):
Not sure if @whokion was planning on adding the muon nuclear processes to our `FtfpBertPhysicsList`?

whokion:
Yes, it is a part of neutron physics implementation (which is on hold for now, but hope to resume soon) which takes some time to have all features including the photon de-exciatation/evaporation process followed by photon-nuclear and lepto-nuclear processes.  It would be better not to support this option until we complete the implementation.

PR #1647, opened 2025-03-02
Fix `getenv_flag`

amandalund:
This fixes the `getenv_flag` helper function, which was always using the default value if the environment variable hadn't been queried yet.

### DIFF
--- a/src/corecel/sys/Environment.cc
+++ b/src/corecel/sys/Environment.cc
@@ -72,18 +72,20 @@ std::string const& getenv(std::string const& key)
 GetenvFlagResult getenv_flag(std::string const& key, bool default_val)
 {
     std::scoped_lock lock_{getenv_mutex()};
-    bool inserted = environment().insert({key, default_val ? "1" : "0"});
-    if (inserted)
+
+    std::string value;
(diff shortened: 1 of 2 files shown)


sethrj (approved):
Good catch!

src/corecel/sys/Environment.cc
```diff
@@ -72,18 +72,20 @@ std::string const& getenv(std::string const& key)
...
-    std::string const& result = environment()[key];
-    if (result.empty())
+    if (value.empty() && environment().insert({key, default_val ? "1" : "0"}))
     {
-        // Variable was queried but is empty
+        // Variable is empty
         return {default_val, true};
     }
 
+    std::string const& result = environment()[key];
```
pcanal, line 88: This calls `std::getenv` a second time (getting the same result :)) for a small inefficiency.

sethrj: Very subtle but yes you're right!

pcanal: A secondary effect is that value/sys_value is set but never used besides checking if it is empty. (Again not an issue besides being a little puzzling on first read)

PR #1648, opened 2025-03-03
Implement detector action for optical physics

lebuller:
This PR creates a new optical step action class for performing a calorimeter action, `CaloAction`. This class executes at the `user_post` order step and the `step ` function of this class calls a `CaloExecutor` for each track. This `CaloExecutor` struct checks for tracks that still have active/alive status at the `user_post` stage and checks there volume id against a list of input optical detector volume ids. If the track is alive and is in a volume corresponding to an optical detector, it contributes to the detector and is killed. The "contribution" is currently just to print out the volume the particle is killed in and the energy of the particle at this stage, will need to be expanded in future MR


sethrj:
I think this is superseded by @2077 and in either way has diverged way too much since this was first pushed ☹️

PR #1649, opened 2025-03-03
Support sensitive detectors in replica/parameterized volumes

### LINKED ISSUE #1637: Support replica/parameterised volumes in SD callback
From https://github.com/celeritas-project/gpu-delta-review-2025/issues/2, the B5 example contains an SD embedded in a replica volume:
```
*** G4Exception : celer.event.flush
      issued by : /mnt/c/Users/srahman1/Documents/celeritas/src/geocel/GeantGeoUtils.cc:360
Celeritas runtime error: sensitive detectors inside of replica/parameterized volumes are not supported: 'cellPhysical' inside {{pv='EMcalorimeterPhysical', lv=10='EMcalorimeterLogical'} -> {pv='fSecondArmPhys', lv=3='secondArmLogical'}}
```

See also https://github.com/celeritas-project/g4vg/pull/25

sethrj:
Geant4 has mutable physical volumes ("replica" and "parameterised") that allow a single volume instance to appear as many different volume instances, with the idea of saving memory. It unfortunately complicates matters a lot because it requires updating the volume on the local thread during tracking. Generating VecGeom versions of these (by literally copy-pasting the volumes) was added in #1071 and https://github.com/celeritas-project/g4vg/pull/25 , but we punted on enabling level-based detector reconstruction (#1495) in such volumes. Unfortunately the B5 benchmark problem (https://github.com/celeritas-project/gpu-delta-review-2025/issues/2) has SDs in these volumes.

This PR:
- adds a new geometry test problem and new capabilities to the geometry test interface;
- changes the GeoParams interface so that VolumeInstanceId corresponds not to just  a `G4PV*` but to a `{G4PV*, int}` pair (written as a struct using an opaque ID) that embeds the copy number needed to update Geant4's geometry
- adds reconstruction support to the `set_history` helper and tests to the level updater

Closes #1637

--- 

ADDENDUM a month later: the discussion about a [touchable IDs user request](https://jira-geant4.kek.jp/browse/UR-100)  brings up this same issue:

> 
> ## An issue / problem
>  
> Looking at G4VPhysicalVolume type (header) reminded me of an inconvenient fact - that only G4PVPlacement volumes are uniquely identified by this instance ID.  Multiple slices in a single ‘Replica’ (or ‘Division’ ) will share this identifier, as they are represented by a single physical volume in memory.  Similarly all instances of a ‘physical volume Parameterisation’ share the same identifier. 
>  
> ### Open question(s):
> Do you allow a user to create a ‘Replica’ or ‘Division’ which slices the original volume into separate sub-volumes?   If so, they share the same physical volume pointer (a G4PVReplica or G4PVDivision) and as such would share the same Instance ID.  So a separate sub-solution would be required for these types.  It is possible to assertain whether a physical volume is one of these types - but likely you already know this during the construction of your geometry. 
>  
> ### Potential extension / solution
>  
> I can imagine a simple extension to the InstanceID which could be used to identify every physical volume uniquely.  For each repetetive volume (replica, division, parameterised volume) it would create a new id combining an offset plus its replica-number.   The offsets could be pre-calculated at the start of the application to provide enough ’space’ to ensure the new / revised identifier is unique, and that either no gaps are created (or that padding is minimal. )
>  
> I hope that I have summarised our discussions well, and provided the means to start a prototype using this ‘hidden’ part of the Geant4 implementation.

### DIFF
--- a/src/geocel/g4/GeantGeoParams.cc
+++ b/src/geocel/g4/GeantGeoParams.cc
@@ -165,8 +165,8 @@ VolumeId GeantGeoParams::find_volume(G4LogicalVolume const* volume) const
 /*!
  * Get the Geant4 physical volume corresponding to a volume instance ID.
  *
- * \warning For Geant4 paramterised/replicated volumes, external state (e.g.
- * the local navigation) *must* be used in concert with this method: i.e.,
+ * \warning For Geant4 parameterised/replicated volumes, external state (e.g.
+ * the local navigation) \em must be used in concert with this method: i.e.,
  * navigation on the current thread needs to have just "visited" the instance.
  *
  * \todo Create our own volume instance mapping for Geant4, where
--- a/src/geocel/g4/GeantGeoParams.cc
+++ b/src/geocel/g4/GeantGeoParams.cc
@@ -165,20 +165,39 @@ VolumeId GeantGeoParams::find_volume(G4LogicalVolume const* volume) const
 /*!
  * Get the Geant4 physical volume corresponding to a volume instance ID.
  *
- * If the input ID is false, a null pointer will be returned.
+ * \warning For Geant4 paramterised/replicated volumes, external state (e.g.
+ * the local navigation) *must* be used in concert with this method: i.e.,
+ * navigation on the current thread needs to have just "visited" the instance.
+ *
+ * \todo Create our own volume instance mapping for Geant4, where
+ * VolumeInstanceId corresponds to a (G4PV*, int) pair rather than just G4PV*
+ * (which in Geant4 is not sufficiently unique).
  */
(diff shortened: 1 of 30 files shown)


src/geocel/g4/GeantGeoParams.cc
```diff
@@ -165,20 +165,39 @@ VolumeId GeantGeoParams::find_volume(G4LogicalVolume const* volume) const
 /*!
  * Get the Geant4 physical volume corresponding to a volume instance ID.
  *
- * If the input ID is false, a null pointer will be returned.
+ * \warning For Geant4 paramterised/replicated volumes, external state (e.g.
```
mrguilima:
```suggestion
 * \warning For Geant4 parameterised/replicated volumes, external state (e.g.
```

mrguilima (approved):
This looks great!
Thanks, @sethrj

PR #1303, opened 2024-07-02
Add detailed error checking for action sequence behavior

sethrj:
As I work toward error correction (#687) I'm increasingly worried about subtly breaking the tracking loop by improperly changing the track status or setting an action ID that's out of order. This PR is a step in the right by adding a *copy* of the pre-action track status, post-step action, and pre-step action; it then checks after the action that each track's state has been consistently updated.

The status checker can be added by the user to the action registry. If added, *the action sequence will run it after every action*. This capability is on by default for unit tests.

Enforcing the status checks gives us some nice documentation on what is allowed and when: my initial implementation got several subtle conditions wrong.
https://github.com/celeritas-project/celeritas/blob/90a01b395e17199aae6fd6b70f967aa35e0dcf5b/src/celeritas/track/detail/StatusCheckExecutor.hh#L61-L118

PR #1650, opened 2025-03-04
Add an option to enable status checker through celer-sim

sethrj:
The status checker was added in #1303 but is currently only enabled for unit tests. This adds an option to build it from the front end and use in multithread contexts.

### DIFF
--- a/src/celeritas/setup/Problem.cc
+++ b/src/celeritas/setup/Problem.cc
@@ -15,11 +15,9 @@
 #include "corecel/Config.hh"
 
 #include "corecel/cont/VariantUtils.hh"
-#include "corecel/cont/detail/VariantUtilsImpl.hh"
 #include "corecel/io/Logger.hh"
 #include "corecel/io/OutputRegistry.hh"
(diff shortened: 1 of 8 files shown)


src/celeritas/setup/Problem.cc
```diff
@@ -447,6 +446,13 @@ problem(inp::Problem const& p, ImportData const& imported)
         ActionDiagnostic::make_and_insert(*core_params);
     }
 
+    if (p.diagnostics.status_checker)
+    {
+        // Add detailed debugging of track states
+        // TODO: this environment variable will be replaced by celer::inp soon!
```
amandalund: Which environment variable?

amandalund (approved):
Thanks!

src/celeritas/setup/Problem.cc
sethrj: Oops stale comment, I originally patched it in with an environment variable but figured it wasn't much more effort to add to the JSON input.

PR #1652, opened 2025-03-04
Improve geometry testing

sethrj:
This pulls another couple of small changes from #1422 . The geometry heuristic test now guarantees reproducibility between CPU and GPU. The VecGeom test is now somewhat simplified. Finally the `ScopedLimitSaver` used by VecgeomParams is now cleaned up a bit.

### DIFF
--- a/src/corecel/sys/ScopedLimitSaver.cuda.cc
+++ b/src/corecel/sys/ScopedLimitSaver.cuda.cc
@@ -11,12 +11,15 @@
 
 namespace celeritas
 {
+namespace
+{
 //---------------------------------------------------------------------------//
-// Static data
-Array<cudaLimit, 2> const ScopedLimitSaver::cuda_attrs_{
+Array<cudaLimit, 2> const g_attrs{
     {cudaLimitStackSize, cudaLimitMallocHeapSize}};
-Array<char const*, 2> const ScopedLimitSaver::cuda_attr_labels_{
-    {"stack size", "heap size"}};
+Array<char const*, 2> const g_labels{{"stack size", "heap size"}};
+
+//---------------------------------------------------------------------------//
+}  // namespace
 
 //---------------------------------------------------------------------------//
 /*!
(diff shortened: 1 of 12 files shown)


test/celeritas/geo/HeuristicGeoTestBase.cc
```diff
@@ -60,41 +49,69 @@ void HeuristicGeoTestBase::run_host(size_type num_states, real_type tolerance)
...
+    }
+    else
+    {
+        std::cout << "Skipping device comparison: device not active\n";
     }
 }
 
 //---------------------------------------------------------------------------//
 /*!
  * Run tracks on device and compare the resulting path length.
```
mrguilima:
```suggestion
* Run tracks on host or device and compare the resulting path length.
```

src/corecel/sys/ScopedLimitSaver.cuda.cc
```diff
@@ -27,8 +29,8 @@ Array<char const*, 2> const g_labels{{"stack size", "heap size"}};
  */
 ScopedLimitSaver::ScopedLimitSaver()
 {
-    static_assert(g_attrs.size() == orig_limits_.size()
-                  && g_labels.size() == orig_limits_.size());
+    static_assert(g_attrs.size() == decltype(orig_limits_){}.size()
+                  && g_labels.size() == g_attrs.size());
```
mrguilima: So this was ultimately the reason for discrepancies in the surface model?

mrguilima (approved):
Awesome!

src/corecel/sys/ScopedLimitSaver.cuda.cc
sethrj: Alas no, this bit here was only related to AdePT's inability to run ATLAS with VecGeom (solid or surface). The TestEM3 discrepancies aren't yet known—but I think it might just be more sensitive to small differences, and for sure the Heuristic test is *very* sensitive (probably testem3 the most sensitive of the three there).

Issue #1397, opened 2024-09-09
Support scoring of optical photons by Geant4 host sensitive detectors


drbenmorgan:
Related to/broken out of #1351 on the general support for optical boundary processes.

## Original discussion

Geant4's `G4OpBoundaryProcess` class provides a capability for user code to use the Sensitive Detector mechanism to score optical photons that are totally absorbed on a given optical surface. Celeritas should support this capability too, and a first implementation could be to follow what was done with EM scoring by sending `G4Step`s of absorbed photons back to the host and passing them to the appropriate user-defined sensitive detectors on the host. It's assumed here that _diagnostic_ like information like number of steps, process counts, etc will be handled internally by Celeritas. 

The basic algorithm used in Geant4 is within [`G4OpBoundaryProcess::PostStepDoIt`](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-11.2-release/source/processes/optical/src/G4OpBoundaryProcess.cc?ref_type=heads#L146) with the key parts being (omitting quite a bit of detail!):

- It will call [`G4OpBoundaryProcess::DoAbsorption`](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-11.2-release/source/processes/optical/include/G4OpBoundaryProcess.hh?ref_type=heads#L303) when the photon is going to be absorbed on the surface.
  - This checks if the photon will be "detected" given the quantum efficiency of the surface.
  - If "detected" it sets the photon's status flag to [`G4OpBoundaryProcessStatus::Detection`](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-11.2-release/source/processes/optical/include/G4OpBoundaryProcess.hh?ref_type=heads#L72), and marks the track as `fStopAndKill`.
- Right at the end of `PostStepDoIt`, if the photon's status is `G4OpBoundaryProcessStatus::Detection`, then [`G4OpBoundaryProcess::InvokeSD`](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-11.2-release/source/processes/optical/src/G4OpBoundaryProcess.cc?ref_type=heads#L1498) is called with the photon's `G4Step`
  -  This adds the photon energy to the step's local energy deposition, then
  - Determines if the _post step point_ has a Sensitive Detector (i.e. the Logical Volume of the post step point), calling its `Hit` member function if so.

That's pretty straightforward/obvious, but with some things to note (which we've also discussed today):

[...]

PR #1653, opened 2025-03-05
Support post-step touchable location and step status

sethrj:
This enables reconstruction of the post-step status (i.e., `fWorldBoundary` or `fGeomBoundary`) and the post-step touchable. The latter capability is needed for https://github.com/celeritas-project/gpu-delta-review-2025/issues/4 , and the former might be needed for #1397. A new option `locate_touchable_post` in the SD options can disable this capability for performance, if needed.


src/celeritas/ext/detail/LevelTouchableUpdater.cc
```diff
@@ -40,26 +40,25 @@ LevelTouchableUpdater::~LevelTouchableUpdater() = default;
...
+    return (*this)(LevelTouchableUpdater::volume_instances(out, i, step_point),
+                   touchable);
 }
 
 //---------------------------------------------------------------------------//
 /*!
  * Initialize from a span of volume instances.
+ *
+ * Since the volume instances are allowed to be padded to better support GPU, a
+ * null ID terminates the sequence. An empty input or one thst starts with a
```
stognini:
```suggestion
 * null ID terminates the sequence. An empty input or one that starts with a
```

stognini:
We do get an energy flow now, but the very first bin is missing in the Celeritas histogram...

benchem3-eflow

sethrj:
> We do get an energy flow now, but the very first bin is missing in the Celeritas histogram...

Easy answer:
https://github.com/celeritas-project/gpu-delta-review-2025/blob/cf5f7776f53a81097d727b7c3860bd5107c9988c/celeritas/BenchmarkEm3/src/TargetSD.cc#L72
the initial point uses the *pre-step* status, which we can't get.

stognini:
Yep, you posted as I was writing too. That first guy is the culprit, and that would mean changing how we're assigning the step  status in our `HitProcessor`. Which I don't think there is any easy way of doing it.

stognini (approved):
Thanks a lot for adding this! Looks great!

PR #1654, opened 2025-03-06
Approximate G4Torus using inner/outer bounding cylinders

elliottbiondo:
This is small PR that serves as a stop-gap until full torus support can be added.


elliottbiondo:
Unexpected downstream CI errors. Will investigate.

sethrj:
@elliottbiondo This is my bad, see #1655

src/orange/g4org/SolidConverter.cc
```diff
@@ -646,9 +661,19 @@ auto SolidConverter::tet(arg_type solid_base) -> result_type
...
+    CELER_LOG(warning) << "G4Torus is not fully supported; approximating with "
+                          "bounding cylinders";
     auto const& solid = dynamic_cast<G4Torus const&>(solid_base);
-    CELER_DISCARD(solid);
-    CELER_NOT_IMPLEMENTED("torus");
+    auto rmax = scale_(solid.GetRmax());
+    auto rtor = scale_(solid.GetRtor());
+
+    std::optional<Cylinder> inner;
+    inner = Cylinder{rtor - rmax, rmax};
```
sethrj: Is it possible to create a degenerate torus? If so, leave it optionaol; if not, you could write this
```suggestion
    std::optional<Cylinder> inner{std::in_place, rtor - rmax, rmax};
```
and just for reference you can also use `inner.emplace(rtor - rmax, rmax)`. Just good to know but these construct in place rather than constructing a standalone object and using the move constructor `Cylinder(Cylinder&&)`.

sethrj (approved):
Nice and easy!

sethrj:
@elliottbiondo for future reference for you and your student, here's some notes I took a while back about torus implementation:

---

# MCNP/Lava

Elliptical or circular torus with axis alignment:
$$(z - \bar z)^2 / b^2 + \left( \sqrt{(x - \bar x)^2 + (y - \bar y)^2} - a\right)^2 /c^2 - 1 = 0$$

where $\bar x, \bar y, \bar z, a, b, c$ specify the ellipse:

$$ \frac{s^2}{b^2} + \frac{(r-a)^2}{c^2} = 1 $$

rotated about the $s$ axis in the $(r, s)$ cylindrical coordinate system whose center is $(\bar x, \bar y, \bar z)$. For a Z torus, $s = z - \bar z$ and $r = \sqrt{(x - \bar x)^2 + (y - \bar y)^2}$

- $a$ is the distance from the centerline of the torus to the center of its ellipse
- $b$ is the ellipse axis along the main axis (perpendicular to the line between the ellipse center and torus center), $c$ along the minor
- MCNP supports "degenerate" torus: like an ellipsoid with a bump taken out of it ($0 < a < c$)
- Lava multiplies through by $c^2$
- UVW transformation: `{{x,y,z}, {y, z, x}, {z, x, y}}`

Normal:
```c
    double dx = pos[0] - sp->xc;
    double dy = pos[1] - sp->yc;
    double dz = pos[2] - sp->zc;
    double tyz = 1.0 - (sp->a * 1.0 / sqrt(dy*dy + dz*dz));

    nrm[0] = sp->c2_b2 * dx;
    nrm[1] = dy * tyz;
    nrm[2] = dz * tyz;
    normalize_vector(nrm);
```

# Geant4/VecGeom

- Only Z axis
- Only circular
- Centered on origin, not translated
- Intersection of +torus, -torus, and wedge
- VecGeom accelerates using bounding cylindrical segment
- Geant4 *prohibits* degenerate (and nearly degenerate!) torus: fails if $a < c + \epsilon$
- Geant4 uses Jenkins-Traub polynomial root solver

| Variable | MCNP | Geant4    |
| -------- | ---- | --------- |
|  Swept center radius | a | fRtor |
|  Radius along axis | b | fRmax |
|  Radius perp to axis | c | fRmax |
|  Inner radius | — | fRmin |

# Possible simplifications

- Centered on origin: x, y, z are zero
- Circular cross section: b = c
- Degenerate $a=0$: convert to sphere or ellipsoid
- $c < a$: delete shape

src/orange/g4org/SolidConverter.cc
sethrj: Nope not optional.

elliottbiondo:
Another good reference is:

Joseph M. Cychosz, editor: James Arvo,  "V.2 - Intersecting a  Ray with an Elliptical Torus," Graphics Gems II, Morgan Kaufmann, pp. 251-256, 1991.

Which I have a hard copy of for my student.

PR #1659, opened 2025-03-08
Support volume-specific uniform field

amandalund:
This add support for a uniform magnetic field in specific logical volumes, needed for https://github.com/celeritas-project/gpu-delta-review-2025/issues/2. Currently the same field strength is used for each volume, though this could be made more general. Tracks in the uniform along-step action are filtered by whether the volume they are in has a field, and a separate kernel is launched for linear propagation. A list of `G4LogicalVolume`s can be specified in the along-step factory.


src/accel/AlongStepFactory.hh
```diff
@@ -9,16 +9,17 @@
 
 #include <functional>
 #include <memory>
+#include <G4LogicalVolume.hh>
```
sethrj: Should just be forward declared

src/celeritas/inp/Field.hh
```diff
@@ -44,6 +47,9 @@ struct UniformField
 
     //! Field driver options
     FieldDriverOptions driver_options;
+
+    //! Volumes where the field is present (optional)
+    std::vector<VolumeId> volumes;
```
sethrj, line 52: Volume ID isn't really a safe way to do this because (as currently implemented) the IDs change based on the implementation (or even by switching the order of volumes in a file).

The equivalent for SDs uses a variant set:
```c++
    using SetVolume = std::unordered_set<G4LogicalVolume const*>;
    using SetString = std::unordered_set<std::string>;
    using VariantSetVolume = std::variant<SetVolume, SetString>;
```
and since "no volumes" really means "all volumes" I'd say it should default to "monostate", which is the standard way of making a variant optional:
```c++
    using VariantSetVolume = std::variant<std::monostate, SetVolume, SetString>;
```

Let's implement this change after this PR is merged (either merging main into #1632 or as a separate PR).

src/celeritas/track/TrackFunctors.hh
```diff
@@ -0,0 +1,120 @@
...
+                     == static_cast<bool>(track.sim().along_step_action()));
+        return track.sim().along_step_action() == this->action;
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply only to tracks in a volume with a field.
+ */
+struct IsInUniformField
```
sethrj: This and the one below should probably live closer to `field` or `alongstep`, not under `track`.

src/celeritas/track/TrackFunctors.hh
```diff
@@ -0,0 +1,120 @@
...
+        return IsAlongStepActionEqual{action}(track)
+               && IsInUniformField{field}(track);
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Apply to tracks in the uniform along-step action in volumes without field.
+ */
+struct IsAlongStepLinear
```
sethrj: You're not going to define a variadic template functor that returns "all of" the `operator()` of its class arguments? 🙃  (j/k let's not)

sethrj (approved):
Excellent!

src/celeritas/inp/Field.hh
amandalund: Nice! I was wondering about a clean way to provide either LVs or names. Sounds good, I'll hold off on updating this.

PR #1660, opened 2025-03-09
Use CLI11 for front end command line parsing

sethrj:
This weekend refactor replaces the custom input parsing for front-end apps (aside from celer-g4, which I'll do later after removing the `.mac` files) with CLI11 to reduce maintenance burden and improve output.
- New external dependency CLI11, automatically downloaded if unavailable (but added to spack environment too)
- New classes `FileOrStdin`, `FileOrStdout` to abstract the "'-' or filename" usage
- Change RootJsonDumper to use reference to ostream rather than pointer
- Use CLI11 to process input arguments


.github/workflows/build-spack.yml
```diff
@@ -82,6 +82,12 @@ jobs:
           buildcache: true
           color: true
           path: spack-src
+      - name: Patch spack # To be deleted; see associated github pulls
+        working-directory: spack-src
+        run: |
+          for id in 49397; do
+            curl -LfsS "https://github.com/spack/spack/pull/${id}.diff" | patch -p1
+          done
```
sethrj: I can revert this for now since the Spack preinstalled version wasn't working

app/celer-geo/celer-geo.cc
```diff
@@ -258,64 +254,24 @@ void print_usage(std::string_view exec_name)
...
-    }
+    std::string filename;
+    cli.add_option("filename", filename, "Input JSON lines")
+        ->required()
+        ->check(CLI::ExistingFile | dash_validator());
 
-    return EXIT_SUCCESS;
+    // Parse and run
+    CELER_CLI11_PARSE(cli, argc, argv);
+    return run_safely(cli, run, filename);
```
sethrj: Check that this prints exceptions

src/corecel/io/FileOrConsole.hh
```diff
@@ -0,0 +1,106 @@
...
+    //! Implicitly cast to the opened stream
+    operator std::istream&() const { return *instream_; }
+
+    //! Get the filename or renamed placeholder
+    std::string const& filename() const { return filename_; }
+
+  private:
+    std::string filename_;
+    std::ifstream infile_;
+    std::istream* instream_{nullptr};
```
sethrj: Delete copy because of this

src/corecel/io/FileOrConsole.hh
```diff
@@ -0,0 +1,106 @@
...
+    //! Implicitly cast to the opened stream
+    operator std::ostream&() const { return *outstream_; }
+
+    //! Get the filename or renamed placeholder
+    std::string const& filename() const { return filename_; }
+
+  private:
+    std::string filename_;
+    std::ofstream outfile_;
+    std::ostream* outstream_{nullptr};
```
sethrj: Same

app/celer-export-geant.cc
```diff
@@ -154,65 +144,45 @@ void run(std::string const& gdml_filename,
...
+    cli.add_option("options", opts_filename, "Options JSON file")
+        ->check(CLI::ExistingFile | dash_validator()
+                | empty_string_validator());
+    auto* output_opt = cli.add_option(
+        "output",
+        out_filename,
+        R"(Output file (ROOT or JSON or '-' for stdout JSON)");
+    cli.add_flag("--gen-test", gen_test, "Generate test data");
+    cli.add_flag("--dump-default", dump_default, "Dump default options")
+        ->excludes(output_opt);
```
amandalund: Could we allow dumping the default options without requiring a GDML?

app/celer-sim/celer-sim.cc
```diff
@@ -159,121 +165,88 @@ std::shared_ptr<OutputRegistry> run(std::istream* is)
...
-    std::ifstream infile;
-    std::istream* instream = nullptr;
-    if (filename == "-")
-    {
-        instream = &std::cin;
-        filename = "<stdin>";  // For nicer output on failure
-    }
-    else
+    // Parse and run
+    CELER_CLI11_PARSE(cli, argc, argv);
```
amandalund: I guess since there are no required options we don't see the parse error and usage printed when running without any arguments, just a runtime error? Should we require that either the input json or one of the flags be provided?

amandalund (approved):
Nice improvement to the input parsing!

app/celer-export-geant.cc
sethrj: Yeah CLI11's option setup is a little limited. I've updated with a manual check.

app/celer-sim/celer-sim.cc
sethrj: Fixed.

PR #1661, opened 2025-03-10
Fix link errors with CUDA RDC and `--as-needed` flag

pcanal:
`-ldl` is neeeded for any library/executable that included the result of nvlink.  Rather than relying on the user to add it, let's add it as part the RDC wrappers.

Note: Without this Celeritas failed to link many exectuable if the user had set LD_FLAGS to include `-Wl,--as-needed\' (some conda environemnt do that ... )

### DIFF
--- a/cmake/CudaRdcUtils.cmake
+++ b/cmake/CudaRdcUtils.cmake
@@ -925,9 +925,7 @@ function(cuda_rdc_target_link_libraries target)
         set_target_properties(${_target_final} PROPERTIES CUDA_RUNTIME_LIBRARY "Shared")
         set_target_properties(${_target_object} PROPERTIES CUDA_RUNTIME_LIBRARY "Shared")
       endif()
-      get_target_property(_link_libraries_final ${_target_final} INTERFACE_LINK_LIBRARIES)
-      list(APPEND _link_libraries_final ${CMAKE_DL_LIBS})
-      set_target_properties(${_target_final} PROPERTIES LINK_LIBRARIES "${_link_libraries_final}")
+      set_property(TARGET ${_target_final} APPEND PROPERTY LINK_LIBRARIES ${CMAKE_DL_LIBS})
     endif()
   else() # We could restrict to the case where the dependent is a static library ... maybe
     set_target_properties(${target} PROPERTIES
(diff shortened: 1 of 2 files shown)


cmake/CudaRdcUtils.cmake
```diff
@@ -925,6 +925,9 @@ function(cuda_rdc_target_link_libraries target)
         set_target_properties(${_target_final} PROPERTIES CUDA_RUNTIME_LIBRARY "Shared")
         set_target_properties(${_target_object} PROPERTIES CUDA_RUNTIME_LIBRARY "Shared")
       endif()
+      get_target_property(_link_libraries_final ${_target_final} INTERFACE_LINK_LIBRARIES)
+      list(APPEND _link_libraries_final ${CMAKE_DL_LIBS})
+      set_target_properties(${_target_final} PROPERTIES LINK_LIBRARIES "${_link_libraries_final}")
```
sethrj:
```suggestion
      set_property(TARGET ${_target_final} APPEND PROPERTY LINK_LIBRARIES ${CMAKE_DL_LIBS})
```

src/corecel/CMakeLists.txt
```diff
@@ -213,8 +213,6 @@ if(CELERITAS_USE_CUDA)
     sys/ScopedLimitSaver.cuda.cc
     sys/ScopedProfiling.cuda.cc
   )
-  # NVTX requires `-ldl`
-  list(APPEND PRIVATE_DEPS ${CMAKE_DL_LIBS})
```
sethrj, line 217: This should be left alone; we still use nvtx when RDC isn't in the picture.

pcanal: Humm ... I expect (it is likely that) if this is needed it will also fail (not be sufficient) when `as-needed` is used.  Which configuration needs this (I am guessing cuda+orange)?

sethrj: "As needed" is only a problem because of the multiple extra incomplete linking steps. With orange CUDA it links directly into the shared lib and is good, or for static libs cmake will propagate it into the final link step.

pcanal: It does indeed fail (without those lines) with cuda+orange+static_libraries ... but with also VecGeom enabled and thus with RDC.  So this PR is not yet fixing the RDC+static_library+as_needed case.

pcanal: So far, it looks like `-ldl` is added in the cuda+only_orange+static_libraries even without those lines (even without this PR (of course)).   `cmake --trace-expand` shows that `FindCUDAToolkit.cmake`  at least in cmake version 3.26.5 (line 1044 and 1201).  So maybe this is needed only for older version of CMake or CUDA sdk.  I will re-add but with a comment.

sethrj (approved):
Thanks!

PR #1662, opened 2025-03-11
Add 3D Cylindrical field map

esseivaju:
Support for 3D field interpolation in a R-Phi-Z map. The implementation is based on the existing RZMapField: we add a nonuniform grid for the phi coordinate and perform trilinear interpolation instead of the bilinear interpolation in RZ.

There is no support for JSON input, instead we provide an utility to sample an existing `G4Field`


src/celeritas/field/RZPhiMapFieldInput.hh
```diff
@@ -0,0 +1,93 @@
...
+ */
+struct RZPhiMapFieldInput
+{
+    unsigned int num_grid_z{};
+    unsigned int num_grid_r{};
+    unsigned int num_grid_phi{};
+    double min_z{};  //!< Lower z coordinate [len]
+    double max_z{};  //!< Last z coordinate [len]
+    double min_r{};  //!< Lower r coordinate [len]
+    double max_r{};  //!< Last r coordinate [len]
```
sethrj: We can definitely use nonuniform grids for any/all of these if we want, implementing on device using `NonuniformGrid`.

src/celeritas/field/RZPhiMapFieldInput.hh
```diff
@@ -0,0 +1,93 @@
...
+    double min_phi{};  //!< Lower phi coordinate [rad]
+    double max_phi{};  //!< Last phi coordinate [rad], typically 2π for full
+                       //!< coverage Note: For a full circle (max_phi - min_phi
+                       //!< = 2π), the field is treated as periodic, with
```
sethrj: To avoid complexity in the build system (including doxygen potentially!), let's avoid unicode characters in source files (especially doxygen comments) until at least C++20. Either encode as HTML `&#x03C0;` or `&pi;` or `\f$\pi\f$`.

src/celeritas/field/RZPhiMapFieldInputIO.json.cc
```diff
@@ -0,0 +1,197 @@
...
+    RZPFI_LOAD(field_phi);
+    if (j.contains("driver_options"))
+    {
+        RZPFI_LOAD(driver_options);
+    }
+
+    // Convert unit systems based on input
+    UnitSystem length_units{UnitSystem::cgs};  // cm
+    UnitSystem field_units{UnitSystem::si};  // tesla
+    if (auto iter = j.find("_units"); iter != j.end())
```
sethrj: Let's extract this block to a "unit utils .json.hh,cc" perhaps to reuse for the other file? Or at the very least let's delete the deprecated part and require a single consistent unit system for this input.

src/celeritas/field/RZPhiMapFieldParams.cc
```diff
@@ -0,0 +1,138 @@
...
+        host.grids.data_z = UniformGridData::from_bounds(
+            inp.min_z, inp.max_z, inp.num_grid_z);
+
+        // For phi, ensure we're creating a periodic grid
+        // If the input specifies a full circle, adjust the grid to handle
+        // periodicity
+
+        bool is_full_circle = soft_zero(
+            std::fabs((inp.max_phi - inp.min_phi) - 2. * constants::pi));
+        if (is_full_circle)
```
sethrj: For the input and runtime let's use `vector<Turn>` to avoid roundoff issues from pi: everything's just an exactly FP-representable fraction of a circle. Check out `SolidEnclosedAngle` in `orange/orangeinp/Solid.{hh,cc}` for examples.

src/celeritas/field/RZPhiMapField.hh
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    // Perform trilinear interpolation for each field component
+    // Define the interpolation weights
+    real_type wr0 = 1.0 - interp_r.fraction;
+    real_type wr1 = interp_r.fraction;
+    real_type wz0 = 1.0 - interp_z.fraction;
+    real_type wz1 = interp_z.fraction;
+    real_type wphi0 = 1.0 - interp_phi.fraction;
+    real_type wphi1 = interp_phi.fraction;
```
sethrj: Er, no offense meant since I think you said you used AI to help, but did claude generate this code? We want to avoid `1.0` since we'll probably want to do these calculations with single precision, and avoid the register usage of having extra variables for `1 - x`.

src/celeritas/field/RZPhiMapField.hh
```diff
@@ -0,0 +1,231 @@
...
+    FindInterp<real_type> interp_z = find_interp<UniformGrid>(grid_z_, pos[2]);
+
+    // Special handling for phi to account for periodicity
+    FindInterp<real_type> interp_phi;
+
+    // Check if we have a full circle (max_phi ~= min_phi + 2π)
+    bool is_full_circle
+        = soft_zero(std::fabs((grid_phi_.back() - grid_phi_.front())
+                              - 2. * constants::pi))
+          || soft_zero(std::fabs((grid_phi_.back() - grid_phi_.front())));
```
sethrj: Perhaps the easiest solution is to require a full circle? If we have a nonuniform grid and they want a partial circle (which seems an unlikely use case) they could add extra points with zeros.

sethrj:
pseudocode from discussion for r/z/phi:

```c++

namespace accel
{

void fill_field(FieldInput& field)
{
    // assert r, z, phi are there
    // resize field, loop over points, store
}

}

namespace inp
{
class FieldInput
{
    using real_type = float;

    vector<real_type> r_grid; // native units (?)
    vector<real_type> z_grid; // native units (?)
    vector<Turn> phi_grid; // turns

    vector<Array<fast_real_type, 3>> field;
};
}

namespace gpu
{
class FieldData
{
    Array<size_type, 3> grid_size; // r, z, phi

    using real_type = float;
    Items<real_type> field;
};

Real3 const& get_field(Real3  const& point)
{
    // calc r, z, phi
    // look up i,j,k
    HyperslabIndexer get_flattened_index{data.grid_size};
    return field[get_flattened_index(i, j, k)];
}
}
```

src/celeritas/field/RZPhiMapFieldInputIO.json.cc
esseivaju: removed for now since we won't do json input in that PR

src/celeritas/field/RZPhiMapFieldParams.cc
esseivaju: with `vector<Turn>`, and equivalently `NonuniformGrid<Turn>`, we have grids of different underlying types for r, z, and phi.  From discussing with @amandalund, I understand you usually have a single collection as backing storage for each grid. If we were to do this here, we'd lose the type safety on `Turn`

src/celeritas/field/RZPhiMapField.hh
esseivaju: requiring a full circle now, I check that the first / last turn are 0, and 1, respectively

esseivaju:
This still need some work, (integration in access, unit tests) but to accelerate this PR, you can already review the current status

src/celeritas/field/RZPhiMapFieldData.hh
```diff
@@ -0,0 +1,114 @@
...
+
+    inline CELER_FUNCTION ElementId id(size_type idx_z,
+                                       size_type idx_r,
+                                       size_type idx_phi) const
+    {
+        CELER_EXPECT(grids);
+        // Index with ordering [Z][R][Phi]
+        return ElementId(idx_z * grids.r.size() * grids.phi.size()
+                         + idx_r * grids.phi.size() + idx_phi);
+    }
```
esseivaju: Should we replace this with the `HyperslabIndexer` (and same in `RZMap`)?

src/celeritas/field/RZPhiMapFieldParams.cc
sethrj: Yeah for quantities on device we usually use `real_type` on the backend anyway, and convert when we pull the data from global memory. I think we could do that here too? (Our grids aren't exactly quantity-friendly.)

src/celeritas/field/RZPhiMapFieldData.hh
sethrj: There are a bunch of places where we do this and yes please do. Also, I know RZ map indexes as [Z][R] which I didn't realize during the original pull, but let's keep the accessors C-like: [Phi][R][Z].

src/celeritas/field/RZPhiMapFieldParams.cc
esseivaju: With the tweak to `find_interp`, a `Grid<Quantity>` would be fine.

esseivaju:
I think this is ready. I have the fullsimlight integration [here](https://github.com/celeritas-project/gpu-delta-review-2025/blob/5964d0a65742a29736bbe49edfef1bebf1d597b6/ATLAS/GeoModel/FullSimLight/fullSimLight.cc#L248-L283)

src/accel/RZPhiMapMagneticField.hh
```diff
@@ -0,0 +1,172 @@
...
+ * sample it.
+ */
+class RZPhiMapFieldSampler
+{
+  public:
+    // Generate field with user-defined grid
+    RZPhiMapFieldParams::Input
+    operator()(std::vector<real_type> r_grid,
+               std::vector<real_type> z_grid,
+               std::vector<real_type> phi_values) const
```
sethrj: Const references, and this should be in a .cc file

src/accel/RZPhiMapMagneticField.hh
```diff
@@ -0,0 +1,172 @@
...
+
+        field_input.field_r.resize(total_points);
+        field_input.field_phi.resize(total_points);
+        field_input.field_z.resize(total_points);
+
+        Array<size_type, 3> const dims{nphi, nr, nz};
+        HyperslabIndexer const flat_index{dims};
+        auto& field = *G4TransportationManager::GetTransportationManager()
+                           ->GetFieldManager()
+                           ->GetDetectorField();
```
sethrj: We should add assertions in here , no telling when some of these may be uninitailized

src/accel/RZPhiMapMagneticField.hh
```diff
@@ -0,0 +1,172 @@
...
+                    field.GetFieldValue(pos.data(), bfield.data());
+
+                    // values in cylindrical vector space
+                    field_input.field_r[idx] = convert_from_geant(
+                        bfield[0] * std::cos(phi) + bfield[1] * std::sin(phi),
+                        clhep_field);
+                    field_input.field_phi[idx] = convert_from_geant(
+                        -bfield[0] * std::sin(phi) + bfield[1] * std::cos(phi),
+                        clhep_field);
+                    field_input.field_z[idx]
```
sethrj: Is it not better to have the 'field input' be a 3-D vector field than having three separate inputs? so [nr][nphi][nz][3], with maybe an enum for the cylindrical axes

sethrj: Also a helper function to convert cartesian->polar/cyl would be good here.

src/accel/SetupOptions.cc
```diff
@@ -132,6 +133,12 @@ void ProblemSetup::operator()(inp::Problem& p) const
         CELER_LOG(debug) << "Getting RZ map field";
         p.field = u->get_field();
     }
+    else if (auto* u
+             = so.make_along_step.target<RZPhiMapFieldAlongStepFactory>())
+    {
+        CELER_LOG(debug) << "Getting RZPhi map field";
```
sethrj: Yeah this won't be used till #1632 sorry for the confusion:(

src/corecel/grid/FindInterp.hh
```diff
@@ -19,11 +21,10 @@ namespace celeritas
  * The resulting index will be in [0, grid.size() - 1)
  * and the fraction will be in [0, 1).
  */
-template<class T>
 struct FindInterp
 {
     size_type index{};  //!< Lower index into the grid
-    T fraction{};  //!< Fraction of the value between its neighbors
+    real_type fraction{};  //!< Fraction of the value between its neighbors
```
sethrj: I think we want to leave this templated? Particularly because for the field I think we want to store and compute in single precision.

test/celeritas/field/Fields.test.cc
```diff
@@ -147,6 +157,146 @@ TEST_F(RZMapFieldTest, all)
...
+                for (real_type f : field)
+                {
+                    actual.push_back(f);
+                }
+            }
+        }
+    }
+
+    // Expected field values at the 8 sample points (2×2×2 grid in r, phi, z)
+    static real_type const expected_field[] = {// r=10cm, phi=0, z=-100cm
```
sethrj: The formatting here got borked

src/celeritas/field/RZPhiMapField.hh
```diff
@@ -0,0 +1,183 @@
...
+    if (r != 0)
+    {
+        cos_phi = pos[0] / r;
+        sin_phi = pos[1] / r;
+    }
+    value[0] = field_r * cos_phi - field_phi * sin_phi;
+    value[1] = field_r * sin_phi + field_phi * cos_phi;
+
+    return value;
+}
```
sethrj: After all the madness of the next two weeks is done it would be interesting to look at how refactoring this might be able to reduce register usage and see how that correlates to performance. Great job on getting this to work!

sethrj (commented):
Couple of comments; I'm adding Quantity support for findinterp and templating Turn on a different branch.

src/celeritas/field/RZPhiMapField.hh
amandalund: We could also probably reduce some of the duplication here—e.g. by storing the `fieldmap` as an array of `Real3` instead of `struct RZPhiMapElement` and iterating over the coordinates.

src/celeritas/field/RZPhiMapFieldData.hh
```diff
@@ -0,0 +1,133 @@
...
+    ItemRange<real_type> phi;  //!< Index range for phi
+    ItemRange<real_type> r;  //!< Index range for r
+    ItemRange<real_type> z;  //!< Index range for z
+
+    //! Check whether the data is assigned
+    explicit inline CELER_FUNCTION operator bool() const
+    {
+        return !storage.empty() && grid_size[0] > 1 && grid_size[1] > 1
+               && grid_size[2] > 1 && phi.size() == grid_size[0]
+               && r.size() == grid_size[1] && z.size() == grid_size[2];
```
amandalund: I wish the `HyperslabIndexer` didn't require creating an extra array for the dimensions... because of that it often looks cleaner/clearer/requires less code just to write out the index calculation (@sethrj).

sethrj: I'm more than happy to redesign the multi-D indexer (new names, arrays by value instead of reference, overloading `operator()` instead of having two separate classes, or just naming them instead of making them function-like). I think we'd also get improved clarity here if we had 
```c++
enum class CylAxis
{
    phi,
    r,
    z
    size_
};
```
with `EnumArray<CylAxis, ItemRange<real_type>> grid`, and (once we pass values into the indexer) a helper function `Indexer::from_sizes(...)` or something.

esseivaju: I'll refactor that to an `EnumArray`

src/celeritas/field/RZPhiMapField.hh
esseivaju: That would make sense, I was following the scheme we have for the `RZMapField`, but for 2D the duplication isn't as bad

src/celeritas/field/RZPhiMapFieldData.hh
esseivaju: should we order it as R-Phi-Z, which is the standard ordering?

src/celeritas/field/RZPhiMapFieldData.hh
```diff
@@ -0,0 +1,126 @@
...
+                                       size_type idx_r,
+                                       size_type idx_z) const
+    {
+        CELER_EXPECT(grids);
+        // HyperSlabIndexer does not take Array<T const>
+        Array<size_type, static_cast<size_type>(CylAxis::size_)> tmp{
+            grids.grid_size[CylAxis::Phi],
+            grids.grid_size[CylAxis::R],
+            grids.grid_size[CylAxis::Z]};
+        return ElementId{HyperslabIndexer{tmp}(idx_phi, idx_r, idx_z)};
```
esseivaju: not the cleanest, I think we have to create a tmp array for now because `HyperslabIndexer` doesn't take `Array<T const>`...

src/celeritas/field/RZPhiMapField.hh
```diff
@@ -0,0 +1,155 @@
...
+    Real3 value{0, 0, 0};
+
+    // Convert Cartesian to cylindrical coordinates
+    real_type r = hypot(pos[0], pos[1]);
+    // Ensure phi is in [0, 2\f$\pi\f$)
+    real_type phi = atan2(pos[1], pos[0]);
+    if (phi < 0)
+    {
+        phi += TwoPi::value();
+    }
```
esseivaju: Not using `atan2Turn` because I am expecting a turn between [0;1] instead of [-0.5;0.5]

esseivaju (commented):
I think I addressed most of the comments

src/celeritas/field/RZPhiMapField.hh
sethrj: You'd have to do the same thing right?
```suggestion
    auto phi = atan2turn(pos[1], pos[0]);
    if (phi < zero_quantity())
    {
        phi += Turn_t<real_type>{1};
    }
```
The whole point of using Turns is you don't have to accumulate floating point error by mucking with pi.

src/celeritas/Types.hh
```diff
@@ -191,6 +191,16 @@ enum class NuclearFormFactorType
     size_
 };
 
+//---------------------------------------------------------------------------//
+//! Cylindrical coordinates indices
+enum class CylAxis
+{
+    Phi = 0,
+    R = 1,
+    Z = 2,
```
sethrj: To conform to our style conventions:
```suggestion
    phi = 0,
    r = 1,
    z = 2,
```

src/celeritas/field/RZPhiMapFieldData.hh
sethrj: Yeah correct. Future work we'll improve this.

src/celeritas/field/RZPhiMapFieldParams.cc
```diff
@@ -0,0 +1,121 @@
...
+        host.grids.grid_size[CylAxis::Phi] = inp.grid_phi.size();
+        host.grids.grid_size[CylAxis::R] = inp.grid_r.size();
+        host.grids.grid_size[CylAxis::Z] = inp.grid_z.size();
+
+        auto grid = make_builder(&host.grids.storage);
+        grid.reserve(inp.grid_phi.size() + inp.grid_r.size()
+                     + inp.grid_z.size());
+        std::transform(inp.grid_phi.cbegin(),
+                       inp.grid_phi.cend(),
+                       std::back_inserter(grid),
```
sethrj: Neat trick!

sethrj (approved):
Amazing work! We can (should) wait until a follow-on to optimize by using single-precision field.

src/celeritas/field/RZPhiMapFieldData.hh
sethrj: The "standard" ordering is, based on my experience, a holdover from fortran indexing. We have the chance to start clean and we should take it :)

src/celeritas/field/RZPhiMapField.hh
esseivaju: Maybe it's actually better to use turn in [-0.5;0.5], that will be one less branch on GPU

src/celeritas/field/RZPhiMapFieldData.hh
amandalund: If we're creating a temporary array anyway, could we remove `grid_size` from the `RZPhiMapGridData` and use  `tmp{grids.axes[CylAxis::Phi].size(), ...}` instead?

src/celeritas/field/RZPhiMapFieldData.hh
esseivaju: should I rename the map `PhiRZ` then?

sethrj: Oh I didn't notice the ordering was flipped around from the name 😭 Maybe we should just call it CylFieldMap... and the RZ one SimpleCylFieldMap (because it's both the field vector and the grid that are azimuthally symmetric)? 

Let's order it naturally: `[x][y][z]` if it's cartesian, `[r][phi/theta][z]` or `[r][z]` if cylindrical, `[r][theta][phi]` if polar.

In my old code base we'd order a cartesian grid `data[z][y][x]` so it would have the same stride pattern as a fortran `data(x, y, z)` grid. I think our `RZField`, which is based on CMS data, uses `[z][r]` for the same reason (and I'd like to change it before v1).

esseivaju:
> optimize by using single-precision field

Do you mean just the interpolation would be done in single precision, but we are still storing the grid, and field using real precision?

sethrj:
> Do you mean just the interpolation would be done in single precision, but we are still storing the grid, and field using real precision?

Everything can/should be stored in single precision. I don't think the grid will ever be fine enough or the field accurate enough to require double. Future optimization work could also actually store the map as a 3D [texture](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#texture-memory) and use built-in CUDA/HIP interpolation for performance.

src/accel/CylMapMagneticField.cc
```diff
@@ -0,0 +1,134 @@
...
+        for (size_type iphi = 0; iphi < nphi; ++iphi)
+        {
+            real_type phi = phi_values[iphi];
+            for (size_type iz = 0; iz < nz; ++iz)
+            {
+                auto* cur_bfield = field_input.field.data()
+                                   + flat_index(ir, iphi, iz, 0);
+                Array<G4double, 4> pos
+                    = {r * std::cos(phi), r * std::sin(phi), z_grid[iz], 0};
+                field.GetFieldValue(pos.data(), bfield.data());
```
sethrj, line 118: Future follow-on work: let's add a little wrapper class in `celeritas/ext` that wraps the Geant4 field manager and returns a Celeritas type.

src/accel/CylMapMagneticField.hh
```diff
@@ -0,0 +1,85 @@
...
+    CylFieldMap calc_field_;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with the Celeritas shared CylFieldMapParams.
+ */
+CylMapMagneticField::CylMapMagneticField(SPConstFieldParams params)
+    : params_(std::move(params))
+    , calc_field_(CylFieldMap{params_->ref<MemSpace::native>()})
```
sethrj: Also follow-on work: we could probably make a template class that does this for both RZ and Cyl.

src/celeritas/alongstep/AlongStepCylFieldMapMscAction.hh
```diff
@@ -0,0 +1,99 @@
...
+    void step(CoreParams const&, CoreStateHost&) const final;
+
+    // Launch kernel with device data
+    void step(CoreParams const&, CoreStateDevice&) const final;
+
+    //! ID of the model
+    ActionId action_id() const final { return id_; }
+
+    //! Short name for the interaction kernel
+    std::string_view label() const final { return "along-step-CylMap-msc"; }
```
sethrj: find-replace error I think
```suggestion
    std::string_view label() const final { return "along-step-cylmap-msc"; }
```

src/celeritas/field/CylFieldMap.hh
```diff
@@ -0,0 +1,151 @@
...
+
+    Real3 value{0, 0, 0};
+
+    // Convert Cartesian to cylindrical coordinates
+    real_type r = hypot(pos[0], pos[1]);
+    // Ensure phi is in [0, 2\f$\pi\f$)
+    Turn phi = atan2turn(pos[1], pos[0]);
+    if (phi < zero_quantity())
+    {
+        phi.value() += 1;
```
sethrj: I forgot that this worked :)

sethrj (approved):
Awesome work!

sethrj:
Follow-on work:
- Single precision, or possibly even texture map
- Field conversion tool: https://github.com/celeritas-project/celeritas/pull/1662#discussion_r2000883142
- Use templating to improve code reuse: https://github.com/celeritas-project/celeritas/pull/1662#discussion_r2000886091

PR #1663, opened 2025-03-12
Move offload template into a celeritas::example namespace

sethrj:
In preparation for a new rather complicated class in the template, I'd like to move the example code into a sub-namespace to avoid potential conflicts and as a best practice.


stognini (approved):
Looks good to me.

PR #1664, opened 2025-03-12
Add an example user action for gathering statistics on GPU

sethrj:
This defines a new set of classes in the `offload-template` example that accumulates step statistics on either device or host using an action that's defined in the user space. There were two additions to `accel` that make this possible:
- The `Integration` classes now have `GetParams` and `GetState` to access the celeritas core params/state (the first being owned by a static global, the second owned by the thread-local transporter).
- A new `add_user_actions` callback has been added to `SetupOptions` that allows the user to modify/extend the core before the states are created.

At some point (probably after further discussion) we will want to factor out most of the code into a templated class that will let users just define the `Data` and `Executor` classes and possibly provide some helpers to make the Data definition/initialization easier.

Output:
```
G4WT3 > EventAction.cc:77: [4/4] info: In event 3:
  average step length = 3.07115 mm
  energy deposition = 62.8922 MeV
  with 13 primaries
  and 55 secondaries
```


src/accel/IntegrationBase.cc
```diff
@@ -10,17 +10,24 @@
...
+using celeritas::detail::IntegrationSingleton;
+
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Edit options before starting the run.
+ * Set options before starting the run.
+ *
+ * This captures the input to indiate that options cannot be modified after
```
mrguilima:
```suggestion
 * This captures the input to indicate that options cannot be modified after
```

example/offload-template/src/StepDiagnostic.hh
```diff
@@ -0,0 +1,117 @@
...
+    static std::shared_ptr<StepDiagnostic>
+    make_and_insert(CoreParams const& core);
+
+    // Construct with IDs and filename base
+    StepDiagnostic(ActionId action_id, AuxId aux_id);
+
+    //// USER INTEGRATION ////
+
+    // Get the statistics and reset them
+    StepStatistics GetAndReset(CoreStateInterface& state) const;
```
mrguilima, line 90: Why not get_and_reset()?

example/offload-template/src/StepDiagnostic.hh
```diff
@@ -0,0 +1,117 @@
...
+    std::string_view description() const final { return sad_.description(); }
+    //!@}
+
+    //!@{
+    //! \name Step action interface
+
+    //! Index of this class instance in its registry
+    ActionId action_id() const final { return sad_.action_id(); }
+    //! Index of this class instance in its registry
+    StepActionOrder order() const final { return StepActionOrder::user_post; }
```
mrguilima: Are those doc strings intended to be identical?

mrguilima (commented):
Just a few comments.

sethrj:
CI failure requires https://github.com/celeritas-project/celeritas/pull/1665 to fix.

example/offload-template/src/StepDiagnostic.hh
sethrj: For our Geant4 interface code I've tried to be consistent with making symbol names "Geant4"-like, and internal ones "celeritas-like". This is the only method in this class that should be called externally.

example/offload-template/src/StepDiagnostic.cc
```diff
@@ -0,0 +1,161 @@
...
+/*!
+ * Get the statistics and reset them.
+ */
+StepStatistics StepDiagnostic::GetAndReset(CoreStateInterface& state) const
+{
+    // Kernel-collected statistics copied to host memory
+    NativeStepStatistics data;
+    HostStepStatistics host_data;
+
+    auto try_copy = [&](auto* core_state) -> bool {
```
stognini: Is it just me or this seems like it could be a separate function? Since the `StepDiagnostic` will be abstracted in the future, seems like this (fetch step state data for the user) is already a useful abstraction "placeholder".

example/offload-template/src/StepDiagnostic.hh
sethrj: Nope, good catch :)

sethrj: Yeah, they're the ID into two differnt registries, but I'll make it more explicit

sethrj: Nope, that's definitely a typo!

example/offload-template/src/StepDiagnosticData.hh
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Only some of the data (\c NativeStepStatistics) is updated on device. The
+ * number of steps must be changed outside the kernel.
+ */
+template<Ownership W, MemSpace M>
+struct StepStateData
+{
+    //// TYPES ////
+
+    using size_type = unsigned long long;
```
amandalund: Is this used anywhere? Should it be in the `HostStepStatistics`?

example/offload-template/src/StepDiagnosticExecutor.hh
```diff
@@ -0,0 +1,58 @@
...
+    // Same here but non-const (state data can change)
+    NativeRef<StepStateData> const state;
+
+    inline CELER_FUNCTION void
+    operator()(celeritas::CoreTrackView const& track);
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Collect distribution of steps per track for each particle type.
```
amandalund: Copy-paste error.

example/offload-template/src/StepDiagnostic.hh
```diff
@@ -0,0 +1,117 @@
...
+                             public AuxParamsInterface,
+                             public ParamsDataInterface<StepParamsData>
+{
+  public:
+    // Construct and add to core params
+    static std::shared_ptr<StepDiagnostic>
+    make_and_insert(CoreParams const& core);
+
+    // Construct with IDs and filename base
+    StepDiagnostic(ActionId action_id, AuxId aux_id);
```
amandalund:
```suggestion
    // Construct with IDs
    StepDiagnostic(ActionId action_id, AuxId aux_id);
```

example/offload-template/src/StepDiagnostic.hh
```diff
@@ -0,0 +1,117 @@
...
+    // Build state data for a stream
+    UPState create_state(MemSpace m, StreamId id, size_type size) const final;
+    //!@}
+
+    //!@{
+    //! \name Data interface
+
+    //! Access physics properties on the host
+    HostRef const& host_ref() const final { return mirror_.host_ref(); }
+    //! Access physics properties on the device
```
amandalund:
```suggestion
    //! Access shared setup data on the host
    HostRef const& host_ref() const final { return mirror_.host_ref(); }
    //! Access shared setup data on the device
```

example/offload-template/src/StepDiagnostic.cc
```diff
@@ -0,0 +1,179 @@
...
+    actions.insert(result);
+    aux.insert(result);
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with IDs and filename base.
+ *
+ * This also writes to the "metadata" JSON suffix.
```
amandalund:
```suggestion
 * Construct with IDs.
```

amandalund (approved):
Nice example, thanks @sethrj! Some more documentation on adding an action (in the example diagnostic classes themselves and in the docs) could be helpful for a user, but that can be added later.

example/offload-template/src/StepDiagnosticData.hh
sethrj: Yep that's where it was meant to be!

sethrj:
Thanks @amandalund ! I've added a little bit in the way of documentation. Next step after this will be extending this and integrating into @esseivaju 's copy of FullSimLight... might not happen in time but we'll try!

PR #1665, opened 2025-03-12
Fix downstream RDC linking and add versioning

pcanal:
This fixes linking of certain downstream applications to celeritas:

From https://github.com/celeritas-project/celeritas/actions/runs/13858181579/job/38780556004:
<details><summary>error message</summary><pre>
FAILED: run-offload 
: && /usr/bin/c++   CMakeFiles/run-offload.dir/main.cc.o CMakeFiles/run-offload.dir/src/ActionInitialization.cc.o CMakeFiles/run-offload.dir/src/DetectorConstruction.cc.o CMakeFiles/run-offload.dir/src/MakeCelerOptions.cc.o CMakeFiles/run-offload.dir/src/PrimaryGeneratorAction.cc.o CMakeFiles/run-offload.dir/src/RunAction.cc.o CMakeFiles/run-offload.dir/src/SensitiveDetector.cc.o CMakeFiles/run-offload.dir/CMakeFiles/run-offload_emptyfile.cu.o CMakeFiles/run-offload.dir/cmake_device_link.o -o run-offload -L/usr/local/cuda/targets/x86_64-linux/lib -Wl,-rpath,/opt/view/lib64:/__w/celeritas/celeritas/install/lib64:/opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/zlib-ng-2.2.1-vykawn4q5senw3ogkdjg4r5xqrwba73t/lib:/opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/expat-2.6.4-iciygczsk5t6jly2v4qsyimjim6yj75x/lib:/opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/clhep-2.4.7.1-3tp5xebivqxft3yw3wonb3nxql5u7mfo/lib:/opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/xerces-c-3.3.0-plthos7ru3cpg4fnxksrxgya5xioye64/lib  /opt/view/lib64/libG4Tree.so  /opt/view/lib64/libG4FR.so  /opt/view/lib64/libG4GMocren.so  /opt/view/lib64/libG4visHepRep.so  /opt/view/lib64/libG4RayTracer.so  /opt/view/lib64/libG4VRML.so  /opt/view/lib64/libG4vis_management.so  /opt/view/lib64/libG4modeling.so  /opt/view/lib64/libG4interfaces.so  /opt/view/lib64/libG4persistency.so  /opt/view/lib64/libG4error_propagation.so  /opt/view/lib64/libG4readout.so  /opt/view/lib64/libG4physicslists.so  /opt/view/lib64/libG4tasking.so  /opt/view/lib64/libG4parmodels.so  /__w/celeritas/celeritas/install/lib64/libaccel.so  /opt/view/lib64/libG4run.so  /opt/view/lib64/libG4event.so  /opt/view/lib64/libG4tracking.so  /opt/view/lib64/libG4processes.so  /opt/view/lib64/libG4analysis.so  /opt/view/lib64/libG4tools.so  /opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/zlib-ng-2.2.1-vykawn4q5senw3ogkdjg4r5xqrwba73t/lib/libz.so  /opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/expat-2.6.4-iciygczsk5t6jly2v4qsyimjim6yj75x/lib/libexpat.so  /opt/view/lib64/libG4digits_hits.so  /opt/view/lib64/libG4track.so  /opt/view/lib64/libG4particles.so  /__w/celeritas/celeritas/install/lib64/libceleritas.so  /__w/celeritas/celeritas/install/lib64/libgeocel.so  /opt/view/lib64/libG4geometry.so  /opt/view/lib64/libG4materials.so  /opt/view/lib64/libG4graphics_reps.so  /opt/view/lib64/libG4intercoms.so  /opt/view/lib64/libG4global.so  /opt/view/lib64/libG4ptl.so.0.0.2  /opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/clhep-2.4.7.1-3tp5xebivqxft3yw3wonb3nxql5u7mfo/lib/libCLHEP-2.4.7.1.so  /opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/xerces-c-3.3.0-plthos7ru3cpg4fnxksrxgya5xioye64/lib/libxerces-c.so  /opt/view/lib64/libvecgeomcuda_objects.so  /opt/view/lib64/libvecgeom.so  -lrt  -lpthread  /__w/celeritas/celeritas/install/lib64/libcorecel.so  -lcudadevrt  -lcudart  -Wl,-rpath-link,/opt/view/lib64 && :
/usr/bin/ld: /opt/view/lib64/libvecgeomcuda_objects.so: undefined reference to `__cudaRegisterLinkedBinary_89b5f6b6_14_UnplacedBox_cu_2b4adfca_685700'
/usr/bin/ld: /opt/view/lib64/libvecgeomcuda_objects.so: undefined reference to `__cudaRegisterLinkedBinary_46fc9466_15_PlacedSphere_cu_2b4adfca_685854'
</pre></details>

Look at both [INTERFACE_]LINK_LIBRARIES in CudaRdc when determining the list of dependencies.


cmake/CudaRdcUtils.cmake
```diff
@@ -591,6 +591,11 @@ function(cuda_rdc_depends_on OUTVARNAME lib potentialdepend)
     get_target_property(_lib_target_type ${lib} TYPE)
     if(NOT _lib_target_type STREQUAL "INTERFACE_LIBRARY")
       get_target_property(lib_link_libraries ${lib} LINK_LIBRARIES)
+      get_target_property(lib_interface_link_libraries ${lib} INTERFACE_LINK_LIBRARIES)
+      if(lib_interface_link_libraries)
+	 list(APPEND lib_link_libraries lib_interface_link_libraries)
+	 list(REMOVE_DUPLICATES lib_link_libraries)
```
sethrj: Formatting I think

sethrj (commented):
Can we also use this opportunity to add versioning to the utility cmake:
```cmake
set(_CUDA_RDC_VERSION 1)
if(CUDA_RDC_VERSION GREATER _CUDA_RDC_VERSION)
  # A newer version has already been loaded
  return()
endif()
set(CUDA_RDC_VERSION ${_CUDA_RDC_VERSION})
message(VERBOSE "Using CUDA_RDC_VERSION ${CUDA_RDC_VERSION}")
```

and in Celeritas, use this instead of the current `include(rdcutils)`:
```c++
if(VecGeom_CUDA_FOUND AND VecGeom_CUDA_VERSION VERSION_LESS 2)
  # Reload the Celeritas RDC
  include("${CMAKE_CURRENT_LIST_DIR}/cmake/CudaRdcUtils.cmake)
else()
  # Use whatever RDC utils is newest
endif()
```

pcanal:
> Can we also use this opportunity to add versioning to the utility cmake:

Sure.

> and in Celeritas, use this instead of the current include(rdcutils):

That won't help the example (using Celeritas as an installed package) though ...

sethrj:
> That won't help the example (using Celeritas as an installed package) though ...

It won't fix old installations, but it'll fix new builds of Celeritas (without having to install new vecgeom, g4vg, etc.). We're the tip of the dependency tree for this.

pcanal:
So I would need to put something like:
```
if(VecGeom_CUDA_FOUND AND VecGeom_CUDA_VERSION VERSION_LESS 2)
  # Reload the Celeritas RDC
  include("${CELERITAS_CMAKE_DIR}/cmake/CudaRdcUtils.cmake)
else()
    # Use whatever RDC utils is newest
  include(CudaRdcUtils.cmake)
endif()
```
in `celeritas/example/offload-template/CMakeLists.txt` ...

Actually with what I coded I think I can just do:
```
  include(CudaRdcUtils.cmake)
  include(${CELERITAS_CMAKE_DIR}/CudaRdcUtils.cmake)
```

cmake/CudaRdcUtils.cmake
```diff
@@ -122,7 +122,19 @@ relocatable device code and most importantly linking against those libraries.
...
+set(_CUDA_RDC_VERSION 1)
+if(CUDA_RDC_VERSION GREATER _CUDA_RDC_VERSION)
+  # A newer version has already been loaded
+  return()
+endif()
+if(NOT CUDA_RDC_VERSION NOT COMMAND cuda_rdc_add_library)
+  # if this is the first we get here use include_guard
+  # to prevent the inclusion of non-version version
+  # of this file
+  include_guard(GLOBAL)
```
sethrj: I think we want to skip the include guard since we might actually want to re-process the file after someone else rededfined cuda_rdc_add_library...

pcanal: This is the case (line 130) where we have neither seen a new version (`NOT CUDA_RDC_VERSION`) nor an ancient version (`cuda_rdc_add_library`) and thus we are the very first to set the guard.
The guard will prevent the inclusion of the ancient version (because they have the guard unconditionally)
The guard have no effect in any future parsing of new version (because they have `CUDA_RDC_VERSION` set and thus would call `include_guard`)

sethrj: But I don't think it will preclude the ancient version: surely it's using an absolute path rather than the base name ?

pcanal: Unfortunately you are right.  In completely related note, since we have to publish/install Celeritas's CudaRdcUtils, it will hide's VecGeom for downstream package whether it is newer or not.  So I need to find a way to make the install conditional again.

sethrj: We can just make sure it's newer right? If people have trouble linking we tell them to update to the latest `.x` revision.

pcanal: > We can just make sure it's newer right? 

I think I did find a decent way.

> If people have trouble linking we tell them to update to the latest .x revision.

How/where?

For now, if the CudaRdcUtils is found (by Celeritas's CMake) to be newer, it will be installed and if (as one would expect) the Celeritas installation path is before VecGeom's install path in the `CMAKE_PREFIX_PATH` it should work out nicely.  (If not, user would indeed have to locally download a newer version of CudaRdcUtils.cmake and put it somewhere that hides the others).

In a next PR, I am thinking of making CudaRdcUtils an external builtin for Celeritas (and VecGeom and g4gv) so that I don't have to change it 4 times ....

sethrj (approved):
Thanks @pcanal !

CMakeLists.txt
```diff
@@ -501,13 +504,33 @@ endif()
...
+# Load RDC utils using CMAKE_MODULE_PATH
+include(CudaRdcUtils)
+
+# Load our local copy if it's newer
+if(NOT CUDA_RDC_VERSION)
+   set(_loaded_CUDA_RDC_VERSION 0)
+else()
+   set(_loaded_CUDA_RDC_VERSION ${CUDA_RDC_VERSION})
+endif()
+include("${_LOCAL_RDCUTILS_FILENAME}")
```
pcanal, line 523: Isn't that always empty/not-set ... at least at the first run?

CMakeLists.txt
```diff
@@ -13,7 +13,10 @@ set(CMAKE_USER_MAKE_RULES_OVERRIDE
 
 project(Celeritas VERSION "${Celeritas_VERSION}" LANGUAGES CXX)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH
+  "${CMAKE_CURRENT_LIST_DIR}/cmake"
+  "${CMAKE_CURRENT_LIST_DIR}/cmake/external"
```
pcanal: I don't think we should add this subdirectory to the the CMAKE_MODULE_PATH, I really meant for the file to only be loaded explicitly.  If it is loaded through the CMAKE_MODULE_PATH, a newer version from VecGeom would never be loaded/seen.

CMakeLists.txt
```diff
@@ -501,13 +504,33 @@ endif()
...
+   set(_loaded_CUDA_RDC_VERSION ${CUDA_RDC_VERSION})
+endif()
+include("${_LOCAL_RDCUTILS_FILENAME}")
+if(CUDA_RDC_VERSION GREATER _loaded_CUDA_RDC_VERSION)
+  # Install our copy since it's newer than what's available elsewhere
+  set(_LOCAL_RDCUTILS_FILENAME "${PROJECT_SOURCE_DIR}/cmake/external/CudaRdcUtils.cmake")
+else()
+  # Don't install
+  set(_LOCAL_RDCUTILS_FILENAME)
+endif()
```
```suggestion
set(_LOCAL_RDCUTILS_FILENAME "${PROJECT_SOURCE_DIR}/cmake/external/CudaRdcUtils.cmake")
include("${_LOCAL_RDCUTILS_FILENAME}")
if(NOT CUDA_RDC_VERSION GREATER _loaded_CUDA_RDC_VERSION)
  # Only install our copy if it's newer than what's available elsewhere
  set(_LOCAL_RDCUTILS_FILENAME)
endif()
```

CMakeLists.txt
```diff
@@ -13,7 +13,10 @@ set(CMAKE_USER_MAKE_RULES_OVERRIDE
 
 project(Celeritas VERSION "${Celeritas_VERSION}" LANGUAGES CXX)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH
+  "${CMAKE_CURRENT_LIST_DIR}/cmake"
+  "${CMAKE_CURRENT_LIST_DIR}/cmake/external"
+)
```
```suggestion
list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
```

sethrj: This fails at the first `include(CudaRdcUtils)` call if you're not building against vecgeom or don't have CUDA enabled:
https://github.com/celeritas-project/celeritas/actions/runs/13856531021/job/38774524699

pcanal: Ok.  So neither option works.  Let me see what I can do.

pcanal: With the last commit, it should work.

cmake/CudaRdcUtils.cmake
sethrj: > How/where?

If they post an issue or ask us on slack 😅 thankfully there aren't many users yet.

> For now, if the CudaRdcUtils is found (by Celeritas's CMake) to be newer, it will be installed and if (as one would expect) the Celeritas installation path is before VecGeom's install path in the `CMAKE_PREFIX_PATH` it should work out nicely.

I think it's dependent not on the `CMAKE_PREFIX_PATH` order but on the dependency order of the `find_package` and `find_dependency` and the fact that `CMAKE_MODULE_PATH` is appended to in the config files. But I believe you're right/

> In a next PR, I am thinking of making CudaRdcUtils an external builtin for Celeritas (and VecGeom and g4gv) so that I don't have to change it 4 times ....

You would still have to submit PRs to update the git hash/version number, but please don't do this because we want to preserve the capability to build *without* having to touch the internet, which can be important for DOE codes.

CMakeLists.txt
sethrj: Assuming no one used that variable name, yeah, but I thought you've knocked me before on not explicitly clearing variables 😉

CMakeLists.txt
sethrj: We have to either do this or escape out the CudaRdcUtils include when vecgeom is disabled.

pcanal: The former is 'deadly' (i.e. it is equivalent to only ever use the version in Celeritas no matter what).
The later is trivial (add `OPTIONAL` to the `include` call).

pcanal: But `OPTIONAL` does not seem to work as [advertised](https://cmake.org/cmake/help/latest/command/include.html#include) out of the box.  I am investigating.

pcanal: Oh.  That's because there is more than one such include :(

CMakeLists.txt
```diff
@@ -501,13 +501,31 @@ endif()
...
+#----------------------------------------------------------------------------#
+
+# Load RDC utils using CMAKE_MODULE_PATH
+include(CudaRdcUtils OPTIONAL)
+
+# Load our local copy if it's newer
+if(NOT CUDA_RDC_VERSION)
+   set(_loaded_CUDA_RDC_VERSION 0)
+else()
+   set(_loaded_CUDA_RDC_VERSION ${CUDA_RDC_VERSION})
```
sethrj, line 520: Wait since now we remembered that `LibraryUtils` loads our own copy of CudaRdcUtils first, won't this always be set and "up to date" (even if a previous older version was later loaded and overrode our functions with old versions?) ?

pcanal: Humm .. no?

> Wait since now we remembered that LibraryUtils loads our own copy of CudaRdcUtils first,

It does not (in the current state).  It loads one on the CMAKE_MODULE_PATH.

sethrj: Ah right I see, because `include(CeleritasLibraryUtils)` comes after the `include(CudaRdcUtils)` stuff. I'm going to update the library utils to just error if CUDA_RDC isn't included, rather than potentially loading the wrong version if someone moves it up to the top of the main cmakelists.

sethrj (approved):

cmake/CudaRdcUtils.cmake
pcanal: > You would still have to submit PRs to update the git hash/version number, 

yes, but more 'trivial' maybe?

> but please don't do this because we want to preserve the capability to build without having to touch the internet

Ok, I will hold off for now ... but we could have the CudaRdcUtils downloaded and installed on it own (yes a single CMake file + tests :) ) for the no-internet case  ....

PR #1666, opened 2025-03-13
Lazily fetch volumes for UniformAlongStepFactory 

rahmans1:
Enable UniformAlongStepFactory to take a function instead of a vector of volumes. This is because the required logical volumes are not initialized at the level where the factory is initialized. This builds on https://github.com/celeritas-project/celeritas/pull/1659.

### DIFF
--- a/src/accel/SetupOptions.cc
+++ b/src/accel/SetupOptions.cc
@@ -117,10 +117,12 @@ void ProblemSetup::operator()(inp::Problem& p) const
         // Check if magnitude is zero
         auto field = u->get_field();
         auto field_val = norm(field.strength);
-        if (field_val > 0)
+	auto volumes = u->get_volumes();
+        if (field_val > 0 && !volumes.empty())
         {
             CELER_LOG(info) << "Using a uniform field: " << field_val << " [T]";
             p.field = std::move(field);
+	    p.volumes = std::move(volumes);
         }
         else
         {
--- a/src/accel/SetupOptions.cc
+++ b/src/accel/SetupOptions.cc
@@ -119,7 +119,17 @@ void ProblemSetup::operator()(inp::Problem& p) const
         auto field_val = norm(field.strength);
         if (field_val > 0)
         {
-            CELER_LOG(info) << "Using a uniform field: " << field_val << " [T]";
+            auto msg = CELER_LOG(info);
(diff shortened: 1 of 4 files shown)


src/accel/SetupOptions.cc
```diff
@@ -118,10 +118,12 @@ void ProblemSetup::operator()(inp::Problem& p) const
         // Check if magnitude is zero
         auto field = u->get_field();
         auto field_val = norm(field.strength);
-        if (field_val > 0)
+	auto volumes = u->get_volumes();
+        if (field_val > 0 && !volumes.empty())
         {
             CELER_LOG(info) << "Using a uniform field: " << field_val << " [T]";
             p.field = std::move(field);
+	    p.volumes = std::move(volumes);
```
sethrj: It's OK to have a field without any volumes (an empty vector or no volumes means "all volumes").
```suggestion
        if (field_val > 0)
        {
            CELER_LOG(info) << "Using a uniform field: " << field_val << " [T]";
            p.field = std::move(field);
	    p.volumes = u->get_volumes();
```

sethrj (commented):
Hey I realize I gave you some bad advice based on my not remembering the new field changes, and by forgetting that #1666 isn't yet merged 🤦  I think if you just make the changes to AlongStepFactory and don't touch the Problem or SetupOptions, everything should work.

P.S. Great first pull request!! 🎉

src/celeritas/inp/Problem.hh
```diff
@@ -52,6 +52,8 @@ struct Problem
     Physics physics;
     //! Set up the magnetic field
     Field field;
+    //! Set up the volumes where field will be present
+    std::vector<G4LogicalVolume const*> volumes;
```
amandalund:
```suggestion
```
I don't think this is needed, is it? We're planning on updating the volumes in the `Field` input itself to use a variant set instead of a `VolumeId` vector (https://github.com/celeritas-project/celeritas/pull/1659#discussion_r1987915912), but I think for now your changes to the along-step factory are enough to get things working!

sethrj: That was fallout from my incorrect suggestion 😅

sethrj (approved):
Yay!

PR #777, opened 2023-05-26
Support running kernels on different CUDA streams

amandalund:
There's a slight speedup compared to launching everything on the default stream, which varies depending on the problem and number/size of the events, but for the demo loop running the cms2018+field+msc-vecgeom-gpu regression problem is ~8%..

PR #1667, opened 2025-03-13
Remove 'default_stream' option

sethrj:
In #777 the "default stream" was useful for performance comparisons of streaming. As part of a refactor to hide Thrust from the host compiler (necessary for new AMD builds, as @wdconinc discovered 😕 ) it will no longer be valid to ask for `StreamId{0}` if `create_streams` hasn't been called.

### DIFF
--- a/app/celer-sim/celer-sim.cc
+++ b/app/celer-sim/celer-sim.cc
@@ -106,7 +106,7 @@ void run(std::shared_ptr<OutputRegistry>& output, std::string const& filename)
     output->insert(std::make_shared<OutputInterfaceAdapter<RunnerInput>>(
         OutputInterface::Category::input, "*", run_input));
 
-    // Allocate device streams, or use the default stream if there is only one.
+    // Allocate device streams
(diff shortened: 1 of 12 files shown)


app/celer-sim/celer-sim.cc
```diff
@@ -108,7 +108,7 @@ void run(std::shared_ptr<OutputRegistry>& output, std::string const& filename)
 
     // Allocate device streams, or use the default stream if there is only one.
     size_type num_streams = run_stream.num_streams();
-    if (run_input->use_device && !run_input->default_stream && num_streams > 1)
+    if (run_input->use_device)
```
amandalund: So when we transport all events on a single stream, we now create a new stream rather than use the default (comment should be updated)?

amandalund (approved):

app/celer-sim/celer-sim.cc
sethrj: Correct. Since I think everything uses the async/stream-based operations now it doesn't make a difference.

PR #1669, opened 2025-03-14
Default downstream CUDA/HIP architectures

sethrj:
If a user builds an app downstream from Celeritas without specifying CUDA/HIP "components", they may not have set the cmake/hip architectures necessary for compatibility.  This PR uses the CeleritasConfig file to set default values before the CUDA/HIP languages are activated. If the user has already declared CUDA/HIP as a language, the `CMAKE_{lang}_ARCHITECTURES` will have already been set by CMake and the Celeritas ones will not be loaded.

When loaded externally, we also now print statements about what variables are being set. Cache variables are printed only on the first configure with a verbosity of "STATUS", and local variables are printed with VERBOSE, matching our other CMake macros.

Also the CI wasn't building the "offload" example for docker because of how the Geant4 version was defined in the CI. This enables `accel` example testing for CUDA/HIP.

These changes are pulled from #1664.


.github/workflows/build-docker.yml
```diff
@@ -96,6 +96,12 @@ jobs:
           }}
         run: |
           . /etc/profile
+          # Get Geant4 version number for use by script
```
pcanal: Other options includes:
```
export G4VERSION_NUMBER=`/opt/view/bin/geant4-config --version | awk -F. '{print $1*100 + $2*10}')"`
```
and (at least in the ATLAS environment):
```
export G4VERSION_NUMBER=`echo $G4VERS  | awk -F. '{print $2*100 + $3*10}')"`
```
albeit the second is probably 'too much' information:
```
$ geant4-config --version
11.1.3
$ echo $G4VERS
geant4.11.1.patch03.atlasmt02
```

sethrj: Neat! didn't know that trick.

sethrj:
Currently fails until #1665 is merged.

scripts/ci/test-examples.sh
```diff
@@ -41,17 +43,25 @@ build_local
 
 # Run Geant4 app examples
 if [ -z "${CELER_DISABLE_ACCEL_EXAMPLES}" ]; then
+  if [ -z "${G4VERSION_NUMBER}" ]; then
+    # Get the geant4 version 11.2.3, replace `.` with ` `, concat as 110203
+    _vers=$(geant4-config --version | tr '.' ' ')
+    G4VERSION_NUMBER=$(printf '%d%02d%02d' ${_vers})
```
pcanal: I don't know that it matters but the `G4Version.h` claims that the 2nd and 3rd numbers are restricted to a single digit and write the version down as `1123`. I.e.:
```suggestion
    # Get the geant4 version 11.2.3, replace `.` with ` `, concat as 110203
    _vers=$(geant4-config --version | tr '.' ' ')
    G4VERSION_NUMBER=$(printf '%d%01d%01d' ${_vers})
```

scripts/ci/test-examples.sh
```diff
@@ -41,17 +43,25 @@ build_local
...
+  fi
+
   # Run small accel examples
   cd "${CELER_SOURCE_DIR}/example/accel"
   build_local
   ctest -V --no-tests=error
 
-  if [ "$(echo $CELER_GEANT4_VERSION | awk -F. '{print $1*100 + $2}')" \
-      -ge 1100 ]; then
+  if [ "${G4VERSION_NUMBER}" -ge 1100 ]; then
```
pcanal, line 57: This fails with the current code at https://github.com/celeritas-project/celeritas/pull/1669/files#r1995958825
but would work with the proposed updated.  Alternatively:
```suggestion
  if [ "${G4VERSION_NUMBER}" -ge 110000 ]; then
```

pcanal (approved):
LGTM (slight doc update needed)

scripts/ci/test-examples.sh
```diff
@@ -41,17 +43,25 @@ build_local
 
 # Run Geant4 app examples
 if [ -z "${CELER_DISABLE_ACCEL_EXAMPLES}" ]; then
+  if [ -z "${G4VERSION_NUMBER}" ]; then
+    # Get the geant4 version 11.2.3, replace `.` with ` `, concat as 110203
```
pcanal:
```suggestion
    # Get the geant4 version 11.2.3, replace `.` with ` `, concat as 1123
```

PR #1670, opened 2025-03-14
Hide Thrust from .cc code

sethrj:
The HipThrust implementation does not like being included from host compilers (i.e., clang without the HIP extension flags enabled): our CelerDeviceApi has been including thrust's memory resource allocator since #908 , and that has occasionally caused builds to fail on Frontier and ROCm CI with errors coming from thrust via `.cc` files about missing types `_bfloat16`, probably because we had to hackily define internal macros such as `THRUST_DEVICE_SYSTEM` to try to match what Thrust does when compiled by HIP.

This refactor, following on to #1667, allows the `Stream` class to be visible by host-only code (without including CUDA headers). The only weird thing is that since we can't forward-declare the CUDA/HIP typedef, a macro `CELER_DEVICE_RUNTIME_INCLUDED` is used to detect whether the headers are available and if not we use a typedef `using MissingDeviceRuntime = void` that helps explain to the user why the interface is misbehaving. The stream class also now exposes a `sync` function to do stream synchronization from non-device code, and the `async()` driver capability query has moved to Device.

All operations that take a valid `StreamId` require a matching valid device stream to be allocated: `StreamId{0}` no longer provides access to the default stream.

I also have updated some of the macros and their locations for clarity:
- `CELER_DEVICE_PREFIX` is renamed `CELER_DEVICE_API_SYMBOL` and moved to `DeviceRuntimeApi.hh`
- `CELER_DEVICE_CALL_PREFIX` is now `CELER_DEVICE_API_CALL`
- `CELER_DEVICE_CHECK_ERROR` is removed in favor of `CELER_DEVICE_API_CALL(PeekAtLastError())`


src/corecel/sys/Stream.cu
```diff
@@ -0,0 +1,119 @@
...
+ * Construct by creating a stream.
+ */
+Stream::Stream()
+{
+    StreamT stream;
+    CELER_DEVICE_API_CALL(StreamCreate(&stream));
+#if CUDART_VERSION >= 12000
+    unsigned long long stream_id = -1;
+    CELER_DEVICE_API_CALL(StreamGetId(stream, &stream_id));
+    CELER_LOG_LOCAL(debug) << "Created stream ID " << stream_id;
```
amandalund: Could we add this to the destructor as well so the stream identifiers are symmetric?

test/celeritas/user/SimpleLoopTestBase.cc
```diff
@@ -37,6 +37,11 @@ void SimpleLoopTestBase::run_impl(size_type num_tracks, size_type num_steps)
     step_inp.stream_id = StreamId{0};
     step_inp.num_track_slots = num_tracks;
 
+    if constexpr (M == MemSpace::host)
```
amandalund:
```suggestion
    if constexpr (M == MemSpace::device)
```

src/corecel/sys/Stream.cu
```diff
@@ -0,0 +1,119 @@
...
+    ResourceT memory_resource;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Destroy the stream.
+ */
+void Stream::ImplDeleter::operator()(Impl* impl) noexcept
+{
+    try
```
amandalund, line 61: Calling `cudaStreamDestroy` on a `nullptr` will  give an invalid resource handle error: since we removed the default stream, is it guaranteed now that the stream will never be null?

src/corecel/sys/Stream.hh
```diff
@@ -6,113 +6,89 @@
...
+    using ResourceT = detail::AsyncMemoryResource<void*>;
     //!@}
 
   public:
-    // Whether asynchronous operations are supported
-    static bool async();
-
     // Construct by creating a stream
     Stream();
+    CELER_DEFAULT_MOVE_DELETE_COPY(Stream);
```
amandalund, line 74: Just to double check: the move assign/construct were implemented using the copy-and-swap idiom instead of the defaults to ensure a stream is always properly destroyed, but the defaults should work now that we're storing the stream as a `unique_ptr<Impl, ImplDeleter>`, correct?

amandalund (commented):
Still working through this, just a couple initial questions/comments:

src/corecel/sys/Stream.hh
sethrj: Yep, exactly.

src/corecel/sys/Stream.cu
sethrj: Ah that's good to know, I'd assumed it was like calling `free` on `nullptr`. But yes, now that we don't ever wrap the default stream, the `impl` should always be there.

amandalund (approved):
Looks good to me!

PR #1673, opened 2025-03-15
Define dependency helper targets and reorganize CMakeLists

sethrj:
This adds libraries `Celeritas::ExtMPI`, `Celeritas::ExtOpenMP`, etc. that are always defined but transitively link to dependencies such as MPI and OpenMP when the package is enabled. It allows easier use of optional code and lets us define object libraries to further restrict the propagation of flags and includes from upstream packages for faster builds and better ccache use. 

- sincospi flag setup moved to where we determine system dependencies
- flag setup moved to a new interface library `BuildFlags` that lives in `src/`
- new external interface libraries in `src/`
- DeviceToolkit definition moved from the main cmakelists to `src/`, and `CMAKE_DL_LIBS` link moved there as well
- Perfetto alias is now `Celeritas::ExtPerfetto`


sethrj:
@pcanal I think our CUDA dev system is being slow, but on wildstyle the develop build (CUDA and vecgeom and unit tests) takes 15s to configure (3.8M lines of cmake trace output), and with this branch it takes 30s (7.5M lines of cmake trace output) even though there should be actually fewer libraries linked than before. Maybe after you review this and the MPI PR you could look and see if there's low-hanging fruit for optimization?

sethrj:
Also this is killing me... https://github.com/celeritas-project/celeritas/actions/runs/13899686495/job/38888586851

CMakeLists.txt
```diff
@@ -497,6 +497,25 @@ if(CELERITAS_BUILD_TESTS)
   celeritas_find_or_builtin_package(GTest 1.10)
 endif()
 
+#----------------------------------------------------------------------------#
+# System property: availability of sincospi
+if(NOT DEFINED CELERITAS_SINCOSPI_PREFIX)
```
pcanal, line 505: For clarify I would have set the full name of the function instead of just the prefix.  (Albeit technically the existing code could be abstract out to test/check another symbol with the same pattern).  Either way minor issue.

sethrj: It's not just one function though, it's several... #1364

cmake/CeleritasLibraryUtils.cmake
```diff
@@ -112,20 +122,28 @@ forward directly to CMake or alternatively to CudaRdc.
...
 
-  The ``<input>`` must be a relative path to the current source directory, and
-  the ``<output>` path is configured to the project build "include" directory.
+  Get a list of all source files in the argument that are CUDA language.
 
-.. command:: celeritas_polysource_append
+    celeritas_get_cuda_source_args(<var> [<source> ...])
 
-  Add C++ and CUDA/HIP source files based on the enabled options.
+.. command:: celeritas_add_library
```
pcanal, line 140: odd placement.  Wouldn't be better next to `celeritas_add_interface_library`?

sethrj: I thought this function was only used internal to this file but looks like it's used elsewhere. I can move it around.

example/accel/CMakeLists.txt
```diff
@@ -17,7 +17,9 @@ if(NOT CELERITAS_USE_Geant4)
     "to be enabled in Celeritas")
 endif()
 
-include(CudaRdcUtils)
+# Use Celeritas CUDA RDC macros to correctly link libraries using VecGeom
+# NOTE: until vecgeom/g4vg get updated, we must use the Celeritas load path
+include("${Celeritas_DIR}/CudaRdcUtils.cmake")
```
pcanal: humm .. that should not be need :(

pcanal: i.e. With the PR we recently merged, it should have installed the Celeritas version and it should have been preferred ...

CMakeLists.txt
```diff
@@ -497,6 +497,25 @@ if(CELERITAS_BUILD_TESTS)
   celeritas_find_or_builtin_package(GTest 1.10)
 endif()
 
+#----------------------------------------------------------------------------#
+# System property: availability of sincospi
```
```suggestion
# System property: availability of sincospi
# The same prefix used for sincospi applies to the whole family of function (sinpi, cospi, sincos).
```

example/accel/CMakeLists.txt
sethrj: Yeah I was flailing ... I thought that maybe if Celeritas loads RdcUtils, then an old VecGeom loads RdcUtils, then we tried to load Celeritas RdcUtils again, it wouldn't work?

Also ... I just realized ... neither of the config files actually include RdcUtils, they just modify the paths. So when we do `include(CudaRdcUtils)` it's based on what config file last did `list(APPEND CMAKE_MODULE_PATH`... which will be Celeritas if you explicitly load vecgeom first, or VecGeom if you let Celeritas loaded it. The versioning guard won't affect that :(

pcanal: > VecGeom if you let Celeritas loaded it.

We could/should fix that.

sethrj: We'd have to unconditionally include CudaRdcUtils from both Config files, right?

pcanal: Yes, I think so.

sethrj: I guess we should've been doing this all along 🤔

CMakeLists.txt
```diff
@@ -512,88 +532,18 @@ endif()
...
 
-# Load RDC utils using CMAKE_MODULE_PATH
-include(CudaRdcUtils OPTIONAL)
-
-# Load our local copy if it's newer
-if(NOT CUDA_RDC_VERSION)
-   set(_loaded_CUDA_RDC_VERSION 0)
-else()
-   set(_loaded_CUDA_RDC_VERSION ${CUDA_RDC_VERSION})
-endif()
```
pcanal, line 523: This was made 'unnecessary' because the optionality of the install of the `CudaRdcUtils` was inadvertently (!?) removed in https://github.com/celeritas-project/celeritas/pull/1674 and in particular [`f6fc29f` (#1674)](https://github.com/celeritas-project/celeritas/pull/1674/commits/f6fc29fa99c2d6bbe465d9ef5f2050f4a8aacc9c).  

That commit seems to address the issue "Always install RdcUtils since the build fails without it " but it is not clear that this was the best solution ... and the very least the commit log could use an extension to explain why installing  an older/obsolete version is useful.

sethrj: It was an expedient to fix the build. I'll update the PR notes. I'd like to get this in to enable `celeritas_add_library` downstream in our examples and tag an `rc`; then could you create a new PR that finesses the installation and reverts this chunk and https://github.com/celeritas-project/celeritas/pull/1674/commits/f6fc29fa99c2d6bbe465d9ef5f2050f4a8aacc9c if you can get everything working without it?

example/accel/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if(NOT CELERITAS_USE_Geant4)
 endif()
 
 add_executable(simple-offload simple-offload.cc)
-cuda_rdc_target_link_libraries(simple-offload
+celeritas_target_link_libraries(simple-offload
```
pcanal: This is a notable strategic change that deserve a mention/justification in the PR synopsis (and maybe the git log).

CMakeLists.txt
pcanal: Since we agreed to always distribute the local version, there is no longer a need to revert.  But whether we use below:
```
include(CudaRdcUtils OPTIONAL)
include("${_LOCAL_RDCUTILS_FILENAME}")
```
or just
```
include("${_LOCAL_RDCUTILS_FILENAME}")
```
depends on whether we choose to trust the `find_package` to already include their own version (i.e. in case theirs is newer).

sethrj: We haven't release any versions of VecGeom nor G4VG yet that include RdcUtils in their Config file. Future VG and G4VG should distribute and include it (and their inclusion will be new enough not to override extant newer versions). Therefore in the future a `find_package` alone will be enough to load the newest RdcUtils among all the dependencies. Present `find_package` is unnecessary because we know no present version is newer than ours.

Does that sound right?

example/accel/CMakeLists.txt
sethrj: It's something of a spinoff from the "documentation needs updating" 😅 You're right that it might be better to move to a separate quick follow-on PR.

sethrj: OK moved to #1678

CMakeLists.txt
pcanal: That sounds about right.  (On a side note, this PR is changing `CudaRdcUtils.cmake` (minorly so it does not actually matter) but does not increase the version number ... at the same time `VecGeom` has a PR that is also changing `CudaRdcUtils.cmake` (with a slight behavior change) and should update the version number too) :( )

.github/workflows/build-docker.yml
```diff
@@ -22,9 +22,9 @@ jobs:
       fail-fast: false
       matrix:
         special: [null]
-        geometry: ["orange", "vecgeom"]
+        geometry: ["vecgeom"]
```
pcanal: Reminder: revert.

pcanal (approved):
LGTM.

.github/workflows/build-docker.yml
sethrj: Yep when I do the "revertme " I delete the "all" job so it's impossible to merge till we restore it :)

PR #1674, opened 2025-03-16
Support VecGeom 2.0

sethrj:
Recent changes to the surface model branch affect the volume models too, including the use of single-precision BVH. This fixes the build with the latest version, improves diagnostic output, adds the "max safety" performance enhancement.


src/geocel/vg/VecgeomParams.cc
```diff
@@ -531,11 +532,7 @@ void VecgeomParams::build_volume_tracking()
 
     {
         ScopedTimeAndRedirect time_and_output_("vecgeom::ABBoxManager");
-#if VECGEOM_VERSION < 0x020000
-        vecgeom::ABBoxManager::Instance().InitABBoxesForCompleteGeometry();
-#else
-        vecgeom::ABBoxManager_t::Instance().InitABBoxesForCompleteGeometry();
-#endif
+        detail::ABBoxManager_t::Instance().InitABBoxesForCompleteGeometry();
     }
```
mrguilima, line 536: Are we still keeping the compatibility with VecGeom 1.2.10 after this PR?

mrguilima (approved):
Sorry that I could not look into this before.
It looks great!
Thanks, @sethrj

src/geocel/vg/VecgeomParams.cc
sethrj: Yep!

sethrj:
@mrguilima The update from surfacedev.1 to surfacedev.2:
```
d175db496 Fixing the check for reference in TestBVHNavigator.
774fbde8b fix debug compilation
83e2236dd [surface model] Mixed precision improvements
dc21658bb Enable single precision in solid BVH
a55c8a79e This fix correctly compensates internal pushes in Boolean subtractions.
695925a66 [surface model] Consolidate cmake options related to surface model, and to navigation state type.
```
has caused several of the tests to start failing, but I think it's important to get the capability in so I'm going to merge. Could you please as a follow-on MR fix the tests and then enable testing on the CI with:
```diff
--- a/.github/workflows/build-spack.yml
+++ b/.github/workflows/build-spack.yml
@@ -222,8 +222,6 @@ jobs:
         continue-on-error: >- # TODO: fix failing tests
           ${{
             fromJSON(matrix.geant || '0') < 11
-            || matrix.geometry == 'vgsurf'
-            || matrix.geometry == 'vecgeom2'
           }}
         run: |
           if ${{matrix.geant == '11.0'
@@ -249,8 +247,6 @@ jobs:
                 || steps.unittest.outcome == 'failure')
             && (!matrix.geant
                 || fromJSON(matrix.geant) >= 11)
-            && !(matrix.geometry == 'vgsurf'
-                 || matrix.geometry == 'vecgeom2')
           }}
         with:
           name: test-results-spack-${{env.CMAKE_PRESET}}-${{matrix.geant}}
```
Thanks!

PR #1675, opened 2025-03-17
Add quantity support to FindInterp and template Turn

sethrj:
To support #1662 this adds support for Quantity division (unit / unit = unitless) . It also templates the `Turn` function so that we can use single-precision turns. A new function `acos2turn` returns `acos2` converted to Turn units.

### DIFF
--- a/test/corecel/grid/FindInterp.test.cc
+++ b/test/corecel/grid/FindInterp.test.cc
@@ -1,0 +1,47 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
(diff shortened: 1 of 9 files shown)


test/corecel/grid/FindInterp.test.cc
```diff
@@ -0,0 +1,144 @@
...
+    {
+        EXPECT_THROW(find_interp(grid, 0.999), DebugError);
+        EXPECT_THROW(find_interp(grid, 5.0), DebugError);
+        EXPECT_THROW(find_interp(grid, 5.001), DebugError);
+    }
+}
+
+// In this case, the fraction is always truncated by integer division to zero
+// If we actually care about this in the future we can return a rational number
+// for the "value"
```
amandalund:
```suggestion
```

amandalund (approved):
Thanks!

esseivaju (approved):

PR #1676, opened 2025-03-17
By default load CudaRdcUtils and print config message

sethrj:
This quality-of-life cmake improvement means users no longer have to provide a `FindCeleritas.cmake` that prints the version, nor do they need to include `CudaRdcUtils` if they hope to use Celeritas with VecGeom+CUDA.


pcanal (approved):
Thanks.

PR #1677, opened 2025-03-17
Fix bremsstrahlung process construction when using only a single model

amandalund:
This fixes the error @whokion reported seeing when trying to use only one of the bremsstrahlung models in a Geant4-integrated application, e.g. with only SB:
```console
/home/alund/celeritas_project/celeritas/app/celer-g4/celer-g4.cc:277: critical: While running testem3-simple-calo-celer-g4.inp.json: Geant4 error: Celeritas runtime error: missing imported model e_brems_lpm
celeritas/phys/ImportedModelAdapter.cc:128: 'celer.init.global' failed
{"result":{"exception":{"condition":"celer.init.global","file":"celeritas/phys/ImportedModelAdapter.cc:128","type":"RuntimeError","what":"Celeritas runtime error: missing imported model e_brems_lpm","which":"Geant4"}}}
```
Since the `GeantPhysicsOptions` can only be specified when we used one of our own physics lists, we don't have an option for brems model selection in accel, and we use the default `ProcessBuilder` options (with `BremsModelSelection::all`). As a result, when we do use our own physics list, the `BremsstrahlungProcess` might try to construct the wrong models. This is fixed by removing the model selection option from the process and process builder and using the imported data to determine which models to construct.

There is a separate issue which I can address as a follow-on: with only relativistic brems enabled, we hit a debug assertion during the discrete model selection because we sample the bremsstrahlung process below the relativistic model limit.


sethrj (approved):
Much more sensible. Thanks!

PR #1680, opened 2025-03-19
Fix static builds and CI reporting

sethrj:
Because static builds require propagating all their dependencies, we have to make sure that all our built-in libraries are installed.

This adds a static build to the CI and rearrange some of the GitHub action workflow code, as well as fixing some of the test reporting. It also slightly shifts the location of the vecgeom options in the output, enables more CI jobs to give useful output on failure, and adds a script for profiling cmake.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -63,19 +63,44 @@ jobs:
         with:
           fetch-depth: 383
           fetch-tags: true
+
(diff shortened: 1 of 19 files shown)


sethrj:
@esseivaju finally 🤪😭

esseivaju (approved):
Looks good on my side, thanks for that fix 😄

PR #1681, opened 2025-03-19
Fix BoundingZone floating point issue

elliottbiondo:
While trying to ray trace the XLDZ geometry I encountered an issue during the GEANT4 conversion process related to BoundingZones being created with interior bboxes slightly larger in one direction than exterior bboxes. I traced the issue back to `IntersectSurfaceBuilder.cc`, and it seems that this issue was already partially remedied in #1257. I simply extended the approach use there for `shrink_exterior` to `grow_interior`.


elliottbiondo:
@sethrj small one,  RTR

sethrj (commented):
What shapes specifically were giving you this problem? If the exterior is the same as the interior on a face of the box, there has to be a planar face at exactly that bbox plane, which seems unlikely.

elliottbiondo:
I ran into it with a cone in which the top and bottom radii are nearly the same. In this case, both the inner and outer bboxes have z-planes equal to +/-hh.

sethrj (approved):

PR #1682, opened 2025-03-20
Enable optical GPU tests

amandalund:
This turns on the optical GPU tests which were unintentionally disabled. It also tweaks a couple of the app CLI11 commands.


sethrj (approved):
Thanks!

PR #1683, opened 2025-03-20
Drastically speed up CudaRdcUtils

pcanal:
A while back a Celeritas CMake invocation with VecGeom (and most the options turned on) was taking about 1.4 millions lines, recent fixes increased this to 11 millions and an intermediary functional fix increase it further to 186 millions !! (some potential dependency were previously skipped due to missing error handling in the code).

Adding caching of the dependency decreased the number of lines to 500k and the narrowing of the search to only RDC libraries decreased it all the down down to 250k lines (However this later change might fail for some untagged commits of VecGeom where the libvecgeom library is missing the CUDA RDC properties).


cmake/FindVecGeom.cmake
```diff
@@ -21,17 +21,19 @@ cmake_policy(POP)
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(VecGeom CONFIG_MODE)
 
-if(VecGeom_FOUND AND TARGET VecGeom::vecgeomcuda)
+if(VecGeom_FOUND AND (TARGET VecGeom::vecgeomcuda OR TARGET VecGeom::vecgeom_final))
```
sethrj, line 26: It would be useful to have version ranges for what vecgeom has the different behavior: if we stop supporting 1.x for example some of this code can go.

cmake/external/CudaRdcUtils.cmake
```diff
@@ -613,24 +642,46 @@ function(cuda_rdc_depends_on OUTVARNAME lib potentialdepend)
       get_target_property(lib_link_libraries ${lib} LINK_LIBRARIES)
       get_target_property(lib_interface_link_libraries ${lib} INTERFACE_LINK_LIBRARIES)
       if(lib_interface_link_libraries)
-        list(APPEND lib_link_libraries lib_interface_link_libraries)
-        list(REMOVE_DUPLICATES lib_link_libraries)
+        if(lib_link_libraries)
+	   list(APPEND lib_link_libraries ${lib_interface_link_libraries})
+        else()
+	   set(lib_link_libraries ${lib_interface_link_libraries})
+	endif()
```
sethrj: This looks like an optimization; does it actually help in practice?

cmake/external/CudaRdcUtils.cmake
```diff
@@ -613,24 +642,46 @@ function(cuda_rdc_depends_on OUTVARNAME lib potentialdepend)
...
+        continue()
+      endif()
+      get_target_property(_finallib ${linklib} CUDA_RDC_FINAL_LIBRARY)
+      # We only consider libraries that might contains RDC Cuda code
+      if(NOT _finallib)
+        continue()
+      endif()
       cuda_rdc_depends_on(${OUTVARNAME} ${linklib} ${potentialdepend})
       if(${OUTVARNAME})
+	if(${${OUTVARNAME}})
```
sethrj: It's not strictly an outvarname then?

sethrj (approved):
Couple of minor comments. Thanks!

cmake/FindVecGeom.cmake
pcanal: With all the streams, it might be a mess :( [ v1, v2, surface ] ) ... I will add a comment with the versions number I can find.

cmake/external/CudaRdcUtils.cmake
pcanal: Which part?  I replace:
```
        if(lib_link_libraries)
	   list(APPEND lib_link_libraries ${lib_interface_link_libraries})
```
with simply
```
	   list(APPEND lib_link_libraries ${lib_interface_link_libraries})
```
I ended up having to deal with a library named `lib_link_libraries-NOTFOUND`

cmake/external/CudaRdcUtils.cmake
pcanal: It is, it is set to `TRUE` or `FALSE`

pcanal: For the recursion, I happen to reuse the same name for all levels.

cmake/FindVecGeom.cmake
sethrj: I think only 1.x and 2.0+ (aka beyond current master) need to be worried about: it's just a comment after all.

cmake/external/CudaRdcUtils.cmake
sethrj: ahhhh that's subtle. I was thinking it was empty. Maybe a little comment would be good? (I wonder if a macro `append_if_found` would slow it down?)

cmake/external/CudaRdcUtils.cmake
sethrj: But it's really `inout` effectively. Most `outvar` functions set but don't get that variable. It would be good to document too.

cmake/external/CudaRdcUtils.cmake
pcanal: I think I will just add a comment.  The `append_if_found` used here would be more aptly named `append_if_list_sofar_is_invalid_replace_if_not`

cmake/external/CudaRdcUtils.cmake
pcanal: I still do not see the `inout` nature.  I only see a reuse of a (user given) name in a different scope.

pcanal: On the other hand, i see that
```
      if(${OUTVARNAME})
	if(${${OUTVARNAME}})
```
is actually redundant ... so the code is wrong ...

pcanal: I updated the code to remove the dead/redundant code.

sethrj: That explains it ;) the original code was "if the outvar has a value, and if the variable reference by outvar is true..." i.e. it's using the pointed-to variable.

PR #1684, opened 2025-03-20
Use float for field interpolation

esseivaju:
doubles are slow, and likely not needed here


amandalund (approved):
Thanks @esseivaju! Hope this helps some 🤞

src/accel/CylMapMagneticField.cc
```diff
@@ -55,31 +55,30 @@ inline void cartesian_to_cylindrical(Array<G4double, 3> const& cart,
  * it will retrieve the G4FieldManager's field to sample it.
  */
 CylMapFieldParams::Input
-MakeCylMapFieldInput(std::vector<real_type> const& r_grid,
-                     std::vector<real_type> const& phi_values,
-                     std::vector<real_type> const& z_grid)
+MakeCylMapFieldInput(std::vector<float> const& r_grid,
+                     std::vector<float> const& phi_values,
+                     std::vector<float> const& z_grid)
```
sethrj: The geant4 interface functions should all take `double` I think. We should convert internally to real_type.

src/celeritas/field/CylMapField.hh
```diff
@@ -52,9 +52,9 @@ class CylMapField
     // Shared constant field map
     FieldParamsRef const& params_;
 
-    NonuniformGrid<real_type> const grid_r_;
-    NonuniformGrid<real_type> const grid_phi_;
-    NonuniformGrid<real_type> const grid_z_;
+    NonuniformGrid<float> const grid_r_;
+    NonuniformGrid<float> const grid_phi_;
+    NonuniformGrid<float> const grid_z_;
```
sethrj: Instead of hardcoding `float`, add `using cylmap_real_type = float;` to the FieldData header, and locally to each class `using real_type = cylmap_real_type;`

src/celeritas/field/CylMapField.hh
```diff
@@ -87,9 +87,9 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
     Real3 value{0, 0, 0};
```
sethrj: You'll probably want  to use `Array<cyl_real_type, 3>` here and convert at the end.

src/celeritas/field/CylMapFieldInput.hh
```diff
@@ -34,12 +34,12 @@ namespace celeritas
...
-    std::vector<Turn> grid_phi;  //!< Phi grid points [AU]
-    std::vector<real_type> grid_z;  //!< Z grid points [len]
+    std::vector<float> grid_r;  //!< R grid points [len]
+    std::vector<Turn_t<float>> grid_phi;  //!< Phi grid points [AU]
+    std::vector<float> grid_z;  //!< Z grid points [len]
 
-    std::vector<real_type> field;  //!< Flattened R-Phi-Z field component
-                                   //!< [bfield]
+    std::vector<float> field;  //!< Flattened R-Phi-Z field component
+                               //!< [bfield]
```
sethrj: It's probably better to have the input be in standard `real_type` precision, so we could in theory check for catastrophic precision loss when we downconvert to double.

test/celeritas/field/Fields.test.cc
```diff
@@ -59,44 +59,44 @@ TEST(CMSParameterizedFieldTest, all)
...
+                                           3.7958506580647,
+                                           0.014834624748597,
+                                           0.014834624748597,
+                                           3.7873486828586,
+                                           0.021253065345318,
+                                           0.021253065345318,
+                                           3.7772444535824,
+                                           0.028935543902684,
+                                           0.028935543902684,
+                                           3.7656950871883};
```
sethrj: Add a trailing comma here to improve fofrmating

test/celeritas/field/Fields.test.cc
```diff
@@ -116,46 +116,46 @@ TEST_F(RZMapFieldTest, all)
...
+                                           3.7912385463715,
+                                           0.011867451667786,
+                                           0.011867451667786,
+                                           3.775991499424,
+                                           0.014240986622126,
+                                           0.014240986622126,
+                                           3.771880030632,
+                                           0.016614892251046,
+                                           0.016614892251046,
+                                           3.757196366787};
```
sethrj: Same, and in this file also use typedefs rather than `float`

src/celeritas/field/CylMapField.hh
```diff
@@ -135,8 +135,8 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
 
     // Project cylindrical components to Cartesian coordinates
     // default for r == 0
-    real_type cos_phi = 1.;
-    real_type sin_phi = 0.;
+    float cos_phi = 1.;
+    float sin_phi = 0.;
```
sethrj: avoiding the doubles (and assuming you've locally aliased `real_type = ...`
```suggestion
    real_type cos_phi = 1;
    real_type sin_phi = 0;
```

sethrj (changes requested):
Type aliases are your friend 😉

esseivaju:
Windows is very pedantic about wanting explicit casts 😅

sethrj:
It is indeed, but every so often it catches a very subtle bug...

sethrj (approved):
Great! Did you have a chance to look at the kernel stats before and after? Was there any performance difference?

esseivaju:
This is the output from compiling with double, and float. I haven't do too much performance comparison
```
// double
ptxas info    : Function properties for _ZN9celeritas6detail26PropagationApplierBaseImplINS0_28CylMapFieldPropagatorFactoryEEclERNS_13CoreTrackViewE
    936 bytes stack frame, 520 bytes spill stores, 520 bytes spill loads

// float
ptxas info    : Function properties for _ZN9celeritas6detail26PropagationApplierBaseImplINS0_28CylMapFieldPropagatorFactoryEEclERNS_13CoreTrackViewE
    824 bytes stack frame, 468 bytes spill stores, 468 bytes spill loads
```

sethrj:
Wonder what it looks like with orange

esseivaju:
For orange:
```
// double
ptxas info    : Compiling entry function '_ZN9celeritas6detail65_GLOBAL__N__d3199362_32_AlongStepCylMapFieldMscAction_cu_aa8d5c2918launch_action_implINS_24ConditionalTrackExecutorINS_22IsAlongStepActionEqualENS0_18PropagationApplierINS0_28CylMapFieldPropagatorFactoryEvEEEELb1EEEvNS_5RangeINS_8OpaqueIdINS_7Thread_EjEEEET_' for 'sm_80'
ptxas info    : Function properties for _ZN9celeritas6detail65_GLOBAL__N__d3199362_32_AlongStepCylMapFieldMscAction_cu_aa8d5c2918launch_action_implINS_24ConditionalTrackExecutorINS_22IsAlongStepActionEqualENS0_18PropagationApplierINS0_28CylMapFieldPropagatorFactoryEvEEEELb1EEEvNS_5RangeINS_8OpaqueIdINS_7Thread_EjEEEET_
    632 bytes stack frame, 44 bytes spill stores, 80 bytes spill loads
ptxas info    : Used 128 registers, used 0 barriers, 632 bytes cumulative stack size, 536 bytes cmem[0], 720 bytes cmem[2]
ptxas info    : Compile time = 11013.957 ms
ptxas info    : Function properties for _ZN9celeritas15FieldPropagatorINS_11FieldDriverINS_20DormandPrinceStepperINS_16MagFieldEquationINS_11CylMapFieldEEEEEEENS_15OrangeTrackViewEEclEd
    0 bytes stack frame, 328 bytes spill stores, 1264 bytes spill loads
ptxas info    : Function properties for _ZNK9celeritas11FieldDriverINS_20DormandPrinceStepperINS_16MagFieldEquationINS_11CylMapFieldEEEEEE16accurate_advanceEdRKNS_8OdeStateEd
    0 bytes stack frame, 190 bytes spill stores, 800 bytes spill loads


// float
ptxas info    : Compiling entry function '_ZN9celeritas6detail65_GLOBAL__N__d3199362_32_AlongStepCylMapFieldMscAction_cu_aa8d5c2918launch_action_implINS_24ConditionalTrackExecutorINS_22IsAlongStepActionEqualENS0_18PropagationApplierINS0_28CylMapFieldPropagatorFactoryEvEEEELb1EEEvNS_5RangeINS_8OpaqueIdINS_7Thread_EjEEEET_' for 'sm_80'
ptxas info    : Function properties for _ZN9celeritas6detail65_GLOBAL__N__d3199362_32_AlongStepCylMapFieldMscAction_cu_aa8d5c2918launch_action_implINS_24ConditionalTrackExecutorINS_22IsAlongStepActionEqualENS0_18PropagationApplierINS0_28CylMapFieldPropagatorFactoryEvEEEELb1EEEvNS_5RangeINS_8OpaqueIdINS_7Thread_EjEEEET_
    632 bytes stack frame, 44 bytes spill stores, 80 bytes spill loads
ptxas info    : Used 128 registers, used 0 barriers, 632 bytes cumulative stack size, 536 bytes cmem[0], 552 bytes cmem[2]
ptxas info    : Compile time = 10007.788 ms
ptxas info    : Function properties for _ZN9celeritas15FieldPropagatorINS_11FieldDriverINS_20DormandPrinceStepperINS_16MagFieldEquationINS_11CylMapFieldEEEEEEENS_15OrangeTrackViewEEclEd
    0 bytes stack frame, 290 bytes spill stores, 1180 bytes spill loads
ptxas info    : Function properties for _ZNK9celeritas11FieldDriverINS_20DormandPrinceStepperINS_16MagFieldEquationINS_11CylMapFieldEEEEEE16accurate_advanceEdRKNS_8OdeStateEd
    0 bytes stack frame, 78 bytes spill stores, 68 bytes spill loads
```

sethrj:
Thanks @esseivaju ! Was the vecgeom output truly limited to the propagation applier? It's quite impressive that the spill load count for the float case decreases by a factor of 10: since that likely means a factor of ~10 reduction in memory access, it could even mean a 10x speedup if the geometry lookups are cheap.

esseivaju:
Yeah that's a pretty big difference, it'll be interesting to compare the two. For VecGeom, it is indeed the only function that outputs any difference in spill when switching between float/double

PR #1686, opened 2025-03-20
Make SB energy limit configurable and partially fix failures when using a single brems model

amandalund:
Follow-on to #1677:
 - Make the Seltzer-Berger upper energy limit configurable.
 - Fix debug assertion failure when only relativistic bremsstrahlung is enabled. This is currently done by extending the lower model limit so it's valid down to the production cut, which prevents us from sampling the brems process and not having any applicable model.

Not fixed:
 - We can still hit the failures with high energy tracks when only Seltzer-Berger enabled, since we can't extend the model limit above the maximum incident energy for the DCS grid (~10 GeV). I did add a warning if only SB is enabled, and a check that the model limit is not above the DCS grid limit.

This should provide a little more flexibility and safety for now, though a real solution will probably need to be a part of our physics grid and model applicability refactor (#907).

Some tests will fail because I regenerated the ROOT test data using a newer Geant4 version—that data will need be updated.


sethrj:
@amandalund regarding the ROOT data: since we have fixed those outstanding 11.3 integration issues, I think it's great to go ahead and start using 11.3 as our "reference" testing data.

amandalund:
Ok will do! I'll open a separate PR to update the data and tests for 11.3.

whokion:
Very nice work!  Looks good to me, but I will actually run a couple of test cases today.  Thanks.

stognini (approved):
Looks great! Thanks @amandalund !

sethrj (approved):

whokion:
Okay, I tested `seltzer_berger_limit` at 1MeV and 10GeV which all look fine.  Thanks for the extending the option.

whokion (approved):

PR #1687, opened 2025-03-21
Update ROOT data and tests with Geant4 version 11.3.0

amandalund:
I reformatted the `RootJsonDumper` and `UrbanMsc` tests in a separate commit so we can easily [see the diff](https://github.com/celeritas-project/celeritas/commit/c79bd6317a8cf07c00e5ef91c1677fc776a24e56), but in the future we can copy/paste the updated results without having to manually reformat them.


.github/workflows/build-spack.yml
```diff
@@ -239,7 +239,7 @@ jobs:
       - name: Regenerate ROOT test data
         if: >-
           ${{
-            matrix.geant == '11.0'
+            matrix.geant == '11.3'
```
sethrj, line 242: In a follow-up (since it will probalby require more tests updating) we need to change all the "11.0" in the `jobs/spack/strategy/matrix/include` to 11.3.

sethrj (approved):
Thanks @amandalund !

PR #1688, opened 2025-03-21
Release v0.6.0

sethrj:
## Pre-merge checklist

- [x] Ensure all CI jobs on the branch to be released (develop or backports/vX.Y) pass.
- [x] If releasing from develop, tag the develop branch with ``vX.Y.Z-rc.N`` where N starts with 1, and increment for every time you return to this step due to new pull requests.
- [x] Run performance regression tests on Perlmutter, Frontier, and an additional machine with debug assertions enabled (e.g., Wildstyle).
- [x] Update documentation with release notes from all pull requests newly included in the release.
- [x] Ensure PDF documentation builds without error.
- [x] For a major release, check for (and delete if found) code marked as "deprecated: to be removed in vX".

## Post-merge checklist

- [ ] ~~If releasing a backported version branch, cherry-pick this documentation commit into the backport branch.~~
- x] Use the [GitHub interface](https://github.com/celeritas-project/celeritas/releases/new) to create a new release with the documentation update that was just added.

## Post-release checklist

- [x] Save the ``tar.gz`` and attach to the release, because the hash changes if the git "describe" function returns a different result for the release tag's hash (e.g., if a collaborative branch on the main repository points to that commit).
- [x] Pull locally (make sure to use the ``--tags`` option) and build PDF user documentation for the release. Ensure breathe is activated (so the API is listed) and that the version is embedded correctly.
- [x] Update the Spack recipe for Celeritas with the new version and sha256 value (either manually or using ``spack checksum``) and submit a [pull request to the Spack project](https://github.com/spack/spack/pull): https://github.com/spack/spack/pull/50197
- [x] Mark the GitHub [release milestone](https://github.com/celeritas-project/celeritas/milestones) as completed.


sethrj:
@mrguilima Github says you left a comment on a commit rather than this PR, but I can't see it. Can you post again using the review functionality above? Thanks!

src/celeritas/em/xs/LPMCalculator.hh
```diff
@@ -26,22 +26,33 @@ namespace celeritas
...
+ * is approximately the energy (using the radiation length \em L ) above which
+ * the LPM effect is significant.
  *
- * See section 10.2.2 of the Geant4 Physics Reference Manual and
- * ComputeLPMfunctions and GetLPMFunctions in G4eBremsstrahlungRelModel and
- * G4PairProductionRelModel. Also see T. Stanev, Ch. Vankov, Development of
- * ultrahigh-energy electromagnetic cascades in water and lead including the
- * Landau-Pomeranchuk-Migdal effect, Phys. Rev. D, 25 (1982), p. 1291.
+ * Calculations of  \f$ \xi(s') \f$ and \f$ s = \frac{s'}{\sqrt{\xi(s')}} \f$
+ * are functional approximations from Eq. 21 in \citet{stanev-lpm-1982,
```
mrguilima:
```suggestion
 * are functional approximations from Eq. 21 in \citet{stanev-lpm-1982},
```

sethrj: `citet` takes two arguments: the first is the bibtex reference; the second (which got wrapped to the next line) is the HTML link (usually DOI). This lets the doxygen code (and vis studio as Stefano pointed out) link directly to the article, while letting the compiled docs point to the bibliography.

sethrj:
@elliottbiondo @amandalund 

GPU throughput for tilecal on frontier dropped substantially:
rel-throughput

Probably due to the reduced occupancy:
rel-occupancy

Here's the increase in memory per thread (again, orange on frontier):
rel-memory

which I assume is due to splines (for `pre-step` and `discrete-select`) and the BVH traversal (for the `along-step`).

elliottbiondo:
@sethrj sorry but can you explain the first plot? I don't understand which results, if any are "before" results.

amandalund:
I believe the first plot is the _ratio_ of the throughputs: new / old.

I wonder why it dropped on the CPU as well for tilecal?

sethrj:
Yes @elliottbiondo , @amandalund has it right: this is the ratio of new to old, so a value of 0.6 means it's 40% slower.

sethrj:
@ptheywood sorry I haven't followed up about the performance results you sent me, but I wanted to ask (since you're the most expert I know in CUDA performance analysis!): do you know of any tool that could use either PTX output or LLVM internals to determine the stack usage for a kernel, resulting in a flame graph that would let us pinpoint why we're using more registers (and driving our occupancy down)? The reduced occupancy seems to have given us a massive performance hit.

esseivaju:
> @ptheywood sorry I haven't followed up about the performance results you sent me, but I wanted to ask (since you're the most expert I know in CUDA performance analysis!): do you know of any tool that could use either PTX output or LLVM internals to determine the stack usage for a kernel, resulting in a flame graph that would let us pinpoint why we're using more registers (and driving our occupancy down)? The reduced occupancy seems to have given us a massive performance hit.

I don't know if you can't get a flame graph out of it, but Nsight compute can display the register usage on a per-line granularity (and I think PTX instructions as well)

ptheywood:
> > @ptheywood sorry I haven't followed up about the performance results you sent me, but I wanted to ask (since you're the most expert I know in CUDA performance analysis!): do you know of any tool that could use either PTX output or LLVM internals to determine the stack usage for a kernel, resulting in a flame graph that would let us pinpoint why we're using more registers (and driving our occupancy down)? The reduced occupancy seems to have given us a massive performance hit.
> 
> I don't know if you can't get a flame graph out of it, but Nsight compute can display the register usage on a per-line granularity (and I think PTX instructions as well)

`ncu`s register usage reporting with `-lineinfo` is the only tool I'm aware of which gives fine-grained register usage, which combined with the mapping to ptx & SASS (when it works) 

I don't think PTX alone is enough to calculate register usage, as SASS is architecture specific and presumably includes optimisations incl register use. It's possible to get the SASS outside the ncu via `cuobjdump`, but SASS is not well documented & there's limited control over how it's generated (from CUDA).

sethrj:
Thanks @ptheywood , I realized that the "use register" vs "spill to local mem" versus "recalculate" is certainly architecture dependent...  but one thing we could do (which spurred this thought) is use the `.reg` listing in the PTX output. The `--keep-device-functions` flag prints out all the device functions, even if in practice they're inlined. For example, the BIH intersecting volume finder includes (with its entire call stack I suppose):
```
    .reg .pred  %p<882>;
    .reg .b16   %rs<433>;
    .reg .f32   %f<389>;
    .reg .b32   %r<637>;
    .reg .f64   %fd<2977>;
    .reg .b64   %rd<1761>;
```
i.e., 882 predicate registers, 2977 double-precision registers... 
and the general quadric sense calculation
```
    .reg .pred  %p<3>;
    .reg .b32   %r<4>;
    .reg .f64   %fd<26>;
    .reg .b64   %rd<3>;
```

But I'm not sure if these numbers are reasonable since they seem to be unoptimized even for an -O3 build, i.e. there seems to be no register reuse.

doc/appendix/release-history/v0.6.rst
```diff
@@ -0,0 +1,263 @@
...
+* Improve error checking and diagnostic output for managing GPU *(@sethrj, #1464)*
+* Add distance-to-bbox function *(@elliottbiondo, #1446)*
+* Add multi-level state to detector step result *(@sethrj, #1471)*
+* Override and save Geant4 dataset variables at configure time *(@pcanal, #1475)*
+* Implement spline interpolation for cross-section calculation *(@lebuller, #1444)*
+* Implement optical Rayleigh model *(@hhollenb, #1469)*
+* Allow core and optical loops to have different numbers of track slots *(@amandalund, #1491)*
+* Add muon decay interactor *(@stognini, #1456)*
+* Add option to use spline-interpolated energy loss *(@lebuller, #1496)*
+* Add option to use spline-interpolated energy loss *(@lebuller, #1496)*
```
amandalund: Duplicate

amandalund (approved):

doc/appendix/release-history/v0.6.rst
sethrj: Yeah that was a weird GitHub glitch. It showed up in the history too

elliottbiondo:
I investigated TileCal more thoroughly (@sethrj, prior, I wasn't "skipping" enough kernels to have enough tracks in flight).

With the BIH there is a 53% decrease in the average active threads per warp in the `AlongStep` kernel, which takes about 40% of the overall runtime. There are also a 66% and 88% decrease in L2 and L1 cache throughput, respectively. The resultant cache misses presumably explain a 4000% increase in warp stalls. Basically, I think for this particular problem, the thread divergence introduced by the BIH outweighs the O(N) -> O(logN) benefits. I am going to look into the structure of the BIH trees for this problem, but I have a few early ideas for fixes:

1) For sufficiently simple universes (e.g., fewer than 50 volumes), don't use a BIH (easy: make every volume an `inf_vol`)
2) If TileCal has universes with huge numbers of volumes, we could try ordering the nodes in [space-filling curve](https://developer.nvidia.com/blog/thinking-parallel-part-iii-tree-construction-gpu/) to get better memory access patterns within a warp (which would only work if threads in a warp represent particles in similar spots spatially).

I will create a BIH -> graphviz utility for this.

amandalund:
@sethrj I dug into the spline a bit; with this diff (removing polynomial spline interpolation) the pre-step kernel occupancy goes up from 0.25 to 0.375:

```diff
diff --git a/src/celeritas/grid/XsCalculator.hh b/src/celeritas/grid/XsCalculator.hh
index 05a477215..ab3cabc56 100644
--- a/src/celeritas/grid/XsCalculator.hh
+++ b/src/celeritas/grid/XsCalculator.hh
@@ -13,7 +13,6 @@
 #include "corecel/grid/UniformGrid.hh"
 #include "corecel/math/Quantity.hh"
 
-#include "SplineCalculator.hh"
 #include "UniformLogGridCalculator.hh"
 #include "XsGridData.hh"
 
@@ -100,16 +99,10 @@ CELER_FUNCTION real_type XsCalculator::operator()(Energy energy) const
     auto const& grid = use_scaled ? data_.upper : data_.lower;
 
     real_type result;
-    if (grid.spline_order == 1)
     {
         // Linear or cubic spline interpolation
         result = UniformLogGridCalculator(grid, reals_)(energy);
     }
-    else
-    {
-        // Spline interpolation without continuous derivatives
-        result = SplineCalculator(grid, reals_)(energy);
-    }
     if (use_scaled)
     {
         result /= value_as<Energy>(energy);
```

Similarly, with this change (removing cubic spline interpolation) the occupancy also goes up to 0.375:

```diff
diff --git a/src/celeritas/grid/InverseRangeCalculator.hh b/src/celeritas/grid/InverseRangeCalculator.hh
index fc514c977..15e5ba173 100644
--- a/src/celeritas/grid/InverseRangeCalculator.hh
+++ b/src/celeritas/grid/InverseRangeCalculator.hh
@@ -12,7 +12,6 @@
 #include "corecel/data/Collection.hh"
 #include "corecel/grid/Interpolator.hh"
 #include "corecel/grid/NonuniformGrid.hh"
-#include "corecel/grid/SplineInterpolator.hh"
 #include "corecel/grid/UniformGrid.hh"
 #include "corecel/math/Algorithms.hh"
 #include "corecel/math/Quantity.hh"
@@ -114,21 +113,12 @@ InverseRangeCalculator::operator()(real_type range) const -> Energy
     CELER_ASSERT(idx + 1 < log_energy_.size());
 
     real_type result;
-    if (deriv_.empty())
     {
         // Interpolate: 'x' = range, y = log energy
         result = LinearInterpolator<real_type>(
             {range_[idx], std::exp(log_energy_[idx])},
             {range_[idx + 1], std::exp(log_energy_[idx + 1])})(range);
     }
-    else
-    {
-        // Use cubic spline interpolation
-        result = SplineInterpolator<real_type>(
-            {range_[idx], std::exp(log_energy_[idx]), deriv_[idx]},
-            {range_[idx + 1], std::exp(log_energy_[idx + 1]), deriv_[idx + 1]})(
-            range);
-    }
     return Energy{result};
 }
 
diff --git a/src/celeritas/grid/NonuniformGridCalculator.hh b/src/celeritas/grid/NonuniformGridCalculator.hh
index 320ff84e4..4e44ad209 100644
--- a/src/celeritas/grid/NonuniformGridCalculator.hh
+++ b/src/celeritas/grid/NonuniformGridCalculator.hh
@@ -125,23 +125,12 @@ CELER_FUNCTION real_type NonuniformGridCalculator::operator()(real_type x) const
     CELER_ASSERT(lower_idx + 1 < x_grid_.size());
 
     real_type result;
-    if (deriv_offset_.empty())
     {
         // Interpolate *linearly* on x using the bin data.
         result = LinearInterpolator<real_type>(
             {x_grid_[lower_idx], (*this)[lower_idx]},
             {x_grid_[lower_idx + 1], (*this)[lower_idx + 1]})(x);
     }
-    else
-    {
-        // Use cubic spline interpolation
-        real_type lower_deriv = reals_[deriv_offset_[lower_idx]];
-        real_type upper_deriv = reals_[deriv_offset_[lower_idx + 1]];
-
-        result = SplineInterpolator<real_type>(
-            {x_grid_[lower_idx], (*this)[lower_idx], lower_deriv},
-            {x_grid_[lower_idx + 1], (*this)[lower_idx + 1], upper_deriv})(x);
-    }
     return result;
 }
 
diff --git a/src/celeritas/grid/RangeCalculator.hh b/src/celeritas/grid/RangeCalculator.hh
index 2066034bb..b20e996d2 100644
--- a/src/celeritas/grid/RangeCalculator.hh
+++ b/src/celeritas/grid/RangeCalculator.hh
@@ -10,7 +10,6 @@
 
 #include "corecel/data/Collection.hh"
 #include "corecel/grid/Interpolator.hh"
-#include "corecel/grid/SplineInterpolator.hh"
 #include "corecel/grid/UniformGrid.hh"
 #include "corecel/math/Quantity.hh"
 
@@ -103,7 +102,6 @@ CELER_FUNCTION real_type RangeCalculator::operator()(Energy energy) const
     CELER_ASSERT(idx + 1 < loge_grid.size());
 
     real_type result;
-    if (data_.derivative.empty())
     {
         // Interpolate *linearly* on energy
         result = LinearInterpolator<real_type>(
@@ -111,17 +109,6 @@ CELER_FUNCTION real_type RangeCalculator::operator()(Energy energy) const
             {std::exp(loge_grid[idx + 1]), this->get(idx + 1)})(
             value_as<Energy>(energy));
     }
-    else
-    {
-        // Use cubic spline interpolation
-        real_type lower_deriv = reals_[data_.derivative[idx]];
-        real_type upper_deriv = reals_[data_.derivative[idx + 1]];
-
-        result = SplineInterpolator<real_type>(
-            {std::exp(loge_grid[idx]), this->get(idx), lower_deriv},
-            {std::exp(loge_grid[idx + 1]), this->get(idx + 1), upper_deriv})(
-            value_as<Energy>(energy));
-    }
     return result;
 }
 
diff --git a/src/celeritas/grid/UniformLogGridCalculator.hh b/src/celeritas/grid/UniformLogGridCalculator.hh
index 297b32f33..05a58a6d0 100644
--- a/src/celeritas/grid/UniformLogGridCalculator.hh
+++ b/src/celeritas/grid/UniformLogGridCalculator.hh
@@ -9,7 +9,6 @@
 #include <cmath>
 
 #include "corecel/grid/Interpolator.hh"
-#include "corecel/grid/SplineInterpolator.hh"
 #include "corecel/grid/UniformGrid.hh"
 #include "corecel/grid/UniformGridData.hh"
 #include "corecel/math/Quantity.hh"
@@ -109,24 +108,12 @@ CELER_FUNCTION real_type UniformLogGridCalculator::operator()(Energy energy) con
     real_type upper_energy = std::exp(loge_grid_[lower_idx + 1]);
 
     real_type result;
-    if (data_.derivative.empty())
     {
         // Interpolate *linearly* on energy
         result = LinearInterpolator<real_type>(
             {lower_energy, (*this)[lower_idx]},
             {upper_energy, (*this)[lower_idx + 1]})(value_as<Energy>(energy));
     }
-    else
-    {
-        // Use cubic spline interpolation
-        real_type lower_deriv = reals_[data_.derivative[lower_idx]];
-        real_type upper_deriv = reals_[data_.derivative[lower_idx + 1]];
-
-        result = SplineInterpolator<real_type>(
-            {lower_energy, (*this)[lower_idx], lower_deriv},
-            {upper_energy, (*this)[lower_idx + 1], upper_deriv})(
-            value_as<Energy>(energy));
-    }
     return result;
 }
 
diff --git a/src/celeritas/grid/XsCalculator.hh b/src/celeritas/grid/XsCalculator.hh
index 05a477215..f34bdf674 100644
--- a/src/celeritas/grid/XsCalculator.hh
+++ b/src/celeritas/grid/XsCalculator.hh
@@ -9,7 +9,6 @@
 #include <cmath>
 
 #include "corecel/grid/Interpolator.hh"
-#include "corecel/grid/SplineInterpolator.hh"
 #include "corecel/grid/UniformGrid.hh"
 #include "corecel/math/Quantity.hh"
```

With both applied it's back up to 0.5. The performance impact seemed relatively small: on testem15 I saw ~3% difference.

sethrj:
Cool! Thanks for that. I would guess we might see the cost a bit more in a "reactor physics"-y problem, with lots of materials/elements but relatively simple geometry. Good to know though!

amandalund:
Yeah, in the types of problems that are relevant to us we probably won't see as much of an effect. To clarify, that few percent is the performance difference with spline disabled vs. removed from the code (the impact of the reduced occupancy, and on NVIDIA, not AMD); disabled vs. enabled there's maybe a percent or so difference, from what I've seen.

PR #1689, opened 2025-03-21
Fix out-of-bounds memory access in `CylMapField`

amandalund:
Because `phi` could be exactly equal to the last grid point, we could try to interpolate off the end of the grid.


src/celeritas/field/CylMapField.hh
```diff
@@ -90,10 +90,7 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
     real_type r = hypot(pos[0], pos[1]);
     // Ensure phi is in [0, 2\f$\pi\f$)
     Turn phi = atan2turn(pos[1], pos[0]);
-    if (phi < zero_quantity())
-    {
-        phi.value() += 1;
-    }
+    phi.value() = eumod(phi.value(), real_type(1));
```
sethrj: Looking at this again: under what circumstances does phi end up as `Turn{1.0f}`?  Is it possible that the grid isn't exactly `real_type{1}` at the high end? I see `CylMapFieldParams` validates soft zero and soft unity; I think we need to *replace* those values with exact zero and exact unity. Maybe then the `eumod` (which may be overkill... especially now that it needs another edge case) will be unnecessary?

esseivaju: yeah, the reason for using soft equivalence was that depending on how users compute their bins, there could be numerical imprecisions. Given that we want a full circle, I think it would make sense to replace these values when building the params.

amandalund: In this case both phi and the last grid point are exactly 1. It's happening when phi is a negative value smaller than machine epsilon, e.g:
```console
pos[0]=1.3210877613973844e+02, pos[1]=-3.7097619497689957e-15, atan2turn=-4.4692485181714004e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.6032440925778337e+02, pos[1]=-1.2431103748723196e-14, atan2turn=-7.6000234297596259e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.6032440925778337e+02, pos[1]=-1.2431103748723196e-14, atan2turn=-7.6000234297596259e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.6032440925778337e+02, pos[1]=-1.2431103748723196e-14, atan2turn=-7.6000234297596259e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.7677069035245967e+02, pos[1]=-1.2499877662246627e-14, atan2turn=-7.1879624083642991e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.7677069035245967e+02, pos[1]=-1.2499877662246627e-14, atan2turn=-7.1879624083642991e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
pos[0]=2.7677069035245967e+02, pos[1]=-1.2499877662246627e-14, atan2turn=-7.1879624083642991e-18, phi=1.0000000000000000e+00, grid.back()=1.0000000000000000e+00
```
Yeah, definitely safer to check _and_ replace those values, I'll add that.

sethrj: I see, that makes sense. What about `if (phi.value() + 1 < 1) { phi.value() += 1; }`? and hope that the optimizers don't mess that up for us? 😅

sethrj: alternatively I guess `phi.value() < -numeric_limits<real_type>::epsilon()`  🤔

amandalund: We could do this 🤷 
```c++
Turn phi = atan2turn(pos[1], pos[0]);
if (phi < zero_quantity())
{
    phi.value() += 1;
}
if (CELER_UNLIKELY(phi.value() == 1)) 
{
    phi.value() = 0;
}
CELER_ASSERT(phi >= zero_quantity() && phi.value() < 1);
```

sethrj: Yeah I like that, simple enough. (Add a comment in the second conditional that this happens for `-epsilon <= phi < 0`.)

sethrj (approved):
Thanks!

PR #1690, opened 2025-03-21
Change default field substeps in accel

esseivaju:
As discussed, we probably missed changing the default number of substeps in accel.


amandalund (approved):
Thanks for catching that!

sethrj (approved):

PR #1691, opened 2025-03-21
Remove deprecations for v0.6

sethrj:
- Remove macro input from celer-g4
- Remove deprecated internal interfaces
- Push back deprecation for the removed shared params values, and add warnings about when they'll be deleted


amandalund (approved):
Thanks @sethrj!

PR #1692, opened 2025-03-22
Force loading of the dl library -- needed in some configs with gcc13

mrguilima:
This change fixes some build errors seen in Perlmutter, using gcc13:
```console
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: test/corecel/CMakeFiles/corecel_data_ObserverPtr.dir/data/ObserverPtr.test.cu.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /lib64/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
It then affected Range, StackAllocator and ObserverPtr tests.

I first saw this a couple weeks ago, and fixed it with this change.  I thought this was fixed some other way, as I cannot see the original problem anymore with current heads of Celeritas (commit f96ecb87) and VecGeom (commit c3ee06912).  But yesterday Soon reported seeing the same error, and this change also fixed his problem.

Should this be merged?

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -49,7 +49,13 @@ function(celeritas_add_device_test base)
   if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
     set(_cuda_args GPU SOURCES "${base}.test.cu")
   endif()
-  celeritas_add_test("${base}.test.cc" ${_cuda_args} ${ARGN})
+    # GCC 13 requires some tests to have -ldl explicitely added
+    if(${base} MATCHES "Range"
+      OR ${base} MATCHES "StackAllocator"
+      OR ${base} MATCHES "ObserverPtr")
+          set(ARGN LINK_LIBRARIES dl)
+    endif()
+    celeritas_add_test("${base}.test.cc" ${_cuda_args} ${ARGN})
 endfunction()
 
 if(CELERITAS_UNITS STREQUAL "CGS")


test/CMakeLists.txt
```diff
@@ -49,7 +49,13 @@ function(celeritas_add_device_test base)
   if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
     set(_cuda_args GPU SOURCES "${base}.test.cu")
   endif()
-  celeritas_add_test("${base}.test.cc" ${_cuda_args} ${ARGN})
+    # GCC 13 requires some tests to have -ldl explicitely added
+    if(${base} MATCHES "Range"
+      OR ${base} MATCHES "StackAllocator"
+      OR ${base} MATCHES "ObserverPtr")
+          set(ARGN LINK_LIBRARIES dl)
```
pcanal, line 56: If it is needed, it should be spelt `${CMAKE_DL_LIBS}`
```suggestion
          set(ARGN LINK_LIBRARIES ${CMAKE_DL_LIBS})
```

pcanal:
What configuration is still problem happening with?

test/CMakeLists.txt
sethrj: Echo: https://github.com/celeritas-project/celeritas/pull/1587#discussion_r1921267188

sethrj:
We've seen this before in https://github.com/celeritas-project/celeritas/pull/1587 . @mrguilima also told me this on slack on 6 March:

> In my Perlmutter build, using gcc 13.2.0, the celeritas build against vg-2.x fails with undefined reference to symbol 
> ```
> 'dlclose@@GLIBC_2.2.5'.
> [3/3] Linking CXX executable test/corecel/data_ObserverPtr
> FAILED: test/corecel/data_ObserverPtr
> : && /usr/bin/g++ -Wall -Wextra -pedantic -O3 -DNDEBUG  test/corecel/CMakeFiles/corecel_data_ObserverPtr.dir/data/ObserverPtr.test.cc.o test/corecel/CMakeFiles/corecel_data_ObserverPtr.dir/data/ObserverPtr.test.cu.o CMakeFiles/corecel_data_ObserverPtr.dir/cmake_device_link.o -o test/corecel/data_ObserverPtr -L/global/homes/l/lima/work/spack-2502/var/spack/environments/cele-gcc13/.spack-env/view/targets/x86_64-linux/lib/stubs   -L/global/homes/l/lima/work/spack-2502/var/spack/environments/cele-gcc13/.spack-env/view/targets/x86_64-linux/lib -Wl,-rpath,/global/homes/l/lima/work/cele/build-vg2.x/test/corecel:/global/homes/l/lima/work/cele/build-vg2.x/test:/global/homes/l/lima/work/cele/build-vg2.x/lib64:/global/homes/l/lima/work/spack-2502/var/spack/environments/cele-gcc13/.spack-env/view/lib64  test/corecel/libtestcel_core.so  test/libtestcel_harness.so  lib64/libcorecel.so  /global/homes/l/lima/work/spack-2502/var/spack/environments/cele-gcc13/.spack-env/view/lib64/libgtest.so.1.15.2  -lpthread  -lcudadevrt  -lcudart && :
> /usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: test/corecel/CMakeFiles/corecel_data_ObserverPtr.dir/data/ObserverPtr.test.cu.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
> /usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /lib64/libdl.so.2: error adding symbols: DSO missing from command line
> collect2: error: ld returned 1 exit status
> ninja: build stopped: subcommand failed.
> ```
> The same message also shows up for cont_Range  and data_StackAllocator tests.
> This patch fixes this problem:
> ```
> diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
> index 0194e35f..a3735ceb 100644
> --- a/test/CMakeLists.txt
> +++ b/test/CMakeLists.txt
> @@ -55,6 +55,12 @@ function(celeritas_add_device_test base)
>    if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
>      set(_cuda_args GPU SOURCES "${base}.test.cu")
>    endif()
> +  # GCC 13 requires some tests to have -ldl explicitely added
> +  if(${base} MATCHES "Range"
> +    OR ${base} MATCHES "StackAllocator"
> +    OR ${base} MATCHES "ObserverPtr")
> +        set(ARGN LINK_LIBRARIES dl)
> +  endif()
>    celeritas_add_test("${base}.test.cc" ${_cuda_args} ${ARGN})
>  endfunction()
> ```
> but maybe this is not the best way to handle this error?  Do you have any other suggestion?

The proposed solution is definitely *not* the right way to fix the problem... it has nothing to do with the names of the test. One difference is that all three libs *do* use thrust, whereas `NumericLimits.test.cu` does *not* use Thrust. Let me try an alternative that defines an ExtThrust and links these libraries against it.

test/CMakeLists.txt
```diff
@@ -49,7 +49,13 @@ function(celeritas_add_device_test base)
   if(CELERITAS_USE_CUDA OR CELERITAS_USE_HIP)
     set(_cuda_args GPU SOURCES "${base}.test.cu")
   endif()
-  celeritas_add_test("${base}.test.cc" ${_cuda_args} ${ARGN})
+    # GCC 13 requires some tests to have -ldl explicitely added
+    if(${base} MATCHES "Range"
+      OR ${base} MATCHES "StackAllocator"
+      OR ${base} MATCHES "ObserverPtr")
```
sethrj, line 55: As noted, the name of the test doesn't determine the requirement for `-ldl`.

sethrj (changes requested):

PR #1693, opened 2025-03-22
Add timers to accel integration classes

amandalund:
This move the timer output class from celer-g4 to our accel library and adds timers to the integration classes.

### DIFF
--- a/src/accel/TimeOutput.hh
+++ b/src/accel/TimeOutput.hh
@@ -2,30 +2,28 @@
 // Copyright Celeritas contributors: see top-level COPYRIGHT file for details
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file celer-g4/TimerOutput.hh
+//! \file accel/TimeOutput.hh
 //---------------------------------------------------------------------------//
 #pragma once
 
 #include <unordered_map>
 #include <vector>
-#include <G4Event.hh>
 
 #include "corecel/Types.hh"
 #include "corecel/io/OutputInterface.hh"
 
 namespace celeritas
 {
-namespace app
-{
 //---------------------------------------------------------------------------//
 /*!
  * Collect timing results and output at the end of a run.
  *
- * Setup time, total time, and time per event are always recorded. The
- * accumulated action times are recorded when running on the host or on the
- * device with synchronization enabled.
+ * Setup time and total time are always recorded. Event time is recorded of \c
+ * BeginOfEventAction and \c EndOfEventAction are called. The accumulated
+ * action times are recorded when running on the host or on the device with
+ * synchronization enabled.
  */
-class TimerOutput final : public OutputInterface
+class TimeOutput final : public OutputInterface
 {
   public:
     //!@{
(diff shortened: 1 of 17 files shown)


amandalund:
Output from a GPU run of the ATLAS benchmark with a single event:
```json
{
  "_index": "thread",
  "actions": [
    {
      "along-step-cylmap-msc": 9.432758654999997,
      "along-step-neutral": 3.626884553999997,
      "annihil-2-gamma": 0.01385919199999998,
      "brems-rel": 0.016702715999999996,
      "brems-sb": 0.02543592499999999,
      "conv-bethe-heitler": 0.01846758300000001,
      "extend-from-primaries": 0.011505055000000002,
      "extend-from-secondaries": 0.08121516800000006,
      "geo-boundary": 1.1247606160000017,
      "initialize-tracks": 0.7602331899999996,
      "ioni-moller-bhabha": 0.017333870000000026,
      "photoel-livermore": 0.04971534700000005,
      "physics-discrete-select": 0.029856576999999995,
      "pre-step": 0.08793576500000004,
      "scat-klein-nishina": 0.017194660999999997,
      "scat-rayleigh": 0.016895440999999997,
      "tracking-cut": 0.010173201000000022
    }
  ],
  "events": [
    []
  ],
  "setup": 127.782590092,
  "total": 45.275981924
}
```

esseivaju:
that's a large fraction in propagation 😅 This is with the "calo" geometry?

src/accel/detail/IntegrationSingleton.hh
```diff
@@ -83,6 +85,12 @@ class IntegrationSingleton
     // Destroy params
     void finalize_shared_params();
 
+    // Start the transport timer
+    void start_timer();
+
+    // Get the transport time
+    real_type stop_timer();
```
sethrj: Might as well inline these; and also, units are seconds right? (not native 😜)

src/accel/TimeOutput.hh
```diff
@@ -2,30 +2,28 @@
...
-namespace app
-{
 //---------------------------------------------------------------------------//
 /*!
  * Collect timing results and output at the end of a run.
  *
- * Setup time, total time, and time per event are always recorded. The
- * accumulated action times are recorded when running on the host or on the
- * device with synchronization enabled.
+ * Setup time and total time are always recorded. Event time is recorded of \c
```
sethrj:
```suggestion
 * Setup time and total time are always recorded. Event time is recorded if \c
```

src/accel/TimeOutput.hh
```diff
@@ -49,16 +47,16 @@ class TimerOutput final : public OutputInterface
     //!@}
 
     // Record the accumulated action times
-    void RecordActionTime(MapStrReal&& time);
+    void record_action_time(MapStrReal&& time);
```
sethrj: I guess this is more of a "power user" class now (anyone downstream who doesn't want to use the integration classes) but we need to differentiate more clearly what the line is between "G4-style" and "Celeritas style"

sethrj (approved):
Nice! Only point of contention is possibly the name update, but I think just marking the class as for "internal use" is good enough for now. Thanks!

src/accel/TimeOutput.hh
amandalund: Yeah I got 90% through and then wondered if I should have changed it... I'm happy to go either way.

sethrj: I'd say since it should be easy to revert it and it lowers the change count and we aren't sure... maybe do a quick reversion? Sorry for the extra work 🫤

src/accel/detail/IntegrationSingleton.hh
amandalund: Added the units here and in the json output.

PR #1694, opened 2025-03-22
Add ExtThrust for more explicit dependencies

sethrj:
This is a proposed solution to the problem motivating #1692  (see https://github.com/celeritas-project/celeritas/pull/1692#issuecomment-2745408133).  ~~@mrguilima @whokion @esseivaju could one of you please see if this works?~~

I think this is safe to merge; it simply makes the dependencies clearer.


esseivaju:
I haven't needed the explicit `ldl` in my CMakePreset, since https://github.com/celeritas-project/celeritas/pull/1661

mrguilima:
I also haven't seen the problem for a while.  I thought the problem was fixed by some other commit, but then I was surprised that @whokion was seeing it.  I hope he can test it.

pcanal:
Are those libraries already depending on `ExtThurst`?   Does the build still suceed if `Thrust` is not installed?

sethrj:
@pcanal these all depend on thrust if cuts or rocm is enabled, but most distributions include them in the paths by default so don't need to explicitly link.

pcanal (approved):
I agree this can be merged even-though it still feels 'weird' to have to be this explicit.

sethrj:
I think @wdconinc saw on some spack setups that an explicit thrust dependency was needed?

PR #1695, opened 2025-03-24
Fix setup options messenger in MT mode

sethrj:
It turns out that macro options to `/celer/...` were not being used in MT mode. The culprit was a secret `protected` variable in the base class that had to be set.

I also incorrectly placed the "slotDiagnostics" option and missed updating some documentation.


amandalund (approved):
Good catch, thanks @sethrj!

src/accel/SetupOptionsMessenger.cc
```diff
@@ -152,6 +153,10 @@ SetupOptionsMessenger::SetupOptionsMessenger(SetupOptions* options)
 {
     CELER_EXPECT(options);
 
+    // Setup options affect the shared parameters: if this base-class protected
+    // variable isn't set, commands will be ignored :(
+    this->commandsShouldBeInMaster = true;
```
pcanal, line 158: Question for Geant4 ... where the is the documentation for that data member!?  Question for Seth, how did you find out?

Side note, the name 'sounds' like it would  execute only the command done/request directly from the main 'thread'  and ignore those issue by other threads. I.e. it sounds like 'true' would be "more" restrictive that 'false' ....

amandalund:
Although actually I had been using these ui commands in MT mode with BenchmarkEm3 and they seem to be working...

src/accel/SetupOptionsMessenger.cc
```diff
@@ -152,6 +153,10 @@ SetupOptionsMessenger::SetupOptionsMessenger(SetupOptions* options)
 {
     CELER_EXPECT(options);
 
+    // Setup options affect the shared parameters: if this base-class protected
+    // variable isn't set, commands will be ignored :(
```
pcanal, line 157: Is there plausible test we can add to avoid this problem from re-occurring?

src/accel/SetupOptionsMessenger.cc
sethrj: I noticed that the "apply" methods weren't being called in our benchmark, I dug around in the UI code and saw this guy... and I remember at some point seeing some really weird logic for the multithread case.

sethrj:
Maybe I imagined it ... I'm not able to reproduce. I think I put quotes in the output names so didn't see them, and I probably missed the debug output message among the barrage of text.

PR #1697, opened 2025-03-26
Allow some "required" SetupOptions to be set via Geant4 UI

sethrj:
This allows the initializers/track slots to be empty when calling SetupOptions.


amandalund (approved):
thanks @sethrj!

example/offload-template/src/MakeCelerOptions.cc
```diff
@@ -43,17 +43,11 @@ celeritas::SetupOptions MakeCelerOptions()
 
     opts.max_num_tracks = 1024 * 16;
     opts.initializer_capacity = 1024 * 128 * 4;
-    opts.secondary_stack_factor = 2.0;
-    opts.ignore_processes = {"CoulombScat"};
```
amandalund, line 47: We'll still need to ignore Coulomb scattering for now (#1391)

stognini (approved):
Nice addition. Thanks!

PR #1700, opened 2025-03-27
Update and fix G4VG external integration

sethrj:
When using G4VG as an "external", it was alway enabling debug and testing, which put an artifical GTest requirement on downstream systems.

This also updates the G4VG version to 1.0.4 which has a few minor RDC fixes and better version.

Finally, it adds build provenance to the config output.


pcanal (approved):
Thanks.

PR #1701, opened 2025-03-27
Improve doc generation and brems-related documentation

sethrj:
Some of these changes were made for the presentation yesterday. It improves the doc auto-regeneration as well.


amandalund (approved):
Thanks @sethrj!

PR #1702, opened 2025-03-27
Support counter with NVTX

esseivaju:
Counters could be useful to have in GPU builds, for example correlating active tracks with kernel duration. 

We don't support perfetto in CUDA build, so use NVTX mark instead. The way NVTX mark are displayed on Nsight system is not as nice as perfetto, as it's just a tick and you need to hover to see the value, whereas perfetto plots the value. 

Eventually, it could be nice to support Perfetto in CUDA build, we'd have to rework our `ScopedProfiling` to pick profiling backend to use instead of automatically defaulting to NVTX if CUDA is enabled. Maybe we could simply pick perfetto over NVTX if `CELERITAS_USE_PERFETTO` is set.

The screenshot below shows how NVTX marks are displayed on Nsight system

image

Note that similarly to Perfetto, we can use sqlite to query events and do some analysis:
```sql
sqlite> 
select start, uint32Value  FROM NVTX_EVENTS as e 
JOIN ENUM_NSYS_EVENT_TYPE as et, StringIds as sid 
ON (et.id = e.eventType AND e.textId = sid.id) 
WHERE (et.name = 'NvtxMark' AND sid.value = 'active-0') 
ORDER BY start ASC 
LIMIT 20;
58849482|100
60825006|100
62153056|196
63172409|297
64229094|400
65184323|497
66358987|607
67496749|747
68631294|901
69972701|1065
71383291|1234
72769584|1413
74268024|1614
75965129|1824
77631404|2082
79464432|2368
81332287|2662
83364150|3053
85727955|3467
88108382|3914
```


src/corecel/sys/TraceCounter.cuda.cc
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Records a named value at the current timestamp which
+ * can then be displayed on a timeline. Only supported on host
+ */
+template<class T>
+void trace_counter(char const* name, T value)
+{
+    static_assert(std::is_arithmetic_v<T>, "Only support numeric counters");
+    if (use_profiling())
+    {
```
sethrj: It would improve performance in the "no profiling" case for the compiler to be able to inline this check, and it would also make the function a little more readable. Would it make sense for the main `trace_counter` template function to be `if (use_profiling()) { detail::trace_counter_impl(value); }`? 

Eh on second thought I think not: if you're building with profiling on you're probably not going to care about performnance with the tools off? And it's going to be coarse grained enough that an extra function call won't matter.

Still though, I think with this function length it would improve readability to use `if (!use_profiling()) { return; }`.

src/corecel/sys/TraceCounter.cuda.cc
```diff
@@ -0,0 +1,89 @@
...
+        else if constexpr (std::is_same_v<T, std::int32_t>)
+        {
+            attributes.payloadType = NVTX_PAYLOAD_TYPE_INT32;
+            attributes.payload.iValue = value;
+        }
+        else if constexpr (std::is_same_v<T, std::int64_t>)
+        {
+            attributes.payloadType = NVTX_PAYLOAD_TYPE_INT64;
+            attributes.payload.llValue = value;
+        }
```
sethrj: Perhaps to improve readability:
```suggestion
#define TC_SET_PAYLOAD(TYPE, NVTX_CONST, MEMBER)            \
    if constexpr (std::is_same_v<T, TYPE>)                  \
    {                                                       \
        attributes.payloadType = NVTX_CONST;                \
        attributes.payload.MEMBER = value; \
    }

    TC_SET_PAYLOAD(std::uint32_t, NVTX_PAYLOAD_TYPE_UNSIGNED_INT32, uiValue)
    else TC_SET_PAYLOAD(std::uint64_t, NVTX_PAYLOAD_TYPE_UNSIGNED_INT64, ullValue)
    else TC_SET_PAYLOAD(float,        NVTX_PAYLOAD_TYPE_FLOAT,          fValue)
    else TC_SET_PAYLOAD(double,       NVTX_PAYLOAD_TYPE_DOUBLE,         dValue)
    else TC_SET_PAYLOAD(std::int32_t, NVTX_PAYLOAD_TYPE_INT32,          iValue)
    else TC_SET_PAYLOAD(std::int64_t, NVTX_PAYLOAD_TYPE_INT64,          llValue)

#undef TC_SET_PAYLOAD
```

sethrj (approved):
Neat!

src/corecel/sys/TraceCounter.cuda.cc
esseivaju: Profiling is always on at build time if we build with CUDA/HIP support, so I refactored that a bit

src/corecel/sys/TraceCounter.cuda.cc
esseivaju: Or we can avoid macros with this 😅 
```c++
template<typename T>
void set_nvtxattributes_payload(nvtxEventAttributes_t& attr, T val);

template<>
void set_nvtxattributes_payload<std::uint32_t>(nvtxEventAttributes_t& attr,
                                               std::uint32_t val)
{
    attr.payloadType = NVTX_PAYLOAD_TYPE_UNSIGNED_INT32;
    attr.payload.uiValue = val;
}

template<>
void set_nvtxattributes_payload<std::uint64_t>(nvtxEventAttributes_t& attr,
                                               std::uint64_t val)
{
    attr.payloadType = NVTX_PAYLOAD_TYPE_UNSIGNED_INT64;
    attr.payload.ullValue = val;
}
```

sethrj:
Please remember not to force push! Screenshot 2025-03-29 at 06 27 50

src/corecel/sys/TraceCounter.cuda.cc
sethrj: Yeah I thought about a struct specialization too (ps you don't even need templating above, you can overload) but for these bunches of changes where we're just trying to compress the text on the screen to *improve* readability, I think macros are appropriate.

PR #1703, opened 2025-03-28
Write unique filenames to JSON output from accel

sethrj:
@whokion pointed out that rerunning the same Celeritas job will delete the previous output file unless done carefully. This adds an option to use a unique filename instead of over-writing. 🐿️


src/corecel/io/detail/UniqueFileNamer.cc
```diff
@@ -0,0 +1,54 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with original filename.
+ */
+UniqueFileNamer::UniqueFileNamer(std::string const& filename)
+{
+    CELER_EXPECT(!filename.empty());
+    std::ostringstream stem;
+
+    if (auto pos = filename.find_last_of('.');
```
pcanal, line 28: Since we already use `std::filesystem` elsewhere we could use `std::filesystem::path::stem` and `std::filesystem::path::extension`

src/corecel/io/FileOrConsole.cc
```diff
@@ -0,0 +1,115 @@
...
+    flags |= std::ios::noreplace;
+#else
+    // C++17/C++20: check non-atomically if the file exists
+    if (std::filesystem::exists(filename))
+    {
+        // Do not open
+        return ofs;
+    }
+#endif
+    ofs.open(filename, flags);
```
pcanal, line 39:
```suggestion
#if __cplusplus >= 202302L
    ofs.open(filename, flags);
#else
    flags |= std::ios::ate;
    ofs.open(filename, flags);
    auto pos = ofs.tellp()
    if (pos != 0) // In case of rare race condition.
       return std::ofstream{};
#endif
```

src/accel/SharedParams.cc
```diff
@@ -771,38 +763,41 @@ void SharedParams::set_num_streams(unsigned int num_streams)
 /*!
  * Write available Celeritas output.
  *
- * This can be done multiple times, overwriting the same file so that we can
- * get output before construction *and* after.
+ * The first call should provide an output filename. If it's to a file, an
+ * initial copy of the output (the setup) will be written in case of later
+ * failure. If the file already exists, a different filename will be used.
```
amandalund, line 472: Is it possible to make the unique filename an option that's disabled by default? It can create a lot of file clutter, and it seems easy enough to change the output filename when needed in a script.

amandalund (commented):
Currently I'm running into:
```console
status: Celeritas core setup complete
terminate called after throwing an instance of 'std::bad_alloc'
```

src/accel/SharedParams.cc
sethrj: The motivating reason for this PR is that it's super easy to run a bunch of Geant4 jobs with the macros and unwittingly lose every output but the last, so I'm hesitant to make the default to overwrite. Maybe I didn't think this through, but perhaps a middle ground (easier to implement too if I'd thought of it before) would be just to generate a default timestamped filename if the user doesn't specify a filename in the input? Should we still make uniquifying an option?

src/corecel/io/FileOrConsole.cc
sethrj: Clever! I guess it only catches the race condition of someone already flushing to disk so the file system can read it: it seems quite possible that multiple could open it but one might not write yet.

src/corecel/io/detail/UniqueFileNamer.cc
sethrj: Good call.  I'll change here and have add a note to follow-on to check for other uses in the codebase.

src/accel/SharedParams.cc
amandalund: I suppose, although any other output (e.g. root output) would still be overwritten, and it also seems super easy for the user to adjust the output filenames to be unique :). I do like that idea for a unique default filename better since if a user does specify a filename the output will be what's expected.

src/corecel/io/FileOrConsole.cc
pcanal: Indeed, it is not foolproof but sounded simple enough while reducing somewhat the probability of trouble.  We could go further (but still not perfect) by immediately writing and flushing some marker content. To get to 'perfection' we would probably need to use some type of full fledge '[file lock' mechanism](https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/sync/file_lock.hpp)' (See link for an example).
But this sounds overboard for code that disappear as soon as C++20.

src/accel/SharedParams.cc
sethrj: Good points! I'll make those changes.

src/corecel/io/FileOrConsole.cc
sethrj: 😄 We'd probably have to hook into the OS. Anyway, from what I've seen with typical id and file access patterns I'm skeptical this suggestion would reduce the race conditions by any appreciable amount... usually the time between querying the os and opening the file would be much less than the time for someone else to flush anything to it? This just seems like overkill.

sethrj:
OK I think this is going to be too difficult to solve well.

src/corecel/io/FileOrConsole.cc
pcanal: It is likely overkill/not-that-useful especially since it goes away after we move to C++23.

>  than the time for someone else to flush anything to it?

It it not about the user doing something explicitly but about this same code running at the same time from multiple threads or processes on the same machine (or not).  For example 2 processes starting at the same time are not unlikely to get here at the 'same' time.

sethrj: I was personifying the other threads 😅

PR #1704, opened 2025-03-28
Fix OBZ bugs

elliottbiondo:
This MR fixes two bugs that pertain to oriented bounding zones (OBZs),  which ostensibly may be used at some point for safety distance calculations.

The first bug is actually in `BoundingBoxUtils`, but was brought to light via as OBZ error when attempting to ray trace XLZD. Basically `calc_center(bbox)` was not properly handling bboxes with infinite values.

The second bug is that safety distances are currently being calculated with respect to the wrong part (inner/outer) of the OBZ. The correct method is illustrated in the diagram below. Even if we end up not using OBZs, I figured I would fix this just so we don't have incorrect code floating around.

safety_distance

@sethrj w.r.t. negation for OBZs: if I understand correctly, `BoundingZones` (i.e., not OBZs) must support negatation because each CSG node has one; anytime you have a subtracted region, the bounding zone is negated. However, we don't have OBZs for every region: OBZs only exist for the physical volumes. So unless I am misunderstanding we should not need a negation flag for OBZs.


src/orange/BoundingBoxUtils.hh
```diff
@@ -76,17 +76,33 @@ inline bool is_degenerate(BoundingBox<T> const& bbox)
...
+
     Array<T, 3> center;
     for (auto ax : range(to_int(Axis::size_)))
     {
-        center[ax] = (bbox.lower()[ax] + bbox.upper()[ax]) / 2;
+        if (CELER_UNLIKELY(isinf(bbox.lower()[ax]) || isinf(bbox.upper()[ax])))
+        {
+            // There is no useful way of defining the midpoint between a finite
+            // and infinite value
+            CELER_ASSERT(isinf(bbox.lower()[ax]) && isinf(bbox.upper()[ax]));
```
sethrj: Should probably be expect

sethrj (approved):
It would be nice to have a higher level (unit proto) test demonstrating the actual failure: showing we get a NaN originally then fixing it. I wonder if the better solution would be to prohibit infinities entirely using the "truncate to parent" logic.

sethrj:
@elliottbiondo feel free to merge if you've verified this fixes the NaN problem and thought about the comment above! You should be able to hit the merge button...

elliottbiondo:
What I have eliminates the NaNs. I agree that we should be intersecting each BIH with the unit's BBox, and it's easy to do (realistically like 1-2 hours by the time I add unit tests). Should I go ahead and do it, or add "TODO" for this in `process_obz_record` since we are unsure I about the future of OBZ?

sethrj:
Let's postpone it. Plenty else to do 😪

PR #1705, opened 2025-03-30
Reduce logging verbosity during Geant4 runs

sethrj:
This improves the consistency of the user of CELER_LOG and CELER_LOG_LOCAL and reduces the logging frequency to improve readability. The "local" output is now only for events that really only happen on a single thread (for a particular event/track). Other things that *should* be the same on all threads, even though they may not occur simultaneously, are now on CELER_LOG. A new environment variable `CELER_LOG_ALL_LOCAL` enables extra verbosity so that all threads will print those messages. In a multithread run, they will still be tagged with `[M]` or `[W]` for manager/worker and may print more than once: for example, the tracking manager setup happens on both a worker and manager thread . (@drbenmorgan is this right? should `TrackingManagerConstructor::ConstructProcess` return early if we're the master thread?)

I've also changed the hit reconstruction code to print once per event rather than once per step, and I've eliminated the debug messages from the Navi reconstructor. The final result of all this is a much cleaner output where it's now feasible to use debug log messages without being completely overwhelmed by output:
Screenshot 2025-03-30 at 10 26 35


amandalund (commented):
Much nicer, thanks! I'd been manually disabling the hit processor logging because it was so much output...

Tested in an external app and it looks good, though in celer-g4 I'm still seeing the log messages on all worker threads.

app/celer-g4/LogHandlers.cc
```diff
@@ -16,11 +16,34 @@ namespace celeritas
...
+               LogProvenance const& prov,
+               LogLevel const& lev,
+               std::string const& msg)
+{
+    // Write main message
+    StreamLogHandler{os}(prov, lev, msg);
+
+    // Lock after building the message while writing
+    static std::mutex log_mutex;
+    std::lock_guard scoped_lock{log_mutex};
```
pcanal, line 36: Do we get here for every 'potential' message or is the code short-circuited early for message that are to be ignored?

sethrj: This only applies if the log level is "visible": the only way it won't write now is to not deliberately not print.

sethrj:
@amandalund The celer-g4 app doesn't use the `accel` logger; it uses its own in-house. I'm updating it to have roughly the same functionality but to be MPI-aware. Woudl you like that in a separate commit?

amandalund:
I see, thanks @sethrj! Whichever you prefer :)

sethrj:
I updated on this branch, and suppressed/deleted even more log messages that aren't very helpful.

sethrj:
I can't reproduce these crashes :(

sethrj:
OK figured out the crash: before this, the world logger didn't route through G4cerr; and now, when diagnostic-level output is shown, the GDML writer's use of ScopedGeantLogger leads to recursion with the log adapter.

src/corecel/io/Logger.hh
```diff
@@ -99,8 +104,11 @@ class Logger
     //! Get the default log level
     static constexpr LogLevel default_level() { return LogLevel::status; }
 
-    // Construct with default celeritas communicator
-    explicit Logger(LogHandler handle);
+    // Create a logger from a handle and level environment variable
+    static Logger from_handle_env(LogHandler&& handle, std::string const& key);
+
+    // Construct from an our put handle
```
amandalund:
```suggestion
    // Construct from an output handle
```

src/celeritas/ext/detail/GeantProcessImporter.cc
```diff
@@ -377,10 +373,6 @@ GeantProcessImporter::operator()(G4ParticleDefinition const& particle,
     {
         if (G4VEmModel* model = process.GetModelByIndex(i))
         {
-            CELER_LOG(debug) << "Saving MSC model '" << model->GetName()
-                             << "' for particle " << particle.GetParticleName()
-                             << " (" << particle.GetPDGEncoding() << ")";
-
```
amandalund: I think the log messages in the importer are still helpful when we're not using one of our own physics lists.

app/celer-g4/LogHandlers.cc
```diff
@@ -0,0 +1,161 @@
...
+#include "geocel/GeantUtils.hh"
+
+namespace celeritas
+{
+namespace app
+{
+namespace
+{
+//---------------------------------------------------------------------------//
+//! Finalize the message, including
```
amandalund: 🐿️

amandalund (approved):
Thanks @sethrj!

src/celeritas/ext/detail/GeantProcessImporter.cc
sethrj: My thought was we're able to see that kind of debugging information in the output json file, and I was seeing a lot of duplication of these messages for each particle type: what about I print one info message per particle type listing the models loaded?

amandalund: Sure, that sounds reasonable.

PR #1707, opened 2025-03-31
Use programmatic optical counters instead of absurdly verbose printout

sethrj:
Following on to #1705 (but to be merged first) this accumulates optical counters for user-facing statistics over the course of the simulation, rather than printing verbose log messages at every core step. It also prints out warnings at the end of the transport loop if some photons were never tracked (a "bug" [it's intentional for now in the celer-sim test] which we need to fix!)

### DIFF
--- a/app/celer-sim/Transporter.cc
+++ b/app/celer-sim/Transporter.cc
@@ -203,22 +203,34 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries) -> TransporterResult
     // Get optical counters
     if (optical_)
     {
-        auto counters
-            = optical_->exchange_counters(stepper_->sp_state()->aux());
+        auto& aux = stepper_->sp_state()->aux();
+        auto counters = optical_->exchange_counters(aux);
         auto const& gen = counters.generators;
 
-        OpticalCounts oc{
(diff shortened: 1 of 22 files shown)


app/celer-sim/Transporter.cc
```diff
@@ -198,6 +200,29 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries) -> TransporterResult
...
+            = optical_->exchange_counters(stepper_->sp_state()->aux());
+        auto const& gen = counters.generators;
+
+        OpticalCounts oc{
+            /* tracks = */ gen.photons,
+            /* generators = */ gen.cherenkov + gen.scintillation,
+            /* steps = */ counters.steps,
+            /* step_iters = */ counters.step_iters,
+            /* flushes = */ counters.flushes,
+        };
```
amandalund: Is this the correct order? Maybe not the safest way to initialize the counts.

amandalund (approved):
Thanks @sethrj!

app/celer-sim/Transporter.cc
sethrj: Oh yeah... good point! Can't wait till we allow c++20 internally 😉

PR #1708, opened 2025-04-01
Fix Geant4 SD hit scoring

amandalund:
This fixes a bug from #1632 in which our SD callbacks were accidentally disabled.


sethrj (approved):
Yikes! Why didn't our tests catch this 😕

amandalund:
Hmm where would it be tested?

sethrj:
I guess our celer-g4 tests don't seem to use the SD callback aside from the ROOT IO which doesn't get checked. We could probably add a trivial check in one of the examples, but I intend sometime soon to revert https://github.com/celeritas-project/celeritas/pull/1632/commits/3b8485562439611bebde0af1003f7e3afd2e14cd so we can get that part of the interface tested.

sethrj:
@amandalund I added a stupid test, changing one of the examples to use electrons as a primary and then asserting if none of the threads saw any energy deposition. It failed before your commit and passes now, but please double check and hit merge if it looks good.

sethrj:
Oops no surprise but we get zero energy if "kill offload" is enabled 🤪

sethrj:
OK @amandalund now it works and catches the error 😅 and I think the abuse of global variables (multiple threads writing simultaneously?) should be safe but I'm not 100% sure. @pcanal is it legal+predictable for multiple threads to write the same thing to the same memory location simultaneously?

amandalund:
Thanks for adding the test! For the global variables, I think even though all threads are writing the same value it's still not guaranteed to be correct, even if it will usually work in practice...

sethrj:
Okeydoke, I will make it atomic? I was originally going to do atomic double but that's not part of the standard 😔.

pcanal:
> is it legal+predictable for multiple threads to write the same thing to the same memory location simultaneously?

Neither unless it is an std::atomic.  When reading a memory area that is being written by 2 thread concurrently, if it is not an atomic, the read value might one or the other or ... a mix of both.

example/accel/trackingmanager-offload.cc
```diff
@@ -68,6 +71,9 @@ class SensitiveDetector final : public G4VSensitiveDetector
 // Simple (not best practice) way of accessing SD
 G4ThreadLocal SensitiveDetector const* global_sd{nullptr};
 
+std::atomic<int> expected_nonzero_energy{0};
+std::atomic<int> actual_nonzero_energy{0};
```
pcanal, line 75: Consider:
```
std::atomic<size_t> expected_nonzero_energy{0};
std::atomic<size_t> actual_nonzero_energy{0};
```
or `unsigned long long`?

sethrj: It's just the number of events times number of threads. Does unsigned give it some advantage? I would think signed would be better because then the compiler doesn't have to worry about atomic unsigned int overflow.

Also my question about the global memory was simply about writing. I know simultaneous read and write is bad, but for the case I had earlier it wasn't read until much later and then on a single thread with no contention.

pcanal: It is still undefined behavior (eg. I guess it could technically write the first 2 bytes in one thread, all 4 bytes in another thread and the last 2 bytes in the first thread, resulting in a mixed value).

Furthermore, the code uses `++variable` which means read then write ....

sethrj: It does now. When I asked the question it was just assigning "true" to a bool.

PR #1709, opened 2025-04-01
Fix sampling of angle in Urban MSC model

amandalund:
Digging into the discrepancy in the energy deposition in the EM3 benchmark, I think I tracked it down to a bug in the sampling of the polar angle in the Urban MSC model introduced in #674 . Lower statistics (only 100k primaries), but here's the before and after:

calo1n-em-dev
calo1n-em-fix


sethrj:
Awesome!! Wonder if there's any provenance for that expression or a clearer way to write it...

whokion:
Great detective work!

I see two critical issues related to this bug:
- refactoring introduces bugs (potentially yes, but surprised here) 
- why is the original code gone (even no log or history)? - I see this in many other places when original codes were renamed or re-organized.  Very bad practice wiping out history of original works! 

The earlier code was:
```
        // The tail should not be too big
        xsi = max<real_type>(xsi, real_type(1.9));

        real_type c = xsi;
        if (std::fabs(xsi - 3) < real_type(0.001))
        {
            c = real_type(3.001);
        }
        else if (std::fabs(xsi - 2) < real_type(0.001))
        {
            c = real_type(2.001);
        }

        real_type ea = std::exp(-xsi);
        // Mean of cos\theta computed from the distribution g_1(cos\theta)
        real_type xmean1 = 1 - (1 - (1 + xsi) * ea) * x / (1 - ea);
```
which was ugly, but the last line is consistent with @amandalund's rollback.

sethrj:
This looks like my fault from #674 😞 

```diff
     // Evaluate parameters for the tail distribution
     real_type xsi = [this] {
         using PolyQuad = PolyEvaluator<real_type, 2>;
@@ -393,7 +379,11 @@ CELER_FUNCTION real_type UrbanMscScatter::sample_cos_theta(Engine& rng) const

     real_type ea = std::exp(-xsi);
     // Mean of cos\theta computed from the distribution g_1(cos\theta)
-    real_type xmean_1 = 1 - (1 - (1 + xsi) * ea) * x / (1 - ea);
+    // small theta => x = theta0^2
+    // large xsi => xmean_1 = 1 - x
+    // small tau => xmean = 1
+    real_type xmean_1 = 1 - x * (1 + (xsi * ea) / (1 - ea));
```

whokion (approved):

sethrj:
@whokion What do you mean by:
> why is the original code gone (even no log or history)? - I see this in many other places when original codes were renamed or re-organized. Very bad practice wiping out history of original works!

Using `git blame` took 2 seconds to find where in the history the original code was. If there had been an explanation of what the lines were doing, or an equation number, that would absolutely be kept in there. But there's not. I spent a long time poring over the literature before and after in this piece of code trying to figure out what the hell it's doing and what the constraints on it are and under what circumstances the different behaviors appear. When rearranging one black box into a (regrettably incorrectly simplified) other black box, there's no sense in keeping a copy of the original except in the git log.

whokion:
Okay, my laziness! I did only `git log UrbanMscScatter.hh ` which only showed history back to commit ac23e66d60fa32c39337663181e4a5ea74d9cc34
As you know, the Urban model is a parameterized model which has many unknown equations and parameterized numbers (and probably many assumptions without rigorous justifications), so your criticism is legitimated - on the other hand, it is a great piece of work/model which we freely borrow from Geant4 and we should give our respect (to Urban).

sethrj:
If we're copying code, we need to respect the license of the original. If we're implementing a method, we cite the reference. I realize there's a bit of a gray area in there, but I would strongly prefer that our work be a new reimplementation of the Geant4 methods (citing Urban's published methods, not copying expressions from his code). If we can't honestly say that, and if it's copy-pasting, then we need a separate directory with the copy-pasted code and a LICENSE file respecting the derivation of the original code.

@amandalund I did a little more refactoring from the original version you fixed. The updated expressions remove several needless temporaries and improve the ability to reason about the values:
https://github.com/sethrj/celeritas/commit/c9319c2bf19798c7320bd315309329ec60eef114

We really ought to break this function into smaller helpers based on the different parameterization/limits of the exiting angular distribution, and compare those distributions near the limits (I think we have four special cases: small end energy, small step, large tau, small angle).

amandalund:
I think the refactors here and elsewhere have significantly improved the code—in terms of clarity, complexity, flexibility, testability, documentation. I do agree it can be difficult to ensure no bugs are introduced (and for the physics models it can sometimes make debugging more challenging). For me, those types of PRs can be some of the hardest to review because they often involve large changesets, and even if most of the changes are purely structural it still requires careful checking to make sure no errors are introduced (for example, in #1632/#1708 I missed the omission of a single but significant line). That said, it's worth it to make the codebase more maintainable in the long term.

We also may have been a bit lax at times about pushing changes after a PR has been approved. It might be good practice to re-request a review or notify the reviewers if additional changes (not directly related to the review feedback) are pushed so things like this don't slip by (although I know we also often approve PRs when there are only minor changes to address/before all tests have passed).

@sethrj thanks, I like those changes! I’d propose we merge this fix (once I’ve updated al lthe tests) and then follow up with another PR with your improvements and potentially additional refactoring of the function/tests?

sethrj:
Sounds good @amandalund . Since the change in results are my fault I volunteer to do the drudgery of updating the regression values 😅

amandalund:
Thanks @sethrj 😆 I've got a start on that, I can update a few more and then call on you for help!

sethrj:
PS I wonder if this bug is the reason we have https://github.com/celeritas-project/celeritas/blob/caa7c8a36f2c73497ccf693a46ad4d907db53aff/src/celeritas/em/msc/detail/UrbanMscScatter.hh#L444-L445

And that is the sort of thing where it *is* important to have clean code for: is that a bug? a numerical roundoff issue? We can't tell. We've caught at least one physics bug that way.

amandalund:
@sethrj I ran a couple quick tests and it looks like `qprob` can still be larger than 1, but I didn't hit any cases where `prob` was... although it also doesn't look like it would be affected by this change.

test/celeritas/em/UrbanMsc.test.cc
```diff
@@ -744,37 +744,37 @@ TEST_F(UrbanMscTest, TEST_IF_CELERITAS_DOUBLE(msc_scattering))
...
+        0.010689973326661,
+        -0.22798069327496,
+        0.59468463817103,
         0,
-        -0.30604942827394,
-        0.40930643343077,
+        0.035407928579158,
+        0.36272875982918,
         0,
         0,
```
sethrj, line 761: FWIW, comparing to the results before [674's changes](https://github.com/celeritas-project/celeritas/pull/674/files#diff-0353155380a6858688ffc76cb1c19bab5d91b39fd072a6b3bb30f5a81d8316ad):
```
 i      expected_angle              angle         Difference
 2   -0.14560882721751   0.198550282263891   -2.36358685155328
 4   -0.32650640360665   0.162186795072201   -1.49673388723975
 5   0.013072020086723   0.475140065630482    35.3478683844032
 8  0.0031128176633161  0.0106899733266619     2.4341790888175
 9   0.055689200859297  -0.227980693274963   -5.09380435986097
10    0.17523769034715    0.59468463817103     2.3935886566009
12   -0.30604942826098  0.0354079285791585   -1.11569349689804
13    0.40930643323792   0.362728759829179  -0.113796582771193
```

so some of the regression values have reverted (yay!)

Full diff of your changes (-) to pre-674 results (+) which *definitely* had some bugs:
```diff
     static double const expected_tstep[] = {
         0.00279169,
         0.15497550035228,
-        0.02893427493639,
-        0.07816386744463,
-        0.0011122660575717,
-        0.00011596861898845,
-        0.0007421564286833,
-        3.1163288786189e-05,
+        0.0376414,
+        0.078163867310103,
+        0.0013704878213315,
+        9.659931080008e-05,
+        0.00074215629579836,
+        3.1163160035578e-05,
         0.00279169,
-        0.18632233731664,
-        0.031954244286546,
-        0.078778123988169,
-        0.0011915432720176,
-        0.00010300856437122,
-        0.00067599187481117,
-        2.6048657211831e-05,
+        0.19292164062171,
+        0.028493079924889,
+        0.078778123985786,
+        0.001181975090476,
+        0.00011040360872946,
+        0.00067599187250339,
+        2.6048655048005e-05,
     };
     static double const expected_gstep[] = {
         0.0027915002818486,
         0.14348626259532,
-        0.027673334586517,
-        0.019196479870158,
-        0.0010444821609262,
-        0.00010644611831056,
-        0.00013922620627564,
-        5.6145657548872e-06,
+        0.035504762192774,
+        0.019196479862044,
+        0.0012685611493895,
+        8.9929525199044e-05,
+        0.00013922620159908,
+        5.6145615756546e-06,
         0.0027915002818486,
-        0.17098608910208,
-        0.030562519433186,
-        0.021108585476009,
-        0.0011228204355752,
-        9.6292200670304e-05,
-        0.00015376538806457,
-        7.1509534497628e-06,
+        0.17644490740217,
+        0.027387302420092,
+        0.021108585475838,
+        0.001114330224665,
+        0.00010271284320013,
+        0.00015376538794516,
+        7.1509532866908e-06,
     };
     static double const expected_alpha[] = {
-        0,
-        1.7708716862051,
-        3.4676312226412,
-        12.793635124418,
-        0,
-        0,
-        1347.4248303342,
-        32089.039345654,
-        0,
-        1.709956627379,
-        3.3376672103125,
-        12.693879333179,
-        0,
-        0,
-        1479.3077213825,
-        38389.694788022,
-    };
-    static double const expected_msc_range_limit[] = {
-        3.5686548881735e-05,
-        4.0510166112733e-05,
-        4.0073894011965e-05,
         -1,
-        2.5270068437103e-05,
-        1.0695397351015e-05,
+        1.7708716862049,
+        3.4500251375335,
+        12.793635146437,
         -1,
         -1,
-        1.6703727150203e-05,
-        2.0782727059196e-05,
-        2.0774322573966e-05,
+        1347.4250715939,
+        32089.171921536,
         -1,
-        1.3419878391705e-05,
-        5.6685934033174e-06,
+        1.7061753085921,
+        3.3439279763905,
+        12.693879333563,
         -1,
         -1,
+        1479.3077264327,
+        38389.697977001,
     };
     static double const expected_angle[] = {
         0.00031474130607916,
-        0.79003683103899,
-        0.19855028226389,
+        0.79003683103898,
+        -0.14560882721751,
         0,
-        0.1621867950722,
-        0.47514006563048,
+        -0.32650640360665,
+        0.013072020086723,
         0,
         0,
-        0.010689973326661,
-        -0.22798069327496,
-        0.59468463817103,
+        0.0031128176633161,
+        0.055689200859297,
+        0.17523769034715,
         0,
-        0.035407928579158,
-        0.36272875982918,
+        -0.30604942826098,
+        0.40930643323792,
         0,
         0,
     };
     static double const expected_displace[] = {
-        8.1986203515053e-06,
+        8.198620357924e-06,
         9.7530617641316e-05,
-        4.8191245399216e-05,
+        -7.1670542039709e-05,
         0,
-        5.948204381332e-05,
-        3.9532158789002e-06,
+        -9.1137823713002e-05,
+        9.7878389032256e-06,
         0,
         0,
-        2.2406571358849e-05,
-        -2.7281539943695e-05,
-        7.6790677851376e-05,
+        5.31692113991705e-06,
+        7.9159745553753e-05,
+        9.8992726876372e-05,
         0,
-        -4.6740645216623e-06,
-        3.1463681655624e-06,
+        -9.0024133671603e-05,
+        2.9542258777685e-05,
         0,
         0,
     };
```

sethrj (approved):
Definitely looks fixed and some of the regression values have been restored. Well done!

amandalund:
Oh yay, that's good to see!

PR #1712, opened 2025-04-02
Update Geant4 version to 11.3 in CI build

amandalund:
Follow-on to #1687: this updates the Geant4 version in the CI build matrix includes from 11.0 to 11.3 and updates the test results.


sethrj (approved):
LGTM, thanks for the upgrade!

.github/workflows/build-docker.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             buildtype: "reldeb"
             image: "rocky-cuda"
     env:
-      CELER_TEST_STRICT: 1
+      CELER_TEST_STRICT: 0 # TODO: enable with Geant4 v11.3
```
amandalund, line 38: Is it ok to disable strict testing for the docker build, which uses 11.0.4?

sethrj: Yep! Since we're not actually running the CUDA.

sethrj (approved):

Issue #459, opened 2022-07-20
Refine MSC implementation


sethrj:
- [ ] Vecgeom crashes on GPU and produces different results from ORANGE (on host) in the stepper test
- [ ] `is_displaced`, `geo_limited`, `action_result` are [not orthogonal choices](https://github.com/celeritas-project/celeritas/pull/454#discussion_r924935600)
- [ ] range is recalculated a couple of different time; and there's a `min` which is necessary only if the range recalculation changes *or* running some kind of test [454](https://github.com/celeritas-project/celeritas/pull/454#discussion_r924929500) [429](https://github.com/celeritas-project/celeritas/pull/429#discussion_r877198327) [398](https://github.com/celeritas-project/celeritas/pull/398#discussion_r843065927)

sethrj:
See #615

sethrj:
Copilot+claude comparison of the three main Geant4 MSC models

# Comparison of ComputeTruePathLengthLimit Methods

| Step | G4GoudsmitSaundersonMscModel | G4WentzelVIModel | G4UrbanMscModel |
|------|------------------------------|------------------|-----------------|
| **Initialize variables** | Sets skin depth and various flags | Sets tlimit | Sets tPathLength to currentMinimalStep |
| **Get track information** | Gets dynamic particle and step point info | ✓ | ✓ |
| **Get material information** | Updates current couple and material index | ✓ | ✓ |
| **Get kinetic energy** | Gets kinetic energy | ✓ | Gets kinetic energy and log kinetic energy |
| **Get range** | Calculates current range from energy and material | ✓ | ✓ |
| **Calculate transport mean free path** | Gets first transport mfp λ₁, also calculates elastic mfp λ₀, screening parameter and G1 | Gets lambdaeff | Gets lambda0 |
| **Check step size vs range** | Limits true path length to currentRange | ✓ | ✓ |
| **Very small step check** | No early return for small steps | Returns early if step < tlimitminfix | Returns early if step < tlimitminfix |
| **Get safety** | Gets presafety from step point | ✓ | ✓ |
| **Safety vs range check** | No early return based on safety/range comparison | Returns early if range < presafety | Returns early if step=range < presafety |
| **Calculate physical limits** | Calculates Zeff and distance considering Z | Calculates rlimit based on range and lambda | Calculates distance with Z-dependent factor |
[...]

PR #1714, opened 2025-04-03
Simplify MSC angular sampling

sethrj:
This adds a power distribution for sampling from $`x^{p}`$ and uses it to clearly express the sampling functions described by Urban. The "simple sampling" function is now standalone and tested thoroughly as a function of tau.

To avoid changing results from AlongStep etc., I've added a `CachedRngEngine` class to enable the out-of-order sampling that the original UrbanMsc had.

See #459 .


src/celeritas/em/msc/detail/UrbanMscScatter.hh
```diff
@@ -397,15 +389,23 @@ CELER_FUNCTION real_type UrbanMscScatter::sample_cos_theta(Engine& rng) const
...
+    // large xsi => xmean_1 = 1
     // small tau => xmean = 1
     real_type x = ipow<2>(2 * std::sin(real_type(0.5) * theta0_));
-    real_type xmean_1 = 1 - x * (1 - (1 + xsi) * ea) / (1 - ea);
 
-    if (xmean_1 <= real_type(0.999) * xmean_)
+    // Calculate intermediate values for the mean of cos(theta)
+    // Since xsi is not near zero (thanks to max), no need to use expm1
+    // The expression in outer parens is in [~0.666, 1]
+    real_type xmean_1 = 1 - x * (1 - xsi / (std::exp(xsi) - 1));
```
sethrj, line 401: let `ex = exp(xsi) = 1/ea`
```
xmean_1
= 1 - x * (1 - (1 + xsi) * ea) / (1 - ea)  // multiply parens by ex/ex
= 1 - x * (ex - (1 + xsi)) / (ex - 1)
= 1 - x * (ex - 1 - xsi) / (ex - 1)
= 1 - x *( (ex - 1)/(ex - 1) - xsi / (ex - 1) )
= 1 - x *(1 - xsi / (ex - 1))
```

src/celeritas/em/msc/detail/UrbanMscScatter.hh
```diff
@@ -397,15 +389,23 @@ CELER_FUNCTION real_type UrbanMscScatter::sample_cos_theta(Engine& rng) const
         return max(result, real_type(1.9));
     }();
 
-    real_type ea = std::exp(-xsi);
     // Mean of cos\theta computed from the distribution g_1(cos\theta)
     // small theta => x = theta0^2
-    // large xsi => xmean_1 = 1 - x
+    // large xsi => xmean_1 = 1
```
sethrj: Oops need to revert this, it does go to 1 - x

sethrj:
@amandalund would you prefer i add the cached rng as a separate commit? @whokion could you review the updated docs?

amandalund:
@sethrj sorry I still have to work through this, but I think it's fine if you want to include it here.

sethrj:
No worries! This has been low on the priority list.

src/corecel/random/distribution/PowerDistribution.hh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct on the interval [a, b).
+ *
+ * It is allowable for the two bounds to be out of order.
+ *
+ * Note that writing as \code (1/a) * b \endcode allows the compiler to
+ * optimize better for the constexpr case a=1.
```
amandalund: Leftover copy-paste.

src/corecel/random/engine/CachedRngEngine.hh
```diff
@@ -0,0 +1,160 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return an RNG with enough space to return Count of type T.
+ */
+template<class T, size_type Count, class Engine>
+inline CELER_FUNCTION auto cache_rng_values(Engine& e)
+{
+    // Accout for the fact that some implementations uses 64-bit integers for
```
amandalund:
```suggestion
    // Account for the fact that some implementations use 64-bit integers for
```

src/celeritas/em/msc/detail/UrbanMscScatter.hh
amandalund: Just curious if it's better _not_ to use `expm1` if `xsi` is guaranteed not to be small?

src/celeritas/em/distribution/UrbanLargeAngleDistribution.hh
```diff
@@ -0,0 +1,142 @@
...
+ * \f$ \lambda_2 \approx 2.5 \lambda_1 \f$.
+ *
+ * Using these moments, Urban calculates: \f[
+   a = \frac{2\langle \mu \rangle + 9\langle \mu^2 \rangle - 3}
+            {2\langle \mu \rangle - 3\langle \mu^2 \rangle + 1}
+ * \f] and
+ * \f[
+   p_1 = \frac{(a + 2)\langle \mu \rangle}{a} \,.
+ * \f]
+ */
```
amandalund, line 80: Nice documentation!

amandalund (approved):
Sorry for the delay on this @sethrj, looks good to me! (though might be good if @whokion could also take a look)

src/celeritas/em/msc/detail/UrbanMscScatter.hh
sethrj: I actually have no idea how it's implemented... but I would assume there's more optimizations for `std::exp` than the special case.

PR #1715, opened 2025-04-03
Improve messages when multiple tracks fail in flight

sethrj:
This improves messages from the multi exception handler, and for the particle debug output, so that we only have to see a few messages (if all tracks fail) rather than a thousand. It also prints the full track state. The debug track output now also prints the material that the track is in.

A new test helper class `StringSimplifier` combines the functionality of removing pointers and ANSI codes with the new ability to round floating point values to reduce the number of slight variants in output between systems.

Here, 1024 threads are failing in the same way. A single failure (on the first thread) is thrown at the end; and before it's thrown, all the other threads' debug messages are consolidated into a single output with a message about the suppression.
Screenshot 2025-04-03 at 08 47 37

### DIFF
--- a/src/celeritas/global/DebugIO.json.cc
+++ b/src/celeritas/global/DebugIO.json.cc
@@ -11,8 +11,11 @@
 #include "corecel/math/QuantityIO.json.hh"
 #include "corecel/sys/ActionRegistry.hh"
 #include "celeritas/UnitTypes.hh"
+#include "celeritas/geo/GeoMaterialView.hh"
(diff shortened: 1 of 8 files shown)


sethrj:
@amandalund I ended up spending way too much time on the string simplifier 🤪 so perhaps I should make that a separate PR and make this a draft until it's completed?

amandalund:
Works for me!

amandalund (approved):
Looks good, thanks!

PR #1717, opened 2025-04-04
Add histogram helper class to simplify distribution tests

amandalund:
Related to #1713: this adds a simple histogram class to simplify our distribution tests. TODO: I think there are also some places in the physics tests this could be used.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -13,7 +13,6 @@ file(TO_CMAKE_PATH "${PROJECT_SOURCE_DIR}" CELERITAS_SOURCE_DIR)
 configure_file(celeritas_test_config.h.in celeritas_test_config.h @ONLY)
 
 set(_harness_sources
-  Histogram.cc
   Test.cc
   testdetail/JsonComparer.json.cc
   testdetail/NonMasterResultPrinter.cc
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -13,6 +13,7 @@ file(TO_CMAKE_PATH "${PROJECT_SOURCE_DIR}" CELERITAS_SOURCE_DIR)
 configure_file(celeritas_test_config.h.in celeritas_test_config.h @ONLY)
 
 set(_harness_sources
+  Histogram.cc
   Test.cc
   testdetail/JsonComparer.json.cc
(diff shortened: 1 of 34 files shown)


test/Histogram.hh
```diff
@@ -0,0 +1,49 @@
...
+
+namespace celeritas
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the histogram of a set of values.
+ */
+class Histogram
```
sethrj: Maybe this class should use `UniformGrid` as an implementation detail? (Then it could be easily templated to allow log/nonuniform grids?) Should the input be a vector by reference (resized during intialization, or with its size used instead of an input number of bins) so this just becomes an accumulator? Should there be a helper function that turns normalizes the result into fractions rather than counts? I guess there's a lot of ways to take this in terms of flexibility vs simplicity... so maybe best to leave as-is?

test/Histogram.hh
```diff
@@ -0,0 +1,49 @@
...
+
+  public:
+    // Construct with number of bins and range
+    Histogram(size_type num_bins, Real2 range);
+
+    // Update the histogram with a value
+    void operator()(real_type value);
+
+    // Get the histogram
+    VecCount const& get_counts() const { return counts_; }
```
sethrj:
```suggestion
    VecCount const& counts() const { return counts_; }
```

sethrj (approved):
Nice simplification.

test/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ file(TO_CMAKE_PATH "${PROJECT_SOURCE_DIR}" CELERITAS_SOURCE_DIR)
 configure_file(celeritas_test_config.h.in celeritas_test_config.h @ONLY)
 
 set(_harness_sources
+  Histogram.cc
```
sethrj: Also since we now have test/corecel/random, should this class live there since it's used for random testing?

test/Histogram.hh
amandalund: Yeah, I considered making it more like a numpy histogram (construct with a vector of bin edges, have a density option) but went for simple because that's all we needed currently. Maybe I'll add the density helper, and if we decide to make this class more flexible later on we can...

PR #1719, opened 2025-04-04
Add string simplifier helper class for testing

sethrj:
This rounds floats down to a few digits, removes trailing `f` from single precision floats, genericizes pointers, and removes ANSI escape sequences.

Screenshot 2025-04-04 at 16 15 09


amandalund (approved):
Nice, thanks @sethrj!

PR #1722, opened 2025-04-07
Estimate scintillation wavelength distribution parameters from tabulated values when only the latter is provided

amandalund:
When generating optical photons from scintillation, Geant4 uses a tabulated distribution for sampling the wavelength (specified by the vector property `SCINTILLATIONCOMPONENT*`) while Celeritas samples from a Gaussian distribution with the mean and standard deviation specified by the custom-defined properties `SCINTILLATIONLAMBDAMEAN*` and `SCINTILLATIONLAMBDASIGMA*`. This adds a little helper class to estimate the mean and sigma if only the tabulated values are provided.


src/corecel/math/PdfUtils.hh
```diff
@@ -27,6 +29,12 @@ struct PostRectangleSegmentIntegrator
     {
         return (hi[0] - lo[0]) * lo[1];
     }
+
+    template<class T>
+    T x_eval(Array<T, 2> lo, Array<T, 2>) const
```
sethrj: Some documentation would be good here... I think I see what you're getting at but does it make sense for anything other than square/trapezoid integration? Perhaps for the moment calculator it would be easier to just hardcode trapezoid for now, and going forward evaluate on an interpolating grid?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -215,11 +217,39 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
+                        &pv, name, {ImportUnits::len, ImportUnits::unitless}))
+                {
+                    auto const& pv_cref = pv;
+                    auto moments
+                        = MomentCalculator{TrapezoidSegmentIntegrator{}}(
+                            make_span(pv_cref.x), make_span(pv_cref.y));
+                    comp.lambda_mean = moments.mean;
+                    comp.lambda_sigma = std::sqrt(moments.variance);
+
+                    CELER_LOG(warning)
```
sethrj: Maybe this should be an info?

sethrj (approved):
LGTM, but see the note on PdfUtils.

src/corecel/math/PdfUtils.hh
amandalund: Sure, we can just hardcode it. A related question: do we plan to expand this to include other integration methods? We don't use rectangular integration anywhere currently; is there a case where it would make sense to?

sethrj: Touché... I think we do need higher order numerical integration for calculating cross sections but certainly not rectangular. I'm certainly guilty of overengineering...

PR #1723, opened 2025-04-09
Improve MSC test code coverage

sethrj:
Out of the 28 (!!!) different branches in the UrbanMscScatter class, only 13 (!!!!!) of the cases were tested in the UrbanMscScatter test. This improves the test coverage so that 19 of the tests are covered by adding new safety and physics steps. It also changes the compared quantities into something more intuitively recognizable instead of just distances:
- pstep: fraction of the MSC mean free path 
- range: fraction of the MSC mean free path 
- msc_range: fraction of the MSC mean free path 
- tstep_frac: fraction of physics step
- gstep_frac: fraction of true path length
- alpha_over: fraction of the MSC mean free path 
- angle: binned polar angle instead of just one lateral component
- displace: displacement distance as fraction of safety instead of one lateral component

Using 8ac59f736b399ad1f1353bce3f491fbc55d5ed8d:
```python
before = {231, 254, 341, 456, 484, 370, 372, 377, 257, 500, 204, 314, 559,}
after = {231, 254, 341, 456, 484, 370, 372, 377, 500, 274, 475, 257, 494, 337, 204, 314, 580, 453, 559,}
expected = {204, 212, 222, 228, 231, 254, 257, 274, 291, 297, 314, 323, 330, 337, 341, 370, 372, 377, 440, 448, 453, 456, 475, 484, 494, 500, 559, 580,}
```

So now we test the following conditions:
- MSC range cache wasn't initialized by UrbanMscStepLimit
- Large energy loss over the step or large angle distribution width
- xsi close to 2
- Selected g_3 in main sampling
- Small `var` in sampling of g_2 (??)
- Correction for very small true path

But we are still missing:
- `true_path_ < shared_.params.geom_limit` (only reachable if MFP < 0.005 cm)
- End-step energy is very low
- Very small mfp travelled
- Arbitrarily small angle change (no displacement)
- Arbitrarily small angle change (with displacement)
- Very small outgoing theta0 average
- Long mean free path, exiting direction isotropic
- <mu^2> mean is more than a little less than <mu> mean
- xsi close to 3

I think this highlights the need for adding code coverage testing to our physics (@pcanal are you familiar with what's in ROOT?).


test/celeritas/em/UrbanMsc.test.cc
```diff
@@ -563,270 +582,564 @@ TEST_F(UrbanMscTest, TEST_IF_CELERITAS_DOUBLE(msc_scattering))
...
+        Histogram bin_action(3, {-0.5, 2.5});
+
+        for (int i = 0; i < num_samples; ++i)
+        {
+            MscInteraction interaction = scatter(rng);
+
+            // Bin change in angle from +z
+            bin_angle(interaction.action != Action::unchanged
+                          ? interaction.direction[2]
+                          : 1);
```
amandalund:
```suggestion
            bin_angle(interaction.direction[2]);
```
should be the same?

amandalund (approved):
Thanks for the additional testing! Having coverage analysis would be helpful...

test/celeritas/em/UrbanMsc.test.cc
sethrj: I think we leave the angle as `{0,0,0}` rather than copying if the interaction is completely skipped.

pcanal:
> I think this highlights the need for adding code coverage testing to our physics (@pcanal are you familiar with what's in ROOT?).

@sethrj If you mean using some tools to discover the amount of code coverage that the test have, but did indeed have some experience with that.

test/celeritas/em/UrbanMsc.test.cc
amandalund: It looks like it is copied? https://github.com/celeritas-project/celeritas/blob/872d52838fcc105c20041b00a8e24d29509ff723/src/celeritas/em/msc/detail/UrbanMscScatter.hh#L292

It also looks like there are some all-zero histograms in the results, maybe because the histogram class ignores values that are exactly equal to the upper bound?

sethrj: Exactly. And I'll update as suggested.

sethrj:
Spun off from #1714.

PR #1724, opened 2025-04-09
Fix field map usage from accel

sethrj:
In #1632 I missed the construction of the new RZ and cylindrical field maps since they're not tested (I still need to do this...).

@bruno-OG This should fix the error you saw.


amandalund (approved):
Thanks @sethrj! Hopefully the last thing we missed :)

PR #1725, opened 2025-04-09
Add quick-start documentation for integrating

sethrj:
There was no simple consolidated explanation of adding Celeritas to a Geant4 app. This adds a small "diff" and explanation showing what needs to be added in the simple case, and a quick start in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,78 @@ If you need to set a default configuration
 $ spack install celeritas +cuda cuda_arch=80
 ```
 
-Then see the "Software library" section of the [installation
-documentation][install] for how to use Celeritas in your application or framework.
-
 [spack-start]: https://spack.readthedocs.io/en/latest/getting_started.html
 [install]: https://celeritas-project.github.io/celeritas/user/usage/installation.html
+
+# Integrating into a Geant4 app
+
+In the simplest case, integration requires a small change to your project's
+CMake file:
+```diff
++find_package(Celeritas 0.6 REQUIRED)
+ find_package(Geant4 REQUIRED)
+@@ -36,3 +37,4 @@ else()
+   add_executable(trackingmanager-offload trackingmanager-offload.cc)
+-  target_link_libraries(trackingmanager-offload
++  celeritas_target_link_libraries(trackingmanager-offload
++    Celeritas::accel
+     ${Geant4_LIBRARIES}
+```
+
+a few includes:
+```diff
+--- example/accel/trackingmanager-offload.cc
++++ example/accel/trackingmanager-offload.cc
+@@ -36,2 +36,10 @@
+
++// Celeritas
++#include <accel/AlongStepFactory.hh>
++#include <accel/SetupOptions.hh>
++#include <accel/TrackingManagerConstructor.hh>
++#include <accel/TrackingManagerIntegration.hh>
++
++using TMI = celeritas::TrackingManagerIntegration;
++
+ namespace
+```
+
+minor additions to the user actions:
+```diff
+--- example/accel/trackingmanager-offload.cc
++++ example/accel/trackingmanager-offload.cc
+@@ -134,2 +142,3 @@ class RunAction final : public G4UserRunAction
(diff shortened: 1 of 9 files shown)


doc/usage/integration.rst
```diff
@@ -8,9 +8,41 @@
...
+The changes for a simple application look like:
+
+.. literalinclude:: ../../example/accel/add-celer.diff
+   :language: diff
+
+More details of integration are described below.
+
+The most stable part of Celeritas is the high-level :ref:`api_g4_interface`.
+Other components of the API are stable and documented in the :code:`api`
+section.
```
stognini: Did you mean something like

```suggestion
The most stable part of Celeritas is the high-level :ref:`api_g4_interface`,
with the API components documented in the :code:`api` section.
```

README.md
```diff
@@ -66,18 +66,83 @@ If you need to set a default configuration
...
-Then see the "Downstream usage as a library" section of the [installation
-documentation][install] for how to use Celeritas in your application or framework.
-
 [spack-start]: https://spack.readthedocs.io/en/latest/getting_started.html
 [install]: https://celeritas-project.github.io/celeritas/user/usage/installation.html
 
+# Integrating into a Geant4 app
+
+In the simplest case, integration requires a small change to your project's
+CMake file:
```
pcanal:
```suggestion
In the simplest case, integration requires a small change to your project's
CMake file (to properly handling linking CUDA libraries):
```

README.md
```diff
@@ -66,18 +66,83 @@ If you need to set a default configuration
...
++#include <accel/SetupOptions.hh>
++#include <accel/TrackingManagerConstructor.hh>
++#include <accel/TrackingManagerIntegration.hh>
++
++using TMI = celeritas::TrackingManagerIntegration;
++
+ namespace
+```
+
+minor additions to the user actions:
```
pcanal: Consider extending the description with a very small explanation of 'why' those changes are needed and/or what they do.  In particular so the user can decide where inside their `BeginOfRunAction` the code belongs (beginning, end or ... middle) or even if they need to create one if they don't have any.

README.md
```diff
@@ -66,18 +66,83 @@ If you need to set a default configuration
...
+--- example/accel/trackingmanager-offload.cc
++++ example/accel/trackingmanager-offload.cc
+@@ -203,4 +235,8 @@ int main()
+
++    auto& tmi = TMI::Instance();
++
+     // Use FTFP_BERT, but use Celeritas tracking for e-/e+/g
+     auto* physics_list = new FTFP_BERT{/* verbosity = */ 0};
++    physics_list->RegisterPhysics(
++        new celeritas::TrackingManagerConstructor(&tmi));
```
pcanal, line 141: Side question: since ADePT says that this is the only part needed for ADePT what are the additional capabilities instrumenting the Run and User action brings to Celeritas?

doc/usage/integration.rst
sethrj: I'll try to clarify a bit more. The geant4 interface is very stable; other components are mostly stable...

README.md
sethrj: Run action allows us to wait until the physics, geometry, etc. are actually set up before use, and unlike adept we actually clean up after ourselves. The `ActionInitialization` lets us set up the logging before anything else but it may not be strictly necessary... maybe we should delete this in 0.7.

README.md
sethrj: I don't know if the readme needs to handle that level of complexity. Is referring to the docs enough?

README.md
sethrj: Maybe adding a "Celeritas.hh" would be good to reduce the includes too?

sethrj: I'll do this extra simplification as a follow-on PR.

stognini (approved):
Looks good to me. Thanks!

README.md
pcanal: Well the part where they need to decide whether to put the Celeritas code in beginning, middle or end of their own customization .... even it it is to say it doesn't matter :)

PR #1727, opened 2025-04-10
Fix failures when integral approach is disabled

amandalund:
When the integral approach is disabled, the pre-step cross section is used to sample the process, and because we don't yet have material-dependent model bounds a model could be selected even when the track's energy was below the model limit. I've added a basic stepper test as well to run with integral approach off since we weren't testing it previously.


sethrj (approved):
Thanks @amandalund ! Under what circumstances in real life do we not want integral xs rejection , by the way?

amandalund:
I think just a tradeoff between accuracy and performance; according to the PRM:
> The integral variant of step limitation is the default for the G4eIonisation, G4hIonisation, G4eBremsstrahlung and some other processes for charged particles. The integral variant of the energy loss sampling process is less dependent on values of the production cuts [eal09] and allows to have less step limitation, however it should be applied on a case-by-case basis because may require extra CPU.

PR #1729, opened 2025-04-10
Simplify Geant4 user application interface

sethrj:
This adds a new `<CeleritasG4.hh>` header that covers most use cases for Geant4 integration, and a more  memorable `Celeritas::G4` library name. It also removes the requirement to initialize logging in `BuildForMaster` and `Build`.


doc/usage/integration.rst
```diff
@@ -21,24 +21,24 @@ recommended for all applications that support Geant4 11.0 or higher.
...
 
    - Find the Celeritas package.
-   - Link ``Celeritas::accel`` .
+   - Link ``Celeritas::G4`` .
    - Use ``celeritas_target_link_libraries`` instead of
      ``target_link_libraries`` when both VecGeom and CUDA are enabled.
 
 2. Set up physics offloading in your "main" function.
 
+   - Include ``<CeleritasG4.hh>``
```
stognini, line 30: I still only took a quick look at the PR, but this seems to focus exclusively on the `TrackingManagerIntegration`, from the previous PR discussion. What happens if we want a single include for the `UserActionIntegration`? Seems like we could wrap each case with its own header

src/CeleritasG4.hh
```diff
@@ -0,0 +1,22 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file G4Celeritas.hh
```
stognini:
```suggestion
//! \file CeleritasG4.hh
```

doc/usage/integration.rst
sethrj: The header supports all the different high level integration methods. The focus in the main integration summary and README is exclusively the tracking manager since that's the "most supported" by geant4.

doc/usage/integration.rst
```diff
@@ -21,28 +21,28 @@ recommended for all applications that support Geant4 11.0 or higher.
 1. Find and link in your CMake project:
```
stognini, line 21: Not this line, but for the Usage overview right above it (that github doesn't let me comment on):

Since the whole section focuses on the tracking manager integration, it may be useful to point to the documentation to the user action case. Something like

> for all applications that support Geant4 11.0 or higher. For older versions, refer to the user action integration documentation and example in `offload-template`

sethrj: The offload template uses the tracking manager, but I'll add a reference to the manual mentioning the other support.

stognini: Oh right, we also moved to the TMI during its PR.

src/corecel/random/distribution/detail/GenerateCanonical32.hh
```diff
@@ -17,38 +17,29 @@ namespace detail
...
-    //! \name Type aliases
     using result_type = float;
-    //!@}
 
-  public:
-    //! Sample a random number with floating point precision
+    // Sample a random number with floating point precision
     template<class Generator>
-    inline CELER_FUNCTION result_type operator()(Generator& rng);
+    inline CELER_FUNCTION float operator()(Generator& rng);
```
amandalund: What's the reason for this change?

amandalund (approved):
I tested with BenchmarkEM3 and all looks good to me.

src/corecel/random/distribution/detail/GenerateCanonical32.hh
sethrj: Trying to fix warnings in the doxygen documentation (I can revert those)

sethrj:
@pcanal It looks like the `Celeritas::G4` alias is not working 😞

amandalund:
Ok yeah sorry I must not have done a clean build after updating the library name because it's definitely _not_ working for me 😅

example/offload-template/CMakeLists.txt
```diff
@@ -64,6 +64,6 @@ target_include_directories(run-offload PRIVATE
 # Use special linking macros to support CUDA+VecGeom when used
 celeritas_target_link_libraries(run-offload
   ${Geant4_LIBRARIES}
-  Celeritas::celeritas
-  Celeritas::accel
+  Celeritas::celeritas # for user action
+  Celeritas::G4
```
pcanal, line 68:
```suggestion
  Celeritas::celeritas # for user action
  Celeritas::accel
  Celeritas::G4
```
here and in geant4.

sethrj: Nope , G4 should alias accel

sethrj: OK fixed it, thanks for the tip about aliasing in the config file!

PR #1730, opened 2025-04-10
Fix pedantic warning in tests

elliottbiondo:
Semicolons after namespaces causes a build error when compiling with `-Werror=pedantic`.


sethrj (approved):
Oops!

PR #1734, opened 2025-04-10
Remove per-process switch for integral cross section method

amandalund:
This removes the option to disable the integral approach on a per-process basis. We're unlikely to need this kind of granularity, and the functionality wasn't exposed to users anyway.


sethrj (approved):
Nice simplification!

PR #1735, opened 2025-04-11
Replace `ImportPhysicsVector` with `inp::Grid`

amandalund:
This proposes replacing `ImportPhysicsVector` with an input `Grid` struct, which moves in the direction of a more Celeritas-like representation of the imported data and should reduce some of the propagation of the spline options in #1633.


sethrj (approved):
Good next step!

Issue #1355, opened 2024-08-05
Optimize field propagation performance


sethrj:
- [ ] Refactor `FieldPropagator` to have internal state and function calls rather than being one huge function with tons of `if` statements
- [x] Rename classes (https://github.com/celeritas-project/celeritas/issues/721)
- [ ] Rethink field intersection algorithm as a root finder (maybe using Regula Falsi iteration)
- [ ] Add a derived field propagator that uses safety: https://github.com/celeritas-project/celeritas/issues/762
- [ ] Perform propagation in local coordinate system  for improved accuracy https://github.com/celeritas-project/celeritas/issues/994#issuecomment-1836508438
  * Perform integration in local coordinate system?
  * Then we could use single precision
- [ ] Make substep length in field driver (due to change in discretization error) persistent across calls in propagator (or even across steps)

---

- Input: step length
- Super small: accept immediately
- Goal: longest step allowed by chord length *and* acceptable discretization error
- If discretization limited, accumulate until step length is reached
- Use distance between endpoints versus dchord to tell if we're within the asymptotic region

Start with last successful step if available
- Converged if chord length is within tolerance, discretization is in limits
- Calculate target next step length for chord length ( * sqrt(eps /
  distance_chord)
- Calculate target next step length for discretization error
- Increase next guess if error is too large??

---

- MicroFieldAdvancer
  - No iteration
  - Success if step < minimum_step (evaluated first of course)
- ChordAdvancer: reduce step until chord is within a tolerance
  - Iteration: reduce step (exponential, no more than halving each iter)
  - Convergence: max_nsteps
  - Convergence: chord distance
  - Success if discretization error is small enough
- Reduce step size based on scale factor
- AccurateAdvancer:
  - sub-sub steps to accumulate to maximum step, but sub-step length changes
  - Success when accumulated length is greater than or equal to requested step
    length * (1 - epsilon_step)
  - StepIntegrator
    - MicroFieldAdvancer if step < minimum_step and update new step
    - GoodAdvancer otherwise:
      - Iteration: reduce step

sethrj:
Not quite the ideas that we discussed with @LSchwiebert today, but at least the same part of the code with the same goal.

PR #1737, opened 2025-04-14
Rename field integration interfaces: Stepper→Integrators, Driver→Substepper

sethrj:
This performs the following renames (see https://github.com/celeritas-project/celeritas/issues/721, https://github.com/celeritas-project/celeritas/issues/1355).

Summary:
	•	“Stepper” → “Integrator”: Renamed all stepper classes to “Integrator” for clarity and consistency with their purpose (e.g., DormandPrinceStepper → DormandPrinceIntegrator).
	•	“Driver” → “Substepper”: Renamed FieldDriver and related tokens to FieldSubstepper to better reflect its role as a substep interface, not a high-level driver.
	•	Result types clarified: FieldStepperResult → FieldIntegration, DriverResult → Substep, and member step → length, improving semantic accuracy of integration result types.
	•	Factory and file renames: Supporting functions and filenames updated to match the above changes.

Old Symbol / Token | New Symbol / Token | Context / Notes
-- | -- | --
FieldDriver | FieldSubstepper | Class rename, represents a field substep integration interface
FieldStepperResult | FieldIntegration | Struct rename, result of a full integrator step
DriverResult | Substep | Struct rename, result of a single substep in the propagator
step (in DriverResult) | length (in Substep) | Member rename for clarity: curved path length of a substep
RungeKuttaStepper | RungeKuttaIntegrator | Class rename for consistency with other integrator names
DormandPrinceStepper | DormandPrinceIntegrator | Class rename for consistency and clarity
ZHelixStepper | ZHelixIntegrator | Class rename for uniform terminology
apply_step_ | integrate_ | Member variable rename in FieldSubstepper
stepper (parameter) | integrate | Variable rename in construction functions and factories
make_mag_field_stepper | make_mag_field_integrator | Factory function rename
DriverT | SubstepperT | Template parameter rename in FieldPropagator
FieldDriver.hh | FieldSubstepper.hh | File rename
RungeKuttaStepper.hh | RungeKuttaIntegrator.hh | File rename
DormandPrinceStepper.hh | DormandPrinceIntegrator.hh | File rename
ZHelixStepper.hh | ZHelixIntegrator.hh | File rename


pcanal:
Can you add an additional column to the table with the equivalent Geant4 concept (if any)?  (It may or may not be that this is identical to the first column in which case updating the heading would be clear enough).

sethrj:
Hmmm not sure, I think most are just functions

pcanal:
What I mean, for example, is that this PR change from "FieldDriver" to "FieldSubstepper", if `Geant4` is calling the equivalent (independently of the technical implementation) concept "field driver" then making the connection between that concept and "FieldSubstepper" will be somewhat harder and we can make it easier to remember by documenting it. (And maybe it is already documented somewhere else ;) ).

src/celeritas/field/FieldSubstepper.hh
```diff
@@ -56,24 +56,27 @@ namespace celeritas
...
-template<class StepperT>
-class FieldDriver
+template<class IntegratorT>
+class FieldSubstepper
 {
   public:
-    // Construct with options data and the stepper
-    inline CELER_FUNCTION
-    FieldDriver(FieldDriverOptions const& options, StepperT&& perform_step);
+    // Construct with options data and the integrate
```
amandalund:
```suggestion
    // Construct with options data and the integrator
```

src/celeritas/field/RungeKuttaIntegrator.hh
```diff
@@ -119,13 +116,13 @@ RungeKuttaStepper<E>::operator()(real_type step,
 
 //---------------------------------------------------------------------------//
 /*!
- * The classical RungeKuttaStepper stepper (the 4th order).
+ * The classical RungeKuttaIntegrator integrate (the 4th order).
```
amandalund:
```suggestion
 * The classical RungeKuttaIntegrator integrator (the 4th order).
```

test/celeritas/field/DiagnosticIntegrator.hh
```diff
@@ -17,24 +17,25 @@ namespace test
 {
 //---------------------------------------------------------------------------//
 /*!
- * Count the number of invocations to the field stepper.
+ * Count the number of invocations to the field integrate.
```
amandalund:
```suggestion
 * Count the number of invocations to the field integrator.
```

test/celeritas/field/DiagnosticIntegrator.hh
```diff
@@ -17,24 +17,25 @@ namespace test
...
     //!@{
     //! \name Type aliases
-    using result_type = typename StepperT::result_type;
+    using result_type = typename IntegratorT::result_type;
     using size_type = std::size_t;
     //!@}
 
   public:
-    //! Forward construction arguments to the original stepper
+    //! Forward construction arguments to the original integrate
```
amandalund:
```suggestion
    //! Forward construction arguments to the original integrator
```

amandalund (approved):
Thanks @sethrj!

I think in some cases, but probably not all, we note the Geant4 classes these are based on... agreed that it would be helpful to document that where applicable.

sethrj:
> What I mean, for example, is that this PR change from "FieldDriver" to "FieldSubstepper", if `Geant4` is calling the equivalent (independently of the technical implementation) concept "field driver" then making the connection between that concept and "FieldSubstepper" will be somewhat harder and we can make it easier to remember by documenting it. (And maybe it is already documented somewhere else ;) ).

Yeah that's what I thought you meant. The corresponding Geant4 code is indeed documented in the classes but doesn't directly correspond to the Substepper (close to `G4MagInt_Driver` mixed with other bits). The DP code is `G4MagIntegratorStepper`. I've added some docs to flesh it out a bit more, and made sure those appear in the generated html.

sethrj:
Actually @amandalund @pcanal I'd like to keep this one a strictly rename PR so I can have it ignored in the git-blame-revs. I've removed the documentation and will cherry-pick to the release-0.6 branch.

PR #1738, opened 2025-04-14
Add convex hull algorithm

elliottbiondo:
This MR adds a 2D convex hull finding algorithm. While this is implemented as a general-purpose capability, the immediate use case is the construction of complex GEANT4 shapes in ORANGE, namely generic polycones and extruded solids. The capabilities in the MR will allow a complex shapes to be constructed recursively, by first constructing convex hulls and next subtracting internal concave regions.

As a general strategy, the convex hull approach is definitely superior to Delaunay triangulation, which requires many more regions. Convex decomposition algorithms can do better in some cases, but for both convex hull and convex decomposition approaches it is easy to come up with pathological cases.

My original design for this code was based on `Spans`. However, points that comprise concave regions are not contiguous in memory in the case where a concave region occurs between the last element and first element in the supplied vector of points.


elliottbiondo:
@sethrj RTR

elliottbiondo:
On thing that is missing is for robustness is a test that the points form a non-self-intersecting polygon, which can be done using the [Bentley-Ottmann Algorithm](https://www.webcitation.org/6ahkPQIsN).

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+
+namespace celeritas
+{
+namespace
+{
+//---------------------------------------------------------------------------//
+/*!
+ * SFINAE struct for determing if a templated type is a valid Real2.
+ */
+template<typename T>
```
elliottbiondo: @sethrj this works everywhere but Windows. There are at least 4 reasonable ways of representing a Real2, so I want to keep this templated, but since is a non-POD type it would be nice to have a check like this.

sethrj: #Since the only R2 we care about (I think??) is `celeritas::Array<T, 2>` why not template on just T?

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+    /// TYPES ///
+    using ConvexMask = std::vector<bool>;
+    using index_type = size_type;
+    using Index3 = std::array<index_type, 3>;
+
+    /// DATA ///
+    Points const& points_;
+    ConvexMask convex_mask_;
+    index_type start_index_;
+    SoftZero<> soft_zero_;
```
sethrj: This should be templated on the class's `real_type`.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+    Points convex_hull() const;
+
+    // Calculate the concave regions, each supplied in clockwise order
+    ConcaveRegions calc_concave_regions() const;
+
+  private:
+    /// TYPES ///
+    using ConvexMask = std::vector<bool>;
+    using index_type = size_type;
+    using Index3 = std::array<index_type, 3>;
```
sethrj: Why bring in std array instead of ours?

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+auto ConvexHullFinder<Real2>::is_clockwise(
+    ConvexHullFinder<Real2>::Index3 const& indices) const -> bool
+{
+    auto const& a = points_[indices[0]];
+    auto const& b = points_[indices[1]];
+    auto const& c = points_[indices[2]];
+
+    auto cross_product = (b[0] - a[0]) * (c[1] - a[1])
+                         - (b[1] - a[1]) * (c[0] - a[0]);
+    return cross_product <= 0 || soft_zero(cross_product);
```
sethrj: We have functions like this (used for validating shapes) in `orange/orangeinp/detail/PolygonUtils.hh`. Can you combine them? See `calc_orientation`, `has_orientation`.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate a mask that indicates which points are on the convex hull.
+ *
+ * This method uses the Gram Scan algorithm.
+ */
+template<class Real2>
+void ConvexHullFinder<Real2>::calc_convex_mask()
+{
+    auto n = points_.size();
```
sethrj: You should be able to put this in the for loop initializer I think.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: That's fine by me. I wanted to support std::vector as well but there isn't that good a reason for it.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,303 @@
...
+ * Contruct with vector of ordered points.
+ *
+ * This function generates a mask that is used to calculate the convex hull
+ * and associated concave regions. Note that this function does not enforce
+ * ordering.
+ */
+template<class Real2>
+ConvexHullFinder<Real2>::ConvexHullFinder(ConvexHullFinder::Points const& points)
+    : points_{points}
+{
```
elliottbiondo: Probably should add a TODO for input validation (heavy lift: requires checking for both counterclockwise order and also non-self-intersection).

src/corecel/math/ConvexHullFinder.hh
sethrj: Yeah let's require array. I don't see much benefit for vectors especially because of all the small allocations. The only reason I could imagine supporting other arrays is if we were a very generic library like boost.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+  public:
+    //!@{
+    //! \name Type aliases
+    using Point = celeritas::Array<T, 2>;
+    using Points = std::vector<Point>;
+    using ConcaveRegions = std::vector<Points>;
+    //!@}
+
+  public:
+    // Contruct with vector of ordered points
```
mrguilima:
```suggestion
    // Construct with vector of ordered points
```

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+    // Determine the next index, with modular indexing
+    size_type calc_next(size_type i) const;
+
+    // Determine the previous index, with modular indexing
+    size_type calc_previous(size_type i) const;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Contruct with vector of ordered points.
```
mrguilima:
```suggestion
 * Construct with vector of ordered points.
```

mrguilima (changes requested):
Just a couple typos on comments.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+    auto n = points_.size();
+
+    // Find the indices of the points on the convex hull. Start from the point
+    // with the lowest y value, which is gaurenteed to be on the hull.
+    std::vector<size_type> hull;
+    auto i = start_index_;
+    hull.push_back(i);
+    i = this->calc_next(i);
+
+    for ([[maybe_unused]] auto _ : range(n - 1))
```
sethrj: I've taken to writing these as old-style `for` loops to avoid the maybe_unused (though I think in C++23, because `_` is a special symbol, it won't be necessary):
```suggestion
    for (size_type i = 0, n = points_.size(); i != n; ++i)
```

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Determine the previous index using modular indexing.
+ */
+template<class T>
+size_type ConvexHullFinder<T>::calc_previous(size_type i) const
+{
+    return (points_.size() + i - 1) % points_.size();
```
sethrj: You should add a note that `i - 1` fails due to unsigned arithmetic :)

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+    start_index_ = this->min_element_idx();
+    convex_mask_ = this->calc_convex_mask();
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return the convex hull.
+ */
+template<class T>
+auto ConvexHullFinder<T>::convex_hull() const -> Points
```
sethrj: Since this is doing something expensive and not returning a reference, use `make_convex_hull` or equivalent. I've seen bugs resulting from surprising return-object-by-value:
```c++
std::copy(finder.convex_hull().begin(), finder.convex_hull().end(), dst);
```

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+ */
+template<class T>
+class ConvexHullFinder
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Point = celeritas::Array<T, 2>;
+    using Points = std::vector<Point>;
+    using ConcaveRegions = std::vector<Points>;
```
sethrj: For consistency and readability, use Real2 for the points; and we generally avoid making a vector of objects by pluralizing the object; and I think since ConcaveRegions isn't an actual class it's clearer how to use it by naming the type based on what it is not what it's intended to store.
```suggestion
    using real_type = T;
    using Real2 = celeritas::Array<real_type, 2>;
    using VecReal2 = std::vector<Real2>;
    using VecVecReal2 = std::vector<VecReal2>;
```

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
...
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Find the convex hull of a sequence of 2D points.
+ *
+ * These points must be supplied in clockwise-order such that segments between
+ * adjacent points, including the last and first points, comprise a
+ * non-self-intersecting polygon. Exploiting this ordering, the Gram Scan
+ * algorithm finds the convex hull with O(N) time complexity.
```
sethrj: Is there a citation to add for this, e.g. Introduction to Algorithms? If so, please add to Zotero and cite here using the updated key.

src/corecel/math/ConvexHullFinder.hh
```diff
@@ -0,0 +1,277 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file corecel/math/ConvexHullFinder.hh
```
sethrj: Since this is geometry specific, and only intended to be used in ORANGE, I think it should live in `orangeinp`.

test/corecel/math/ConvexHullFinder.test.cc
```diff
@@ -0,0 +1,237 @@
...
+        for (auto i : range(expected.size()))
+        {
+            EXPECT_EQ(expected[i][0], actual[i][0]);
+            EXPECT_EQ(expected[i][1], actual[i][1]);
+        }
+    }
+
+    void compare_concave_regions(ConcaveRegions expected, ConcaveRegions actual)
+    {
+        EXPECT_EQ(expected.size(), actual.size());
```
sethrj: Must be `ASSERT_EQ`, otherwise the test will segfault. Same with the size check above.

test/corecel/math/ConvexHullFinder.test.cc
```diff
@@ -0,0 +1,237 @@
...
+ * Test harness for ConvexHullFinder
+ */
+class ConvexHullFinderTest : public ::celeritas::test::Test
+{
+  public:
+    using CHF = ConvexHullFinder<double>;
+    using Points = CHF::Points;
+    using ConcaveRegions = CHF::ConcaveRegions;
+
+    void compare_convex_hulls(Points expected, Points actual)
```
sethrj:
```suggestion
    void compare_convex_hulls(Points const& expected, Points const& actual)
```
and also const references below.

It might be easier for both cases to just provide a `linearize` function that turns the reference and actual into vectors that you can then EXPECT_VEC_EQ.

Also rename this function (and the next) `expect_eq` since that's a better description of what they're doing.

test/corecel/math/ConvexHullFinder.test.cc
```diff
@@ -0,0 +1,237 @@
...
+ *   1                 |\
+ *   \\                | \
+ *    \ \      3  5    |  \
+ *     \  \    /\/\    |   \
+ *      \   \/   4  \  |    \ 8
+ *       \   2        \|    /
+ *        \            6   /
+ *         \ 11           /
+ *          \/\__________/
+ *          0  10        9
```
sethrj: Great ASCII art inclusions!

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: I believe L-to-R order is respected for arithmetic operations operations with the same precedence, so this is equivalent to:

```
(points_.size() + i) - 1
```

where `(points_.size() + i)` is guaranteed to be at least 3.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: Yes it wasn't quite expensive enough for a `calc_`, so I dropped that without considering `make_`. Will fix.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: (removed this type independently of this comment)

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: It's used in both for loops. I don't think it makes any different, but I will put it in each loop initializer.

src/corecel/math/ConvexHullFinder.hh
sethrj: Right, but I mean add a note so that no one unwittingly says "oh, with the modulo we can rewrite this as `(i - 1) % points.size()`" and breaks it.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: Is it worth propagating the template to `calc_orientation`?

Also, the way the Graham Scan works, collinear points need to be kept, which requires soft equality.

I could do something link this:

```
template<class T>
inline Orientation
calc_orientation(Real2<T> const& a, Real2<T> const& b, Real2<T> const& c, SoftZero<T> soft_zero = SoftZero<T>())
```

which also introduces a `Real2<T>`, propagating the template even further...

elliottbiondo: The default value would also introduce soft equality in places where there wasn't soft equality before, which I could imagine having unintended consequences.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: I added it to Zotero. Remind me: is the key from exporting the citation via Bibtex?

test/corecel/math/ConvexHullFinder.test.cc
elliottbiondo: > It might be easier for both cases to just provide a linearize function that turns the reference and actual into vectors that you can then EXPECT_VEC_EQ.

While this would be fine for the VecReal2, it could cause problems for VecVecReal2 as:

```
{{point_a, point_b, point_c}, {point_e, point_f, point_g, point_h}}
```

and

```
{{point_a, point_b, point_c, point_e}, {point_f, point_g, point_h}}
```

would return equal. It's not impossible that a bug introduced to the Graham Scan could result in something like this.

elliottbiondo:
@sethrj @mrguilima this is ready for another look

test/corecel/math/ConvexHullFinder.test.cc
elliottbiondo: Fixed name and changed to const ref (the latter was of course an oversight)

src/corecel/math/ConvexHullFinder.hh
sethrj: I think it might be clearer to have a "remove+erase" like functionality outside of the hull finder to delete collinear (or nearly) since that may also need to take the user input geometry tolerance into account. I would do that at the shape level and then the hull finder would assert non-collinear. Thoughts?

elliottbiondo: Yes, originally I was going to either disallow or remove collinear points. However, due to a corner case, the algorithm itself needs to support collinearity. Consider the geometry:

```
 *                     7
 *   1                 |\
 *   \\                | \
 *    \ \      3  5    |  \
 *     \  \    /\/\    |   \
 *      \   \/   4  \  |    \ 8
 *       \   2        \|    /
 *        \            6   /
 *         \ 11           /
 *          \/\__________/
 *          0  10        9
 ```
 When creating the convex hull while traversing clockwise, points 9, 10, and 11, are added to the stack. Once 0 is encountered, 11 is popped of the stack. If you don't allow collinearity, 10 will also be popped off the stack. That is actually okay if you only want is a convex hull. However, if you want all the concave regions (which we do) we need to keep point 10 in order to create the 0-11-10 convex region.

They may be more elegant ways of handling this case, but even *detecting* this case requires checking for collinearity.

I think a reasonable option would be:
1) template calc_orientation on T, just because we always want the option of using float,
2) write a free-standing "is_soft_collinear" method in `orange/orangeinp/detail/PolygonUtils.hh`.

Comment (sethrj) [reply src/corecel/math/ConvexHullFinder.hh]: Is 11 there an "epsilon" sized chunk? I don't know if we *do* want all the concave regions in that case? 🤔

Comment (elliottbiondo) [reply src/corecel/math/ConvexHullFinder.hh]: Nope. It's the same issue here, drawn to scale:

```
 *                     7
 *   1                 | \
 *   \\                |  \
 *    \ \      3  5    |   \
 *     \  \    /\/\    |    \
 *      \   \/   4  \  |     \ 8
 *       \   2        \|     /
 *        \     / \     6   /
 *         \   /11 \       /
 *          \/      \____/
 *          0      10   9
 ```

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,282 @@
...
+/*!
+ * Calculate a mask that indicates which points are on the convex hull.
+ *
+ * This method uses the Graham Scan algorithm.
+ */
+template<class T>
+auto ConvexHullFinder<T>::calc_convex_mask() const -> ConvexMask
+{
+    // Find the indices of the points on the convex hull. Start from the point
+    // with the lowest y value, which is gaurenteed to be on the hull.
```
mrguilima:
```suggestion
    // with the lowest y value, which is guaranteed to be on the hull.
```

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,282 @@
...
+
+  private:
+    /// TYPES ///
+    using ConvexMask = std::vector<bool>;
+
+    /// DATA ///
+    VecReal2 const& points_;
+    ConvexMask convex_mask_;
+    size_type start_index_;
+    SoftZero<real_type> soft_zero_;
```
mrguilima: I don't see where this is used, is it really necessary?
later on there is a `soft_zero()` call, but it is not actually using this.

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,282 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Determine the previous index using modular indexing.
+ */
+template<class T>
+size_type ConvexHullFinder<T>::calc_previous(size_type i) const
+{
+    // add points_.size() before subtracting 1 to avoid a negative unsigned int
+    return (i + points_.size() - 1) % points_.size();
```
mrguilima: This looks simpler, if not also faster:
```suggestion
    return (i > 0 ? i : points_.size()) - 1;
```

test/orange/orangeinp/ConvexHullFinder.test.cc
```diff
@@ -0,0 +1,235 @@
...
+        ASSERT_EQ(expected.size(), actual.size());
+        for (auto i : range(expected.size()))
+        {
+            EXPECT_EQ(expected[i][0], actual[i][0]);
+            EXPECT_EQ(expected[i][1], actual[i][1]);
+        }
+    }
+
+    void expect_eq(VecVecReal2 const& expected, VecVecReal2 const& actual)
+    {
```
mrguilima: Not obvious to me... why these two functions should have different names (trailing s in 'eqs')?

mrguilima (dismissed):

src/orange/orangeinp/ConvexHullFinder.hh
sethrj: Good point, the same could be done for the other side too `(i + 1 == points_.size() ? 0 : i + 1)`, since there's only a max offset of 1...

src/corecel/math/ConvexHullFinder.hh
sethrj: I see. How about in polygon utils you define a helper like SoftZero but with the same signature as calc_orientation:
```c++
/*!
 * Functor for calculating orientation with a tolerance for collinearity.
 */
template<class T>
class SoftOrientation
{
  public:
    using real_type = T;
    using Real2 = Array<real_type, 2>;

    // Construct with default tolerance
    CELER_CONSTEXPR_FUNCTION SoftOrientation()
        : soft_zero_{}
    {
    }

    // Construct with specified absolute tolerance
    explicit CELER_FUNCTION SoftOrientation(real_type abs_tol)
        : soft_zero_{abs_tol}
    {
    }

    // Calculate orientation with tolerance for collinearity
    CELER_FUNCTION Orientation
    operator()(Real2 const& a, Real2 const& b, Real2 const& c) const
    {
        auto crossp = (b[0] - a[0]) * (c[1] - b[1]) - (b[1] - a[1]) * (c[0] - b[0]);
        return soft_zero_(crossp) ? Orientation::collinear
               : crossp < 0         ? Orientation::clockwise
                                    : Orientation::counterclockwise;
    }

  private:
    SoftZero<real_type> soft_zero_;
};
```

And perhaps for completeness it should construct the tolerance using the bump calculator with the user-supplied rel+abs tolerance?

src/corecel/math/ConvexHullFinder.hh
sethrj: Do you have the "better bibtex" extension? It'll provide a default, otherwise the key will be in the "extra" field. Cite it as `graham-convexhull-1972`.

src/orange/orangeinp/ConvexHullFinder.hh
sethrj: yep that's a bug, it should've been using the local soft zero rather than the global function.

test/orange/orangeinp/ConvexHullFinder.test.cc
sethrj: I think another typo, @elliottbiondo please call `expect_eq`.

elliottbiondo: Find-and-replace error

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -263,7 +263,7 @@ auto ConvexHullFinder<T>::is_clockwise(size_type i_prev,
 template<class T>
 size_type ConvexHullFinder<T>::calc_next(size_type i) const
 {
-    return (i + 1) % points_.size();
+    return (i < points_.size() - 1 ? i + 1 : 0);
```
sethrj: Even though we can guarantee `points_` isn't empty it's a good practice to add one to the LHS rather than risk underflowing an unsigned integer
```suggestion
    return (i + 1 < points_.size() ? i + 1 : 0);
```

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: I just installed Better BibTeX. I will use `graham-convexhull-1972` for this one, but it looks like the keys generated by Better BibTeX are in a different format: e.g., `heckbertGraphicsGemsIV1994`.

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,281 @@
...
+            VecReal2 concave_region;
+            concave_region.push_back(points_[calc_next(i)]);
+            do
+            {
+                concave_region.push_back(points_[i]);
+                i = this->calc_previous(i);
+            } while (!convex_mask_[i]);
+
+            concave_region.push_back(points_[i]);
+            concave_regions.push_back(concave_region);
```
sethrj:
```suggestion
            concave_regions.push_back(std::move(concave_region));
```

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,281 @@
...
+
+        i = i_next;
+    }
+
+    // Convert convex hull indices to a mask
+    ConvexMask convex_mask(points_.size(), false);
+    for (auto h : hull)
+    {
+        convex_mask[h] = true;
+    }
```
sethrj: Why discard the hull in favor of the mask, when you have a separate function that later reconstructs the hull from the mask? It seems to me your `calc_concave_regions` could use the hull indices rather than the mask without much loss increase in complexity.

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,281 @@
...
+
+    // Calculate the concave regions, each supplied in clockwise order
+    VecVecReal2 calc_concave_regions() const;
+
+  private:
+    /// TYPES ///
+    using ConvexMask = std::vector<bool>;
+
+    /// DATA ///
+    VecReal2 const& points_;
```
sethrj: Note in the documentation that this helper class should have a shorter lifetime than the passed points. Alternatively, if the input is a vector that you will own, you could reverse and/or rotate the points so that they meet the "clockwise and lowest y is first" requirements, obviating `start_index_`.

src/corecel/math/ConvexHullFinder.hh
elliottbiondo: > And perhaps for completeness it should construct the tolerance using the bump calculator with the user-supplied rel+abs tolerance?

I am not sure what you mean by this. Are you saying that `ConvexHullFinder` should calculate `abs_tol_` using the `BumpCalculator` in order to construct via `SoftOrientation(real_type abs_tol)`?

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,281 @@
...
+    //! \name Type aliases
+    using real_type = T;
+    using Real2 = celeritas::Array<real_type, 2>;
+    using VecReal2 = std::vector<Real2>;
+    using VecVecReal2 = std::vector<VecReal2>;
+    //!@}
+
+  public:
+    // Construct with vector of ordered points
+    explicit ConvexHullFinder(VecReal2 const& points);
```
sethrj: Perhaps also pass in a `Tolerance` object? Deletion of small concave regions (or near collinearity of points) may be affected by the user tolerance.

src/orange/orangeinp/ConvexHullFinder.hh
elliottbiondo: I considered storing the originally created hull. However, writing this as a general-purpose algorithm, it felt more natural to store the mask, for several regions:

1) The originally created hull is typically rotated. I.e. it might be (10, 12, 0, 1, 2) rather than (0, 1, 2, 10, 12). It's nicer to return the points in order. So you still have to further process the originally created hull in some way no matter what your approach.
2) `calc_concave_regions` is simplest if you have a mask. If you store the hull, your probably going to make a mask anyway.
3) Lower storage (`bool` vs. `Real2`). Storing two vectors of `Real2` (`points_` and `hull_`) felt wasteful.  

Since I don't expect this to be a performance-critical part of the code, I don't think their is big difference either way, but this is the side I landed on.

sethrj: Yeah I was thinking more in terms of simplicity than performance: this function was getting kind of long. I think it's good enough and we can move on.

elliottbiondo:
@sethrj @mrguilima this is ready for another look

src/orange/orangeinp/ConvexHullFinder.hh
```diff
@@ -0,0 +1,289 @@
...
+                                       size_type i_next) const -> bool
+{
+    auto const& a = points_[i_prev];
+    auto const& b = points_[i];
+    auto const& c = points_[i_next];
+
+    // Create a SoftOrientation object based on the bump distance of the middle
+    // point. Since the BumpCalculator operates on Real3, use a z-value of 0.
+    auto abs_tol = calc_bump_({b[0], b[1], 0});
+    detail::SoftOrientation<real_type> soft_ori_(abs_tol);
```
sethrj: Is it reasonable to precalculate this based on the extents of the whole polygon? It seems like the tolerance should be relative to that, not the points being compared.

sethrj (approved):
Just the last question about soft orientation. (Also is the CHF going to be an implementation detail for the genpocone/shape constructors? If so it can+shoudl go into the detail dir.)

src/orange/orangeinp/ConvexHullFinder.hh
elliottbiondo: Yes, I think that would be better. `BumpCalculator::operator()` takes a point as argument though, so I wasn't sure what you were getting at. I could make the point `(max_x - min_x, max_y - min_y, 0)` if this doesn't feel like too much of a hack.

elliottbiondo:
Yes this will be used an an implementation detail; moved to detail dir.

sethrj (approved):
Very nice! Thanks for your patience with all the nitpicking.

PR #1739, opened 2025-04-14
Add uniform grid input and continue to simplify grid construction

amandalund:
Follow-on to #1735 and precursor to #1633, this:
  - Adds a `inp::UniformGrid` struct that's closer to our Celeritas grids/less like a Geant4 physics log vector
  - Simplifies the grid inserter classes
  - Removes most of the `ValueGridBuilder` code
  - No longer imports the (unused) range table from Geant4


app/celer-dump-data.cc
```diff
@@ -368,9 +368,9 @@ Microscopic cross sections:
                 cout << "| " << setw(20) << std::left
                      << (i == 0 ? geo_mat.name : std::string{}) << " | "
                      << setw(13) << std::left << elements[el_id].name << " | "
-                     << setprecision(3) << setw(12) << xs[i].front() / barn
+                     << setprecision(3) << setw(12) << xs[i].y.front() / barn
                      << " -> " << setprecision(3) << setw(12)
-                     << xs[i].back() / barn << " |\n";
+                     << xs[i].y.back() / barn << " |\n";
```
sethrj: If this utility is becoming annoying to maintain I think we can remove it. When we started ROOT didn't have a json export. On the future I think we could transition to using a python tool for formatting the export files.

src/celeritas/io/ImportModel.hh
```diff
@@ -70,16 +70,15 @@ struct ImportModelMaterial
...
 
     static constexpr auto energy_units{ImportUnits::mev};
     static constexpr auto xs_units{ImportUnits::len_sq};
 
-    std::vector<double> energy;  //!< Energy grid for the material
-    std::vector<MicroXs> micro_xs;  //!< Cross sections for each element
-
-    explicit operator bool() const { return energy.size() >= 2; }
+    double energy_min;  //!< Lower energy bound for the material
+    double energy_max;  //!< Upper energy bound for the material
```
sethrj: Default initialize to zero

test/celeritas/SimpleTestBase.cc
```diff
@@ -109,7 +109,8 @@ auto SimpleTestBase::build_physics() -> SPConstPhysics
         kn_model.materials.resize(this->material()->size());
         for (ImportModelMaterial& imm : kn_model.materials)
         {
-            imm.energy = {1e-4, 1e8};
+            imm.energy_min = 1e-4;
+            imm.energy_max = 1e8;
```
sethrj: This makes me wonder if the input would be better written as `Array<double, 2> energy; // min, max` or possibly `EnumArray<Bound, double>` (we'd move `Bound = {lo, hi}` from geocel into corecel).

sethrj (approved):
This is a great simplification! Thank you!

app/celer-dump-data.cc
amandalund: It is, and I was thinking the same.

sethrj: New `removal` pr? 😈

src/celeritas/inp/Grid.hh
```diff
@@ -55,6 +55,22 @@ struct Grid
...
+//---------------------------------------------------------------------------//
+/*!
+ * A uniform grid of increasing, sorted 1D data.
+ */
+struct UniformGrid
+{
+    using VecDbl = std::vector<double>;
+
+    double xmin{};
+    double xmax{};
```
sethrj: Should these also be an `EnumArray<Bounds, real_type> x;`?

amandalund: I was debating that since we're mostly operating on these individually, but yeah `x[Bound::lo]` isn't much longer than `xmin`.

sethrj: I was thinking particularly of the symmetry of `x = {0.1, 1e6};` which would be the same if it were a uniform grid or an arbitrary one? Maybe that's actually worse since the number of points changes the actual grid...

amandalund: Good points. I made the change; if it seems better we'll keep it, otherwise I can revert. I also simplified the `CalculatorTestBase`.

sethrj (approved):

PR #1742, opened 2025-04-16
Rename material ID types

sethrj:
At the very beginning of our project we made the mistake of not differentiating between material definitions and "material cuts couple" IDs, which have EM physics cross sections altered by physics cutoffs. We already defined a geometry material ID (#1288) and this one defines a physics material ID. I don't know if we want to do GeometryMaterialId/GeoMaterialId/GeoMatId; PhysicsMaterialId/PhysMaterialId/PhysMatId; OpticalMaterialId/OptMaterialId/OptMatId or what... so we can revert 5a32f7a52 if the names are too abbreviated (@pcanal )

This in preparation for work on #983 and wanted the big renaming done before 0.6.

### DIFF
--- a/src/celeritas/Types.hh
+++ b/src/celeritas/Types.hh
@@ -38,8 +38,7 @@ using UniqueEventId = OpaqueId<struct Event_, std::uint64_t>;
 using IsotopeId = OpaqueId<struct IsotopeRecord>;
 
 //! Opaque index of a material modified by physics options
-// TODO: rename to PhysMatId; equivalent to "material cuts couple"
-using MaterialId = OpaqueId<struct Material_>;
+using PhysicsMaterialId = OpaqueId<struct PhysicsMaterial_>;
 
 //! Opaque index of model in the list of physics processes
 using ModelId = OpaqueId<struct Model_>;
--- a/src/celeritas/Types.hh
+++ b/src/celeritas/Types.hh
@@ -38,13 +38,13 @@ using UniqueEventId = OpaqueId<struct Event_, std::uint64_t>;
 using IsotopeId = OpaqueId<struct IsotopeRecord>;
 
 //! Opaque index of a material modified by physics options
-using PhysicsMaterialId = OpaqueId<struct PhysicsMaterial_>;
+using PhysMatId = OpaqueId<struct PhysicsMaterial_>;
 
 //! Opaque index of model in the list of physics processes
 using ModelId = OpaqueId<struct Model_>;
 
 //! Opaque index to a material with optical properties
-using OpticalMaterialId = OpaqueId<struct OpticalMaterial_>;
+using OptMatId = OpaqueId<struct OpticalMaterial_>;
 
 //! Opaque index to ParticleRecord in a vector: represents a particle type
 using ParticleId = OpaqueId<struct Particle_>;
--- a/src/celeritas/optical/MaterialData.hh
+++ b/src/celeritas/optical/MaterialData.hh
@@ -36,7 +36,7 @@ struct MaterialParamsData
 
     OpticalMaterialItems<NonuniformGridRecord> refractive_index;
     VolumeItems<OpticalMaterialId> optical_id;
-    OpticalMaterialItems<CoreMaterialId> core_material_id;
+    OpticalMaterialItems<CorePhysicsMaterialId> core_material_id;
 
     // Backend data
     Items<real_type> reals;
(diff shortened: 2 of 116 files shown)


src/celeritas/optical/MaterialData.hh
```diff
@@ -28,15 +28,15 @@ struct MaterialParamsData
...
     template<class T>
     using VolumeItems = celeritas::Collection<T, W, M, VolumeId>;
 
     //// MEMBER DATA ////
 
     OpticalMaterialItems<NonuniformGridRecord> refractive_index;
-    VolumeItems<OpticalMaterialId> optical_id;
-    OpticalMaterialItems<CoreMaterialId> core_material_id;
+    VolumeItems<OptMatId> optical_id;
+    OpticalMaterialItems<CorePhysicsMaterialId> core_material_id;
```
amandalund: `CorePhysicsMaterialId ` -> `CorePhysMatId`?

src/celeritas/phys/CutoffData.hh
```diff
@@ -63,7 +63,8 @@ struct CutoffParamsData
     ParticleItems<size_type> id_to_index;
 
     ParticleId::size_type num_particles;  //!< Particles with production cuts
-    MaterialId::size_type num_materials;  //!< All materials in the problem
+    PhysMatId::size_type num_materials;  //!< All materials in the
+                                         //!< problem
```
amandalund: Weird clang-format change...

test/corecel/data/Collection.test.hh
```diff
@@ -31,7 +31,7 @@ struct MockMaterial
     ItemRange<MockElement> elements;
 };
 
-using MockMaterialId = ItemId<MockMaterial>;
+using MockPhysicsMaterialId = ItemId<MockMaterial>;
```
amandalund:
```suggestion
using MockPhysMatId = ItemId<MockMaterial>;
```
(but just a test so doesn't really matter)

amandalund (approved):
Thanks @sethrj! I'm happy with the abbreviated names.

src/celeritas/Types.hh
```diff
@@ -38,14 +38,13 @@ using UniqueEventId = OpaqueId<struct Event_, std::uint64_t>;
...
-// TODO: rename to PhysMatId; equivalent to "material cuts couple"
-using MaterialId = OpaqueId<struct Material_>;
+using PhysMatId = OpaqueId<struct PhysicsMaterial_>;
 
 //! Opaque index of model in the list of physics processes
 using ModelId = OpaqueId<struct Model_>;
 
 //! Opaque index to a material with optical properties
-using OpticalMaterialId = OpaqueId<struct OpticalMaterial_>;
+using OptMatId = OpaqueId<struct OpticalMaterial_>;
```
pcanal, line 47: This is fine(ish) as long as we don't use the `Opt` prefix anywhere else to mean `Optional`

amandalund: Hopefully we never need an optional optimized optical material ID

src/celeritas/Types.hh
```diff
@@ -38,14 +38,13 @@ using UniqueEventId = OpaqueId<struct Event_, std::uint64_t>;
 using IsotopeId = OpaqueId<struct IsotopeRecord>;
 
 //! Opaque index of a material modified by physics options
-// TODO: rename to PhysMatId; equivalent to "material cuts couple"
-using MaterialId = OpaqueId<struct Material_>;
+using PhysMatId = OpaqueId<struct PhysicsMaterial_>;
```
pcanal, line 41:
```suggestion
using PhysCutsMatId = OpaqueId<struct PhysicsCutsMaterial_>;
```
I am leaning toward `Cuts` being an important part of the name/concept.   Albeit that order of the 3 concepts is a tad bit hard on the tongue.

src/celeritas/Types.hh
mrguilima: That would be the easier one: Opt3MatId!

src/celeritas/phys/CutoffData.hh
sethrj: I messed up the find-replace the first time, making the name longer, then it reformatted on commit  😔 thanks!

src/celeritas/Types.hh
sethrj: I don't think `Cuts` is useful at all except to people intimately acquainted with geant4. The key difference between two "physics materials" that have the same "geometry material" is that the physics treatment is different in the two. That could be different secondary production cutoff, or a different physics list.

PR #1743, opened 2025-04-16
Simplify grid construction in a few more places

amandalund:
One more round of simplifications in a few places I missed...


src/celeritas/grid/NonuniformGridBuilder.hh
```diff
@@ -6,58 +6,41 @@
...
 #include "celeritas/inp/Grid.hh"
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Construct a generic grid.
- *
- * This uses a deduplicating inserter for real values to improve cacheing.
+ * Construct a nonuniform grid.
```
sethrj, line 23: Was there a reason to delete the text about deduplication, or do all our builders do that now?

src/corecel/grid/detail/GridAccessor.hh
```diff
@@ -49,19 +58,18 @@ class GridAccessor
 
 //---------------------------------------------------------------------------//
 /*!
- * Helper class for accessing grid data from spans.
+ * Helper class for accessing grid data from a generic grid.
```
sethrj: I think "generic grid" is obsolete (#1616) so perhaps make sure all the stuff in corecel/grid and celeritas/grid say "nonuniform grid".

sethrj (approved):
You're killing it with the code removal 😄

src/celeritas/grid/NonuniformGridBuilder.hh
amandalund: Not intentional, I must have just gotten carried away with the deleting. But yes, all our grid inserters should be using deduplicating builders.

PR #1753, opened 2025-04-20
Support testing nested container equivalence and tweak `Histogram` class

amandalund:
This tweaks a couple of the testing utilities:
  - Adds support for checking nested container equivalence
  - Makes the `Histogram` helper behave more like a numpy histogram
    - The density is normalized such that the _integral_ over the range is 1 (it is a probability density function rather than a probability mass function)
    - Bins are half-open except for the righthand-most bin, which is closed. Values equal to the upper range value will be included
    - Values outside the range are ignored
  - Adds a `Histogram` test


sethrj (approved):

sethrj:
Neat! From a testing standpoint I kind of prefer mass over density (especially when choosing power of 10 samples so that we get an exact number of digits out) and count "outside the distribution " as an explicit bin (so we don't need to manually check the min/max of the result). However consistency is valuable so we might as well do what numpy does.

amandalund:
Yeah, my thinking was along the same lines. I also thought the exact representation and fewer digits with mass was nicer for testing, and often easier to visually inspect and compare (I tried to recover some of that clarity with the nested container equivalence). Since we're mostly dealing with continuous distributions, density probably makes more sense, and as you said consistency with libraries like numpy is valuable because it makes the behavior of the utility more predictable and familiar for anyone using it. I see how counting values outside the distribution could be useful for testing... maybe we could add a method to return the number of "out of range" values?

sethrj:
Maybe overkill, but what about min/max bins that expand outward from the user-provided domain? Those would keep the spirit of the PDF and give information about the extended range too.
```
EXPECT_EQ(0, hist.count(Bounds::lo)) << "Sampled result as low as " << hist.domain(Bounds::lo);
// or
EXPECT_EQ(0, hist.min().second) << "Sampled result as low as " << hist.min().first;
```

amandalund:
Yeah I think something like that would be useful.

pcanal:
> count "outside the distribution " as an explicit bin (so we don't need to manually check the min/max of the result). However consistency is valuable so we might as well do what numpy does.

Knowing how much of the data is outside the range is an interesting debug tool and I would indeed recommend adding the underflow and overflows bins.  If those bins contains more than a few values, their might be something very wrong with the filling code (e.g. the range is wrong or the data is completely off) that might not be appearing in some case in the shape of the histogram.

test/corecel/random/Histogram.hh
```diff
@@ -39,39 +42,79 @@ class Histogram
...
-    ++total_counts_;
     double frac = (value - offset_) * inv_width_;
-    if (frac < 0.0 || frac >= 1.0)
+    if (frac < 0.0)
     {
+        ++out_of_range_[Bound::lo];
+        extrema_[Bound::lo] = std::min(extrema_[Bound::lo], value);
+        return;
+    }
+    else if (frac > 1.0)
```
sethrj: Minor thing to watch out for: it's a bit safer to order this as `if (frac < 0) else frac < 1 else frac == 1 else ...`. This way, NaN (which is false for all comparisons) falls into the upper extremum and it would stay there. As written it's dangerous because `static_cast<size_type>(NaN)` is UB.

sethrj (approved):
Nice! Just watch out for NaN cat.

test/corecel/random/Histogram.hh
amandalund: Good point!

PR #1758, opened 2025-04-22
Round all floating point numbers in test output to a few digits

sethrj:
This finishes up a little work I was doing earlier this month to make stringy results less sensitive to small numerical differences (including float vs double). It now rounds all floating point values in filtered string outputs to the specified number of digits (by default).


amandalund (approved):
Thanks @sethrj!

PR #1759, opened 2025-04-22
Add helper for accumulating histograms

sethrj:
I spent way too much time on this (ugh) but it incorporates the "expect ref equal" that makes it easier to compare structs of data, and a new function `accumulate_n` and helper class `HistogramSampler` that are quite useful for looping over different particle types, energy distributions, etc.


amandalund (approved):
Thanks @sethrj, nice extensions!

test/testdetail/TestMacrosImpl.hh
```diff
@@ -287,19 +290,19 @@ using ValueTypeT = typename ValueType<T>::type;
 /*!
  * Recursively get the underlying scalar type of a container.
  */
-template<typename T, typename = void>
+template<class T, class = void>
```
pcanal, line 293: This change seems odd.
```suggestion
template<typename T, typename = void>
```
Since this seems to be the tail end of the recursion where `T` will/might be a scalar type rather than a class.

test/testdetail/TestMacrosImpl.hh
```diff
@@ -724,6 +727,81 @@ template<class ContainerE, class ContainerA, class T>
...
+            else
+            {
+                item_result = IsRefEq(expr1, expr2, *iter1, *iter2);
+            }
+            if (!item_result)
+            {
+                if (failures++ < max_printable_failures)
+                {
+                    result.fail() << item_result << "\n(Failed in element "
+                                  << i << " of " << expr2 << ")";
```
pcanal, line 776: Should we also mention `expr1` here?

test/testdetail/TestMacrosImpl.hh
sethrj: As a rule we use class rather than type name for template arguments. We use type name only for aliases/type resolution.

pcanal: A bit 'confusing' but since this is the convention the rest of the code use then fine :).

test/testdetail/TestMacrosImpl.hh
sethrj: It should be the reference and is long, so I think better to omit to avoid confusion. It often shows the diff anyway. This is mostly for facilitating copy paste to update reference values.

PR #1761, opened 2025-04-23
Fix missing header and typo 

esseivaju:
Fix a missing include causing a compile error with apple-clang 17 and a typo from #1744


pcanal (approved):

sethrj (approved):
Thanks!

PR #1764, opened 2025-05-01
Adds Sensitive Detector parameter Class

lebuller:
Creates a new `SDParamsData` struct which stores a Collection for mapping between volume IDs and detector IDs. A New `SDParams` class is created which stores the `SDParamsData` which creates the detector collection based on an input vector of volume labels. An accessor function for the detector collection is also present in the new `SDParams` class


src/celeritas/user/SDData.hh
```diff
@@ -0,0 +1,52 @@
...
+
+    // ! boolean for assignment of struct.
+    // ! TODO: this is a placeholder for now since COllection doesn't case to
+    // bool and is the only member
+    bool assigned;
+
+    //! Whether the data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return static_cast<bool>(assigned);
```
sethrj: You should be able to use `detector.empty()` and eliminate the `assigned` variable.

src/celeritas/user/detail/SDParams.hh
```diff
@@ -0,0 +1,59 @@
...
+
+#include "../SDData.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+class GeoParamsInterface;
+
+namespace detail
+{
```
sethrj: I think we can put the SD params in the top-level database, and in fact we may want to add it to core params.

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,102 @@
...
+    {
+        detector_map[volume_ids_[didx]] = DetectorId{didx};
+    }
+
+    auto this_has_detectors = detector_map.empty() ? HasDetectors::none
+                                                   : HasDetectors::all;
+    if (has_det == HasDetectors::unknown)
+    {
+        has_det = this_has_detectors;
+    }
```
sethrj: This variable was only used for the StepCollector stuff: basically to determine whether or not to filter on volume IDs. It can be deleted from this constructor.

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,102 @@
...
+                   VecLabel labels,
+                   GeoParamsInterface const& geo)
+    : output_label_{std::move(output_label)}, volume_labels_{std::move(labels)}
+{
+    CELER_EXPECT(!output_label_.empty());
+    if (volume_labels_.empty())
+    {
+        return;
+    }
+    CELER_EXPECT(!volume_labels_.empty());
```
sethrj: Let's require that the user provide at least one volume with this constructor. If the problem has no SDs, we could just use the default constructor.

src/celeritas/user/detail/SDParams.hh
```diff
@@ -0,0 +1,59 @@
...
+
+    //! Access sensitive detector properties on the host
+    HostRef const& host_ref() const final { return mirror_.host_ref(); }
+    //! Access sensitive detector properties on the device
+    DeviceRef const& device_ref() const final { return mirror_.device_ref(); }
+    //!@}
+
+  private:
+    std::string output_label_;
+    VecLabel volume_labels_;
```
sethrj: We don't really needto keep the volume labels around: those are provided by the geometry given a volume ID.

src/celeritas/user/detail/SDParams.hh
```diff
@@ -0,0 +1,59 @@
...
+    std::vector<VolumeId> volume_ids_;
```
sethrj: Let's make an accessor for this: `detector_to_volume_id(DetectorId)`.

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,102 @@
...
+namespace celeritas
+{
+
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from list of volume labels.
+ */
+SDParams::SDParams(std::string output_label,
```
sethrj: I don't think this is needed; we'll only have one SD params.

sethrj:
Tagging https://github.com/celeritas-project/celeritas/issues/1397 : also @lebuller can you sketch out on that page the execution plan we discussed? (Or maybe we need to sit down and write one out if we only discussed this as a next step?)

sethrj (commented):
Solid start!

src/celeritas/user/detail/SDParams.hh
```diff
@@ -0,0 +1,59 @@
...
+class GeoParamsInterface;
+
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Manage params and state data for sensitive detectors.
+ *
+ */
+class SDParams : public ParamsDataInterface<SDParamsData>
```
sethrj:
```suggestion
class SDParams final : public ParamsDataInterface<SDParamsData>
```
To fix the build warning/error.

src/celeritas/user/SDData.hh
lebuller: Good point, changed to remove `assigned` and use `detector.empty()` in this operator.

src/celeritas/user/detail/SDParams.cc
lebuller: Got it, `has_det` and the `HasDetectors` object have been removed.

src/celeritas/user/detail/SDParams.cc
lebuller: Default constructor added and is now used in optical `CoreParams` when no detector_labels are included in `CoreParams::input`

src/celeritas/user/detail/SDParams.hh
lebuller: removed as a member

src/celeritas/user/detail/SDParams.hh
lebuller: added

src/celeritas/user/detail/SDParams.cc
lebuller: Ok, I have removed the output_label, both as an input parameter and a member of `SDParams`

src/celeritas/user/detail/SDParams.hh
lebuller: got it, fixed

src/celeritas/user/detail/SDParams.hh
lebuller: Not sure I understand this comment correctly. To clarify, remove the detail namespace here to keep the `SDParams` just in the `celeritas` namespace? Would I need to move the file to a different directory, like `global/detail` where the existing global `CoreParams` are? and when you say add it to Core params, do you mean add it to the global core params similar to how it is added to the optical version here already?

sethrj: Sorry, I meant to write top level *namespace*, so just move it to `celeritas/user`. In a follow-on commit, we might want to add this to the global core params.

src/celeritas/user/SDData.hh
```diff
@@ -0,0 +1,46 @@
...
+#include "corecel/data/CollectionBuilder.hh"
+#include "celeritas/Quantities.hh"
+#include "celeritas/Types.hh"
+
+namespace celeritas
+{
+
+namespace detail
+{
+
```
sethrj: Neither SDParams nor SDData should be in the `detail/` subdirectory nor be in the `detail` namespace. "detail" essentially corresponds to "only used by code in the parent directory" (i.e., highly localized) but we will be using the SD params elsewhere.

src/celeritas/user/SDData.hh
```diff
@@ -0,0 +1,46 @@
...
+{
+    //// DATA ////
+
+    // ! Mapping for volume -> sensitive detector
+    Collection<DetectorId, W, M, VolumeId> detector;
+
+    //! Whether the data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return static_cast<bool>(detector.empty());
```
sethrj: Since `detector.empty()` returns a bool the static cast is superfluous.

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,83 @@
...
+namespace celeritas
+{
+
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from list of volume labels.
+ */
+SDParams::SDParams(VecLabel volume_labels, GeoParamsInterface const& geo)
```
sethrj: Since `volume_labels` is only used locally and never modified, the argument should be a const ref.

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,83 @@
...
+    std::map<VolumeId, DetectorId> detector_map;
+    for (auto didx : range<DetectorId::size_type>(volume_ids_.size()))
+    {
+        detector_map[volume_ids_[didx]] = DetectorId{didx};
+    }
+
+    mirror_ = CollectionMirror{[&] {
+        HostVal<SDParamsData> host_data;
+        if (!detector_map.empty())
+        {
```
sethrj: This should always be true since we require at least one volume label

src/celeritas/user/detail/SDParams.cc
```diff
@@ -0,0 +1,83 @@
...
+
+DetectorId SDParams::volume_to_detector_id(VolumeId vol_id)
+{
+    return host_ref().detector[vol_id];
+}
+
+VolumeId SDParams::detector_to_volume_id(DetectorId det_id)
+{
+    return volume_ids_[det_id.get()];
+}
```
sethrj: Simple accessors like these should generally be inline in the .hh file. Also, don't forget to do bounds checks (`CELER_EXPECT(det_id < volume_ids_.size());`) and add doxygen docs.

sethrj (commented):
Could you also add a simple unit test? It should be pretty easy to make a geometry-only harness. I think this should give you what you need to get the geometry:
```c++
class SDParamsTest : public GlobalGeoTestBase,
                             public OnlyGeoTestBase,
                             public OnlyCoreTestBase
{
    public:
    std::string_view geometry_basename() const override
    {
        return "testem3-flat"sv;
    }
};
```
See `celeritas/geo/Geometry.test.cc` (the HeuristicGeoTestBase is also set up like the above class)

sethrj:
@lebuller Did you see the comments above?

lebuller:
> @lebuller Did you see the comments above?

Yes, Just working on getting the unit test working, then will push

src/celeritas/user/SDParams.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "corecel/io/Join.hh"
+#include "geocel/GeoVolumeFinder.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from list of volume labels.
+ */
+SDParams::SDParams(VecLabel const volume_labels, GeoParamsInterface const& geo)
```
sethrj:
```suggestion
SDParams::SDParams(VecLabel const& volume_labels, GeoParamsInterface const& geo)
```

sethrj (approved):
Great! I'll swing by to talk unit test and then we can get this in.

sethrj (approved):
Thanks @lebuller !

PR #1765, opened 2025-05-02
Replace hand-rolled hooks with pre-commit and CI

### LINKED ISSUE #1760: Use `pre-commit` tool for formatting and linting
The ACTS folks (@stephenswat) have set up a couple of best practices that we should also adopt.

`pre-commit`, which is actually used in celerpy, will do linting/formatting:
- [ ] Replace our `install-commit-hooks.sh` and pre/post commit scripts
- [ ] Add `pre-commit` to the CI to check for (but not fix) errors
- [ ] Have it generate patches that are uploaded and suggested as comments with a `pull_request_completed` action (or pushed directly to the branch)

sethrj:
This adds support for [pre-commit](https://pre-commit.com) to automatically clang-format and do other validation. I've also added the [pre-commit.ci](https://pre-commit.ci/) app which should push fixes to PRs in case the formatting checks fail.

Closes #1760 .


test/celeritas/data/event-novtx.hepmc3
```diff
@@ -22,4 +22,3 @@ P 3 0 22 5.1336099422574807e+02 5.4197742781708928e+02 6.6537279576514345e+02 1.
 P 4 0 22 -3.6655746358147650e+02 8.0035990693978067e+02 4.7440451647940733e+02 1.0000000000000000e+03 0.0000000000000000e+00 1
 P 5 0 22 -7.8969793730748802e+02 -5.4961247282688134e+02 -2.7258631206540650e+02 1.0000000000000000e+03 0.0000000000000000e+00 1
 HepMC::Asciiv3-END_EVENT_LISTING
-
```
pcanal: The hepmc3 and dat files should probably be excluded from the reformatting.

app/CMakeLists.txt
```diff
@@ -4,54 +4,69 @@
...
 
 if(BUILD_TESTING)
-  if(CELERITAS_USE_Python)
-    set(CELER_PYTHON "$<TARGET_FILE:Python::Interpreter>")
-    set(CELER_NEEDS_PYTHON)
-  else()
-    set(CELER_PYTHON "python3")
-    set(CELER_NEEDS_PYTHON DISABLED true)
-  endif()
+    if(CELERITAS_USE_Python)
```
pcanal: Do we mean to indent by 4?

pcanal:
For clarity I would consider merging the active part of the PR separately from the application of the commit hooks (for example, we could merge the commit hooks induces changes in a first PR (i.e. a NFC PR) and the active changes in a second PR).

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+      args: ['--maxkb=100']
+    - id: check-case-conflict
+    - id: check-merge-conflict
+    - id: check-shebang-scripts-are-executable
+    - id: check-json
+    - id: check-yaml
+    - id: end-of-file-fixer
+      exclude: '\.(diff|patch)$'
+    - id: trailing-whitespace
+      exclude: '\.(diff|patch)$'
```
pcanal, line 15:
```
    - id: end-of-file-fixer
      exclude: '\.(diff|patch|dat|hepmc3|root)$'
    - id: trailing-whitespace
      exclude: '\.(diff|patch|dat|hepmc3|root)$'
```

esseivaju:
Regarding the auto-fix to PR, do we always have to explicitly approve the workflow run before auto fixes are applied? If not, that could be an issue if you make a bunch of changes locally, but then end up with conflicts because `pre-commit` pushed a bunch of formatting changes.

sethrj:
> For clarity I would consider merging the active part of the PR separately from the application of the commit hooks (for example, we could merge the commit hooks induces changes in a first PR (i.e. a NFC PR) and the active changes in a second PR).

I didn't realize that would happen: that's how I planned it. Sorry!

app/CMakeLists.txt
sethrj: Yep thank you, I've added it to the rc file.

.pre-commit-config.yaml
sethrj: ROOT doesn't get touched since it's a binary file, and since the hepmc3 files are text I think those are OK to clean.

sethrj:
> Regarding the auto-fix to PR, do we always have to explicitly approve the workflow run before auto fixes are applied? If not, that could be an issue if you make a bunch of changes locally, but then end up with conflicts because `pre-commit` pushed a bunch of formatting changes.

I added a section to the CONTRIBUTING about how to fix this :)

test/celeritas/data/event-novtx.hepmc3
sethrj: I think it should be OK since they're just plain text?

pcanal: Are the `dat` file plain text?
I don't know the syntax of the hepmc3 files but whatever automatic formatting is allowed to be run on them should be checked against that syntax (and if it is not clear, I would side on the side of caution (no auto updates)).

pcanal (approved):
This is a very nice improvement and seem to improve the hooks manageability.  The only caveat is that I would exclude the files for which we are not clear on the syntax requirement from auto-formatting.

test/celeritas/data/event-novtx.hepmc3
sethrj: It's just a matter of trailing whitespace and newlines and such; the CI would catch any error that's there. From what I know of the EVT/HEPMC3 formats there aren't any whitespace changes that would matter.

PR #1767, opened 2025-05-06
Require covfie v0.14

esseivaju:
Covfie [v0.13](https://github.com/acts-project/covfie/releases/tag/v0.13.0), and [v0.14](https://github.com/acts-project/covfie/releases/tag/v0.14.0) both have many important bug fixes. Since this is an external dependency that will get downloaded if not found, we can require `v0.14.0`.

Once [this](https://github.com/acts-project/covfie/pull/85) is merged, we can move away from using `std::unique_ptr<field>` in the field params data, and instead use field move semantics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,7 +409,7 @@ if(CELERITAS_USE_ROOT)
 endif()
 
 if(CELERITAS_USE_covfie)
-  celeritas_find_or_builtin_package(covfie)
+  celeritas_find_or_builtin_package(covfie 0.14.0)
 
   if(CELERITAS_USE_HIP AND covfie_VERSION VERSION_LESS "0.13.0")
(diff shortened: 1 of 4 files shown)


esseivaju:
Not sure why `-lcore  -lcpu` are being linked. I cannot reproduce this locally and it's working... even with covfie 0.12.1 installed in Spack, and fetching 0.14.0 using FetchContent

pcanal (approved):
Thanks.

sethrj:
We should remember to update the spack recipe when updating dependency requirements: also it looks like this release isn't yet available from spack?

esseivaju:
> We should remember to update the spack recipe when updating dependency requirements: also it looks like this release isn't yet available from spack?

Yep, it's not released on Spack yet, the builtin will override the Spack installation for now. Should be there soon.

PR #1768, opened 2025-05-06
Refactor micro xs CDF calculation and add test

amandalund:
This moves the calculation of the CDF from the model microscopic cross sections used for sampling an element from the `PhysicsParams` to a helper class and adds a test for it.


pcanal (approved):
LGTM.

PR #1769, opened 2025-05-07
Add missing typename and const to CartMapField

esseivaju:
~~The CartMapFieldParamsData have a base class that wasn't correctly initialized as we need to specify the base class members first in aggregate initialization~~

It turns out that this should be fine because the compiler must use RVO. Repurpose this PR to minor tweaks: missing typename on type alias, and const deleter function.


amandalund (approved):
Thanks @esseivaju!

pcanal:
Is it possible to test for this problem?

pcanal:
The failure seems somehow related to the covfie PR ....

esseivaju:
This is caused by the tar archive's shasum changing. I asked @stephenswat and he hasn't retagged anything so not sure what triggered the change. Github has a [policy](https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/#future-stability-of-archives-and-hashes) of keeping the hash stable for at least a year...

esseivaju:
> Is it possible to test for this problem?

I'm not sure why this wasn't causing a crash in the first place since the c++ standard clearly says that the elements are:

> for a class, the direct base classes in declaration order, followed by the direct non-static data members that are neither anonymous [bit-fields](https://en.cppreference.com/w/cpp/language/bit_field) nor members of an [anonymous union](https://en.cppreference.com/w/cpp/language/union#Anonymous_unions), in declaration order.

pcanal:
>> I'm not sure why this wasn't causing a crash

It might/should lead to the wrong value being used (i.e. the value being stored in the wrong place) ....

esseivaju:
Actually, this might be fine because of named return value optimization, [this example](https://godbolt.org/z/P7o9KPr73) assigns correctly but calls neither copy nor move constructors

pcanal:
Fair enough, that is a good reason why we may not have seen it ... but then ... how did you notice there was an issue?

esseivaju:
> Fair enough, that is a good reason why we may not have seen it ... but then ... how did you notice there was an issue?

I had a warning about object slicing (value-initializing base from derived object), then looked up the rules for aggregate initialization, but that's not what's happening here since the struct is using an implicitly-declared constructor which then gets optimized away. We have a `CELER_ENSURE` that [checks](https://github.com/celeritas-project/celeritas/blob/ae44eed9342a9ea6aa2f75895ec57473b8aa0dd8/src/celeritas/field/CartMapFieldParams.covfie.cc#L104) that the field pointer is [set](https://github.com/celeritas-project/celeritas/blob/ae44eed9342a9ea6aa2f75895ec57473b8aa0dd8/src/celeritas/field/CartMapFieldData.covfie.hh#L43)

Maybe it's better to be explicit about using aggregate initialization? I'm not 100% sure on the rules defining which initialization the compiler must pick first, both (aggregate initialization of base class from derived object, or copy/move constructor) would be valid there.

3rd edit 😅 : actually, since we're using an IIFE, the compiler would have to use RVO and initialize `host_` in place from the returned value.

We can close or repurpose this PR

pcanal:
@esseivaju The PR seems to have change a bit from the start.  Can you verify that the description still match the changes?

pcanal (approved):

PR #1770, opened 2025-05-07
Fail early if find-or-builtin package is missing

sethrj:
In a build failure on https://github.com/spack/spack/pull/50197 I realized that if the BUILTIN cmake setting is provided by the user, and the package is requested but unavailable, the `find_or_builtin` call will not fail like it should.


cmake/CeleritasOptionUtils.cmake
```diff
@@ -168,16 +173,19 @@ macro(celeritas_find_or_builtin_package package)
       # Celeritas is a subproject)
       find_package(${package} ${ARGN})
     endif()
+    set(_found "${${package}_FOUND}")
```
pcanal, line 176: Since this is macro, it might be helpful to obfuscate the name more.
```suggestion
    set(_celeritas_builtin_found "${${package}_FOUND}")
```
(and of course the other related changes)

pcanal (approved):
Thanks.

esseivaju:
Did we have a motivation for implementing our own `celeritas_find_or_builtin_package` instead of using CMake's [try_find_package](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE)?

cmake/CeleritasOptionUtils.cmake
sethrj: These local variables are "unset" immediately after. You'll see the same idiom in official `FindX.cmake` packages too.

sethrj:
> Did we have a motivation for implementing our own `celeritas_find_or_builtin_package` instead of using CMake's [try_find_package](https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE)?

First thing I can think of is that CMake 3.24 would be required (released Aug 2022 so we could probably bump our minimum) and I haven't yet looked into how the behavior of that feature might differ. If it's effectively the same, and won't be interfered with by our `try_find_package`, then we can switch anytime and bump the cmake minimum.

cmake/CeleritasOptionUtils.cmake
pcanal: hummm ... If somebody make the 'mistake' to write:
```
set(_found SomethingClever)
celeritas_find_or_builtin_package(....)
message(STATUS "Is ${_found} still something clever")?
```

sethrj: Then it doesn't show anything, because `_found` is unset.

pcanal: exactly my point :).  In my opinion, it should print 'SomethingClever'; the caller of `celeritas_find_or_builtin_package` should not be expected to know the internal variables used and wiped by `celeritas_find_or_builtin_package` :)

PR #1771, opened 2025-05-07
Patch covfie shasum

esseivaju:
The shasum for covfie tarball changed overnight. The current Github [policy](https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/#future-stability-of-archives-and-hashes) should have the hash stable for a year

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -44,7 +44,7 @@ if(CELERITAS_BUILTIN_covfie)
   celer_fetch_external(
     covfie
     "https://github.com/acts-project/covfie/archive/refs/tags/v0.14.0.tar.gz"
-    7a6331526a4870caa980959fdbd67057de0ef4665740ae81251e8310bf386bc6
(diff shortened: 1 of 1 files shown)


pcanal (approved):
Good to go if the CI is green

sethrj:
Oops, this is likely a side effect of https://github.com/acts-project/covfie/pull/83 . The official recommendation of the spack folks has been to download and re-upload the tarball to keep it consistent, because if github changes their compression level/zip implementation then the hashes also change...

sethrj:
@stephenswat I forgot to mention that using the `export-subst` git attribute makes the repository tarball less stable. The `.tar.gz` is not guaranteed to be stable in the first place (it's caused some consternation once or twice in Spack when GitHub changed their compression level/zip implementation), so if people rely on checksums then you should upload a copy of the repository as an artifact and make *that* the primary object to download. I've up the process for Celeritas with a semi-automated GH API script at https://github.com/celeritas-project/release-scripts/blob/main/release_notes.py#L458 if you want to have a look.

PR #1772, opened 2025-05-08
Add default values for track slots and initializer capacity

amandalund:
This adds reasonable default values (based on our scaling studies) for `num_track_slots` and `initializer_capacity` in accel, celer-g4, and celer-sim. I've also added a default `secondary_stack_factor` in celer-sim and made that value consistent everywhere.


sethrj (approved):
Nice!

src/accel/SetupOptions.hh
```diff
@@ -238,6 +239,15 @@ struct SetupOptions
     {
         return max_num_tracks > 0 && static_cast<bool>(make_along_step);
     }
+
+    //! Initialize runtime-dependent defaults
+    SetupOptions()
+    {
+        // Set the default number of tracks per stream on device or host
+        max_num_tracks = celeritas::Device::num_devices() ? 262144 : 1024;
```
sethrj: This will need to be handled in the `inp` stuff or deferred till setup time: since people sometimes have static global `SetupOptions`, possibly before CUDA/MPI are initialized, `Device` can't be used here. You could make `max_num_tracks` and `initializer_capacity` a `std::optional<size_type>` and then set defaults when loading them?

src/corecel/io/JsonUtils.json.hh
```diff
@@ -33,6 +33,23 @@
...
+    do                                                      \
+    {                                                       \
+        if (auto iter = OBJ.find(#NAME);                    \
+            iter != OBJ.end() && !iter->is_null())          \
+        {                                                   \
+            iter->get_to(STRUCT.NAME);                      \
+        }                                                   \
+        else                                                \
+        {                                                   \
+            STRUCT.NAME = DEFAULT;                          \
```
sethrj: Maybe also log a debug message `Set <NAME> = <DEFAULT>`?

esseivaju:
Since these are changes to `accel`, should we maybe have a class there that centralizes good default parameters, e.g.
```cpp
struct DefaultOptions
{
    constexpr int stack_factor{8};
};
``` 
Then all downstream applications can set their input from there. You could even later add template parameters to specialize for the build options / system we're building on.

src/accel/SetupOptions.hh
amandalund: Ah right, damn. I initially set them in the `inp` in `ProblemSetup`, but we also need `max_num_tracks` from `SetupOptions` in the `LocalTransporter`. I'll rework it...

amandalund:
> Since these are changes to accel, should we maybe have a class there that centralizes good default parameters

Yeah, definitely good to have them all centralized. That should be where we're headed with the `inp` refactor: I think eventually we'll be removing the `SetupOptions` and the app inputs altogether and replacing them with the `inp` classes, which will be reused across the different front ends and should reduce duplication and improve consistency..

esseivaju (approved):

PR #1773, opened 2025-05-08
Apply pre-commit hooks across repository

sethrj:
This applies the pre-commit hooks (#1765) to the whole repository. After this is merged, we can enable `pre-commit.ci`.

The `gersemi` tool generated too many warnings and weird formatting, so I've removed it (we can later revert 8be53467a0b7404e447be375a4f8369806927f6a).

@celeritas-project/core After this you *must* install `pre-commit` (your old pre-commit hooks should auto-update or inform you how to do so: for mac users, `brew install pre-commit`); you should also update your clang-format installation to 20.1.3 which is the hardcoded version used by pre-commit.


esseivaju:
> you should also update your clang-format installation to 20.1.3 which is the hardcoded version used by pre-commit.

I think pre-commit should automatically download its own clang-format installation.

sethrj:
> I think pre-commit should automatically download its own clang-format installation.

Sure, I was just thinking for example my vim/vscode built-in formatter; there's sometimes small differences.

PR #1776, opened 2025-05-10
Fix and refactor maximum cross section energy calculation for integral approach

amandalund:
This moves the calculation of the maximum cross section for the integral approach from the physics params to a helper class and fixes a couple of bugs:
  - Scaled cross section wasn't being scaled by energy
  - Final comparison of upper and lower grid values was of energies, not xs

None of our processes that use the integral approach have scaled cross section grids, so we won't have run into this in practice, but I added a test that failed before the fixes.

### DIFF
--- a/test/celeritas/MockTestBase.cc
+++ b/test/celeritas/MockTestBase.cc
@@ -13,6 +13,7 @@
 #include "celeritas/mat/MaterialParams.hh"
 #include "celeritas/phys/CutoffParams.hh"
 #include "celeritas/phys/ParticleParams.hh"
+#include "celeritas/phys/PhysicsOptions.hh"
 #include "celeritas/phys/PhysicsParams.hh"
 #include "celeritas/track/SimParams.hh"
 #include "celeritas/track/TrackInitParams.hh"
@@ -223,8 +224,9 @@ auto MockTestBase::build_physics() -> SPConstPhysics
         inp.materials = this->material();
         inp.interact = interact;
         inp.label = "barks";
-        inp.applic = {make_applicability("electron", 1e-5, 10)};
+        inp.applic = {make_applicability("electron", 1e-5, 1e9)};
         inp.xs = {Barn{0}, Barn{6.0}, Barn{12.0}, Barn{6.0}};
+        inp.xs_scaled = {Barn{6.0}, Barn{4.0}, Barn{2.0}, Barn{1}, Barn{0}};
         inp.energy_loss = MevCmSqLossDens{0.5 * 1e-20};
         inp.interp = this->interpolation();
         physics_inp.processes.push_back(std::make_shared<MockProcess>(inp));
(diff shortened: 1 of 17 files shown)


test/celeritas/MockTestBase.cc
```diff
@@ -223,8 +224,9 @@ auto MockTestBase::build_physics() -> SPConstPhysics
         inp.materials = this->material();
         inp.interact = interact;
         inp.label = "barks";
-        inp.applic = {make_applicability("electron", 1e-5, 10)};
+        inp.applic = {make_applicability("electron", 1e-5, 1e9)};
```
amandalund: The test failure in `PhysicsStepUtilsTest.calc_mean_energy_loss`:
```console
celeritas/phys/PhysicsStepUtils.hh:268:
celeritas: internal assertion failed: range > step" thrown in the test body.
```
happens with a CGS build too depending on what this upper applicable energy is set to. For some values it passes with:
```console
range=2.0200000000000010e-03, step=2.0200000000000001e-03, eloss=1.0100000000000000e-03, pre_step_energy=1.0000000000000000e-03
```
and for others it fails with:
```console
range=2.0199999999999966e-03, step=2.0200000000000001e-03, eloss=1.0100000000000000e-03, pre_step_energy=1.0000000000000000e-03
```

sethrj: Um. Can we tell where the discrepancy is occurring? I thought we were strict about this even accounting for floating point arithmetic 🤔

amandalund: Just a testing issue I think. The value of the range can change slightly depending on the energy bounds or number of grid points, and in this case it looks like it became slightly less than the step; in `calc_physics_step_limit()` the step limit would have been set equal to the range so we wouldn't have hit that assertion, but this test only calls `calc_mean_energy_loss()`.

I'm assuming this test is meant to test [this case](https://github.com/celeritas-project/celeritas/blob/20d2ba457817d76f607573de757d140c9cc877d1/src/celeritas/phys/PhysicsStepUtils.hh#L270-L272)...

sethrj (approved):
Cool! With this fix can we finally do:
```diff
diff --git a/src/celeritas/phys/PhysicsStepUtils.hh b/src/celeritas/phys/PhysicsSt>
index 05c81e2cd..15d5d0179 100644
--- a/src/celeritas/phys/PhysicsStepUtils.hh
+++ b/src/celeritas/phys/PhysicsStepUtils.hh
@@ -77,7 +77,7 @@ find_ppid(MaterialView const& material,
         // \sigma_{\max} \f$. Note that it's possible for \f$ \sigma(E_1) \f$
         // to be larger than the estimate of the maximum cross section over the
         // step \f$ \sigma_{\max} \f$.
-        if (generate_canonical(rng) * pstep.per_process_xs(ppid) > xs)
+        if (RejectionSampler(xs, pstep.per_process_xs(ppid))(rng))
         {
             // No interaction occurs; reset the physics state and continue
             // tracking
```
or does that fail because of non-monotonic cross sections?

amandalund:
Yeah, that could still fail if the cross sections are not monotonic in the interval and the global maximum is not in the interval.

PR #1778, opened 2025-05-11
Adjust version compatibility for covfie

sethrj:
Following on to #1767: since covfie 0.13 still compiles with CPU-only, and 0.14 isn't yet available (see https://github.com/spack/spack/pull/50417 ), it would be good not to fail to configure on spack environments. Version minimum requirements should reflect a minimum capability. This updates the requirements so that CUDA/HIP need 0.14, but CPU only still works with 0.13. I also failed to update the g4vg spack files when updating in the main CMakelists, so those are now also built.

### DIFF
--- a/.github/workflows/build-spack.yml
+++ b/.github/workflows/build-spack.yml
@@ -131,7 +131,7 @@ jobs:
                  || matrix.geometry == 'vecgeom2'
                  || matrix.geometry == 'vgsurf')
                 && matrix.geant != ''}}; then
-            spack -e . add g4vg@1.0.3:
+            spack -e . add g4vg@1.0.4
           fi
           spack -vd -e . compiler find --mixed-toolchain
           # Add the spack ref so that updating spack will reconcretize
--- a/.github/workflows/build-spack.yml
+++ b/.github/workflows/build-spack.yml
@@ -130,8 +130,8 @@ jobs:
           if ${{(matrix.geometry == 'vecgeom'
                  || matrix.geometry == 'vecgeom2'
                  || matrix.geometry == 'vgsurf')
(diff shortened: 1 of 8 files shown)


esseivaju (approved):
Makes sense. I was planning on requiring v0.14 for CPU too but a bug prevented that so need to wait for the next release. Is it a requirement to have the package on Spack for external dependencies though since they will be downloaded if not found

sethrj:
Our "spack" workflow build should include everything directly from spack, I'm a little confused why it didn't fail with the covfie 0.14 upgrade... I'll look into it. The problem I encountered was that my spack environment already has covfie, and so I would have to switch from an external to a builtin whenever the code upgrades past what spack has availble, which is less than ideal.

.github/workflows/build-spack.yml
```diff
@@ -96,8 +97,9 @@ jobs:
       ### ENVIRONMENT ###
 
       - name: Check out Spack
-        uses: spack/setup-spack@5792f7c7055c3707819380b5e3831d2be6e64b6c # 2024/12
+        uses: sethrj/setup-spack@200141be54665bd75af4dbb131d08d24ea606a9c
         with:
+          repository: celeritas-project/spack
```
esseivaju, line 102: Is this temporary because Spack doesn't have g4vg@1.0.4 yet, or do we want to use our own fork of Spack. If so, maybe we should transfer the `sethrj/setup-spack` to Celeritas organization?

sethrj: I'm just waiting for https://github.com/spack/setup-spack/pull/19 to get merged in... hopefully won't take long!

PR #1779, opened 2025-05-11
Add tests and documentation for Tsai nuclear screening functions

sethrj:
This moves some screening functions out of the internals of the relative brems calculator. It documents the units, demonstrates exactly what equations the classes implement, and adds testing.

### DIFF
--- a/src/celeritas/em/xs/ScreeningFunctions.hh
+++ b/src/celeritas/em/xs/ScreeningFunctions.hh
@@ -49,8 +49,9 @@ struct BhwlScreeningFactors
  * The calculator argument is the fraction \f[
  * \delta = \frac{k}{E(k - E)} \equiv \frac{2\delta_\mathrm{Tsai}}{m_e}
  * \f]
- * where \f$k\f$ is the kinetic plus rest mass energy of the incident electron
- * and \f$E\f$ is the exiting gamma energy.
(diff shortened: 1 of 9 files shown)


src/celeritas/em/xs/ScreeningFunctions.hh
```diff
@@ -0,0 +1,153 @@
...
+ * This calculates atomic screening factors given by  \citet{tsai-1974,
+ * https://doi.org/10.1103/RevModPhys.46.815} Eq. 3.30-31, as part of the
+ * relativistic bremsstrahlung cross section calculation.
+ * This model is valid for \f$ Z \ge 5 \f$.
+ *
+ * The calculator argument is the fraction \f[
+ * \delta = \frac{k}{E(k - E)} \equiv \frac{2\delta_\mathrm{Tsai}}{m_e}
+ * \f]
+ * where \f$k\f$ is the kinetic plus rest mass energy of the incident electron
+ * and \f$E\f$ is the exiting gamma energy.
```
amandalund: Should it be the other way around (k is the photon energy and E is the total energy  of the incident electron)?

amandalund (approved):
Nice documentation!

src/celeritas/em/xs/ScreeningFunctions.hh
sethrj: Yep, good catch!

PR #1781, opened 2025-05-11
Use correct types for physics grids

amandalund:
Currently all physics grids use cross section grid data even though only macroscopic cross sections can have both scaled and unscaled values. This updates the physics data and classes so that energy loss, range, and microscopic cross sections use uniform grids instead.


sethrj (approved):
Nice! We should keep in mind how to eventually allow different grid types for cross sections (though I suppose range and energy loss might always be best in log E?)

PR #1783, opened 2025-05-13
Create IntersectRegion for a convex extruded polygon

elliottbiondo:
This MR creates an `IntersectRegion` for a convex, extruded polygon, with scaling along the z-axis. This is the next step toward supporting Geant4 `xtru` solids, which will consist of unions of these extruded polygons over all z-levels, with non-convex support added by subtracting out concave regions. At first glance it appears that extruded polygons could be used to model Geant4 `GenTrap`s, but this is not the case, as the latter allows for the top and bottom quadrilaterals to have fundamentally different shapes.


src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +535,141 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+{
+    CELER_VALIDATE(polygon_.size() >= 3,
+                   << "polygon must consist of at least 3 vertices");
+    CELER_VALIDATE(scaling_factors[0] > 0 && scaling_factors[1] > 0,
+                   << "scaling factors must be positive");
+    CELER_VALIDATE(line_segment[BOT][Z] < line_segment[TOP][Z],
+                   << "line segment must begin with lower z value");
+
+    // Calculate min/max x/y values, used as both characteristic lengths for
+    // tolerfance considerations, and bounding box surface locations
```
mrguilima:
```suggestion
    // tolerance considerations, and bounding box surface locations
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +535,141 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+void ExtrudedPolygon::build(IntersectSurfaceBuilder& insert_surface) const
+{
+    // Insert the upper and lower bounding planes
+    insert_surface(Sense::outside, PlaneZ{line_segment_[BOT][Z]});
+    insert_surface(Sense::inside, PlaneZ{line_segment_[TOP][Z]});
+
+    // Insert all vertical bounding planes
+    for (auto i : range(polygon_.size()))
+    {
+        // Specify points in such an order that guarentees and outward-facing
```
mrguilima:
```suggestion
        // Specify points in such an order that guarantees an outward-facing
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +535,141 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+//---------------------------------------------------------------------------//
+/*!
+ * Whether this encloses another extruded polygon
+ *
+ * TODO: *extremely* difficult
+ */
+bool ExtrudedPolygon::encloses(ExtrudedPolygon const& other) const
+{
+    CELER_NOT_IMPLEMENTED("ExtrudedPolygon::encloses");
+}
```
mrguilima: If _this_ is a convex polygon, just check that all points from the _other_ polygon are inside _this_ one.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,74 @@ class EllipticalCylinder final : public IntersectRegionInterface
     real_type hh_;
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Region formed by extruding + scaling a convex polygon along a line segment.
+ *
+ * The convex polygon is supplied as a set of points on the XY plane in
+ * clockwise order. The line segment is supplied as a length 2 array of 3D
+ * points denoting the extrusion path travesed by the polygon's origin,
```
mrguilima:
```suggestion
 * points denoting the extrusion path traversed by the polygon's origin,
```

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,74 @@ class EllipticalCylinder final : public IntersectRegionInterface
...
+ * lower polygons.
+ */
+class ExtrudedPolygon final : public IntersectRegionInterface
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using VecReal2 = std::vector<Real2>;
+    using ArrayReal3 = std::array<Real3, 2>;
+    using ArrayReal = std::array<real_type, 2>;
```
mrguilima: Why using std::array here rather than celeritas::Array<>?
In ConvexHullFinder.hh, Real2 is defined as celeritas::Array<real_type, 2>, can't that definition be reused instead of ArrayReal?

src/orange/surf/Plane.hh
```diff
@@ -117,6 +122,23 @@ CELER_FUNCTION Plane::Plane(Real3 const& n, real_type d) : normal_{n}, d_{d}
     CELER_EXPECT(is_soft_unit_vector(normal_));
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from three points.
+ *
+ * The direciton of the normal is dictated by the right-hand rule, assuming
```
mrguilima:
```suggestion
 * The direction of the normal is dictated by the right-hand rule, assuming
```

mrguilima (commented):
A first review.

elliottbiondo:
@mrguilima this is still a draft MR, but I will address your comments.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Yeah sorry, originally this was going to be for an arbitrary polygon, but with convexity it becomes pretty easy

elliottbiondo: It's easy, but you do have to actually build the solid first, which feels like the wrong thing to do here. Since `encloses` is just part of the template interface and is currently only used when creating excluded interior regions---which are not applicable for `xtru`---I am going to leave this unimplemented.

elliottbiondo:
This is going to remain as draft for now. I realize that I should actually be discarding colinear points.

elliottbiondo:
@mrguilima @sethrj this is ready to review

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,143 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+    auto colinear_mask
+        = detail::make_colinear_mask(make_span(polygon), abs_tol);
+    CELER_ASSERT(colinear_mask.size() == polygon.size());
+    for (auto i : range(polygon.size()))
+    {
+        if (!colinear_mask[i])
+        {
+            polygon_.push_back(polygon[i]);
+        }
+    }
```
sethrj: I think an [erase remove idiom](https://en.wikipedia.org/wiki/Erase–remove_idiom) might be better than a two-pass filter. Supposing that your tolerance were somewhat loose and you made a circle of points where each triplet is collinear, the current implementation would delete all the points. An algorithm that updates the list continuously as it's processing would still keep a rough circle shape (and as a bonus will avoid the extra vector creation and a second pass).

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,82 @@ class EllipticalCylinder final : public IntersectRegionInterface
...
+    {
+        BOT = 0,
+        TOP = 1
+    };
+
+    //// DATA ////
+
+    VecReal2 polygon_;
+    ArrayReal3 const& line_segment_;
+    ArrayReal const& scaling_factors_;
```
sethrj: Definitely don't use references here! The solids *always* have a lifetime larger than their construction argument.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,82 @@ class EllipticalCylinder final : public IntersectRegionInterface
...
+
+  private:
+    //// TYPES ////
+
+    // Enum for differenting between the bottom and top instance of the polygon
+    enum
+    {
+        BOT = 0,
+        TOP = 1
+    };
```
sethrj: Avoid all-caps symbol names. Use the existing `enum class Bound` in corecel/grid/GridTypes.hh (an odd place for it, we could move that...) and as a shorthand you can locally use `constexpr auto bot = Bound::lo`. (And you would define `Real3Bound = EnumArray<Bound, Real3>` and `RealBound = EnumArray<Bound, real_type>` to store the data under the hood.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,82 @@ class EllipticalCylinder final : public IntersectRegionInterface
...
+ * polygon is linearly scaled in accordance with scaling factors supplied for
+ * each line segment point.
+ *
+ * As is done in Geant4, construction is done by first applying scaling factors
+ * to the upper and lower polygons via scalar multiplication with each polygon
+ * point, then the points on the line are used to offset the upper and lower
+ * polygons.
+ *
+ * The line_segment and scaling parameters are stored as const references and
+ * must have lifetimes longer than this class.
```
sethrj: Not gonna happen 😅 besides, they're just a few doubles so it's not harmful to copy.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,143 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+    x_range_[1]
+        = std::max(poly_max_x * scaling_factors_[BOT] + line_segment_[BOT][X],
+                   poly_max_x * scaling_factors_[TOP] + line_segment_[TOP][X]);
+    y_range_[0]
+        = std::min(poly_min_y * scaling_factors_[BOT] + line_segment_[BOT][Y],
+                   poly_min_y * scaling_factors_[TOP] + line_segment_[TOP][Y]);
+    y_range_[1]
+        = std::max(poly_max_y * scaling_factors_[BOT] + line_segment_[BOT][Y],
+                   poly_max_y * scaling_factors_[TOP] + line_segment_[TOP][Y]);
+}
```
sethrj, line 663: Make this a const function that returns `T_range` for T in  {X, Y} (given as an argument) and just call it twice in the constructor:
```
x_range_ = calc_ranges(polygon, X);
y_range_ = calc_ranges(polygon, Y);
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,143 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+                  + line_segment_[BOT];
+
+        insert_surface(Sense::inside, Plane{p0, p1, p2});
+    }
+
+    // Insert additional planes for bounding box creation
+    insert_surface(Sense::outside, PlaneX{x_range_[0]});
+    insert_surface(Sense::inside, PlaneX{x_range_[1]});
+    insert_surface(Sense::outside, PlaneY{y_range_[0]});
+    insert_surface(Sense::inside, PlaneY{y_range_[1]});
```
sethrj: Don't add unused surfaces just for the bbox; the `insert_surface(Sense, BBox)` form should be used (which just adjusts the bbox, not creates any surfaces):
```
    constexpr auto inf = numeric_limits<real_type>::infinity();
    insert_surface(Sense::inside,
                   BBox::from_unchecked({x_range_[0], y_range_[0], -inf}, {x_range_[1], y_range_[1], inf});
```

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -239,6 +239,82 @@ class EllipticalCylinder final : public IntersectRegionInterface
...
+    using VecReal2 = std::vector<Real2>;
+    using ArrayReal3 = celeritas::Array<Real3, 2>;
+    using ArrayReal = celeritas::Array<real_type, 2>;
+    //!@}
+
+  public:
+    // Construct from a convex polygon, line segment, and scaling factors
+    ExtrudedPolygon(VecReal2 const& polygon,
+                    ArrayReal3 const& line_segment,
+                    ArrayReal const& scaling);
```
sethrj: The `ArrayReal3` and `ArrayReal` definition is not obvious from the names: the second is really a Real2 but it's a lo/hi rather than x/y. I think maybe use an approach like `GenPrism`: combine segment + scaling into a struct, and pass two of those to the constructor: `lo` and `hi`. Then your internal data structures can use `EnumArray<Bound,  real_type>` etc.

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +149,28 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+    real_type tol = 0.01;
+
+    // Nothing colinear
+    VecReal2 corners{{0, 0}, {0, 1}, {1, 1}, {1, 0}};
+    expect_eq(corners, make_colinear_mask(make_span(corners), tol));
+
+    // Nothing colinear
+    VecReal2 corners{{0, 0}, {0, 1}, {1, 1}, {1, 0}};
+    expect_eq(corners, make_colinear_mask(make_span(corners), tol));
+}
```
sethrj, line 275: An interesting test to add would be to have a very wide tolerance and generate 1000 points on a circle and see what they filter down to.

elliottbiondo: Yep I rewrote the code yesterday to handle this case. Will add.

elliottbiondo:
@sethrj I am also going to change up what it means to be soft colinear. For hard collinearity, the cross product makes sense because only the sign matters. For soft collinearity, we should be comparing α to an absolute tolerance (i.e., the bump distance), where α is:

Screenshot 2025-05-21 at 11 59 42 AM

sethrj:
Good point!

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: Sure --- but then what should I do for the accessors? If we only store EnumArrays (which I see the value in to catch indexing mistakes) then the accessors would have to reconstruct the `PolyFaces`, which seems messy. Returning an EnumArray seems bad too.

elliottbiondo:
@mrguilima @sethrj this should be ready

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -150,6 +192,55 @@ inline bool is_convex(Span<Real2 const> corners, bool degen_ok = false)
     return true;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Return a mask that is true if a point is colinear with its 2 neighbors.
+ *
+ * Points are checked for colinearly dynamically, i.e, if a point is
```
mrguilima:
```suggestion
 * Points are checked for colinearity dynamically, i.e, if a point is
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -150,6 +192,55 @@ inline bool is_convex(Span<Real2 const> corners, bool degen_ok = false)
...
+ * found to be colinear, it is not used for future colinearity checks.
+ */
+std::vector<Real2>
+filter_collinear_points(std::vector<Real2> const& corners, double abs_tol)
```
mrguilima, line 216: Both colinear and collinear (single or double L) are correct, but one should stick to only one of them.  I prefer colinear (no double L) myself...

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,140 @@ void EllipticalCone::output(JsonPimpl* j) const
...
 
+//---------------------------------------------------------------------------//
+// ExtrudedPolygon
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from a convex polygon, line segment, and scaling factors.
+ */
+ExtrudedPolygon::ExtrudedPolygon(ExtrudedPolygon::VecReal2 const& polygon,
+                                 ExtrudedPolygon::PolygonFace bot_face,
+                                 ExtrudedPolygon::PolygonFace top_face)
```
mrguilima:
```suggestion
                                 ExtrudedPolygon::PolygonFace const& bot_face,
                                 ExtrudedPolygon::PolygonFace const& top_face)
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,140 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+{
+    CELER_EXPECT(dir == X || dir == Y);
+
+    constexpr auto bot = Bound::lo;
+    constexpr auto top = Bound::hi;
+
+    Range range;
+
+    // Find min/max x/y values of the polygon itself
+    auto [poly_min_x_it, poly_max_x_it] = std::minmax_element(
```
mrguilima: Since x/y is given by dir, probably should drop the _x_ in variable names like poly_min_x_it and others.

src/orange/orangeinp/ObjectIO.json.cc
```diff
@@ -197,6 +197,14 @@ void to_json(nlohmann::json& j, EllipticalCone const& cr)
          SIO_ATTR_PAIR(cr, upper_radii),
          SIO_ATTR_PAIR(cr, halfheight)};
 }
+void to_json(nlohmann::json& j, ExtrudedPolygon const& cr)
+{
+    j = {
+        {"_type", "extrudedpolygon"}, SIO_ATTR_PAIR(cr, polygon),
+        // SIO_ATTR_PAIR(cr, line_segment),
+        // SIO_ATTR_PAIR(cr, scaling_factors)
```
mrguilima: Are these supposed to be commendted out?

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +83,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
+ *     .  t     .        .
+ *   a . . . . . . . . . . c
+ *
+ * The displacement is calculated as follows.
+ *
+ * Let:
+ * u = b - a
+ * v = c - a
+ *
+ * In 2D, the cross product is can be written as,
```
mrguilima:
```suggestion
 * In 2D, the cross product can be written as,
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +83,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
+ *
+ * u x v = |u| |v| sin(t),
```
mrguilima: LaTeX-like formatting should be used for math equations in comments:
```suggestion
 * \f[
 *  \vec u \times \vec v = |\vec u| |\vec v| \sin(t),
 * \f]
```
See src/celeritas/field/MagFieldEquation.hh for examples.

src/orange/surf/Plane.hh
```diff
@@ -117,6 +122,33 @@ CELER_FUNCTION Plane::Plane(Real3 const& n, real_type d) : normal_{n}, d_{d}
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from three points.
+ *
+ * The direction of the normal is dictated by the right-hand rule, assuming
+ * the normal vector, C, is given by:
+ * A = (p1 - p0),
+ * B = (p2 - p0),
+ * C = A x B.
```
mrguilima:
```suggestion
 * \f[
 * \vec A = (\vec{p_1} - \vec{p_0}) ,
 * \vec B = (\vec{p_2} - \vec{p_0}) ,
 * \vec C = \vec A \times \vec B .
 * \f]
```

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +162,118 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+    expect_polygon_equal(exp, filter_collinear_points(points, tol));
+}
+
+/* Test pathological case consisting of a many-sided regular polygon with every
+ * point locally soft collinear with its neighbors due to a large tolerance.
+ */
+TEST(PolygonUtilsTest, filter_collinear_points_pathological)
+{
+    // Create a many-sided regular polygon by placing 20 equally spaced points
+    // on a circule of radius = 1, in clockwise order.
```
mrguilima:
```suggestion
    // on a circle of radius = 1, in clockwise order.
```

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +162,118 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+    }
+
+    // Verify choose a tolerance such that the points are locally soft
+    // collinear
+    real_type tol = 0.1;
+    SoftOrientation soft_ori(tol);
+    EXPECT_EQ(Orientation::collinear,
+              soft_ori(points[0], points[1], points[2]));
+
+    // Check that filtering gives us more that zero points, in this case 7
```
mrguilima:
```suggestion
    // Check that filtering gives us more than zero points, in this case 7
```

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +162,118 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+    VecReal2 points(20);
+    real_type step = static_cast<real_type>(2 * constants::pi / points.size());
+
+    for (auto i : range(points.size()))
+    {
+        real_type theta = -step * i;
+        points[i] = {std::cos(theta), std::sin(theta)};
+    }
+
+    // Verify choose a tolerance such that the points are locally soft
```
mrguilima:
```suggestion
    // Choose a tolerance such that the points are locally soft
```

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +162,118 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+
+/* Test pathological case consisting of a many-sided regular polygon with every
+ * point locally soft collinear with its neighbors due to a large tolerance.
+ */
+TEST(PolygonUtilsTest, filter_collinear_points_pathological)
+{
+    // Create a many-sided regular polygon by placing 20 equally spaced points
+    // on a circule of radius = 1, in clockwise order.
+    VecReal2 points(20);
+    real_type step = static_cast<real_type>(2 * constants::pi / points.size());
```
mrguilima, line 258: At this point, the cast is redundant.
```suggestion
    real_type step = 2 * constants::pi / static_cast<real_type>(points.size());
```

mrguilima (changes requested):
Thanks for the implementation!

test/orange/orangeinp/detail/PolygonUtils.test.cc
```diff
@@ -149,6 +162,117 @@ TEST(PolygonUtilsTest, convex_self_intersect)
...
+ *    .               |           .
+ *   .                |          .
+ *   .                |         .
+ *   b________________a____g__f_________
+ *   |        |       |       |        |
+ *  -1       -0.5     0      0.5       1
+ */
+TEST(PolygonUtilsTest, filter_collinear_points)
+{
+    // Point locations, as labled above
```
mrguilima:
```suggestion
    // Point locations, as labeled above
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +87,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
+ *         .    .    .
+ *       .      . d    .
+ *     .  t     .        .
+ *   a . . . . . . . . . . c
+ *
+ * The displacement is calculated as follows.
+ *
+ * Let:
+ * u = b - a
+ * v = c - a
```
mrguilima:
```suggestion
 * Let: \f[
 * \vec u = \vec b - \vec a
 * \vec v = \vec c - \vec a
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +87,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
+ *
+ * In 2D, the cross product is can be written as,
+ *
+ * u x v = |u| |v| sin(t),
+ *
+ * noting that this is a different cross product (different vectors) compared
+ * to the cross product used for orientation determination. Geometrically, the
+ * displacement can be calculated as,
+ *
+ * d = |u| sin(t).
```
mrguilima:
```suggestion
 * d = |vec u| sin(t).
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +87,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
+ *
+ * Therefore,
+ *
+ * d = |u| (u x v) / (|u| |v|)
+ *   = (u x v)/|v|.
```
mrguilima:
```suggestion
 * \f[
 * d = |\vec u| (\vec u \times \vec v) / (|\vec u| |\vec v|)
 *   = (\vec u \times \vec v)/|\vec v|.
 * \f]
```

mrguilima (dismissed):
More LaTeX suggestions.

src/orange/orangeinp/detail/PolygonUtils.hh
sethrj: Needs to be either two separate equations or use the split environment. Also I'm not too worried about equation rendering in detail classes; it's most important for classes rendered in the user manual.

src/orange/orangeinp/ObjectIO.json.cc
elliottbiondo: This is stale; it was resolved in a subsequent commit.

test/orange/orangeinp/detail/PolygonUtils.test.cc
elliottbiondo: (also stale)

test/orange/orangeinp/detail/PolygonUtils.test.cc
elliottbiondo: (also stale)

test/orange/orangeinp/detail/PolygonUtils.test.cc
elliottbiondo: (also stale)

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -316,6 +318,97 @@ class EllipticalCone final : public IntersectRegionInterface
...
+
+    //! Top scaling factor
+    real_type top_scaling_factor() const
+    {
+        return scaling_factors_[Bound::hi];
+    }
+
+  private:
+    //// TYPES ////
+    using Range = celeritas::Array<real_type, 2>;
```
sethrj, line 394: Range/range are confusing given `corecel/cont/Range.hh`... I might've gone with `Real2 lo_, hi_;` for points but whatever, let's just get this in.

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -119,6 +87,84 @@ is_same_orientation(Orientation a, Orientation b, bool degen_ok = false)
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Functor for calculating orientation with a tolerance for collinearity.
+ *
+ * Collinearity is based on a supplied absolute tolerance. For three ordered
+ * points a, b, c, point b is collinear if the displacement, d, is less than
+ * the absolute tolerance.
+ *
```
sethrj: Can you put this (and future ASCII art) in [`\verbatim` blocks](https://www.doxygen.nl/manual/commands.html#cmdverbatim)? Otherwise it will render as mush.

sethrj (approved):
Sorry for prolonging this review!

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: I think using the most appropriate data type (i.e., `celeritas::Array<real_type, 2>`) is more important than coming up with an ideal name for the data type. It's not ideal to give it a name that conflicts with another type elsewhere, but there is an extremely limited number of words that can describe this concept. `Range` and `Span` are taken. Maybe `Interval`.

FWIW I prefer more descriptive type names. As an example elsewhere in this MR, I would prefer call `Array<real_type, 2>` a `LineSegment`. I specifically used `ArrayReal` for this within in the MR based on previous MR comments (https://github.com/celeritas-project/celeritas/pull/1738#discussion_r2044206267).

src/orange/orangeinp/detail/PolygonUtils.hh
elliottbiondo: @sethrj why does this need to be two separate equations?

elliottbiondo:
@mrguilima @sethrj I fixed the equations in `src/orange/surf/Plane.hh` to use LaTeX. For the stuff in `detail` directory, I am actually going to advocate for not using LaTeX. It comes down to how important it is to have nice Doxygen for detail classes. Personally, I never use Doxygen; I always just go into the code, even for projects that I am a developer of (e.g., Geant4). The equations I have in `PolygonUtils` are really "developer equations," describing gritty implementation details, and I would prefer to read them in plain text rather than unrendered LaTeX.

sethrj:
I'm fine with this. Just add verb markup or html pre tags around it so it doesn't mess up the developer documentation which does include details

mrguilima:
> @mrguilima @sethrj The equations I have in `PolygonUtils` are really "developer equations," describing gritty implementation details, and I would prefer to read them in plain text rather than unrendered LaTeX.

Sure.  I didn't know about the 'detail' classes not having their docs rendered, that's why I suggested the LaTeX formatting.

mrguilima (approved):
LGTM, Thanks!

src/orange/orangeinp/IntersectRegion.hh
mrguilima: Actually `Array<real_type, 2>` is equivalent to Real2, it is a 2D-point, not a line segment.

elliottbiondo: Yep, I said "elsewhere in this MR"

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,140 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+
+    // After removing collinear points, at least 3 points must remain
+    CELER_VALIDATE(polygon_.size() >= 3,
+                   << "polygon must consist of at least 3 points");
+
+    // After removing collinear points, the polygon should have a *strictly*
+    // clockwise orientation, which also guarantees it is convex.
+    CELER_VALIDATE(
+        has_orientation(make_span(polygon_), detail::Orientation::clockwise),
+        << "polygon must be specified in strictly clockwise order");
```
sethrj, line 576: Our discussion on https://github.com/celeritas-project/celeritas/pull/1840/files#r2210769390 made me take a closer look at this. You're correct that Geant4 always stores these points as clockwise, but that seems to be backward from general computer graphics algorithms:  it seems like most assume counterclockwise winding of exterior faces. (ChatGPT's API summary table seems to, and all occurrences of "clockwise" in [Ericson's real-time collision detection](https://www.taylorfrancis.com/books/9780080474144) are in the word "counterclockwise".

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -527,6 +528,140 @@ void EllipticalCone::output(JsonPimpl* j) const
...
+    Real3 const extents{
+        x_range_[1] - x_range_[0], y_range_[1] - y_range_[0], 0};
+    real_type abs_tol = ::celeritas::detail::BumpCalculator(
+        Tolerance<>::from_default())(extents);
+
+    polygon_ = detail::filter_collinear_points(polygon, abs_tol);
+
+    // After removing collinear points, at least 3 points must remain
+    CELER_VALIDATE(polygon_.size() >= 3,
+                   << "polygon must consist of at least 3 points");
```
sethrj, line 570: Again based on our discussion about tolerances (https://github.com/celeritas-project/celeritas/pull/1840/files#r2210049221): I think it would be more correct to do this filtering in the `build` function, and only check for exactly collinear/concave points in the constructor. In practice though I think it's fine.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Yeah, maybe we should change it so we do everything counterclockwise, then just reverse the points that Geant4 gives us. I agree that it's not good to inherit design decisions from other projects.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Maybe we should touch base on this. I think the the convex hull code was the first place you suggested calculating an absolute tolerance based on the length scale. In principle, shouldn't we be doing this everywhere? My design decisions have been influenced by "where is the best place to calculate the abs tol." I was viewing collinear points as invalid data, and I don't like storing invalid data on a class. Why not DBC-expect that the supplied polygon is valid? Well, if you bring in the machinery for detecting polygon validity, you might as well just fix in place.

PR #1784, opened 2025-05-14
Simplify nested join nodes

sethrj:
This adds a new "simplifier" that I added almost a year ago but neglected to merge. When adding a "join" node to an existing join with the same operator, it inlines the daughter nodes, turning `all(all(1, 2, 3), all(4, 5))` to `all(1, 2, 3, 4, 5)` which also allows simplifications like `all(all(-1, 2), all(2,-3))` to `all(-1, 2, -3)`. One additional lookup will be needed to simplify `all(all(-1, 2), all(1, 3))` to `false`.

Changes to existing trees/expressions are isolated to 080babfda906a3e8169d113d43ea23ddc3036f09.


esseivaju (approved):
Looks good, and should help a bit with better short-circuiting once we get that in use!

PR #1785, opened 2025-05-14
Store hardwired model data by reference

amandalund:
This updates the physics hardwired models so the model data is stored by reference rather than being duplicated. It also cleans up some of the old model data aliases.

### DIFF
--- a/src/celeritas/em/interactor/CombinedBremInteractor.hh
+++ b/src/celeritas/em/interactor/CombinedBremInteractor.hh
@@ -54,7 +54,7 @@ class CombinedBremInteractor
   public:
     // Construct with shared and state data
     inline CELER_FUNCTION
-    CombinedBremInteractor(CombinedBremRef const& shared,
+    CombinedBremInteractor(NativeCRef<CombinedBremData> const& shared,
                            ParticleTrackView const& particle,
                            Real3 const& direction,
                            CutoffView const& cutoffs,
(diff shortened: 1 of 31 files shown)


src/celeritas/em/interactor/CombinedBremInteractor.hh
```diff
@@ -54,7 +54,7 @@ class CombinedBremInteractor
   public:
     // Construct with shared and state data
     inline CELER_FUNCTION
-    CombinedBremInteractor(CombinedBremRef const& shared,
+    CombinedBremInteractor(NativeCRef<CombinedBremData> const& shared,
```
sethrj, line 57: Since the combined brem executor currently only works for the single element case (is this because we never ported the per-element xs?) should we delete the combined brem interactor?

sethrj (approved):
Thanks!

src/celeritas/em/interactor/CombinedBremInteractor.hh
amandalund: Right, it's because we don't have micro xs for the combined model. From our past discussions it sounded like we didn't see much benefit from the combined model when we tested very early on. We've talked about taking another look once the geometry's sped up and the physics is a more significant fraction of the time, and possibly also testing combined interactors for the muon models. I think we should either build the combined cross sections or remove the model, and given that this is pretty low on our list of performance concerns (and that we could always revert it later) I'm more in favor of the latter.

sethrj: Agreed!

PR #1789, opened 2025-05-18
Organize optical directory

amandalund:
Our optical directory is getting a little cluttered; this starts to organize it a bit by adding a subdirectory for all the classes associated with optical photon generation. I've moved the Cherenkov and scintillation  `Data`, `Params` and `Generator` out of the `optical` namespace (the  `Offload`, `OffloadAction` and `GeneratorAction` were already in the `celeritas` namespace). I also cleaned up some documentation here and there.


sethrj (approved):
Nice improvement! It's good to have the intermediate generation layer separate from the optical loop internals. Heads up @hhollenb when merging develop after this.

PR #1791, opened 2025-05-20
Fix `sincospi` on device

amandalund:
`sincospi` was producing garbage values on device: it looks like the only place it's currently used on the GPU is calculating the polarization in the scintillation generator, which wasn't previously checked. This fixes it and adds a device algorithms test.

Not directly related, but while working through this I came across this note in the [CUDA Math API](https://docs.nvidia.com/cuda/cuda-math-api/index.html):

> Note also that due to implementation constraints, certain math functions from std:: namespace may be callable in device code even via explicitly qualified std:: names. However, such use is discouraged, since this capability is unsupported, unverified, undocumented, not portable, and may change without notice.

Is this something we should worry about?

### DIFF
--- a/src/celeritas/optical/ParticleTrackView.hh
+++ b/src/celeritas/optical/ParticleTrackView.hh
@@ -81,6 +81,8 @@ CELER_FUNCTION ParticleTrackView&
 ParticleTrackView::operator=(Initializer const& init)
 {
     CELER_EXPECT(init.energy >= zero_quantity());
+    CELER_EXPECT(is_soft_unit_vector(init.polarization));
+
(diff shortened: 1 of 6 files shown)


sethrj (approved):
Yikes, good catch. And although I could have sworn that the CUDA guide said that all of the math functions also live in `std::` I can't find it now 😓

Maybe we need to globally replace all the `std::` math functions to be safe, but also include `using std::sqrt` inside the celeritas namespace...

amandalund:
Hmm yeah I haven't been able to find any other mention of it in the docs either...

PR #1792, opened 2025-05-20
Complete optical stepping loop

amandalund:
This fills in the rest of the pre-step and along-step actions in the optical loop, enabling the volumetric models. It also fixes a few assertion failures that showed up in the process.

### DIFF
--- a/src/celeritas/optical/interactor/RayleighInteractor.hh
+++ b/src/celeritas/optical/interactor/RayleighInteractor.hh
@@ -92,13 +92,16 @@ CELER_FUNCTION Interaction RayleighInteractor::operator()(Engine& rng) const
         else
         {
             // Project polarization onto plane perpendicular to new direction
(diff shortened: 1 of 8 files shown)


src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -92,13 +92,16 @@ CELER_FUNCTION Interaction RayleighInteractor::operator()(Engine& rng) const
         else
         {
             // Project polarization onto plane perpendicular to new direction
-            new_pol = make_unit_vector(inc_pol_ - projected_pol * new_dir);
+            new_pol = inc_pol_ - projected_pol * new_dir;
+            new_pol = make_unit_vector(
+                new_pol - dot_product(new_pol, new_dir) * new_dir);
```
sethrj: I didn't notice in the original but these should use axpy instead of operators and temporaries

sethrj (approved):
Nice! Thank you.

PR #1793, opened 2025-05-21
Add make_orthogonal helper function

sethrj:
Following on to a minor change in #1792 this consolidates orthogonalization into a helper function. It also removes a (incorrectly written, my bad on the review) no longer necessary check in the rayleigh scattering.


src/celeritas/optical/gen/ScintillationGenerator.hh
```diff
@@ -159,8 +159,7 @@ ScintillationGenerator::operator()(Generator& rng)
             pol[j] = cosphi * pol[j] + sinphi * perp[j];
         }
         // Enforce orthogonality
-        axpy(-dot_product(pol, photon.direction), photon.direction, &pol);
-        return make_unit_vector(pol);
+        return = make_unit_vector(make_orthogonal(pol, photon.direction));
```
amandalund:
```suggestion
        return make_unit_vector(make_orthogonal(pol, photon.direction));
```

src/orange/MatrixUtils.cc
```diff
@@ -134,20 +134,14 @@ void orthonormalize(SquareMatrix<T, N>* mat)
...
+            cur = make_orthogonal(cur, (*mat)[ip]);
         }
 
         // Normalize row
-        T inv_mag = 1 / norm(cur);
-        for (size_type j = 0; j != N; ++j)
-        {
-            cur[j] *= inv_mag;
-        }
+        cur[j] = make_unit_vector(cur[j]);
```
amandalund:
```suggestion
        cur = make_unit_vector(cur);
```

src/corecel/math/ArrayUtils.hh
```diff
@@ -149,6 +155,28 @@ CELER_FUNCTION Array<T, N> make_unit_vector(Array<T, N> const& v)
...
+//---------------------------------------------------------------------------//
+/*!
+ * Return the component of \em x that is orthogonal to the unit vector \em y.
+ *
+ * In this implementation, \em y must be normalized, and the result is normal.
+ *
+ * \f[
+  \mathbf{x}' \gets
+ \mathbf{x} - \frac{\mathbf{x} \cdot \mathbf{y}}{\mathbf{y} \cdot \mathbf{y}}
+\mathbf{y}
```
amandalund: Maybe just to be explicit that this won't normalize y:
```suggestion
  \mathbf{x}' \gets \mathbf{x} - (\mathbf{x} \cdot \mathbf{y}) \mathbf{y}, \quad \|\mathbf{y}\| = 1
```

src/corecel/random/distribution/RejectionSampler.hh
```diff
@@ -85,8 +84,8 @@ CELER_FUNCTION
 RejectionSampler<RealType>::RejectionSampler(real_type f, real_type fmax)
     : f_{f}, fmax_{fmax}
 {
-    CELER_EXPECT(f_ >= 0);
-    CELER_EXPECT(fmax_ >= f_);
+    CELER_EXPECT(!(f_ < 0));  // Written to allow f=NaN
```
amandalund: We might want to test this case.

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -80,25 +80,19 @@ CELER_FUNCTION Interaction RayleighInteractor::operator()(Engine& rng) const
     do
     {
         new_dir = sample_direction(rng);
+        // Project polarization onto plane perpendicular to new direction.
+        // In the unlikely case that the incident polarization and new
+        // direction are coincident, make_orthogonal will return a zero vector,
+        // and all elements of new_pol will be NaN, which will reject the
+        // sample.
```
amandalund: Won't `nan < fmax * u` return false and accept the sample?

amandalund (commented):
Thanks! The helper is nice, just a question about propagating nans to the rejection sampler.

src/corecel/math/ArrayUtils.hh
sethrj: Oh yep indeed.

src/celeritas/optical/interactor/RayleighInteractor.hh
sethrj: Yep, "false" means "reject" rather than "accept" 🤪  I've allowed it and updated the rayleigh sampler.

sethrj:
Incidentally I tested the new `make_orthogonal` implementation of the matrix orthonormalize, and it has the same instruction count (but just a very slightly different instruction ordering).

src/celeritas/optical/interactor/RayleighInteractor.hh
amandalund: Does it make sense to always accept nan? I think rejecting it is reasonable, but I can't think of a case where we'd want to accept it as a sample.

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -80,25 +80,21 @@ CELER_FUNCTION Interaction RayleighInteractor::operator()(Engine& rng) const
     do
     {
         new_dir = sample_direction(rng);
-
-        auto projected_pol = dot_product(new_dir, inc_pol_);
-
-        if (soft_zero(projected_pol))
+        if (CELER_UNLIKELY(dot_product(inc_pol_, new_dir) == 0))
```
amandalund: Previously we used `soft_zero`, not sure if that makes a difference?

src/celeritas/optical/interactor/RayleighInteractor.hh
```diff
@@ -80,25 +80,21 @@ CELER_FUNCTION Interaction RayleighInteractor::operator()(Engine& rng) const
...
-            // If new direction is parallel to incident polarization, then
-            // randomly sample azimuthal direction
-            new_pol = ExitingDirectionSampler{0, new_dir}(rng);
+            // In the edge case that the incident polarization and new
+            // direction are coincident, reject the sample
+            continue;
         }
-        else
+        // Project polarization onto plane perpendicular to new direction
+        new_pol = make_unit_vector(make_orthogonal(inc_pol_, new_dir));
```
amandalund: Also back to seeing this assertion failure (in the optical collector test with a single precision build):
```console
celeritas/optical/interactor/RayleighInteractor.hh:105:
celeritas: postcondition failed: soft_zero(dot_product(new_pol, new_dir))
```

src/celeritas/optical/interactor/RayleighInteractor.hh
sethrj: We did, though Geant4 uses a hard zero (which seems even more "edge"y than I'd think necessary). My thought was that rather than eliminating a whole range of directions, we could eliminate just one (the renormalization would result in some loss of precision but should work??) or rely on nan.

src/celeritas/optical/interactor/RayleighInteractor.hh
sethrj: OK so perhaps even though it was orthogonalizing, it couldn't be both a unit normal and orthogonal within precision. That makes sense. I've updated the rejection so that *this* is the condition we require, rather than rejecting based on the input values.

src/celeritas/optical/interactor/RayleighInteractor.hh
sethrj: Accepting it "propagates" the value, which some of our algorithms (e.g. clamp/min/max do) do to make sure we don't hide errors. In contrast some functions such as fmax ignore nans (so `f(nan, x) -> x`). I've reverted the behavior so we throw on it.

sethrj:
Aw *shit*, `continue` inside a `do while` doesn't jump *up* to the `do`, it jumps *down* to the `while`! 😨 How many bugs have I written/missed because of this?

amandalund:
Oops 😅 I didn't catch that

amandalund (approved):
Thanks @sethrj!

PR #1794, opened 2025-05-22
Move core track initialization to `CoreTrackView`

amandalund:
Similar to our optical loop, this moves the core track initialization code from the executor into an `operator=` in  the `CoreTrackView`. It also stores the parent ID for "detailed" geometry initialization directly in the `TrackInitializer` instead of a separate state collection in the `TrackinitData`. This means we'll be allocating `capacity` rather than `state` number of IDs, but it should be the direction we're headed with storing unique volume instance IDs for fast geometry initialization.

### DIFF
--- a/src/geocel/Types.hh
+++ b/src/geocel/Types.hh
@@ -85,6 +85,11 @@ struct GeoTrackInitializer
     {
         return dir[0] != 0 || dir[1] != 0 || dir[2] != 0;
     }
+
+    // Constructors
+    inline CELER_FUNCTION GeoTrackInitializer();
+    inline CELER_FUNCTION GeoTrackInitializer(Real3, Real3);
+    inline CELER_FUNCTION GeoTrackInitializer(Real3, Real3, TrackSlotId);
 };
 
 //---------------------------------------------------------------------------//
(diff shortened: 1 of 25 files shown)


src/geocel/Types.hh
```diff
@@ -77,6 +78,7 @@ struct GeoTrackInitializer
 {
     Real3 pos{0, 0, 0};
     Real3 dir{0, 0, 0};
+    TrackSlotId parent;
```
sethrj, line 81: Given how this is being used (i.e., it's not actually used by `operator=(Initializer_t const&)`, but rather used by the core track to construct `AdvancedInitializer const&`) I don't think it belongs here. (Unless the usage in core track view is "temporary" in which case it might need a to-do. I think a change that would "fix" this problem is to remove the "advanced initializer" nested structs entirely and instead have the GeoTrackInitializer directly pull from the other state. The downside is that would require keeping a pointer to the whole state vector rather than just the local data.

sethrj (commented):
This is sensible. One thing that I think should change is to define an explicit constructor for the `GeoTrackInitializer`, because the updated tests (besides being numerous) are a bit less readable.

src/geocel/Types.hh
amandalund: Right, my understanding was we're hoping to eventually replace this optimization with a fast initialization of the geometry state, and whatever information we need for that would live here. I'll try removing the detailed initializers and see how that looks.

amandalund: Does that seem any better? If not I can revert it and just move the parent ID up a level in the track initializer.

sethrj: Looks good to me, thanks!

test/orange/Orange.test.cc
```diff
@@ -73,7 +73,7 @@ TEST_F(OneVolumeTest, track_view)
     EXPECT_FALSE(geo.is_on_boundary());
 
     // Initialize from a pre-existing OrangeTrackView object
-    geo = OrangeTrackView::DetailedInitializer{geo, Real3({1, 0, 0})};
+    geo = Initializer_t{geo.pos(), Real3{1, 0, 0}, TrackSlotId{0}};
```
sethrj, line 76: Oh yeah, I forgot if we're using the track slot ID we don't need the position... future work I guess!

sethrj (approved):
LGTM!

test/orange/Orange.test.cc
amandalund: Oh good point... yeah I guess I'll leave it for now, for debugging at least it's helpful to assert that the parent's position matches.

PR #1795, opened 2025-05-23
Fix assigning parent ID to track initializer

amandalund:
This fixes a dumb mistake in #1794 where the parent ID wasn't actually being set in the track initializer. I also forgot we do  allow copying the parent geometry when partitioning tracks by charge, just not for in-place initialization. This also simplifies the `ProcessSecondariesExecutor` a bit by using `CoreTrackView`.


sethrj (approved):
Looks good, but are there any quick tests that could be added which caught the original failure?

amandalund:
I added a simple check in the track init test, which also caught that the ID wasn't default-initialized to invalid when inserting primaries.

Issue #1749, opened 2025-04-17
Map ORANGE volumes to G4 VolumeId/VolumeInstanceId


sethrj:
Needed for optical physics (#1351 ); we can save the g4 properties during the `g4org` construction. The `VolumeId` in ORANGE needs to be changed from an unfolded Universe+Volume ID to arrays mapping {universe ID, local volume ID} -> {VolumeId, PlacedVolumeId}.

- Rename VolumeId to InternalVolumeId (universe indexed) inside ORANGE and/or delete it entirely, but check SCALE first
- Add device params data: `Collection<PhysicalVolumeId, ..., LocalVolumeId>` (i.e. storing the volume ID for each local volume inside each universe), and update the track view accessors to use them to return the VolumeId and PhysicalVolumeId.(We could do another VolumeId array for now too
- Make `VolumeInput::label` a `variant<VolumeId, Label>` and add a preprocessing step to create internal VolumeParams from the labels and universe hierarchy?

PR #1796, opened 2025-05-25
Keep persistent GeantGeoParams for geometry translation

sethrj:
Toward https://github.com/celeritas-project/celeritas/issues/1749 and optical physics, this changes the internals of the code so we always have a global GeantGeoParams to handle mapping of volume IDs, instance IDs, and (soon) surface IDs.


amandalund (approved):
Looks great, thanks!

PR #1798, opened 2025-05-28
Move magnetic field retrieval to a helper function

sethrj:
This tiny PR defines a more robust helper function for getting the magnetic field from the transportation field manager.


esseivaju (approved):
Looks good!

PR #1800, opened 2025-05-29
Refactor optical generation to reduce duplication

amandalund:
This is a first step in refactoring the optical generators. The main changes are:

  - Eliminate the duplication in the offload and generator actions
  - Store the distribution buffers as aux data in the generator action. ~This should allow us to move the generation from the core to the optical loop~ (to be done in a follow on). Actually it won't since we can't access any aux data without the core state, so we'll still need a way to access the distribution data.
  - Remove the `OffloadParams` and store the step data as aux data in the gather action


src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -41,8 +40,7 @@ auto get_core_sizes(OpticalLaunchAction const& ola)
 
     // NOTE: quantities are *per-process* quantities: integrated over streams,
     // but not processes
-    result.generators = result.streams
-                        * ola.offload_params().host_ref().setup.capacity;
+    //! \todo Store generator sizes
```
amandalund, line 43: Will restore this in the follow on.

src/celeritas/optical/OpticalCollector.cc
```diff
@@ -152,21 +149,49 @@ AuxId OpticalCollector::optical_aux_id() const
...
+    if (auto id = this->cherenkov_aux_id())
+    {
+        auto& gen = dynamic_cast<GeneratorStateBase const&>(aux.at(id));
+        result.distributions += gen.buffer_size;
+    }
+    if (auto id = this->scintillation_aux_id())
+    {
+        auto& gen = dynamic_cast<GeneratorStateBase const&>(aux.at(id));
+        result.distributions += gen.buffer_size;
+    }
```
sethrj, line 192: Is this possible to write as just a dynamic cast?
```suggestion
    if (auto* gen = dynamic_cast<GeneratorStateBase const*>(&aux.at(id)))
    {
        result.distributions += gen->buffer_size;
    }
```

sethrj (approved):
Thanks for the arduous refactor!

src/celeritas/optical/OpticalCollector.cc
amandalund: Hmm maybe not since the IDs need to be valid?

Issue #1512, opened 2024-11-19
Add basic optical surface physics


sethrj:
Next steps:
- [x] Boundary interaction result for classical surface physics
- [x] Apply interaction result to optical track
- [x] Extend the boundary action so we can have it in the stepping loop twice
- [x] Geometry surface specification: #1351 @sethrj 
- [x] Multiple surface physics, surface<->physics mappings, with future extension to user physics
- [x] Importing surface properties from Geant4: @stognini 

Surface physics types:
- reflection/refraction
- diffuse reflective/lambertian (LZ)
- polished crystal (calvision)
- user customizable
- (GLISUR, unified models?)

Boundary physics class:
- Map surface ID -> {physics ID, local physics surface index}
- Determines whether multiple bounces are necessary etc.
- Manages temporary data (sampled local normal, ...)

End of step if encountered boundary:
- Save previous volume
- Cross boundary into next volume
- Look up physics on the boundary (https://github.com/celeritas-project/celeritas/issues/1351 )
- Select boundary physics action based on current volume and  (default: just crossing; )
- Sample physics (or whatever, may scatter or absorb), return interaction result
- Apply result using geometry normal: change geometry physical state + particle state (polarization), logical state if reflection
- Cross boundary again for particles reflecting (reuse boundary crossing kernel)

Boundary interaction result:
- No change
- Internal reflection
- Refraction with some polar angle change
- Absorption
- Absorption + detection (hit scoring: call back to geant4/host)?
- Custom?
- Isotropic?

Future optimizations:
- Peek at next volume rather than crossing in and out for reflection
- Definitely profile internal reflection in fiber optics: average reflections per volume before leaving

Future work:
- Special volume physics for total internal reflection

hhollenb:
After reviewing some of the content of `G4OpBoundaryProcess.cc`, there seems to be a rough pattern between most of the models:

```
while dot(newMomentum, globalNormal) <= 0:
    calculate reflectivity and transmissivity
    sample reflection / absorption / transmission
    if reflection:
        choose reflection type
        calculate reflection
    else if absorption:
        do absorption based on material efficiency
    else if transmission:
        sample facet normal based on momentum and surface type
        sample new momentum direction from distribution
        determine new polarization from momentum proposal
        update old momentum and polarization with new quantities for next round of sampling
```

The main differences between the surface models enters for:
- How they calculate probabilities for transmission, absorption, reflection
[...]

PR #1801, opened 2025-05-30
Add optical surface physics

hhollenb:
# Adding optical surface physics

For issue #1512 

An initial draft of optical surface physics. Still early (missing documentation, asserts, etc) but I'm making sure it's up here since I've been having some issues with my work computer the past week.

Thoughts
- Having a separate surface physics params since the data is laid out significantly differently than how the volumetric physics params data is organized.
- [ ] Need to handle a more general boundary action selection process
- [ ] Write surface normal functions for VecGeom GeoTrackView
- [ ] Make surface normal calculations const?
- [ ] May need to include some optical boundary physics state (e.g. for air gaps in back painted surfaces in UNIFIED model)
- [ ] Build surface data from imported geant4 data


src/orange/OrangeTrackView.hh
```diff
@@ -1207,5 +1212,36 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(LevelId lev)
     return this->get_transform(this->get_daughter(lsa));
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Get the surface normal vector of the surface the particle is currenlty on.
+ */
+CELER_FUNCTION Real3 OrangeTrackView::local_surface_normal()
```
sethrj: I think this and the global surface normal can be `const`.

src/orange/OrangeTrackView.hh
```diff
@@ -1207,5 +1212,36 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(LevelId lev)
...
+{
+    Real3 normal = this->local_surface_normal();
+
+    // Rotate up local surface normal to get the global surface normal
+    auto apply_transform = TransformVisitor{params_};
+    auto rotate_up = [&normal](auto&& t) { normal = t.rotate_up(normal); };
+    for (auto level : range<int>(this->level().unchecked_get()).step(-1))
+    {
+        apply_transform(rotate_up, this->get_transform(LevelId(level)));
+    }
```
sethrj: This is used in `OrangeTrackView::set_dir`; can you replace the similar chunk of code in that function to call this one?

sethrj (commented):
The new local/normal surface code looks good! Can you put that in a separate PR? (I can add the testing code for it since I know you haven't dug much into the geometry.)

src/celeritas/optical/surface/SimpleReflectionExecutor.hh
```diff
@@ -0,0 +1,93 @@
...
+    auto rng = track.rng();
+
+    // Get surface normal from geometry
+    Real3 surface_normal = geo.global_surface_normal();
+
+    real_type reflectivity = params.scalars.global_reflectivity;
+    real_type refl_and_trans = reflectivity
+                               + params.scalars.global_transmittance;
+
+    CELER_EXPECT(0 <= reflectivity && reflectivity <= refl_and_trans
```
stognini: Easier to read if `reflectivity >= 0`.

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -0,0 +1,84 @@
...
+using ValudGridId = OpaqueId<ValueGrid>;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Scalar quantities used by optical surface physics.
+ */
+struct SurfacePhysicsParamsScalars
+{
+    //!@{
+    //! \name Global surface properties for testing purposes
```
stognini: How did you test this?

hhollenb: These were meant to be a constant reflectivity and transmittance for all surfaces, just as a temporary stand in to get the framework for optical surface physics in. They'll be removed once we have imported optical surface data with actual reflectivity / transmittance values.

src/celeritas/optical/surface/SimpleReflectionExecutor.hh
sethrj: I dunno, I've seen before (and don't object to) a < b<= c in that form

src/orange/OrangeTrackView.hh
sethrj: Just saw this as a to-do :)

src/celeritas/optical/surface/SimpleReflectionExecutor.hh
stognini: Yeah, I don't really object to as well. My preference comes from my brain always defaulting to [my_parameter] on the left and [what_should_be] on the right when reading.

src/celeritas/optical/surface/SimpleReflectionExecutor.hh
```diff
@@ -0,0 +1,93 @@
...
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ */
+struct SimpleReflectionExecutor
+{
+    inline CELER_FUNCTION SurfaceInteraction operator()(CoreTrackView&) const;
```
stognini: Nitpick: usually on structs we have data first, then functions. Is this a code guidance that we'd like to set?

src/celeritas/optical/surface/SurfaceInteractionApplier.hh
```diff
@@ -0,0 +1,130 @@
...
+            geo.cross_boundary();
+
+            if (CELER_UNLIKELY(geo.failed()))
+            {
+                track.apply_errored();
+                return;
+            }
+        }
+
+        // Update direction and polarization
```
stognini: So this happens on both reflected and transmitted actions, with `cross_boundary` only being called if transmitted, right?
If so, I'd make a small note to make it easy to follow that otherwise is just reflection.

src/celeritas/optical/surface/SimpleReflectionExecutor.hh
sethrj: Yeah that would be a good one to add to the "Odds and ends" section of the style guide.

sethrj:
Sketches and comments from today's discussion:
```diff
diff --git a/src/celeritas/inp/Physics.hh b/src/celeritas/inp/Physics.hh
index 96f4ab625..7edcc9c47 100644
--- a/src/celeritas/inp/Physics.hh
+++ b/src/celeritas/inp/Physics.hh
@@ -57,7 +57,7 @@ struct EmPhysics
 /*!
  * Optical physics processes and options.
  */
-struct OpticalPhysics
+struct OpticalGeneratorPhysics
 {
     //!@{
     //! \name Optical photon generation
@@ -74,6 +74,35 @@ struct OpticalPhysics
     //!@}
 };
 
+struct Polished
+{
+};
+struct Glisur
+{
+    real_type roughness{};
+};
+struct Unified
+{
+    real_type sigma_alpha{};  // gaussian sigma
+};
+
+using UserCallback = std::function < something() > ;
+
+struct OpticalInterfacePhysics
+{
+    // Roughness model: variant{polished, glisur [uniform smear], unified [gaussian], user callback}
+    // Reflectivity model: variant{glisur, unified, user}
+    // Interaction model: variant{dielectric, ...}
+};
+
+// Surface physics defines *exiting* properties:
+// interface (border) surface: {pre} -> {layer 0} -> {layer 1} ->
+struct OpticalSurfacePhysics
+{
+    OpticalInterfacePhysics volume;
+    std::vector<std::pair<optical::MaterialId, OpticalInterfacePhysics>> layers;
+};
+
 //---------------------------------------------------------------------------//
 /*!
  * Hadronic physics processes and options.
@@ -107,8 +136,13 @@ struct Physics
     //! Enable electromagnetic physics
     std::optional<EmPhysics> em{std::in_place};
 
-    //! Enable optical photon physics
-    std::optional<OpticalPhysics> optical;
+    //! Enable generation of optical photons from EM tracks
+    std::optional<OpticalGeneratorPhysics> optical_gen;
+
+    // TODO: Select volumetric optical physics
+
+    //! Define physics for each optical surface
+    std::vector<OpticalSurfacePhysics> optical_surface;
 
     //! Enable hadronic physics
     std::optional<HadronicPhysics> hadronic;
diff --git a/src/celeritas/optical/surface/SurfacePhysicsData.hh b/src/celeritas/optical/surface/SurfacePhysicsData.hh
index f4d769ede..d92e37bb0 100644
--- a/src/celeritas/optical/surface/SurfacePhysicsData.hh
+++ b/src/celeritas/optical/surface/SurfacePhysicsData.hh
@@ -20,21 +20,46 @@ namespace optical
 //---------------------------------------------------------------------------//
 
 using SurfaceId = OpaqueId<struct OpticalSurface_>;
+using ModelDataId = OpaqueId<struct ModelData_>;
 
 using ValueGrid = NonuniformGridRecord;
 using ValudGridId = OpaqueId<ValueGrid>;
 
+// A map of {surface id: model data} given to a specific model's constructor results in a vector; each
+// index in the vector corresponds to ModelDataId
+/*
+example with surface IDs mapping to properties:
+ model 0 [unified roughness]: {1: ua, 2: ub, 4: uc}
+ model 1 [glisur roughness]: {0: ga, 3: gb}
+ model 2 [user roughness]: {5: xa, 6:xb}
+
+results in 7 SurfaceInterfaceModels:
+ [{SM{1}, MD{0}}, {SM{0}, MD{0}}, {SM{0}, MD{1}}, {SM{1}, {MD{1}}, {SM{0}, MD{2}},
+  {SM{2}, MD{0}}, {SM{2}, MD{1}}]
+*/
+struct ModelAndData
+{
+    //! Action ID of the physics model to be applied
+    ActionId model;
+    //! Index local to the physics model: the 'i'th surface using that model
+    ModelDataId data;
+};
+
+// stupid simple : scalar
+// wavelength -> scalar
+
+/*
+ 
+
 //---------------------------------------------------------------------------//
 /*!
- * Store optical physics data for a given surface layer.
+ * Redirect to optical physics models and data for a volume or layer.
  */
-struct SurfaceLayerRecord
+struct SurfaceInterfaceModels
 {
-    // TODO: Per model layer IDs
-
-    ActionId roughness_model{};
-    ActionId reflectivity_model{};
-    ActionId interaction_model{};
+    // TODO: ModelAndData roughness{};
+    ModelAndData reflectivity{};
+    // ModelAndData interaction{}; // refraction or reflection, model specififc?
 
     //! Whether data is assigned and valid
     inline CELER_FUNCTION operator bool() const { return false; }
@@ -48,8 +73,8 @@ struct SurfaceLayerRecord
  */
 struct SurfaceRecord
 {
-    ItemRange<SurfaceLayerRecordId> layers;
-    ItemRange<OpticalMaterialId> optical_materials;
+    SurfaceInterfaceModels volume;
+    // TODO: ItemRange<SurfaceInterfaceModels> layers;
 };
 
 //---------------------------------------------------------------------------//
@@ -101,6 +126,28 @@ struct SurfacePhysicsParamsData
     }
 };
 
+/*
+ * Surface properties:
+ * - If no optical surface ID at a boundary crossing, we kill the track
+ * - 
+* (todo: possible consistency track for optical materials -> volumes -> surfaces)
+
+Initial impl:
+1. Add surface/volume properties to core and optical params
+2. Extend "init boundary" to look up surface ID and kill (no need for SD check) if not found
+3. Hardcoded smooth surface, scalar reflection probability, action for refract/reflect/transmit
+4. Grid reflection (requires moving data from input to device, loading surface data)
+5. Surface roughness: require careful "on the surface multiple reflection" logic
+6. Multiple roughness models (polished + uniform smearing)
+7. Multiple reflection models (?)
+8. Interaction models
+9. Layers?
+
+interface surfaces (aka boundary): know before/after materials at construction time
+initially exiting from skin surface: know before material, have to save "after" from post-crossing volume
+initially entering into skin surface: know after material, have to save "before" from post-crossing volume
+*/
+
 //---------------------------------------------------------------------------//
 /*!
  * Dynamic optical surface physics state data.
@@ -118,6 +165,7 @@ struct SurfacePhysicsStateData
 
     //// Persistent State Data ////
 
+    StateItems<MaterialId> unknown_material; // null for interface?
     StateItems<SurfaceId> surface;
     StateItems<SurfaceLayerId> current_layer;
     StateItems<Real3> surface_normal;
```

hhollenb:
Ok added some hefty changes to reflect some of the most recent PRs and incorporate `inp::SurfacePhysics` and `SurfacePhysicsMapBuilder` from discussions with Stefano and Seth. Made a lot of small changes to unify some of the OpaqueId's used throughout the code and get a standard sense of `SurfaceId` vs `PhysSurfaceId`. I've got a lot more features to add that depend on this PR so it's best to get this in before working on the next steps.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -633,8 +633,20 @@ inp::OpticalPhysics import_optical_physics()
     }
     log_and_rethrow(std::move(handle));
 
-    CELER_LOG(debug) << "Loaded " << geo->num_surfaces()
-                     << " optical physics surfaces";
+    // Default Geant4 surface
+    PhysSurfaceId default_surface{result.surfaces.num_phys_surfaces()};
```
sethrj: For the surface map stuff, I am using the null SurfaceId to indicate the "default" surface (which is consistent with the return result from the VolumeSurfaceView). For this PR, to keep it simple, can we stick to using the GeoSurfaceIds and postpone the multiple layers? We will need to add a geo->phys mapping for the "last surface" default to be sensible (the surface map keeps the extra surface ID as an implementation detail). Then we can define a `{multiple geo surface IDs and/or null geo surface ID} -> {physics surface ID vector}` mapping in a follow-on.

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.hh
```diff
@@ -39,7 +39,7 @@ class GeantSurfacePhysicsHelper
 
     // Insert a value into a map in place
     template<class T>
-    inline void emplace(std::map<SurfaceId, T>& m, T&& value) const;
+    inline void emplace(std::map<PhysSurfaceId, T>& m, T&& value) const;
```
sethrj, line 42: When we switch from geo->phys properties we'll also need to update the fact that the surface physics helper is "duplicating" surface properties: per our original discussion, rather than have an indirection of geometry surface -> optical surface property, we duplicate the properties for each geometry surface.

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -219,22 +221,39 @@ struct InteractionModels
...
+
+    //! Number of physics surfaces
+    PhysSurfaceId::size_type num_phys_surfaces() const
+    {
+        PhysSurfaceId::size_type num = 0;
+        for (auto const& mats : materials)
+        {
+            num += mats.size() + 1;
+        }
+        return num;
```
sethrj: I think this should only be calculated during setup, not input. A geometry `SurfaceId{}` to physics `PhysSurfaceId{...}` mapping is needed.

src/celeritas/optical/Types.hh
```diff
@@ -33,18 +34,34 @@ enum class GeneratorType
 namespace optical
 {
 
+//! Sub-steps of surface physics boundary crossing
+enum class SurfacePhysicsStep
```
sethrj: `Step` could be confusing, this is more analogous to ActionOrder (but inside a single action). Maybe SurfaceActionOrder?

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -0,0 +1,211 @@
...
+{
+    ItemMap<SubsurfaceMaterialId, OpaqueId<OptMatId>> subsurface_materials;
+    ItemMap<SubsurfaceInterfaceId, PhysSurfaceId> subsurface_interfaces;
+
+    //! Construct record directly from ranges
+    SurfaceRecord(Range<OpaqueId<OptMatId>> materials,
+                  Range<PhysSurfaceId> interfaces)
+        : subsurface_materials(materials), subsurface_interfaces(interfaces)
+    {
+    }
```
sethrj: This actually might not be necessary since #1916 was merged.

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -0,0 +1,211 @@
...
+
+    //! Whether data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return !subsurface_interfaces.empty()
+               && subsurface_materials.size() + 1
```
sethrj: Technically the first condition is redundant 🙃 
```suggestion
        return subsurface_materials.size() + 1
```

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -0,0 +1,211 @@
...
+{
+    //!@{
+    //! \name Type aliases
+    using ModelMap = SurfacePhysicsMapData<W, M>;
+
+    template<class T>
+    using GeoSurfaceItems = Collection<T, W, M, SurfaceId>;
+
+    template<class T>
+    using SurfaceStepArray = EnumArray<SurfacePhysicsStep, T>;
```
sethrj: I don't think we'll have any of these arrays aside from the physics models, so it might be clearer to just use the type directly as `EnumArray<SurfacePhysicsStep, ModelMap>`?

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -0,0 +1,211 @@
...
+    operator=(SurfacePhysicsParamsData<W2, M2> const& other)
+    {
+        CELER_EXPECT(other);
+        scalars = other.scalars;
+        surfaces = other.surfaces;
+        subsurface_materials = other.subsurface_materials;
+        for (auto step : range(SurfacePhysicsStep::size_))
+        {
+            model_maps[step] = other.model_maps[step];
+        }
```
sethrj, line 126: Ah this is because the assignment `operator=` doesn't do the conversion automatically 🤔

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+    CELER_EXPECT(track_id_ < states_.size());
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize track state with given initializer data.
+ */
+CELER_FUNCTION SurfacePhysicsView&
+SurfacePhysicsView::operator=(Initializer const& init)
+{
```
sethrj, line 151: Should probably have `CELER_EXPECT(init.surface)` since there's a separate `reset`

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Reset the state of a track.
+ *
+ * Invalidates the surface ID, indicating the track is no longer undergoing
+ * boundary crossing.
+ */
+CELER_FUNCTION void SurfacePhysicsView::reset() const
```
sethrj: Shouldn't be const; and I wonder if there's a more descriptive way of saying we're no longer on a surface...

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+    return states_.surface_orientation[track_id_];
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Whether the track is undergoing boundary crossing.
+ *
+ * Returns true if there's a valid surface ID, otherwise false.
+ */
+CELER_FUNCTION bool SurfacePhysicsView::is_crossing_boundary() const
```
sethrj, line 203: Not your problem (future to-do for https://github.com/celeritas-project/celeritas/issues/721 I guess) but we should probably rename boundary to surface...

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Whether the direction is exiting the surface.
+ */
+CELER_FUNCTION bool SurfacePhysicsView::is_exiting(SubsurfaceDirection d) const
+{
+    return (this->subsurface_position().get() + to_signed_offset(d))
+           >= this->num_positions();
```
sethrj: This looks susceptible to signed/unsigned overflow errors (at the very least it's a signed/unsigned compare which can give warnings).  I think static casting the sizes to `int` is appropriate.

(Or is this cleverly relying on `size_type(-1) >= num_positions` overflowing? If so, I think a static cast  of the left size to `size_type` and a note would be great.)

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+    {
+        return states_.pre_volume_material[track_id_];
+    }
+    if (this->in_post_volume())
+    {
+        return states_.post_volume_material[track_id_];
+    }
+
+    auto material_record_id = this->to_record_index(
+        SurfaceTrackPosition(this->subsurface_position().get() - 1),
```
sethrj: You should be able to write 
```suggestion
        this->subsurface_position() - 1,
```
and you'd want an extra check against the positoin being zero.

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+    // Whether the track is in the post-volume
+    inline CELER_FUNCTION bool in_post_volume() const;
+
+    //// ACCESS PHYSICS DATA ////
+
+    // Position of the track in the surface crossing
+    inline CELER_FUNCTION SurfaceTrackPosition subsurface_position() const;
+
+    // Position of the track in the surface crossing
+    inline CELER_FUNCTION SurfaceTrackPosition& subsurface_position();
```
sethrj: We usually have a setter like so:
```suggestion
    inline CELER_FUNCTION void subsurface_position(SubsurfaceTrackPosition);
```
that can do error checking.

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convert track-local position to index in a surface record.
+ */
+template<class T, class U>
+CELER_FUNCTION U SurfacePhysicsView::to_record_index(
+    SurfaceTrackPosition pos, ItemMap<T, U> const& map) const
+{
```
sethrj, line 386: This function's a bit dangerous because it's discarding the type safety of the opaque IDs. Why don't you use `SubsurfaceInterfaceId` to access the map directly, and have a function `subsurface_interface()` that does the mapping of `(pos, orientation) -> interface` (which can `EXPECT(on_boundary)` and `ENSURE(result < num_interfaces())`" )?

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -0,0 +1,391 @@
...
+    // Reset surface physics state of the track
+    inline CELER_FUNCTION void reset() const;
+
+    //// STATE INVARIANTS ////
+
+    // Get current geometric surface
+    inline CELER_FUNCTION SurfaceId surface() const;
+
+    // Get surface orientation
+    inline CELER_FUNCTION SubsurfaceDirection orientation() const;
```
sethrj, line 63: The traversal through the discretized "lineland" might merit a helper class of its own with  the same interface as the geo track to help with clarity. The "position" is an oriented subsurface `{-0, +0, -1, +1, ..., -N, +N}`, "surface" is the subsurface ID `{0, 0, 1, 1, ..., N, N}`, "volume" (which gives material ID) is `{0, 1, 1, 2, ... N-1, N}`, `is_outside` is true for position <= -0 or >= +N, and "exiting" I guess would be if we're outside heading away?

sethrj (commented):
I think this looks really good, but I think the multi-layer stepping needs a little more consideration before implementing. Is the multi-layer infrastructure higher priority than getting the basic surface physics? I know it's exciting because it's novel (I'm excited too!!) but when we drew up a plan for progressive capability we said single layers first so we can start some testing simultaneously.

Can we discuss removing the multi-layer changes (going back to a single SurfaceId type, a single layer) for this PR and immediately reverting that commit and opening a PR for follow-up work? Then the surface model implementations can proceed in parallel.

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.hh
hhollenb: What surface properties is the surface physics helper duplicating here?

src/celeritas/inp/SurfacePhysics.hh
hhollenb: The `SurfaceId` to `PhysSurfaceId` is determined by `std::vector<InterstitialMaterials> materials` (which is indexed by SurfaceId) since the number of physics surfaces is required to be the number of interstitial materials + 1 for every geometry surface. We can't have more / less physics surfaces than that so this is a quick helper to initialize collections which are indexed by PhysSurfaceId.

src/celeritas/optical/surface/SurfacePhysicsData.hh
hhollenb: AFAIK `EnumArray` is implemented as a C-style array so it doesn't support copy assignment and doesn't have an overriding `operator=`.

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: Maybe `clear`? It's only purpose is to reset the surface physics state when called by `PostBoundaryExecutor`

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: Yeah it was intentionally using overflowing for that reason. I'll add a note and cast.

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: `if (this->in_pre_volume())` is a specific check against the position being zero. I'll add an extra assert on this.

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: The problem is then that you're mixing up the `SubsurfaceInterfaceId` which indexes how the data is laid out in storage with the track's position in the materials, which depends on the surface orientation. If there's a type-safe reverse ItemMap then we could use that instead of the calculation?

You also need to do the same calculations for both the interfaces map and the materials map, which is why I tried to pull it out into a single function. As with `ItemMap` the goal of this function is to wrap the type safety of `SurfaceTrackPosition` and do the type-unsafe calculation in a single place, and return a type safe ID that's valid in the given map.

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: Is the position an oriented subsurface? The orientation here refers to whether the surface record is laid out `A -> B` or `B -> A` in storage. The next interface is determined by the surface normal and the track momentum direction, which doesn't depend on how the surface is laid out in storage.

We might be able to refactor out a helper class and have the position be the pair of the volume and the track direction (stored from normal + momentum calculation)?

src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
```diff
@@ -66,20 +68,27 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
...
-    else
+    auto oriented_surface
+        = select_surface(track.surface(), geo.volume_instance_id());
+    if (!oriented_surface)
     {
-        // If there's no surface, mark photon as killed
-        // TODO: Add default behavior
-        track.sim().status(TrackStatus::killed);
+        // Use default surface data
+        oriented_surface.surface = surface_physics.default_surface();
```
amandalund, line 89: I think we want to use the default surface only if the new volume has an optical material, and otherwise mark the track as killed.

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.hh
sethrj: Moved to #1920 : currently a single G4 optical property is being reproduced multiple times (one per geo surface that uses it)

src/celeritas/inp/SurfacePhysics.hh
sethrj: TODO: rename SurfaceId to GeoSurfaceId

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -219,22 +221,39 @@ struct InteractionModels
...
 
     //! Whether the data are assigned
     explicit operator bool() const
     {
-        return reflectivity && roughness && interaction;
+        return reflectivity && roughness && interaction && !materials.empty();
+    }
+
+    //! Number of physics surfaces
+    PhysSurfaceId::size_type num_phys_surfaces() const
```
sethrj: Reminder to remove this function (should be used during surface physics construction)

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -219,22 +221,39 @@ struct InteractionModels
...
+ *
+ * Interstitial materials are the interstitial materials per geometric surface.
+ * The last entry is used as the default surface.
  */
 struct SurfacePhysics
 {
     //!@{
     //! \name type aliases
     using DetectionEfficiency = std::map<SurfaceLayer, Grid>;
+    using InterstitialMaterials = std::vector<OptMatId>;
```
sethrj: Reminder to rename `VecOptMatId` and delete the unused alias above it

src/celeritas/io/ImportDataTrimmer.cc
```diff
@@ -226,6 +251,24 @@ void ImportDataTrimmer::operator()(ImportProcess& data)
...
 
+//---------------------------------------------------------------------------//
+void ImportDataTrimmer::operator()(inp::ReflectionForm& data)
+{
+    filter_out_infs(data.specular_spike.x);
+    filter_out_infs(data.specular_spike.y);
+    filter_out_infs(data.specular_lobe.x);
+    filter_out_infs(data.specular_lobe.y);
+    filter_out_infs(data.backscatter.x);
+    filter_out_infs(data.backscatter.y);
```
sethrj, line 262: Not super important but I think this should be `(*this)(data.specular_spike)` etc., and then the deeper calls (from inp::grid to std::vector) should handle the filtering on x/y.

src/celeritas/optical/surface/SurfacePhysicsParams.hh
```diff
@@ -0,0 +1,118 @@
...
+    ActionId surface_stepping_action() const
+    {
+        return surface_stepping_action_->action_id();
+    }
+
+    //! Action ID for finishing boundary interactions
+    ActionId post_boundary_action() const
+    {
+        return post_boundary_action_->action_id();
+    }
```
sethrj, line 88: Since we've decided to do surface models as something that is "inside" an action, is there an advantage to having these three actions separately rather than having a single action that launches multiple kernels? If not, we can simplify in a follow-up PR.

src/celeritas/optical/surface/SurfacePhysicsParams.cc
```diff
@@ -0,0 +1,224 @@
...
+
+    // Finalize data
+    CELER_ENSURE(data);
+
+    data_ = CollectionMirror<SurfacePhysicsParamsData>{std::move(data)};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build surface data form inputs.
```
amandalund:
```suggestion
 * Build surface data from inputs.
```

sethrj (approved):
Excellent work! It feels great to finally get this done :)

src/celeritas/Types.hh
```diff
@@ -80,6 +83,9 @@ using ParticleProcessId = OpaqueId<ProcessId>;
 //! Opaque index of a model applicable to a single particle type
 using ParticleModelId = OpaqueId<ModelId>;
 
+//! Opaque index of subsurface interface with physics models
+using PhysSurfaceId = OpaqueId<struct PhysSurface_, unsigned int>;
```
amandalund, line 87: I think this might resolve the `RootJsonDumper` test failures, unless there's a reason for making it an unsigned int?
```suggestion
using PhysSurfaceId = OpaqueId<struct PhysSurface_>;
```

amandalund (approved):
Awesome @hhollenb, this was a ton of work and looks great!

src/celeritas/io/ImportDataTrimmer.cc
hhollenb: Honestly what I wrote is probably a pretty bad hack. When passing down to grid via `(*this)(data.specular_spike)` you run into the issue that it might perform the vector size filtering if that's defined, even if `options_.physics == false`. This code really doesn't fit well into the overall logic of the `ImportDataTrimmer`.

src/celeritas/optical/surface/SurfacePhysicsParams.hh
hhollenb: I personally like having an init action exposed so that the `PropagateExecutor` can just assign geometry limited tracks to `InitBoundaryAction` explicitly, and we don't need to keep track inside the stepping action as to whether tracks are queued to be initialized or not.

Otherwise I don't thinks there's any performance differences, and logically they're pretty much equivalent.

src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
hhollenb: Yes we definitely do. A small caveat is that for some surfaces (e.g. dieletric-metal) there can be a valid surface but invalid post-volume optical material (metals without index of refraction) so they should only be killed if there's an invalid surface. In the `PostBoundaryAction` we can do an extra check if a photon crossed into an invalid optical material through a valid surface and kill it there as well.

src/celeritas/Types.hh
hhollenb: This was something I added after Seth's suggestion about the mismatches I was having with `PhysSurfaceId` using `unsigned long` in my builds (without device enabled) while in other builds it uses `unsigned int`. I had your suggestion there before but was getting failing CI tests with it. I don't think I added the regenerated root test data to the commit with this change so let me try that real quick

amandalund: Looks like you updated the ROOT data in your last commit but not this change? I tested it out on a CUDA build and it did seem to work (after initially failing the same as the CI).

src/celeritas/optical/surface/SurfacePhysicsParams.hh
sethrj: I was thinking they'd be assigned the surface physics action, and they would be initialized if their current surface state is empty. My concern wasn't performance (since the number of kernels wouldn't change) but just reducing the extra scaffolding.

src/celeritas/io/ImportDataTrimmer.cc
sethrj: Yuck, fair enough. Hack it is!

src/celeritas/Types.hh
hhollenb: Ok looks like it works now, just need to also add it to `OptMatId`!

PR #1802, opened 2025-05-30
Print Celeritas configuration options when loading downstream

sethrj:
This extends the `COMPONENTS` for downstream `find_package` commands to include "forced" packages (e.g. CLI11) and options (e.g. REAL_TYPE). It additionally prints the list of enabled options when loading Celeritas, useful for downstream diagnostics.

(The motivation of this PR is prototyping adding this to DD4HEP.)

When configuring the first time, if COMPONENTS not set, YOu will see:
```
-- Found Celeritas: /Users/seth/Code/celeritas-temp/build/lib/cmake/Celeritas/CeleritasConfig.cmake (found suitable version "0.6.0.1", minimum required is "0.6") found components: covfie Geant4 HepMC3 PNG Python ROOT CLI11 GTest CORE_RNG=xorwow CORE_GEO=ORANGE OPENMP=disabled UNITS=CGS REAL_TYPE=double
```
If COMPONENTS *is* set, you will only see the ones you asked for.


esseivaju (approved):
LGTM!

PR #1803, opened 2025-05-30
Move optical photon generation to optical stepping loop

amandalund:
Follow-on to #1800.

This moves the optical generator actions from the core to the optical loop. Currently track initializers are still created and buffered at the start of the first optical step, but as a next step we can explore generating the photons directly in the track slots.

### DIFF
--- a/src/celeritas/optical/CoreState.hh
+++ b/src/celeritas/optical/CoreState.hh
@@ -8,6 +8,7 @@
 
 #include "corecel/cont/Span.hh"
 #include "corecel/data/AuxInterface.hh"
+#include "corecel/data/AuxStateVec.hh"
 #include "corecel/data/CollectionStateStore.hh"
 #include "corecel/data/ObserverPtr.hh"
 #include "celeritas/Types.hh"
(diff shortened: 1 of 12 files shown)


src/celeritas/optical/CoreState.hh
```diff
@@ -136,6 +138,14 @@ class CoreState final : public CoreStateBase
     // Inject primaries to be turned into TrackInitializers
     void insert_primaries(Span<TrackInitializer const> host_primaries) final;
 
+    //// AUXILIARY DATA ////
+
+    //! Access auxiliary core state data
+    SPAuxStateVec const& aux() const { return aux_state_; }
+
+    //! Access auxiliary core state data (mutable)
+    SPAuxStateVec& aux() { return aux_state_; }
```
sethrj, line 147: You probably want a const reference to the shared pointer; otherwise someone could `reset` it which I assume is undesirable?
```suggestion
    SPAuxStateVec const& aux() { return aux_state_; }
```

sethrj (approved):

src/celeritas/optical/CoreState.hh
amandalund: This part is a little hacky, but we need to be able to assign it at the start of the optical loop so we can access the aux state data.

sethrj: Hack away!

sethrj:
You may have to push an empty commit to get it to build the CUDA/HIP tests (since the last commit was before the draft status got removed). But if you have tested this on a CUDA platform I can just force-merge it.

PR #1804, opened 2025-05-31
Add surface normal calculations in GeoTrackView

hhollenb:
# Add surface normal calculations to GeoTrackView

Surface normals are necessary for calculating optical surface physics. When a track is on a boundary, we will need to calculate the surface normal in global coordinates so that it can be compared to the track's direction. This is an initial draft to separate it out from #1801 .


src/geocel/g4/GeantGeoTrackView.hh
```diff
@@ -340,6 +328,35 @@ CELER_FORCEINLINE bool GeantGeoTrackView::is_on_boundary() const
     return safety_radius_ == 0.0;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Get the surface normal of the boundary the track is currently on, in local
+ * coordinates.
+ */
+CELER_FUNCTION Real3 GeantGeoTrackView::local_surface_normal() const
```
hhollenb: The `GetLocalExitNormal` and `GetGlobalExitNormal` in the `G4Navigator` can't be used here since they expect that the step moves into a new volume, but when we move to the boundary we don't change volumes so internal state of the navigator isn't correct. This was an initial attempt at getting the surface normal but doesn't work for being on the boundary of a daughter volume.

src/geocel/vg/VecgeomTrackView.hh
```diff
@@ -330,6 +334,33 @@ CELER_FUNCTION bool VecgeomTrackView::is_on_boundary() const
     return vgstate_.IsOnBoundary();
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Get the surface normal of the boundary the track is currently on, in local
+ * coordinates.
+ */
+CELER_FUNCTION Real3 VecgeomTrackView::local_surface_normal() const
```
hhollenb: Haven't tested this at all yet, not as familiar with vecgeom.

src/orange/OrangeTrackView.hh
```diff
@@ -1209,12 +1189,51 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(DaughterId daughter_id
...
     LSA lsa(&states_, track_slot_, lev);
     return this->get_transform(this->get_daughter(lsa));
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Get the surface normal vector of the surface the particle is currenlty on,
+ * in local coordinates.
+ */
+CELER_FUNCTION Real3 OrangeTrackView::local_surface_normal() const
```
hhollenb: Mainly copied from the `set_dir` function.

src/geocel/g4/GeantGeoTrackView.hh
```diff
@@ -51,6 +51,13 @@ class GeantGeoTrackView
     using Real3 = Array<real_type, 3>;
     //!@}
 
+    //! Helper struct for initializing from an existing geometry state
+    struct DetailedInitializer
```
amandalund: It looks like some unrelated recent changes to the geo track views might have been accidentally reverted here.

src/geocel/g4/GeantGeoTrackView.hh
sethrj: Maybe we should just replace the implementation with `CELER_NOT_IMPLEMENTED` for now: we primarily use GeantGeo for comparison with ray tracing and never with actual simulation.

sethrj (commented):
For now, let's replace the implementations in the GeantGeo and VecGeom track views with "not implemented" and just worry about ORANGE.  Let's drop the `global_` from that accessor since by default everything in the geo track view is "global" (viz., `pos` and `dir`). Since we refer to being on a `boundary` sometimes let's simplify further to `geo.normal` and `geo.local_normal`.

You should be quickly able to add some basic tests: search for `move_to_boundary` and `cross_boundary` in `orange/OrangeJson.test.cc`. Your tests will look like:
```diff
--- a/test/orange/OrangeJson.test.cc
+++ b/test/orange/OrangeJson.test.cc
@@ -551,12 +551,16 @@ TEST_F(NestedRectArraysTest, leaving)
     EXPECT_EQ("arrfill.px", this->surface_name(geo));
     EXPECT_EQ("Bfill", this->volume_name(geo));
     EXPECT_VEC_SOFT_EQ(Real3({4, 1.5, 0.5}), geo.pos());
+    EXPECT_VEC_SOFT_EQ(Real3({....}), geo.normal());
+    EXPECT_VEC_SOFT_EQ(Real3({....}), geo.local_normal());

     // Cross universe boundary
     geo.cross_boundary();
     EXPECT_EQ("arrfill.px", this->surface_name(geo));
     EXPECT_EQ("interior", this->volume_name(geo));
     EXPECT_VEC_SOFT_EQ(Real3({4, 1.5, 0.5}), geo.pos());
+    EXPECT_VEC_SOFT_EQ(Real3({....}), geo.normal());
+    EXPECT_VEC_SOFT_EQ(Real3({....}), geo.local_normal());

     next = geo.find_next_step();
     EXPECT_SOFT_EQ(16, next.distance);
```

sethrj:
TODO: only global normal calculation, no local-to-global, local direction, etc.

sethrj:
We start the boundary surface process after the surface crossing, so the queried normal on the + side of the surface should point back toward the original volume (aka outward)
IMG_8132

src/geocel/g4/GeantGeoTrackView.hh
```diff
@@ -97,10 +97,8 @@ class GeantGeoTrackView
     inline bool is_on_boundary() const;
     //! Whether the last operation resulted in an error
     CELER_FORCEINLINE bool failed() const { return false; }
-    // Get the local normal vector of the current surface
-    inline CELER_FUNCTION Real3 local_surface_normal() const;
-    // Get the global normal vector of the current surface
-    inline CELER_FUNCTION Real3 global_surface_normal() const;
+    // Get the normal vector of the current surface
+    inline CELER_FUNCTION Real3 surface_normal() const;
```
sethrj: I think we discussed just calling it `normal` since the geo track view uses the word "boundary" rather than "surface" , and technically geant4 doesn't have surfaces.

test/orange/OrangeJson.test.cc
```diff
@@ -282,6 +285,8 @@ TEST_F(UniversesTest, cross_into_parent_non_coincident)
     EXPECT_EQ("inner_a.my", this->surface_name(geo));
     EXPECT_EQ("johnny", this->volume_name(geo));
     EXPECT_VEC_SOFT_EQ(Real3({2, -4, 0.75}), geo.pos());
+    EXPECT_VEC_SOFT_EQ(Real3({0, 1, 0}), geo.surface_normal());
+    EXPECT_LT(dot_product(geo.surface_normal(), geo.dir()), 0);
```
sethrj: Let's also test the normal at the end of the previous block: after moving to the boundary, before crossing it. (I forgot whether we decided to make the normal change direction when we cross sides and/or change directions. Do we let the surface physics code flip it based on the dot product?)

src/orange/OrangeTrackView.hh
```diff
@@ -1209,12 +1199,63 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(DaughterId daughter_id
...
+ */
+CELER_FUNCTION Real3 OrangeTrackView::surface_normal() const
+{
+    CELER_EXPECT(this->is_on_boundary());
+
+    auto normal = this->geo_surface_normal();
+    // Flip direction if on the outside of the surface
+    if (this->sense() == Sense::outside)
+    {
+        normal *= real_type{-1};
```
sethrj: To avoid signed zeros, I recommend using `negate` from `corecel/math/Algorithms.hh`

hhollenb:
Updated so that the Geant4 and VecGeom track views have unimplemented asserts in their `surface_normal` methods. In Orange I split the surface normal function into a helper which retrieves the surface normal as defined by the geometry in global coordinates. The public `surface_normal` method used by the physics then flips the normal based on the surface sense so that if this method is called after a `cross_boundary` then it will return a normal pointing into the previous volume (which is the convention we'll be using for the physics code).

test/orange/OrangeJson.test.cc
hhollenb: Currently I have the code set up so that it will return the normal directed into the volume the track is not currently in. So before crossing a boundary, it will point into the new volume (positive dot product) and after crossing it points into the previous volume (negative dot product). My understanding is that in general the normal won't be calculated until after we cross the boundary.

sethrj: OK perfect. That makes intuitive sense: even if we're numerically and logically 'on' the surface and headed away from it, the normal is toward the surface based on the stored sense. Nice!

src/orange/OrangeTrackView.hh
```diff
@@ -1209,12 +1200,63 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(DaughterId daughter_id
...
+    return normal;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the normal vector of the current surface.
+ *
+ * When called after \c cross_boundary this function gives the surface
+ * normal pointed into the previous volume. This ensures that the dot product
+ * of the surface normal and the track direction is negative.
```
sethrj: This is true only if it hasn't changed direction while on the boundary; it could "outside pointing in" and still have the same normal. I think your github comment phrased it better.

I'd add to the class documentation up top (since this function is hard to find): Even if we're numerically and logically 'on' the surface and the direction is headed away from it, the normal is toward the surface *no matter what*  based on the stored sense.

src/orange/OrangeTrackView.hh
```diff
@@ -1209,12 +1200,63 @@ CELER_FUNCTION TransformId OrangeTrackView::get_transform(DaughterId daughter_id
...
+
+    auto normal = this->geo_normal();
+    // Flip direction if on the outside of the surface
+    if (this->sense() == Sense::outside)
+    {
+        normal = negate(normal);
+    }
+
+    return normal;
+}
```
sethrj: Oh one final thing, can you move this function corresponding to where it is in the header, just below the definition for `is_on_boundary`?

sethrj (approved):
Awesome! Nice work. Just the last comment about placement of the function and the documentation.

src/orange/OrangeTrackView.hh
hhollenb: Got it, made the documentation a bit clearer

PR #1805, opened 2025-05-31
Make optical `max_step_iters` configurable and properly reset state after aborting

amandalund:
- Make `max_step_iters` in the optical loop a user-configurable option rather than hardcoding it to 1024
- Reset the state data and counters when the optical loop is aborted early
- Fix the assignment of the maximum step iterations in the core problem input
- Fix an assertion failure in the Cherenkov generator where the polarization was not orthogonal to the direction
- Activate the optical loop in the celer-sim lAr unit tests


sethrj (approved):
LGTM!

PR #1806, opened 2025-06-01
Fix orange polyhedron orientation and roundoff error

sethrj:
This fixes two errors:
1. The orientation for polyhedra was calculated correctly only for hexes. Now it works for all polys, with extended testing.
2. Roundoff error sometimes caused solid construction to fail with errors such as `Failed to convert solid type 'G4Polyhedra' named 'quad_third': invalid interior angle 1 [turns]: must be in (0, 1]`. A new soft equivalence test fixes it.

Before (spheres are for where the assertion failed):
polyhedra-old

After:
polyhedra-new

### DIFF
--- a/test/geocel/CMakeLists.txt
+++ b/test/geocel/CMakeLists.txt
@@ -81,6 +81,7 @@ if(CELERITAS_USE_VecGeom)
       "TwoBoxesVgdmlTest.*"
       "SimpleCmsVgdmlTest.*"
       "FourLevelsVgdmlTest.*"
+      # Polyhedra doesn't work because of anonymous postions
     )
     if(NOT VecGeom_SURF_FOUND)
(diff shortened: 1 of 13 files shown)


sethrj:
@elliottbiondo This should be a pretty easy review (mostly test, a few lines of change).

test/geocel/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ if(CELERITAS_USE_VecGeom)
       "TwoBoxesVgdmlTest.*"
       "SimpleCmsVgdmlTest.*"
       "FourLevelsVgdmlTest.*"
+      # Polyhedra doesn't work because of anonymous postions
```
elliottbiondo: positions

elliottbiondo (approved):

PR #1808, opened 2025-06-02
Full G4ExtrudedSolid support

elliottbiondo:
This MR adds full support for G4ExtrudedSolids, defined by extruding an arbitrary polygon along a polyline, with scaling factors applied at each polyline point. Convexity within polygons is handled via a convex decomposition approach [1], as demonstrated below.

image

[1] S. B. Tor and A. E. Middleditch, "Convex Decomposition of Simple Polygons," ACM Transactions on Graphics,"  volume 3(4), 224-265, 1984, https://doi.org/10.1145/357346.357348.


elliottbiondo:
@sethrj @mrguilima this is ready for your review

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -0,0 +1,243 @@
...
+                                               VecReal3&& polyline,
+                                               VecReal&& scaling)
+    : label_{std::move(label)}
+    , polygon_{std::move(polygon)}
+    , polyline_{std::move(polyline)}
+    , scaling_{std::move(scaling)}
+{
+    CELER_VALIDATE(polygon_.size() >= 3,
+                   << "polygon must have at least 3 vertices");
+    CELER_VALIDATE(polyline_.size() >= 2,
```
elliottbiondo, line 91: My next MR will contain utils for validating polylines.

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -155,10 +155,10 @@ NodeId StackedExtrudedPolygon::make_levels(
 
     // Create a vector of all concave regions, via recursion
     std::vector<NodeId> concave_nodes(concave_regions.size());
-    for (auto const i : range(concave_nodes.size()))
+    for (size_type i : range(concave_nodes.size()))
```
sethrj: Don't bother changing it, but the "best practices" (Scott Myers) is to static cast where needed and use auto rather than an implicit cast:
```suggestion
    for (auto i : range(static_cast<size_type>(concave_nodes.size())))
```

src/orange/orangeinp/StackedExtrudedPolygon.hh
```diff
@@ -0,0 +1,160 @@
...
+         ____ . . . . .
+        |   | level 1, stack 0, segment 0
+        |___|         .
+        .             .
+        .             .
+        .       ______
+        .      |      | level 1, stack 1, segment 0
+        . . .  |______|
+   \endverbatim
+ */
```
sethrj, line 87: This is great documentation! Can you add the class to the list in the `Objects` section of `doc/implementation/geometry/orange.rst`?

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -0,0 +1,243 @@
...
+                   << "polygon must have at least 3 vertices");
+    CELER_VALIDATE(polyline.size() >= 2,
+                   << "polyline must have at least 2 vertices");
+    CELER_VALIDATE(polyline.size() == scaling.size(),
+                   << "polyline and scaling must be the same size");
+
+    // If the polygon is convex and the polyline is a single segment, make an
+    // ExtrudedPolygon
+    bool degen_ok = true;
+    if (polyline.size() == 2 && detail::is_convex(make_span(polygon), degen_ok))
```
sethrj: Either make the bool constexpr or:
```suggestion
    if (polyline.size() == 2 && detail::is_convex(make_span(polygon), /* degen_ok = */ true))
```

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -0,0 +1,243 @@
...
+    {
+        concave_nodes[i] = this->make_levels(
+            vb, concave_regions[i], SubRegionIndex{si.level + 1, i});
+    }
+
+    auto level_label = this->make_level_label(si);
+
+    // Create a union of all concave regions
+    NodeId concave_union
+        = vb.insert_region(Label{label_, level_label + "_concave_union"},
```
sethrj: Instead of underscores, follow the polysolid and use a period `.` as a separator. To keep surface names short, and because they're really only used for debugging internally (we throw them out after construction), how about `".cu"` and the labels below `".ncu"` and `".d"`?

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -0,0 +1,243 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Make a label for a level.
+ */
+std::string StackedExtrudedPolygon::make_level_label(
+    StackedExtrudedPolygon::SubRegionIndex si) const
+{
+    return "level" + std::to_string(si.level);
```
sethrj: Same thing here ; and you might even consider skipping the level/stack/segment text and just ending up with nodes named `shape@1`, `shape@1.2`, `shape@1.2.3`, `shape@1.2.3.p0` where you have level, stack, and segment. They're always nested in that order right? You could add a line in the docs for people who really are curious.

src/orange/orangeinp/ObjectIO.json.cc
```diff
@@ -123,6 +124,16 @@ void to_json(nlohmann::json& j, SolidBase const& obj)
     }
 }
 
+void to_json(nlohmann::json& j, StackedExtrudedPolygon const& cr)
+{
+    j = {
+        {"_type", "stackedextrudedpolygon"},
```
sethrj, line 130: You missed the opportunity to abbreviate this as `SExPoly` 😂

sethrj (approved):
Really excellent work! The only thing is quibbling about the label length and consistency.

src/orange/orangeinp/StackedExtrudedPolygon.cc
elliottbiondo: Gotcha. I will change that way I can reference it in the future.

elliottbiondo:
(@sethrj done)

src/orange/orangeinp/StackedExtrudedPolygon.hh
```diff
@@ -0,0 +1,160 @@
...
+ * polyline:
+ * \verbatim
+        . . ._________
+        .   |         |
+        .___|         |
+        |             |
+        |             |
+        |       ______|
+        |      |      .
+        |______|. . . .
```
mrguilima: Wouldn't the convex hull be shaped more like
```verbatim
      ________
    /          |
   /           |
   |           |
   |           |
   |          /
   |________/ 
   ```
   than a perfect rectangle?

mrguilima (approved):
Just one question.

src/orange/orangeinp/StackedExtrudedPolygon.hh
sethrj: I think this is right; while fixing the docs I updated the nice ASCII art. @elliottbiondo can you check that I did it right? Then you can resolve this comment and it should merge.

elliottbiondo: Yes my original ASCII art was wrong --- it was not using the actual convex hull. Fixed now.

PR #1809, opened 2025-06-02
Make Geant4 volume IDs consistent wrt load order

sethrj:
This minor change updates the Geant4 geometry to keep volume IDs consistent for use in ORANGE volume ID mapping.


pcanal (approved):
LGTM.

amandalund (approved):

Issue #1351, opened 2024-08-02
Support optical surface interfaces


sethrj:
The LZ and ePIC forward detectors both specify volume types not seen in the main LHC: "bordersurface" in the former, and "skinsurface" for the latter. Both are related to surface-processes for #886 . These two special GDML structure entries seem only to declare interfaces rather than "wrap" a volume: the logical volume it creates is not referenced by other parts of the geometry.

LZ, for example, has the following:
```xml
  <solids>
    <opticalsurface finish="4" model="1" name="liquidXeTeflonSurfacePMT" type="1" value="1"/>
  </solids>
  <structure>
    <bordersurface name="lXePMTSurface_1" surfaceproperty="liquidXeTeflonSurfacePMT">
      <physvolref ref="Bottom_PMT_Body_3000x76afca0"/>
      <physvolref ref="LiquidSkinXenonBank0x76058b0"/>
    </bordersurface>
  </structure>
```

From [the geant4 application guide](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/physicsProcess.html#specifying-the-surface):

> There are three methods of specifying an optical surface.
> 
> 1. For the simple case of a perfectly smooth interface between two dielectric materials, all the user needs to provide are the refractive indices (RINDEX) of the two materials stored in their respective G4MaterialPropertiesTables.
> 2. A skin surface is the surface entirely surrounding a logical volume. This is useful where a volume is coated with a reflector and placed into many different mother volumes. A limitation is that the skin surface can only have one and the same optical property for all of the enclosed volume’s sides. An optical surface is created and defined using G4OpticalSurface. This surface is assigned to a logical volume using the class G4LogicalSkinSurface.
> 3. A border surface is defined by specifying the ordered pair of physical volumes touching at the surface. Because the pair of physical volumes is ordered, the user may specify different optical properties for photons arriving from the reverse side of the same interface. For the optical boundary process to use a border surface, the two volumes must have been positioned with G4PVPlacement. The ordered combination can exist at many places in the simulation. An optical surface is created and defined using G4OpticalSurface. This surface is assigned to the ordered pair of physical volumes using the class G4LogicalBorderSurface.

[...]

PR #1812, opened 2025-06-04
Add geometry surface definitions to runtime

sethrj:
This defines device-friendly storage, access, and setup for mapping between volumes and surface IDs. The key data structure to use in the downstream physics is the `VolumeSurfaceView`: from a pre-step volume ID, it gives the boundary surface of that volume, and lets you search from the pre-step volume instance to a post-step volume.

In our surface physics, we'll end up storing the pre-step volume and volume instance ID, and post-step VI ID, then looking through the boundary and interface surfaces accordingly.

The current implementation uses a linear search for the interface map lookup, so it will be slow for LHCb which has hundreds of pre->post pairs from the same volume.

Needed for #1351.

### DIFF
--- a/src/geocel/SurfaceData.hh
+++ b/src/geocel/SurfaceData.hh
@@ -1,0 +1,107 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file geocel/SurfaceData.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "corecel/data/Collection.hh"
+#include "corecel/data/CollectionBuilder.hh"
+#include "geocel/Types.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Store surface data corresponding to a volume.
+ *
+ * This stores information about the surfaces (both boundary and interface)
+ * of a volume. The \em boundary is a single optional surface ID, and the
+ * \em interface is an unzipped map \code (pre, post) -> surface \endcode.
+ *
+ * If \c interface_pre and \c interface_post are zipped, the result is \em
+ * sorted.  In other words, the pre-step surface can be searched with
+ * bisection, and the resulting subrange can also be searched with bisection to
+ * find the post-step surface. This then corresponds to the \c SurfaceId of
+ * that interface.
+ */
+struct VolumeSurfaceRecord
+{
+    //! Surface identifier for the volume boundary
+    SurfaceId boundary;
+
+    //! Sorted range of exiting volume instances (from this volume)
+    ItemRange<VolumeInstanceId> interface_pre;
+
+    //! Corresponding range of entering volume instances (to other volumes)
+    ItemRange<VolumeInstanceId> interface_post;
+
+    //! Surface IDs for the pre->post mapping
+    ItemRange<SurfaceId> interface;
+
+    //! True if valid data is present
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return boundary && !interface_pre.empty()
+               && interface_pre.size() == interface_post.size()
+               && interface_pre.size() == interface.size();
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Persistent data for mapping between volumes and their surfaces.
+ *
+ * This structure manages the relationship between volumes and their surfaces,
+ * particularly for tracking optical photons as they cross from one volume
+ * instance to another through shared surfaces.
+ *
+ * If no interface surfaces (aka 'border surface') are present then the backend
+ * storage arrays will be empty.
+ */
+template<Ownership W, MemSpace M>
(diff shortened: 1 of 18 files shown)


stognini:
Looking good. So next steps would be to load the imported surface data to `inp::` in `FrameworkInput` and expand `SurfaceParamsData` to store all the model and surface information, correct?

sethrj:
> Looking good. So next steps would be to load the imported surface data to `inp::` in `FrameworkInput` and expand `SurfaceParamsData` to store all the model and surface information, correct?

@stognini This gets loaded through the existing code in GeantGeoParams; the next step is adding these params to core and optical params and pulling them in during `setup`. Then the surface physics should be able to use it.

All *you* need to do for your surface physics is, given the `geant_geo` params:
1. Loop through `range(SurfaceId{geo.num_surfaces()})`
2. Run `geo.id_to_geant(surf_id)` to get `G4LogicalSurface const* g4surf`
3. Construct physics for each surface from `g4surf->GetSurfaceProperty()`

So you can actually do all of this right now before the PR; this one is just needed for @hhollenb 's work.

src/geocel/SurfaceData.hh
```diff
@@ -0,0 +1,110 @@
...
+    //! Corresponding range of entering volume instances (to other volumes)
+    ItemRange<VolumeInstanceId> interface_post;
+
+    //! Surface IDs for the pre->post mapping
+    ItemRange<SurfaceId> surface;
+
+    //! True if valid data is present
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return boundary && !interface_pre.empty()
```
amandalund: Should it be `boundary || ...`?

src/geocel/detail/SurfaceParamsBuilder.hh
```diff
@@ -0,0 +1,130 @@
...
+ * VolumeSurfaceData objects that map surfaces to volumes and their interfaces.
+ * It performs error checking against duplicate surface assignments.
+ */
+class SurfaceInputInserter
+{
+  public:
+    // Construct with pointers to target data and volume params
+    SurfaceInputInserter(VolumeParams const& volumes,
+                         std::vector<Label>* labels,
+                         std::vector<VolumeSurfaceData>* volume_surfaces);
```
amandalund: Are we moving towards using [mutable references rather than mutable pointers](https://celeritas-project.github.io/celeritas/user/development/style.html#function-inputs-and-outputs)?

src/geocel/VolumeSurfaceView.hh
```diff
@@ -0,0 +1,180 @@
...
+    // ID of the Volume
+    CELER_FORCEINLINE_FUNCTION VolumeId volume_id() const;
+
+    // ID of the boundary surface for this volume, if any
+    CELER_FORCEINLINE_FUNCTION SurfaceId boundary_id() const;
+
+    // Check if this volume has least one interface surface
+    CELER_FORCEINLINE_FUNCTION bool has_interface() const;
+
+    // Find surface ID for a transition to another volume instances
```
amandalund:
```suggestion
    // Find surface ID for a transition to another volume instance
```

src/geocel/VolumeSurfaceView.hh
```diff
@@ -0,0 +1,180 @@
...
+CELER_FUNCTION auto VolumeSurfaceView::volume_id() const -> VolumeId
+{
+    return volume_;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the optional boundary surface ID for this volume.
+
+ * If the result is null,
```
amandalund: 🐿️

src/geocel/VolumeSurfaceView.hh
```diff
@@ -0,0 +1,180 @@
...
+/*!
+ * Find the surface ID for a transition between volume instances.
+ *
+ * This searches for the surface ID associated with a pre->post
+ * volume instance transition.
+ * \todo The current implementation as a linear search, which is unsuitable for
+ * complex detectors such as LHCB's RICH, whose pvRichGrandPMTQuartz has 770
+ * specific interfaces. We should either implenent an \c equal_range function
+ * for searching these sorted arrays, or (better) use a hash lookup for {pre,
+ * post} -> surface.
```
amandalund:
```suggestion
 * \todo The current implementation is a linear search, which is unsuitable for
 * complex detectors such as LHCB's RICH, whose pvRichGrandPMTQuartz has 770
 * specific interfaces. We should either implement an \c equal_range function
 * for searching these sorted arrays, or (better) use a hash lookup for {pre,
 * post} -> surface.
```
Since we have `lower_bound` and `upper_bound`, seems like it should be pretty simple to make the search more efficient using the first option for now... any reason to hold off on that?

amandalund (approved):
Looks good to me!

src/geocel/VolumeSurfaceView.hh
sethrj: I wanted to get the correct implementation in first, but I'll update this now.

sethrj: Actually, it'll take a bit of work to update this. Maybe could I add to the optimization page? https://github.com/celeritas-project/celeritas/wiki/Potential-optimizations

PR #1814, opened 2025-06-06
Add options to forcibly disable Cherenkov and scintillation

amandalund:
Currently we always enable Cherenkov in the optical loop and enable scintillation if the necessary optical properties are present. This makes it an input option instead.

### DIFF
--- a/src/celeritas/inp/Physics.hh
+++ b/src/celeritas/inp/Physics.hh
@@ -59,6 +59,15 @@ struct EmPhysics
  */
 struct OpticalPhysics
 {
+    //!@{
+    //! \name Optical photon generation
+
+    //! Generate Cherenkov photons
+    bool cherenkov{true};
+
+    //! Generate scintillation photons
+    bool scintillation{true};
+    //!@}
 };
 
 //---------------------------------------------------------------------------//
--- a/src/celeritas/inp/Physics.hh
+++ b/src/celeritas/inp/Physics.hh
@@ -62,6 +62,10 @@ struct OpticalPhysics
     //!@{
     //! \name Optical photon generation
 
+    /*! \todo Replace with a mapping of \c VolumeId to \c ScintillationPhysics
+     * or \c CherenkovPhysics
+     */
+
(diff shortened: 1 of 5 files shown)


src/celeritas/inp/Physics.hh
```diff
@@ -59,6 +59,15 @@ struct EmPhysics
...
 struct OpticalPhysics
 {
+    //!@{
+    //! \name Optical photon generation
+
+    //! Generate Cherenkov photons
+    bool cherenkov{true};
+
+    //! Generate scintillation photons
+    bool scintillation{true};
```
sethrj, line 73: Worth noting at some point that these will be replaced with objects at some point: probably a mapping `VolumeId -> ScintillationPhysics`

src/celeritas/setup/Problem.cc
```diff
@@ -335,12 +335,22 @@ auto build_optical_offload(inp::Problem const& p,
         !imported.optical_materials.empty(),
         << R"(an optical tracking loop was requested but no optical materials are present)");
 
+    CELER_ASSERT(p.physics.optical);
```
sethrj: This should be a validate as well, since the collector is built based on the presence of `optical_capacity`.

sethrj (approved):

pcanal:
> enable scintillation if the necessary optical properties are present. This makes it an input option instead.

This is an odd wording (and/or missing additional description) as this sounds like we are now allowing the user to enable scintillation "without the **necessary** optical properties" ... which sounds like a bad idea ... what am I missing?

amandalund:
Sorry, I was not very clear :). Currently scintillation is always enabled in any material where the user has provided the scintillation optical properties. Cherenkov is enabled in _any_ material with optical properties, because it only needs the refractive index, which is required for any optical material. We had no way of disabling scintillation without changing the detector definition, and no way of disabling Cherenkov at all. This change lets us disable them (but we still validate that all the necessary optical properties are present when enabled).

PR #1559, opened 2025-01-02
Prepare to update copyrights

sethrj:
Following https://github.com/spack/spack/pull/48352 and [open source best practices](https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects) we will removing the copyright year from our headers and reformatting it for more simplicity. I've also added some text to the top-level COPYRIGHT file and updated the generation script.

PR #1816, opened 2025-06-08
Add WLS model class, generator, and executor and support second WLS process

amandalund:
This converts the WLS params class into a model class, adds an executor, and splits the interactor into an interactor and a generator, which in the future will be used to generate WLS secondaries in the same way we generate Cherenkov and scintillation photons. The WLS secondary's time is now correctly set as the pre-step time plus the sampled delta time.


test/celeritas/optical/OpticalMockTestBase.cc
```diff
@@ -186,6 +188,15 @@ void OpticalMockTestBase::build_import_data(ImportData& data) const
         data.optical_materials[0].rayleigh.scale_factor = 1;
         data.optical_materials[0].rayleigh.compressibility
             = native_value_from(Compressibility{7.658e-23});
+        data.optical_materials[0].wls.mean_num_photons = 2;
+        data.optical_materials[0].wls.time_constant
+            = native_value_from(TimeSecond(1e-9));
+        // Reemitted photon energy range (visible light)
+        data.optical_materials[0].wls.component.x
+            = {1.65e-6, 2e-6, 2.4e-6, 2.8e-6, 3.26e-6};
```
sethrj, line 196: I feel a pang of regret for not using eV for the optical tracking loop 😅 (not like it matters in most cases). These are in MeV I guess? And does the spectrum below integrate to 1 over the energy range?

sethrj (approved):
LGTM! Thanks for the refactor. I'll wait for @whokion to take a look though (no rush).

test/celeritas/optical/OpticalMockTestBase.cc
amandalund: Yep, in MeV :). We use the spectrum to calculate the CDF, so it does get normalized during setup time.

src/celeritas/optical/Interaction.hh
```diff
@@ -36,8 +36,8 @@ struct Interaction
 
     Real3 direction;  //!< Post-interaction direction
     Real3 polarization;  //!< Post-interaction polarization
-    Span<TrackInitializer> secondaries;  //!< Emitted secondaries
     Action action{Action::scattered};  //!< Flags for interaction result
+    WlsDistributionData distribution;  //!< Data for generating WLS secondaries
```
whokion, line 40: This introduces an extra null entry for other optical interactors. I don’t have a better alternative at the moment to avoid this, but we might consider reorganizing the data structure to handle it more cleanly.

src/celeritas/optical/WavelengthShiftData.hh
```diff
@@ -10,12 +10,32 @@
...
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Input data for sampling WLS optical photons.
+ */
+struct WlsDistributionData
+{
+    size_type num_photons{};  //!< Sampled number of photons to generate
+    units::MevEnergy energy;
+    real_type time{};  //!< Pre-step time
```
whokion: The generation time for WLS photons should be set at the post-step point—i.e., the WLS photon’s time should be post_step_time + delta_time (or delayed time for remission).

src/celeritas/optical/WavelengthShiftData.hh
```diff
@@ -10,12 +10,32 @@
...
+    Real3 position{};
+    OptMatId material;
```
whokion, line 30: Unlike scintillation or Cerenkov photons, WLS photons are generated during stepping on the device. Would it be possible to update these information directly from the geometry state after execution?

src/celeritas/optical/interactor/WavelengthShiftGenerator.hh
```diff
@@ -0,0 +1,131 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with shared and state data.
+ */
+CELER_FUNCTION
+WavelengthShiftGenerator::WavelengthShiftGenerator(
+    NativeCRef<WavelengthShiftData> const& shared,
+    WlsDistributionData const& distribution)
+    : distribution_(distribution)
+    , sample_time_(real_type{1}
```
whokion, line 82: One additional consideration is the timing distribution for the delta-time (i.e., the pure emission delay). The sample_time_ should either follow an exponential distribution or be a fixed time offset, depending on the specified distribution type - which may require adding this parameter to the WlsMaterialRecord.

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -86,17 +56,19 @@ class WavelengthShiftInteractor
...
+    SimTrackView const& sim,
+    Real3 const& pos,
+    OptMatId const& mat_id)
+    : sample_num_photons_(shared.wls_record[mat_id].mean_num_photons)
 {
-    CELER_EXPECT(inc_energy_.value() > calc_cdf_.grid().front());
+    CELER_EXPECT(mat_id);
+
+    distribution_.energy = particle.energy();
+    distribution_.time = sim.time() - sim.step_length() / constants::c_light;
```
whokion: Again, the reference generation time for WLS photons is the time at which the original photon is absorbed (assuming I understand sim.time correctly).

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
```diff
@@ -108,67 +80,15 @@ WavelengthShiftInteractor::WavelengthShiftInteractor(
...
-    // Sample wavelength shifted optical photons
+    // Sample the number of photons generated from WLS.
     Interaction result = Interaction::from_absorption();
-    result.secondaries = {secondaries, num_photons};
-
-    IsotropicDistribution sample_direction{};
-    NonuniformGridCalculator calc_energy = calc_cdf_.make_inverse();
-    for (size_type i : range(num_photons))
+    size_type num_photons = sample_num_photons_(rng);
+    if (num_photons > 0)
```
whokion, line 88: In most typical cases, the average number of WLS photons is around one. Therefore, the probability of generating more than one photon in a given interaction is relatively small—approximately 25% in typical applications. An alternative approach could be to apply additional WLS photon generation only when the sampled number exceeds one (i.e., a mixed strategy). However, it’s unclear whether this added complexity would offer any performance benefits or logical advantages in the overall design.

src/celeritas/optical/model/WavelengthShiftModel.cu
```diff
@@ -0,0 +1,40 @@
+//---------------------------------*-CUDA-*----------------------------------//
+// Copyright 2024 UT-Battelle, LLC, and other Celeritas developers.
+// See the top-level COPYRIGHT file for details.
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
whokion:
```suggestion
//------------------------------- -*- C++ -*- -------------------------------//
// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
// SPDX-License-Identifier: (Apache-2.0 OR MIT)
```

whokion (commented):
The overall refactoring looks good to me, except for the handling of the resulting time for WLS photons. We could also consider adding support for WLS2—even as a dummy copy of WLS for now—to ensure compatibility with all features currently supported by Geant4.  Of course, the addition of WLS2 can be in a separate merge request.

src/celeritas/optical/Interaction.hh
amandalund: Yes, I think this might be moved out of the `Interaction` once we implement the WLS secondary generation. Just needed somewhere to store it for now.

src/celeritas/optical/WavelengthShiftData.hh
amandalund: Oops you are right, it's post-step!

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
amandalund: Good point, I do think there would be a benefit (and not too much added complexity) to initializing the first secondary directly in the parent's track slot.

src/celeritas/optical/model/WavelengthShiftModel.cu
amandalund: I think this is consistent with our other .cu files?

src/celeritas/optical/WavelengthShiftData.hh
amandalund: It depends on how we'll do the generation... if we create track initializers for the WLS secondaries after the post-step actions (and are careful about initializing the first secondary directly in the parent's slot) then we should be able to just store the number of photons to generate. If we want to move towards eliminating the track initializers and instead initializing tracks directly in the vacant slots from the distribution data, we'd need to store this information.

amandalund:
Yes, I think we could pretty easily support WLS2 even without the code duplication. Happy to either try adding that here or as a follow on.

sethrj:
Let's leave extensions for future work! We should validate what we have first. @whokion have you got any representative WLS problems of interest?

whokion:
> Let's leave extensions for future work! We should validate what we have first. @whokion have you got any representative WLS problems of interest?

The current unit tests, along with some extensions, may be sufficient for verifying basic WLS input and output behavior. More sophisticated testing—such as scenarios based on the DUNE X0ARAOUCA setup—would require boundary processes and WLS2  involving a sensitive detector (SiPM).

sethrj:
Cool! Let's wait then till those are in and tested.

src/celeritas/optical/model/WavelengthShiftModel.cu
sethrj: I think the text got simplified in #1559 ; the format should still be CUDA though. I guess there's some copy-pasting from old .cu files involved? 😅

amandalund: I guess so, I copy-pasted the optical Rayleigh model .cu :)

src/celeritas/optical/interactor/WavelengthShiftInteractor.hh
amandalund: Added a TODO for this for when we implement the secondary generation.

src/celeritas/optical/interactor/WavelengthShiftGenerator.hh
amandalund: Added a TODO for this as well.

amandalund:
@sethrj @whokion since supporting a second WLS process is trivial (all we were really missing was the data import) I just went ahead and did it here.

app/celer-sim/simple-driver.py
```diff
@@ -42,6 +42,8 @@
     'optical': {
         'absorption': True,
         'rayleigh_scattering': True,
+        'wavelength_shifting': "exponential",
+        'wavelength_shifting2': "exponential"
```
whokion, line 46: These parameter names are somewhat confusing, as they don’t clearly reflect the underlying time structure of remission. Names like wls_delay_time or wls_remission_time might sound better for their meaning (?).  

I assumed these parameters were intended for use in the WavelengthShiftGenerator when sampling time, which doesn't appear to be implemented yet - or am I missing something?

whokion (approved):
This is a very nice extension for WLS2—thanks for introducing the model to avoid code duplication. I’ve left a minor comment regarding the sampling of remission times for WLS/WLS2.  Otherwise, good to merge!  Thanks.

app/celer-sim/simple-driver.py
amandalund: These are part of `GeantOpticalPhysicsOptions` and are used to both enable the wavelength shifting process and choose the time profile (the "none" option disables the process)---I think they were implemented this way to simplify things, but I agree it's a little confusing. Maybe having an additional struct like this would be clearer:

```c++
struct WavelengthShiftingOptions
{
    //! Enable the process
    bool enable{true};
    //! Select a model for sampling reemission time
    WLSTimeProfileSelection time_profile{WLSTimeProfileSelection::delta};

    //! True if the process is activated
    explicit operator bool() const { return enable; }
};
```

I have a little follow-on PR to support the fixed reemission time as well, so I could update this there.

whokion: No problem merging it as is and  extending the time sampling in a follow-up MR. Thanks!

amandalund:
@sethrj I'll wait for you to take another look before merging since I made some non-trivial updates.

sethrj (approved):
Excellent! I think the changes to support wls2 show the fragility of the process/import system but that's what we have so far 😅

PR #1819, opened 2025-06-09
Replace EXPECT_VEC_CLOSE with soft comparator

sethrj:
Small commit I forgot to open a PR for a while back: it just simplifies some of the macros and allows us to extend the vector/scalar comparisons.


amandalund (approved):
Thanks!

PR #1822, opened 2025-06-11
Support configurable WLS reemission time distribution

amandalund:
This adds support for the constant WLS reemission time and refactors the WLS Geant physics options slightly for clarity.


src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -93,7 +80,27 @@ constexpr bool operator==(ScintillationPhysicsOptions const& a,
 }
 
 //---------------------------------------------------------------------------//
-//! Optical Boundary process options
+//! Optical wavelength shifting process options
+struct WavelengthShiftingOptions
```
sethrj, line 84: is it feasible to start moving this stuff to an `inp`-like format? This struct could be replaced by `std::optional< WlsTimeProfile>` (or the original enum). Or are you planning to add more data to this?

src/celeritas/optical/interactor/WavelengthShiftGenerator.hh
```diff
@@ -122,7 +120,11 @@ CELER_FUNCTION TrackInitializer WavelengthShiftGenerator::operator()(Engine& rng
     result.polarization = ExitingDirectionSampler{0, result.direction}(rng);
 
     // Sample the delta time (based on the exponential relaxation)
-    result.time = distribution_.time + sample_time_(rng);
+    result.time
+        = distribution_.time
+          + (time_profile_ == WlsTimeProfile::delta
+                 ? time_constant_
+                 : ExponentialDistribution(real_type{1} / time_constant_)(rng));
```
sethrj, line 127: Wish we could use std::variant in device code 😅

sethrj (approved):
Looks good, but I'm not sureI understand the motivation of changing from an enum to a struct with an "enable" flag, which I don't like because it eliminates the one-to-one mapping from input to problem definition. That is, `{"enabled": false, "distribution": "delta"}` and `{"enabled": false, "distribution": "exponential"}` create a identical problem and we shouldn't allow a "distribution" if enabled is false.

amandalund:
> I'm not sure I understand the motivation of changing from an enum to a struct with an "enable" flag

My thinking was:
- It could be confusing to users (see https://github.com/celeritas-project/celeritas/pull/1816#discussion_r2138653893) that the process is enabled with a string that specifies the reemission time distribution
- It's consistent with the other optical physics options (e.g. `BoundaryPhysicsOptions`)
- I didn't like `none` in the `WlsTimeProfile` enum since that's not a valid option for the time distribution

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
amandalund: Yes, I do think we should start making these more `inp`-like... there's no other data that would go here that I'm aware of.

sethrj:
Fair enough! Future work can be more inp-like, let's push it out the door 🚢

PR #1823, opened 2025-06-12
Speed up CI by not autoupdating man db

sethrj:
The user manual job was taking 1.5 minutes to set up, apparently due to an IO-intensive step: see https://askubuntu.com/questions/272248/processing-triggers-for-man-db .  Sometimes that part can take 5 seconds, sometimes 50 seconds. So let's just skip it.


sethrj:
Apparently doesn't work with the ubuntu/whatever we have installed :)

PR #1824, opened 2025-06-12
Read and export Geant4 surface definitions

sethrj:
This maps `G4LogicalSkinSurface` and `G4LogicalBorderSurface` to a reproducible `SurfaceId` by sorting using the linearized Volume[Instance]Ids. It adds a test of translating surface information to and from surface IDs.

### DIFF
--- a/doc/implementation/geometry.rst
+++ b/doc/implementation/geometry.rst
@@ -87,11 +87,12 @@ Unique instance
 Surface
    A *surface* is defined as a contiguous area on the boundary of a volume,
    sometimes on only a single side of the volume. Note that this definition
-   different to the infinite surfaces of ORANGE and the surface frames of
+   differs from the infinite surfaces of ORANGE and the surface frames of
    VecGeom. Surfaces currently are defined in two ways: *Interface* surfaces
-   are one-directional surfaces defined as the interface from one
-   volume instance to another, called "border surfaces" in Geant4. *Boundary*
-   surfaces surrounding an entire volume and are called "skin" in Geant4.
+   ("border" surfaces in Geant4) are one-directional surfaces defined as the
+   interface from one volume instance to another. *Boundary* surfaces ("skin"
+   surfaces in Geant4) surround an entire volume, and their properties apply
+   symmetrically to tracks entering or exiting.
 
 
 .. [#cn] A *volume instance* has a one-to-one mapping for ``G4PVPlacement``,
--- a/doc/implementation/geometry.rst
+++ b/doc/implementation/geometry.rst
@@ -18,28 +18,38 @@ their enclosing parent volume.
 
 .. table:: Celeritas nomenclature tends toward computer science terminology.
 
-   +------------------+------------------------+----------------+--------------------+
-   | Celeritas        | Geant4                 | VecGeom        | KENO [#sc]_        |
-   +==================+========================+================+====================+
-   | (not used)       | Solid                  | Unplaced       | Shape              |
-   +------------------+------------------------+----------------+--------------------+
-   | Volume           | Logical volume         | Logical volume | Unit/array/media   |
-   +------------------+------------------------+----------------+--------------------+
-   | Volume instance  | Physical volume [#cn]_ | Placed volume  | Hole/array element |
-   +------------------+------------------------+----------------+--------------------+
-   | Child            | Daughter               | Daughter       | Hole/placement     |
-   +------------------+------------------------+----------------+--------------------+
-   | Parent           | Mother                 | Mother         | ---                |
-   +------------------+------------------------+----------------+--------------------+
+   +------------------+-------------------------+----------------+--------------------+
+   | Celeritas        | Geant4                  | VecGeom        | KENO [#sc]_        |
+   +==================+=========================+================+====================+
+   | (not used)       | Solid                   | Unplaced       | Shape              |
+   +------------------+-------------------------+----------------+--------------------+
+   | Volume           | Logical volume          | Logical volume | Unit/array/media   |
+   +------------------+-------------------------+----------------+--------------------+
+   | Volume instance  | Physical volume [#cn]_  | Placed volume  | Hole/array element |
+   +------------------+-------------------------+----------------+--------------------+
+   | Child            | Daughter                | Daughter       | Hole/placement     |
+   +------------------+-------------------------+----------------+--------------------+
+   | Parent           | Mother                  | Mother         | ---                |
+   +------------------+-------------------------+----------------+--------------------+
+   | Interface        | Border surface          | ---            | Hole/array element |
+   +------------------+-------------------------+----------------+--------------------+
+   | Boundary         | Skin surface            | ---            | Hole/placement     |
+   +------------------+-------------------------+----------------+--------------------+
+   | Surface          | Surface property [#sp]_ | ---            | ---                |
+   +------------------+-------------------------+----------------+--------------------+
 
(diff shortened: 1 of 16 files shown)


doc/implementation/geometry.rst
```diff
@@ -74,6 +84,16 @@ Unique instance
    and indirect children for each node.  Celeritas always uses 64-bit integers
    to store the ``VolumeUniqueInstanceId``.
 
+Surface
+   A *surface* is defined as a contiguous area on the boundary of a volume,
+   sometimes on only a single side of the volume. Note that this definition
+   different to the infinite surfaces of ORANGE and the surface frames of
```
amandalund:
```suggestion
   differs from the infinite surfaces of ORANGE and the surface frames of
```

doc/implementation/geometry.rst
```diff
@@ -74,6 +84,16 @@ Unique instance
...
+   VecGeom. Surfaces currently are defined in two ways: *Interface* surfaces
+   are one-directional surfaces defined as the interface from one
+   volume instance to another, called "border surfaces" in Geant4. *Boundary*
+   surfaces surrounding an entire volume and are called "skin" in Geant4.
```
amandalund: A little more documentation (not necessarily in this PR) on the interface/border and boundary/skin surfaces could be very helpful---e.g. something on the order of precedence for surfaces or even a table like in section 2.4 [here](https://www.fe.infn.it/u/paterno/Geant4_tutorial/slides_further/Physics/Dietz-Laursonn_2016_Peculiarities%20in%20the%20Simulationof%20Optical%20Physics%20with%20Geant4.pdf).
```suggestion
   volume instance to another, called "border surfaces" in Geant4. *Boundary*
   surfaces ("skin" surfaces in Geant4) surround an entire volume and their properties apply symmetrically to photons entering or exiting.
```

src/geocel/GeantGeoParams.hh
```diff
@@ -102,6 +112,14 @@ class GeantGeoParams final : public GeoParamsInterface,
     // Get the Geant4 logical volume corresponding to a volume ID
     G4LogicalVolume const* id_to_geant(VolumeId vol_id) const;
 
+    //// SURFACES ////
+
+    //! Get the number of surfaces (TODO: maybe live in surface params?)
+    SurfaceId::size_type num_surfaces() const { return surfaces_.size(); }
+
+    // Get the Geant4 logical volume corresponding to a volume ID
```
amandalund:
```suggestion
    // Get the Geant4 logical surface corresponding to a surface ID
```
And the definition below

src/geocel/GeantGeoParams.cc
```diff
@@ -131,6 +133,98 @@ std::vector<Label> make_physical_vol_labels(G4VPhysicalVolume const& world,
...
+        auto* lv = surf->GetLogicalVolume();
+#endif
+
+        {
+            CELER_ASSERT(lv);
+            auto vol_id = geo.geant_to_id(*lv);
+            CELER_ASSERT(vol_id);
+            auto iter_inserted = temp.insert({vol_id, surf});
+            CELER_ASSERT(iter_inserted.second);
+        }
```
amandalund:
```suggestion
        CELER_ASSERT(lv);
        auto vol_id = geo.geant_to_id(*lv);
        CELER_ASSERT(vol_id);
        auto iter_inserted = temp.insert({vol_id, surf});
        CELER_ASSERT(iter_inserted.second);
```
Redundant inner block

src/geocel/GeantGeoParams.hh
```diff
@@ -46,6 +46,15 @@ namespace celeritas
  * placements, a single PV may correspond to multiple spatial placements and is
  * disambiguated with \c ReplicaId , which corresponds to the PV's "copy
  * number".
+ *
+ * Each \c SurfaceId maps to a \c G4LogicalSurface instance, which is ether a
+ * \c G4LogicalBorderSurface (an "interface" surface between two volume
+ * instances) or a \c G4LogicalSkinSurface (a "boundary" surrounding a single
+ * logical volume). To ensure reproducible surface IDs across runs, we put
+ * boundaries before interfaces, and sort within each set by IDs (not by
```
amandalund:
```suggestion
 * boundaries before interfaces, and sort within each set by volume IDs (not by
```

amandalund (approved):
Nice next step!

doc/implementation/geometry.rst
sethrj: Britishisms sneak into my documentation every so often 😉

doc/implementation/geometry.rst
sethrj: Since the surface order precedence is a particular (and peculiar) implementation detail of the Geant4 optical boundary process, it isn't part of the "surface params" as such. If we wanted to, we could change the precedence of the boundaries at runtime or in different physics implementations.

doc/implementation/optical-physics.rst
```diff
@@ -107,24 +107,29 @@ optical materials.
...
+
+1. The interface surface from P0 to P1
+2. If L1 is the child of L0 (crossing into an "enclosed" volume), then the
+   boundary surface of L1
+3. The boundary surface of L0
+4. The boundary surface of L1
+5. The volumetric material properties of L0 and L1
+
+.. todo:: Once surface models are implemented, move the
+   precedence above into SurfacePhysics documentation.
```
amandalund, line 135: Nice!

sethrj:
@stognini wanna give this a quick look before I merge?

doc/implementation/optical-physics.rst
```diff
@@ -107,24 +107,29 @@ optical materials.
...
+Optical photons also have special interactions at material boundaries,
+specified largely by user-provided material properties. The surface
+definitions are translated from Geant4 "skin" and "border" surfaces to
+Celeritas "boundary" and "interface" surfaces, respectively (see
+:ref:`api_geometry`). The "boundary" of a volume is currently defined, from
+Geant4 input, as a *directional* property.
+
+Celeritas surface physics currently uses the following heuristic to reproduce
+Geant4 boundary physics behavior.  Given a pair of old→new volume instances
+P0→P1 corresponding to volumes L0→L1, the surface properties are determined in
```
stognini, line 124: Nitpick to make it more clear to less experienced users:
```suggestion
Celeritas surface physics currently uses the following heuristic to reproduce
Geant4 boundary physics behavior.  Given a pair of old→new physical volume instances
P0→P1 corresponding to logical volumes L0→L1, the surface properties are determined in
```

src/geocel/inp/Model.hh
```diff
@@ -98,6 +123,7 @@ struct Model
     // TODO: Materials
     // TODO: Regions
     Volumes volumes;
+    Surfaces surfaces;
```
stognini, line 126: In the new `inp` design, are we storing also physical properties here, or just ID mapping?
I considered using `OpticalPhysics` for the data, but then lots of the info is not really physics, but geometry. So I also considered storing some of the data in `Model` as well. So I think this looks good, and we can discuss some of this after this PR.

stognini (approved):
Thanks @sethrj ! This looks great.

src/geocel/inp/Model.hh
sethrj: _just_ geometry connectivity (ID mapping). All physics will be loaded into your surface physics input.

doc/implementation/optical-physics.rst
sethrj: I'm using celeritas terminology here

PR #1825, opened 2025-06-13
Rename solid angle class

sethrj:
LHCb's RICH detector requires the polar truncation of a sphere. This will need to create a cone rather than a wedge, and it will have different extents. Therefore I'll be adding a new class that's also a solid angle.

This is strictly renaming, no other changes.

### DIFF
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -137,8 +137,9 @@ EnclosedAzi enclosed_azi_from_poly(S const& solid)
  * Get the enclosed polar angle by a solid.
  */
 template<class S>
-EnclosedAzi make_wedge_polar(S const& solid)
+EnclosedAzi enclosed_pol_from(S const& solid)
 {
+    // FIXME: polar angle will be a different class
     return enclosed_azi_radians(solid.GetStartThetaAngle(),
                                 solid.GetDeltaThetaAngle());
 }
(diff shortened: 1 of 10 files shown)


amandalund (approved):
Looks good, though I think `EnclosedAzimuthal` might be worth the tradeoff in length for the added clarity.

sethrj:
You're ruining my dream of making our variable names compatible with FORTRAN IV...

amandalund:
You don't dream about being more like Geant4? `GimmeThePointsAndForgetThem`? `DefineSphericalAdjointSourceWithCentreAtTheCentreOfAVolume`??

sethrj:
🤣 ok we can name the functions that way, use six letter for variables, and use three letters for user input parameters since those can be encoded into a 16-bit integer.

sethrj:
Ooops sorry @amandalund , I intended to but did not commit  + push the change Azi -> Azimuthal. Maybe this is ok? The only other word that starts with `azi` is azide so there hopefully won't be any confusion. The question is if we keep this, do we use `pol` and `polar`?

amandalund:
No worries... I'm not particularly concerned about the meaning being ambiguous in this case (I don't think AZI is a common acronym either), I just have a slight preference for spelling out the full word when we can because I feel it improves readability/reduces mental overhead a bit. Consistency would suggest `pol`, but again I prefer `polar`. :)

PR #1826, opened 2025-06-13
Use variant set for uniform field input

amandalund:
This is a little update to the volume-specific field input (see https://github.com/celeritas-project/celeritas/pull/1659#discussion_r1987915912) I'd forgotten about.


sethrj (approved):
Nice!

PR #1827, opened 2025-06-17
Refactor solid Z slab as Truncated object

sethrj:
While working on the polar orange slice (ha!) I realized that the solid isn't the right place to put the Z slab: the only solid in Geant4 that has it is an ellipsoid; and we need a similar capability for SCALE. So I defined a `Truncated` class that allows up to 6 aligned planes to be added and applied.


src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1177,6 +1145,51 @@ void Parallelepiped::output(JsonPimpl* j) const
     to_json_pimpl(j, *this);
 }
 
+//---------------------------------------------------------------------------//
+// PLANEALIGNED
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with sense, axis, and position.
+ */
+PlaneAligned::PlaneAligned(Sense sense, Axis axis, real_type position)
```
elliottbiondo: The top level doc string says that an `IntersectRegionInterface` is a "non-reentrant spatial [region]." Thus I think this should be renamed `Halfspace` or `AlignedHalfSpace`.

src/orange/orangeinp/Solid.hh
```diff
@@ -21,12 +21,12 @@ namespace orangeinp
...
+ * Define the azimuthal truncation of a solid.
  *
  * This is a pie slice infinite along the z axis and outward from it. Its cross
  * section is in the \em x-y plane, and a start
  * angle of zero corresponding to the \em +x axis. An interior angle of one
- * results in no radial excluded in the resulting solid. A interior angle of
+ * results in no radial exclusion from the resulting solid. A interior angle of
  * more than 0.5 turns (180 degrees) results in a wedge being subtracted from
  * the solid, and an angle of less than or equal to 0.5 turns results in the
  * intersection of the solid with a wedge.
```
elliottbiondo, line 32: Not in this MR, but I think in line 36:

```
EnclosedAzi{Turn{-0.125}, Turn{0.25}};
```

The starting point should be -0.25

test/orange/orangeinp/IntersectTestResult.hh
```diff
@@ -0,0 +1,45 @@
...
+
+namespace celeritas
+{
+namespace orangeinp
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+
+struct IntersectTestResult
```
elliottbiondo, line 32: Can you document this a bit better and/or level it as a devops helper file?

test/orange/orangeinp/Truncated.test.cc
```diff
@@ -0,0 +1,105 @@
...
+        "el@trunc.pz",
+        "",
+        "el@trunc.mz",
+        "el@trunc",
+        "el",
+    };
+    static char const* const expected_volume_strings[] = {"all(-0, -1, +2)"};
+
+    auto const& u = this->unit();
+    //::celeritas::orangeinp::test::print_expected(u);
```
elliottbiondo: Do you mean to leave this in?

elliottbiondo (dismissed):

src/orange/orangeinp/IntersectRegion.cc
sethrj: Fair enough; though our plane surface is also effectively a half-space. I don't think halfspace is descriptive enough; we'd have to name it something like "axis-aligned  semi-infinite slab"...

src/orange/orangeinp/Solid.hh
sethrj: Actually no , starting at -0.25 and an interior of 0.25 would be the bottom right (southeast) quadrant; this means from SE to NE.

src/orange/orangeinp/IntersectRegion.cc
sethrj: `PlaneAlignedHalfspace`? Since `PlaneAligned<Axis>` is our templated class I wanted to keep the name consistent.

elliottbiondo: What information is communicated by "axis-aligned semi-infinite slab" that is not captured "AlignedHalfSpace"?

It looks like officially a half-space is bound by a plane; i.e. a quadric surface partitions R3, but each side isn't technically a half-space.

src/orange/orangeinp/Solid.hh
elliottbiondo: Oh yes, my mistake!

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: (My previous message was sent before I saw your last one, due to being interrupted by a meeting)

`PlaneAlignedHalfspace` is fine. My main comment is that is confusing to have a file full or different convex regions and have one of them designed as a plane.

sethrj: Actually I am going for `InfPlane` analogous with `InfWedge` etc.

sethrj: Didn't see your messages either, actually, lol... I think that since they are all convex regions the concept is conveyed clearly? The semi-infinite ones (wedge, plane) are also convex.

elliottbiondo: Sorry, yes, everything in this file is convex, and indeed a plane is convex too. I guess moreso I mean that everything in this file is a 3D shape with volume, whereas `PlaneAligned` and `InfPlane` are names that don't evoke something with volume.

elliottbiondo (approved):

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Just approved MR. I do think `Halfspace` is more descriptive than `Plane` but ultimately it doesn't matter too much.

sethrj: I think enough people think "halfspace" in the sense of _any_ quadric surface dividing the universe, so I see the term informally used to refer to surface plus sense.

Issue #1752, opened 2025-04-18
Implement on-device general tally/histogram


sethrj:
For #1751 , #809, #261 we will want some multi-dimensional (potentially ragged) data structures for binning tracks, scoring them and their moments, reducing and normalizing in parallel (threads + MPI processes), and efficiently transferring between CPU and GPU. We might be able to use some other software (ADIOS?) as a backend for reducing and I/O.

## Indexers

Continuous:
- One-dimensional nonuniform grid (for x, y, z, E, or t)
- One-dimensional uniform grid (optimization of 1D nonuniform, but common)
- One-dimensional uniform log space (optimization of 1D nonuniform)
- Particle attributes (charged vs neutral, em vs hadron, see https://github.com/celeritas-project/celeritas/issues/1594)

Discrete:
- Particle type
- Volume or region

Multi-dimensional:
- Multi-D regular grid (x * y * z)
- Multi-D ragged (particle type * energy per particle)
- Concatenated grid (e.g. LZ time grid, just an optimization for 1D nonuniform grid)

## Index quantity

Continuous:
- Track position
- Track kinetic energy
- Source position (aka primary vertex)
- Angle (either dot product with surface normal or absolute direction?)
- Custom, e.g. `(primary position, detector id) -> hypot(detector center - primary position)` for LZ (#1751)

Discrete:
- Particle type
- Track's pre-step volume
- (For surface detectors) post-step volume?
- Particle charge

Tallying quantities and estimators:
- Weighted path length (#809)
- Reaction rate (path length estimator)
- Reaction rate (collision estimator)
- Energy deposition
- Change in angle (?)
- Step count (for steps-per-track diagnostic)
- Number of secondaries
- Interacting process (post-step action)
- Loss in energy

Accumulation event:
- Lifetime (for steps-per-track diagnostic, observables)
- Surface crossing (optical detectors, or cell based detectors)
- Virtual surface crossing (mesh tally)
- Every step
- Every interaction

Accumulation type:
- Mean
- Mean + variance
- If we ever do variance reduction schemes, a bin structure that could resolve low-variance but accurately tally high-variance events

Combining dimensions:
- N-dimensional binning: arbitrary number of axes but optimize for a few?
- Ragged binning, e.g. different energy bins for different particle types; perhaps as sub-dimensions (i.e. two concatenated uniform grids would have sub-indices (0, 0..4) and (1, 0..9); the particles could have different sub-dimensions).
[...]

PR #1828, opened 2025-06-19
Add in-situ cartesian field map construction from Geant4

esseivaju:
Implements `CartMapMagneticField` as a `G4MagneticField` wrapper around Celeritas `CartMapField`. Includes:

  - `CartMapMagneticField` class using pimpl to hide C++20 dependencies
  - `MakeCartMapFieldInput` function to query and build Celeritas field from an existing `G4Field`
  - Refactored `CylMapMagneticField` to use shared field sampling utilities
  - Added `covfie` dependency to accel


src/accel/CartMapMagneticField.hh
```diff
@@ -0,0 +1,81 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+//! POD struct for CartMap field grid parameters
+struct CartMapFieldGridParams
+{
+    G4double min_x{};  //!< Minimum X coordinate
+    G4double max_x{};  //!< Maximum X coordinate
+    size_type num_x{};  //!< Number of grid points in X direction
```
sethrj: Should we be using G4int for these, or real_type for the coordinates? Would it make sense to define a helper struct and just do:
```c++
struct Uniform { double min; double max; int num; };
Uniform x;
Uniform y;
Uniform z;
```

so they do `x.min` rather than `min_x?

src/accel/CartMapMagneticField.hh
```diff
@@ -0,0 +1,81 @@
...
+    // Destructor
+    ~CartMapMagneticField() override;
+
+    // Calculate values of the magnetic field vector
+    void GetFieldValue(G4double const point[3], G4double* field) const override;
+
+  private:
+    // Forward declaration for PIMPL
+    struct Impl;
+    std::unique_ptr<Impl> pimpl_;
```
sethrj: You'll want to delete the move operator if you're defining this way.

src/accel/detail/MagneticFieldUtils.hh
```diff
@@ -0,0 +1,80 @@
...
+ *                        native units in the correct coordinate space. Must
+ *                        have signature: void(Array<G4double, 3> const&,
+ *                        real_type*) taking G4 field [Bx, By, Bz] and writing
+ *                        converted values to output pointer
+ */
+template<typename PositionCalc, typename FieldConverter>
+inline void setup_and_sample_field(real_type* field_data,
+                                   Array<size_type, 4> const& dims,
+                                   PositionCalc const& position_calculator,
+                                   FieldConverter const& field_converter)
```
sethrj: Instances should be named like nouns:
```suggestion
                                   PositionCalc const& calc_position,
                                   FieldConverter const& convert_field)
```

src/accel/CartMapMagneticField.cc
```diff
@@ -0,0 +1,152 @@
...
+{
+    // Validate input parameters
+    CELER_VALIDATE(params, << "invalid CartMapFieldGridParams provided");
+
+    CartMapFieldParams::Input field_input;
+
+    // Convert from Geant4 units to native units
+    field_input.min_x = convert_from_geant(params.min_x, clhep_length);
+    field_input.max_x = convert_from_geant(params.max_x, clhep_length);
+    field_input.num_x = params.num_x;
```
sethrj: Maybe we should make the field input use a "uniform grid" input struct too...

src/accel/CartMapMagneticField.cc
```diff
@@ -0,0 +1,152 @@
...
+    G4double const dy = (params.max_y - params.min_y) / (params.num_y - 1);
+    G4double const dz = (params.max_z - params.min_z) / (params.num_z - 1);
+
+    // Position calculator for Cartesian grid
+    auto position_calculator = [&](size_type ix, size_type iy, size_type iz) {
+        G4double x = params.min_x + ix * dx;
+        G4double y = params.min_y + iy * dy;
+        G4double z = params.min_z + iz * dz;
+        return Array<G4double, 4>{x, y, z, 0};
+    };
```
sethrj: Future work: https://github.com/celeritas-project/celeritas/issues/1752

sethrj (approved):
Looks great! This is a nice addition. @rahmans1 @sanmayphy this should be useful in your framework integration.

src/accel/CartMapMagneticField.hh
esseivaju: Refactored to use a custom deleter so that we can keep the move semantics

src/accel/CartMapMagneticField.hh
```diff
@@ -59,20 +45,31 @@ class CartMapMagneticField : public G4MagneticField
...
 
+  private:
+    // Forward declaration for pImpl
+    struct Impl;
+
+    // Custom deleter for pImpl
+    struct ImplDeleter
+    {
+        void operator()(Impl* ptr) const;
+    };
```
sethrj, line 61: Put private stuff after public unless absolutely necessary; it makes the code easier to read.

src/accel/CartMapMagneticField.hh
```diff
@@ -59,20 +45,31 @@ class CartMapMagneticField : public G4MagneticField
...
+
   public:
     // Construct with CartMapFieldParams
     explicit CartMapMagneticField(SPConstFieldParams field_params);
 
+    // Default move semantics work with custom deleter
+    CELER_DEFAULT_MOVE_DELETE_COPY(CartMapMagneticField);
```
sethrj: You should be able to delete this and the default destructor and it'll be the same.

sethrj (approved):
Thanks!

sethrj:
I did not do a thorough job with this review. I know we were in a rush to get this in, but this code is virtually untested. I can't edit any of it for fear of breaking something. Our helper functions are using different units internally. ☹️

esseivaju:
Did you spot a bug where we have a unit mismatch? I can take a look at adding unit test but since we're getting the field from Geant4, some part would require a mock Geant4 run

sethrj:
~~While working on #2209 I needed to touch `convert_to_geant` and `convert_from_geant`. This led me down a rabbit hole into the field code since I couldn't test it at all, and I think I see an order-of-10 magnitude error *when we use CGS* but not when we use CLHEP...~~ 

Never mind, I am an idiot and it was a problem with my test: I had the wrong units for the scale factor (neglecting a 1/length).

PR #1830, opened 2025-06-20
Optimize cosine for turn

sethrj:
This is a stupid minor change while I was diddling with some other Turn related stuff; as a diversion I tried to improve the integer sin/cos theta to use simple instructions rather than a static array lookup.

https://github.com/sethrj/testsnippets/commit/87d733274d90fe83c0c106ca5d766e1a2d6eba0e

### DIFF
--- a/src/corecel/math/Turn.hh
+++ b/src/corecel/math/Turn.hh
@@ -95,8 +95,16 @@ CELER_FORCEINLINE_FUNCTION void sincos(Turn_t<T> r, T* sinv, T* cosv)
 
 CELER_CONSTEXPR_FUNCTION int cos(IntQuarterTurn r)
 {
-    constexpr int cosval[] = {1, 0, -1, 0};
-    return cosval[(r.value() > 0 ? r.value() : -r.value()) % 4];
+    // Get the last two bits
+    auto i = std::abs(r.value()) & 0x3;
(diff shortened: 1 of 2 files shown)


amandalund (approved):
Can't say it improves readability, but it's clever 😁

src/corecel/math/Turn.hh
```diff
@@ -95,14 +95,23 @@ CELER_FORCEINLINE_FUNCTION void sincos(Turn_t<T> r, T* sinv, T* cosv)
 
 CELER_CONSTEXPR_FUNCTION int cos(IntQuarterTurn r)
 {
-    constexpr int cosval[] = {1, 0, -1, 0};
-    return cosval[(r.value() > 0 ? r.value() : -r.value()) % 4];
+    // Cosine is symmetric and periodic: modulo with 4
+    // (note: avoid std::abs which isn't constexpr till C++23)
+    auto i = (r.value() > 0 ? r.value() : -r.value()) % 4;
+    // Map to {1, 0, -1, 0}[i] by encoding the 2-bit values into an int
+    // offset by one: {2, 1, 0, 1}. Since ints are written big endian,
```
pcanal, line 102: The text can be interpreted as hinting that this code might not work on a little endian machine (a priori it should work), is there a better way to phrase this?

PR #1831, opened 2025-06-22
Use start/stop instead of start/interior for azimuthal wedge

sethrj:
Chained on #1827 , more stupid refactoring to try to get the polar wedge right. This changes the azimuthal slice for a solid to use a start/stop range rather than start/interior, since the polar range is on a non-periodic boundary.


sethrj:
@elliottbiondo This one next please! And remember, "request changes" is "reject unless revised".

src/orange/g4org/SolidConverter.cc
```diff
@@ -81,16 +81,17 @@ namespace
  * This constructs from native Geant4 radians and truncates to \c real_type,
  * ensuring that roundoff doesn't push the turn beyond a full one.
  */
-EnclosedAzi enclosed_azi_radians(double start_rad, double delta_rad)
+auto enclosed_azi_radians(double start_rad, double stop_rad)
```
elliottbiondo, line 84: Missing `-> EnclosedAzi`

elliottbiondo (approved):

src/orange/g4org/SolidConverter.cc
sethrj: Actually with C++17 you don't even need it! :)

elliottbiondo: Sure, but don't you think it's nice to be able to tell what the return type is by just looking at the definition?

sethrj: True... I'll consider that going forward!

Issue #1473, opened 2024-10-29
LHCb Integration


sethrj:
This is the primary tracking issue for integrating Celeritas with the simulation workflow for [LHCb](). The primary Celeritas contact points are the "assignees" to the right.

# Core capabilities

- [ ] History parentage bookkeeping for tracker region (?)
- [ ] Magnetic field?

# Geometry

- [ ] Sphere with polar limits

# Extras

- [ ] Enable Cherenkov on a per-volume basis (Geant4 UR-92), though could this be set as per-region physics? (Or do we give an extra option that adds new volume-related physics toggles and expand in region with the capability @amandalund added in  #2023 ?)

PR #1832, opened 2025-06-22
Add polar truncation for sphere shape

sethrj:
Chained on #1831.

This adds the ability to truncate a sphere's polar region to a particular set of latitudes. This is used by LHCb's RICH detector (#1473 )

### DIFF
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -85,7 +85,11 @@ auto enclosed_azi_radians(double start_rad, double stop_rad)
 {
     auto start = native_value_to<RealTurn>(start_rad);
     auto stop = native_value_to<RealTurn>(stop_rad);
-    if (soft_equal(stop.value() - start.value(), real_type{1}))
+    auto delta_turn = value_as<RealTurn>(stop - start);
+    CELER_VALIDATE(delta_turn <= 1 || soft_equal(delta_turn, real_type{1}),
+                   << "azimuthal restriction [" << start.value() << ", "
+                   << stop.value() << "] [turn] exceeds 1 turn");
+    if (delta_turn >= real_type{1})
     {
         // Avoid roundoff error: return full region, *but* keep orientation,
         // needed for polyhedra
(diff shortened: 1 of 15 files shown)


sethrj:
@elliottbiondo Finally ready!

src/orange/g4org/SolidConverter.cc
```diff
@@ -94,6 +94,27 @@ auto enclosed_azi_radians(double start_rad, double stop_rad)
...
+ * This constructs from native Geant4 radians and truncates to \c real_type,
+ * ensuring that roundoff doesn't push the turn beyond a full one. The
+ * G4Sphere::CheckThetaAngles implementation prevents the endpoint being
+ * greater than 180 degrees, so we do the same here.
+ */
+auto enclosed_polar_radians(double start_rad, double stop_rad)
+{
+    auto start = native_value_to<RealTurn>(start_rad);
+    auto stop = native_value_to<RealTurn>(stop_rad);
+    if (stop > RealTurn{0.5})
```
elliottbiondo: Seems like there should be a DBC to check that stop isn't something way off, like 0.6

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1033,6 +1041,67 @@ void InfAziWedge::output(JsonPimpl* j) const
...
+//---------------------------------------------------------------------------//
+// INFPOLARWEDGE
+//---------------------------------------------------------------------------//
+/*!
+ * Construct from a starting angle and stop angle.
+ */
+InfPolarWedge::InfPolarWedge(Turn start, Turn stop)
+    : start_{start}, stop_{stop}
+{
+    CELER_VALIDATE(start_ >= zero_quantity() && start_ < Turn{0.5},
```
elliottbiondo: south_pole

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -572,6 +572,43 @@ class InfAziWedge final : public IntersectRegionInterface
     Turn stop_;
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Use a cone and plane to select a polar (latitudinal) region.
```
elliottbiondo: Clarify that it could be two cones, not a cone in a plane. The plane is really the special case, even if that's what is used most often.

src/orange/orangeinp/Solid.hh
```diff
@@ -70,6 +71,54 @@ class EnclosedAzi
...
+ *
+ * This subtracts up to two infinite cones centered along the z axis from the
+ * origin.
+ *
+ * A start angle of zero corresponding to the \em +z axis. An interior angle of
+ * 0.5 results in no exclusion from the resulting solid.
+ * \code
+  // Truncates a solid to the top hemisphere
+  EnclosedPolar{Turn{0}, Turn{0.25}};
+  // Truncates a solid to the equatorial region (18 degrees N/S)
```
elliottbiondo: Specify that this is the two-cone case

elliottbiondo (approved):

src/orange/orangeinp/IntersectRegion.hh
sethrj: The intersect region is always one plane for the equator, and one cone (except for the degenerate case of a polar endpoint).

PR #1834, opened 2025-06-24
Flush optical tracks at the end of core stepping loop

amandalund:
While [updating the LAr sphere optical benchmark](https://github.com/celeritas-project/benchmarks/commit/8fb39e0b92aa08083d35c9fb76e046c76f6f31af) to account for some recent optical physics changes, I realized we don't flush the remaining optical tracks at the end of the core stepping loop: this fixes that.

### DIFF
--- a/app/celer-sim/Transporter.cc
+++ b/app/celer-sim/Transporter.cc
@@ -166,7 +166,8 @@ auto Transporter<M>::operator()(SpanConstPrimary primaries)
     append_track_counts(track_counts);
     record_step_time();
 
-    while (track_counts)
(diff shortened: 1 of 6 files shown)


sethrj (approved):
Definitely a step in the right direction, but I wonder if we should have a more direct way of flushing the optical loop

amandalund:
Yep agreed we should. I think all of this optical launch/generation code is still very in flux 🙃

sethrj:
I talked with @lebuller about slightly refactoring the optical launch action so we could call it directly from the unit test; I think that's the next step in this directino.

PR #1836, opened 2025-06-26
Output optical actions

amandalund:
This tiny update lets us write the optical actions (and processes) that are enabled to the output.


sethrj (approved):
Nice!

Issue #1810, opened 2025-06-03
Map track IDs between Geant4 and Celeritas


sethrj:
As part of #1746 we want to restore as much "MC truth" information as possible. We want to:
- Map the Geant4 track ID to the Celeritas primary ID when buffering (need new data structure) #1837
- When creating a hit, use the track ID of the primary (?) #1837
- Add an interface to call back to the framework that we're adding a Celeritas-created secondary from a particular Geant4 track ID

esseivaju:
> * Add an interface to call back to the framework that we're adding a Celeritas-created secondary from a particular Geant4 track ID

Do we want to add a callback to the Geant4 Tracking/Stacking action for secondaries here, or rather our own callback mechanism?

sethrj:
I was thinking specifically geant tracking action for setting up MC truth data.

esseivaju:
ATLAS currently uses stacking actions for truth linking, see [here](https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/Simulation/G4Utilities/G4UserActions/src/AthenaStackingAction.cxx#0119) where the track user info is set.

sethrj:
> // TODO: Why is this here? Can I remove it?
> // TODO What does this condition mean?

😅

esseivaju:
> > // TODO: Why is this here? Can I remove it?
> > // TODO What does this condition mean?
> 
> 😅

Here be dragons 🐉

esseivaju:
I'll close this, we have #1745 to track callback to the user framework

PR #1837, opened 2025-06-27
Reconstruct Geant4 TrackID

esseivaju:
Use the primary's parent TrackID to keep track of the source track, e.g. Geant4 when using `accel`. Then, assign the TrackID when reconstructed in the `HitProcessor`


src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -380,14 +380,20 @@ void HitProcessor::operator()(DetectorStepOutput const& out, size_type i) const
...
+    ParticleId id = out.particle[i];
     CELER_EXPECT(id < tracks_.size());
     G4Track& track = *tracks_[id.unchecked_get()];
     step_->SetTrack(&track);
 
     G4ParticleDefinition const& pd = *track.GetParticleDefinition();
 
+    if (!out.parent_id.empty())
+    {
+        track.SetTrackID(out.parent_id[i].unchecked_get());
```
esseivaju: I suppose there would be an issue there when reconstructing hit from a secondary, in which case the parent id points to Celeritas' TrackID

amandalund: Just so I understand the intent, are we trying to associate the reconstructed Geant4 track with the original offloaded track/Celeritas primary that it came from? As opposed to trying to assign it a unique ID associated with the internal Celeritas track ID (which I guess is not currently possible because of our non-reproducible track IDs).

src/celeritas/track/detail/ProcessPrimariesExecutor.hh
```diff
@@ -63,6 +63,7 @@ CELER_FUNCTION void ProcessPrimariesExecutor::operator()(ThreadId tid) const
     TrackInitializer ti;
     ti.sim.track_id
         = make_track_id(params->init, state->init, primary.event_id);
+    ti.sim.parent_id = primary.track_id;
```
amandalund: I think our parent IDs should always be either an invalid ID (primary) or an ID associated with a Celeritas track, not a Geant4 track. Maybe we need to store some additional data mapping the Geant4 track ID to the Celeritas primary?

src/celeritas/ext/detail/HitProcessor.cc
esseivaju: Right, as part of #1810 we want to preserve the original Geant4 TrackID of offloaded primaries when reconstructing the track. I suppose it would be fine if internally Celeritas still uses its own primary ID as long as we can map it back to the Geant4 TrackID

src/celeritas/track/detail/ProcessPrimariesExecutor.hh
esseivaju: Ah is that because an invalid ID is how you tell that a Track Initializer is a primary? We could have a mechanism similar to `AuxStateData` to attach arbitrary information to a Primary but that could add quite a bit of memory overhead..

src/celeritas/phys/Primary.hh
```diff
@@ -27,6 +27,7 @@ struct Primary
     Real3 direction{0, 0, 0};
     real_type time{};
     EventId event_id;
+    TrackId track_id;
```
sethrj: No, this is different from the Celeritas track ID (in fact we previously had this field and removed it). If we want to add a field here it should be `int geant_primary_id{-1};`. Or we could add a `PrimaryId` and then as you say keep a host-only `aux` data mapping to Geant4.

src/celeritas/track/detail/ProcessPrimariesExecutor.hh
sethrj: We could add a `primary_id` field to SimData to easily propagate this information. (We could even make it optional if we want to optimize for future non-g4 simulations.) The `ExtendFromPrimariesAction` can store aux information (a vector of [primary id] -> [G4 track ID]) that can be referenced by the hit processor.

sethrj (changes requested):
This changes the meaning of parent_id, which is currently used to store the relationship between a secondary track and the one that created it:
https://github.com/celeritas-project/celeritas/blob/47f47c8892e00066bfe66e3a8390b424beb29015/src/celeritas/track/SimTrackView.hh#L151

which is needed for reconstructing MC truth. What we need is a new property for storing the original primary ID and mapping it to geant4.

esseivaju:
Okay I have reworked this to use an extra field to store the originating `PrimaryID`. The `PrimaryID` is propagated down to secondaries when initializing new tracks. 

I have made the simplification, that the `PrimaryID` is set to the Geant4 TrackID in accel, which removes the need for a [primary id] -> [G4 track ID] lookup data structure, the Geant4 TrackID is propagated down to the secondaries.

esseivaju:
Alternatively, if we want to have separate Celeritas PrimaryId, and Geant4 TrackId 
5dc143a stores the lookup vector in HitProcessor. This seems better than using `aux` state data because `HitProcessor` and `LocalTransporter` are both geant4-specific, while `ExtendFromPrimariesAction` isn't and can just transparently propagate primary ID regardless of wether it's a Geant4 trackId, or another user-defined id.

src/accel/LocalTransporter.cc
```diff
@@ -259,6 +269,16 @@ void LocalTransporter::Push(G4Track const& g4track)
...
     PDGNumber const pdg{g4track.GetDefinition()->GetPDGEncoding()};
     track.particle_id = particles_->find(pdg);
+
+    // Generate Celeritas-specific PrimaryID and notify HitProcessor of mapping
+    PrimaryId celeritas_primary_id{next_celeritas_primary_id_++};
+    track.primary_id = celeritas_primary_id;
+    if (hit_processor_)
+    {
+        hit_processor_->register_primary_id_mapping(celeritas_primary_id,
+                                                    g4track.GetTrackID());
```
sethrj: I think `[[nodiscard]] PrimaryId register_primary(G4Track const&)` would be better: the hit processor keeps track of IDs (instead of having to keep the sizes and counters in two places), and the track reference allows us to later extend the hit processor to do other mapping (weight, user track info,  other MC truth, ...).

src/accel/LocalTransporter.cc
```diff
@@ -220,6 +223,13 @@ void LocalTransporter::InitializeEvent(int id)
     event_id_ = id_cast<UniqueEventId>(id);
     ++run_accum_.events;
 
+    // Clear PrimaryID mapping in HitProcessor and reset counter for new event
+    if (hit_processor_)
+    {
+        hit_processor_->clear_primary_id_mapping();
```
sethrj:
```suggestion
        hit_processor_->begin_event();
```

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -226,6 +229,32 @@ HitProcessor::~HitProcessor()
...
+                                               int geant4_track_id)
+{
+    auto id = celeritas_id.unchecked_get();
+    if (id >= celeritas_to_g4_track_id_.size())
+    {
+        celeritas_to_g4_track_id_.resize(
+            std::max(celeritas_to_g4_track_id_.size() * 2,
+                     static_cast<std::size_t>(id + 1)));
+    }
+    celeritas_to_g4_track_id_[id] = geant4_track_id;
```
sethrj: This seems like overkill since `vector` automatically manages capacities to amortize costs. If we update to use the signature I recommended, this could just be 
```c++
auto primary_id = id_cast<PrimaryId>(c2g4ti_.size());
celeritas_to_g4_track_id_.push_back(g4track->GetTrackId());
return primary_id;
```

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -380,14 +409,23 @@ void HitProcessor::operator()(DetectorStepOutput const& out, size_type i) const
...
     CELER_EXPECT(id < tracks_.size());
     G4Track& track = *tracks_[id.unchecked_get()];
     step_->SetTrack(&track);
 
     G4ParticleDefinition const& pd = *track.GetParticleDefinition();
 
+    if (!out.primary_id.empty())
+    {
+        PrimaryId celeritas_primary_id = out.primary_id[i];
+        CELER_EXPECT(celeritas_primary_id < celeritas_to_g4_track_id_.size());
```
sethrj:
```suggestion
        CELER_ASSERT(celeritas_primary_id < celeritas_to_g4_track_id_.size());
```
for internal consistency checks

sethrj (commented):
Looks good! A fair bit of bookkeeping, we should probably replace the "detector steps" stuff with something simpler that always has all the data...

esseivaju:
Are we ready on this one? I suppose any refactor to the "detector steps" infrastructure would come in a subsequent PR.

sethrj:
Sorry @esseivaju , this one slipped my mind.

sethrj (approved):
Nice, thanks!

PR #1838, opened 2025-06-28
Add surface properties to core and optical params

amandalund:
I haven't added the `VolumeParams` as it's currently only used to construct the surfaces, but I can if needed.

### DIFF
--- a/src/geocel/VolumeSurfaceView.hh
+++ b/src/geocel/VolumeSurfaceView.hh
@@ -72,7 +72,7 @@ VolumeSurfaceView::VolumeSurfaceView(SurfaceParamsRef const& params,
                                      VolumeId id)
     : params_(params), volume_(id)
 {
-    CELER_EXPECT(id < params.volume_surfaces.size());
+    CELER_EXPECT(id < params.volume_surfaces.size() || !params_);
 }
 
 //---------------------------------------------------------------------------//
@@ -92,6 +92,11 @@ CELER_FUNCTION auto VolumeSurfaceView::volume_id() const -> VolumeId
  */
 CELER_FUNCTION SurfaceId VolumeSurfaceView::boundary_id() const
 {
+    if (!params_)
+    {
+        // No surfaces are present
+        return {};
+    }
     return this->volume_record().boundary;
 }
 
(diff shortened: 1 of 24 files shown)


src/geocel/VolumeSurfaceView.hh
```diff
@@ -72,7 +72,7 @@ VolumeSurfaceView::VolumeSurfaceView(SurfaceParamsRef const& params,
                                      VolumeId id)
     : params_(params), volume_(id)
 {
-    CELER_EXPECT(id < params.volume_surfaces.size());
+    CELER_EXPECT(id < params.volume_surfaces.size() || !params_);
```
sethrj: I don't think we should be allowed to create a volume surface view without having any surfaces 🤔

sethrj (commented):
Awesome! Maybe I'm missing something about the surface view construction though. The only reason we should create a surface volume view (right now) is for surface physics, and surface physics need surfaces to be defined.  When would we want to create a surface view but not know whether we have surfaces?

Also indeed don't worry about the volume params for now, but we will probably need that when we want to reconstruct "unique volume IDs".

src/geocel/VolumeSurfaceView.hh
amandalund: My reasoning was we could still simulate optical photons without any optical surfaces (e.g. our simple LAr sphere), but we'd need to check for a surface when crossing a boundary. Or did we land on always requiring a volume with optical properties to have a surface?

sethrj: In [the notes from our hacking friday](https://github.com/celeritas-project/celeritas/pull/1801#issuecomment-3014023505) I have "if no optical surface ID at a boundary crossing, we kill the track". Based on that, I think we should require SurfaceParams if optical physics is enabled. (And during construction, we can validate that surfaces are defined, and/or we can warn when volumes that are associated with optical materials have no optical surfaces. Or we just create default skin surfaces with no physics for volumes that have no optical materials.) Since the LAr sphere is an edge case, we shouldn't focus on supporting it.

amandalund: Sounds good, I'll warn and add invalid surface IDs if none are defined.

sethrj (approved):
Great, thanks!

Issue #1815, opened 2025-06-07
Refactor volume management to improve mapping across the codebase


sethrj:
Currently there are two ways to match properties between `ImportData` and problem input (as well as across geometry implementations), mostly handled by `GeantVolumeMapper`:
- The `GeoParamsInterface::find_volume(G4LogicalVolume const*)` method
- Searching for a label

The searching by label is done to support the ability to load through VecGeom GDML and via a separate ORANGE geometry, both of which are complicated by:
- Duplicate volume names inside a geometry (CMS)
- Changing pointer addresses across runs or from GDML to in-memory Geant4

Going from a volume/volume instance ID to a geant4 pointer, needed for touchable reconstruction, can also be done with:
- `GeoParamsInterface::id_to_geant(VolumeInstanceId id)` (which right now has weirdness due to replica volumes)
- `GeantGeoParams::id_to_geant(VolumeId vol_id)`

Geometry labels are currently made via `make_logical_vol_labels` and `make_physical_vol_labels` which use `detail::make_label_vector` to use *occurrences* in the ordered hierarchy rather than *pointers* to construct unique labels. It requires that the list of daughters is ordered  consistently between VecGeom and Geant4 (guaranteed by G4VG). It does not apply to ORANGE which isn't constructed using the same hierarchy except as an implementation detail in g4org.

## Proposed new implementation

The `inp::Model` will now contain a `Volumes` struct that embeds the graph of nodes (G4LV) and edges (G4PV + copy number) and can be used to index between them. Eventually the main geometry classes will not reference Geant4 objects *or* labels: such metadata will be handled by GeantGeoParams and VolumeParams, respectively.

This will enable the solution of:
- #1749 by providing a geometry-agnostic input for mapping internal volume IDs to global volume IDs
- #1748 by calculating and storing the subgraph sizes and mapping to and from physical ID stacks
- #1351 by mapping volume instance IDs (used by boundary surfaces) to volume IDs
- The touchable ID uniquification https://github.com/celeritas-project/celeritas/pull/1649

Steps:
1. Import volumes (g4lv*) and instances (g4pv*, copy number) from Geant4 into the new VolumeParams class. Initially for simplicity, keep the volume IDs consistent with the old volume IDs (which may have gaps due to 'unused' LV/PV) and use duplicate instances for problems with replica IDs. (#1821)
2. Use this class to set up SurfaceParams for use in optical physics. (#1824)
[...]

PR #1839, opened 2025-06-30
Define geometry implementation IDs

sethrj:
The next step of https://github.com/celeritas-project/celeritas/issues/1749 is separating ORANGE's implementation volume IDs (which are the "global" implementation volume index, the unfolded local volume + universe ID) from the `VolumeParams` IDs which are the geometry's "structural" IDs. Structural here means deriving directly from the user-input geometry structure, which in Geant4's case is the "mother-daughter" volume hierarchy. In SCALE, the structural IDs will only encompass user-placed media and holes, not intermediate volumes like "exterior".

This renames `InternalSurfaceId -> ImplSurfaceId` and defines `ImplVolumeId` and `ImplVolumeUniqueInstanceId`. Currently those two are direct aliases and are just used as code "comments" rather than type enforcements. When we complete the separation of IDs (#1815) they will become an independent type.

I've also made aliases to distinguish the geometry params interface from the "core" (stepping algorithm) selection: `CoreGeoX` replaces `GeoX` in all applicable places but I've left the old files and aliases for compatibility.


sethrj:
@elliottbiondo This is ready for review

elliottbiondo (approved):

PR #1840, opened 2025-07-01
Add RevolvedSpecialTrapezoid as a first step towards G4GenericPolycone

elliottbiondo:
Per issue #1843, this MR adds a new `IntersectRegion` called a `RevolvedSpecialTrapezoid`, which is a trapezoid or flat bottom/top triangle, revolved around the z axis.


elliottbiondo:
I finished this, but at the very end I realized I don't think this actually works: double-sheeted cones can break things by trimming the polygon with BOTH sheets!

sethrj:
As discussed, this will be used as part of #1843.

elliottbiondo:
@sethrj @mrguilima is this ready for your review

src/orange/orangeinp/CsgTypes.hh
```diff
@@ -146,6 +148,67 @@ inline constexpr bool is_boolean_node(Node const& n)
...
+//---------------------------------------------------------------------------//
+/*!
+ * Polygon bound by top/bottom z segments, one of which may be a single point.
+ *
+ * Thus, valid polygons are:
+ *
+ * 1) quadrilaterals with two segments parallel to the z axis,
+ * 2) triangles with one segment parallel to the z axis.
+ */
+class SpecialTrapezoid
```
elliottbiondo, line 162: This class is arguable large enough to test on it's own, but it is tested implicitly via `RevolvedSpecialTrapzoid`

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1212,10 +1205,10 @@ void Prism::build(IntersectSurfaceBuilder& insert_surface) const
     insert_surface(Sense::outside, PlaneZ{-hh_});
     insert_surface(Sense::inside, PlaneZ{hh_});
 
-    // Offset (if user offset is zero) is calculated to put a point at y=0 on
-    // the +x axis. An offset of 1 would produce a shape congruent with an
-    // offset of zero, except that every face has an index that's decremented
-    // by 1. We prevent this by using fmod.
+    // Offset (if user offset is zero) is calculated to put a point at y=0
+    // on the +x axis. An offset of 1 would produce a shape congruent with
```
elliottbiondo: Not sure what is happening by my `clang-format`, but this is an ongoing issue. I can fix this at the end if it matters.

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -254,6 +254,23 @@ filter_collinear_points(std::vector<Real2> const& corners, double abs_tol)
     return result;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the min/max values of a polygon for a given dimension.
+ */
+inline std::pair<Real2::value_type, Real2::value_type>
+find_extrema(std::vector<Real2> const& polygon, size_type dim)
```
elliottbiondo, line 263: I added this back when I was using a different approach. It is used in `ExtrudedPolygon` within this MR. I still think it's a good refactor, even if it's not used in `RevolvedSpecialTrapezoid`

src/corecel/math/SoftEqual.hh
```diff
@@ -115,7 +115,43 @@ class EqualOr : public F
 
 //---------------------------------------------------------------------------//
 /*!
- * Functor for floating point equality.
+ * Functor for comparing values with an absolute precision.
+ */
+template<class RealType = ::celeritas::real_type>
+class SoftClose
```
sethrj, line 121: I wish I hadn't suggested this name, because now it sounds like a drawer hardware damper 😂

src/orange/orangeinp/CsgTypes.hh
```diff
@@ -146,6 +148,67 @@ inline constexpr bool is_boolean_node(Node const& n)
...
+    {
+        quad,
+        pointy_top,
+        pointy_bot
+    };
+    //!@}
+
+  public:
+    // Construct from bottom/top z segments
+    SpecialTrapezoid(ZSegment&& bot, ZSegment&& top);
```
sethrj, line 185: Since these arguments are (and always will be) POD it's just as good to use `const&`, but based on the way we're constructing it I think it's OK to leave as is.

src/orange/orangeinp/CsgTypes.cc
```diff
@@ -79,6 +81,85 @@ std::string to_string(Node const& n)
...
+    CELER_VALIDATE(bot_.z < top_.z,
+                   << "bottom segment must have a lower z than the top "
+                      "segment");
+
+    // Calculate abs_tol_ based on extents
+    auto r_min = std::min(bot_.r[left], top_.r[left]);
+    auto r_max = std::max(bot_.r[right], top_.r[right]);
+    Real3 const extents{r_max - r_min, top_.z - bot_.z, 0};
+    abs_tol_ = ::celeritas::detail::BumpCalculator(
+        Tolerance<>::from_default())(extents);
```
sethrj, line 101: If the tolerance is to be part of this class, it should be passed into it from upstream (where the construction is occurring and has knowledge of the geometry tolerance). I think it would behoove you to do fuzzy deduplication logic outside of this class so that you change the coordinates in question consistently across adjacent trapezoidal regions.

src/orange/orangeinp/CsgTypes.cc
```diff
@@ -79,6 +81,85 @@ std::string to_string(Node const& n)
...
+        points.push_back({bot_.r[left], bot_.z});
+    }
+
+    // Add top points
+    points.push_back({top_.r[left], top_.z});
+    if (variety_ != Variety::pointy_top)
+    {
+        points.push_back({top_.r[right], top_.z});
+    }
+
```
sethrj, line 156: Having an extra `variety_` class and that logic really seems like overengineering; if you rely on point deduplication upstream you would just have:
```suggestion
    points.push_back({bot_.r[right], bot_.z});
    if (bot_.r[right] != bot_.r[left])
    {
        points.push_back({bot_.r[left], bot_.z});
    }
    points.push_back({top_.r[left], top_.z});
    if (top_.r[right] != top_.r[left])
    {
        points.push_back({top_.r[right], top_.z});
    }
```
which seems just as clear.

Also, not a big deal, but do we have a "default" ordering for polygon points? For the extruded prism and other shapes, we move counterclockwise from the top right quadrant (i.e. theta = 0, theta increasing) rather than bottom right clockwise (theta = -pi/2, theta decreasing).

src/orange/orangeinp/detail/PolygonUtils.hh
sethrj: Should `size_type dim` be `Axis ax` with an `CELER_EXPECT(ax == Axis::x || ax == Axis::y)`? There's no validation on the dimension as is.

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -254,6 +254,23 @@ filter_collinear_points(std::vector<Real2> const& corners, double abs_tol)
...
+{
+    CELER_VALIDATE(polygon.size() >= 3,
+                   << "polygon must consist of at least 3 points");
+
+    auto [poly_min_it, poly_max_it] = std::minmax_element(
+        polygon.begin(), polygon.end(), [&dim](auto const& a, auto const& b) {
```
sethrj: For trivially copyable variables like `dim` it's better to just copy rather than reference them if const (same as you would an int data member in a struct).
```suggestion
        polygon.begin(), polygon.end(), [dim](auto const& a, auto const& b) {
```

test/corecel/math/SoftEqual.test.cc
```diff
@@ -115,6 +115,50 @@ TYPED_TEST(FloatingTest, soft_equal)
     EXPECT_FALSE(comp(inf, inf));
 }
 
+TYPED_TEST(FloatingTest, soft_near)
```
sethrj:
```suggestion
TYPED_TEST(FloatingTest, soft_close)
```

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -1344,17 +1344,17 @@ TEST_F(GenPrismTest, trap_even_twist)
...
+
+   x=-1      x=1           x=-0.5
+   +----+----+ y=1      +--+------+ y=1
+   |    |    |          |   \     |
+   |    |  R |          |    \  R |
+   |  L |    |          |  L  \   |
+   |    |    |          |      \  |
+   +----+----+ y=-1     +-------+-+ y=-1
+        x=0                     x=0.5
+   \endverbatim
```
sethrj, line 1319: FWIW it renders the same in doxygen (but yes it's easier to read). I tend to leave them in because of the inconsistency over the leading `*` before the verbatim/endverbatim commands.

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -735,6 +739,57 @@ class Prism final : public IntersectRegionInterface
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * A region formed by revolving a SpecialTrapezoid around the z-axis.
+ *
+ * The special trapezoid constraints are imposed by the general quadratic
+ * capability in orange/surf, which only supports double-sheeted conical
+ * surfaces. As result, polygons more complex than special trapezoids cannot be
+ * created in general case, as the undesired sheet can erroneously clip the
+ * polygon.
```
sethrj:
```suggestion
 * The trapezoid being revolved always has a flat top and bottom along the
 * \em z axis never crosses the \em z axis. These special constraints ensure
 * that the revolved trapezoid's sides never touch both sheets of a cone, which
 * as a quadric is always double-sheeted about the cone's vanishing point.
```

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -735,6 +739,57 @@ class Prism final : public IntersectRegionInterface
...
+ *
+ * As a convention, all r values must be positive, which also guarentees the
```
sethrj: Is `r=0` valid? (I would imagine it's possible in the geant4 solid?)

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -735,6 +739,57 @@ class Prism final : public IntersectRegionInterface
...
+ * polygon does not cross the z axis. An example of the revolution
+ * process is shown below.
+ * \verbatim
+                              |
+     ........            ^    |     .           ________
```
sethrj, line 800: There you go rotating things clockwise again 😆

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1271,6 +1264,112 @@ bool Prism::encloses(Prism const& other) const
...
+ */
+void RevolvedSpecialTrapezoid::build(IntersectSurfaceBuilder& insert_surface) const
+{
+    auto const& bot = trap_.bot();
+    auto const& top = trap_.top();
+    constexpr auto left = Bound::lo;
+    constexpr auto right = Bound::hi;
+
+    // Create z planes first for faster short circuiting
+    if (trap_.variety() != SpecialTrapezoid::Variety::pointy_bot)
```
sethrj: I think we still want to keep the z planes even if it is pointy: the z slice will constrain the resulting bbox and it will accelerate the intersection test.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1271,6 +1264,112 @@ bool Prism::encloses(Prism const& other) const
...
+        // Bottom z plane
+        insert_surface(Sense::outside, PlaneZ{bot.z});
+    }
+    if (trap_.variety() != SpecialTrapezoid::Variety::pointy_top)
+    {
+        // Top z plane
+        insert_surface(Sense::inside, PlaneZ{top.z});
+    }
+
+    SoftClose soft_close(trap_.abs_tol());
```
sethrj: We should be using the tolerances from the `IntersectSurfaceBuilder` propagated from the geometry.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1271,6 +1264,112 @@ bool Prism::encloses(Prism const& other) const
...
+        = [&](real_type r_bot, real_type r_top, Sense sense) {
+              if (!soft_close(r_bot, r_top))
+              {
+                  // Conical surface
+                  insert_surface(
+                      sense, this->make_cone({r_bot, bot.z}, {r_top, top.z}));
+              }
+              else if (!soft_close(real_type{0}, r_bot))
+              {
+                  // Cylindrical surface
```
sethrj, line 1409: I agree it's good to have this, but I wonder how it overlaps with the surface simplification (which only uses relative tolerances and may be a little fudgy for general quadrics).

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1271,6 +1264,112 @@ bool Prism::encloses(Prism const& other) const
...
+                  insert_surface(sense, CCylZ(r_bot));
+              }
```
sethrj, line 1411: For clarity I would also write
```suggestion
              }
              else
              {
                  // r_top = r_bottom = 0: do not create a surface
              }
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1271,6 +1264,112 @@ bool Prism::encloses(Prism const& other) const
...
+
+    auto delta_r = p1[R] - p0[R];
+    auto delta_z = p1[Z] - p0[Z];
+    auto tangent = delta_r / delta_z;
+    auto intercept = p0[Z] - p0[R] * delta_z / delta_r;
+
+    // The tangent value given to ConeZ must be positive. However, since
+    // ConeZ creates a double-sheeted cone, the negative cone will be
+    // properly produced as well.
+    return ConeZ{Real3{0, 0, intercept}, std::abs(tangent)};
```
sethrj:
```suggestion
    // ConeZ creates a double-sheeted cone, the negative cone will be
    // properly produced as well.
    return ConeZ{Real3{0, 0, intercept}, tangent};
```
The positivity condition was removed somewhat recently (#1832); the restriction now is the tangent is nonzero (and not NaN).

sethrj (commented):
Looking good!

src/orange/orangeinp/CsgTypes.cc
elliottbiondo: The reason I do point deduplication on the class is to prevent construction in an invalid state, i.e. two sets of equal points. There will be a second class, `SpecialTrapezoidDecomposer`, that will actually construct these objects, so having robust DBC for construction will lessen the DBC burden within `SpecialTrapezoidDecomposer`.

src/orange/orangeinp/IntersectRegion.hh
elliottbiondo: Yes, typo

src/orange/orangeinp/CsgTypes.cc
elliottbiondo: For arbitrary polygons, e.g., `ExtrudedPolygon` we use clockwise, same a Geant4. I was thinking that this function would return such a polygon. However, it's easy enough to change.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: I am not sure I understand. Insert a bbox a few lines down:

```
    // Establish bbox
    auto r_max = std::max(bot.r[right], top.r[right]);
    insert_surface(
        Sense::inside,
        BBox::from_unchecked({-r_max, -r_max, bot.z}, {r_max, r_max, top.z}));
```

sethrj: Good point. I was thinking for short circuit logic. For the axis aligned case, the BIH should take care of acceleration too.

elliottbiondo: Adding both planes for faster intersection tests

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Closing based offline discussion. Basically here we have access to a properly scaled tolerance, so it makes sense it use it rather hoping things work out downstream.

src/orange/orangeinp/CsgTypes.cc
```diff
@@ -79,6 +81,85 @@ std::string to_string(Node const& n)
...
+ */
+SpecialTrapezoid::SpecialTrapezoid(ZSegment&& bot, ZSegment&& top)
+    : bot_(std::move(bot)), top_(std::move(top))
+{
+    constexpr auto left = Bound::lo;
+    constexpr auto right = Bound::hi;
+
+    CELER_VALIDATE(bot_.z < top_.z,
+                   << "bottom segment must have a lower z than the top "
+                      "segment");
```
sethrj: Since this mostly used is an internal implementation class, I don't think we really need to VALIDATE as much in this object. These should be internal assertions if they're coming out of a builder helper class. Maybe we should spend a few minutes discussing this and your comment below about DBC.

src/orange/orangeinp/detail/PolygonUtils.hh
elliottbiondo: This is a carry-over from the original implementation, in which this function would apply to both polygons on the XY plane and RZ plane. Now it's only used in the XY plane, so `Axis` technically works, but it makes it less generic.

src/orange/orangeinp/IntersectRegion.cc
sethrj: Strange, is this happening from visual studio code? (I've seen it disagree with clang-format sometimes)

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -1999,6 +1999,155 @@ TEST_F(PrismTest, rhex)
...
+//---------------------------------------------------------------------------//
+// REVOLVEDSPECIALTRAPEZOID
+//---------------------------------------------------------------------------//
+using RevolvedSpecialTrapezoidTest = IntersectRegionTest;
+
+//---------------------------------------------------------------------------//
+/*
+ * Test revolving a square around z.
+ */
+TEST_F(RevolvedSpecialTrapezoidTest, simple_cube)
```
sethrj:
```suggestion
TEST_F(RevolvedSpecialTrapezoidTest, square)
```

test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -1999,6 +1999,155 @@ TEST_F(PrismTest, rhex)
...
+{
+    SpecialTrapezoid trap({{1, 1}, 0}, {{0, 4}, 3});
+    auto result = this->test(RevolvedSpecialTrapezoid(std::move(trap)));
+
+    static char const expected_node[] = "all(+0, -1, +2, -3)";
+    static char const* const expected_surfaces[] = {"Plane: z=0",
+                                                    "Plane: z=3",
+                                                    "Cone z: t=0.33333 at "
+                                                    "{0,0,3}",
+                                                    "Cone z: t=1 at {0,0,-1}"};
```
sethrj: You might get some complains from clang about odd linebreak; add a trailing comma before the end brace and re-join the broken string:
```suggestion
    static char const* const expected_surfaces[] = {
        "Plane: z=0",
        "Plane: z=3",
        "Cone z: t=0.33333 at {0,0,3}",
        "Cone z: t=1 at {0,0,-1}",
    };
```

src/orange/orangeinp/detail/PolygonUtils.hh
```diff
@@ -254,6 +254,23 @@ filter_collinear_points(std::vector<Real2> const& corners, double abs_tol)
     return result;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the min/max values of a polygon for a given dimension.
+ */
+inline std::pair<Real2::value_type, Real2::value_type>
```
sethrj: Unless you plan to template this function...
```suggestion
inline std::pair<real_type, real_type>
```

src/orange/orangeinp/CsgTypes.cc
sethrj: Let's add a to-do in the code consistency post to change these to counterclockwise later: https://github.com/celeritas-project/celeritas/issues/721 so we can do it at the same time as the existing poly

test/orange/orangeinp/IntersectRegion.test.cc
elliottbiondo: My clang-format automatically does this line break, not me!

test/orange/orangeinp/IntersectRegion.test.cc
elliottbiondo: RE: clang-format: this is on the command line with `clang-format version 19.1.7`. No VS code involved.

sethrj: Weird

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Since I am not doing this though VS code,  maybe my clang-format is actually more correct. I updated to `clang-format version 20.1.8` via `brew` and I am still getting clashes.

src/orange/orangeinp/CsgTypes.cc
elliottbiondo: We discussed the offline. Basically, floating point equality should in general be checked at build-time using the tolerance on `insert_surface`. In this case, it makes sense to have SOME validation on `SpecialTrapezoid`, and in the absence of an `insert_surface` object, this is a okay way of doing it. Maybe this should be a minimal struct --- we can revisit this possibility on the next MR, which will involved building a bunch of `SpecialTrapezoids`.

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Fixed and will use `pre-commit` henceforth.

elliottbiondo:

elliottbiondo:
@sethrj I will probably refactor this in the MR after next; i.e. actually constructing the the general polycone.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1362,6 +1355,109 @@ bool Prism::encloses(Prism const& other) const
...
+        << "r values must be positive");
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Build surfaces.
+ *
+ * Surface are constructed by revolving each segment around the z axis. Thus:
+ * - segments parallel to z become z-aligned cylinders,
+ * - segments perpendicular to z become z-orthagonal planes,
```
mrguilima:
```suggestion
 * - segments perpendicular to z become z-orthogonal planes,
```

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -1362,6 +1355,109 @@ bool Prism::encloses(Prism const& other) const
...
+ */
+void RevolvedSpecialTrapezoid::build(IntersectSurfaceBuilder& insert_surface) const
+{
+    auto const& bot = trap_.bot();
+    auto const& top = trap_.top();
+    constexpr auto left = Bound::lo;
+    constexpr auto right = Bound::hi;
+
+    // Create both z planes first, even if one end is pointy, for short
+    // circutting
```
mrguilima:
```suggestion
    // circuiting
```

sethrj (approved):

mrguilima (commented):
Just a couple typos for now.

src/orange/orangeinp/IntersectRegion.cc
sethrj: Good catch!

PR #1842, opened 2025-07-02
Add a helper function for launching a CPU kernel with an arbitrary number of threads

amandalund:
This simplifies some of our CPU action launches a bit.


sethrj (approved):
Very nice! Thanks!

PR #1844, opened 2025-07-04
Generate optical photons directly into track slots for 2× GPU speedup

amandalund:
This changes how we do the optical photon generation: instead of creating track initializers for all photons from all distributions at the start of the optical loop, optical tracks are initialized directly in vacant track slots from distributions at each step. The requirement that all tracks be generated at once and the memory needed for the buffered initializers previously limited the size of the problems we could run.

With this change, on the simple LAr sphere benchmark I saw a 10% slowdown for CPU track-level parallelism, a 15% speedup for CPU event-level parallelism, and over 2x speedup on the GPU (now 820x compared to Geant4).


test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -591,8 +587,8 @@ TEST_F(LArSphereOffloadTest, host_generate)
     if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)
     {
         EXPECT_SOFT_NEAR(
-            462263.0, static_cast<double>(result.accum.steps), 1e-4);
-        EXPECT_EQ(37, result.accum.step_iters);
+            4593666, static_cast<double>(result.accum.steps), 1e-4);
```
amandalund, line 590: I'm not sure why this value (sum of the number of active tracks over all step iterations) changed by an order of magnitude, but given the number of step iterations and track slots the new value seems more reasonable.

src/celeritas/optical/gen/detail/GeneratorAction.cc
```diff
@@ -149,30 +150,49 @@ void GeneratorAction<G>::step_impl(optical::CoreParams const& params,
...
-    // Update cumulative statistics
-    aux_state.accum.distributions += aux_state.buffer_size;
-    aux_state.accum.photons += count;
-
-    state.counters().num_initializers += count;
-    state.counters().num_pending -= count;
-    aux_state.buffer_size = 0;
+    if (aux_state.num_generated == 0)
+    {
+        // On the first step iteration, calculate the cumulative sum of the
```
sethrj, line 155: Do we guarantee that one optical launch always tracks all optical photons to completion? We might want a case where we pause the optical loop when the track slots are no longer full, and then at the end of the event we flush any that remain.

Right now if `num_step_iters == max_step_iters_` in the launch action, it looks like we simply kill any remaining tracks?

test/celeritas/optical/OpticalCollector.test.cc
sethrj: Weird but I agree!

sethrj (approved):
Good work, great result!

src/celeritas/optical/gen/detail/GeneratorAction.cc
amandalund: Right, currently we always run the optical loop to completion, and if we reach the max iterations we kill the tracks. Yes, we should definitely do that: maybe make it the default behavior to pause when there are no more pending tracks and refill the distributions. Increasing the auto flush can help reduce the tails to some extent, but not as effectively in all cases.

amandalund: I'm probably not resetting the counters correctly if we set `max_step_iters`, so let me double check that. I also hadn't tuned the benchmark settings at all: doing it all in a single flush gets us another 20% speedup on GPU.

src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -137,48 +128,40 @@ void OpticalLaunchAction::execute_impl(CoreParams const&,
...
                 << ": aborting optical transport loop with "
                 << counters.num_generated << " generated tracks, "
                 << counters.num_active << " active tracks, "
                 << counters.num_alive << " alive tracks, "
                 << counters.num_vacancies << " vacancies, and "
-                << counters.num_initializers << " queued";
+                << counters.num_pending << " queued";
 
             state.reset();
+            this->reset_generators(core_state.aux());
```
amandalund, line 164: We may want a separate class to manage the generators, but this should work for now.

sethrj (approved):
Awesome! Seems like a `GeneratorInterface` would be a good follow-on.

PR #1845, opened 2025-07-08
Update build script and user presets

sethrj:
Based on recent feedback from @Rashika-Gupta and @mrguilima, this improves the documentation and attempts to streamline the use of CMake presets with Celeritas, especially by setting appropriate "development" defaults. @lebuller since you prefer to use a homemade shell script to compile/build/test, I'd appreciate your feedback about this as well.


Rashika-Gupta:
Tested locally, builds and runs fine.

Rashika-Gupta (approved):

scripts/build.sh
```diff
@@ -1,5 +1,20 @@
 #!/bin/sh -e
 
+function log() {
```
lebuller: On my system, using Ubuntu through Windows subsystem for linux, `/bin/sh` by default points to dash, which does not have many modern scripting features, such as functions, so this script errors with `./scripts/build.sh: 3: Syntax error: "(" unexpected`. dash is the shell that `/bin/sh` points to by default on Ubuntu and other Debian-based linux distros, so many linux and windows (through WSL) users will run into this error, which is not trivial to debug. Using `/bin/bash` works for me. It would be good to either switch the shell utility used here or to add some detail to the README about this error

sethrj: Thanks! Made POSIX compliant.

lebuller (approved):

scripts/cmake-presets/_dev_.json
```diff
@@ -0,0 +1,55 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0},
+  "configurePresets": [
+    {
+      "name": "dev",
+      "displayName": "Default development options",
+      "inherits": [".debug", "default"],
+      "binaryDir": "${sourceDir}/build",
```
pcanal, line 9: Side note.  I tend to prefer to keep the source directory untouched by build and install artifact.  (In addition more often than not, I was several different build configuration using the same source).  [Not requesting any changes just describing distinct use case]

sethrj: Fair enough. Many tools (clangd, vscode) "default" to a subdirectory named `build`, which I find pretty convenient. To support multiple build directories I use `build-${presetName}`—my mac usually has two or three of these.

esseivaju:
We could also auto-detect any existing ccache install and configure Celeritas to use it, e.g. 
```diff
diff --git a/CMakePresets.json b/CMakePresets.json
index 9feff14a..fd589e20 100644
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -2,12 +2,22 @@
   "version": 3,
   "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0},
   "configurePresets": [
+    {
+      "name": ".ccache",
+      "hidden": true,
+      "description": "Use ccache compiler launcher if available",
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER_LAUNCHER": { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"},
+        "CMAKE_C_COMPILER_LAUNCHER":   { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"}
+      }
+    },
     {
       "name": "default",
       "displayName": "Automatic options",
       "description": "Dependencies are enabled based on environment probing",
       "binaryDir": "${sourceDir}/build-${presetName}",
       "generator": "Ninja",
+      "inherits": [".ccache"],
       "cacheVariables": {
         "Celeritas_GIT_DESCRIBE": "",
         "CELERITAS_BUILD_TESTS":{"type": "BOOL", "value": "ON"},
@@ -19,7 +29,7 @@
       "displayName": "Full",
       "description": "All external dependencies are expected",
       "binaryDir": "${sourceDir}/build",
-      "inherits": "default",
+      "inherits": ["default", ".ccache"],
       "cacheVariables": {
         "CELERITAS_BUILD_DOCS":  {"type": "BOOL", "value": "ON"},
         "CELERITAS_BUILD_TESTS": {"type": "BOOL", "value": "ON"},
@@ -37,6 +47,7 @@
       "description": "Smallest capabilities for useful development",
       "generator": "Ninja",
       "binaryDir": "${sourceDir}/build-${presetName}",
+      "inherits": [".ccache"],
       "cacheVariables": {
         "CELERITAS_BUILD_DOCS":  {"type": "BOOL", "value": "OFF"},
         "CELERITAS_USE_CUDA":    {"type": "BOOL", "value": "OFF"},
diff --git a/scripts/build.sh b/scripts/build.sh
index 83f01b14..ecec0d15 100755
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -23,6 +23,14 @@ if [ -f "${_ENV_SCRIPT}" ]; then
   . "${_ENV_SCRIPT}"
 fi
 
+# Auto-detect and configure ccache if available
+if command -v ccache >/dev/null 2>&1; then
+  export CCACHE_PROGRAM="ccache"
+  echo "ccache detected and enabled" >&2
+else
+  export CCACHE_PROGRAM=""
+fi
+
 # Check arguments and give presets
 if [ $# -eq 0 ]; then
   echo "Usage: $0 PRESET [config_args...]" >&2

``` 

I can create a separate PR since this is already setup to auto-merge

PR #1846, opened 2025-07-08
Add volume params to core

sethrj:
This adds volume parameters to the core celeritas global params. Rather than add another unit test extension point, I've started factoring out some of the `setup`/`inp` code to be used in the test harness.

Needed for https://github.com/celeritas-project/celeritas/issues/1749 .


src/geocel/VolumeParams.hh
```diff
@@ -66,16 +74,16 @@ class VolumeParams
...
-    inline Span<VolumeInstanceId const> children(VolumeId vid) const;
+    inline Span<VolumeInstanceId const> children(VolumeId v_id) const;
 
     // Get the geometry material of a volume
-    inline GeoMatId material(VolumeId vid) const;
+    inline GeoMatId material(VolumeId v_id) const;
 
     // Get the volume being instantiated (outgoing node)
-    inline VolumeId volume(VolumeInstanceId vid) const;
+    inline VolumeId volume(VolumeInstanceId viid) const;
```
amandalund:
```suggestion
    inline VolumeId volume(VolumeInstanceId vi_id) const;
```

sethrj:
@amandalund This should be ready now.

src/geocel/vg/VecgeomParams.cc
```diff
@@ -329,10 +329,15 @@ VecgeomParams::~VecgeomParams()
 //---------------------------------------------------------------------------//
 /*!
  * Create model parameters corresponding to our internal representation.
+ *
+ * This could be used to eliminate the "gaps" from the `[TEMP]` volumes.
  */
 inp::Model VecgeomParams::make_model_input() const
 {
-    CELER_NOT_IMPLEMENTED("model input from VecGeom");
+    CELER_LOG(warning) << "ORANGE cannot yet construct model input";
```
amandalund:
```suggestion
    CELER_LOG(warning) << "VecGeom cannot yet construct model input";
```

src/geocel/SurfaceParams.cc
```diff
@@ -42,24 +41,34 @@ SurfaceParams::SurfaceParams(inp::Surfaces const& input,
...
+        {
+            // Only resize surface array if surfaces are defined
+            detail::VolumeSurfaceRecordBuilder build_record(
+                &host_data.volume_surfaces,
+                &host_data.volume_instance_ids,
+                &host_data.surface_ids);
+            std::for_each(temp_volume_surfaces.begin(),
+                          temp_volume_surfaces.end(),
+                          build_record);
+        }
```
amandalund:
```suggestion
        detail::VolumeSurfaceRecordBuilder build_record(
            &host_data.volume_surfaces,
            &host_data.volume_instance_ids,
            &host_data.surface_ids);
        std::for_each(temp_volume_surfaces.begin(),
                      temp_volume_surfaces.end(),
                      build_record);
```
We settled on creating empty boundary surface IDs if optical physics was enabled but no surfaces were defined, right? Otherwise we can't construct the `VolumeSurfaceView`.

amandalund (commented):
Looks good! just a question about the surface params behavior.

src/geocel/SurfaceParams.cc
sethrj: I've forgotten and confused myself: I guess "no surfaces" is the special case for optical physics, but the regular case for EM physics (which now will always also create surface params).

With the `if`, we can't instantiate a `VolumeSurfaceView`, which I thought was what we agreed but maybe not. Without the if, we have an unused volumes-length collection of volume-to-surface data. If we don't have surfaces, we won't have surface physics, and we shouldn't have to have to check the surface view at the boundary. But I suppose for optical physics, that would mean adding a special no-surface case.

So maybe I should restore the default constructor for surface physics—to be used if optical tracking is off and doesn't create any data—and implement the above change?

amandalund: > So maybe I should restore the default constructor for surface physics—to be used if optical tracking is off and doesn't create any data—and implement the above change?

I think that makes sense: empty surface data and no instantiating a `VolumeSurfaceView` if optical physics is off, and invalid boundary surfaces for any volumes that don't have surface physics if it's on (which could in the special case be all of them).

amandalund (approved):
Great, thanks!

PR #1847, opened 2025-07-08
Add track weight propagation to secondaries track

Rashika-Gupta:
This PR adds support for propagating particle weights from initial generation through secondary creation and into HitProcessor [Issue #1268].

- The weight field is added to `SimTrackView`, `SimData`, and included in `DetectorSteps`.
- Parent Track weight is propagated to secondary track in `InteractionApplier` and energy is updated by multiplying the deposited energy with the weight.
- Copied to `HitProcessor`.
- Test added in `Sim.Test`.


src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -326,8 +326,11 @@ void HitProcessor::operator()(DetectorStepOutput const& out, size_type i) const
          * \todo Celeritas currently ignores incoming particle weight and
          * does not perform any variance reduction. See issue #1268.
          */
-        g4sp->SetWeight(1.0);
-
+        // g4sp->SetWeight(1.0);
```
sethrj: Remove dead code

src/celeritas/phys/Primary.hh
```diff
@@ -27,7 +27,7 @@ struct Primary
     Real3 direction{0, 0, 0};
     real_type time{};
     EventId event_id;
+    real_type weight{1.0};
 };
-
```
sethrj, line 31: We tend to leave some whitespace in between blocks of code for style and vim navigation (left and right braces move between blank lines) so please revert this whitespace change and the one in Secondary.

sethrj (approved):
Truly an excellent first pull request! (Great even for a PR for a longtime team member!) Once the minor changes are addressed we can merge this.

In the next pull request, you should modify `test/celeritas/ext/HitProcessor.test.cc` to make one of the tracks a non-unit weight, and update the `SimpleSensitiveDetector` to multiply the weight into the energy deposition.

PR #1848, opened 2025-07-08
Detect and use ccache

esseivaju:
This automatically configure `CMakePresets` to use cache if available in the user environment. Also parse the cache utilization and print a message recommending to increase the cache size if almost full

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -10,6 +10,8 @@
       "cacheVariables": {
         "Celeritas_GIT_DESCRIBE": "",
         "CELERITAS_BUILD_TESTS":{"type": "BOOL", "value": "ON"},
+        "CMAKE_CXX_COMPILER_LAUNCHER": { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"},
+        "CMAKE_C_COMPILER_LAUNCHER":   { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"},
         "CMAKE_INSTALL_PREFIX": "${sourceDir}/install-${presetName}"
       }
     },
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -16,21 +16,18 @@ log() {
 
 check_ccache_usage() {
     # Check if ccache is full and warn user
-    if ccache --print-stats >/dev/null 2>&1; then
-        cache_stats=$(ccache --print-stats)
-        current_kb=$(echo "$cache_stats" | grep "^cache_size_kibibyte" | cut -f2)
-        max_kb=$(echo "$cache_stats" | grep "^max_cache_size_kibibyte" | cut -f2)
+    cache_stats=$(ccache --print-stats)
+    current_kb=$(echo "$cache_stats" | grep "^cache_size_kibibyte" | cut -f2)
+    max_kb=$(echo "$cache_stats" | grep "^max_cache_size_kibibyte" | cut -f2)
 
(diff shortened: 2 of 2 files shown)


scripts/build.sh
```diff
@@ -14,6 +14,35 @@ log() {
...
+    max_kb=$(echo "$cache_stats" | grep "^max_cache_size_kibibyte" | cut -f2)
+
+    if [ -n "$current_kb" ] && [ -n "$max_kb" ] && [ "$max_kb" -gt 0 ] 2>/dev/null; then
+        usage_percent=$((current_kb * 100 / max_kb))
+
+        if [ "$usage_percent" -gt 90 ] 2>/dev/null; then
+            # Convert to human-readable sizes (GiB)
+            current_gb=$((current_kb / 1024 / 1024))
+            max_gb=$((max_kb / 1024 / 1024))
+            log warning "ccache is ${usage_percent}% full (${current_gb}/${max_gb} GiB). Consider increasing cache size with 'ccache -M <size>G'"
```
sethrj, line 30: Since this gets piped into printf, better escape the `%`:
```suggestion
            log warning "ccache is ${usage_percent}%% full (${current_gb}/${max_gb} GiB). Consider increasing cache size with 'ccache -M <size>G'"
```

scripts/build.sh
```diff
@@ -14,6 +14,35 @@ log() {
...
+        fi
+    fi
+}
+
+setup_ccache() {
+    # Auto-detect and configure ccache if available
+    if command -v ccache >/dev/null 2>&1; then
+        export CCACHE_PROGRAM="ccache"
+        log info "Using ccache"
```
sethrj: Maybe use `which` so we get the full path stored in the cmake configure file

CMakePresets.json
```diff
@@ -10,6 +10,8 @@
       "cacheVariables": {
         "Celeritas_GIT_DESCRIBE": "",
         "CELERITAS_BUILD_TESTS":{"type": "BOOL", "value": "ON"},
+        "CMAKE_CXX_COMPILER_LAUNCHER": { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"},
+        "CMAKE_C_COMPILER_LAUNCHER":   { "type": "STRING", "value": "$env{CCACHE_PROGRAM}"},
```
sethrj, line 14: Did you test that this works even when ccache isn't available?

scripts/build.sh
esseivaju: Does it really need escaping if it's not the printf format string? Testing that change I see a double '%%' being printed

CMakePresets.json
esseivaju: Yes, tested and this will simply be an empty CMake variable is the env variable isn't set

scripts/build.sh
sethrj: Oh apologies, I didn't realize it was getting substituted as an `%s` argument: good the way you wrote it.

sethrj (approved):
LGTM, though I worry we're over-engineering the build script... also I will have to update my presets to *disable* ccache where not wanted.

esseivaju:
> LGTM, though I worry we're over-engineering the build script... also I will have to update my presets to _disable_ ccache where not wanted.

I agree that it is a downside,  but I think it's easier for power-users to selectively turn-off ccache in select presets, and this would benefit users that would otherwise not have known about ccache, especially if we include it in our spack dev environment.

PR #1849, opened 2025-07-08
Add generator interface and registry

amandalund:
This adds an `Interface` and `Registry` for the generators and refactors the counters a bit.


test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -58,15 +59,9 @@ class LArSphereOffloadTest : public LArSphereBase
...
         // Optical distribution data
-        size_type num_photons{0};
         OffloadResult cherenkov;
         OffloadResult scintillation;
         OpticalAccumStats accum;
-
-        // Step iteration at which the optical tracking loop first launched
-        size_type optical_launch_step{numeric_limits<size_type>::max()};
-
-        void print_expected() const;
```
sethrj, line 69: Why is print_expected no longer needed?

sethrj (approved):
This is great! Nice work.

test/celeritas/optical/OpticalCollector.test.cc
amandalund: It was just getting a little long and hard to maintain, and we don't use the full output anywhere, just selected results in each of the tests. This test could also probably use an overhaul; as we've added more features and refactored things I keep patching it to keep up and could probably do things more cleanly now.

PR #1852, opened 2025-07-10
Add surface physics map

sethrj:
Completes item 2 of https://github.com/celeritas-project/celeritas/issues/1512#issuecomment-3053851560 . This defines a collection group that we can use in the surface physics params: we'll have one surface physics map each for roughness, reflectivity, and interaction.

I've put this class in the core physics for now, since it's conceivable we'd want other physics (neutrons etc) to support surface-based behavior. We should be able to inherit `optical::SurfaceModel` from this class and an optical stepper.

### DIFF
--- a/src/celeritas/phys/SurfacePhysicsMapBuilder.cc
+++ b/src/celeritas/phys/SurfacePhysicsMapBuilder.cc
@@ -60,7 +60,7 @@ void SurfacePhysicsMapBuilder::operator()(SurfaceModel const& model)
         // Assign and check the action ID
         ActionId prev_id
             = std::exchange(data_.action_ids[surface_id], model.action_id());
-        CELER_VALIDATE(prev_id == ActionId{},
+        CELER_VALIDATE(!prev_id,
                        << "multiple surface physics models were assigned to "
                           "the same surface");
 
--- a/src/celeritas/phys/SurfacePhysicsMapBuilder.cc
+++ b/src/celeritas/phys/SurfacePhysicsMapBuilder.cc
@@ -1,0 +1,76 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/phys/SurfacePhysicsMapBuilder.cc
+//---------------------------------------------------------------------------//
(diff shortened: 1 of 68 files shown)


sethrj:
@amandalund @hhollenb This should be good to go!

src/celeritas/phys/SurfacePhysicsMapView.hh
```diff
@@ -0,0 +1,84 @@
...
+
+#include "corecel/Types.hh"
+
+#include "SurfacePhysicsMapData.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Brief class description.
```
hhollenb: Add missing documentation

test/geocel/SurfaceTestBase.hh
```diff
@@ -0,0 +1,84 @@
...
+    cc2 : interface 1 -> 2
+    c3c : interface 3 -> 1
+    bc  : interface 0 -> 1
+    bc2 : interface 0 -> 2
+    ec  : interface 5 -> 1
+    db  : interface 4 -> 1
+ * \endverbatim
+ *
+ */
+class SurfaceTestBase : public ::celeritas::test::Test
```
hhollenb, line 69: Does this match the data used in `test/geocel/Surface.test.cc`?

hhollenb (approved):
Looks good! Should be straightforward to port into SurfacePhysicsParams later

src/celeritas/phys/SurfacePhysicsMapBuilder.cc
```diff
@@ -0,0 +1,76 @@
...
+    for (SurfaceId surface_id : model.get_surfaces())
+    {
+        CELER_VALIDATE(surface_id < surfaces_.num_surfaces(),
+                       << "surface physics model " << model.label()
+                       << " contained invalid surface indices");
+
+        // Assign and check the action ID
+        ActionId prev_id
+            = std::exchange(data_.action_ids[surface_id], model.action_id());
+        CELER_VALIDATE(prev_id == ActionId{},
```
amandalund:
```suggestion
        CELER_VALIDATE(!prev_id,
```

amandalund (approved):
Nice, thanks!

test/geocel/SurfaceTestBase.hh
sethrj: Oh yeah, forgot to update that 🤔 good catch!

sethrj:
Thanks guys!

PR #1853, opened 2025-07-10
Update HitProcessor test to support non unit weights

### LINKED ISSUE #1268: Add particle weight to simulations

Rashika-Gupta:
This PR updates the energy deposition in Sensitive Detector and adjusts the corresponding the test expectations in Hit Processor by (See issue #1268): 
- Scaling energy deposition in `SimpleSensitiveDetector` by the track weight `edep = edep * weight` (if weight exists and is non-zero).
- Initializing `DetectorStepOutput::weight` in `make_dso()` within `HitProcessor.test.cc`. (It is assumed that weight is same for different hits)
- Updating test expectations in `HitProcessor.test.cc` to account for weighted energy deposition:
`expected_energy = weight * unweighted_energy `.

### DIFF
--- a/test/celeritas/ext/SimpleSensitiveDetector.cc
+++ b/test/celeritas/ext/SimpleSensitiveDetector.cc
@@ -99,10 +99,8 @@ bool SimpleSensitiveDetector::ProcessHits(G4Step* step, G4TouchableHistory*)
     {
         hits_.particle.push_back(track->GetDefinition()->GetParticleName());
         real_type weight = track->GetWeight();
-        if (weight != 0.0)
-        {
-            edep *= weight;
-        }
+        CELER_ASSERT(weight > 0);
+        edep *= weight;
     }
     hits_.energy_deposition.push_back(edep / CLHEP::MeV);
     hits_.pre_energy.push_back(pre_step->GetKineticEnergy() / CLHEP::MeV);
--- a/test/celeritas/ext/SimpleSensitiveDetector.cc
+++ b/test/celeritas/ext/SimpleSensitiveDetector.cc
@@ -91,14 +91,20 @@ bool SimpleSensitiveDetector::ProcessHits(G4Step* step, G4TouchableHistory*)
     auto* pre_step = step->GetPreStepPoint();
     CELER_ASSERT(pre_step);
 
-    hits_.energy_deposition.push_back(step->GetTotalEnergyDeposit()
-                                      / CLHEP::MeV);
     hits_.step_length.push_back(step->GetStepLength() / CLHEP::cm);
+
+    real_type edep = step->GetTotalEnergyDeposit();
+
     if (auto* track = step->GetTrack())
     {
         hits_.particle.push_back(track->GetDefinition()->GetParticleName());
+        real_type weight = track->GetWeight();
+        if (weight != 0.0)
+        {
+            edep *= weight;
+        }
     }
-
+    hits_.energy_deposition.push_back(edep / CLHEP::MeV);
     hits_.pre_energy.push_back(pre_step->GetKineticEnergy() / CLHEP::MeV);
 
     for (int i : range(3))
--- a/test/celeritas/ext/SimpleSensitiveDetector.cc
+++ b/test/celeritas/ext/SimpleSensitiveDetector.cc
@@ -93,16 +93,16 @@ bool SimpleSensitiveDetector::ProcessHits(G4Step* step, G4TouchableHistory*)
 
     hits_.step_length.push_back(step->GetStepLength() / CLHEP::cm);
 
-    real_type edep = step->GetTotalEnergyDeposit();
+    double edep = step->GetTotalEnergyDeposit() / CLHEP::MeV;
 
     if (auto* track = step->GetTrack())
     {
         hits_.particle.push_back(track->GetDefinition()->GetParticleName());
-        real_type weight = track->GetWeight();
+        double weight = track->GetWeight();
         CELER_ASSERT(weight > 0);
         edep *= weight;
     }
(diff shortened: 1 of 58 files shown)


test/celeritas/ext/SimpleSensitiveDetector.cc
```diff
@@ -91,14 +91,20 @@ bool SimpleSensitiveDetector::ProcessHits(G4Step* step, G4TouchableHistory*)
...
-                                      / CLHEP::MeV);
     hits_.step_length.push_back(step->GetStepLength() / CLHEP::cm);
+
+    real_type edep = step->GetTotalEnergyDeposit();
+
     if (auto* track = step->GetTrack())
     {
         hits_.particle.push_back(track->GetDefinition()->GetParticleName());
+        real_type weight = track->GetWeight();
+        if (weight != 0.0)
```
sethrj: When would Geant4 have a particle with weight of zero? This might be good to replace with an assertion (`CELER_ASSERT(weight != 0);`). In Celeritas we use "design by contract" to check assumptions rather than ignoring or propagating potentially bad data.

sethrj (commented):
I mistakenly thought your last pull request only pulled weights into Celeritas, not sent them back, which is why I didn't ask to add these test to that PR. There is one piece missing at least: the weights aren't being copied from the track state in `StepGatherExecutor`, and they're not copied from GPU to CPU in the `DetectorSteps` classes. I think you also need to extend the DetectorSteps test and double check that the `weights` option shows up everywhere `step_length` does.

test/celeritas/ext/SimpleSensitiveDetector.cc
Rashika-Gupta: Ideally, it should not be. But, added as a precautionary check just in case the weights were not initialised properly or were set to some default value of 0.0.

sethrj: Let's replace it with `CELER_ASSERT(weight > 0);` so that if we do get an uninitailized/improper result, we can track it back to the source and fix it there. Code gets bloated and hard to reason about otherwise :)

Rashika-Gupta: sure will do ! Thanks

Rashika-Gupta:
All done as suggested.

test/celeritas/ext/SimpleSensitiveDetector.cc
```diff
@@ -91,14 +91,18 @@ bool SimpleSensitiveDetector::ProcessHits(G4Step* step, G4TouchableHistory*)
     auto* pre_step = step->GetPreStepPoint();
     CELER_ASSERT(pre_step);
 
-    hits_.energy_deposition.push_back(step->GetTotalEnergyDeposit()
-                                      / CLHEP::MeV);
     hits_.step_length.push_back(step->GetStepLength() / CLHEP::cm);
+
+    real_type edep = step->GetTotalEnergyDeposit();
```
sethrj: Convert to the native unit system here, rather than when appending to the vector. Generally, anything inside celeritas should have a native type or a Quantity to specify its unit system. Also, if we're running with `real_type = float`, then the implementation here is `float( float(edep) /  double(MeV))` rather than `float(double(edep) / deouble(MeV))` so there's some precision loss.

test/celeritas/user/DetectorSteps.test.cc
```diff
@@ -255,6 +258,7 @@ TEST_F(DetectorStepsTest, TEST_IF_CELER_DEVICE(device))
     EXPECT_VEC_EQ(host_output.event_id, output.event_id);
     EXPECT_VEC_EQ(host_output.track_step_count, output.track_step_count);
     EXPECT_VEC_EQ(host_output.step_length, output.step_length);
+    EXPECT_VEC_EQ(host_output.weight, output.weight);
```
sethrj, line 261: I'm not sure if we've got you access to a GPU-enabled machine yet (it would be great to try using CADES?) but this fails due to missing code in `DetectorSteps.cu`, which is the CUDA implementation corresponding to `DetectorSteps.cc`:
```
289: [ RUN      ] DetectorStepsTest.device
289: info: Creating 1 device streams
289: /home/s3j/Code/celeritas/test/celeritas/user/DetectorSteps.test.cc:261: Failure
289:  Size of: output.weight
289:   Actual: 0
289: Expected: host_output.weight.size()
289: Which is: 180
289: Actual values: {};
289:
289:
289: [  FAILED  ] DetectorStepsTest.device (15 ms)
```

sethrj (changes requested):
Almost there 😅 unfortunately our CI doesn't have any actual GPUs on it, so we can only compile (rather than run) tests.

sethrj (approved):
It took me a bit to figure out why the test was still failing: there's yet another place to implement in the "compacting" kernel : see 6b7b6b121c025360a0e9305086be96b52f46879a . I do apologize for overthinking this implementation...

PR #1857, opened 2025-07-12
Fix counters in optical generator action

amandalund:
In the optical generator action we weren't updating the number of active tracks in the case that the generating process had no pending photons. This is fixed here and explains the change in https://github.com/celeritas-project/celeritas/pull/1844#discussion_r2186078020.


sethrj (approved):
Good catch!

PR #1858, opened 2025-07-12
Add optical primary generator

amandalund:
This adds a basic primary generator action for the optical loop. There are some limitations currently, which should be addressed in future work:
  -  The action is currently only used in testing and hasn't been incorporated into the main code/apps
  - The distributions for sampling the primary energy, position, etc. are currently hardcoded (monoenergetic, isotropic point source) rather than being runtime-configurable


src/celeritas/optical/detail/OpticalLaunchAction.hh
```diff
@@ -52,7 +52,7 @@ class OpticalLaunchAction : public AuxParamsInterface,
     {
         SPOpticalParams optical_params;
         size_type num_track_slots{};
-        size_type max_step_iters{};
+        size_type max_step_iters{static_cast<size_type>(-1)};
```
sethrj: If this is meant to be "maximum" rather than "invalid," it'd be better to write using `numeric_limits`.

src/celeritas/optical/gen/GeneratorData.hh
```diff
@@ -20,6 +20,26 @@
 
 namespace celeritas
 {
+//---------------------------------------------------------------------------//
```
sethrj, line 23: Should this class be in the optical namespace? Cherenkov and Scintillation are not because they interact first and foremost with "core" tracks. Maybe this guy should be OpticalPrimaryDistributionData (corresponding to the class below)? Or do we change the namespace for our other generators?

src/celeritas/optical/gen/OpticalPrimaryGenerator.hh
```diff
@@ -0,0 +1,81 @@
...
+    optical::TrackInitializer result;
+    result.energy = data_.energy;
+    result.position = data_.position;
+    result.direction = sample_angle_(rng);
+    do
+    {
+        result.polarization = make_unit_vector(
+            make_orthogonal(sample_angle_(rng), result.direction));
+    } while (CELER_UNLIKELY(
+        !soft_zero(dot_product(result.polarization, result.direction))));
```
sethrj, line 79: We do something like this in the other generators and the Rayleigh interactor (which additionally resamples the direction if orthogonality fails). We should probably add a note here about the semi-duplication. Once we have some additional interactors (including surface physics) that modify the polarization we could consider a helper function/class to eliminate the duplication.

src/celeritas/optical/gen/detail/PrimaryGeneratorAction.cc
```diff
@@ -0,0 +1,209 @@
...
+        counters.num_pending -= num_gen;
+        counters.num_generated += num_gen;
+        counters.num_vacancies -= num_gen;
+
+        // Update the generator counters and statistics
+        gen_counters.num_pending -= num_gen;
+        gen_counters.num_generated += num_gen;
+        aux_state.accum.num_generated += num_gen;
+    }
+    counters.num_active = state.size() - counters.num_vacancies;
```
sethrj: We should think about how to make this a helper class/function (maybe using CRTP) because this is all duplicated in `gen/detail/GeneratorAction.cc`.

sethrj (commented):
Looking good, but is there a straightforward way to reduce duplication?

src/celeritas/optical/gen/GeneratorData.hh
amandalund: I was considering moving the generator classes/data to the optical namespace now that all the generation is done in the optical loop (then `OpticalPrimaryGenerator` would be `optical::PrimaryGenerator`), but I'm undecided since the line for scintillation/Cherenkov is kind of fuzzy (the "offload" should stay in the celeritas namespace, and probably the params and data too... so does it make more sense to keep it all together?) What do you think?

sethrj: Yeah it's definitely gray areas, which means we should just stick with what you have for now without changing it (maybe adding a note for later). From a purely theoretical (but not the most practical) perspective I'd say all the generation and data is in `optical`, and all the offload data is in the main namespace. I think the main cherenkov/scint params themselves are main namespace? 🤔 

Definitely don't change it at this point 🙃 because we'll probably be leaning another way tomorrow.

amandalund: I tried it out earlier just to see: https://github.com/amandalund/celeritas/commit/c8bd7c3b7657d1e053a4942cd6a1854a8d61a0d0 but I wasn't really feeling it

amandalund:
@sethrj I added an interface class for the optical generators to try to reduce the duplication (open to other ideas though).

src/celeritas/optical/gen/detail/OpticalGeneratorBase.hh
```diff
@@ -0,0 +1,119 @@
...
+#include "celeritas/optical/action/ActionInterface.hh"
+#include "celeritas/phys/GeneratorInterface.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Common interface for genearating optical photons.
```
sethrj, line 23:
```suggestion
 * Common interface for generating optical photons.
```

sethrj (approved):
Awesome!

src/celeritas/optical/gen/GeneratorData.hh
sethrj: I actually like it better; it points out a few inconsistencies in our `detail` stuff and suggests a structure that might be better. Don't change it for this PR though :)

src/celeritas/optical/gen/detail/OpticalGeneratorBase.hh
sethrj: Re-running a CI cycle for such a tiny typo pains me 😆 I'm just going to resolve and merge

sethrj:
@lebuller

PR #1859, opened 2025-07-15
Add VolumeSurfaceSelector

### LINKED ISSUE #1351: Support optical surface interfaces

hhollenb:
Add selector for surface IDs given a track's pre-volume and post-volume instances. Encapsulates the logic for the order of preference for boundary and interface surfaces when multiple are present. Convenience functions for directly using a `GeoTrackView` are implemented as well.

The `InitBoundaryAction` will use this selector to determine the optical surface ID after a boundary crossing. Missing surfaces will be returned as invalid surface IDs.


src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convenience constructor to use IDs from geometry.
+ */
+CELER_FUNCTION VolumeSurfaceSelector::VolumeSurfaceSelector(
+    NativeCRef<SurfaceParamsData> const& params, GeoTrackView const& geo)
+    : VolumeSurfaceSelector(params, geo.volume_id(), geo.volume_instance_id())
```
sethrj, line 92: Won't the geometry usually be in the "post-volume" state? I think you may just keep the one constructor and then overload `operator()`.

test/celeritas/optical/VolumeSurfaceSelector.test.cc
```diff
@@ -0,0 +1,225 @@
...
+  public:
+    std::string_view geometry_basename() const override
+    {
+        return "optical-surfaces";
+    }
+
+  protected:
+    void SetUp() override
+    {
+        if (CELERITAS_CORE_GEO == CELERITAS_CORE_GEO_GEANT4)
```
amandalund: To remove this restriction (I think we'll want to test our other geometry engines) we could:
  - Have the test inherit from `GeantTestBase` instead
  - Get the volume and surface params from `this->core()->volume()` and `this->core()->surface()`
  - Require Geant4 for this test

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Select surface based on post-volume IDs.
+ *
+ * Returns an invalid \c SurfaceId if no surface data exists for the volumes.
+ */
+CELER_FUNCTION SurfaceId VolumeSurfaceSelector::operator()(
+    VolumeId post_volume, VolumeInstanceId post_volume_inst) const
+{
```
amandalund, line 105: Can we assume this?
```suggestion
{
    CELER_EXPECT(pre_volume_inst_ != post_volume_inst);
```

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+#include "celeritas/geo/GeoTrackView.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Retrieve the surface ID between two volume instances based on Geant4's
+ * interface / boundary priority.
```
amandalund: Even though we have it in the user manual, it might be helpful to explicitly state what that precedence is here.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+ * Construct with pre-volume IDs.
+ */
+CELER_FUNCTION VolumeSurfaceSelector::VolumeSurfaceSelector(
+    NativeCRef<SurfaceParamsData> const& params,
+    VolumeId pre_volume,
+    VolumeInstanceId pre_volume_inst)
+    : params_(params)
+    , pre_volume_(pre_volume)
+    , pre_volume_inst_(pre_volume_inst)
+{
```
amandalund, line 81:
```suggestion
{
    CELER_EXPECT(pre_volume_ && pre_volume_inst_);
```

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    if (!surface_id && post_volume < params_.volume_surfaces.size())
+    {
+        VolumeSurfaceView post_surface{params_, post_volume};
+        surface_id = post_surface.boundary_id();
+    }
+
+    return surface_id;
```
amandalund: Since we can always instantiate a `VolumeSurfaceView` when optical physics is enabled, we might be able to simplify/clarify the surface selection heuristic a bit:

```c++
VolumeSurfaceView pre_surface{params_, pre_volume_};
if (auto id = pre_surface.find_interface(pre_volume_inst_, post_volume_inst))
{
    // Interface surface from P0 -> P1
    return id; 
}
if (auto id = pre_surface.boundary_id())
{
    // Boundary surface of L0
    return id; 
}
// Boundary surface of L1
return VolumeSurfaceView(params_, post_volume).boundary_id();
```

Do we not currently handle this case?
>  If L1 is the child of L0 (crossing into an “enclosed” volume), then the boundary surface of L1

amandalund (commented):
Great work and nice addition @hhollenb!

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,125 @@
...
+ * Select surface based on post-volume IDs.
+ *
+ * Returns an invalid \c SurfaceId if no surface data exists for the volumes.
+ */
+CELER_FUNCTION SurfaceId VolumeSurfaceSelector::operator()(
+    VolumeId post_volume, VolumeInstanceId post_volume_inst) const
+{
+    SurfaceId surface_id{};
+
+    if (pre_volume_ < params_.volume_surfaces.size())
```
sethrj: I don't think this condition is necessary: per https://github.com/celeritas-project/celeritas/pull/1846#discussion_r2202968720 we should always have volume surfaces when using optical physics (and pre_volume should always be defined and should never be out of range).

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
sethrj: Unfortunately not; since the volume instances are the edges to the last node of the path, it's possible to have the same physical volume before and after.
```mermaid
graph TD
    Top --> B
    Top --> C
    C --> B
    B --> |vol inst| A
```

As a concrete example, think about the fuel region of a pin cell, for many pin cells stacked vertically.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
sethrj: Yeah I was going to ask about the child/mother priority: I think in practice it's poorly defined and implemented because it's using the *direct* parent/daughter relation but with the *final* volumes in the hierarchy. We could get that to work with additional help from a VolumeView (extension of VolumeParams) but I just don't think it's the right approach.

test/celeritas/optical/VolumeSurfaceSelector.test.cc
sethrj: Even better, we could simply use/extend the SurfaceTestBase from #1852. We only need surface IDs and structure for the surface selector, not any real geometry navigation.

test/celeritas/optical/VolumeSurfaceSelector.test.cc
```diff
@@ -0,0 +1,225 @@
...
+                          this->select_surfaces(VolumeInstanceId{4}));
+        }
+    }
+}
+
+//---------------------------------------------------------------------------//
+// Check selector correctly uses volume instances from a geo view
+TEST_F(VolumeSurfaceSelectorTest, geo_view_wrapper)
+{
+    if (CELERITAS_CORE_GEO == CELERITAS_CORE_GEO_GEANT4)
```
sethrj: Even if we did need the g4 geo, you should disable this at the cmake level: otherwise it looks like this test is running and passing when it's not doing anything at all.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
amandalund: Ah right, the unique instance is  the full path

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
hhollenb: We'd need to cache the pre-volume information from the geometry somehow, so I just had it draw directly from the track e.g.

```
// in pre-volume state
auto select_surface = track.surface_selector();
track.geo().cross_boundary();
// in post-volume state
SurfaceId surface = select_surface(track.geometry());
```
where the `surface_selector()` function in CoreTrackView looks like
```
return VolumeSurfaceSelector{params_.surface, this->geometry()};
```

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
hhollenb: I think this ambiguity should be resolved when we introduce layers. Multiple boundaries and interfaces will just be traversed depending on their ordering between the pre and post volumes. This code will then be responsible for handling whether boundaries or interfaces contribute to the current layer.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return surface_id;
+    }
+
+    if (auto surface_id = pre_surface.boundary_id())
+    {
+        return surface_id;
+    }
+
+    if (post_volume < params_.volume_surfaces.size())
```
amandalund: Are there cases where the post volume ID might be invalid?

hhollenb: I think you're right - previously I had run a test where the track was outside the geometry but that case is now covered by the `GeoTrackView` overload.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
sethrj: Cool! Makes sense.

test/celeritas/optical/VolumeSurfaceSelector.test.cc
```diff
@@ -0,0 +1,129 @@
...
+            SurfaceId{7},
+            SurfaceId{},
+            SurfaceId{},
+            SurfaceId{},
+            SurfaceId{},
+        };
+
+        EXPECT_VEC_EQ(expected_surfaces, select_surfaces(VolumeInstanceId{5}));
+    }
+}
```
amandalund, line 124: We could also add a test documenting a case where the result is different from what Geant4 would produce (e.g. going from a mother to a daughter where both have boundary surfaces).

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -0,0 +1,128 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Retrieve the surface ID between two volume instances based on Geant4's
+ * interface / boundary priority.
+ *
+ * Given (old, new) physical volumes P0, P1 corresponding to logical volumes
+ * L0, L1
+ * - Ordered (P0, P1) border surface
+ * - Skin surface of L1 if it's the daughter of L0
```
amandalund: Let's add a note for now that we haven't implemented this case, so this differs a bit from Geant4's heuristic.

amandalund (approved):
Thanks @hhollenb!

test/celeritas/optical/VolumeSurfaceSelector.test.cc
sethrj: Good idea; that way when/if we add that capability we can show the change in behavior.

sethrj (approved):
Awesome!

Issue #1860, opened 2025-07-15
Add Celeritas optical-only integration with existing Geant4


sethrj:
In the future, due to ongoing work by @whokion , we expect a new Geant4 feature to allow G4 optical generation processes to emit distributions of photons rather than photons. These distributions can then be offloaded to Opticks, Celeritas, etc. and will be supported in the long term by Geant4.

However, in the meantime for validation and for compatibility with older frameworks, we need to support an independent way to integrate Celertias' optical physics into a Geant4 application. The obvious way to do this is with an approach like Opticks, having a custom G4VProcess that replaces Cerenkov for example.

*Unlike* Opticks, this will *not* be just a copy of G4Cerenkov/G4Scintillation. Instead, it will convert Geant4 data into Celeritas types and use the Celeritas CherenkovOffload class to add generator distributions to the Celeritas optical loop. This will allow validation of more of Celeritas from inside a Geant4 context.

Note that this concept is similar to https://github.com/celeritas-project/celeritas/issues/1396, but this will be for optical processes rather than EM. Some technical challenges may be shared: for example, whether we can both pull data from *and* replace a process.

sethrj:
Proposed steps:
- Make a G4 example with optical physics; figure out how to cleanly replace the cerenkov/scintillatoin with Celeritas (custom modular physics list?)
- Add adapter classes for Geant4 classes that can be reused in geant importer, primary construction, hit processing etc (@sethrj )
- Add adapter class for HEP RNG (@sethrj )
- Create generator distributions in a new G4VProcess subclass using celeritas data
- Add an optical-only mode for setting up celeritas whose "flush" at end of event will run optical photons

amandalund:
I'm not sure we'll actually need to create custom Cherenkov/Scintillation processes for this. I think we should be able to set the "stacking" flag to false, so the number of photons is sampled in the Geant4 processes but the secondaries aren't generated. Then we can retrieve the number via `GetNumPhotons()`.

sethrj:
@amandalund Great idea, but we may want to consider both options since with a custom process the Celeritas imported materials and dn/dx calculator are used rather than Geant4 (could be good for validation).

amandalund:
True, that could be useful for validation. We might be able to get something working more quickly starting with the existing processes (wouldn't have to worry about changing the physics list or replacing the processes when integrating with an external application).

sethrj:
Yep, great point!

sethrj:
[...]

PR #1861, opened 2025-07-15
Begin exposing Cherenkov/optical loop outside of the optical collector 

sethrj:
Maybe I should break these into a few micro-PRs:
- Enable running (but not testing results) for cherenkov; it exposes the minor oddness that 10 GeV electrons have beta=1.0f in single precision
- Allow construction of ~~optical collector~~ offload distribution builder from low-level data, to be used in #1860
- Allow access to the optical collector from accel SharedParams, and allow access to the cherenkov/scintillation params from inside that

### DIFF
--- a/src/celeritas/optical/OpticalCollector.cc
+++ b/src/celeritas/optical/OpticalCollector.cc
@@ -174,6 +174,17 @@ auto OpticalCollector::cherenkov() const -> SPConstCherenkov
 
 //---------------------------------------------------------------------------//
 /*!
+ * Access scintillation params (may be null).
+ */
+auto OpticalCollector::scintillation() const -> SPConstScintillation
+{
+    if (!scint_offload_)
+        return nullptr;
+    return scint_offload_->params();
+}
+
+//---------------------------------------------------------------------------//
+/*!
  * Get the generator registry.
  */
 GeneratorRegistry const& OpticalCollector::gen_reg() const
--- a/src/celeritas/optical/OpticalCollector.cc
+++ b/src/celeritas/optical/OpticalCollector.cc
@@ -6,6 +6,7 @@
 //---------------------------------------------------------------------------//
 #include "OpticalCollector.hh"
 
+#include "corecel/Assert.hh"
 #include "corecel/data/AuxParamsRegistry.hh"
 #include "corecel/data/AuxStateVec.hh"
 #include "corecel/io/OutputInterfaceAdapter.hh"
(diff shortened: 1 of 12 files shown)


src/celeritas/optical/OpticalCollector.hh
```diff
@@ -111,6 +111,22 @@ class OpticalCollector
...
+    SPConstOpticalParams const& optical_params() const
+    {
+        return optical_params_;
+    }
+
+    // Access Cherenkov params (may be null)
+    SPConstCherenkov cherenkov() const;
+
+    // Access scintillation params (may be null)
+    SPConstScintillation scintillation() const;
```
amandalund, line 127: Needs to be implemented?

src/celeritas/optical/gen/CherenkovOffload.hh
```diff
@@ -37,14 +37,24 @@ namespace celeritas
...
-    // Construct with optical material, Cherenkov, and step data
+    // Construct directly
     inline CELER_FUNCTION
-    CherenkovOffload(ParticleTrackView const& particle,
-                     SimTrackView const& sim,
-                     optical::MaterialView const& mat,
-                     Real3 const& pos,
-                     NativeCRef<CherenkovData> const& shared,
-                     OffloadPreStepData const& step_data);
+    CherenkovOffload(units::ElementaryCharge charge,
```
amandalund, line 42: It doesn't look like this new constructor is used anywhere. What's the reason for adding it?

amandalund (approved):

src/celeritas/optical/gen/CherenkovOffload.hh
amandalund: > Allow construction of optical collector from low-level data

Or maybe this was a typo :). So this would be used if we created a custom Geant4 Cherenkov process?

sethrj: Exactly. It's currently delegated to by the Celeritas "view" constructor, so in that sense it is indeed "used". I can improve the documentation but I'm confused about what's a typo?

amandalund: "optical collector" in the PR description/blockquote, nothing in the code!

src/celeritas/optical/OpticalCollector.cc
```diff
@@ -162,7 +178,7 @@ OpticalCollector::OpticalCollector(CoreParams const& core, Input&& inp)
  */
 GeneratorRegistry const& OpticalCollector::gen_reg() const
 {
-    return *launch_->optical_params().gen_reg();
+    return *optical_params_->gen_reg();
```
amandalund, line 192: Any reason you prefer to store the optical params pointer in this class rather than get it from the launch action?

sethrj: Honestly I didn't notice at first. I'll fix.

sethrj: Oh it only exposes it as a const reference, and store it as a mutable pointer. I think keeping an SP const copy is easy enough

PR #1862, opened 2025-07-15
Add optical surface physics input and import surface physics data from Geant4

### LINKED ISSUE #1850: Import Geant4 optical surface physics data/selection
For #1512 we need to be able to import optical surface model selection and data for the various surfaces.

## KEY POINTS
- Optical models in Geant4 are "bundles" of separately selectable Celeritas models: see https://github.com/celeritas-project/celeritas/issues/1512#issuecomment-3019564068 
- Each surface will have one surface physics "optical model" for each of three groups of optical properties:
  * roughness: a description relating physical surface facet normals to average geometric normal
  * reflectivity: the probability of reflecting as a function of wavelength etc.
  * interaction: the resulting angular/polarity distribution from interacting with the surface
 - 

## Input structure

There's one group for each surface physics model type (roughness, reflection, interaction):
```c++
struct SurfacePhysics
{
    RoughnessModels roughness;
    ReflectivityModels reflectivity;
    InteractionModels interaction;
};
```

Each model contains one data member per predefined model, *plus* (eventually) a callback for users to add models. Each builtin member maps a surface layer (i.e., surface ID *plus* [eventually] layer ID) to model properties.
```c++
struct RoughnessModels
{
    using SurfaceLayer = SurfaceId; // currently
    //  using SurfaceLayer = pair<SurfaceId, SurfaceLayerId>; // someday
    map<SurfaceLayer, PolishedRoughness> polished;
    map<SurfaceLayer, SmearRoughness> smear;
    map<SurfaceLayer, GaussianRoughness> gaussian;
    // vector<OpticalModelFactory> custom; // someday
};
```

Each 
```c++
struct PolishedRoughness
{
};
struct SmearRoughness  // used by GLISUR
{
    real_type smeariness{};
};
struct GaussianRoughness  // used by UNIFIED
{
    real_type sigma_alpha{};
};
```

Structs can be reused as well, as needed for interaction:
```c++
struct ReflectionForm
{
    real_type specular_lobe{};
    real_type specular_spike{};
    real_type backscatter{};
    real_type lambertian{};
};

struct InteractionModels
{
    map<SurfaceLayer, ReflectionForm > dielectric_dielectric;
    map<SurfaceLayer, ReflectionForm > dielectric_metal;
    // vector<function<SP<OpticalModel>>> custom; // someday
};
```

### Helpers

Since this layout "unfolds" the physics so that a single surface will appear once per optical model type, we could add a helper function for constructing (and unit tests)

```c++
struct SurfacePhysicsSetter
{
inp::SurfacePhysics& phys;
template<class Roughness, class Reflectivity, class Interaction>
void operator()(SurfaceLayer layer, Roughness&& rough, Reflectivity&& refl, Interaction&& inter)
{
    if constexpr (std::is_same_v<Roughness, Polished>)
    {
        surface_physics.roughness.polished[id] = rough;
    }
    // ... 
}
};
```

## Importing from Geant4

Done via GeantGeoParams (@sethrj  in #1812):
- {pv, pv} -> border (interface) -> SurfaceId
- {lv} -> skin (boundary) -> SurfaceId

To access surface properties via other code (see https://github.com/celeritas-project/celeritas/pull/1812#issuecomment-2977507793):
1. Loop through `range(SurfaceId{surf_params.num_surfaces()})`
2. Run `geant_geo.id_to_geant(surf_id)` to get `G4LogicalSurface const* g4surf`
3. Load physics properties for each surface from `g4surf->GetSurfaceProperty()`

Importing is to be done in `celeritas/ext/detail` alongside the GeantImporter which does similar work.

## [Obsolete discussion from May 23](https://github.com/celeritas-project/celeritas/issues/1512#issuecomment-2905491800) with @stognini and @hhollenb 

<details>
No common properties

- One "inp" for each surface physics model
- Each surface physics model has a mapping of SurfaceId -> properties

```c++
namespace inp
{
enum class GlisurSurfaceType {
polished,
rough
};

struct GlisurModel
{ 
    using surface_idx = unsigned int;
    std::map<surface_idx, GlisurSurfaceType> surfaces;
};

struct UnifiedModelSurface
{
    std::vector<double> spike;
    std::vector<double> lobe;
    ...
};

struct UnifiedModel
{
    std::map<SurfaceId, UnifiedModelSurface> surfaces;
};

struct SurfacePhysics
{
    GlisurModel glisur;
    UnifiedModel unified;

    std::vector<std::string> surface_names; // indexed by SurfaceId
};

}
```

```c++
// rename ImportData -> LegacyImportData (supports ROOT IO)

struct ImportData
{
    LegacyImportData legacy;
    inp::SurfacePhysics surface_physics;
};
```

Inside GeantImport:
- loop over surface IDs
- Add to model-specific map for that surface if that surface model is supportd
- Warn if model is unavailable

During build:
- Construct GlisurModel from inp::GlisurModel
- Construct UnifiedModel from inp::UnifiedModel
- SurfacePhysics will check that surfaces have only a single model that applies
</details>

stognini:
This PR adds `inp::SurfacePhysics` to `inp::OpticalPhysics` and imports data from Geant4 accordingly.

The new `SurfacePhysics` untangles the existing Geant4 code design (with high code branching based on surface type/finish for each surface model) into an interactor/kernel-like structure to make it more suitable for GPUs.

The import process has to move away from `ImportData` to ultimately populate `inp::Problem`. To start this process, the `ImportData` is now starting to be populated with `inp` objects.


stognini:
TODO: 
- Create tests with glisur/unified model data
- Correctly verify that inputed data matches minimum requirements for each model
- Add  detector efficiency to Surface Physics?
   ```cpp
   SurfacePhysics
   {
       using DetectionEfficiency = std::map<SurfaceLayer, Grid>;
       ...
       DetectionEfficiency efficiency;
   };
   ```

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,221 @@
...
+ *   deep groove.
+ * - \c diffuse_lobe : Probability of internal Lambertian reflection.
+ *
+ * \note The sum of \c specular_lobe + \c specular_spike + \c back_scatter +
+ * \c diffuse_lobe probabilities must be equal to one. Diffuse lobe is not
+ * user-defined in Geant4 and is calculated from the other three.
+ */
+struct ReflectionForm
+{
+    real_type lambertian_roughness{-1};  //! \todo: range?
```
hhollenb: I'll suggest leaving out this parameter for now. I believe this refers to the `kSURFACEROUGHNESS` property which is another characterization of the material's reflectivity (i.e. in comparison of specular lobe / spike / back-scatter / diffuse). The Lambertian distribution itself doesn't take any parameters, and I can't find any discussion of what model this parameter came from or any examples of it being used. I don't think we lose anything by leaving it out for now, and that way we don't have hanging TODOs in the code

stognini: Correct. That ends up being used here:
https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/processes/optical/src/G4OpBoundaryProcess.cc?ref_type=heads#L1098

For this variable I did not check defaults in the manual yet, just seemed reasonable to add it. If no one really uses it, we just never hit that `if` statement. So happy to remove it from the struct/import and revisit this if we need in the future.

src/celeritas/io/ImportData.hh
```diff
@@ -106,6 +107,17 @@ struct ImportData
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * \c ImportData will eventually become \c inp::Problem once the import
+ * system populates all its \c inp:: objects.
+ */
+struct ImportData
+{
+    LegacyImportData legacy;  //!< Temporary container; Will be deprecated
+    inp::OpticalPhysics optical_physics;  //!< \todo: replace by inp::Physics
```
amandalund, line 114: To avoid changing `data` to `data.legacy` all over the code, is there any reason we can't keep the old `ImportData` struct and just put the `inp::OpticalPhysics` directly in there? We've already started replacing some of the old `Import` structs with `inp` there (e..g. the various grid classes).

doc/usage/input/physics.rst
```diff
@@ -10,18 +10,52 @@ Physics
...
 
+Optical
+^^^^^^^
+
+Optical photon _generation_ is a part of the standard stepping loop that manages
+EM, decay, and hadronic physics, but its _transport_ has its own separate
+stepping loop, where surface physics is the most complex part. Therefore, the
+``OpticalPhysics`` input includes optical photon _generation_ processes (such as
+Cherenkov and scintillation) and surface physics information. The latter
+describing how optical photons should interact with it.
```
sethrj, line 42: I wonder if we want separate "optical generation" physics, volumetric physics for optical photons, and surface physics for optical photons...

src/celeritas/io/ImportData.hh
stognini: I can do that. That was more of an organizational choice---as `inp` objects are pulled out, they start to show up outside `legacy`. But again, we can just make a clear separation in the original `ImportData` and write the todo documentation. No strong preference here. :)

src/celeritas/io/detail/ImportDataConverter.cc
```diff
@@ -33,39 +33,39 @@ ImportDataConverter::ImportDataConverter(UnitSystem usys)
 //---------------------------------------------------------------------------//
 void ImportDataConverter::operator()(ImportData* data)
 {
-    for (auto& m : data->geo_materials)
+    for (auto& m : data->legacy.geo_materials)
```
sethrj: Instead of all these changes, just use `LegacyImportData*` as the argument and call it from a new `ImportData*` function. Same for the data trimmer.

src/celeritas/mat/MaterialParams.cc
```diff
@@ -62,14 +62,14 @@ MatterState to_matter_state(ImportMaterialState state)
 std::shared_ptr<MaterialParams>
 MaterialParams::from_import(ImportData const& data)
 {
-    CELER_EXPECT(!data.geo_materials.empty());
-    CELER_EXPECT(!data.phys_materials.empty());
-    CELER_EXPECT(!data.elements.empty());
+    CELER_EXPECT(!data.legacy.geo_materials.empty());
+    CELER_EXPECT(!data.legacy.phys_materials.empty());
+    CELER_EXPECT(!data.legacy.elements.empty());
```
sethrj: Again, fewer changes by doing
```c++
MaterialParams::from_import(ImportData const& new_data)
{
    auto const& data = new_data.legacy;
    CELER_EXPECT(!data.geo_materials.empty());
    CELER_EXPECT(!data.phys_materials.empty());
    CELER_EXPECT(!data.elements.empty());
```

src/celeritas/ext/RootInterfaceLinkDef.h
```diff
@@ -47,17 +47,33 @@
...
+#pragma link C++ class celeritas::inp::OpticalPhysics+;
+#pragma link C++ class celeritas::inp::Polished+;
+#pragma link C++ class celeritas::inp::ReflectionAnalytic+;
+#pragma link C++ class celeritas::inp::ReflectionForm+;
+#pragma link C++ class celeritas::inp::ReflectionGrid+;
+#pragma link C++ class celeritas::inp::ReflectivityModels+;
+#pragma link C++ class celeritas::inp::RoughnessModels+;
+#pragma link C++ class celeritas::inp::SmearRoughness+;
+#pragma link C++ class celeritas::inp::SurfacePhysics+;
+#pragma link C++ class celeritas::inp::SurfaceNames+;
```
sethrj: Do we really want to add all these types to ROOT? I think we should consider abandoning the ROOT output for the time being. Or at least not writing any of the optical stuff to ROOT: we require Geant4 to be loaded to get the surface geometry information regardless.

sethrj (commented):
Still reviewing this, but when you see massive find-replace changes of the kind `data` -> `data.legacy` all over the code base, we should pause to consider whether it's the right approach, and probably open a separate PR for the refactoring aspect.

I'm not sure if we need ImportData to include the optical physics data, or if we can keep it alongside it for time being, building it in the same fashion that we do the `inp::Model` components.

If we do end up keeping the surface physics data alongside the import data, let's at least separate out the "ImportData is now ImportLegacyData" into a separate PR.

src/celeritas/io/ImportData.hh
sethrj: Just saw these comments: see below. Maybe we should have a brief meeting to discuss the restructuring. The whole ImportData refactoring is possibly a big jump in the scope of the next step.

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,250 @@
...
+ */
+struct SurfacePhysics
+{
+    //!@{
+    //! \name type aliases
+    using SurfaceNames = std::map<SurfaceLayer, std::string>;
+    using DetectionEfficiency = std::map<SurfaceLayer, Grid>;
+    //!@}
+
+    SurfaceNames names;
```
sethrj: We don't need surface names; those are input by the surface params. If the G4 properties here had separate names we could use those, but they don't.

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
...
+/*!
+ * Global surface normal with smearing.
+ *
+ * Polishness range is [0, 1], where 1 is maximum polishness.
+ *
+ * \note Used by the GLISUR model in Geant4.
+ */
+struct SmearRoughness
+{
+    real_type polishness{-1};  //!< [0, 1] where 1 means maximum polishness
```
sethrj: `polishness` isn't a word in English; `polish` is, but it's confusing (both a verb and a nationality!). It seems like computer graphics has the answer; it's more standard to use "roughness" (0 is fully specular, 1 fully diffuse) so even though it's backward from Geant4 maybe we shoul duse that instead.

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
...
+ * \note The sum of \c specular_lobe + \c specular_spike + \c back_scatter +
+ * \c diffuse_lobe probabilities must be equal to one. Diffuse lobe is not
+ * user-defined in Geant4 and is calculated from the other three.
+ */
+struct ReflectionForm
+{
+    // The sum of these properties must be equal to 1
+    Grid specular_lobe;  //!< [0, 1] probability
+    Grid specular_spike;  //!< [0, 1] probability
+    Grid back_scatter;  //!< [0, 1] probability
```
sethrj:
```suggestion
    Grid backscatter;  //!< [0, 1] probability
```
but maybe could also be specular_backscatter? Look at the original paper to see what options

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
...
+            }
+        }
+        return true;
+    }
+
+    //! Whether the data are assigned
+    explicit operator bool() const
+    {
+        return unity(specular_lobe) && unity(specular_spike)
+               && unity(back_scatter) && unity(diffuse_lobe)
```
sethrj: These are way too complicated for input. I'm not sure what level of validation we need, but we don't want to be doing summations over all the elements to determine whether they're "assigned". **All `operator bool` should be executable in constant time** i.e., they can't have loops.

Validation of this level should always be where they're used, so we can use this code later when implementing the roughness physics. For now just return `specular_lobe  && specular_spike && ... ;` if all grids are needed.

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * \todo Future work will allow the of use multiple surface paints/wrappings
+ * managed by different models. \c SurfaceLayer will pair a \c SurfaceId with a
+ * \c SurfaceLayerId that defiens paint/wrapping combinations.
+ */
+struct RoughnessModels
+{
+    std::map<SurfaceLayer, Polished> polished;
+    std::map<SurfaceLayer, SmearRoughness> smear;
+    std::map<SurfaceLayer, GaussianRoughness> gaussian;
```
sethrj: Instead of these maps of surface layers, since all surfaces should be assigned and we don't know if any/many layers will be present: let's define a typedef
```c++
template<class T>
using SurfaceLayerVector = std::vector<T>;
```
so for the present it's one element per surface ID, indexed by surface ID; then when we add layers we can easily do
```c++
template<class T>
using SurfaceLayerVector = std::vector<std::variant<T, std::vector<T>>;
```
so that for a given surface we can assign a layer of data but by default it's just a single element.

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/inp/Surfaces.hh
```
sethrj: Rename `SurfacePhysics` ; it's not how surfaces are defined (geometry) but physics (what happens on them)

src/celeritas/inp/Surfaces.hh
```diff
@@ -0,0 +1,251 @@
...
+    //! \name type aliases
+    using SurfaceNames = std::map<SurfaceLayer, std::string>;
+    using DetectionEfficiency = std::map<SurfaceLayer, Grid>;
+    //!@}
+
+    SurfaceNames names;
+    ReflectivityModels reflectivity;
+    RoughnessModels roughness;
+    InteractionModels interaction;
+    DetectionEfficiency efficiency;  //!< \todo: Keep as optional?
```
sethrj: I think efficiency is a terrible name for what this does in Geant4: from what I can tell, if you define it at all, it immediately absorbs the track without any other physics being applied. I almost think it should be part of a separate "surface detector" group in scoring that forces the surface to become a "detector surface" and then with a given probability passes the energy on to the detector. @hhollenb what do you think?

sethrj (commented):
Some preliminary comments on the input; haven't looked at the loading code. Since the `codespell` pre-commit was added on develop, you can run `pre-commit run -a codespell` to fix the typo(s). Also @pcanal can you help determine why the addition of the new fields (empty for most of the probelms) increased file size by a factor of 2?

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+{
+//---------------------------------------------------------------------------//
+// HELPER FUNCTIONS
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a string corresponding to the \c G4SurfaceType selection.
+ */
+char const* to_cstring(G4SurfaceType value)
```
sethrj: These should all be in an anonymous namespace since they're only used within this translation unit. I would also try to use a macro to avoid duplication/typos:
```c++
#define GSPL_SURFACE_PAIR(ENUMVALUE) {G4SurfaceType:: ENUMVALUE, #ENUMVALUE}
```
and write `to_cstring_impl = {GSPL_SURFACE_PAIR(dielectric_metal), ...}`.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+    return to_cstring_impl.find(value)->second.c_str();
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get a string corresponding to the \c G4OpticalSurfaceFinish selection.
+ */
+char const* to_cstring(G4OpticalSurfaceFinish value)
+{
+    static std::map<G4OpticalSurfaceFinish, std::string> const to_cstring_impl = {
```
sethrj: Also use unordered maps if they don't have to be sorted

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Populate surface physics data.
+ */
+inp::SurfacePhysics GeantSurfacePhysicsLoader::operator()()
+{
+    inp::SurfacePhysics result;
+
+    for (auto sid : range(SurfaceId(geo_->num_surfaces())))
+    {
```
sethrj: It might be more natural (and similar to the other builders) to take a `inp::SurfacePhysics&` in the constructor, and call `operator()` for each `SurfaceId` to add it.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+        auto const* g4log_surf = geo_->id_to_geant(sid);
+        CELER_ASSERT(g4log_surf);
+        auto* g4surf_prop = g4log_surf->GetSurfaceProperty();
+        CELER_ASSERT(g4surf_prop);
+        auto* g4opt_surf = dynamic_cast<G4OpticalSurface*>(g4surf_prop);
```
sethrj: Always check a dynamic cast.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+        auto const* g4mpt = g4opt_surf->GetMaterialPropertiesTable();
+        CELER_ASSERT(g4mpt);
+        detail::GeantMaterialPropertyGetter get_property{*g4mpt};
+
+        try
+        {
+            result.names.insert({sid, g4opt_surf->GetName()});
```
sethrj: remove this since we don't need the names (those come from the geometry)

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+            this->validate_model(sid, *g4opt_surf, result);
+        }
+        catch (RuntimeError const& e)
+        {
+            CELER_VALIDATE(
+                false,
+                << "failed to convert surface " << g4opt_surf->GetName()
+                << " with model " << to_cstring(g4opt_surf->GetModel())
+                << " and finish " << to_cstring(g4opt_surf->GetFinish())
+                << ": " << e.details().which << ", " << e.details().what);
```
sethrj: I know I suggested this in https://github.com/celeritas-project/celeritas/issues/1850#issuecomment-3057294846, but what we could also do is `CELER_LOG(error)` here instead of usurping the exception, then just `throw;` to keep the original exception propagating correctly. We could also define a `RichContextException` class and use `std::throw_with_nested`.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+                    << "' not available";
+                break;
+        }
+    }
+    else
+    {
+        CELER_LOG(error) << "inp::ReflectionForm incorrectly set up. Verify "
+                            "that all parameters have the same grid sizes and "
+                            "that their probability sums (for each energy "
+                            "point in the grid) are equal to 1";
```
sethrj: Should these be error messages or validation failures? And this one in particular, the validation criteria is so complicated that I think we want a greater level of detail, which we should probably do in the physics itself.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate diffuse lobe from the rest of the imported data.
+ *
+ * Since the total probability for all 4 properties is equal to one, the
+ * diffuse lobe can be calculated by subtracting the other three.
+ */
+inp::Grid GeantSurfacePhysicsLoader::calc_diffuse_lobe(
```
sethrj: This should be constant or (since it  doesn't even use class data) a free function

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+    auto const size = sl.x.size();
+    CELER_ASSERT(ss.x.size() == size && bc.x.size() == size);
+
+    inp::Grid result;
+    result.x = sl.x;
+    result.y.resize(size);
+    for (auto i : range(size))
+    {
+        // diffuse_lobe = 1 - specular_lobe - specular_spike - back_scatter
+        result.y[i] = real_type{1} - sl.y[i] - ss.y[i] - bc.y[i];
```
sethrj: I don't think we want to have overconstrained input like this. Eq 60 from [Nayar et al](https://doi.org/10.1109/34.85654) has $`I_{im} = C_{dl} + \frac{C_{sl}}{\cos \theta_r} \exp\left(-\frac{\alpha^2}{2\sigma_\alpha^2} \right) + C_{ss} \delta(\theta_i - \theta_r) \delta(\phi_r)`$ it looks like the diffuse lobe is an additional fudge factor, so I think this is the most natural term to drop.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+// PRIVATE MEMBER FUNCTIONS
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
```
sethrj:
```suggestion
//---------------------------------------------------------------------------//
```

There are a couple of these in the PR

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+/*!
+ * Collect reflectivity information from a given optical surface.
+ */
+void GeantSurfacePhysicsLoader::insert_reflectivity(
+    SurfaceId sid,
+    G4OpticalSurface const& surf,
+    detail::GeantMaterialPropertyGetter& get_property,
```
sethrj: These arguments get passed a lot. I think a helper class (in a local anonymous namespace) that temporarily stores the optical surface, the surface ID, the property getter, and a reference to `result` would help clear up the code. That kind of refactoring could be easily done inside the translation unit by Copilot (I highly recommend claude 3.7 'thinking').

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+ * Collect roughness information from a given optical surface.
+ */
+void GeantSurfacePhysicsLoader::insert_roughness(SurfaceId sid,
+                                                 G4OpticalSurface const& surf,
+                                                 inp::SurfacePhysics& result)
+{
+    using G4OSM = G4OpticalSurfaceModel;
+    auto const g4model = surf.GetModel();
+
+    if (g4model == G4OSM::glisur)
```
sethrj: For enums try to use `switch`.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,435 @@
...
+        if (polishness == 1)
+        {
+            // Perfectly polished surface
+            result.roughness.polished.insert({sid, inp::Polished{}});
+        }
+        else
+        {
+            // Smearing is available
+            inp::SmearRoughness smear{polishness};
+            CELER_ASSERT(smear);
```
sethrj: Assertions about user input need to be VALIDATE; or we should validate in the model later on. (This is probably a bigger picture question about where and how much to validate user input; especially as we add a python front end.)

src/celeritas/inp/Surfaces.hh
sethrj: Actually I forgot that each of these is going to be sparse, and since the models should build this in sequential/reproducible order we *should* keep it sorted as map. We also have to think about how we'd handle multiple layers: would we need a separate vector at the top level to keep track of how many interfaces per surface there are, so we can correctly search for the entries in the model maps? Or do we build those on the fly when we construct the surface physics?

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
sethrj: And we could also wrap this loop over surfaces in a `MultiExceptionHandler` so that we could print errors for all surfaces.

src/celeritas/inp/Surfaces.hh
hhollenb: I believe efficiency refers to the quantum efficiency of the detecting surface. I've definitely heard it used regularly when people talk about SiPMs so I think it's readily understood by people doing instrumentation.

I personally think that detection efficiency should be a property of the sensitive detector associated with that surface / volume. It receives hits that are absorbed photons and can apply whatever quantum efficiency it wants, but that's the responsibility of scoring / digitization and not surface physics.

hhollenb:
I remember having some issues with regenerating the root test data and having larger files. There might be an option set somewhere that outputs extra MC truth data or something? I tried tracking it down in my past branches but couldn't find it -- might want to check your simple-driver.py file and whatnot

src/celeritas/inp/Surfaces.hh
sethrj: Yeah I definitely am familiar with "intrisic detection efficiency", defined by Knoll as "ε = # pulses / number of radiation quanta incident on the detector" which is the input here. The weird side effect is that it forces the surface to be completely absorbent. So I agree, it is a detector property that forces a surface to be strictly a detector with a given efficiency.

sethrj: So @stognini leave this out for now but add a to-do in the docs for `inp::Scoring`

stognini: Yep. I agree that it makes sense for the quantum efficiency to be in sensitive detectors. I'll make a note in the scoring bit for it

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+           GSPL_ST_PAIR(dielectric_LUT),
+           GSPL_ST_PAIR(dielectric_LUTDAVIS),
+           GSPL_ST_PAIR(dielectric_dichroic),
+           GSPL_ST_PAIR(firsov),
+           GSPL_ST_PAIR(x_ray),
+#if G4VERSION_NUMBER >= 1110
+           GSPL_ST_PAIR(coated)
+#endif
+        };
+    return to_cstring_impl.find(value)->second.c_str();
```
sethrj: I also didn't notice: instead of std::string just use `const char*`. Since these are "static" strings rather than local, the pointers to them will still be valid. Also, formatting will be improved if you add a trailing comma, and finally make sure to check for invalid/unknown enums.
```suggestion

    static std::unordered_map<G4SurfaceType, char const*> const names = {
        GSPL_ST_PAIR(dielectric_metal),
        GSPL_ST_PAIR(dielectric_dielectric),
        GSPL_ST_PAIR(dielectric_LUT),
        GSPL_ST_PAIR(dielectric_LUTDAVIS),
        GSPL_ST_PAIR(dielectric_dichroic),
        GSPL_ST_PAIR(firsov),
        GSPL_ST_PAIR(x_ray),
#if G4VERSION_NUMBER >= 1110
        GSPL_ST_PAIR(coated),
#endif
    };
    if (auto iter = names.find(value); iter != names.end())
    {
        return iter->second;
    }
    return "UNKNOWN";
```

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+
+    static std::unordered_map<G4OpticalSurfaceModel, std::string> const to_cstring_impl
+        = {GSPL_OSM_PAIR(glisur),
+           GSPL_OSM_PAIR(unified),
+           GSPL_OSM_PAIR(LUT),
+           GSPL_OSM_PAIR(DAVIS),
+           GSPL_OSM_PAIR(dichroic)};
+    return to_cstring_impl.find(value)->second.c_str();
+
+#undef GSPL_OSM_PAIR
```
sethrj: Same comments here

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+ */
+bool total_prob_is_unity(inp::ReflectionForm const& rf)
+{
+    auto const& sl = rf.specular_lobe;
+    auto const& ss = rf.specular_spike;
+    auto const& bc = rf.backscatter;
+    auto const& dl = rf.diffuse_lobe;
+    auto const size = sl.x.size();
+    CELER_ASSERT(ss.x.size() == size && bc.x.size() == size
+                 && dl.x.size() == size);
```
sethrj: Since these are all user input, these shoudl be validate not assert. Also you're assuming they're on the same grid (what if they have same # points but different values). Maybe this isn't something we can easily validate?

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Verify that the sum of all \c ReflectionForm probabilities adds up to one.
+ */
+bool total_prob_is_unity(inp::ReflectionForm const& rf)
+{
+    auto const& sl = rf.specular_lobe;
+    auto const& ss = rf.specular_spike;
+    auto const& bc = rf.backscatter;
+    auto const& dl = rf.diffuse_lobe;
```
sethrj: Not sure if you saw the comment about diffuse lobe; I think you should just delete it entirely, and just assert here that each grid point is nonnegative and <= 1.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+ * Construct empty.
+ */
+GeantSurfacePhysicsLoader::GeantSurfacePhysicsLoader() {}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Populate surface physics data for a given \c SurfaceId .
+ */
+void GeantSurfacePhysicsLoader::operator()(SurfaceId sid,
+                                           inp::SurfacePhysics& result)
```
sethrj: Pass result into the constructor since it's "invariant"; just give it surface ID.

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+    }
+    catch (RuntimeError const& e)
+    {
+        auto const& surf = helper.surface();
+        CELER_LOG(error) << "failed to convert surface " << surf.GetName()
+                         << " (id " << sid.unchecked_get() << ") with model "
+                         << to_cstring(surf.GetModel()) << " and finish "
+                         << to_cstring(surf.GetFinish()) << ": "
+                         << e.details().which << ", " << e.details().what;
+    }
```
sethrj: I meant to log the error's context and then keep propagating the exception so the code doesn't keep going
```suggestion
        throw;
    }
```

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.hh
```diff
@@ -0,0 +1,64 @@
...
+
+  private:
+    SurfaceId const sid_;
+    G4OpticalSurface const* surface_;
+    G4MaterialPropertiesTable const* mpt_;
+};
+
+//---------------------------------------------------------------------------//
+#if !CELERITAS_USE_GEANT4
+inline GeantSurfacePhysicsHelper::GeantSurfacePhysicsHelper(SurfaceId)
```
sethrj: I don't think these are necessary; the class is only ever used within g4-only files.

src/celeritas/ext/GeantSurfacePhysicsLoader.hh
```diff
@@ -0,0 +1,74 @@
...
+// Geant4 forward declaration
+class G4OpticalSurface;  // IWYU pragma: keep
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Brief class description.
+ */
+class GeantSurfacePhysicsLoader
```
sethrj: Update docs, and this should probably be a detail class?

src/celeritas/ext/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,469 @@
...
+void GeantSurfacePhysicsLoader::operator()(SurfaceId sid,
+                                           inp::SurfacePhysics& result)
+{
+    CELER_EXPECT(sid);
+    detail::GeantSurfacePhysicsHelper helper(sid);
+    try
+    {
+        this->insert_reflectivity(helper, result);
+        this->insert_roughness(helper, result);
+        this->insert_interaction(helper, result);
```
sethrj: Maybe we should order these like they're used at runtime: roughness, reflectivity, interaction

amandalund:
FWIW, it looks like the root file sizes only increased by ~20% rather than 2x, which looks like it might be from the additional branches created for the optical surface physics. #1885 should bring this back down.

I also ran across this in the ROOT documentation:

> To prevent a data member from being written to the file, insert a ! as the first character in a single-line comment (//) following the declaration of the member.
>
> To accommodate doxygen-style documentation, this annotation can also be written as ///<!.

So apparently if we did want to e.g. exclude the optical physics date from being written to the ROOT file we could just make this change:

```diff
-    inp::OpticalPhysics optical_physics;  //!< \todo: replace by inp::Physics
+    inp::OpticalPhysics optical_physics;  ///<! \todo: replace by inp::Physics
```

sethrj:
Neat but ridiculous!

amandalund:
Haha, right? An LLM told me like two weeks ago I could disable output that way and I assumed it was hallucinating and gave up asking about ROOT. Then I saw that today in the docs, and... it works 🙃

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.cc
```diff
@@ -0,0 +1,60 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get property from material properties table.
+ */
+bool GeantSurfacePhysicsHelper::get_property(inp::Grid* dst,
+                                             std::string const& name)
+{
+    GeantMaterialPropertyGetter get_property{*mpt_};
+    return get_property(dst, name, {ImportUnits::mev, ImportUnits::unitless});
```
sethrj: Are all the properties MeV on x, unitless on y?

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Verify that all elements of a grid are within range [0, 1].
+ *
+ * Used to verify that \c ReflectionForm Grids are within the expected range.
+ */
+bool unity(inp::Grid const& grid)
+{
+    return std::any_of(grid.y.begin(), grid.y.end(), [](real_type const& val) {
```
sethrj: The name isn't right ; and it should be checking all the points not "any". Maybe:
```suggestion
bool is_fractional(inp::Grid const& grid)
{
    return std::all_of(grid.y.begin(), grid.y.end(), [](real_type const& val) {
```

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+            result_.interaction.dielectric_dielectric.insert(
+                {sid, std::move(refl_form)});
+            break;
+        case G4ST::dielectric_metal:
+            result_.interaction.dielectric_metal.insert(
+                {sid, std::move(refl_form)});
+            break;
+        default:
+            CELER_LOG(error) << "G4SurfaceType '" << to_cstring(interface_type)
+                             << "' not available";
```
sethrj: Shouldn't these be runtime errors, not error messages?

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+//! \todo: Expand \c SurfaceLayer to a `map<SurfaceId, SurfaceLayerId>`, where
+//! the \c SurfaceLayerId describes a set of layers.
+using SurfaceLayer = SurfaceId;
+//!@}
+
+//---------------------------------------------------------------------------//
+// SURFACE PHYSICS: interaction mechanisms / reflection models.
+//---------------------------------------------------------------------------//
+/*!
+ * Paramaters used by different reflection mechanisms.
```
sethrj:
```suggestion
 * Parameters used by different reflection mechanisms.
```
and also you can note that the three probabilities must sum to less than one at every grid point; and the remainder is diffuse scattering

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Interaction models for different interface types.
+ *
+ * Existing interface types are dielectrict-dielectric and dielectric-metal.
+ *
+ * \todo Future work may allow for custom interfaces.
```
sethrj:
```suggestion
 * This will be extended to allow user-provided interaction kernels.
```

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+struct SurfacePhysics
+{
+    //!@{
+    //! \name type aliases
+    using DetectionEfficiency = std::map<SurfaceLayer, Grid>;
+    //!@}
+
+    ReflectivityModels reflectivity;
+    RoughnessModels roughness;
+    InteractionModels interaction;
```
sethrj: Let's order them with the sequence they're called; same in the operator bool.
```suggestion;
    RoughnessModels roughness;
    ReflectivityModels reflectivity
    InteractionModels interaction;
```

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+ * Global surface normal with smearing.
+ *
+ * Roughness range is [0, 1], where 1 is maximum roughness. In Geant4 parlance,
+ * \code roughness = 1 - GetPolish(); \endcode.
+ *
+ * \note Used by the GLISUR model in Geant4.
+ */
+struct SmearRoughness
+{
+    real_type roughness{-1};  //!< [0, 1] where 1 means maximum roughness
```
sethrj:
```suggestion
    real_type roughness{-1};  //!< Scale from 0 = specular to 1 = diffuse
```

stognini:
Yep, I had problems in the past because of that Doxygen inconsistency and took me awhile to figure out why my dictionaries were failing. And `//!<` actually works, so just this guy `//!` is the problem.

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+
+    //! Whether the data are assigned
+    explicit operator bool() const { return static_cast<bool>(grid); }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Analytic reflectivity: use Fresnel equations.
+ */
+struct ReflectionAnalytic
```
sethrj: Perhaps FresnelReflection? Could there be other analytic models?

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+//---------------------------------------------------------------------------//
+// SURFACE DESCRIPTION: Reflectivity and models for surface normals.
+//---------------------------------------------------------------------------//
+/*!
+ * Surface reflectivity index, which can be a user-defined Grid, which is a
+ * function of energy (wavelength), or constant.
+ */
+struct ReflectionGrid
+{
+    Grid grid;
```
sethrj:
```suggestion
    Grid reflectivity;
```

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+ */
+struct ReflectionAnalytic
+{
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Simplest surface normal treatment.
+ */
+struct Polished
```
sethrj: What about
```suggestion
 * A polished (perfectly smooth) surface, where the facet normal is the macroscopic normal.
 */
struct NoRoughness
```

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+namespace celeritas
+{
+namespace inp
+{
+//---------------------------------------------------------------------------//
+// SURFACE DESCRIPTION: Reflectivity and models for surface normals.
+//---------------------------------------------------------------------------//
+/*!
+ * Surface reflectivity index, which can be a user-defined Grid, which is a
+ * function of energy (wavelength), or constant.
```
sethrj: I think these should be `XReflectivity` since we have `XRoughness`. ReflectionGrid sounds like it's a `Grid ` about reflection.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+}
+
+//---------------------------------------------------------------------------//
+// PRIVATE MEMBER FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Collect reflectivity information from a given optical surface.
+ */
+void GeantSurfacePhysicsLoader::insert_reflectivity(
+    detail::GeantSurfacePhysicsHelper& helper)
```
sethrj: Since the surface physics loader is in the detail namespace, you don't have to add `detail::` here.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+    inp::ReflectivityModels refl_mods;
+    if (!this->analytic_reflection_only(helper.surface()))
+    {
+        // Insert any model that includes user-defined grid reflectivity
+        inp::ReflectionGrid refl_grid;
+        helper.get_property(&refl_grid.grid, "REFLECTIVITY");
+        refl_mods.grid.insert({sid, std::move(refl_grid)});
+    }
+    refl_mods.analytic.insert({sid, inp::ReflectionAnalytic{}});
+    result_.reflectivity = std::move(refl_mods);
```
sethrj: This is wrong! You're replacing all the models so you'll only ever end up with the last surface added. Instead, use `auto& refl_mods = result_.reflectivity;` at the start of the function.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+    detail::GeantSurfacePhysicsHelper& helper)
+{
+    auto sid = helper.surface_id();
+    inp::ReflectivityModels refl_mods;
+    if (!this->analytic_reflection_only(helper.surface()))
+    {
+        // Insert any model that includes user-defined grid reflectivity
+        inp::ReflectionGrid refl_grid;
+        helper.get_property(&refl_grid.grid, "REFLECTIVITY");
+        refl_mods.grid.insert({sid, std::move(refl_grid)});
```
sethrj: You could define a templated method for your helper class to help
```c++

template<class T>
void GeantSurfacePhysicsHelper::insert(std::map<SurfaceLayer, T>& models, T&& surface_model)
{
    auto&& [iter, inserted] = models.insert({this->surface_id(), std::forward<T>(surface_model)});
    CELER_VALIDATE(inserted,
                   << "duplicate model inserted for surface ID " << this->surface_id());
}
```
so this would become the safer:
```suggestion
        helper.insert(refl_mods.grid, std::move(refl_grid));
```

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return true for Geant4 models/finishes that *ONLY* use analytical
+ * reflection.
+ *
+ * Currently, only the Unified model with [polished/ground]backpainted undergo
+ * uniquely through analytical reflection (i.e. Fresnel equations).
+ */
+bool GeantSurfacePhysicsLoader::analytic_reflection_only(
```
sethrj: This can be a free function

stognini:
Still about the ROOT files: After reorganization and what is likely the final format of `SurfacePhysics`, the `lar-sphere.root` goes from 33k to 35k, which seems reasonable for the extra surface info + overhead.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,427 @@
...
+
+    auto sid = helper.surface_id();
+    auto const& surf = helper.surface();
+    auto const g4model = surf.GetModel();
+    switch (g4model)
+    {
+        case G4OSM::glisur: {
+            // Get GLISUR surface polish
+            real_type roughness = real_type{1} - surf.GetPolish();
+            if (roughness == soft_equal(real_type{0}, roughness))
```
sethrj:
```suggestion
            if (soft_zero(roughness))
```

sethrj (commented):
*please* next time submit a shorter pull request. We should've started with a single model and built up, not thrown *everything* all at once.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,426 @@
...
+            if (surf.GetFinish() == G4OSF::polished)
+            {
+                // Perfectly polished surface
+                result_.roughness.polished.insert({sid, inp::Polished{}});
+            }
+            else
+            {
+                // Smearing is available (surf.GetFinish() == G4OSF::ground)
+                // Celeritas' roughness is the complement of Geant4 polish
+                inp::SmearRoughness smear{real_type{1} - surf.GetPolish()};
```
sethrj: Is  a polish of 1 used in practice? Should we replace it with a polished model?

pcanal:
> Neat but ridiculous!

Marking as transient should really be an attribute but they did not exist 30 years ago and we have not had the effort to properly introduce/use them.  Related (and reducing the urgency of such a move) is that the transientness can also be marked in the dictionary request file (when it is in the xml format)

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
stognini: Actually, re-re-reading the code, having `ground` finish and `polish = 1.0` is still possible. So you can have a perfectly polished surface that undergoes spike reflection (with finish == polished), or a ground finish that is also perfectly polished (with `GetPolish()` returning 1), but undergoing lobe reflection...

sethrj: Aren't spike vs lobe the reflection model, not the surface facet distribution?

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -0,0 +1,206 @@
...
+    {
+        return specular_lobe && specular_spike && backscatter;
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Interaction models for different interface types.
+ *
+ * Existing interface types are dielectrict-dielectric and dielectric-metal.
```
sethrj:
```suggestion
 * Existing interface types are dielectric-dielectric and dielectric-metal.
```

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -0,0 +1,426 @@
...
+        }
+
+        case G4OSM::unified: {
+            // Insert Gaussian if available
+            inp::GaussianRoughness gauss;
+            gauss.sigma_alpha = surf.GetSigmaAlpha();
+            if (gauss)
+            {
+                result_.roughness.gaussian.insert({sid, std::move(gauss)});
+            }
```
sethrj: What's the `else` case? You need a rougnness model right? I guess this gets checked in the validation? (That seems error prone.)

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
stognini: Correct. But when Geant's `G4OpBoundaryProcessStatus` is set as `LobeReflection` (for `ground` finish)  it also calls `GetFacetNormal()`, which uses `GetPolish()`. So the finish determines if it uses spike or lobe, and in turn the lobe model calls `GetFacetNormal()`.

The `GetFacetNormal()` checks `if (polish < 1)` to calculate `facetNormal = normal + (1. - polish) * smear`, otherwise manually sets `facetNormal = normal`, which is an unnecessary if/else statement, since if polish is 1 the extra term is just zero.

src/celeritas/inp/SurfacePhysics.hh
stognini: I like that! Even though negating roughness seems more work than just polished, it does unify the nomenclature throughout all the possible options.

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
hhollenb: Maybe `is_probability`?

sethrj:
> > Neat but ridiculous!
> 
> Marking as transient should really be an attribute but they did not exist 30 years ago and we have not had the effort to properly introduce/use them. Related (and reducing the urgency of such a move) is that the transientness can also be marked in the dictionary request file (when it is in the xml format)

@pcanal The more disruptive aspect is that it's a "surprise": I don't think most people would think a doxygen-like comment or an unusual comment character would trigger a behavior change in *anything*, unless there's a pragma-style comment a la clang-tidy or other source-slurping tools: have you considered deprecating that behavior and using `// root: hide` instead?

sethrj (commented):
Some probably-stale comments I left in from Thursday; let's try to get a final draft ready today?

pcanal:
> The more disruptive aspect is that it's a "surprise":

'Technically' we were there first :)  and doxygen is the one that 'surprisingly' disable a correctly functioning 'thing' :) :)

 > I don't think most people would think a doxygen-like comment or an unusual comment character would trigger a behavior change in anything

Yep, the original choice was indeed probably too succinct :)
 
>  unless there's a pragma-style comment a la clang-tidy or other source-slurping tools: have you considered deprecating that behavior and using // root: hide instead?

Some variations of this has been considered, it is believe that using `attribute` might make the most sense.  The 'hard' part is to finalize a coherent new 'grammar' covering all the options (there is also syntax to express length of array or to specify storage precision of floating numbers).  So it is non-zero effort that has not yet bumbled up on top of other priorities :(

src/celeritas/ext/detail/GeantSurfacePhysicsHelper.cc
stognini: Yes. At least for now. Added note.

src/celeritas/inp/SurfacePhysics.hh
stognini: This and the previous comment made me realize the nomenclature inconsistency. How about this instead, since both grid/analytic are reflectivity models:

```cpp
struct ReflectivityModels
{
    std::map<SurfaceLayer, GridReflection> grid;
    std::map<SurfaceLayer, FresnelReflection> fresnel;
};
```

hhollenb (approved):
Looked at the recent changes and looks good!

PR #1864, opened 2025-07-17
Refactor geometry construction to use static methods and shared pointers

sethrj:
This unpleasant refactor is necessary to pass the GeantGeo-constructed volumes into the ORANGE params constructor. It simplifies and clarifies the persistent existence of a Geant4 geometry (and wrapper functions) and improves construction of all the geometry types.

Geo params constructors now take only internal construction properties (vecgeom geo manager, geant4 physical volume, ORANGE input) for their constructor, rather than a filename. Additional static helper functions `from_geant`, `from_gdml`, etc. expose the ability to load that input and pass it into the constructor.

I also changed the global `geant_geo` to a weak_ptr from a raw pointer, which makes the code safer.


test/PersistentSP.hh
```diff
@@ -0,0 +1,141 @@
...
+    CELER_EXPECT(ptr);
+    CELER_LOG(debug) << "Updating persistent " << env_->desc << " to '" << key
+                     << "'";
+    env_->key = std::move(key);
+    env_->ptr = std::move(ptr);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Reset the shared poniter.
```
amandalund:
```suggestion
 * Reset the shared pointer.
```

amandalund (approved):
Looks good (and yes unpleasant), thanks @sethrj!

PR #1865, opened 2025-07-18
Move generators into optical namespace and refactor track generation utilities

amandalund:
Following up on https://github.com/celeritas-project/celeritas/pull/1858#discussion_r2203333108, this refactors some of the optical generation classes and core track initialization helper functions:
  - Moves the generators into the optical namespace
  - Moves the actions out of the detail namespace
  - Splits the generator/offload traits and algorithms into two separate files and namespaces
  - Moves the core track initialization utilities out of the detail namespace and eliminates a few
  - Adds `LogicalFalse` to replace two similar helper functions


src/celeritas/optical/gen/GeneratorData.hh
```diff
@@ -123,6 +125,19 @@ struct GeneratorStateData
...
+ * Get the number of photons from a distribution.
+ */
+struct GetNumPhotons
+{
+    // Return the number of photons to generate
+    CELER_FUNCTION size_type operator()(GeneratorDistributionData const& data) const
+    {
+        return data.num_photons;
+    }
+};
```
sethrj, line 142: Could be worth pointing out you can't just use a function here because it's passed from host to device by thrust.

sethrj (approved):
I think the organization is much better! Thanks for that.

src/celeritas/optical/gen/GeneratorData.hh
amandalund: Good point! And yeah I did end up liking it better after your suggestions.

src/corecel/math/Algorithms.hh
```diff
@@ -142,6 +142,30 @@ struct LogicalTrue<void>
     }
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Evaluate whether the argument is "false".
+ */
+template<class T = void>
+struct LogicalFalse
```
pcanal, line 150: Is it a `LogicalFalse` or a `LogicalNot` (or `LogicalNegate`) ? (it look like the later, given the former I sorta assumed the implementation to be `return false;`)

sethrj: Yeah good point, this should be https://en.cppreference.com/w/cpp/utility/functional/logical_not.html . Maybe we should also just change LogicalTrue to `Identity` with an implementation like https://en.cppreference.com/w/cpp/utility/functional/identity.html , which should work with all_of/any_of.

@pcanal could you

pcanal: sure, I can ...

sethrj: lol... 🐿️ 

Can you open up a PR that has the find-replace plus the geant4 typo uncorrection (you probably need to add it to the codespell dictionary in scripts/dev)?

pcanal: > this should be [en.cppreference.com/w/cpp/utility/functional/logical_not.html](https://en.cppreference.com/w/cpp/utility/functional/logical_not.html)

Do you mean literally (`celeritas::LogicalFalse` -> `std::logical_not` (C++14) or in spirit -> `celeritas::LogicalNot`)?

amandalund: Hmm yeah maybe we should get rid of it and use `std::logical_not` and `thrust::logical_not`.

sethrj: I think `thrust::identity` got removed though, so we may want to make our own to be generic. (And sorry @pcanal I forgot to answer you clearly; in spirit as celeritas::LogicalNot)

amandalund: and `std::identity` requires C++20

pcanal: See https://github.com/celeritas-project/celeritas/pull/1872

PR #1866, opened 2025-07-19
Reconstruct G4VTrackUserInfo and G4VProcess

esseivaju:
This save and restore user-defined track user info and process.

Because the `G4Track` owns the user info, we need to make sure that it is not deleted by Geant4 when the track is deleted, so we have to unset the pointer. The fastsim and user action offload interface only have access to a `G4Track const&` so I went with a const-cast but this is not ideal...

Alternatively, we could:

1. Update the `LocalTransporter` interface to `Push(G4Track&)` and const-cast in the fast-sim/user actions. 
2. Only support restoring track info when using the tracking manager offload.
3. With the tracking manager offload, we control the lifetime of the G4Track, so we could store the entire object. Unfortunately that's not the case in other offload mode and Geant4 will probably delete the track after we set the status to `fStopAndKill`


src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -83,6 +83,40 @@ get_step_status(DetectorStepOutput const& out, size_type step_index)
...
+/*!
+ * Restore the G4Track from the reconstruction data.
+ */
+void restore(G4TrackReconstructionData const& track_data, G4Track& track)
+{
+    CELER_EXPECT(track_data);
+    track.SetTrackID(track_data.track_id);
+    if (track_data.user_info)
+    {
+        track.SetUserInformation(track_data.user_info.get());
```
esseivaju: The ownership of the returned track user info is tricky to manage, we can't have the G4Track be the owner  because we'll need the user info again in the next steps...

sethrj: Yes, I never found a satisfactory way around that. I guess I didn't try resetting the G4Track pointer to tell Geant4 we were taking ownership? 🤔

src/celeritas/ext/detail/HitProcessor.hh
```diff
@@ -36,6 +38,20 @@ struct DetectorStepOutput;
 
 namespace detail
 {
+//---------------------------------------------------------------------------//
+//! Data needed to reconstruct a G4Track from Celeritas transport
+struct G4TrackReconstructionData
```
sethrj: This looks like a Geant4 built-in class name; let's call it `GeantTrackReconstructionData`. It could also probably be a nested class since it's just used to save/restore internally.

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -249,7 +290,7 @@ PrimaryId HitProcessor::register_primary(G4Track const& primary)
  */
 void HitProcessor::begin_event()
 {
-    celeritas_to_g4_track_id_.clear();
+    celeritas_to_g4_track_data_.clear();
```
sethrj: Should we try to clear user track information before the end of the previous event? Are there any user info classes that have destructors with side effects that assume the event is the same?

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -234,12 +274,13 @@ HitProcessor::~HitProcessor()
...
- * Register mapping from Celeritas PrimaryID to Geant4 TrackID.
+ * Register mapping from Celeritas PrimaryID to Geant4 TrackID and user info.
  */
 PrimaryId HitProcessor::register_primary(G4Track const& primary)
 {
-    auto primary_id = id_cast<PrimaryId>(celeritas_to_g4_track_id_.size());
-    celeritas_to_g4_track_id_.push_back(primary.GetTrackID());
+    auto primary_id = id_cast<PrimaryId>(celeritas_to_g4_track_data_.size());
+
+    celeritas_to_g4_track_data_.push_back(save(const_cast<G4Track&>(primary)));
```
sethrj: We could probably shorten that data member name now ;)

sethrj (approved):
Really cool capability!

src/celeritas/ext/detail/HitProcessor.cc
esseivaju: It's probably better, I guess there could be given the user can implement their custom class.

esseivaju: There's also the auxiliary track information that are deleted when the track is destroyed

src/celeritas/ext/detail/HitProcessor.cc
esseivaju: Yeah that's not ideal but not really another way around it. It would be nice if Geant4 had shared ownership of user info classes...

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -238,18 +244,29 @@ HitProcessor::~HitProcessor()
...
 {
-    auto primary_id = id_cast<PrimaryId>(celeritas_to_g4_track_id_.size());
-    celeritas_to_g4_track_id_.push_back(primary.GetTrackID());
+    auto primary_id = id_cast<PrimaryId>(g4_track_data_.size());
+    g4_track_data_.emplace_back(GeantTrackReconstructionData{
+        primary.GetTrackID(),
+        std::unique_ptr<G4VUserTrackInformation>(primary.GetUserInformation()),
+        primary.GetCreatorProcess()});
+    // Clear user information so that it doesn't get deleted with the G4Track
+    primary.SetUserInformation(nullptr);
```
esseivaju: The const_cast isn't actually needed because the user info are mutable

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -86,6 +86,36 @@ get_step_status(DetectorStepOutput const& out, size_type step_index)
...
+ */
+HitProcessor::GeantTrackReconstructionData::GeantTrackReconstructionData(
+    G4Track const& track)
+    : track_id{track.GetTrackID()}
+    , user_info{track.GetUserInformation()}
+    , creator_process{track.GetCreatorProcess()}
+{
+    CELER_EXPECT(*this);
+    // Clear user information so that it doesn't get deleted with the G4Track
+    track.SetUserInformation(nullptr);
```
sethrj, line 102: Of course geant4 has `mutable G4VUserTrackInformation* fpUserInformation` 🙄 

Even though it means people can avoid const casts, we're actually changing the member data. I think constructing with `G4Track&` and calling `GeantTrackReconstructionData` using a `const_cast` would make it clearer in the code that we're actually changing the track.

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -238,18 +274,24 @@ HitProcessor::~HitProcessor()
  */
 PrimaryId HitProcessor::register_primary(G4Track const& primary)
```
sethrj: This should also be mutable reference to indicate we're modifying the track by claiming ownership of its data.

sethrj (approved):
Sweet.

src/celeritas/ext/detail/HitProcessor.cc
esseivaju: Good point, better to be explicit in the interface. Updated.

esseivaju: Should we also propagate the mutable ref all the way to `LocalTransporter::Push`?

PR #1868, opened 2025-07-19
Add `begin_run` to optical launch action

amandalund:
This adds a `begin_run` to the optical launch action which sets the aux state pointer and constructs the action groups so that additional actions can be added after the launch action's construction. Combined with the previous refactoring of the generators, this means there’s no longer any restriction on the order in which we construct the actions in the optical collector.


sethrj (approved):
Very nice!

PR #1869, opened 2025-07-20
Add TrackProcessor class for reconstructing Geant4 tracks

esseivaju:
This moves the track reconstruction functionality to a new `TrackProcessor` class.


src/accel/LocalTransporter.cc
```diff
@@ -263,10 +263,11 @@ void LocalTransporter::Push(G4Track& g4track)
     PDGNumber const pdg{g4track.GetDefinition()->GetPDGEncoding()};
     track.particle_id = particles_->find(pdg);
 
-    // Generate Celeritas-specific PrimaryID and notify HitProcessor of mapping
+    // Generate Celeritas-specific PrimaryID
     if (hit_processor_)
     {
-        track.primary_id = hit_processor_->register_primary(g4track);
+        track.primary_id
+            = hit_processor_->track_processor().register_primary(g4track);
```
esseivaju, line 270: Instead of accessing the track_processor through the hit_processor, we could have both `LocalTransporter` and `HitProcessor` store a shared pointer to the `TrackProcessor`

sethrj: I think the current way is good: It's critical that the track processor is deleted on the corresponding worker thread because of Geant4-allocated thread-local objects. There's also shared data between the hit and track processor (G4Step) so I think it's best for it all to be retained in the same place for now.

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -219,18 +189,8 @@ HitProcessor::HitProcessor(SPConstVecLV detector_volumes,
...
-        auto track = std::make_unique<G4Track>(
-            new G4DynamicParticle(pd, G4ThreeVector()), 0.0, G4ThreeVector());
-        track->SetTrackID(0);
-        track->SetParentID(0);
-        track->SetStep(step_.get());
-
-        tracks_.emplace_back(std::move(track));
-    }
+    // Set the step for all tracks managed by TrackProcessor
+    track_processor_.set_step_for_tracks(step_.get());
```
sethrj: The step should be part of the TrackProcessor constructor, not set after the fact (splits initialization in to multiple steps, easy to forget). Either defer construction of the track processor (using unique_ptr/shared_ptr/optional) or have the TrackProcessor own the corresponding step (which I think makes more sense).

test/celeritas/ext/detail/TrackProcessor.test.cc
```diff
@@ -0,0 +1,349 @@
...
+    primary_track->SetParentID(0);
+
+    // Add user information
+    auto user_info = std::make_unique<MockUserTrackInformation>(42);
+    primary_track->SetUserInformation(user_info.release());
+
+    // Set creator process using mock process pointer
+    auto mock_process = std::make_unique<MockProcessPointer>("TestCompton");
+    auto* process_ptr = reinterpret_cast<G4VProcess*>(mock_process.get());
+    primary_track->SetCreatorProcess(process_ptr);
```
sethrj: This looks illegal. Can't you just create a minimal G4VProcess ?

sethrj (approved):
Great refactor and excellent testing! I think the step ownership should live with the track ownership, but other than that great!

test/celeritas/ext/detail/TrackProcessor.test.cc
esseivaju: I've added a minimal process, it's just a lot of boilerplate with all the pure virtual functions. `reinterpret_cast` between pointer to unrelated types (T1*->T2*->T1*) is legal, but accessing members is UB.

src/celeritas/ext/detail/HitProcessor.cc
esseivaju: Moved the step to the `TrackProcessor`

test/celeritas/ext/detail/TrackProcessor.test.cc
esseivaju: Seems that there is an issue linking G4VProcess:
```
/usr/bin/ld: test/celeritas/CMakeFiles/celeritas_ext_detail_TrackProcessor.dir/ext/detail/TrackProcessor.test.cc.o: undefined reference to symbol '_ZN10G4VProcess16SetMasterProcessEPS_'
/usr/bin/ld: /opt/view/lib64/libG4processes.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

src/celeritas/ext/detail/TrackProcessor.hh
```diff
@@ -0,0 +1,92 @@
...
+    void end_event();
+
+    // Register mapping from Celeritas PrimaryID to Geant4 TrackID
+    [[nodiscard]] PrimaryId register_primary(G4Track&);
+
+    // Restore track information for given primary and particle IDs
+    G4Track& restore_track(ParticleId, PrimaryId) const;
+
+    // Get the owned step
+    G4Step* step() const { return step_.get(); }
```
sethrj: Since this class owns the pointer, and it's never null, the interface should be returning a reference (and you can take the address of it whenever a pointer is needed).

src/celeritas/ext/detail/TrackProcessor.cc
```diff
@@ -0,0 +1,144 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Restore the G4Track from the reconstruction data. Returns the track for the
+ * given particle ID with restored primary track information if a valid
+ * PrimaryId is provided.
+ */
+G4Track& TrackProcessor::restore_track(ParticleId particle_id,
+                                       PrimaryId primary_id) const
+{
+    CELER_EXPECT(particle_id.unchecked_get() < tracks_.size());
```
sethrj:
```suggestion
    CELER_EXPECT(particle_id < tracks_.size());
```
will work just as well

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -138,29 +108,25 @@ HitProcessor::HitProcessor(SPConstVecLV detector_volumes,
...
+    {                                                                    \
+        if (!selection.points[StepPoint::LOWER])                         \
+        {                                                                \
+            HP_CLEAR_STEP_POINT(Reset##TITLE##StepPoint);                \
+        }                                                                \
+        else                                                             \
+        {                                                                \
+            auto* sp = track_processor_.step()->Get##TITLE##StepPoint(); \
+            sp->SetStepStatus(fUserDefinedLimit);                        \
+            step_points_[StepPoint::LOWER] = sp;                         \
```
sethrj: OK the hit processor has a non-owning reference to the steps (owned by the G4Step I think?) so perhaps it could also have a `G4Step* step_` that references the TrackProcessor-owned step?

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -448,26 +370,13 @@ void HitProcessor::operator()(DetectorStepOutput const& out, size_type i) const
...
+void HitProcessor::update_track(G4Track& track) const
 {
-    ParticleId id = out.particle[i];
-    CELER_EXPECT(id < tracks_.size());
-    G4Track& track = *tracks_[id.unchecked_get()];
-    step_->SetTrack(&track);
-
     // Copy data from step to track
-    track.SetStepLength(step_->GetStepLength());
+    track.SetStepLength(track_processor_.step()->GetStepLength());
```
sethrj: I think having a non-owning `G4Step* step_` would also be better here: nothing else in this function touches the track processor (it's only copying from lower-down structures up to the track).

src/celeritas/ext/detail/TrackProcessor.hh
```diff
@@ -0,0 +1,92 @@
...
+    CELER_DEFAULT_MOVE_DELETE_COPY(TrackProcessor);
+
+    // Clear G4Track reconstruction data
+    void end_event();
+
+    // Register mapping from Celeritas PrimaryID to Geant4 TrackID
+    [[nodiscard]] PrimaryId register_primary(G4Track&);
+
+    // Restore track information for given primary and particle IDs
+    G4Track& restore_track(ParticleId, PrimaryId) const;
```
esseivaju, line 56: Not sure if we want to have this be a const function. It modifies the temporary step/tracks so it's not const but it's called from const functions in the HitProcessor

src/celeritas/ext/detail/HitProcessor.cc
```diff
@@ -255,46 +210,13 @@ HitProcessor::~HitProcessor()
     try
     {
         CELER_LOG(debug) << "Deallocating hit processor";
```
sethrj: Maybe this try/log should live in the track processor and wrap `this->end_event()`? The log message was meant to highlight crashes that would happen from allocating/deallocating on different threads.

src/celeritas/ext/detail/TrackProcessor.hh
sethrj: Good question, it's the old "const pointer to mutable data" conundrum... leave it as you have now :)

sethrj (approved):
Last comments I swear 😆

PR #1870, opened 2025-07-21
Support single-precision Geant4->ORANGE conversion

sethrj:
The ability to make Geant4 geometry from ORANGE (#1169) with single precision was disabled due to a bit of laziness on my part. Occasionally this would cause surprise test failures (#1869). This adds the casts necessary to do double-> conversion.


test/orange/g4org/SolidConverter.test.cc
```diff
@@ -93,6 +93,14 @@ SignedSense to_signed_sense(EInside inside)
     CELER_ASSERT_UNREACHABLE();
 }
 
+bool sense_equal(SignedSense lhs, SignedSense rhs)
+{
+    // Disagreeing about'on' is fine
```
elliottbiondo: on

elliottbiondo (approved):

PR #1867, opened 2025-07-19
Add pre-commit spell checking

sethrj:
Inspired by a few recent obvious typos, and referencing https://github.com/WordPress/openverse/issues/378 ,  it's a good idea to add `codespell` to our pre-commit hook. It doesn't rigorously check all spelling (and thus has far fewer false positives), but it seems to catch obvious typos. If desired I can split the typo fixes into a separate PR.

PR #1871, opened 2025-07-21
Revert one unintentional typo change from #1867

pcanal:
This is small partial revert of commit 2b343e9f925c4650991bdf938bcd406d51c11480


sethrj (approved):
Thanks!

PR #1872, opened 2025-07-21
Rename LogicalFalse and LogicalTrue

pcanal:
Change LogicalFalse to LogicalNot
and LogicalTrue -> Identity
to clarify the semantic.


src/corecel/math/Algorithms.hh
```diff
@@ -123,22 +123,22 @@ struct Less<void>
...
-struct LogicalTrue<void>
+struct Identity<void>
 {
     template<class T>
-    CELER_CONSTEXPR_FUNCTION bool operator()(T const& value) const noexcept
+    CELER_CONSTEXPR_FUNCTION const T&& operator()(T const& value) const noexcept
     {
-        return static_cast<bool>(value);
+        return std::forward<T const>(value);
     }
```
sethrj: In C++20 `identity` isn't templated on anything; only its `operator()` is. Should we do the same (mirroring the inconsistency between the identity and other functional unary operators?)

pcanal: It is a bit more intrusive but seems to work.

src/corecel/math/Algorithms.hh
```diff
@@ -122,23 +122,12 @@ struct Less<void>
...
-template<>
-struct LogicalTrue<void>
+struct Identity
 {
     template<class T>
-    CELER_CONSTEXPR_FUNCTION bool operator()(T const& value) const noexcept
+    CELER_CONSTEXPR_FUNCTION const T&& operator()(T const& value) const noexcept
     {
-        return static_cast<bool>(value);
+        return std::forward<T const>(value);
```
amandalund: Should this be
```suggestion
    CELER_CONSTEXPR_FUNCTION const T&& operator()(T&& value) const noexcept
{
        return std::forward<T>(value);
```
?

src/corecel/math/Algorithms.hh
```diff
@@ -122,23 +122,12 @@ struct Less<void>
...
-    CELER_CONSTEXPR_FUNCTION bool operator()(T const& value) const noexcept
-    {
-        return static_cast<bool>(value);
-    }
-};
-
-//! Specialization of LogicalTrue with template deduction
-template<>
-struct LogicalTrue<void>
+struct Identity
```
amandalund, line 125: We should update the description for this and `LogicalNot`.

amandalund (approved):
Thanks!

src/corecel/math/Algorithms.hh
sethrj: Yes definitely should not be static_cast, sorry I missed that!

amandalund: Sorry, `static_cast` should be from the original, I'm not sure why the suggested change came out all mangled... I was wondering if we want the parameter to be a forwarding reference.

pcanal: The pattern is from https://en.cppreference.com/w/cpp/utility/functional/identity.html.

amandalund: Right, there it is:
```c++
template<class T>
constexpr T&& operator()(T&& t) const noexcept { return std::forward<T>(t); }
```
and we have:
```c++
template<class T>
constexpr const T&& operator()(T const& t) const noexcept { return std::forward<T const>(t); }
```

pcanal: yep.  This was actually a left over to make it 'work' with the template struct.  Once that was removed, the C++20 version seem to work as is.

src/corecel/math/Algorithms.hh
pcanal: Which description still need to be updated?

amandalund: A couple lines above here: `Evaluate whether the argument is "true".`, and the one above `LogicalNot`: `Evaluate whether the argument is "false".`

pcanal: Fair enough. Done.

PR #1873, opened 2025-07-22
Convert to using counterclockwise order for polygons

elliottbiondo:
This MR changes recent additions to ORANGE to use polygons specified in Geant4's clockwise order to instead use counterclockwise order, which is standard within in the computational geometry community. Polygons arriving from Geant 4 have their points reversed accoringly.

### DIFF
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -519,10 +519,9 @@ auto SolidConverter::extrudedsolid(arg_type solid_base) -> result_type
     // ordering.
     std::vector<G4TwoVector> g4polygon = solid.GetPolygon();
     VecReal2 polygon;
-    size_type n = g4polygon.size();
-    for (auto i : range(n))
+    for (auto i : range<int>(g4polygon.size()).step(-1))
     {
-        auto point = g4polygon[n - 1 - i];
+        auto point = g4polygon[i];
         polygon.push_back(Real2{scale_(point[0]), scale_(point[1])});
     }
 
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -514,7 +514,7 @@ auto SolidConverter::extrudedsolid(arg_type solid_base) -> result_type
 
     auto const& solid = dynamic_cast<G4ExtrudedSolid const&>(solid_base);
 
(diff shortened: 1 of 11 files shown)


test/orange/orangeinp/IntersectRegion.test.cc
```diff
@@ -636,11 +638,11 @@ TEST_F(ExtrudedPolygonTest, skewed)
     static char const* const expected_surfaces[]
         = {"Plane: z=10",
            "Plane: z=15",
-           "Plane: n={0.3152,0.3152,-0.89516}, d=-6.9658",
-           "Plane: n={-0.8165,0.40825,0.40825}, d=3.4701",
-           "Plane: n={0.35448,-0.70895,0.6097}, d=3.6511",
-           "Plane: n={0.45718,0.22859,-0.8595}, d=-5.1204",
            "Plane: n={-0.85138,0.42569,0.3065}, d=0.34055",
```
elliottbiondo, line 641: I thought something was wrong here because the senses don't change on any surfaces but it actually correct.

PXL_20250722_194120876~2

elliottbiondo:
@sethrj @mrguilima small one ready for review

src/orange/g4org/SolidConverter.cc
```diff
@@ -514,12 +514,15 @@ auto SolidConverter::extrudedsolid(arg_type solid_base) -> result_type
...
+    // counterclockwise ordering for polygons whereas GEANT4 uses clockwise
+    // ordering.
     std::vector<G4TwoVector> g4polygon = solid.GetPolygon();
     VecReal2 polygon;
-    for (auto i : range(g4polygon.size()))
+    size_type n = g4polygon.size();
+    for (auto i : range(n))
     {
-        auto point = g4polygon[i];
+        auto point = g4polygon[n - 1 - i];
```
sethrj: You should be able to use `range(g4polygon.size()).step(-1)` to do the same thing.

elliottbiondo:
P.S. obviously this wasn't a big priority, but I didn't like the feeling of digging myself into a deeper hole with the next MR I am working on.

mrguilima (approved):
Looks good to me.
All tests passing locally (GPU-based tests not tried though).

sethrj (approved):

elliottbiondo:
@sethrj RTM pending CI

PR #1874, opened 2025-07-23
Add volume params visitor

sethrj:
This refactors the volume visitor to support VolumeParams as a type to visit, allowing us to unfold and manipulate the geometry structure outside geant4/vecgeom. Since the `max_depth` in the geometry interface had different meanings between the geometry types, it is now removed from the geo params interface.

It also adds `VolumeToString`, a `std::visit`-oriented visitor class for the upcoming use of volume instance IDs in place of labels for ORANGE input.

These changes are both to implement #1749 .

### DIFF
--- a/src/geocel/g4/VisitVolumes.hh
+++ b/src/geocel/g4/VisitVolumes.hh
@@ -31,7 +31,7 @@ class GeantVolumeAccessor final
         return result;
     }
 
-    //! Outgoing instance nodes from a volume
+    //! Outgoing instance edges from a volume
     ContainerVolInstRef children(VolumeRef parent) final
     {
         CELER_EXPECT(parent);
--- a/src/geocel/g4/VisitVolumes.hh
+++ b/src/geocel/g4/VisitVolumes.hh
@@ -32,10 +32,10 @@ class GeantVolumeAccessor final
     }
 
     //! Outgoing instance nodes from a volume
-    VecVolInstRef children(VolumeRef parent) final
+    ContainerVolInstRef children(VolumeRef parent) final
     {
         CELER_EXPECT(parent);
-        VecVolInstRef result;
+        ContainerVolInstRef result;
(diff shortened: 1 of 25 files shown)


test/geocel/Volume.test.cc
```diff
@@ -174,6 +306,41 @@ TEST_F(VolumeTest, complex_hierarchy)
...
+        visit(nv, VolumeInstanceId{0});
+
+        static std::string const expected_names[]
+            = {"0:0", "1:2", "2:4", "2:6", "1:3", "2:4", "2:6"};
+        EXPECT_VEC_EQ(expected_names, nv.names);
+    }
+
+    {
+        MaxVisitor mpv{volumes_.volume_labels(), {}};
+        visit(mpv, VolumeId{0});
```
amandalund, line 342: Should we be checking the result?

src/geocel/g4/VisitVolumes.hh
```diff
@@ -17,22 +17,30 @@
...
+    VolumeRef volume(VolumeInstanceRef parent) final
+    {
+        CELER_EXPECT(parent);
+        auto result = parent->GetLogicalVolume();
+        CELER_ENSURE(result);
+        return result;
+    }
 
-    static void get_children(PV const& parent, std::vector<PV const*>& dst)
+    //! Outgoing instance nodes from a volume
```
amandalund: Edges?

amandalund (approved):
Very nice!

src/geocel/g4/VisitVolumes.hh
sethrj: Yep, I'd been thinking about reformulating the geometry representation as a 'bipartite graph', where volumes can only point to instances and instances to volumes, because the "world" volume in Geant4 has a volume instance as a placement and that can't be represented if the instances are edges. But I'll save the change for another time.

PR #1875, opened 2025-07-24
Refactor optical params and stepping loop for optical-only mode

amandalund:
As a first step towards enabling optical-only runs, this decouples the optical params from the optical collector and the stepping loop from the launch action.


src/celeritas/optical/Transporter.hh
```diff
@@ -0,0 +1,80 @@
...
+{
+class CoreParams;
+template<MemSpace M>
+class CoreState;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Transport all pending optical tracks to completion.
+ */
+class Transporter
```
amandalund, line 33: We may want to call this something else.

src/celeritas/inp/Physics.hh
```diff
@@ -72,6 +73,9 @@ struct OpticalPhysics
     //! Generate scintillation photons
     bool scintillation{true};
     //!@}
+
+    //! Volume labels for sensitive detectors
+    std::optional<std::vector<Label>> detector_labels;
```
sethrj: Can we pass `SDParams` through from the core params? This isn't exactly physics input.

src/celeritas/optical/Transporter.hh
sethrj: Stepper maybe if we were doing a single loop, but since we're transporting all the tracks and incrementing `accum.flushes` we could call it `Flusher` 😂

src/celeritas/optical/Transporter.hh
```diff
@@ -0,0 +1,80 @@
...
+
+  public:
+    // Construct with problem parameters and setup options
+    explicit Transporter(Input);
+
+    // Transport all pending optical tracks on the host
+    void operator()(CoreParams const&, CoreStateHost&) const;
+
+    // Transport all pending optical tracks on the device
+    void operator()(CoreParams const&, CoreStateDevice&) const;
```
sethrj: Since we're constructing with SP<CoreParams> maybe we should just call it with the state?

src/celeritas/setup/Problem.cc
```diff
@@ -279,62 +281,96 @@ auto build_along_step(inp::Field const& var_field,
...
-        }
-    }
+    auto num_streams = inp.optical_params->max_streams();
+    inp.num_track_slots = ceil_div(cap.tracks, num_streams);
+    inp.buffer_capacity = ceil_div(cap.generators, num_streams);
+    inp.auto_flush = ceil_div(cap.primaries, num_streams);
+    inp.max_step_iters = p.tracking.limits.optical_step_iters;
 
-    CELER_ASSERT(oc_inp);
+    CELER_ENSURE(inp);
```
sethrj: The reason I avoided `inp` here is because it's also a valid namespace; so this is now shadowing `celeritas::inp`.

test/celeritas/LArSphereBase.hh
```diff
@@ -45,6 +48,28 @@ class LArSphereBase : public GeantTestBase
         result.processes |= GeantImportDataSelection::optical;
         return result;
     }
+
+    SPConstOpticalPhysics build_optical_physics() override
```
sethrj: If we add another optical physics test, this should probably move to GeantPhysics base class with a flag for enabling optical physics (or a new OpticalGeantPhysics base...)

sethrj (approved):
Nice, thanks for slogging through that refactor! Great next step.

src/celeritas/optical/Transporter.hh
amandalund: I hate it 😂 but I've got nothing better

src/celeritas/inp/Physics.hh
amandalund: Yeah, definitely not the right place. I could put it in `inp::Scoring`?

sethrj: Ah right, this is where I told @lebuller to for the time being use a manual list of volumes rather than pulling from the geant4 simulation. I think these honestly can be pulled from `VariantSetVolume force_volumes`, or maybe we should just change that to `VariantSetVolume volumes` to which we append the Geant4 SDs during setup time... the SdParams should be building from that. I think that should be a separate PR; for now since we don't even use the SDs let's just leave this out and leave SD params blank. I will add a task to https://github.com/celeritas-project/celeritas/issues/1397 ; it might be a good next step for @lebuller since it doesn't require anything in the optical physics itself

PR #1876, opened 2025-07-25
Update VolumeSurfaceSelector and optical BoundaryAction for oriented surfaces

hhollenb:
A small PR to break up #1801 with some of our recent discussions.

- Add `SubsurfaceDirection` enum for forward and reverse directions
- `VolumeSurfaceSelector` now returns an `OrientedSurface` which is a surface ID along with a direction. Previously it only searched for interface surfaces indexed by (pre,post). Now also searches for (post,pre) interface surfaces and returns them along with a reverse direction.
- Moved `BoundaryAction` to `InitBoundaryAction` and updated the executor to use the `VolumeSurfaceSelector`. Doesn't initialize the surface physics state (that is in #1801 where the params are implemented), but does queue the next post step action to be the `PostBoundaryAction`.
- Added `PostBoundaryAction`. Currently it will just kill the photon, but the executor will be updated to unset the surface ID, update geometry, and set the optical material.


src/celeritas/optical/surface/PostBoundaryAction.hh
```diff
@@ -2,26 +2,30 @@
 // Copyright Celeritas contributors: see top-level COPYRIGHT file for details
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 //---------------------------------------------------------------------------//
-//! \file celeritas/optical/action/BoundaryAction.hh
+//! \file celeritas/optical/surface/PostBoundaryAction.hh
```
amandalund: Since this class is identical to `InitBoundaryAction` except for the executors and labels, could we have a single templated `BoundaryAction` instead of duplicating it?

src/celeritas/optical/CoreTrackData.hh
```diff
@@ -33,15 +33,16 @@ struct CoreScalars
...
+    ActionId post_boundary_action{};
     ActionId tracking_cut_action;
 
     StreamId::size_type max_streams{0};
 
     //! True if assigned and valid
     explicit CELER_FUNCTION operator bool() const
     {
-        return boundary_action && max_streams > 0;
+        return init_boundary_action && post_boundary_action && max_streams > 0;
```
amandalund: Since they're both post-step actions we can check that they're executed in the correct order:
```suggestion
        return post_boundary_action && init_boundary_action < post_boundary_action && max_streams > 0;
```

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+        return sim.post_step_action() == track.init_boundary_action()
+               && sim.status() == TrackStatus::alive;
+    }());
+
+    auto geo = track.geometry();
+    CELER_EXPECT(geo.is_on_boundary());
+
+    // SurfaceParams not built for all cases, use this to temporarily
+    // disable surface physics and kill tracks immediately
+    if (!surface_physics_enabled)
```
amandalund: Should be fixed by #1878: if optical physics is enabled we should always have surface data, even if all surfaces are invalid.

amandalund (commented):
I still need to make sure I understand the surface selector changes, but this is looking great!

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+
+    // Move the particle across the boundary
+    geo.cross_boundary();
+    if (CELER_UNLIKELY(geo.failed()))
+    {
+        track.apply_errored();
+        return;
+    }
+    else
+    {
```
sethrj: Return/else is unnecessary; let's just return early as the exceptional case

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+            track.apply_errored();
+        }
+        else
+        {
+            track.sim().status(TrackStatus::killed);
+        }
+        return;
+    }
+
+    auto select_surface = track.surface_selector();
```
sethrj: Add a comment that the surface selector *must* be created locally before geometry changes; unlike most of the views, it holds persistent information that's temporary (and for this reason, I think we should actually *not* make it part of the track; let's make the track strictly return views, and the surface selector let's build from those views here).

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+    // Move the particle across the boundary
+    geo.cross_boundary();
+    if (CELER_UNLIKELY(geo.failed()))
+    {
+        track.apply_errored();
+        return;
+    }
+    else
+    {
+        auto oriented_surface = select_surface(geo);
```
sethrj: Note that the surface is found using the post-crossing geometry status

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+
+        if (!oriented_surface.surface)
```
sethrj: Maybe we should have the struct define an operator bool like `OnTface` from `orange/univ/detail/Types.hh`

src/celeritas/optical/Types.hh
```diff
@@ -20,6 +20,17 @@ using ScintParticleId = OpaqueId<struct ScintParticle_>;
...
+{
+
+enum class SubsurfaceDirection : int
+{
+    forward = 1,
+    reverse = -1
+};
+
+}  // namespace optical
+
```
sethrj: Should be moved down to the `enumerations` section.

Also it may be premature optimization, but let's implement it as a bool flag and converted to integers only when indexing is needed.
```c++
enum Direction : bool { reverse, forward }; // Note that reverse is less than forward as is numeric
CELER_FORCEINLINE_FUNCTION int to_offset(Direction d) { return 2 * static_cast<int>(d) - 1; }
```

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -100,34 +109,46 @@ CELER_FUNCTION VolumeSurfaceSelector::VolumeSurfaceSelector(
...
  * Convenience function to use IDs from geometry.
  */
-CELER_FUNCTION SurfaceId
+CELER_FUNCTION auto
 VolumeSurfaceSelector::operator()(GeoTrackView const& geo) const
+    -> OrientedSurface
 {
     if (geo.is_outside())
-        return SurfaceId{};
+        return OrientedSurface{SurfaceId{}, SubsurfaceDirection::forward};
```
sethrj:
```suggestion
        return {};
```

src/celeritas/optical/surface/PostBoundaryAction.hh
hhollenb: I wasn't 100% certain when writing it that they'd be the same, but I think that'd be the way to go!

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -22,9 +23,10 @@ namespace optical
  *
  * Given (old, new) physical volumes P0, P1 corresponding to logical volumes
  * L0, L1
- * - Ordered (P0, P1) interface surface
- * - Boundary surface of L0
- * - Boundary surface of L1
+ * - Ordered (P0, P1) interface surface, forward
+ * - Ordered (P1, P0) interface surface, reverse
+ * - Boundary surface of L0, forward
+ * - Boundary surface of L1, reverse
```
amandalund: I'm not sure I understand the second case ((P1, P0) interface surface). Why would we select a surface the track doesn't "see", and why does it take precedence over the boundary surfaces?

src/celeritas/optical/surface/InitBoundaryExecutor.hh
```diff
@@ -0,0 +1,100 @@
...
+        {
+            // If there's no surface, mark photon as killed
+            track.sim().status(TrackStatus::killed);
```
amandalund: I know we discussed this in one of our meetings, but I'm having trouble finding where in the boundary process the track is killed if there's no optical surface... as far as I can tell it's still tracked as long as the new volume has an optical material.

sethrj: Yeah if there's no surface defined it defaults to:
```c++
  fModel             = glisur;
  fFinish            = polished;
  G4SurfaceType type = dielectric_dielectric;
```
and that uses the before/after materials. So I guess we'll need a "default" model for our use cases as well.

sethrj: Actually for our input (#1862) since we're pulling in `std::map`s of SurfaceId -> model, what about 'pre-populat'ing an empty `SurfaceId{}` key in the Geant4 importer? That would preserve the default Geant4 behavior while allowing users to change it. Our surface physics class can use this empty surface ID to store the default models. Or maybe it would make more sense to add an imaginary "default" surface with a real ID, which would make the data access pattern for the surface physics map a lot easier.

hhollenb: It might make more sense to have a default surface settings as a parameter for `SurfacePhysicsParams` to keep it separated from the geometry description. It's a default physics model so we can keep defaulted roughness/reflectivity/interaction IDs in the scalars and have them be assigned to missing surface IDs. Mainly I just want to avoid propagating defaults or fictitious surfaces out into the VolumeParams or SurfaceParams.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
hhollenb: This is based off my understanding that we're storing interfaces layered from P0 -> P1 but we don't store the same surface P1 -> P0. Using examples from Geant4 unified model:

1. unified, polished, dielectric-dielectric: Scintillator --> WLS fiber

If the user defines a surface (Scintillator, WLS fiber) then there won't be a corresponding surface (WLS fiber, Scintillator) unless explicitly implemented. Photons originating in the scintillator and crossing into the fiber will see the surface but photons in the fiber and crossing into the scintillator won't. Physically it's the same interface surface so it should still take precedence over boundary surfaces. Users can still override it by explicitly adding a (WLS fiber, Scintillator) interface surface.

2. unified, polished back painted, dielectric-dielectric: Crystal --> Glass

The back painted surface represents a wrapping around the crystal with an air gap, so the back painted surface (Crystal, Glass) physically represents Crystal --> Gap --> Wrapping --> Glass. A photon in the crystal crossing into glass should traverse the layers in that order. A photon originating in the glass and crossing into the crystal should traverse the layers as Glass --> Wrapping --> Gap --> Crystal. If we don't look up the (P1, P0) surface then the photon just passes through the wrapping and gap.

src/celeritas/optical/surface/InitBoundaryExecutor.hh
amandalund: We'll need to differentiate between the case where there's no optical surface and no optical material in the new volume (kill the track, like we do currently) and the case where there's no optical surface but the new volume does contain an optical material (use the default model). Would it be straightforward/possible to categorize "missing" surfaces this way during import?

src/celeritas/optical/surface/PostBoundaryAction.hh
hhollenb: Tried to break up the boundary action into a templated version. Could use some feedback on it since I used type traits and it ended up being a bit messy.

hhollenb:
Added `volume_id()` to `GeantGeoTrackView` and `VecgeomTrackView` so that `VolumeSurfaceSelector` is able to use it and compile. Currently they're just recasting `impl_volume_id()` so either we'll have to update their mappings to canonical volumes eventually or have some switch for compiling the surface physics with orange volume ids.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
amandalund: In the first example, if a photon is crossing from the fiber into the scintillator, won't it select the (P1 scintillator, P0 fiber) interface surface even though it's not supposed to see it?

sethrj:
Yep I've got a branch working on updating those.

src/celeritas/optical/CoreTrackView.hh
```diff
@@ -65,17 +65,17 @@ class CoreTrackView
     // Return a physics view
     inline CELER_FUNCTION PhysicsTrackView physics() const;
 
-    // Return a surface selector for the current volume
-    inline CELER_FUNCTION VolumeSurfaceSelector surface_selector() const;
```
amandalund, line 69: We might want to add a method here to return a `VolumeSurfaceView`. Then the `VolumeSurfaceSelector` could be constructed with pre- and post-`VolumeSurfaceView` instances and a pre volume instance ID (or possibly cache the instance ID in the surface view?)

This way we wouldn't need to explicitly pass the surface params data to the boundary executor and could instead get everything from the `CoreTrackView`. It might also be a little clearer than constructing and invoking`operator()` with the same geometry state before and after it's changed.

src/celeritas/optical/surface/detail/BoundaryActionTraits.hh
```diff
@@ -0,0 +1,62 @@
...
+    constexpr static char const* action_name = "optical-boundary-init";
+    constexpr static char const* action_desc
+        = "Initialize optical boundary crossing action";
+
+    template<MemSpace M>
+    static InitBoundaryExecutor
+    construct(CoreParams const& params, CoreState<M>&)
+    {
+        return InitBoundaryExecutor{params.surface()->ref<M>()};
+    }
```
amandalund: If the init- and post-boundary executors have the same member layout (currently I don't think they need any data members), we could remove these `construct()` methods.

src/celeritas/optical/CoreTrackView.hh
sethrj: 100% agree. As I think I commented earlier, all other track views have no temporary state.

src/celeritas/optical/surface/InitBoundaryExecutor.hh
sethrj: The no-material case should be pretty easy to differentiate; in the InitBoundary we just check the new optical ID and absorb if it's absent. If I'm understanding what you mean by "missing" surfaces, almost every surface in the geometry (i.e. a boundary or before-after pair) is going to be missing, and I think we want to avoid creating one border surface for every volume.

Do we want the default surface physics model to be configurable? I think it would be more clear to the user if we do store the default as part of the input, which we can later export/output for feedback.

amandalund: Yes, I think having the default in the input is a good idea, and it wouldn't hurt to make it configurable if we can. Agreed about checking the optical ID in the executor; I think I misunderstood what you were proposing in your earlier comment. Would the `SurfaceParamsData` remain unchanged, we'd just have one extra "default" model in the input/surface physics data?

sethrj: That would be ideal, but I'm not sure how to do this cleanly: the geometry specifies the surfaces, but the surface physics needs an extra surface ID, and we pass around SurfaceParams to resize stuff to fit. We could add one "pretend surface" to the end of the list (maybe using `SurfaceId{}` for the input data and `SurfaceModel`-returned IDs, but reassigning it `SurfaceId{surface_params.num_surfaces()}` afterward). However, I fear that approach could be confusing: maybe less so if we have a scalar `SurfaceId default_surface` in the physics params data?

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -40,28 +41,31 @@ namespace optical
...
-    // Convenience function to use IDs from geometry
-    inline CELER_FUNCTION SurfaceId operator()(GeoTrackView const&) const;
+    inline CELER_FUNCTION OrientedSurface
+    operator()(VolumeSurfaceView const& post_volume,
+               VolumeInstanceId post_volume_inst) const;
 
   private:
-    NativeCRef<SurfaceParamsData> const& params_;
-    VolumeId pre_volume_;
+    VolumeSurfaceView pre_surface_;
```
sethrj, line 68: Views are always pointing to data managed externally, and their implementations may not "cache" the starting values, so it's dangerous to assume a view won't change after it's created. (I know that's not the case here but it still seems error prone.) How about we add a `SurfaceParamsRef const&` accessor to `VolumeSurfaceView`, then in this constructor we can still pass the view but simplmy keep the starting volume ID and a reference to the SurfaceParamsRef.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -71,69 +75,39 @@ class VolumeSurfaceSelector
...
-    if (auto surface_id = pre_surface.boundary_id())
+    else if (auto surface_id = pre_surface_.boundary_id())
     {
-        return surface_id;
+        return {surface_id, SubsurfaceDirection::forward};
+    }
+    else
+    {
+        // If no boundary in post-volume then return invalid SurfaceId
+        return {post_surface.boundary_id(), SubsurfaceDirection::reverse};
```
sethrj: At first this comment sounds like it's inconsistent with the code but I get what you mean. Maybe
```suggestion
        // Return volume boundary from the opposite direction, which is just a "false"
        // OrientedSurface if there is no post-volume boundary.
        return {post_surface.boundary_id(), SubsurfaceDirection::reverse};
```

sethrj (commented):
Great! One last comment about the views, which I think might need another revision on my part since we probably should make them about volume instance rather than volume.

Speaking of, the other reason I wanted to write the SurfaceParams indexed by Volume is that we could conceivably add different surface specifications beyond PV/PV interfaces and LV boundaries. For example, with ORANGE, we could specify individual faces for each volume (using `ImplSurfaceId`): this capability is necessary for nuclear reactor-style problems where you have reflecting boundaries on one surface of the  world box and vacuum boundaries on the others.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
amandalund: That's definitely true of the `TrackView`s which are constructed with params and state data, but won't views that are only constructed with the constant params data (like `ParticleView`, `MaterialView`, this one) not change?

sethrj: OK, that's a good point. Sure, let's keep as is.

hhollenb: Yeah I was thinking that the `VolumeSurfaceView` was independent of the track and acts like a `MaterialView`. This should also be pretty straightforward to update if we decide to refactor `SurfaceParams` later.

src/celeritas/optical/surface/VolumeSurfaceSelector.hh
```diff
@@ -71,69 +75,43 @@ class VolumeSurfaceSelector
...
-        return surface_id;
+        return {surface_id, SubsurfaceDirection::forward};
+    }
+    // L1 boundary surface
+    else
+    {
+        // Return the L1 boundary surface from the opposite direction. If no
+        // boundary surface exists, an invalid OrientedSurface is returned.
+        return {post_surface.boundary_id(), SubsurfaceDirection::reverse};
     }
```
amandalund: No `else` needed for either of these cases

amandalund (approved):
Great, thanks @hhollenb!

sethrj (approved):

PR #1879, opened 2025-07-28
Protect Celeritas::G4 CMake target against double creation

drbenmorgan:
Discovered during integration in the ATLAS Athena framework, which calls `find_package(Celeritas)` more than once. In this case, CMake will error out on the second call, as the `Celeritas::G4` `ALIAS` target already exists.

The fix here simply checks whether the `Celeritas::G4` target exists first before creating it if it doesn't. The problem and fix can be confirmed by adding two `find_package(Celeritas)` calls in any of the examples and running the install tests, but I wasn't sure if that would be acceptable given this might confuse users.


sethrj (approved):
Thanks @drbenmorgan ! Since `find_package` tends to be expensive it might behoove Athena to move that up toward the top level, but it's good to have this fix. No test necessary.

PR #1880, opened 2025-07-29
Add volume test base

sethrj:
Refactor the `VolumeTestBase` to look more like `VolumeTest`:
- The base class stores a shared_ptr to a `VolumeParams`
- Replace `make_volume_input` with `virtual shared_ptr<VolumeParams> build_volumes() const =0` that lives in the test base, and define a base class SetUp that saves `volumes_ = this->build_volumes()` and checks with CELER_ASSERT.
- The `volumes()` accessor has `CELER_EXPECT(volumes_)` and returns the dereferenced shared pointer
- The refactored construction methods define `inp::Volumes in; ....; return std::shared_ptr<VolumeParams>(std::move(in));` .

It's probably going to interfere with #1876 so we can wait till that's in first.


hhollenb:
#1876 is going to be waiting until VolumeInstanceIds are implemented for ORANGE, so I'm fine if you want to publish this when you're ready and I can update my PR later when it can pass the ORANGE tests

amandalund (approved):
Nice refactor!

PR #1881, opened 2025-07-29
Remove spurious optical error messages

amandalund:
There were a couple errors and warnings when running with optical physics enabled that @Rashika-Gupta and others have reported that aren't actually errors.


sethrj (approved):
Thanks!

PR #1883, opened 2025-07-29
Refactor Implementation Volume IDs

hhollenb:
This is a quick draft of refactoring the volume IDs and the implementation volume IDs for geometries and their mappings to `VolumeParams`. Will update info soon when we figure out what specifically this PR will consist of.


hhollenb:
Superceded by #1886 and #1889

PR #1884, opened 2025-07-29
Rename volumes to "implementation volumes"

sethrj:
Working with @hhollenb on #1883 toward #1749, we're doing a global find-replace so that we correctly reflect that the "volume IDs" being returned from the geometry and the geo track views are actually the *implementation* volume IDs (how each geometry engine represents the model) rather than the *canonical* volume IDs (which are how the user specified the model).


src/celeritas/inp/Physics.hh
```diff
@@ -63,8 +63,8 @@ struct OpticalPhysics
     //!@{
     //! \name Optical photon generation
 
-    /*! \todo Replace with a mapping of \c VolumeId to \c ScintillationPhysics
-     * or \c CherenkovPhysics
+    /*! \todo Replace with a mapping of \c ImplVolumeId to \c
```
amandalund:
```suggestion
    /*! \todo Replace with a mapping of volume to \c
```

src/orange/OrangeTrackView.hh
```diff
@@ -453,7 +453,10 @@ CELER_FUNCTION Real3 const& OrangeTrackView::dir() const
 
 //---------------------------------------------------------------------------//
 /*!
- * The current volume ID.
+ * The current canonical volume ID.
+ *
+ * This is the volume identifier in the user's geometry model, not the ORANGE
+ * implementation of it.
```
amandalund: Why does ORANGE define a `VolumeId` that's the same as its `ImplVolumeId`?

amandalund (approved):

src/orange/OrangeTrackView.hh
sethrj: Right, this can be deleted. I've got an *actual* volume ID implementation upcoming.

hhollenb (approved):

PR #1885, opened 2025-07-30
Prevent splitting during ROOT I/O to reduce file size

amandalund:
By [preventing splitting](https://root.cern/manual/io/#preventing-splitting) when writing the imported data to ROOT, we can reduce the size of files. From the ROOT docs:

> If told, TTree will create (sub-) branches for each member of a class and its base classes. If such a member is a class itself, that member’s type can also be split. The recursion level of nested splitting is called the “split level”; it can be configured during branch creation.
>
>If the split level is set to 0, there is no splitting: all data members are stored in the same branch. Data members can also be configured to be non-split as part of the dictionary; see → [I/O](https://root.cern/manual/io_custom_classes). The default split level of 99 means to split all members at any recursion level.

This should help with the increased file sizes in #1862.


sethrj (approved):
Thanks @amandalund !

PR #1887, opened 2025-07-31
Use class constructors rather than initialization for Array to fix CMS

### LINKED ISSUE #1863: Missing braces warning from GCC
From @sanmayphy while trying to build CMS:

```
Error message :
>> Compiling  src/SimG4Core/Application/src/Phase2EventAction.cc
In file included from /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/celeritas/Types.hh:19,
                 from /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/celeritas/global/ActionInterface.hh:10,
                 from /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/accel/AlongStepFactory.hh:15,
                 from /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/CeleritasG4.hh:16,
                 from src/SimG4Core/Application/interface/Phase2EventAction.h:22,
                 from src/SimG4Core/Application/src/Phase2EventAction.cc:1:
/cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/geocel/Types.hh:75:22: error: missing braces around initializer for 'double [3]' [-Werror=missing-braces]
   75 |     Real3 pos{0, 0, 0};
      |                      ^
/cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02898/el8_amd64_gcc12/external/celeritas/v0.6.0-cb1099433c65b894c311c917958aa58f/include/geocel/Types.hh:76:22: error: missing braces around initializer for 'double [3]' [-Werror=missing-braces]
   76 |     Real3 dir{0, 0, 0};
      |                      ^
cc1plus: some warnings being treated as errors
gmake: *** [config/SCRAM/GMake/Makefile.rules:1920: tmp/el8_amd64_gcc12/src/SimG4Core/Application/src/SimG4CoreApplication/Phase2EventAction.cc.o] Error 1
```

We've seen this before, it's an overzealous/opinionated GCC warning (see [GCC 95330](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95330), [Clang 21629](https://bugs.llvm.org/show_bug.cgi?id=21629) ). Options to pursue are:
- Suppressing it inside Celeritas CMake (https://github.com/celeritas-project/celeritas/blob/4f17b671fcb5ffd031a49f890e1c49edb1246b48/src/CMakeLists.txt#L41) but this may not fix CMS which isn't using CMake to set flags
- Writing new Array constructors that don't use aggregate initialization
- Changing all use cases of this in Celeritas header files (yuck)

sethrj:
This changes the `Array` class from using aggregate initialization, which was supported in `std::array` from C++14 to simplify construction, to using varargs-based construction. It's mostly backward compatible, but it now fails if you try to write `Real3 foo{1, 2};` which previously was equivalent to writing `Real3 foo{1, 2, 0}`.

Should close #1863 ; @whokion or @sanmayphy could you try building CMS without the `-Wno-error=missing-braces` from https://github.com/cms-sw/cmsdist/pull/9973  and verify this works?


whokion:
> Should close #1863 ; @whokion or @sanmayphy could you try building CMS without the `-Wno-error=missing-braces` from [cms-sw/cmsdist#9973](https://github.com/cms-sw/cmsdist/pull/9973) and verify this works?

Yes, this was discussed in this morning’s meeting. Sanmay will test it using the latest CMSSW pre-release.

src/corecel/random/params/XorwowRngParams.cc
```diff
@@ -43,39 +43,40 @@ XorwowRngParams::XorwowRngParams(unsigned int seed)
...
-         {0x79a31c08u, 0xc28d5d18u, 0x91649708u, 0x8b1bb8f8u, 0xf9158a86u},
-         {0x670f870eu, 0xa7bb9766u, 0xef013c78u, 0xeb4a1373u, 0x256f3323u},
-         {0xf333af18u, 0x3b4e266bu, 0xa872663eu, 0x3888cd82u, 0x4daf13ecu},
-         {0x75689dc9u, 0x036bf3a9u, 0x64ce979cu, 0x3bdff14fu, 0x51e43048u},
-         {0xef06fe75u, 0xefd6d1d8u, 0x09319075u, 0x69d568f5u, 0x5b2bd898u},
-         {0xf05ae255u, 0xc4df4ca0u, 0xf032420cu, 0xf44ae9f0u, 0xe0298de2u},
-         {0x02308bc9u, 0xdbe74deeu, 0xf4c5fe7du, 0xaac571a1u, 0xaa1f5f8bu},
-         {0x6e8043cau, 0x0ed24ff6u, 0x1e668b6au, 0x538fc45fu, 0x4bfb509du},
-         {0x9f564f7bu, 0x543973e4u, 0x9b33ee2cu, 0x149df73au, 0x58f31585u}}};
+    static ArrayJumpPoly const jump = {{
```
whokion, line 46: Is `'{{'` still required instead of `'{'`?

whokion (approved):
Looks good to me!

src/corecel/random/params/XorwowRngParams.cc
sethrj: Yes unfortunately; it's a consequence of using an Array inside an Array; they both have variable-argument templates, so the compiler can't deduce it. Using the extra braces initializes using the `T const (&)[N]` constructor rather than the varargs constructor.

Alternatively we could write:
```c++
    using JP = JumpPoly;
    static ArrayJumpPoly const jump = {
        JP{0x00000002u, 0x00000000u, 0x00000000u, 0x00000000u, 0x00000000u},
        JP{0x00000010u, 0x00000000u, 0x00000000u, 0x00000000u, 0x00000000u},
....
```
but I think just including the extra braces is easier.

src/corecel/cont/Array.hh
```diff
@@ -45,88 +49,116 @@ struct Array
...
+    CELER_CEF Array(T first) : d_{first} {}
 
-    T data_[N];  //!< Storage
+    //! Construct with the array's data
+    template<class... Us>
+    CELER_CEF Array(T first, Us... rest) : d_{first, static_cast<T>(rest)...}
+    {
+        // Protect against leaving off an entry, e.g. Array<int, 2>(1)
+        static_assert(sizeof...(rest) + 1 == N,
+                      "All array entries must be explicitly specified");
```
esseivaju, line 78: Wouldn't missing entries simply be zero-initialized?

sethrj: With aggregate initialization yes, but I figured it'd be more safe+clear for us to make it explicit. There was a `Real3 point{x, y};` in the code. C/C++ initialization rules are so quirky: I also saw a `VecReal results{8}` which, unlike `VecReal results(8)` constructs a vector with a single element 8 rather than a vector of size 8 ☹️

src/corecel/cont/Array.hh
```diff
@@ -45,88 +49,116 @@ struct Array
...
+
+  private:
+    T d_[N];  //!< Storage
 };
 
+//---------------------------------------------------------------------------//
+// DEDUCTION GUIDES
+//---------------------------------------------------------------------------//
+
+// Note: this differs from std::array, which deduces from T rather than the
```
pcanal, line 148: Probably I just lost in the English grammar ... does the 'which' applies/describes `std::array` (my reading) or `celeritas::Array` (what the text seems to describe).

src/corecel/cont/Array.hh
esseivaju: Yeah it does get very tricky when you start mixing aggregate initialization, default initialization, I need to refresh my memory on this 😅  I wonder if `Array(std::initializer_list<T>)` constructor could be used instead of variadic template/C-array, it could simplify things.

sethrj: I don't think it works on device :(

src/corecel/cont/Array.hh
sethrj: Your reading is correct: `std::array<T, N>`; vs `celeritas::Array<std::common_type_t<T, Us...>, N>`. Maybe you missed "rather than"?

pcanal: Not really, what I actually missed is the actual purpose of this declaration :(

esseivaju (approved):
Looks good!

PR #1888, opened 2025-07-31
Add RevolvedPolygon class

elliottbiondo:
This MR adds the `RevolvedPolygon` class, analogous to `G4GeneralPolycone`, in which an arbitrary polygon is revolved around the `z` axis. This is done using a Convex Differences Tree [1], as was done to handle `G4ExtrudedSolid`.

The MR does not provide an `or_solid` method, nor handle azimuthal slices, both of which will be added to the next MR, which will hook this new capability up to `SolidConverter`.

[1] S. B. Tor and A. E. Middleditch, "Convex Decomposition of Simple Polygons," ACM Transactions on Graphics," volume 3(4), 224-265, 1984, https://doi.org/10.1145/357346.357348.


elliottbiondo:
@sethrj @mrguilima this is ready for review

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
+// b------------------------------- -*- C++ -*-
+// -------------------------------//
```
sethrj: I think a cat sat on your keyboard

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
...
+{
+    CELER_VALIDATE(polygon_.size() >= 3,
+                   << "polygon must have at least 3 vertices");
+
+    // All points must be positive
+    CELER_VALIDATE(
+        std::all_of(polygon_.begin(),
+                    polygon_.end(),
+                    [](Real2 const& p) { return p[R] >= 0 && p[Z] >= 0; }),
+        << "polygon must consist of only positive r and z values");
```
sethrj: I don't think Z needs to be positive

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
...
+    // Subtract concave regions from the convex hull
+    return vb.insert_region(Label{label_, level_ext + ".d"},
+                            Joined{op_and, {result, sub_node}});
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Revolved a convex polygon around the \em z axis.
+ *
+ * The polygon be strictly counterclockwise, i.e., it must not contain any
```
sethrj:
```suggestion
 * The polygon must be strictly counterclockwise, i.e., it must not contain any
```

Unless ye be a pirate

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Make a translated cylinder node.
+ */
+NodeId RevolvedPolygon::make_cylinder(detail::VolumeBuilder& vb,
+                                      Real2 const& p0,
+                                      Real2 const& p1,
+                                      SubIndex const& si) const
+{
+    real_type hh = std::abs(p1[Z] - p0[Z]) / 2;
```
sethrj:
```suggestion
    real_type hh = std::fabs(p1[Z] - p0[Z]) / 2;
```

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
...
+ * Make a translated cone node.
+ */
+NodeId RevolvedPolygon::make_cone(detail::VolumeBuilder& vb,
+                                  Real2 const& p0,
+                                  Real2 const& p1,
+                                  SubIndex const& si) const
+{
+    auto [r_bot, r_top] = (p0[Z] < p1[Z]) ? std::pair{p0[R], p1[R]}
+                                          : std::pair{p1[R], p0[R]};
+    auto [z_bot, z_top] = std::minmax(p0[Z], p1[Z]);
```
sethrj: This might be a bit less error-prone to pass in the Real2 by value as `p_bot` and `p_top`, then `if (p_bot[Z] > p_top[Z]) { std::swap(p_bot, p_top); }`

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -0,0 +1,273 @@
...
+                      "order");
+
+    SoftEqual<real_type> soft_equal(vb.tol().rel, vb.tol().abs);
+    SoftZero<real_type> soft_zero(vb.tol().abs);
+
+    auto n = polygon.size();
+    std::vector<NodeId> outer_nodes;
+    std::vector<NodeId> inner_nodes;
+
+    auto next_idx = [n](size_type i) { return (i + 1 < n) ? i + 1 : 0; };
```
sethrj: I think the intent matches the semantics of modulo (wrapping around) and improves readability:
```suggestion
    auto next_idx = [n](size_type i) { return i % n; };
```
or since this is only used in one place, you can put it inline there.

test/orange/orangeinp/RevolvedPolygon.test.cc
```diff
@@ -0,0 +1,285 @@
...
+                                                      "rp@0.0.0.pz",
+                                                      "",
+                                                      "rp@0.0.0.cz",
+                                                      "",
+                                                      "rp@0.0.0,rp@0.0.ou"};
+
+    auto const& u = this->unit();
+    EXPECT_VEC_EQ(expected_surface_strings, surface_strings(u));
+    EXPECT_VEC_EQ(expected_volume_strings, volume_strings(u));
+    EXPECT_VEC_EQ(expected_md_strings, md_strings(u));
```
sethrj, line 91: I think it would be good to also test the constructed bounding boxes.

test/orange/orangeinp/RevolvedPolygon.test.cc
```diff
@@ -0,0 +1,285 @@
...
+        = {"Plane: z=0",
+           "Plane: z=2",
+           "Cone z: t=0.5 at {0,0,0}",
+           "Cyl z: r=3",
+           "Plane: z=1.5",
+           "Cone z: t=0.8 at {0,0,0}",
+           "Cone z: t=0.4 at {0,0,4.5}"};
+    static char const* const expected_volume_strings[]
+        = {"all(+0, -1, -3, !all(+0, -1, -2), !all(!all(+0, -1, -2), "
+           "any(all(+0, -4, -5), all(-1, +4, -6))))"};
```
sethrj: Sometimes you need to manually check for weird string wrapping before committing...
```suggestion
           "Cone z: t=0.4 at {0,0,4.5}",};
    static char const* const expected_volume_strings[]
        = {R"(all(+0, -1, -3, !all(+0, -1, -2), !all(!all(+0, -1, -2), any(all(+0, -4, -5), all(-1, +4, -6)))))",};
```

test/orange/orangeinp/RevolvedPolygon.test.cc
```diff
@@ -0,0 +1,285 @@
...
+                     {3, 1},
+                     {3, 2},
+                     {2, 2},
+                     {2, 1},
+                     {1, 1},
+                     {1, 3},
+                     {0.33, 3},
+                     {0.33, 0.5}};
+
+    this->build_volume(RevolvedPolygon{"rp", std::move(polygon)});
```
sethrj: It would also be good to test a few points to make sure the constructed logic is correct (besides eyeballing it). Include `"orange/orangeinp/detail/SenseEvaluator.hh"` and do something like:
```suggestion
    auto vol_id = this->build_volume(RevolvedPolygon{"rp", std::move(polygon)});
    auto eval_sense = [&u = this->unit(), vol_id](Real3 const& point) {
        CELER_ASSERT(vol_id < u.tree.volumes().size());
        auto n = u.tree.volumes()[vol_id.get()];
        detail::SenseEvaluator eval_sense(u.tree, u.surfaces, pos);
        return eval_sense(n);
    };
```
(you could make this a class method to reuse across the different cases).

sethrj (approved):
Awesome! Just a couple minor comments.

src/orange/orangeinp/RevolvedPolygon.cc
elliottbiondo: Agreed, but I want to keep the API the same as make_cylinder, and also avoid explaining in the doc string that `p_bot` and `p_top` can be supplied in any order, which I think is an implementation detail. Here is what I came up with:

```
NodeId RevolvedPolygon::make_cone(detail::VolumeBuilder& vb,
                                  Real2 const& p0,
                                  Real2 const& p1,
                                  SubIndex const& si) const
{
    Real2 p_bot = p0;
    Real2 p_top = p1;

    if (p_bot[Z] > p_top[Z])
    {
        std::swap(p_bot, p_top);
    }
 ...
 ```

File: src/orange/orangeinp/RevolvedPolygon.cc (hunk 6 above)
Comment (elliottbiondo) [reply src/orange/orangeinp/RevolvedPolygon.cc]: I only did it this way on the basis of comments on a previous MR. Switched to modulus.

File: src/orange/orangeinp/RevolvedPolygon.cc (hunk 5 above)
Comment (sethrj) [reply src/orange/orangeinp/RevolvedPolygon.cc]: Looks good to me!

File: test/orange/orangeinp/RevolvedPolygon.test.cc (hunk 8 above)
Comment (elliottbiondo) [reply test/orange/orangeinp/RevolvedPolygon.test.cc]: `pre-commit` reformats what you have. I unwrapped the two offending strings in this file, added trailing `,`, and then let pre-commit handle it. Anything different would require `clang-format off`.

**Comment by elliottbiondo** [comment]
@sethrj ready for another look

File: src/orange/orangeinp/RevolvedPolygon.hh [hunk 10]
```diff
@@ -0,0 +1,161 @@
...
+   i |                      |                             |
+   s |_____________         |_____________                |_____________
+        r axis ->
+   \endverbatim
+ * In this example, level 1 region 0 is formed from only two subregions, but
+ * the general case is handed via:
+ *
+ * region = union(outer subregions) - union(inner subregions).
+ *
+ * \internal When labeing nodes in the CSG output, the following shorthand
```
mrguilima, line 63:
```suggestion
 * \internal When labeling nodes in the CSG output, the following shorthand
```

test/orange/orangeinp/RevolvedPolygon.test.cc
sethrj: Yep, that's all I was suggesting!

sethrj (approved):
Awesome!

src/orange/orangeinp/RevolvedPolygon.hh
```diff
@@ -0,0 +1,161 @@
...
+   z |       \|             |________| subregion 0        |_______\ subregion 1
+     |                 =    |          (a cylinder)  -    |         (a cone)
+   a |                      |                             |
+   x |                      |                             |
+   i |                      |                             |
+   s |_____________         |_____________                |_____________
+        r axis ->
+   \endverbatim
+ * In this example, level 1 region 0 is formed from only two subregions, but
+ * the general case is handed via:
```
mrguilima, line 59:
```suggestion
 * the general case is handled via:
```

src/orange/orangeinp/RevolvedPolygon.hh
```diff
@@ -0,0 +1,161 @@
...
+                     Real2 const& p1,
+                     SubIndex const& si) const;
+
+    // Make a label extension for a level
+    std::string make_level_ext(SubIndex si) const;
+
+    // Make a label extension for a region within a level
+    std::string make_region_ext(SubIndex si) const;
+
+    // Make a extension label for a subregion within a region
```
mrguilima, line 150:
```suggestion
    // Make an extension label for a subregion within a region
```

mrguilima (commented):
This PR has been merged, but I found a few typos.
Also, all tests contain only positive z for polygon points, it would be nice to have a couple tests with negative z points.

Sorry for the late review.

sethrj:
@mrguilima Could you submit a little PR with these fixes and a new test? Thanks!

elliottbiondo:
Fixing typos in #1892

PR #1886, opened 2025-07-30
Provide canonical volume and instance IDs from ORANGE

sethrj:
Closes #1749 by mapping during construction the volume params' IDs to ORANGE input. Two extra arrays indexed by ImplVolumeId ; they're constructed via the ORANGE input, where the "label" is now a choice between `Label` *or* `VolumeInstanceId`. The g4org conversion propagates the volume instances upward through construction, and the volume IDs are extracted from those.

PR #1889, opened 2025-08-01
Add mapping from ImplVolumeId to VolumeId

sethrj:
Currently most of our runtime code uses ImplVolumeId to map from the geometry state to attributes like detector volumes, materials, etc. To set up the materials correctly for #1886 we'll store `ImportVolume` properties corresponding to `VolumeParams` volumes rather than Geant4 implementation (or geometry implementation) volumes. Up to this point we've been relying on labels to construct those.

This adds a `volume_id(ImplVolumeId) -> VolumeId` method for GeoParamsInterface that maps from the implementation ID to the canonical ID. It also adds support for the GeoMaterialParams to be constructed from a map of ImplVolumeId (basically the original way of doing it), a map of labels (which ORANGE currently relies on), and a map of VolumeId (which #1886 requires).

### DIFF
--- a/src/celeritas/setup/Problem.cc
+++ b/src/celeritas/setup/Problem.cc
@@ -435,7 +435,7 @@ ProblemLoaded problem(inp::Problem const& p, ImportData const& imported)
 
     // Create geometry/material coupling
     params.geomaterial = GeoMaterialParams::from_import(
-        imported, params.geometry, params.material);
+        imported, params.geometry, params.volume, params.material);
 
     // Construct particle params
     params.particle = ParticleParams::from_import(imported);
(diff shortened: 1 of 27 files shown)


src/celeritas/setup/Problem.cc
```diff
@@ -434,8 +434,17 @@ ProblemLoaded problem(inp::Problem const& p, ImportData const& imported)
     params.material = MaterialParams::from_import(imported);
 
     // Create geometry/material coupling
-    params.geomaterial = GeoMaterialParams::from_import(
-        imported, params.geometry, params.material);
+    params.geomaterial = [&] {
+        GeoMaterialParams::SPConstVolume volume_params;
+        if constexpr (CELERITAS_CORE_GEO != CELERITAS_CORE_GEO_ORANGE)
+        {
+            // FIXME: ORANGE doesn't support volume/material conversion
```
hhollenb, line 441: Is this because ORANGE is getting the ImplVolumeId -> VolumeId mapping in #1886 ?

hhollenb (approved):
Nice changes, looks like a good addition!

src/celeritas/setup/Problem.cc
sethrj: Indeed. Almost got that working...

PR #1892, opened 2025-08-04
Add full support for G4GenericPolycone

elliottbiondo:
This MR does two things:
1) add `EnclosedAzi` support to `RevolvedPolygon`,
2) hook up `RevolvedPolygon` to `SolidConverter` to add full support for `G4GenericPolycone`.

One thing I didn't implement here is an `or_solid`-like method, that only makes cylinders or cones. I don't think the actual surface logic changes significantly in these cases. An `or_polycone` could be added for cases where generic polygons are not required, but detecting this case would take some work.

### DIFF
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -57,6 +57,7 @@
 #include "orange/orangeinp/CsgObject.hh"
 #include "orange/orangeinp/IntersectRegion.hh"
 #include "orange/orangeinp/PolySolid.hh"
+#include "orange/orangeinp/RevolvedPolygon.hh"
 #include "orange/orangeinp/Shape.hh"
 #include "orange/orangeinp/Solid.hh"
 #include "orange/orangeinp/StackedExtrudedPolygon.hh"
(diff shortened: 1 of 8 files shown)


elliottbiondo:
@sethrj @mrguilima RTR

src/orange/g4org/SolidConverter.cc
```diff
@@ -551,8 +552,22 @@ auto SolidConverter::extrudedsolid(arg_type solid_base) -> result_type
...
+    // Get the polygon. Although Geant4 prefers clockwise order upon input,
+    // GetCorner actually returns points in counterclockwise order, as used
+    // by ORANGE.
+    size_type num_points = solid.GetNumRZCorner();
+    std::vector<Real2> polygon;
+    polygon.reserve(num_points);
+    for (auto i : range(num_points))
+    {
+        auto point = solid.GetCorner(i);
+        polygon.push_back(Real2{scale_(point.r), scale_(point.z)});
```
sethrj: This is fine as is, but for your enrichment there's also a shortcut:
```suggestion
        polygon.push_back(scale_.to<Real2>(point.r, point.z));
```

src/orange/orangeinp/RevolvedPolygon.cc
```diff
@@ -62,7 +66,23 @@ NodeId RevolvedPolygon::build(VolumeBuilder& vb) const
...
+    auto result = this->make_levels(vb, filtered_polygon, ri);
+
+    if (enclosed_)
+    {
+        // Handle azimuthal truncation
+        auto&& [sense, wedge] = enclosed_.make_sense_region();
+        NodeId wedge_id = build_intersect_region(vb, label_, "angle", wedge);
+        if (sense == Sense::outside)
+        {
+            wedge_id = vb.insert_region(Label{label_, "negated_angle"},
```
sethrj: To be consisted with Solid (and actually could you update PolySolidBase L97-98 too?)
```suggestion
        char const* ext = (sense == Sense::outside ? "~azi" : "azi");
        NodeId wedge_id
            = build_intersect_region(vb, label_, ext, wedge);
        if (sense == Sense::outside)
        {
            wedge_id = vb.insert_region({}, Negated{wedge_id});
```

sethrj (approved):
Very nice!

elliottbiondo:
@sethrj RFAL

PR #1895, opened 2025-08-06
Refactor geo testing to use persistent Geant4 geometry

sethrj:
This horrific refactor is necessary to test core geometry with volume mapping (to be added right after this). There are two small other changes too:
- `demangle_type` now properly works when passing a polymorphic class rather than printing type information from the base class (oops)
- Geant4 error messages during setup are now sent to the world logger, not the self logger


amandalund (approved):
Thanks @sethrj!

PR #1896, opened 2025-08-07
Add smear and Gaussian surface roughness calculators

hhollenb:
Added simple implementations and tests for the smear and Gaussian surface roughness calculators. These calculators take a their corresponding parameter (roughness or sigma alpha), the global surface normal, and the incident track direction, and return a sampled facet normal.

I haven't optimized the performance of the implementations because I don't want to accidentally modify the resulting probability distributions. It would be interesting to have some type of validation framework to take calculators, run them over a large number of samples, and show the resulting distributions in some validation program. Would make finding discrepancies between Geant4, benchmarking different implementations, and regressions easier if that were automated somehow.


src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
```diff
@@ -0,0 +1,96 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Sample a facet normal via the Gaussian roughness model.
+ */
+template<class Engine>
+CELER_FUNCTION Real3 GaussianRoughnessCalculator::operator()(Engine& rng) const
+{
+    NormalDistribution<real_type> sample_alpha(0, sigma_alpha_);
+    real_type f_max = min(real_type{1}, 4 * sigma_alpha_);
```
sethrj: Make these class member data instead of local.

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * The Gaussian roughness model is parameterized by a positive real number \c
+ * sigma_alpha. The angle \c alpha between the facet normal and the global
+ * normal is sampled from the distribution
+ * \f[
+ * p(\alpha) = N(\alpha; 0, \sigma_\alpha) * \sin(\alpha)
+ * \f]
+ * where alpha is in the range [0, pi/2).
+ */
+class GaussianRoughnessCalculator
```
sethrj: Make sure to add these classes to `doc/implementation/optical-physics.rst`.

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
```diff
@@ -0,0 +1,96 @@
...
+    do
+    {
+        real_type cos_alpha, sin_alpha;
+        do
+        {
+            cos_alpha = cos(sample_alpha(rng));
+            sin_alpha = sqrt(1 - ipow<2>(sin_alpha));
+        } while (cos_alpha <= 0 || RejectionSampler{sin_alpha, f_max}(rng));
+
+        local_normal = ExitingDirectionSampler{cos_alpha, normal_}(rng);
```
sethrj: A quick comment here reminding people of what this sampler does (rotates the exiting direction relative to the surface normal, and samples azimuthally)

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
```diff
@@ -0,0 +1,96 @@
...
+
+    Real3 local_normal;
+    do
+    {
+        real_type cos_alpha, sin_alpha;
+        do
+        {
+            cos_alpha = cos(sample_alpha(rng));
+            sin_alpha = sqrt(1 - ipow<2>(sin_alpha));
+        } while (cos_alpha <= 0 || RejectionSampler{sin_alpha, f_max}(rng));
```
sethrj: Is this too early of an optimization? Since the gaussian and cosine are symmetric... 
```suggestion
            cos_alpha = fabs(cos(sample_alpha(rng)));
            sin_alpha = sqrt(1 - ipow<2>(sin_alpha));
        } while (RejectionSampler{sin_alpha, f_max}(rng));
```

 Also what is this rejection criterion? Is it just trying to keep the sampled angle from being too close to grazing?

src/celeritas/optical/surface/calc/SmearRoughnessCalculator.hh
```diff
@@ -0,0 +1,95 @@
...
+        0, real_type(2 * constants::pi));
+    UniformRealDistribution<real_type> sample_cos_theta(-1, 1);
+    UniformRealDistribution<real_type> sample_r(0, 1);
+
+    Real3 local_normal;
+    do
+    {
+        local_normal = normal_;
+        axpy(cbrt(sample_r(rng)) * roughness_,
+             from_spherical(sample_cos_theta(rng), sample_phi(rng)),
```
sethrj: This results in unspecified behavior, because the ordering of the function calls in here isn't specified by the language (i.e. one or the other arguments could be evaluated first), and since those both change RNG, this can lead to different compilers sampling different results from the same stream! 😨  Instead of this, use `geocel/random/IsotropicDistribution.hh`.

src/celeritas/optical/surface/calc/SmearRoughnessCalculator.hh
```diff
@@ -0,0 +1,95 @@
...
+    UniformRealDistribution<real_type> sample_phi(
+        0, real_type(2 * constants::pi));
+    UniformRealDistribution<real_type> sample_cos_theta(-1, 1);
+    UniformRealDistribution<real_type> sample_r(0, 1);
+
+    Real3 local_normal;
+    do
+    {
+        local_normal = normal_;
+        axpy(cbrt(sample_r(rng)) * roughness_,
```
sethrj: I think it's equivalent and more readable to have `PowerDistribution sample_r_{2}` (even though I think it's slightly less efficient).

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
```diff
@@ -0,0 +1,96 @@
...
+
+        local_normal = ExitingDirectionSampler{cos_alpha, normal_}(rng);
+    } while (!is_entering_surface(local_normal, dir_));
```
sethrj: Won't all the roughness calculations going to be predicated on this? Maybe this rejection loop and the use of the incident direction should be outside the roughness sampler itself.

test/celeritas/optical/RoughnessCalculator.test.cc
```diff
@@ -0,0 +1,128 @@
...
+  private:
+    RandomEngine rng_;
+};
+
+//---------------------------------------------------------------------------//
+// Test smear roughness model distribution
+TEST_F(RoughnessCalculatorTest, smear)
+{
+    real_type roughness = 0.3;
+    Real3 normal = make_unit_vector(Real3{0, 0, 1});
```
sethrj: It might be good to use an unusual surface normal.

test/celeritas/optical/RoughnessCalculator.test.cc
```diff
@@ -0,0 +1,128 @@
...
+            real_type normal_cos = dot_product(normal, local_normal);
+            real_type momentum_cos = dot_product(momentum, local_normal);
+
+            EXPECT_TRUE(is_soft_unit_vector(local_normal));
+            EXPECT_TRUE(momentum_cos < 0);
+
+            moments[0] += normal_cos;
+            moments[1] += ipow<2>(normal_cos);
+            moments[2] += momentum_cos;
+            moments[3] += ipow<2>(momentum_cos);
```
sethrj: Is this the best way of measuring the sampled normal since the incident normal and momentum are basically hard-coded? (With the values chosen, the `x` component isn't even contributing to the sample result.) I think a better test might be to bin the dot product with the surface normal (and you can put the rejection loop in here). Have you seen the `HistogramSampler` helper class? That could help you easily test a couple of different roughness values of the smearing and gaussian to really observe the effects..

sethrj (commented):
Looking good;  I have a few thoughts. I can volunteer to refactor the test (or help copilot do it) to show you what that would look like.

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
hhollenb: This keeps alpha in the range (-pi/2, pi/2), which can be written as a conditional probability. The probability is then

$$
P(|x| < \alpha) = \frac{\text{erf}(\alpha / \sigma_\alpha)}{\text{erf}((\pi/2)/\sigma_\alpha)}
$$

but if you take the absolute value of the cosine, then the PDF looks something like

$$
p(\alpha) = \sum_{n=-\infty}^{\infty} \mathcal{N}(\alpha + 2 \pi n; 0, \sigma_\alpha)
$$

This is partially why I think a validation tool would be useful because these distributions get subtle quickly and I'm not the best at working with them.

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
hhollenb: This rejection is only because the distributions might generate non-sensical normals. You can likely write better implementations that don't generate non-sensical normals and this would be just a `CELER_ENSURE` check. I could see this loop being broken out but then you'd have to be careful wherever you use them to correct them for being anti-aligned with the momentum.

src/celeritas/optical/surface/calc/SmearRoughnessCalculator.hh
hhollenb: Ah I was looking for that distribution but missed that it was in geocel

sethrj: Yeah, it requires Real3 which is only in geocel and higher... I'm ok with that since it's a geometry-related distribution.

src/celeritas/optical/surface/calc/GaussianRoughnessCalculator.hh
sethrj: I guess the question is whether we want the roughness calculator to *only* sample the microsurface normal, or to sample a normal and ensure consistency with the sense. If it's *just* the surface structure, the direction shouldn't come into play at all (aside from the polarity/sign/sense for the global surface normal).

test/celeritas/optical/RoughnessCalculator.test.cc
hhollenb: I just picked these two vectors since they're (hard coded as) a linearly independent 2D subspace and the local normal is 2D since it's normalized. The `HistogramSampler` looks great, but I might try extending it to something like `HistogramSampler2d`? I'll probably also remove the sampling against momentum here and just check the distributions about the global normal.

sethrj: I think that might be a step too far. Are there any correlated azimuthal+polar effects that we need to worry about, or can we take the dot product with the global normal and bin the cosine?

sethrj:
@hhollenb May I help speed this along a bit by updating the calculator test? I also take back my suggestion about the do/while loop; I think it's now a bit more awkward.

sethrj:
@hhollenb Since the sampled normals no longer depend on the direction, I went ahead and simplified the tested values, and expanded the range of allowable sigma_alpha and roughness. I've also moved the calculators up one directory to avoid it being too deep, and finally I've changed the direction rejection to be a wrapper class that first checks that the global and direction are set up for "entering" (otherwise we risk an infinite loop if there's an error in the surface sign).

sethrj:
@hhollenb also for large sigma_alpha and a larger sample count, I hit errors in the rejection sampling. Does the fmax calculation assume that we're never going to sample past 4 sigma?

hhollenb:
Awesome, that looks great thanks!

For the fmax calculation, I was a bit confused by that in the Geant code. I had a vague idea of what it was trying to do last week but going back to my notes it doesn't make sense. I think you run into an issue that the distribution is very sharply peaked and awkward to sample so they automatically accept any angle > 4 sigma_alpha away?

sethrj:
I think what we should do is just limit the sampled surface angle to `[-pi/2, pi/2]` (rejecting greater than vertical), chopping off the tail of the gaussian, because anything outside that is probably unphysical (the photon would have to be "underneath" an outcropping of the surface, trying to get out). The approximation of the surface normals as gaussian breaks down before that point, clearly... the paper that introduced the unified model had a sigma of ~12 degrees = 0.07 rad , so it's unlikely we'll ever see anything close to that.

sethrj:
Oh wait, that's what the cos_alpha < 0 rejection does anyway. 🙃

hhollenb:
Yeah that's what the cos_alpha < 0 rejection does. 

As far as I can tell, the Gaussian term wants to sample angles close to 0, but the sin(alpha) term suppresses them but accepts angles closer to pi/2. I believe the logic is that if the Gaussian samples a large angle (unlikely) then it's very likely to be accepted by the sin(alpha) rejection so just accept and move on. For smaller angles, do the rejection sampling as normal. For small angles sin f_max ~ f_max.

Although at that point you're basically sampling the a linear distribution sin x ~ x so there's probably a more efficient transform you can do?

sethrj:
Looking at [levin-morephysical-1996](https://doi.org/10.1109/NSSMIC.1996.591410), I'm confused about where the sin(alpha) even comes from? The gaussian distribution is just the angle relative to the global normal, so can't we just take the cosine of it and move on with our lives?
fig 5

Is the Geant4 `sin(alpha)` distribution some other model? Or am I missing something subtle like a $`\sin \theta`$ popping out from a polar integration...

EDIT: yep I sure am. The extra `sin(θ)` isn’t a different model—it’s the Jacobian from sampling on the sphere. The Levin “σₐ” is defined from a 1D Gaussian slope distribution, but when extended isotropically to 3D, a `sin(θ)` is needed for proper normalization in spherical coordinates.

sethrj:
OK @hhollenb , please take another look at my changes... to avoid the `f <= fmax` error we basically have to sample from a truncated gaussian.

hhollenb:
@sethrj the Jacobian term makes sense. I found a paper on viewing the sin(alpha) term as a correction for the projected surface area of the facet: https://arxiv.org/abs/2503.11567
I'll go ahead and take another look at the most recent changes.

hhollenb:
Ok looks pretty clean, I doubt it'll get too much simpler unless we start changing the actual distribution. I think it's good for a baseline implementation if we want to iterate on it in the future once we start doing validation tests for surface physics.

sethrj (approved):

PR #1897, opened 2025-08-08
Fix oversimplification of simple quadrics

sethrj:
A recent SCALE bug report (see [internal issue](https://code-int.ornl.gov/rnsd/scale/-/issues/1988)) demonstrated that small ellipsoids are being incorrectly oversimplified. This is not as I thought a bug in the SurfaceSimplifier but actually an issue with how the quadrics are normalized. The ellipsoid implementation resulted in second-order terms scaling with $`r^2`$ rather than being first-order as is the case with transformed cylinders etc.; this resulted in the terms being incorrectly dropped. Further exploration showed that such scaling also negatively affects the quadratic solver, which drops second order terms when they're below a certain absolute magnitude.

With this change, the ellipsoid quadric equation is normalized by the product of the min and max radii so that it will produce the same equation as a sphere when the radii are equal. The elliptical cylinder needed a similar correction.

### DIFF
--- a/src/orange/surf/SimpleQuadric.hh
+++ b/src/orange/surf/SimpleQuadric.hh
@@ -40,7 +40,7 @@ class Sphere;
  * cylinders, etc.  The quadric is ill-defined if all non-constants are zero.
  *
  * Even though this equation can be scaled arbitrarily, the quadratic solver
- * is sensitive to the scale of the coefficients to avoid numerical precisoin
+ * is sensitive to the scale of the coefficients to avoid numerical precision
  * loss. If present, the first-order coefficients should be on the length scale
  * of the problem (i.e., translations avoid catastrophic precision loss).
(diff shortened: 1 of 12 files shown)


src/orange/surf/SimpleQuadric.hh
```diff
@@ -37,7 +37,15 @@ class Sphere;
   \f]
  *
  * This can represent axis-aligned hyperboloids, ellipsoids, elliptical
- * cylinders, etc.
+ * cylinders, etc.  The quadric is ill-defined if all non-constants are zero.
+ *
+ * Even though this equation can be scaled arbitrarily, the quadratic solver
+ * is sensitive to the scale of the coefficients to avoid numerical precisoin
```
elliottbiondo: precision

elliottbiondo (approved):

PR #1899, opened 2025-08-11
Fix normalization and edge cases on twisted GenPrism faces

sethrj:
Related to #1897, #1232, and #1337, this addresses a long-standing issue with the scaling of the "twisted" quadric surface (hyperbolic paraboloid) created by the GenPrism: congruent surfaces would have different scaling and thus not be deduplicated.

This fixes that scaling by normalizing the linear component to unity, so that as the twist angle goes to zero, the surface approaches a correctly normalized plane. It also rigorously determines when to approximate the surface as a plane based on the tolerance, and it gives a more accurate plane when the two normals aren't exactly equal, by averaging the normals at two opposite corners of the face.

(I also had to fix the string simplifier test helper code; that can be moved to another unit test if needed.)


src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -630,7 +641,8 @@ void ExtrudedPolygon::build(IntersectSurfaceBuilder& insert_surface) const
         auto p2 = scaling_factors_[top] * Real3{p_a[X], p_a[Y], 0}
                   + line_segment_[top];
 
-        insert_surface(Sense::inside, Plane{p0, p1, p2});
+        insert_surface(Sense::inside,
+                       Plane{detail::normal_from_triangle(p0, p1, p2), p0});
```
sethrj, line 645: The complex constructor for Plane shouldn't have been inlined; and the underlying code was also used in the GenPrism face construction.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -657,25 +669,25 @@ void ExtrudedPolygon::output(JsonPimpl* j) const
...
 {
-    CELER_EXPECT(dir == X || dir == Y);
+    CELER_EXPECT(dim == X || dim == Y);
 
     constexpr auto bot = Bound::lo;
     constexpr auto top = Bound::hi;
 
     // Find extrema of unextruded polygon
-    auto [poly_min, poly_max] = detail::find_extrema(polygon, dir);
+    auto [poly_min, poly_max] = detail::find_extrema(make_span(polygon), dim);
```
sethrj, line 681: I generalized this usage and changed the dimension variable to be consistent with the other uses.

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -767,9 +779,9 @@ GenPrism GenPrism::from_trap(
  * Construct from half Z height and 1-4 vertices for top and bottom planes.
  */
 GenPrism::GenPrism(real_type halfz, VecReal2 const& lo, VecReal2 const& hi)
-    : hz_{halfz}, lo_{lo}, hi_{hi}
+    : hh_{halfz}, lo_{lo}, hi_{hi}
```
sethrj, line 782: This was changed for consistency with the other regions.

elliottbiondo (approved):

PR #1901, opened 2025-08-12
Use canonical volumes for VecGeom

sethrj:
As part of #1815, following on to #1886 (ORANGE), this implements canonical volumes and volume instances for VecGeom.

- Remove `volume_instances` from GeoParmasInterface; these will be replaced by VolumeParams.
- Add `impl_volume_instances` to vecgeom params
- Prefix ORANGE/VG data members with `impl_` as approporiate
- Add ImplX -> X mapping collections in vecgeom params
- Move volume stack test into geotests and enable for vecgeom

> [!WARNING]
> Problems with replica volumes (e.g. Makoto's B5) will temporarily be wrong because we map all replicas to the same volume instance ID. I will fix this in the next pull request.


src/geocel/vg/VecgeomData.hh
```diff
@@ -20,36 +20,77 @@
...
+        return host_world != nullptr && max_depth > 0;
+    }
+};
+
 //---------------------------------------------------------------------------//
 /*!
  * Persistent data used by VecGeom implementation.
+ *
+ * The volumes and volume instances are only available if generating the
+ * geometry directly from Geant4.
```
hhollenb: In the future will we plan on being able to build volume and volume instances without Geant4? E.g. running optical physics with just VecGeom.

src/geocel/vg/VecgeomParams.hh
```diff
@@ -197,16 +217,17 @@ auto VecgeomParams::volume_instances() const -> VolInstanceMap const&
  */
 inline VolumeId VecgeomParams::volume_id(ImplVolumeId iv_id) const
 {
-    CELER_EXPECT(volume_id_map_.empty() || iv_id < volume_id_map_.size());
+    auto const& vol_ids = this->host_ref().volumes;
+    CELER_EXPECT(vol_ids.empty() || iv_id < vol_ids.size());
 
-    if (CELER_UNLIKELY(volume_id_map_.empty()))
+    if (CELER_UNLIKELY(vol_ids.empty()))
```
hhollenb: It seems inconsistent for this to be returning a different value than `VecgeomTrackView::volume_id()` for the same `ImplVolumeId`. If VGDML loads the geometry should it also initialize `vol_ids`?

src/geocel/vg/VecgeomTrackView.hh
```diff
@@ -271,92 +278,103 @@ VecgeomTrackView::operator=(Initializer_t const& init)
...
+ * Get the physical volume ID in the current cell.
  */
-CELER_FUNCTION ImplSurfaceId VecgeomTrackView::next_impl_surface_id() const
+CELER_FUNCTION VolumeInstanceId VecgeomTrackView::volume_instance_id() const
 {
-#ifdef VECGEOM_USE_SURF
-    if (!this->is_on_boundary())
+    CELER_EXPECT(!this->is_outside());
+    ImplVolumeInstanceId ipv_id{this->physical_volume().id()};
+    if (CELER_UNLIKELY(params_.volume_instances.empty()))
```
hhollenb: Same concern here as with `volume_id`, should VGDML initialize a default for `volume_instances`?

src/geocel/vg/VecgeomTrackView.hh
```diff
@@ -271,92 +278,103 @@ VecgeomTrackView::operator=(Initializer_t const& init)
...
-    return id_cast<ImplVolumeId>(this->volume().id());
+    return id_cast<LevelId>(vgstate_.GetLevel());
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Get the canonical volume ID in the current implementation volume.
+ * Get the volume instance ID for all levels.
+ *
+ * This is only useful for Geant4 geometry, so shouldn't be used if the mapping
```
hhollenb: I think this code is a bit funky. It's Geant4 specific but is available for multiple geometry implementations, so I have to be cautious about configurations and the state of VecgeomParams if it has been built with `volumes` and `volume_instances`. If it's only useful for Geant4 geometry shouldn't it be only available when there's a Geant4 geometry and mapping?

hhollenb (commented):
Good work! It's been a minute since I've looked at the volume params code so some of our objectives are a bit fuzzy to me still, so I could use a refresher at some point.

src/geocel/vg/VecgeomData.hh
sethrj: As I see it, we should be able to implement `make_model_inp` in VecGeomParams when loading directly from GDML rather than Geant4: it would return a volume mapping similar to the Geant4 one. However, this capability isn't needed at the moment: we are exclusively building with Geant4 (and for "ultralite" testing, we don't have vecgeom enabled). I'll update the comment to be clearer though.

src/geocel/vg/VecgeomParams.hh
sethrj: Looks like this was no longer necessary! If not loading with an input model (e.g. Geant4) that has canonical volume information, this accessor should not be used.

src/geocel/vg/VecgeomTrackView.hh
sethrj: If we build a "canonical" model different from the vecgeom internal mapping, then yes we should build it. Otherwise, as a bit of a hack to support unit testing, we leave the field empty and return "implementation" volume IDs.

src/geocel/vg/VecgeomTrackView.hh
sethrj: This function is only used for mapping between VecGeom and a "canonical" geometry (in real use case, the Geant4 navigation history reconstruction). I don't think there's a good way of switching between TrackView implementations; and again, I don't think we should prioritize VGDML construction. This code will also change soon.

sethrj:
@hhollenb I'm working on an update that'll eliminate the weird volume ID casting and VGDML exceptions. Should be done soon.

src/geocel/vg/VecgeomData.hh
sethrj: To clean up the VGDML cases and optimize for Geant4 cases, I now have VecGeom generating a model if running standalone.

src/geocel/vg/VecgeomTrackView.hh
sethrj: Fixed!

src/geocel/vg/VecgeomGeoTraits.hh
```diff
@@ -41,6 +41,8 @@ struct GeoTraits<VecgeomParams>
 
     //! Surfaces are available when using frame implementation
     static constexpr bool has_impl_surface{CELERITAS_VECGEOM_SURFACE};
+    //! VecGeom has "physical volumes"
```
mrguilima:
```suggestion
    //! VecGeom has "placed (physical) volumes"
```

src/geocel/vg/VecgeomParams.cc
```diff
@@ -111,6 +112,8 @@ int vecgeom_verbosity()
 //---------------------------------------------------------------------------//
 /*!
  * Get a reproducible vector of LV instance ID -> label from the given world.
+ *
+ * This create the "implementation" volume map.
```
mrguilima:
```suggestion
 * This creates the "implementation" volume map.
```

src/geocel/vg/VecgeomParams.cc
```diff
@@ -558,9 +575,47 @@ VecgeomParams::~VecgeomParams()
...
+        if (label.name.empty())
+        {
+            continue;
+        }
+
+        auto const& placed_vol = get_placed_volume(geo, ivi_id);
+
+        v.volume_instances[ivi_id.get()].label = label;
+        // Map the corresponding logical volume ID
+        v.volume_instances[ivi_id.get()].volume
```
mrguilima, line 601: I thought this should store the placed vol ID, but the log vol ID is being stored. Is this correct?

mrguilima (approved):
LGTM.
Thanks @sethrj !

src/geocel/vg/VecgeomParams.cc
sethrj: Here we're just associating the (volume index = logical volume ID) with the (volume instance definition = physical volume). So it's both!

PR #1903, opened 2025-08-14
Uniquely and consistently identify and map volume instances

### LINKED ISSUE #1815: Refactor volume management to improve mapping across the codebase

sethrj:
This chonky refactor completes the transition from implementation volume IDs to actual volume IDs, fixing some longstanding issues with ORANGE volume mapping (a detector could only map to a single ImplVolumeId instead of multiple implementation IDs that represent a single logical volume) and label mapping. It replaces all the label- and implementation-based volume mapping with VolumeId mapping. (It's possible that trying to run a real problem without Geant4 may no longer work, but that is no longer a use case we're considering aside from basic unit testing.)

- VolumeId and ImplVolumeId are now separate opaque IDs! (See #1839, #1884, #1889)
- VolumeInstanceIds now encode the copy number of a replica volume! This eliminates a lot of weirdness when reconstructing navigation hierarchy. (See https://jira-geant4.kek.jp/browse/UR-100?focusedId=20220&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20220 .)
- Some helper classes such as the `GeantVolumeMapper` have been removed. Functions that used to need the geometry now (under the hood) use the global geant4 geometry and volumes.
- A new `VolumeIdBuilder` class constructs Volume IDs from user-provided labels or G4LV*.
- A new `build_volume_collection` constructs `Collection<..., ImplVolumeId>` using VolumeId-based data.

Follow-up work:
- Construct SDParams in core and pass to optical params; currently these are unused.
- Replace use cases of `find_geant_volumes` with `VolumeIdBuilder`
- Add volume hierarchy reconstruction to ORANGE


doc/implementation/geometry.rst
```diff
@@ -64,24 +67,28 @@ fields as "cells") and "surfaces" defined by the relationships between volumes.
...
-   *implementation* volumes.
+   instances but is treated identically.
+   It has a specific shape, material, metadata, and associated scoring/sensitive region.
+   Each volume is simply a *node* in the detector geometry graph.
+   This definition differs slightly from Geant4 and VecGeom, where the
+   ``G4LogicalVolume`` and ``vecgeom::UnplacedVolume`` classes directly
+   reference the child geometry nodes and thus implicitly include the objects
+   embedded in a volume.
+   Celeritas refers to user-defined volumes as *canonical* to differentiate
+   them from *implementation* volumes.
```
pcanal: Are ``G4LogicalVolume`` and ``vecgeom::UnplacedVolume`` considered `implementation volumes`?

doc/implementation/geometry.rst
```diff
@@ -64,24 +67,28 @@ fields as "cells") and "surfaces" defined by the relationships between volumes.
...
-   thus implicitly include the objects embedded in a volume. Celeritas refers
-   to user-defined volumes as *canonical* to differentiate them from
-   *implementation* volumes.
+   instances but is treated identically.
+   It has a specific shape, material, metadata, and associated scoring/sensitive region.
+   Each volume is simply a *node* in the detector geometry graph.
+   This definition differs slightly from Geant4 and VecGeom, where the
+   ``G4LogicalVolume`` and ``vecgeom::UnplacedVolume`` classes directly
+   reference the child geometry nodes and thus implicitly include the objects
+   embedded in a volume.
```
pcanal: > classes directly reference the child geometry nodes and thus implicitly include the objects
> embedded in a volume.

This seems to conversely implies that the relationship between the `Celeritas Volumes` is external to the `node`, is that the intent?

doc/implementation/geometry/geant4.rst
```diff
@@ -7,7 +7,7 @@ Geant4 geometry
 Celeritas defines mappings to the underlying Geant4 geometry objects for
 integration in the rest of the code base. There is limited support for
 executing the main simulation engine using Geant4 navigation states, but this
```
pcanal: why `but`? Is it in opposition to 'there is support' or to 'limited' or something else?  And/or is it actually mean something like "to allow implementation of that, this class can be/is used ..."?

doc/implementation/geometry.rst
```diff
@@ -64,24 +67,28 @@ fields as "cells") and "surfaces" defined by the relationships between volumes.
...
+   Celeritas refers to user-defined volumes as *canonical* to differentiate
```
pcanal: Does this mean that the heading of this paragraph (line 68) is essentially `Canonical volume` or are the canonical volumes a subset of volumes (the sentence on line 77 seems to associate the `user-defined` quality as `canonical` ... but then again which volumes are not user-defined?)

pcanal:
>  Functions that used to need the geometry now (under the hood) use the global geant4 geometry and volumes.

What happens if the user has 'broken' code that trigger/use does functions before the `Geant4` geometry is setup?  Or is it already impossible to get in this situation by construction?

src/geocel/GeantGeoParams.cc
```diff
@@ -228,6 +180,15 @@ void append_border_surfaces(GeantGeoParams const& geo,
...
             continue;
         }
+        if (key.first->IsReplicated() || key.second->IsReplicated())
+        {
+            CELER_LOG(error) << "G4 border surface '" << surf->GetName()
+                             << "' uses replica/parameterised volumes: these "
+                                "will be ignored!";
+            continue;
+        }
+        // TODO: if these are replica volumes, warn the user? or loop over all
```
pcanal: How is that different from the above condition (line 183)?

src/geocel/GeantGeoParams.cc
```diff
@@ -309,70 +246,96 @@ std::vector<inp::Volume> make_inp_volumes(GeantGeoParams const& geo)
...
             continue;
         }
 
-        auto* g4lv = geo.id_to_geant(vol_id);
-        CELER_ASSERT(g4lv);
-        result[vol_id.get()] = inp_from_geant(geo, label, *g4lv);
+        auto vol_id = geo.volume_id(iv_id);
+        auto& g4lv = *geo.id_to_geant(vol_id);
+
+        // TODO: do we want to strip pointers?
```
pcanal: Why? :)

doc/implementation/geometry.rst
sethrj: Yes. Both Geant4 and VecGeom create "extra" volumes that aren't reachable directly through volume graph (via reflection and/or boolean operations).

doc/implementation/geometry.rst
sethrj: I am also confused upon re-reading this sentence... but I *think* I meant that if there's a G4 logical volume `A` that has solid `S_A`, and `A` has daughters `B` and `C` with their corresponding solids: our definition is that `A` is the region `S_A - (S_B | S_C)` (i.e. the daughters subtracted out), and it doesn't include the full region enclosed by `S_A`.

src/geocel/GeantGeoParams.cc
```diff
@@ -309,70 +246,96 @@ std::vector<inp::Volume> make_inp_volumes(GeantGeoParams const& geo)
...
+            CELER_ASSERT(g4pv);
+            for (int j = 0, jmax = g4pv->GetMultiplicity(); j < jmax; ++j)
+            {
+                if (g4pv->IsReplicated())
+                {
+                    g4pv->SetCopyNo(j);
+                }
+                auto vol_inst_id = geo.geant_to_id(*g4pv);
+                vol_inp.children.push_back(vol_inst_id);
+            }
```
pcanal, line 288: Is the intent to have at the end of the loop `g4pv`' copy number to `jmax-1` or should it be reset or does it not matter?

doc/implementation/geometry/geant4.rst
sethrj: I meant that even though the use of navigation as a "track view" is limited, the metadata/object translation is quite complete. I'll clarify.

doc/implementation/geometry.rst
sethrj: I will also update this for clarity. I meant that the "volumes" in the header can be referred to as "canonical volumes" for disambiguation. The "user-defined" is meant to differentiate from "implementation" (which includes temporaries for booleans/reflection) but that's imprecise.

src/geocel/GeantGeoParams.cc
sethrj: Yep I must've written the comment before adding the warning; I'll remove that. If (god forbid) a user tries to define a replica surface with an optical boundary then we can add that capability.

src/geocel/GeantGeoParams.cc
sethrj: If the GDML is loaded without the pointer stripping option (which I don't think is done in practice) then this can result in nonreproducible volume names. I don't think we need to worry about this so I'll delete the comment.

src/geocel/GeantGeoParams.cc
sethrj: It would only matter if the GeantGeoParams construction were done on a worker thread while a thread-local navigator is inside a replica volume (impossible with how Celeritas is used). In fact in for older Geant4 (before 11.2 I think) the copy number is never initialized and can result in undefined behavior by uninitialized memory access.

src/geocel/GeantGeoParams.cc
```diff
@@ -799,26 +675,27 @@ void GeantGeoParams::build_metadata()
...
     {
         CELER_LOG(debug) << "Building after volume stores were cleared: "
                          << "lv_offset=" << this->lv_offset()
-                         << ", pv_offset=" << this->pv_offset()
                          << ", mat_offset=" << this->mat_offset();
     }
 
+    // Construct volume instance mapper
+    vi_mapper_ = detail::GeantVolumeInstanceMapper(*this->world());
+    data_.vi_mapper = &vi_mapper_;
```
pcanal, line 688: Is there any case where we need `vi_mapper != data_.vi_mapper` (i.e. do we really need both)?

sethrj: The `data_.vi_mapper` is (unfortunately) necessary to sneakily propagate the pointer into the GeantGeoTrackView and hackishly access it from the `VecgeomParams` setup...

pcanal: Would it then be possible to 'only' keep `data_.vi_mapper` (and drop `vi_mapper_)?

sethrj: I was hesitant to put a shared_ptr in the params data, but since the state data has unique pointers I guess it couldn't hurt!

src/geocel/VolumeCollectionBuilder.hh
```diff
@@ -0,0 +1,94 @@
...
+    auto num_impl_volumes = geo.impl_volumes().size();
+    resize(&result, num_impl_volumes);
+    for (auto iv_id : range(ImplVolumeId{num_impl_volumes}))
+    {
+        T value{};
+        if (auto vol_id = geo.volume_id(iv_id))
+        {
+            value = fill_value(vol_id);
+        }
+        result[iv_id] = value;
```
pcanal: Is `T` guaranteed to be trivial/fast to copy?  Otherwise consider:
```suggestion
        if (auto vol_id = geo.volume_id(iv_id))
        {
            result[iv_id] = fill_value(vol_id);
        } else {
	        result[iv_id] = T{};
	    }
```
or if `Collection` has `emplace`:
```suggestion
        if (auto vol_id = geo.volume_id(iv_id))
        {
            result.emplace(std::make_pair(iv_id, fill_value(vol_id)));
        } else {
	        result.emplace(std::make_pair(iv_id, T{}));
	    }
```

src/geocel/VolumeIdBuilder.hh
```diff
@@ -0,0 +1,76 @@
...
+    //! \name Type aliases
+    using G4LV = G4LogicalVolume;
+    //!@}
+
+  public:
+    // Construct using "global" values (NOT PREFERRED)
+    VolumeIdBuilder();
+
+    // Construct using geant4 params and/or volume params
+    inline VolumeIdBuilder(VolumeParams const*, GeantGeoParams const*);
```
pcanal, line 45: Is this constructor used (it does not seem to be directly used but might be indirectly used)?

src/geocel/detail/GeantVolumeInstanceMapper.cc
```diff
@@ -0,0 +1,244 @@
...
+ */
+GeantVolumeInstanceMapper::GeantVolumeInstanceMapper(G4PV const& world)
+{
+    // Track volumes that aren't normal/replica/param
+    std::vector<std::string> invalid_volumes;
+
+    // Visit (depth-first) volume instances inside the world
+    VolumeVisitor visit_vol{GeantVolumeAccessor{}};
+    visit_vol(
+        [&](G4VPhysicalVolume const* pv, int) {
```
pcanal:
```suggestion
        [&invalid_volumes](G4VPhysicalVolume const* pv, int) {
```
makes it easier to distinguish internal and external variable for the lambda.

src/geocel/vg/VecgeomParams.cc
```diff
@@ -89,6 +89,29 @@ namespace
...
+        CELER_EXPECT(parent);
+        auto result = parent->GetLogicalVolume();
+        CELER_ENSURE(result);
+        return result;
+    }
+
+    //! Outgoing instance nodes from a volume
+    ContainerVolInstRef children(VolumeRef parent) final
+    {
+        auto&& daughters = parent->GetDaughters();
```
pcanal, line 111:
```suggestion
        CELER_EXPECT(parent);
        auto&& daughters = parent->GetDaughters();
```
for symmetry with the above routine?

src/geocel/vg/VecgeomParams.cc
```diff
@@ -118,28 +141,30 @@ std::vector<Label> make_logical_vol_labels(vecgeom::VPlacedVolume const& world)
...
-                name.erase(name.begin() + pos, name.end());
-            }
 
-            // Add to name map
-            names[name].push_back(lv);
-        },
-        &world);
+    VolumeVisitor visit_vol{VecgeomVolumeAccessor{}};
+    visit_vol(make_visit_volume_once<vecgeom::LogicalVolume const*>(
+                  [&](VolT const* lv) {
```
pcanal:
```suggestion
                  [&names](VolT const* lv) {
```

src/geocel/VolumeCollectionBuilder.hh
sethrj: Yes, since our collections have to be allocated uninitialized and filled with `memcpy`, `T` here must be trivially copyable and constructable.

src/geocel/VolumeIdBuilder.hh
sethrj: I intended the other constructor to delegate to this one, good catch!

src/geocel/detail/GeantVolumeInstanceMapper.cc
sethrj: Also requires capturing `this`

src/geocel/GeantGeoParams.cc
sethrj: Oh yeah, the `vi_mapper` is stored in the class by value, not shared pointer.

src/geocel/VolumeCollectionBuilder.hh
sethrj: I've rearranged it to improve clarity and eliminate the unnecessary assignment.

doc/implementation/geometry.rst
```diff
@@ -16,14 +16,15 @@ instantiation is associated with a transformation and
 other metadata. In HEP geometries, child nodes may not overlap each other or
 their enclosing parent volume.
 
-.. table:: Celeritas nomenclature tends toward computer science terminology.
+.. table:: Approximately equivalent geometry structure terminology for
+   Celeritas, Geant4, VecGeom, and SCALE.
```
amandalund, line 20: SCALE has the concept of interface and boundary surfaces?

doc/implementation/geometry.rst
```diff
@@ -35,21 +36,21 @@ their enclosing parent volume.
...
    substantially from Geant4 geometry definitions. In KENO-VI :cite:`kenovi`
    geometry, parent units mask (rather than strictly contain) child units.
 
-.. [#ob] :ref:`api_orange_objects` are used strictly for construction in ORANGE and are not
-   identifiable during runtime.
+.. [#ob] :ref:`api_orange_objects` are used strictly for ORANGE construction in
+   ORANGE and are not identifiable during runtime.
+
+.. [#lv] The ``G4LogicalVolume`` class is equivalent to a volume except that
+   points inside the children are not considered to be part of the volume.
```
amandalund: Does a VecGeom logical volume behave like `G4LogicalVolume`?

src/celeritas/setup/Problem.cc
```diff
@@ -213,7 +213,8 @@ auto build_track_init(inp::Control const& c, size_type num_streams)
         {
             input.track_order = TrackOrder::none;
         }
-        CELER_LOG(debug) << "Set default track order " << input.track_order;
+        CELER_LOG(debug) << "Set default track order "
+                         << to_cstring(input.track_order);
```
amandalund, line 218: Why not use the `operator<<` for enum here and above?

amandalund (commented):
Didn't quite make it through tonight, but what I've seen looks great!

src/celeritas/setup/Problem.cc
sethrj: It turns out the `EnumStringMapper` was "leaking" a generic `operator<<`... so removing that IWYU header forced this change. We should probably remove that helper function...

doc/implementation/geometry.rst
sethrj: Copy-paste error, thanks!

src/geocel/VolumeCollectionBuilder.hh
```diff
@@ -0,0 +1,101 @@
...
+        {
+            return fill_value(vol_id);
+        }
+        return {};
+    };
+
+    // Create a collection sized for each implementation volume
+    Collection<T, Ownership::value, MemSpace::host, ImplVolumeId> result;
+    auto num_impl_volumes = geo.impl_volumes().size();
+    resize(&result, num_impl_volumes);
```
pcanal, line 53: Side note unrelated to the PR: can you remind me why `resize` take a pointer rather than a reference? (Since at least some of the resize have a `CELER_EXPECT(data);` the reason does not appears to be to support potential `nullptr` values).

Sorta never mind, I finally remembered we discussed it and went to look it up, checked https://celeritas-project.github.io/celeritas/user/development/style.html and got my answer (paraphrasing): `passing mutable value by pointer used to be recommended by the Google style but they now have updated the recommendation and passing non-optional (mutable or not) values should be done by reference`

src/geocel/detail/GeantVolumeInstanceMapper.hh
```diff
@@ -0,0 +1,100 @@
...
+    explicit GeantVolumeInstanceMapper(G4PV const& world);
+
+    //! Number of volume instances
+    size_type size() const { return g4pv_.size(); }
+
+    // Get the volume instance using the pv and its copy number if applicable
+    VolumeInstanceId geant_to_id(G4PV const&) const;
+
+    // Get the volume instance without altering state
+    VolumeInstanceId geant_to_id(G4PV const&, int copy_no) const;
```
amandalund: Descriptions were a bit confusing as they seemed flipped, but the definitions clarified it.

src/geocel/GeantGeoParams.cc
pcanal: Fair enough, that's a bit too bad but since `data_` cannot/should-not` own the mapper, I guess this the best we can do ...

pcanal (approved):
LGETM.

amandalund (approved):
Awesome, thanks for all the work @sethrj!

sethrj:
Thanks a bunch @pcanal and @amandalund ! I'm sure it was as unpleasant to review as it was to refactor 🤪

PR #1904, opened 2025-08-14
Allow constant values for R index

Rashika-Gupta:
This PR updates the optical material validation logic to allow refractive index (RINDEX) values that are constant or monotonically increasing, instead of strictly increasing.

- Updated `MaterialParams` validation to use `is_monotonic_nondecreasing` for `RINDEX` values.
- Adjusted Cherenkov test input data to use constant RINDEX for low energy and then increasing.
- Updated affected expected test values in `Cherenkov.test.cc` to match new output.


sethrj (approved):
Nice fix, thanks!

sethrj:
@Rashika-Gupta Looking further into your test, I'm not 100% sure that the constant R index you added is being tested in the troublesome spot: I expect if there are any errors, it'll occur if an electron slows down over a step so that it's superluminal at the beginning and subluminal at the end. The test has only 500 keV (1/beta ~ 1.58) , 10 GeV (1/beta ~ 1), and subluminal (1/beta ~ 1.81).

In dfa5fbbe1fe76f856870ad868575737bcec5613b I made the test extensible by supporting a different optical material during construction. Can you add a test like:
```

class CherenkovAirTest : public CherenkovTest
{
  public:
    ImportOpticalProperty build_import_property() const final
    {
        ImportOpticalProperty prop;
        for (double wl : {1.29, 0.82})
        {
            prop.refractive_index.x.push_back(um_to_mev(wl));
        }
        prop.refractive_index.y = {1.3, 1.3};
        return prop;
    }
};
```
and add a dn/dx calculator test for 1/beta = {1.2, 1.3, 1.4}?

src/celeritas/optical/MaterialParams.cc
```diff
@@ -120,8 +120,8 @@ MaterialParams::MaterialParams(Input const& inp)
         CELER_VALIDATE(is_monotonic_increasing(make_span(ri.x)),
                        << "refractive index energy grid values are not "
                           "monotonically increasing");
-        CELER_VALIDATE(is_monotonic_increasing(make_span(ri.y)),
-                       << "refractive index values are not monotonically "
+        CELER_VALIDATE(is_monotonic_nondecreasing(make_span(ri.y)),
```
amandalund, line 123: Our Cherenkov dn/dx calculator uses an inverse grid (flips the energy and rindex values) to find the energy at a given rindex, relying on both being monotonic increasing. Won't this change mean we could create a grid with constant x values and increasing y values?

test/celeritas/optical/Cherenkov.test.cc
```diff
@@ -81,12 +81,12 @@ Span<real_type const> get_wavelength()
...
-Span<real_type const> get_refractive_index()
+Span<real_type const> get_water_refractive_index()
 {
     static real_type const refractive_index[]
-        = {1.3235601610672, 1.3236962786529, 1.3238469492274, 1.3239820826015,
-           1.3241317601229, 1.3242660923031, 1.3244149850321, 1.3245487081924,
-           1.3246970353146, 1.3248303521764, 1.3249783454392, 1.3251114708334,
+        = {1.3235601610672, 1.3235601610672, 1.3235601610672, 1.3235601610672,
+           1.3235601610672, 1.3235601610672, 1.3235601610672, 1.3235601610672,
+           1.3235601610672, 1.3235601610672, 1.3235601610672, 1.3235601610672,
```
amandalund: Now that we have the additional test with a material with a constant rindex I think we can revert these changes; otherwise we should add a note in the comment above that these were modified from the values obtained from that source.

amandalund (commented):
> I expect if there are any errors, it'll occur if an electron slows down over a step so that it's superluminal at the beginning and subluminal at the end. The test has only 500 keV (1/beta ~ 1.58) , 10 GeV (1/beta ~ 1), and subluminal (1/beta ~ 1.81).

The 500 keV generator test should test this case (pre-step energy = 500 keV, 1/beta ~1.16; post-step energy = 150 keV, 1/beta ~ 1.58), but I agree about adding a separate test for a constant refractive index material.

src/celeritas/optical/MaterialParams.cc
sethrj: That was my concern. But maybe it's ok to have coincident points that result in a discontinuity? Would be worth checking in a unit test of the inverse grid I think...

test/celeritas/optical/Cherenkov.test.cc
Rashika-Gupta: I was thinking more of a case when refractive index in constant for some energy range and then starts to increase as a function of energy.

src/celeritas/optical/MaterialParams.cc
Rashika-Gupta: Do you think that test should be included as part of this PR, or would you prefer it to be in a follow-up?

amandalund: One issue I see is our `Interpolator` currently expects the left grid point to be strictly less than the right, and can result in nan values if not.

test/celeritas/optical/Cherenkov.test.cc
amandalund: I think it would be good to add that to the `CherenkovAirTest`, and also at least one value before the constant rindex values: then I think we'll hit some assertion failures in the dn/dx calculation.

Rashika-Gupta: I did try as suggested - starting with one refractive index value, then keeping it constant, and then increasing it. This did not trigger an assertion failure. However, I did see the assertion error yesterday when I used a decreasing refractive index function - 
```
CherenkovTest.generator
unknown file: Failure
C++ exception with description "src/corecel/random/distribution/RejectionSampler.hh:88:
celeritas: precondition failed: fmax_ >= f_"

```

src/celeritas/optical/MaterialParams.cc
sethrj: But we can't eve be "inside" a grid cell with zero width can we? I think a test as part of this PR would be great.

amandalund: I don't think we have a special case in the grid calculator for landing exactly on a grid point and skipping interpolation, we only do that if we're on one of the grid endpoints (or out of bounds).

test/celeritas/optical/Cherenkov.test.cc
amandalund: Was the first refractive index value less than the constant values? If I change the test grid data to e.g. `wl = {1.3, 1.29, 0.82, 0.8}` and `ri = {1.0, 1.3, 1.3, 1.4}` I hit:
```console
celeritas/src/corecel/grid/Interpolator.hh:112: celeritas: precondition failed: left[X] < right[X]
```

Rashika-Gupta: Another set that i tried for is : `ri = {1.1, 1.2, 1.31, 1.31, 1.320, 1.330}` and it didn't throw any error. Perhaps, the change in the refractive index values matter too.

amandalund: Yeah, if we're not testing one of the constant refractive index values (in that case beta = 1 / 1.31) then we won't see any failures.

src/celeritas/optical/MaterialParams.cc
sethrj: We don't, but I think "nan" might be a reasonable answer for the question "give me the value of f(x) at a discontinuity `x_d`"... but since we define the whole grid on `[x_min, x_max)` (closed lower) then maybe we should define `f(x_d)` as the value on the higher point?

test/celeritas/optical/Cherenkov.test.cc
sethrj: This should be fixed with #1906.

sethrj:
@Rashika-Gupta This discussion has probably sounded like a lot of hand-wringing about an astronomically unlikely edge case (stopping at exactly the discontinuity), especially if the "worst case" is that we get a NaN. But it's been useful because (1) with enough particles and enough users we'll eventually hit it (suppose somebody has a tracking cut that leaves particles at 1/beta) or if we try out single precision, and (2) outside of the cherenkov case it forced us to think about and document the use case of representing discontinuous functions, which often show up in user scoring and nuclear data.

Now if we represent a total cross section as a sum of an elastic (nonzero at all energies) and inelastic (nonzero only after a threshold energy) we can get the correct and consistent behavior by taking the union of their grids and adding another coincident point at the start of the threshold energy.

src/celeritas/optical/MaterialParams.cc
sethrj: This should be fixed with https://github.com/celeritas-project/celeritas/pull/1906.

sethrj (approved):
Now this should work 😊

test/celeritas/optical/Cherenkov.test.cc
```diff
@@ -483,6 +498,54 @@ TEST_F(CherenkovTest, generator)
...
+                     1e-6);
+
+    optical::MaterialView mat_view{material->host_ref(), material_id};
+    CherenkovDndxCalculator calc_dndx(
+        mat_view,
+        params->host_ref(),
+        this->particle_params()->get(ParticleId{0}).charge());
+
+    std::vector<real_type> betas{
+        real_type{1.0 / 1.2}, real_type{1.0 / 1.3}, real_type{1.0 / 1.4}};
```
amandalund: We should test one of the constant values, it should work now:
```suggestion
        real_type{1.0 / 1.2}, real_type{1.0 / 1.3}, real_type{1.0 / 1.31}, real_type{1.0 / 1.4}};
```

amandalund (approved):
Thanks @Rashika-Gupta!

test/celeritas/optical/Cherenkov.test.cc
Rashika-Gupta: Updated. Thank you 😊

PR #1906, opened 2025-08-17
Support discontinuities in nonuniform grid calculator

sethrj:
This should address the issues with inverting a non-monotonic-increasing grid in https://github.com/celeritas-project/celeritas/pull/1904. Since we define the grid on an interval `[a, b)` it follows that each grid "cell" should be defined on `[a, b)` as well, implying that the value at the discontinuity should be closed on the right. 

discontinuity

a82ac785de6378a822c3d1445f68e9f0d6166830 accomplishes this by using `upper_bound` instead of `lower_bound`, and 3556047eaa335389e784b73578a06b0a0bf3f23b reduces the search space required for the grid search, with the nice side effect of (if debug assertions are disabled) falling back to making out-of-bounds grid point searches appear at the end grid points.

This change also fixes an edge case in FindInterp: see https://github.com/celeritas-project/celeritas/pull/1906/commits/ca744cf6fb51ab2a8a79b9f9b1b8e38ec6da6adf  ; and since *that* is used in our RZ map field, cyl map field, etc., we should now be able to represent discontinuities in those.

### DIFF
--- a/src/celeritas/grid/NonuniformGridInserter.hh
+++ b/src/celeritas/grid/NonuniformGridInserter.hh
@@ -24,7 +24,6 @@ namespace celeritas
 /*!
  * Construct a nonuniform grid and add it to the specified grid collection.
  *
- * \todo Use dedupe collection builder
  * \todo Validate input doesn't contain more than two coincident x grid points?
  */
 template<class Index>
--- a/src/celeritas/grid/NonuniformGridInserter.hh
+++ b/src/celeritas/grid/NonuniformGridInserter.hh
@@ -23,6 +23,9 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 /*!
  * Construct a nonuniform grid and add it to the specified grid collection.
+ *
+ * \todo Use dedupe collection builder
+ * \todo Validate input doesn't contain more than two coincident x grid points?
  */
 template<class Index>
 class NonuniformGridInserter
--- a/test/celeritas/grid/NonuniformGridCalculator.test.cc
+++ b/test/celeritas/grid/NonuniformGridCalculator.test.cc
@@ -98,17 +98,9 @@ TEST_F(NonuniformGridCalculatorTest, discontinuous)
     EXPECT_EQ(2.0, calc[3]);
 
     // Test on grid points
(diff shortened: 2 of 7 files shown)


src/celeritas/grid/NonuniformGridInserter.hh
```diff
@@ -23,6 +23,9 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 /*!
  * Construct a nonuniform grid and add it to the specified grid collection.
+ *
+ * \todo Use dedupe collection builder
```
amandalund: We do already use the dedupe collection builder in the `NonuniformGridBuilder` (used by this class).

test/celeritas/grid/NonuniformGridCalculator.test.cc
```diff
@@ -84,6 +84,51 @@ TEST_F(NonuniformGridCalculatorTest, nonmonotonic)
...
+    // Test out-of-bounds
+    EXPECT_SOFT_EQ(1.0, calc(0.0));
+    EXPECT_SOFT_EQ(2.0, calc(4.0));
+}
+
+TEST_F(NonuniformGridCalculatorTest, discontinuous_end)
+{
+    inp::Grid grid;
+    grid.x = {1.0, 2.0, 2.0};
+    grid.y = {1.0, 1.0, 2.0};
```
amandalund, line 114: Is a grid defined on only two points at a discontinuity allowed? E.g.:
```c++
grid.x = {2.0, 2.0};
grid.y = {1.0, 2.0};
```
Thinking about #1904, this could happen with a constant refractive index. Currently `calc(2)` on this grid would return the lower rather than upper value.

amandalund (commented):
Seems like reasonable behavior, thanks!

test/celeritas/grid/NonuniformGridCalculator.test.cc
sethrj: Added a test and fixed this for consistency.

amandalund (approved):

Rashika-Gupta:
Thanks for clarifying ! 
Makes sense 😊

PR #1907, opened 2025-08-18
Add uniform JSON IO to OpaqueId

sethrj:
This writes OpaqueIDs as integers or "null" values to JSON. It also makes the nomenclature more consistent about the ID being "null", adjusts the template parameter names to prepare for a `std::optional`-like interface, and implements `id_cast` to reduce the number of template instantiations.

### DIFF
--- a/src/corecel/OpaqueId.hh
+++ b/src/corecel/OpaqueId.hh
@@ -21,7 +21,7 @@ namespace detail
 //---------------------------------------------------------------------------//
 //! Sentinel value for an unassigned opaque ID
 template<class T>
-inline constexpr T invalid_opaqueid_value{static_cast<T>(-1)};
+inline constexpr T null_opaqueid_value{static_cast<T>(-1)};
 
 //---------------------------------------------------------------------------//
 //! Safely cast from one integer T to another U, avoiding the sentinel value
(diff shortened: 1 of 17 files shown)


src/corecel/OpaqueId.hh
```diff
@@ -16,11 +16,50 @@
...
 {
+namespace detail
+{
+//---------------------------------------------------------------------------//
+//! Sentinel value for an unassigned opaque ID
+template<class T>
+inline constexpr T nullid_value{static_cast<T>(-1)};
+
+//---------------------------------------------------------------------------//
+//! Safely cast from one integer T to another U, avoiding the sentinel value
```
pcanal, line 27: Should this function also check for overflow?

pcanal (approved):
LGTM

amandalund (approved):

src/corecel/OpaqueId.hh
sethrj: It does :)
```c++
        // Check that the cast value is within the integer range [0, N-1)
        using C = std::common_type_t<T, std::make_unsigned_t<U>>;
        if constexpr (std::is_signed_v<C>)
        {
            CELER_EXPECT(static_cast<C>(value) >= 0);
        }
        CELER_EXPECT(static_cast<C>(value) < static_cast<C>(nullid_value<T>));
```
using `<` rather than `!=` protects against overflow for both kinds of ints, and the `static_cast... >= 0` checks against an signed->unsigned overflow.

pcanal: Right since `sizeof(C) >= sizeof(T)`.

PR #1908, opened 2025-08-18
Define throw helper functions to reduce code bloat and improve performance

sethrj:
This uses an old trick from [Draco](https://github.com/lanl/Draco/blob/28ab37f48196b4f2eb2de475f86303e7969b1d10/src/dsxx/Assert.cc#L56)/Nemesis that defines a `toss_cookies` helper function to throw an exception, reducing the amount of duplicate exception handling code across the code base. With the advent of the `[[noreturn]]` annotation it's now possible to inform the compiler that the data with function call can be moved to a "cold" code section, improving efficiency. 

Measuring the optical collector "host generate" test (the most expensive "unit" test in our code) shows a 1% performance boost for debug+assert build, and about a 2% improvement for a reldeb (RelWithDebInfo plus assertions) build. The code size for a debug build is reduced by 2% as well.

### DIFF
--- a/src/corecel/Assert.hh
+++ b/src/corecel/Assert.hh
@@ -142,7 +142,7 @@
 #if !defined(__HIP__) && !defined(__CUDA_ARCH__)
 // Throw in host code
 #    define CELER_DEBUG_THROW_(MSG, WHICH) \
-        throw ::celeritas::DebugError(     \
+        ::celeritas::throw_debug_error(    \
             {::celeritas::DebugErrorType::WHICH, MSG, __FILE__, __LINE__})
 #elif defined(__CUDA_ARCH__) && !defined(NDEBUG)
 // Use the assert macro for CUDA when supported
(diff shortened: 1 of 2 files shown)


src/corecel/Assert.hh
```diff
@@ -142,7 +142,7 @@
 #if !defined(__HIP__) && !defined(__CUDA_ARCH__)
 // Throw in host code
 #    define CELER_DEBUG_THROW_(MSG, WHICH) \
-        throw ::celeritas::DebugError(     \
+        ::celeritas::throw_debug_error(    \
```
pcanal, line 145: The minor loss here is the ability (not used) of being able to (eventually) add the function name in which the exception is thrown.  If that function name was captured, it would be helpful in case the exception is caught and either re-thrown or printed by some experiment's framework.  (and yes, this is minor since the file name and line number is somewhat equivalent even if it is less expressive as it does not express the template parameters).

pcanal (approved):
LGTM (Can't argue against the gains)

src/corecel/Assert.hh
sethrj: Wouldn't that be trivially easy to add? You could just add another `char const* func_name;` to `DebugErrorDetails` which would be expanded inside the macro just like `__FILE__`.

pcanal: Indeed.

PR #1909, opened 2025-08-18
Make surface model its own hierarchy and add default surface

sethrj:
This addresses the last physics meeting's discussion  https://github.com/celeritas-project/celeritas/issues/1512#issuecomment-3160606607 : surface models are no longer actions, and we now support a "default" surface model.


src/celeritas/phys/SurfacePhysicsMapView.hh
```diff
@@ -16,27 +17,36 @@ namespace celeritas
 /*!
  * Access surface physics mappings for a particular surface.
  *
- * This simply encapsulates the \c SurfaceParamsData class.
+ * This simply encapsulates the \c SurfaceParamsData class. The "default"
+ * surface
```
amandalund: 🐿️

amandalund (approved):
Looks good to me!

PR #1910, opened 2025-08-19
Enable muon physics and add user-defined list of offloaded particles 

stognini:
This PR enables muon physics by default and adds the new `SetupOptions::offload_particles` which allows specifying a subset of particles to be offloaded to Celeritas via the Tracking Manager Integration.

This new feature allows for easier individual physics validations in Geant4-Celeritas applications.


src/accel/TrackingManagerIntegration.cc
```diff
@@ -111,14 +111,10 @@ void verify_tracking_managers(Span<G4PD const* const> expected,
...
-                      "active in Celeritas: missing"
-                   << join(missing.begin(), missing.end(), ", ", printable_pd));
-    CELER_VALIDATE(all_attached_correctly,
-                   << "tracking manager(s) are not attached correctly "
-                      "(maybe add TrackingManagerConstructor to your physics "
-                      "list?)");
+    CELER_LOG(warning)
+        << "not all particles from TrackingManagerConstructor are active in "
+           "Celeritas: missing "
+        << join(missing.begin(), missing.end(), ", ", printable_pd);
```
sethrj: Are you sure these changes are correct? Celeritas can't do anything with track types that it doesn't know about. Also the tracking managers should definitely be attached right? I don't understand this change at all.

src/accel/TrackingManagerIntegration.cc
```diff
@@ -51,7 +51,7 @@ void verify_tracking_managers(Span<G4PD const* const> expected,
     bool all_attached_correctly{true};
     auto log_tm_failure = [&all_attached_correctly](G4PD const* p) {
         all_attached_correctly = false;
-        auto msg = CELER_LOG(error);
+        auto msg = CELER_LOG(warning);
```
sethrj: This also seems wrong; how can you offload a particle to celeritas if the tracking manager isn't attached when using tracking manager integration?

src/accel/TrackingManagerConstructor.cc
```diff
@@ -98,14 +162,16 @@ void TrackingManagerConstructor::ConstructProcess()
...
+    for (auto* p : offload_particles_)
     {
         CELER_EXPECT(p);
         // Memory for the tracking manager should be freed in
         // G4VUserPhysicsList::TerminateWorker from G4WorkerRunManager
         // by constructing a 'set' of all tracking managers.
         // (Note that it is leaked in Geant4 11.0 and 11.1 for MT mode.)
         p->SetTrackingManager(manager ? manager.release() : manager_ptr);
+        CELER_LOG(info) << "Added particle \"" << p->GetParticleName()
+                        << "\" to Tracking Manager ";
```
sethrj: I think we want one log message, not one per particle:
```c++
}
CELER_LOG(info) << "Built Celeritas tracking managers for "
    << join(offload_particles_.begin(), offload_particles_.end(), ", ",
               [](G4ParticleDefinition const* pd)
               {
                   return p->GetParticleName();
               });
```

src/accel/TrackingManagerConstructor.cc
```diff
@@ -70,6 +118,22 @@ TrackingManagerConstructor::TrackingManagerConstructor(
...
+/*!
+ * Construct particles before \c ::ConstructProcess since any
+ * \c G4ParticleDefinition other than ions and shortlived must be created
+ * during Geant4's \c Pre_Init state.
+ */
+void TrackingManagerConstructor::ConstructParticle()
+{
+    auto& singleton = detail::IntegrationSingleton::instance();
+    auto const& user_offload = singleton.setup_options().offload_particles;
+    // Construction of particles happens at offload_particles_ assignment
```
sethrj: What does this comment mean?

src/accel/TrackingManagerConstructor.cc
```diff
@@ -70,6 +118,22 @@ TrackingManagerConstructor::TrackingManagerConstructor(
     CELER_EXPECT(tmi == &TrackingManagerIntegration::Instance());
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Construct particles before \c ::ConstructProcess since any
+ * \c G4ParticleDefinition other than ions and shortlived must be created
+ * during Geant4's \c Pre_Init state.
```
sethrj, line 60: Is this because calling `Blah::Definition` "constructs" it and will error if it's called for the first time during `ConstructProcess`?

src/accel/SetupOptions.cc
```diff
@@ -255,8 +255,10 @@ inp::FrameworkInput to_inp(SetupOptions const& so)
     inp::FrameworkInput result;
     result.system = load_system(so);
     result.geant.ignore_processes = so.ignore_processes;
-    result.geant.data_selection.particles = GeantImportDataSelection::em_basic;
-    result.geant.data_selection.processes = GeantImportDataSelection::em_basic;
+    //! \todo: Expose DataSelection to SetupOptions or default to all
+    // This avoids missing a process/particle import from GeantPhysicsOptions
+    result.geant.data_selection.particles = GeantImportDataSelection::em;
+    result.geant.data_selection.processes = GeantImportDataSelection::em;
```
sethrj: I think we want to rework the `inp::GeantImport` so that the process and particle lists live there. Or does the construction ordering mean that the tracking manager setup happens before the celeritas setup? 🤪

src/accel/TrackingManagerConstructor.cc
stognini: Since `G4Particle::Definition()` is called when setting up the `offload_particles_` list, a singleton is instantiated for each of those particles in the G4 application.

src/accel/TrackingManagerConstructor.cc
stognini: Yep. If we do not initialize them during that call, and do it during `ConstructProcess` we get

```
G4ParticleDefinition (other than ions and shortlived) should be created in Pre_Init state - mu-

-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : PART101
      issued by : G4ParticleDefintion::G4ParticleDefinition()
G4ParticleDefinition should be created in PreInit state
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------
```

src/accel/TrackingManagerConstructor.cc
sethrj: Oh, ok the `G4Particle` particle singletons... there are way too many particles types and instances and such to keep track of 😉

stognini: I'll reword that comment, but thought was useful to have something there, because yeah, it's far from obvious that that call would technically construct the particles (as the `ConstructParticle()` method implies).

src/accel/TrackingManagerIntegration.cc
stognini: Oh you have a point. That was a trailing change from my initial tests before I had the user list. We should only have managers to the offloaded particles, so I have to update the `TrackingManagerIntegration::BeginOfRunAction`:

```cpp
verify_tracking_managers(
                make_span(TrackingManagerConstructor::OffloadParticles()), // <-- Use user-defined list here 
                make_span(singleton.shared_params().OffloadParticles()),
                singleton.shared_params(),
                singleton.local_transporter())
```

src/accel/TrackingManagerConstructor.cc
sethrj: Totally agree...

src/accel/TrackingManagerConstructor.cc
```diff
@@ -25,15 +28,88 @@ namespace celeritas
...
+
+    return {std::begin(default_particles), std::end(default_particles)};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the list of user-defined particles that will be offloaded.
+ *
+ * \warning This assumes that the \c TrackingManagerIntegration::SetOptions has
+ * been called.
```
sethrj: We should dedicate a section of the offloading part of the user manual to discuss the order of operations... maybe a little flow chart for the dependencies of "A must be called before B"? I think it should be set at this point.

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -200,6 +212,30 @@ struct GeantPhysicsOptions
     //! Optical physics options
     GeantOpticalPhysicsOptions optical{
         GeantOpticalPhysicsOptions::deactivated()};
+
+    //! Initialize with no physics
```
sethrj, line 218: I think we've been meaning to move most of that list into a separate "EM physics" struct... wonder if now is the right time to do it?

src/accel/TrackingManagerConstructor.cc
```diff
@@ -25,15 +28,95 @@ namespace celeritas
...
+                   << "List of particles defined in "
+                      "SetupOptions::offload_particles is larger than the "
+                      "list of supported particles in Celeritas, which is: "
+                   << join(supported.begin(),
+                           supported.end(),
+                           ", ",
+                           [](G4ParticleDefinition const* pd) {
+                               return pd->GetParticleName();
+                           }););
+
```
sethrj: I think this check is redundant.

src/accel/TrackingManagerConstructor.hh
```diff
@@ -50,11 +50,18 @@ class TrackingManagerConstructor final : public G4VPhysicsConstructor
     //!@{
     //! \name Type aliases
     using LocalTransporterFromThread = std::function<LocalTransporter*(int)>;
+    using VecG4PD = std::vector<G4ParticleDefinition*>;
```
sethrj: Since the ordering for these doesn't matter (I just used a `Span` originally because the data was easily stored as a static array) let's use
```suggestion
    using SetG4PD = std::unordered_set<G4ParticleDefinition*>;
```
and return const references to static SetG4PD. That should also simplify your testing for user-requested particles.

src/accel/TrackingManagerIntegration.cc
```diff
@@ -156,9 +158,16 @@ void TrackingManagerIntegration::BeginOfRunAction(G4Run const*)
         // Set tracking manager on workers when Celeritas is not fully disabled
         CELER_LOG(debug) << "Verifying tracking manager";
 
+        auto const& opts
+            = detail::IntegrationSingleton::instance().setup_options();
+        auto const offload_particles
+            = opts.offload_particles.empty()
+                  ? TrackingManagerConstructor::DefaultOffloadParticles()
+                  : TrackingManagerConstructor::UserOffloadParticles();
```
sethrj: This is pretty awkward since it reconstructs the offload particles again. How about having `TrackingManagerConstructor`'s `ConstructParticle` method setting some static data in `TrackingManagerIntegration` ? It might be better to have a conversation about this.

src/celeritas/ext/GeantPhysicsOptions.hh
sethrj: Let's defer.

src/accel/SharedParams.cc
```diff
@@ -258,6 +250,20 @@ SharedParams::SharedParams(SetupOptions const& options)
...
+        //! \todo: Set up default or all supported particles?
+        particles_ = IS::default_offload_particles();
+    }
+
+    if (mode_ == Mode::enabled)
+    {
+        // Set up offloaded particles based on user input
+        auto const& user_offload = options.offload_particles;
+        particles_ = user_offload.empty() ? IS::default_offload_particles()
+                                          : user_offload;
```
sethrj:
```suggestion
    if (mode_ == Mode::enabled || mode_ == Mode::kill_offload)
    {
        // Set up offloaded/killed particles based on user input
        auto const& user_offload = options.offload_particles;
        particles_ = user_offload.empty() ? IS::default_offload_particles()
                                          : user_offload;
```

Also maybe SharedParams is a better place to store the hardcoded "supported particles" vetors? We were originally doing that at one point, and it seems like a more accessible+appropriate place to get the list of default+supported.

src/accel/TrackingManagerConstructor.cc
```diff
@@ -70,6 +53,20 @@ TrackingManagerConstructor::TrackingManagerConstructor(
...
+ * Construct particles before \c ::ConstructProcess since any
+ * \c G4ParticleDefinition other than ions and shortlived must be created
+ * during Geant4's \c Pre_Init state.
+ */
+void TrackingManagerConstructor::ConstructParticle()
+{
+    // Construction of particles happens at offload_particles_ assignment,
+    // since it will instantiate the G4Particle::Definition() singletons
+    offload_particles_
+        = detail::IntegrationSingleton::instance().offloaded_particles();
```
sethrj: Wait, this should be assigned only in the delegating `TrackingManagerConstructor(TrackingManagerIntegration` constructor, right? Otherwise this class can't be used without the singleton.

src/accel/detail/IntegrationSingleton.cc
```diff
@@ -23,6 +29,46 @@ namespace celeritas
...
+            });
+    };
+
+    for (auto const& pd : user)
+    {
+        CELER_ASSERT(pd);
+        CELER_VALIDATE(find(pd),
+                       << "Particle \"" << pd->GetParticleName()
+                       << "\" (PDG = " << pd->GetPDGEncoding()
+                       << ") is not available in Celeritas");
```
sethrj: I forgot about PrintablePD in `geocel/GeantUtils.hh`: use that here.

sethrj (commented):
One more set of small changes I think? Remember we need to support the case *without* the singleton.

sethrj (approved):
Nice! Just a typo in the doxygen causing failures.

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -53,11 +53,11 @@ enum class RelaxationSelection
...
 {
     //! Enable muon pair production
-    bool pair_production{false};
+    bool pair_production{true};
     //! Enable muon ionization
-    bool ionization{false};
+    bool ionization{true};
     //! Enable muon bremsstrahlung
-    bool bremsstrahlung{false};
+    bool bremsstrahlung{true};
```
amandalund, line 60: Can we also enable MSC with the Urban model?

sethrj:
@stognini The final failure is in a diff used to show how to add celeritas in the documentation (it's now wrong!) Could we relax the ordering of `tmi.SetOptions(MakeOptions())`: check the user input particles, and if they're the default ones (or empty) then allow it as usual; if not, then fail and tell the user to move before the tracking constructor?

PR #1912, opened 2025-08-20
Fix volume instance IDs for background volumes in ORANGE

sethrj:
In ORANGE, even though most implementation volumes represent "volume instances", "background" volumes (fill volumes for a universe) are actually "volumes", because they're filling the parent universe rather than the local universe. This means that the volume instance ID can only be found by going up one universe level in the ORANGE tracker.

This is something of a hack: the background volume in a `orangeinp::UnitProto` is annotated by setting the label to `VolumeInstanceId{}` in `g4org::ProtoConstructor`; then converted from a proto to a `UnitInput` by the `InputBuilder`, and finally in `g4org::Converter` the empty volume instance IDs are replaced by (1) the world `VolumeInstanceId` for the top-level background volume, or (2) the `VolumeId` corresponding to the unit's label.


src/orange/detail/UniverseInserter.cc
```diff
@@ -140,22 +139,34 @@ UniverseId UniverseInserter::operator()(UniverseType type,
...
 
                 // Replace with volume label corresponding to it
                 var_label = volume_params_->volume_labels().at(vol_id);
             }
             else
             {
-                // No physical volume corresponds to it
-                volume_ids_.push_back(VolumeId{});
-                volume_instance_ids_.push_back(VolumeInstanceId{});
+                // No special metadata: just an impl volume e.g.  [EXTERIOR]
```
elliottbiondo: extra space

src/orange/g4org/ProtoConstructor.cc
```diff
@@ -111,11 +118,14 @@ auto ProtoConstructor::operator()(PhysicalVolume const& pv) -> SPUnitProto
 
 //---------------------------------------------------------------------------//
 /*!
- * Place this physical volume into the proto.
+ * Place this physical volume into a proto being constructed.
+ *
+ * It will return a "local child ID" if it generates a material input, but no t
```
elliottbiondo: not

elliottbiondo (approved):

Issue #1900, opened 2025-08-12
Refactor optical model input


sethrj:
We're having problems separating out model construction and materials and adding new models because they're all entangled.

## Short term

Support empty MFP grids for materials without data (needed for rayleigh):
- `PhysicsTrackView::calc_mfp` should check grid first, return inf if not present
- Loosen assertions in RayleighModel constructors and `build_mfps` and elsewhere
- Add a test

## Long term

We should move to a model like `inp::SurfacePhysics`. To `inp::OpticalPhysics` add:
```c++
using OpticalRayleighMaterial = ImportOpticalRayleigh;
using OpticalProperties = ImportOpticalProperty;

struct OpticalRayleighProcess
{
    std::map<OpticalMaterialId, OpticalRayleighMaterial> materials;
};

//! Processes and models for volumetric interactions of photons and material
// (add to inp::OpticalPhysics)
struct OpticalBulkPhysics
{
   OpticalRayleighProcess rayleigh;
   OpticalWlsProcess wls;
   OpticalWlsProcess wls2;
// later, add Mie
};

// Add to inp::OpticalPhysics:

  // Optical properties for each optical material
  std::vector< OpticalProperties > properties;
```

Steps:
1. Move optical properties from `ImportOpticalMaterial` to `OpticalPhysics` and update the rest of the code
2. Add `inp::OpticalBulkPhysics ` to constructor of `optical::PhysicsParams` and update `build_optical_params` in `celeritas/setup/Problem.cc` (eventually get rid of `ModelBuilder`)
3. Replace Rayleigh model constructor to take `inp::OpticalRayleighModel` and optical material params; remove from `imported.optical_models`, update GeantImporter to load data into the new `inp`  and build Rayleigh inside `optical::PhysicsParams` if any materials are present

hhollenb:
Related, I'm working on a simple surface physics model that does extremely trivial interactions for optical physics. Adding it to the current optical models requires hard coding it into a few places in the input and model builders, which isn't ideal for something that we'll use for debugging / validation / benchmarking but very rarely for production runs. I think it might be able to serve as a simple case for user-defined models while refactoring inputs?

sethrj:
@hhollenb I think that's a great idea, and do you think it could wait until we have the basic surface crossing "working" (or at least to a point where it doesn't crash and we can test it?)

hhollenb:
[...]

PR #1913, opened 2025-08-22
Enable optical SetupOptions and allow Rayleigh when some materials lack MFP

Rashika-Gupta:
This PR does three things:
**1. Enable optical options**
 - Allows optical photons and optical processes in `SetupOptions.cc.`
 - Adds and initializes the `optical_capacity` setup option.
 - Enables optical physics when materials have refractive indexes in `ImportedMaterials.cc`

**2. Build MFP when Rayleigh scattering data is missing** (with ref to issue #1900 )
 - Allows optical materials without Rayleigh models by removing the strict validation check.
 - Extends `MfpBuilder` to allow empty grids and initialize them appropriately. 
 - When Rayleigh data is unavailable, builds an MFP grid with infinite values instead of failing .
 
**3. Special case of `water`**
 - In `GeantImporter.cc::import_optical_material` for `rayleigh` model even if the `compressibility == 0` but if the `material = water` set the compressibility to default value.
 - Include a warning for overriding the temperature of water if no `Rayleigh` mean free paths are present.


src/celeritas/optical/MfpBuilder.hh
```diff
@@ -71,7 +71,13 @@ MfpBuilder::MfpBuilder(Values* real_data, GridValues* grid_data)
...
 void MfpBuilder::operator()(Args const&... args)
 {
-    insert_grid_(args...);
+    inp::Grid const& g = std::get<0>(std::tie(args...));
+    if (g.x.empty())
+    {
+        insert_grid_();
+    }
+    else
+        insert_grid_(args...);
```
sethrj: I think this function was when we had more complicated ways of building the grids (now the only two ways to build are `inp::Grid const&`  or with no arguments at all). Can you remove the templates and just change `operator()(inp::Grid const&)`, and add an `operator()()` that does "insert empty grid"?

src/accel/SetupOptions.hh
```diff
@@ -160,7 +160,7 @@ struct SetupOptions
 
     //!@{
     //! \name Celeritas stepper options
-
+    std::optional<inp::OpticalStateCapacity> optical_capacity;
```
sethrj, line 167: Make sure to document this with Doxygen; and perhaps it should go in a new "optical photon" block?

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -88,7 +88,7 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
     {
         build(mfp);
     }
-    else
+    else if (input_ && input_.imported_materials->rayleigh(mat))
```
sethrj, line 75: Some comments inside each of the if blocks would be good; on line 89, "// User explicitly provided Rayleigh mfp"; here, "MFPs can be calculated from user-given properties"; below, "// Build a grid with infinite MFP to prevent selection of the model"

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -111,6 +111,19 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
...
     }
+    else
+    {
+        auto mat_view = input_.materials->get(mat);
+        auto rindex_calc = mat_view.make_refractive_index_calculator();
+        auto energies = rindex_calc.grid().values();
+        if (!energies.empty())
+        {
+            inp::Grid g;
+            g.x.assign(energies.begin(), energies.end());
```
sethrj: Better to just make two points:
```suggestion
            g.x = {energies.front(), energies.back()};
```

src/celeritas/ext/GeantImporter.cc
```diff
@@ -520,7 +520,6 @@ import_optical_materials(detail::GeoOpticalIdMap const& geo_to_opt)
 
     auto const& mt = *G4Material::GetMaterialTable();
     CELER_ASSERT(mt.size() == geo_to_opt.num_geo());
-
```
sethrj: Revert unrelated change

src/accel/SetupOptions.hh
amandalund: The blank line after the group name is also needed here for the Doxygen to render correctly.

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -61,11 +61,12 @@ RayleighModel::RayleighModel(ActionId id, SPConstImported imported, Input input)
     {
         if (input_)
         {
-            CELER_VALIDATE(
-                imported_.mfp(mat) || input_.imported_materials->rayleigh(mat),
-                << "Rayleigh model requires either imported MFP or "
-                   "material parameters to build MFPs for each optical "
-                   "material");
+            if (!(imported_.mfp(mat)
+                  && input_.imported_materials->rayleigh(mat)))
```
amandalund: Should be `||` not `&&` since either the MFP or the material properties for calculating it are required, not both.

We might also need to relax the `VALIDATE` below, so the body of this loop can just be:
```c++
if (!(imported_mfp(mat) || input_.imported_materials->rayleigh(mat)))
{
    CELER_LOG(debug) << ...
}
```

src/celeritas/optical/ImportedMaterials.cc
```diff
@@ -26,7 +26,8 @@ ImportedMaterials::from_import(ImportData const& data)
     if (!std::any_of(data.optical_materials.begin(),
                      data.optical_materials.end(),
                      [](auto const& mat) {
-                         return mat.rayleigh || mat.wls || mat.wls2;
+                         return mat.rayleigh || mat.wls || mat.wls2
+                                || mat.properties.rindex;
```
amandalund: I'm not sure I understand the reason for this change... why do we want to build the `ImportedMaterials` for the models (or models themselves) if there is no model data in any of the materials?

Rashika-Gupta: We import materials even if Rayleigh data isn’t explicitly provided, because RINDEX is enough for photons to undergo Rayleigh scattering (as in the OpNovice example, where only refractive index is given) and undergoes rayleigh scattering. This will be further extended to mie scattering.

src/celeritas/optical/model/RayleighModel.cc
Rashika-Gupta: Yes, sorry ! that was a typo

src/celeritas/optical/model/RayleighModel.cc
Rashika-Gupta: Good point - i was thinking for the case when the shape of the grid had to be same as the other optical tables, but i am fine reducing it to two points if that is not the concern.

src/celeritas/optical/ImportedMaterials.cc
amandalund: Ok, I see---Geant4 will add the Rayleigh process and let it compete with the other processes even if the model data isn't present, but it won't actually undergo Rayleigh scattering unless either the MFP or the isothermal compressibility used to calculate it are provided (if they aren't, it will always return an infinite MFP). Do we still need/want to construct the model if it will never be sampled, or can we avoid adding it if no model data is present?

sethrj: Yeah, let's avoid constructing the model if there's no data to construct any nonzero cross sections.

Rashika-Gupta: Removing the refractive index property!

src/celeritas/optical/model/RayleighModel.cc
Rashika-Gupta: Done! Thanks

src/celeritas/optical/model/RayleighModel.cc
Rashika-Gupta: Done - thanks!

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -111,6 +105,27 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
         }
         build(grid);
     }
+    // Build a grid with infinite MFP to prevent selection of the model
+    else
+    {
```
sethrj: Comments belong inside the block to which they apply
```suggestion
    else
    {
        // Build a grid with infinite MFP to prevent selection of the model
```

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -111,6 +105,27 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
...
+        {
+            inp::Grid g;
+
+            // Using min and max energy values from the refractive index grid
+            // and setting MFP to infinity
+            g.x = {energies.front(), energies.back()};
+            g.y = {std::numeric_limits<real_type>::infinity(),
+                   std::numeric_limits<real_type>::infinity()};
+            CELER_LOG(debug) << "Material " << mat.get()
+                             << " has no MFP data - setting mfp to infinity";
```
sethrj: Minor change for consistency with our usual messages
```suggestion
                             << " has no MFP data: setting mfp to infinity";
```

src/celeritas/optical/model/RayleighModel.cc
sethrj: This isn't quite right; should be:
```c++
    else if (input_ && input_.imported_materials->rayleigh(mat))
    {
        // MFPs can be calculated from user given properties
```

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -84,11 +75,14 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
 {
     CELER_EXPECT(mat < imported_.num_materials());
 
+    // User explicitly provided Rayleigh MFP
```
sethrj: Move two lines down

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -59,23 +59,14 @@ RayleighModel::RayleighModel(ActionId id, SPConstImported imported, Input input)
...
-        {
-            CELER_VALIDATE(imported_.mfp(mat),
-                           << "Rayleigh model requires imported MFP for each "
-                              "optical material");
+            CELER_LOG(debug)
+                << "Rayleigh model has no MFP data for material " << mat.get();
         }
     }
 }
-
```
sethrj, line 78: Leave blank lines between functions to assist in code navigation

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -59,23 +59,14 @@ RayleighModel::RayleighModel(ActionId id, SPConstImported imported, Input input)
...
-                   "material parameters to build MFPs for each optical "
-                   "material");
-        }
-        else
-        {
-            CELER_VALIDATE(imported_.mfp(mat),
-                           << "Rayleigh model requires imported MFP for each "
-                              "optical material");
+            CELER_LOG(debug)
+                << "Rayleigh model has no MFP data for material " << mat.get();
```
sethrj: This log message is redundant with the one in the constructor; I'd delete this one to remove the complex conditionals.

src/celeritas/optical/MfpBuilder.hh
```diff
@@ -68,10 +67,17 @@ MfpBuilder::MfpBuilder(Values* real_data, GridValues* grid_data)
...
 {
-    insert_grid_(args...);
+    if (!grid)
+    {
+        // insert empty grid
+        insert_grid_();
+        return;
+    }
+
+    CELER_EXPECT(grid);
```
sethrj: This is redundant (and no longer a precondition for calling the function)
```suggestion
```

sethrj (approved):
Just a few minor changes; thanks for making this work!

PR #1918, opened 2025-08-28
Modularize and expose physics lists and PG generator

sethrj:
This improves the modularity of a few Celeritas classes and allows them to be used downstream, to be added as part of an `accel` testing framework.

- Moves the Celeritas->Geant4 primary generator class from celer-g4 to accel
- Remove redundant suffixes from inp-related classes (we had "monoenergetic" but "isotropicangle")
- Moves the EmStandardPhysics to `g4` (where future adapter classes will go) and sets its "type" so that it can be used to replace EM physics for *any* Geant4 modular constructor
- Makes MuHadPhysics inherit from EmPhysics so it can be a partial replacement for EM physics as well (it basically just omits "extra" EM)
- Reorganize EmStandardPhysics docs


src/celeritas/g4/EmStandardPhysics.cc
```diff
@@ -459,24 +430,8 @@ void EmStandardPhysics::add_e_processes(G4ParticleDefinition* p)
...
- * | Multiple scattering          | G4WentzelVIModel             |
- *
- * \note Currently all muon processes are disabled by default
- *
- * \note Prior to version 11.1.0, Geant4 used the \c G4BetheBlochModel for muon
- * ionization between 200 keV and 1 GeV and the \c G4MuBetheBlochModel above 1
- * GeV. Since version 11.1.0, the \c G4MuBetheBlochModel is used for all
- * energies above 200 keV.
- *
- * \todo Implement energy loss fluctuation models for muon ionization.
```
amandalund: Was the removal of this todo intentional?

src/celeritas/g4/EmStandardPhysics.hh
```diff
@@ -0,0 +1,101 @@
...
+/*!
+ * Construct Celeritas-supported Geant4 EM physics processes.
+ *
+ * This is a configurable plug-in replacement for \c G4EmStandardPhysics.
+ * Limitations:
+ * - No support for generic ions
+ * - No hadronic EM interactions
+ * - Limited support for muons (currently disabled by default)
+ * - Coulomb scattering and multiple scattering (high E) are currently
+ *   disabled.
```
amandalund: If we want to be more precise, Wentzel VI is unsupported but high E may be enabled with Urban.

src/celeritas/ext/GeantPhysicsOptions.hh
```diff
@@ -174,7 +177,7 @@ struct GeantPhysicsOptions
     MevEnergy lowest_electron_energy{0.001};  // 1 keV
     //! Tracking cutoff kinetic energy for muons/hadrons
     MevEnergy lowest_muhad_energy{0.001};  // 1 keV
-    //! Kill secondaries below the production cut
+    //! Kill secondaries from gamma processes below the production cut
```
amandalund: It's not necessarily just those from gamma processes, e.g. decay and annihilation can also produce secondaries below the production cut.

src/celeritas/g4/EmStandardPhysics.cc
```diff
@@ -459,24 +430,8 @@ void EmStandardPhysics::add_e_processes(G4ParticleDefinition* p)
...
+ * \note Remove processes from celeritas::detail::MuHadEmStandardPhysics when
+ * added here.
```
amandalund:
```suggestion
 * \note Remove processes from celeritas::detail::MuHadEmStandardPhysics when
 * muon physics is fully supported in Celeritas.
```

src/celeritas/ext/detail/MuHadEmStandardPhysics.hh
```diff
@@ -6,24 +6,20 @@
...
 //---------------------------------------------------------------------------//
 /*!
- * Construct EM standard physics not implemented in Celeritas.
- *
- * \todo Remove muon physics from this constructor once it is fully supported
- * in Celeritas.
+ * Construct EM standard physics, including those not implemented in Celeritas.
  */
-class MuHadEmStandardPhysics : public G4VPhysicsConstructor
+class MuHadEmStandardPhysics : public EmStandardPhysics
```
amandalund: Maybe we want to name this class something else now that it's all EM standard physics?

amandalund (approved):
Thanks @sethrj!

src/celeritas/ext/GeantPhysicsOptions.hh
sethrj: I may have misunderstood this then: https://geant4-forum.web.cern.ch/t/understanding-how-production-cuts-are-used/4665 

> you need to take into account, that ApplyCuts flag enable/disable secondary thresholds for gamma processes only (photo-effect, Compton,…). For ionisation and bremsstrahlung this flag not used at all, only value of a cut affects number of secondaries.

src/celeritas/ext/detail/MuHadEmStandardPhysics.hh
sethrj: True. Should we rename EmStandardPhysics something like "BasicEmStandardPhysics" since the G4EmStandardPhysics includes mu/had/ion?

amandalund: Sounds good, matches our `em_basic` data selection.

src/celeritas/g4/EmStandardPhysics.cc
sethrj: Yeah, there's enough un-implemented models and processes that I didn't think this one deserved a special to-do. Thoughts @stognini ?

src/celeritas/g4/EmStandardPhysics.hh
sethrj: I think this comment wasn't updated when we put that in.

src/celeritas/g4/EmStandardPhysics.cc
amandalund: It doesn't have to be a todo, but maybe a note: I think it's generally a good idea to keep track of any known differences between our implementation and Geant4, and this differs from e+/e- where we do support energy loss fluctuations.

src/celeritas/ext/GeantPhysicsOptions.hh
amandalund: Yeah it definitely won't affect ionization or brems since the production cut inherently prevents secondary production below that threshold, but I'm not sure that's true of all processes. I don't think there's anything preventing e.g. a positron at rest annihilating and producing secondaries below the gamma cut?

whokion: Yes, no production cut is applied to the two secondary gammas from e+e– annihilation at rest (at least in the model we adopted). The total gamma energy is to be above 2 × electron mass, which may be sufficient to keep both photons. On the other hand, if the energy of one photon is below the user-defined production cut, it can be deposited locally as energy deposition without loss of generality for HEP applications—this may be just a matter of choice.

src/celeritas/ext/detail/MuHadEmStandardPhysics.hh
whokion: **Basic**EmStandardPhysics is somewhat vague to me too - may be DefaultEmStandarPhysics, which I do not like either.  Not a big fan of the naming anyway^^.

whokion (approved):
Looks good to me.  Thanks.

src/celeritas/g4/EmStandardPhysics.cc
stognini: Agree with @amandalund . In the G4 physics list construction there is a fluctuation model assigned (see [here](https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-11.3-release/source/processes/electromagnetic/muons/src/G4MuIonisation.cc?ref_type=heads#L144)), but as long as we have a \note or \warning making sure we/users know the difference, I'm fine with that.

Another way of keeping track of this: add to our `CELER_LOG(debug)` message that fluctuation is not implemented for muons.

src/celeritas/ext/detail/MuHadEmStandardPhysics.hh
sethrj: Renaming to SupportedEmStandradPhysics (only celeritas-supported processes) and EmStandardPhysics (all EM processes)

PR #1919, opened 2025-08-28
Add custom error handler to unify G4 exception handlers

sethrj:
This pulls out a bit of work I left on a branch some time back and replaces the `celer-g4` exception handler by extending the main one slightly. I tested a number of exception-handling cases in celer-g4 and they all work. This extended class is to be used by a pending `accel` test harness.


amandalund (approved):
Looks good to me!

Issue #1921, opened 2025-08-28
Support surface normals in VecGeom


hhollenb:
Optical surface physics requires getting the surface normal from the geometry implementation. Currently there's a working surface normal implementation in ORANGE but we should also add one to VecGeom (the `VecGeomTrackView::normal()` function throws `CELER_NOT_IMPLEMENTED` at the moment). This limits any of the integration tests using the surface normal from the track's geometry, including:

- `celeritas/optical/PrimaryGenerator`
- `celeritas/optical/OpticalCollector`
- `app/celer-sim:lar`

Namely they call `InitBoundaryExecutor` as part of the optical photon transportation which (in an upcoming PR) will call `GeoTrackView::normal()`. For now I'll suggest limiting these tests to just running in ORANGE and then include VecGeom once surface normals are supported.

The convention we're using for surface normals is that they should always point from the post-volume into the pre-volume. In other words, the following assertion should pass after a call to `cross_boundary`: 
```
#include "celeritas/optical/surface/SurfacePhysicsUtils.hh"
CELER_ASSERT(is_entering_volume(track.geometry().dir(), track.geometry().normal());
```

sethrj:
@mrguilima Could you make this your highest priority?

sethrj:
This was accidentally closed; it isn't complete since normals still work only under certain circumstances (depending on whether entering/exiting a parent/child).

PR #1922, opened 2025-08-28
Add surface normal state to optical surface physics

hhollenb:
Small changes to add the global and local facet surface normals to the surface physics state. These will be necessary to implement roughness / reflectivity / interaction models downstream. Since only ORANGE supports surface normals right now, I'll limit the tests to only run with ORANGE core geometry, and I've opened an issue #1921 for support of normals in VecGeom.

Because this depends on #1801 I'm leaving it as a draft, but should be able to be reviewed soon after that's merged.


src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
```diff
@@ -77,14 +77,6 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
         = select_surface(track.surface(), geo.volume_instance_id());
     if (!oriented_surface)
     {
-        // Kill the track if the post-volume doesn't have a valid optical
-        // material and there's no surface
-        if (!post_volume_material)
-        {
-            track.sim().status(TrackStatus::killed);
-            return;
-        }
```
amandalund: Looks like this might have been reverted unintentionally.

src/celeritas/optical/surface/SurfacePhysicsParams.cc
```diff
@@ -154,7 +154,7 @@ SurfacePhysicsParams::SurfacePhysicsParams(ActionRegistry* action_reg,
 
 //---------------------------------------------------------------------------//
 /*!
- * Build surface data from inputs.
+ * Build surface data form inputs.
```
amandalund: Same here.

src/celeritas/optical/surface/SurfacePhysicsData.hh
```diff
@@ -143,17 +143,22 @@ struct SurfacePhysicsStateData
 
     StateItems<SurfaceId> surface;
     StateItems<SubsurfaceDirection> surface_orientation;
-    StateItems<SurfaceTrackPosition> surface_position;
+    StateItems<Real3> global_normal;
```
amandalund, line 146: Is it necessary to cache the global normal in the state data? I.e., can the geometry state change before we are finished using the normal?

amandalund (commented):
Thanks @hhollenb!

src/celeritas/optical/surface/SurfacePhysicsData.hh
sethrj: I suppose this is an optimization? Also I wonder if it would be better to transform the direction/polarization into the surface's reference frame and do all the work there, then transform back at the end (?) The normal here also might have a factor of -1 compared to the geometry-returned normal.

hhollenb: I was caching the normal because I thought the computation of surface normals in global coordinates was generally expensive. I'm not sure if the surface reference frame is generally well defined, and it might be more complicated to do the transformations?

sethrj: I think it might not be too expensive, but on reflection it *would* be good to isolate the 'calculating' geometry calls from the rest of the surface physics (otherwise we risk destroying our kernel occupancy when using vecgeom).

amandalund: Yeah, the occupancy is a good point. Generally unless we know something will be detrimental to performance I'd be inclined to save on memory for anything that scales with the number of threads (and add it to the list of [potential optimizations](https://github.com/celeritas-project/celeritas/wiki/Potential-optimizations)), but I don't have a great sense in this case so I'll leave it up to you guys.

app/celer-sim/CMakeLists.txt
```diff
@@ -133,4 +133,9 @@ celer_sim_test(lar
   lar-sphere.gdml gamma-4evt-15prim.hepmc3 lar-mctruth.root
 )
 
+if(NOT CELERITAS_CORE_GEO STREQUAL "ORANGE")
+    set_tests_properties("app/celer-sim:lar:cpu" PROPERTIES DISABLED True)
```
amandalund: Is the time frame for implementing VecGeom surface normals short enough that it would be reasonable to hold off on this until that's in (so we can continue to run optical problems with VecGeom)?

amandalund (approved):

app/celer-sim/CMakeLists.txt
sethrj: As a temporary hack (as long as we don't care about correct answers) we could return `geo.dir()` inside the call to `geo.normal()`... so that way we always get something physical.

src/celeritas/optical/surface/SurfacePhysicsData.hh
hhollenb: Since we no longer guarantee the convention of the geometry returned normal, we'll either have to store a surface orientation and recompute the normal and correct it every time, or store a corrected copy of the global normal. Definitely something to benchmark

sethrj (approved):
I think this is good! The traversal direction/orientation code may warrant further review but this is a step in the right direction. Since the global normal is only used to sample the facet normal, we could perhaps use a single "normal" buffer that is initialized every step with the global normal before the start of the facet sampling, locally copied into a register (and sign-corrected) to check for validity, and replaced with the facet normal for use in the reflection kernel.

PR #1924, opened 2025-08-29
Refactor Geant4 launching macros into a reusable script

sethrj:
Our example cmake currently has code to launch all combinations of Geant4 serial/mt/task with Celeritas CPU/GPU/none/kill-offload . This cleans up and extends the code and moves it to the main `cmake/` directory so that it can be used by an upcoming PR for running Geant4 integration tests.


cmake/CeleritasUtils.cmake
```diff
@@ -131,20 +141,45 @@ endfunction()
...
-  set(_CELER_G4ENV)
-  foreach(_ds IN LISTS Geant4_DATASETS)
-    set(_key ${Geant4_DATASET_${_ds}_ENVVAR})
-    set(_val ${Geant4_DATASET_${_ds}_PATH})
-    set(_env "$ENV{${_key}}")
-    if(_env AND NOT _env STREQUAL _val)
-      message(VERBOSE "CELER_G4ENV: using ${_key}=${_env} from user environment")
-      list(APPEND _CELER_G4ENV "${_key}=${_env}")
+function(celeritas_get_g4env var)
+  if(NOT DEFINED CELERITAS_G4ENV)
```
pcanal, line 145: Since `CELERITAS_G4ENV` is immutable (per the `if` on line 145), wouldn't it make more sense to just set it once when finding Geant4?  Then the (current) callers of this function would just copy `CELERITAS_G4ENV` ... 

Humm ... right ... This is to support the 'reusable' part of the scheme where there is no good place to set `CELERITAS_G4ENV` unless we require `Celeritas` to be "search for" only after search for `Geant4` ....

cmake/CeleritasG4Tests.cmake
```diff
@@ -0,0 +1,124 @@
...
+  set_tests_properties("${test_name}" PROPERTIES
+    ENVIRONMENT "${_env}"
+    LABELS "${labels}"
+    ${_extra_props}
+    ${ARGN}
+  )
+endfunction()
+
+function(celeritas_g4_add_tests target)
+  cmake_parse_arguments(PARSE "DISABLE" "NAME" "ARGS;LABELS" ${ARGN})
```
pcanal: This looks correct but is also not yet exercised (except for the DISABLE option)

pcanal (approved):
LGTM.

cmake/CeleritasG4Tests.cmake
sethrj: I can delete these parts for now and restore them in #1925 where they're used.

pcanal: No need.  Guaranteed future testing is good enough :)

PR #1925, opened 2025-08-29
Add minimal extensible testing for Geant4 integration

sethrj:
~~CHAINED on #1918, #1919, #1924, #1932~~

This adds test harnesses for setting up and running a Geant4 application through googletest. It allows us to create one unit test for each offload method (tracking manager, fast sim, user action) and test all combinations of geant4 threads with all combinations of celeritas offload. This will be extended to support error checking, different physics options (e.g., optical only, muon only), etc.  It's intended more for comprehensive sanity checks than detailed physics comparisons or user examples, but I could imagine it being used for those as well.

To be fixed in follow-on PRs:
- Weight flag isn't being activated in step collector (needs to be set in GeantSd.cc)
- EM Rayleigh scattering cross sections are inconsistent for Geant4 11.0 (lower and upper grids overlap)
- Aborting the run manager (due to a caught exception) during `BeginOfRun` doesn't stop the first event from being run, which leads to many follow-on exceptions when the initial setup fails
- Primary generator energy is "real_type" (a quantity under the hood) but should be "double"

### DIFF
--- a/src/accel/TrackingManagerIntegration.cc
+++ b/src/accel/TrackingManagerIntegration.cc
@@ -12,6 +12,8 @@
 #include <G4Threading.hh>
 #include <G4Version.hh>
 
+#include "geocel/ScopedGeantExceptionHandler.hh"
+
 #if G4VERSION_NUMBER >= 1100
 #    include <G4VTrackingManager.hh>
 
@@ -167,6 +169,13 @@ void TrackingManagerIntegration::BeginOfRunAction(G4Run const*)
     if (G4Threading::IsMasterThread())
     {
         singleton.initialize_shared_params();
+    }
+
+    if (CELER_UNLIKELY(ScopedGeantExceptionHandler::suppressed_fatal()))
+    {
+        CELER_LOG(debug)
+            << R"(Canceling BeginOfRunAction: initialization failed)";
+        return;
     }
 
     bool enable_offload = singleton.initialize_local_transporter();
--- a/src/accel/TrackingManagerIntegration.cc
+++ b/src/accel/TrackingManagerIntegration.cc
@@ -12,8 +12,6 @@
 #include <G4Threading.hh>
 #include <G4Version.hh>
 
-#include "geocel/ScopedGeantExceptionHandler.hh"
-
 #if G4VERSION_NUMBER >= 1100
 #    include <G4VTrackingManager.hh>
 
@@ -169,13 +167,6 @@ void TrackingManagerIntegration::BeginOfRunAction(G4Run const*)
     if (G4Threading::IsMasterThread())
     {
         singleton.initialize_shared_params();
-    }
-
-    if (CELER_UNLIKELY(ScopedGeantExceptionHandler::suppressed_fatal()))
-    {
-        CELER_LOG(debug)
-            << R"(Canceling BeginOfRunAction: initialization failed)";
-        return;
     }
 
     bool enable_offload = singleton.initialize_local_transporter();
--- a/test/accel/IntegrationTestBase.cc
+++ b/test/accel/IntegrationTestBase.cc
@@ -53,22 +53,21 @@ namespace test
 namespace
 {
 //---------------------------------------------------------------------------//
+using namespace std::string_literals;
+
 std::string thread_label()
 {
-    if (int thread = G4Threading::G4GetThreadId(); thread >= 0)
-    {
-        return std::to_string(thread);
-    }
-    return "M";
+    return G4Threading::IsMasterThread()
+               ? "M"s
+               : std::to_string(G4Threading::G4GetThreadId());
 }
 
 std::string thread_description()
 {
-    if (int thread = G4Threading::G4GetThreadId(); thread >= 0)
-    {
-        return "worker thread " + std::to_string(thread);
-    }
-    return "main thread";
+    return G4Threading::IsMasterThread()
+               ? "main thread"s
+               : "worker thread "s
+                     + std::to_string(G4Threading::G4GetThreadId());
 }
 
 //---------------------------------------------------------------------------//
--- a/test/accel/IntegrationTestBase.cc
+++ b/test/accel/IntegrationTestBase.cc
@@ -336,9 +336,11 @@ auto LarSphereIntegrationMixin::make_physics_input() const -> PhysicsInput
  */
 auto LarSphereIntegrationMixin::make_primary_input() const -> PrimaryInput
 {
+    using MevEnergy = Quantity<units::Mev, double>;
+
     PrimaryInput result;
     result.pdg = {pdg::electron()};
-    result.energy = inp::MonoenergeticDistribution{units::MevEnergy{10}};
+    result.energy = inp::MonoenergeticDistribution{MevEnergy{10}};
     result.shape = inp::PointDistribution{from_cm({99, 0.1, 0})};
     result.angle = inp::IsotropicDistribution{};
     result.num_events = 4;
@@ -364,10 +366,12 @@ auto LarSphereIntegrationMixin::make_sens_det(std::string const& sd_name)
  */
 auto TestEm3IntegrationMixin::make_physics_input() const -> PhysicsInput
 {
+    using MevEnergy = Quantity<units::Mev, double>;
+
     PhysicsInput result = Base::make_physics_input();
     result.em_bins_per_decade = 14;
-    result.min_energy = units::MevEnergy{0.1};
-    result.lowest_electron_energy = units::MevEnergy{1};
+    result.min_energy = MevEnergy{0.1};
+    result.lowest_electron_energy = MevEnergy{1};
     result.default_cutoff = 0.1 * units::centimeter;
     return result;
 }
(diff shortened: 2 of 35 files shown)


test/accel/IntegrationTestBase.cc
```diff
@@ -0,0 +1,411 @@
...
+namespace
+{
+//---------------------------------------------------------------------------//
+std::string thread_label()
+{
+    if (int thread = G4Threading::G4GetThreadId(); thread >= 0)
+    {
+        return std::to_string(thread);
+    }
+    return "M";
```
stognini: I think this is easier to read:

```suggestion
    return G4Threading::IsMasterThread() ? "M" : std::to_string(G4Threading::G4GetThreadId());
```

test/accel/IntegrationTestBase.cc
```diff
@@ -0,0 +1,411 @@
...
+}
+
+std::string thread_description()
+{
+    if (int thread = G4Threading::G4GetThreadId(); thread >= 0)
```
stognini: Similar here.

example/geant4/CMakeLists.txt
```diff
@@ -56,15 +56,15 @@ celeritas_get_disable_device(CELER_DISABLE_DEVICE)
 
 #-----------------------------------------------------------------------------#
 
+set(_needs_g410)
+if(Geant4_VERSION VERSION_LESS 11.0)
+  set(_needs_g410 DISABLE)
+endif()
```
amandalund: `_needs_g410` is a bit confusing since it requires >=11.0; maybe `_needs_geant4_11_0`?

src/accel/TrackingManagerIntegration.cc
```diff
@@ -142,6 +154,11 @@ TrackingManagerIntegration& TrackingManagerIntegration::Instance()
  */
 void TrackingManagerIntegration::BeginOfRunAction(G4Run const*)
 {
+    CELER_VALIDATE(G4VERSION_NUMBER >= 1100,
+                   << "the current version of Geant4 (" << G4VERSION_NUMBER
+                   << ") is too old to support the tracking manager offload "
+                      "interface");
```
stognini: I'd also add "... interface. Requires 11.0.0 or newer."

Same for the Constructor.

src/accel/TrackingManagerConstructor.cc
```diff
@@ -93,6 +101,7 @@ void TrackingManagerConstructor::ConstructProcess()
     auto* transporter = this->get_local_transporter();
     CELER_VALIDATE(transporter, << "invalid null local transporter");
 
+#if G4VERSION_NUMBER >= 1100
```
stognini: If there is a validate that should prevent the construction of the class, why do we have this here?

EDIT: I see your point. Since we keep it during cmake, we need it to compile. And hitting validate seems more useful than removing it as a source in the cmakelists.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -0,0 +1,104 @@
...
+        EXPECT_EQ(0, local_transport.GetBufferSize());
+    }
+};
+
+//---------------------------------------------------------------------------//
+class LarSphere : public LarSphereIntegrationMixin, public TMITestBase
+{
+};
+
+TEST_F(LarSphere, run)
```
stognini, line 73: I think a useful test addition would be a subset of physics/particles:
Maybe a single particle with a single process:

```cpp
auto physics_input = PhysicsInput::deactivate();
physics_input.ionization = true;
```

and
```cpp
auto setup_options = make_setup_options()
setup_options.offload_particles = {G4Electron::Definition()};
```

src/accel/TrackingManagerConstructor.cc
sethrj: The code below it requires methods/types that won't compile in the newer versions.

test/accel/TrackingManagerIntegration.test.cc
sethrj: There are *many* useful test additions to be made. I wanted to get the basic structure in here and then we can expand and fix the errors that will inevitably crop up :)

test/accel/IntegrationTestBase.cc
```diff
@@ -0,0 +1,410 @@
...
+#else
+            std::shared_ptr<G4RunManager>()
+#endif
+        };
+        CELER_ASSERT(rm);
+
+        // Access the particle table before creating the run manager, so that
+        // missing environment variables like G4ENSDFSTATEDATA get caught
+        // cleanly rather than segfaulting
+        G4ParticleTable::GetParticleTable();
```
amandalund: Sounds like this should be moved up before the run manager creation (or the comment should be modified).

test/accel/IntegrationTestBase.hh
```diff
@@ -0,0 +1,135 @@
...
+    //! \name Dispatch from user run/event actions
+    virtual void BeginOfRunAction(G4Run const* run) = 0;
+    virtual void EndOfRunAction(G4Run const* run) = 0;
+    virtual void BeginOfEventAction(G4Event const* event) = 0;
+    virtual void EndOfEventAction(G4Event const* event) = 0;
+    //!@}
+};
+
+//---------------------------------------------------------------------------//
+//! Generate TestEM3 geometry with 100 MeV electrons
```
amandalund:
```suggestion
//! Generate LAr sphere geometry with 10 MeV electrons
```

test/accel/IntegrationTestBase.cc
```diff
@@ -0,0 +1,410 @@
...
+/*!
+ * Create physics list: default is EM only using make_physics_input.
+ */
+auto TestEm3IntegrationMixin::make_physics_input() const -> PhysicsInput
+{
+    using MevEnergy = Quantity<units::Mev, double>;
+
+    PhysicsInput result = Base::make_physics_input();
+    result.em_bins_per_decade = 14;
+    result.min_energy = MevEnergy{0.1};
```
amandalund: Can you remind me what the `min_energy` option does in practice?

amandalund (commented):
Nice, thanks @sethrj! Looks like the GPU tests are currently failing (`(M,celer.init.global): Celeritas runtime error: activate_device may be called only once per application`); otherwise just a few minor comments.

test/accel/TrackingManagerIntegration.test.cc
stognini: That sounds good to me.

test/accel/UserActionIntegration.test.cc
```diff
@@ -0,0 +1,102 @@
...
+{
+};
+
+TEST_F(TestEm3, run)
+{
+    auto& rm = this->run_manager();
+    UAI::Instance().SetOptions(this->make_setup_options());
+
+    rm.Initialize();
+    rm.BeamOn(2);
```
Rashika-Gupta, line 97: should you print anything about `Em3` here like `LarSphere` ?

Rashika-Gupta (approved):
Just one minor comment but, can be completely ignored. 
Thanks

test/accel/IntegrationTestBase.cc
amandalund: Ok right, I think it sets the lower energy limit for the physics tables.

So it seems like for Geant4 <11.1 (when the boundary between the Rayleigh low and high energy lambda tables changed from 100 to 150 keV) we're setting the low energy limit exactly equal to that 100 keV prime value in this test. This is allowed by `G4VEmProcess::SetMinKinEnergyPrim()` and won't produce a warning, but when the lambda table is built, if the min energy is >= the prime energy the upper limit for the lower table is increased to twice the min energy. This would cause the tables to overlap.

If instead the min energy is greater than the prime energy, `G4VEmProcess::SetMinKinEnergyPrim()` will warn and won't set the prime value. I think in this case the high energy table won't be built at all, so we won't hit the contiguous increasing error. I think to get around that error we can check if the tables completely overlap and ignore the lower one if so. I can try updating that after this is merged if you'd like.

amandalund:
Looks like the GPU tests are still failing?

sethrj:
Yep @amandalund , working on the GPU tests now.  (It's just because googletest also does device initialization. We may end up wanting to loosen that restriction anyway.)  Thanks!

test/accel/IntegrationTestBase.cc
sethrj: Ahhh okay. I was just trying with this command to reduce the simulation time (but inadvertently caused more problems). Let me just revert the min_energy change here and we can add a check in the physics importer.

test/accel/UserActionIntegration.test.cc
sethrj: Yep! Added.

sethrj:
@amandalund I've reverted all the "fixes" I made to get this working by disabling some of the tests (like GPU). To improve visibility in the release notes I'll add those back as small PRs.

EDIT: I'll revert the quantity fix too.

amandalund (approved):

PR #1720, opened 2025-04-04
Update VecGeom support for 2.0 solid and surface

mrguilima:
Creating this (draft) PR as suggested yesterday at the standup meeting.

I tried today to merge in the develop branch, and then it does not compile.  However, this was passing all tests (including orange, e.g. VecGeom_DIR undefined).

For surface model, tested against VecGeom commit c3ee06912.
For solids model, tested agains VecGeom commit 7b8b07ba4 (then the head of v1-patches).

All tests also build ok for solids model, against the same commit above used for the surface model (c3ee06912), however the solids model fail several tests.  Overall, the solids model currently looks better (e.g. has better agreement agains surface model and Orange) using v1-patches than using VecGeom's master branch.

PR #1926, opened 2025-08-29
Implement VecgeomTrackView::normal()

mrguilima:
This implements the VecgeomTrackView::normal(), which dispatches to VecGeom's PlacedVolume()::Normal() function, where the physical volume called depends on which volume (surface) the track is on, as the track can be either leaving current volume, or going deeper into a daughter volume.

A very basic test was added, which just checks that the normal() returns a unit-vector at every boundary crossing during the track() function of many geometry *Test functions.


src/geocel/vg/VecgeomTrackView.hh
```diff
@@ -397,7 +397,13 @@ CELER_FUNCTION bool VecgeomTrackView::is_on_boundary() const
...
 CELER_FUNCTION Real3 VecgeomTrackView::normal() const
 {
-    CELER_NOT_IMPLEMENTED("VecgeomTrackView::normal");
+    CELER_EXPECT(this->is_on_boundary() || this->is_next_boundary());
+    vecgeom::Vector3D<vecgeom::Precision> normal;
+    auto pos = detail::to_vector(this->pos());
+    auto ok = this->is_on_boundary() ? vgstate_.Top()->Normal(pos, normal)
+                                     : vgnext_.Top()->Normal(pos, normal);
+    CELER_ENSURE(ok);
+    return Real3(normal);
```
sethrj: I think we can `CELER_EXPECT(!this->is_outside() && this->is_on_boundary())`

Also I'm surprised the last line works... it shouldn't! I see that Vector3D has an *implicit* `operator bool`, perhaps your vector is being converted to a bool which is being cast to a scalar that's filling the array's first element (i.e. returning {1,0,0} always). Try using `detail::to_array(normal)`.

test/geocel/GenericGeoTestBase.t.hh
```diff
@@ -318,6 +318,8 @@ auto GenericGeoTestBase<HP>::track(Real3 const& pos,
         try
         {
             geo.cross_boundary();
+            auto normal = geo.normal();
+            CELER_ASSERT(is_soft_unit_vector(normal));
```
sethrj: Since this is in a test, use `EXPECT_TRUE(...) << repr(normal)` to get useful output.

src/geocel/vg/VecgeomTrackView.hh
mrguilima: So in VecgeomTrackView::is_on_boundary() only checks vgstate_, not vgnext_, to be on the boundary.
I wanted to make sure that a track going from mother to daughter would be covered, as I would expect vgstate_ not to be on the boundary, in this case.

mrguilima: As for the vector, I used a printout, and I got many normal vectors different from (1,0,0). Anyway, I am fixing that line.

sethrj: vgnext_ isn't "mother to daughter", it's "what could the next track be along this direction". You could be exiting one volume (on the boundary), then search for another along the same direction, and that would be vgnext. You could be in the middle of the volume looking for a potential crossing, and that's vgnext. It's not about the parent/daughter.

But after some debugging on this over the weekend (see #1934) I did find that indeed the issue is that the `normal` call only works properly if the track is inside the daughter volume when it's called. There's going to have to be some nasty logic for vg 1.x to get this working correctly. It looks like vg 2.x has allowed the "SetLastExited" function additional flexibility (I think it's effectively unused in 1.x) so maybe we can use that to store the previous daughter if we just exited.

test/geocel/GenericGeoTestBase.t.hh
sethrj: This is all going to be replaced by #1934.

src/geocel/vg/VecgeomTrackView.hh
```diff
@@ -394,10 +399,41 @@ CELER_FUNCTION bool VecgeomTrackView::is_on_boundary() const
...
+    auto* tr = pv.GetTransformation();
+    CELER_ASSERT(tr);
+
+    // Get local position
+    auto local_pos = tr->Transform(detail::to_vector(pos_));
+
+    // Calculate normal
+    vecgeom::Vector3D<real_type> local_normal{0, 0, 0};
+    CELER_ASSERT(pv.GetUnplacedVolume());
+    auto success = pv.GetUnplacedVolume()->Normal(local_pos, local_normal);
```
hhollenb, line 422: My understanding is that this will attempt to calculate the surface normal of the current volume, and fail if the point is too far from the surface. This makes sense if going from the current volume to the mother volume, but I'm not sure it works for current volume to a daughter?

hhollenb (commented):
Revisiting this PR since it will be relevant for optical surface physics integration tests.

src/geocel/vg/VecgeomTrackView.hh
sethrj: That's correct; I tried to communicate this in my comment above:
> But after some debugging on this over the weekend (see https://github.com/celeritas-project/celeritas/pull/1934) I did find that indeed the issue is that the normal call only works properly if the track is inside the daughter volume when it's called. There's going to have to be some nasty logic for vg 1.x to get this working correctly. It looks like vg 2.x has allowed the "SetLastExited" function additional flexibility (I think it's effectively unused in 1.x) so maybe we can use that to store the previous daughter if we just exited.

After #1720 is merged, I think the next task for @mrguilima will be to tear apart the solid navigator so that it's more closely tailored to our use case (finding the distance to the next volume but not exiting until we cross the boundary).

PR #1928, opened 2025-09-02
Add builtin surface roughness models

hhollenb:
Adds support for built-in roughness models: polished, uniform smear, and Gaussian roughness.

The `BuiltinSurfaceModelBuilder` builds surface models from supported input data in the form of `std::map<PhysSurfaceId, InputT>`. For reflectivity and interaction models (not yet supported), `FakeSurfaceModel`s are built instead. `BuiltinSurfaceModel` provides a common interface for accessing surfaces from the input data and wrapping executors with an appropriate applier (cf. `InteractionApplier` in volumetric physics models).

The `SurfaceModelView` is a wrapper view to directly access physics relevant for a surface model. This encapsulates a lot of the bookkeeping of the surface traversal and represents a single physical surface crossing. This also hides some of the previous complications of `SurfacePhysicsView`, which now primarily handles subsurface traversal.

Should be direct to add reflectivity and interaction models once we get calculators for them implemented.


app/celer-sim/CMakeLists.txt
```diff
@@ -133,4 +133,9 @@ celer_sim_test(lar
   lar-sphere.gdml gamma-4evt-15prim.hepmc3 lar-mctruth.root
 )
 
+if(NOT CELERITAS_CORE_GEO STREQUAL "ORANGE")
+    set_tests_properties("app/celer-sim:lar:cpu" PROPERTIES DISABLED True)
+    set_tests_properties("app/celer-sim:lar:gpu" PROPERTIES DISABLED True)
+endif()
```
sethrj: This is my top priority.

src/celeritas/optical/CoreTrackView.hh
```diff
@@ -227,6 +234,20 @@ CELER_FUNCTION auto CoreTrackView::surface_physics() const -> SurfacePhysicsView
...
+ * Return a surface model view for the given step.
+ */
+CELER_FUNCTION auto CoreTrackView::surface_model(SurfacePhysicsOrder step) const
+    -> SurfaceModelView
+{
+    auto s_physics = this->surface_physics();
+    CELER_EXPECT(s_physics.is_crossing_boundary());
+
+    return s_physics.surface_model(
+        s_physics.traversal_direction(this->geometry().dir()), step);
```
sethrj: I think the core track view is best left as a very thin wrapper that provides access to the most immediate track data (anything we store directly in the core state/params). It's got enough data to manage that adding more functionality risks overburdening it.

src/celeritas/optical/action/ActionLauncher.device.hh
```diff
@@ -44,15 +44,15 @@ namespace optical
...
+template<class F, class StepActionT = OpticalStepActionInterface>
 class ActionLauncher : public KernelLauncher<F>
 {
     static_assert(
         (std::is_trivially_copyable_v<F> || CELERITAS_USE_HIP
          || CELER_COMPILER == CELER_COMPILER_CLANG)
             && !std::is_pointer_v<F> && !std::is_reference_v<F>,
         R"(Launched action must be a trivially copyable function object)");
-    using StepActionT = OpticalStepActionInterface;
+    // using StepActionT = OpticalStepActionInterface;
```
sethrj: I think only the constructors need to be templated on the step action. It's purely for obtaining the label for diagnostic export.

src/celeritas/optical/surface/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,194 @@
...
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using SPModel = std::shared_ptr<SurfaceModel>;
+    //!@}
+
+  public:
+    // Construct with storage to fill
+    inline BuiltinSurfaceModelBuilder(std::vector<SPModel>& models);
```
sethrj:
```suggestion
    explicit inline BuiltinSurfaceModelBuilder(std::vector<SPModel>& models);
```

src/celeritas/phys/SurfaceModel.hh
```diff
@@ -45,7 +45,7 @@ class SurfaceModel
     virtual ~SurfaceModel() = 0;
 
     //! Get the list of surfaces/layers this applies to
-    virtual VecSurfaceLayer get_surfaces() const = 0;
+    virtual VecSurfaceLayer const& get_surfaces() const = 0;
```
sethrj, line 46: My thought for this was we don't want to force the underlying class to store them as vectors (suppose they have a map?) and we only need to call it ~once per run.

src/celeritas/optical/surface/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,194 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/optical/surface/BuiltinSurfaceModelBuilder.hh
```
sethrj: This should be a detail I think; most data structure builders are.

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
```diff
@@ -0,0 +1,150 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/optical/surface/model/BuiltinSurfaceModel.hh
```
sethrj: To avoid making these paths too deep, what do you think of having `bulk` and `surface` as the two main top-level directories for physics, with mostly detail classes underneath?

src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
```diff
@@ -93,8 +93,13 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
     surface_physics
         = SurfacePhysicsView::Initializer{oriented_surface.surface,
                                           oriented_surface.orientation,
+                                          geo.normal(),
                                           pre_volume_material,
                                           post_volume_material};
+
+    CELER_ASSERT(
+        is_entering_surface(geo.dir(), surface_physics.global_normal()));
```
sethrj, line 110: Working on the "geo.normal" stuff now, I don't think we should force the geometry to flip the sign of the normal based on what direction a track is traveling/what side of the surface the particle is. Here I think we should update the normal's sign as needed at the start of the boundary crossing based on the track's physical direction, and we might also need/want to keep track of the traversal direction, rather than using the physical direction `s_physics.traversal_direction(this->geometry().dir())`.

sethrj (commented):
I'm still working through this but it's looking fantastic! I've got a few initial comments to suggest while I wrap up the vecgeom surface normal work.

src/celeritas/phys/SurfaceModel.hh
hhollenb: If their input data is a map, then they'd need to build the surface model data on the host/device and the original input map won't need to be stored. Since models will need to build a collection indexed by `InternalSurfaceId` they'll have to iterate over surface IDs to implicitly build this collection, so they might as well store a vector since that's the only data not saved to host/device data.

sethrj: That's true!

hhollenb: Still something we can revert if we encounter a use case, but at least for the built-in models this will be true.

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
hhollenb: I'm generally fine however we want to divvy it up, this was mainly to help me keep the directories clean while drafting the model files and testing different ways to implement them.

hhollenb:
@sethrj I've been seeing some errors with finding `sincos` in the `GaussianRoughnessSampler` when trying to build some tests. Not sure what the regression was, maybe a missing include?

sethrj:
@hhollenb Can you post the errors? The only ones I'm seeing in the CI (about missing operators) is due to not including `corecel/math/ArrayOperators.hh` in `surface/detail/InitBoundaryExecutor.hh`.

hhollenb:
Sorry I was looking at the build-docker tests:
```
 [429/1146] Building CUDA object src/celeritas/CMakeFiles/celeritas.dir/optical/surface/model/GaussianRoughnessModel.cu.o
FAILED: src/celeritas/CMakeFiles/celeritas.dir/optical/surface/model/GaussianRoughnessModel.cu.o 
/usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -Dceleritas_EXPORTS -I/__w/celeritas/celeritas/src -I/__w/celeritas/celeritas/build/include -isystem /usr/local/cuda/targets/x86_64-linux/include -isystem /opt/view/include -isystem /opt/software/linux-rocky9-x86_64_v3/gcc-11.5.0/openmpi-5.0.6-qzolp62fnhdcym2hkvnbvxujqzgeqtf2/include -Werror all-warnings -g -std=c++17 "--generate-code=arch=compute_70,code=[compute_70,sm_70]" -Xcompiler=-fPIC -MD -MT src/celeritas/CMakeFiles/celeritas.dir/optical/surface/model/GaussianRoughnessModel.cu.o -MF src/celeritas/CMakeFiles/celeritas.dir/optical/surface/model/GaussianRoughnessModel.cu.o.d -x cu -c /__w/celeritas/celeritas/src/celeritas/optical/surface/model/GaussianRoughnessModel.cu -o src/celeritas/CMakeFiles/celeritas.dir/optical/surface/model/GaussianRoughnessModel.cu.o
/__w/celeritas/celeritas/src/celeritas/optical/surface/GaussianRoughnessSampler.hh(106): error: no instance of overloaded function "celeritas::sincos" matches the argument list
            argument types are: (celeritas::real_type, celeritas::real_type *, celeritas::real_type *)
          sincos(alpha, &sin_alpha, &cos_alpha);
          ^
/__w/celeritas/celeritas/src/corecel/math/Turn.hh(124): note #3326-D: function "celeritas::sincos(celeritas::IntQuarterTurn, int *, int *)" does not match because argument #1 does not match parameter
                          void sincos(IntQuarterTurn r, int* sinv, int* cosv)
                               ^
/__w/celeritas/celeritas/src/corecel/math/Turn.hh(98): note #3327-D: candidate function template "celeritas::sincos(celeritas::Turn_t<T>, T *, T *)" failed deduction
                            void sincos(Turn_t<T> r, T* sinv, T* cosv)
                                 ^
```
I'll fix the `InitBoundaryExecutor` missing include as well

sethrj:
Aha, I see what's happening... the error is when building with NVCC, and there's a `celeritas::sincos` that shadows the builtin nvidia `sincos` 🤔 we need to get the std/global/nvidia namespace issues worked out. Maybe we just add celeritas functions that call the appropriate global/std namespace functions for host/nvidia? This issue is more complicated than it sounds due to functions like  https://en.cppreference.com/w/cpp/numeric/math/fabs being for a single type in the global (C) namespace but overloaded in the `std` (C++) namespace.

amandalund:
> we need to get the std/global/nvidia namespace issues worked out

Not sure we made an issue for this, but a related reminder from the CUDA math API:

> Note also that due to implementation constraints, certain math functions from std:: namespace may
be callable in device code even via explicitly qualified std:: names. However, such use is discouraged,
since this capability is unsupported, unverified, undocumented, not portable, and may change without
notice

sethrj:
Yep I do remember that @amandalund . Created https://github.com/celeritas-project/celeritas/issues/1950 (which also explains why @hhollenb 's latest build is failing the CPU build...)

app/celer-sim/CMakeLists.txt
sethrj: I think we should be able to delete it now? Certainly it'll at least work with Geant4...

src/celeritas/optical/surface/SurfaceModelView.hh
```diff
@@ -0,0 +1,126 @@
...
+                                           OptMatId post_mat);
+
+    // Get subsurface track direction
+    inline CELER_FUNCTION SubsurfaceDirection direction() const;
+
+    // Get surface model ID
+    inline CELER_FUNCTION SurfaceModelId surface_model() const;
+
+    // Get internal surface ID for the model
+    inline CELER_FUNCTION InternalSurfaceId internal_surface_id() const;
```
sethrj: It might be good to clarify whether this is "internal" as in "implementation" or "sub-surface" (I can't remember what we ended up deciding or postponing...)

EDIT: I remember, originally this was a type alias inside `SurfaceModel` , and it's just a local index for the data. I wonder if it'd be better to call it ModelSurfaceId or something after all? :\

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -10,13 +10,33 @@
...
 #include "SurfacePhysicsData.hh"
 #include "SurfacePhysicsUtils.hh"
 
 namespace celeritas
 {
 namespace optical
 {
+
+namespace
+{
```
sethrj: It's generally not a good idea to put anonymous namespace/static functions in a header file because the code will be duplicated across multiple translation units (rather than eliminated by the linker). Since only `operator+` is defined (no other math operators, and only with a mixed type) I recommend putting the function in SurfaceUtils and giving it a descriptive name like `advance_subsurface_along(pos, dir)`

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -94,21 +114,18 @@ class SurfacePhysicsView
...
+    // Get surface model for the given step
     inline CELER_FUNCTION
-        SurfacePhysicsMapView surface_physics_map(SurfacePhysicsOrder,
-                                                  PhysSurfaceId) const;
+        SurfaceModelView surface_model(SubsurfaceDirection,
+                                       SurfacePhysicsOrder) const;
+
+    // Get surface model for the given step
+    inline CELER_FUNCTION SurfaceModelView
+    surface_model(Real3 const&, SurfacePhysicsOrder) const;
```
sethrj: To improve the readability of the class synopsis, 
```suggestion
    surface_model(Real3 const& normal, SurfacePhysicsOrder) const;
```

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
sethrj: Let's wait until this and Mie scattering is in (and whatever other relevant major PRs) and then reorganize.

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
```diff
@@ -0,0 +1,146 @@
...
+ */
+template<SurfacePhysicsOrder>
+struct BuiltinApplier;
+
+template<>
+struct BuiltinApplier<SurfacePhysicsOrder::roughness>
+{
+    template<class T>
+    using Applier = RoughnessApplier<T>;
+};
```
sethrj: I think you could make this the default behavior of the BuiltinApplier and eliminate the specialization classes (and the specialization class on `size_`, though you may want to specialize that one with a bogus type like `using Applier = void;`)

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,196 @@
...
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+template<>
+struct BuiltinApplier<SurfacePhysicsOrder::size_>
+{
+    template<class T>
+    using Applier = TrivialApplier<T>;
+};
```
sethrj: I'm not sure what this specialization is here for or why it's needed. Add a comment if it is?

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,196 @@
...
+
+namespace detail
+{
+namespace
+{
```
sethrj: Since it's a header file (not cc), even though it's probably going to be used only by a single .cc file,  better to avoid the `namespace {` (the primary purposes of which are to help the compiler inline more effectively, to catch errors about unused functions, and to document that the content is only ever used inside the current file)

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,196 @@
...
+    // Construct with storage to fill
+    explicit inline BuiltinSurfaceModelBuilder(std::vector<SPModel>& models);
+
+    // Construct a fake surface model
+    template<class T>
+    void build_fake(std::string_view label, std::map<PhysSurfaceId, T> const&);
+
+    // Construct a built-in surface model
+    template<class M>
+    void build(std::map<PhysSurfaceId, typename M::InputT> const&);
```
sethrj: These should also be inline if they're in the header file here

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
```diff
@@ -0,0 +1,146 @@
...
+ * surface physics steps.
+ */
+template<SurfacePhysicsOrder S>
+class BuiltinSurfaceModel : public SurfaceModel
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using Applier = typename BuiltinApplier<S>::template Applier<T>;
```
sethrj: The templates here are getting pretty complicated. I'm not sure I see what the need for a template parameter S versus a value are? (I guess it depends on how BuiltinApplier is supposed to be different based on the type?)

sethrj (commented):
Still working my way through this, and it's looking great! I'm really impressed.

src/celeritas/optical/surface/SurfacePhysicsView.hh
sethrj: I guess it's direction not normal? Regardless I think the traversal direction should be an enum (boolean) value on the state, not a property of the geometry direction.

src/celeritas/optical/surface/TrackSlotExecutor.hh
```diff
@@ -0,0 +1,64 @@
...
+{
+    SurfacePhysicsOrder step;
+    SurfaceModelId model;
+
+    //! Whether the surface model should be executed for the track
+    CELER_FUNCTION bool operator()(CoreTrackView const& track) const
+    {
+        return track.surface_physics().is_crossing_boundary()
+               && track.surface_physics()
+                          .surface_model(track.geometry().dir(), step)
```
sethrj: We don't want to have to query the geometry to determine whether the kernel should apply to the current thread...

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,196 @@
...
+
+}  // namespace
+
+//---------------------------------------------------------------------------//
+/*!
+ * Utility for building built-in surface models from input data.
+ *
+ * Wraps the call to build a model with a check on whether the input data is
+ * empty. If empty, then the model is not built. Keeps track of number of
+ * models built and constructs new models with the next ID.
```
sethrj, line 63: Another important function is it unzips the `<surface ID, input>` map into two separate arrays: a vector of surface IDs (sent to the base class) and a vector of per-surface inputs (kept in the daughter).

src/celeritas/optical/surface/model/GaussianRoughnessModel.cc
```diff
@@ -0,0 +1,73 @@
...
+ * Construct model from surfaces and inputs.
+ */
+GaussianRoughnessModel::GaussianRoughnessModel(
+    SurfaceModelId model,
+    std::vector<PhysSurfaceId> surfaces,
+    std::vector<InputT> const& inputs)
+    : BuiltinRoughnessModel(model, "gaussian", std::move(surfaces))
+{
+    HostVal<GaussianRoughnessData> data;
+    auto build_sigma_alpha = ::celeritas::make_builder(&data.sigma_alpha);
```
sethrj: This should work now that we use C++17:
```suggestion
    auto build_sigma_alpha = CollectionBuilder{&data.sigma_alpha};
```

src/celeritas/optical/surface/model/GaussianRoughnessModel.cc
```diff
@@ -0,0 +1,73 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construct model from surfaces and inputs.
+ */
+GaussianRoughnessModel::GaussianRoughnessModel(
+    SurfaceModelId model,
+    std::vector<PhysSurfaceId> surfaces,
+    std::vector<InputT> const& inputs)
+    : BuiltinRoughnessModel(model, "gaussian", std::move(surfaces))
```
sethrj: This shows some of the danger of splitting data into the base protected class and daughter, and relying on a third class to unzip an input map (!) because now we have the implementation details in three places, and there's no check that the surfaces size (`BuiltinSurfaceModel::get_surfaces()`) is equal to the inputs size. I think it would be better to keep the map input until it's given to this class, and *then* have a helper function split out the surface IDs (keys) from the input (values).

src/celeritas/optical/surface/SurfaceModelView.hh
hhollenb: Maybe `PerModelSurfaceId`?

src/celeritas/optical/surface/SurfacePhysicsView.hh
hhollenb: Because the traversal direction changes whenever a surface interaction occurs, we always call the function with
```
auto surface_physics = track.surface_physics();
auto surface_model = surface_physics.surface_model(surface_physics.traversal_direction(track.geometry().dir()), step);
```
so I collapsed it down to
```
auto surface_model = track.surface_physics().surface_model(track.geometry().dir(), step);
```
but I can go back and expand it again for readability.

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
hhollenb: Moved this functionality into the model's constructs as per your other comment.

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
hhollenb: This code should get removed once we have the other builtin models implemented, so I wanted to make sure it didn't accidentally get used elsewhere.

hhollenb:
Added some changes to simplify the applier code, use maps as inputs to models, and cache the track traversal direction. This should remove most of the ambiguities about whether vectors refer to the surface normal or track direction in function arguments, and also remove the need to query geometry in the track slot executors. The traversal direction will need to be updated after a reflection / refraction in the interaction step which will be implemented later.

src/celeritas/optical/surface/model/BuiltinSurfaceModel.hh
```diff
@@ -0,0 +1,123 @@
...
+/*!
+ * Templated base class for built-in optical surface physics models.
+ *
+ * Built-in surface physics models share a common format for their input data
+ * as well as the parameters to pass to their kernel launchers. The \c
+ * Applier wrappers executor calls to factor out common behavior between
+ * surface physics steps.
+ */
+template<SurfacePhysicsOrder S, template<class> class Applier>
+class BuiltinSurfaceModel : public SurfaceModel
```
sethrj: I think this is too much complexity for incomplete functionality (i.e., we have to inherit from it in the Roughness model, so we end up with an extra model class that doesn't do anything on its own: it's not an interface, it's an implementation).

We should choose to:
1. Continue like we've done with the other actions in the code and have a bunch of similar-looking code (~50% looks the same but it's spread among many different lines)
2. Commit fully to a template+traits approach: each traits instance will have surface physics order, input type, "internal surface" record type, input-to-record builder class, (likely another params-type type for "backend storage"), and executor.

For this PR, is it at all possible to do the first one, including only a single model, and using the current workaround for the "fake" processes? Then we can generalize in the next PR.

sethrj (commented):
@hhollenb This really is great work, but to make it easier to review we need to try to split these into smaller requests. I think there's a quadratic scaling with PR review time...

Still working on finishing this up but we need to decide about this `Builtin` thing first. (Or maybe we should just change that in a follow-on...)

sethrj:
@hhollenb @amandalund Notes from today, which should be incorporated into documentation once stable:

# Physics surface crossing state

Logical directions:
- traversal direction: forward and reverse, based on pre -> post volume crossing
- orientation: whether we're going from low -> high subsurface IDs, or high ->
  low

Physical directions:
- geometry direction (updated during tracking at end of substep)
- polarization (also updated at end of substep)
- geometry normal is constant throguhout the surface traversal and cached, with
  sign normalized to point from post into pre
- Local 

Changing during traversal:
- surface position
- facet normal

Cached:
- geometry normal
- pre and post volume material

# Executor/applier/etc diagram

This is the best I could come up with since there's a mix of functional programming ($`f(x; p) \to y`$) and procedural programming ($`f(), g(), h()`$) that gets mixed together at the Applier level, as well as a mix of objects with different levels of "temporariness" (executor/applier/trackexecutor):

surface physics

The red bold arrows are return values, dashed arrows are "creates during call operator", black arrows are data passing. Construction arguments are to the left, call arguments to the right. I'll create an equivalent for the EM interactor for comparison... maybe that'll help inform our design choices.

src/celeritas/optical/surface/model/RoughnessApplier.hh
```diff
@@ -0,0 +1,69 @@
...
+    auto s_physics = track.surface_physics();
+    auto model_view = s_physics.surface_model(SurfacePhysicsOrder::roughness);
+    auto rng = track.rng();
+
+    // Ensure the local normal follows the entering surface convention
+    auto normal = s_physics.global_normal();
+    if (s_physics.traversal_direction() == SubsurfaceDirection::reverse)
+    {
+        normal = -normal;
+    }
```
sethrj: Should this be done when entering a surface, reflecting, or here? It doesn't seem related to the roughness per se.

sethrj:
For comparison, here's a typical EM interactor (bethe-heitler):
EM helper nesting

In the EM case, there's a clean break between the procedural components and functional: the executor doesn't return an interactor, it *executes* the interaction. I actually think philosophically the choice @hhollenb is nicer. Instead of building the executor and, embarrassingly, calling it only once:
```c++
    BetheHeitlerInteractor interact(
        params, particle, dir, allocate_secondaries, material, element);

    auto rng = track.rng();
    return interact(rng);
```
you would construct the interactor and let someone else handle it:
```c++
    return BetheHeitlerInteractor{...};
```
However, this *isn't* an executor since it's not really executing (though in the EM case it could also be sampling the element!). The "Executor" misnomer is one issue causing confusion in this PR.

I think the other complication is that forwarding and combining arguments can be confusing, especially if nested. That's done a few mote times here than in EM: EnteringSurfaceNormalSampler, BSM::make_executor in particular. And the templates on BSM make that even more confusing.

Finally, the roughness "executor" takes a few select track states rather than the whole track view. Even though this is the "correct" thing to do (pass around only the data you need, not all the data you could possibly want, in order to make the data flow and operations clearer) I think it adds another layer of cognitive burden.

So I think this design is good at heart, but to improve clarity we should maybe:
- Rename the Executor to RoughnessSampler{Builder/Factory/...} if its job is to return a functor  that samples given an RNG. 
- Consider having `RoughnessApplier` do more, or maybe eliminating it/combining it as a helper in the executor rather than the other way around.

sethrj:
I think this has been superseded by #2006, #2008, #2020, #1984.

sethrj:
Never mind, looks like only the polished was added.

sethrj (approved):
Thanks for adding those @hhollenb ! I hope the conflict resolution in 52f10bdd526d4fd6a961410cddc64b0372b06506 is to your satisfaction.

whokion (commented):
I do not have any specific line-by-line comments, but I have a few general comments:
1) The code structure of the two model classes (GaussianRoughnessModel and SmearRoughnessModel) is almost identical, except one data parameter (sigma_alpha vs. roughness). This appears somewhat structrally redundant and could potentially be further consolidated.
2) The class naming, particularly the term “Roughness”, is somewhat confusing and may not be the best choice. Names such as GaussianSmearingModel and CubicSmearingModel might better reflect the actual behavior of surface or UnifiedSmearingModel and GlisurSmearingModel to attribute origin of the models. Additionally, PolishedRoughnessModel is somewhat odd, as it represents the no-smearing (roughness) case.
3) It is unclear how much validation has been performed on the sampling results, especially for GaussianRoughnessModel as the sampling logic has been already refactored. I performed a rough validation of the final surface normal with a specific setup (included below), while I cannot say which result is more physically accurate, the results differ between Geant4 and Celeritas driven by the max of the sampled alpha.

Image

sethrj:
Thanks @whokion , those are interesting results. Can you please explain your methodology and post code for the celeritas/g4 cases? The "clipping" you see in Celeritas might be implemented elsewhere in the Geant4 surface physics (some other outer rejection loop; it's hard to tell). I think we might need to validate the overall roughness model with a virtual experimental setup using a dielectric-metal interface and 100% specular reflection.

Regarding the name "roughness": we all agreed on this name six months ago, so there would have to be a very strong reason to change it at this point. What is confusing about it? We're sampling a microfacet normal using a parameterized distribution of the roughness of the surface.

I agree the code can use simplification: this is just the first step, completing the work @hhollenb started. In a follow-on PR I'll do a bit of simplification, but we still can't require exactly a single user input parameter, because there are infinite possible representations of surface roughness. For example, instead of approximating the facet slopes as a gaussian (fig 5, levin-morephysical-1996, https://doi.org/10.1109/NSSMIC.1996.591410):
Screenshot 2025-12-24 at 07 22 59
we could store the measured slopes as a discrete PDF and sample from that. Such a distribution wouldn't have a single scalar describing the surface roughness.

whokion:
I added a simple ROOT comparison macro below. As noted earlier, the discrepancy simply comes from the refactoring of the alpha sampling. Regarding the naming, sorry that I missed the earlier discussion - I am just wondering how these relatively simple sampling functions turned into “models.” For example, PolishedRoughnessModel can hardly be considered a model,
as it effectively performs no meaningful operation. To me, a roughness model should describe how roughness is represented physically, rather than how the degree of roughness is smeared via a particular sampling method, which could vary independently. I do not want to argue further if this naming and modeding have already been agreed upon
[facet_normal_unified.C](https://github.com/user-attachments/files/24330778/facet_normal_unified.C)
.

sethrj:
"Polished" is a model of a surface's texture, as are all the other mechanisms of approximating the local surface normal using the geometric. The mechanism is general enough that we could define anisotropic or even spatially dependent microstructure.

sethrj:
PS thanks for the script and have a merry Christmas!

sethrj:
Happy end of 2025 @whokion ! I'm looking at your code and comment again. I think your analysis belongs on #1896 when we added the roughness sampling a few months ago. The differences are probably due to my changes: https://github.com/celeritas-project/celeritas/pull/1896/changes/91ece1c889b99b035ce64c6a9235a7a6d8617423..c876fc1d77f200fa7d5ad006a013e48d5d243009 . I'm not sure I understand your plot for alpha: why are they both between 0.6 and 0.8? alpha is an angle that should be between 0 and pi/2 radians, isn't it? Could you please look at the [gaussian roughness documentation](https://celeritas-project.github.io/celeritas/user/implementation/optical-physics.html#_CPPv4N9celeritas7optical24GaussianRoughnessSamplerE) and tell me if I have a conceptual misunderstanding?

whokion:
As I understand it, $\alpha$ is the normal distribution with the standard deviation $\sigma_{\alpha}$.  The 0.8 cut in celeritas is due to the first sampling condition,
```
alpha_max_(fmin(real_type(constants::pi / 2), 4 * sigma_alpha))
...
do
{
   alpha = std::fabs(sample_alpha_(rng));
} while (alpha >= alpha_max_);
```
where $|\alpha| < 4 \sigma_{\alpha} = 4 \times 0.2 = 0.8$ in the example test case.  The lower bound of 0.6 is from the final condition that the dot product between the sampled angle rotated along the original (momentum) direction and the surface normal should be negative. I was not trying to argue which sampling model is more physically correct, but simply wanted to point out that the final sampling result differs between Geant4 and celeritas, which will lead to differences in subsequent validation with optical boundary processes.

hhollenb:
Looks like we just got the `alpha_max` mixed up in `GaussianRoughnessSampler`. We want a restricted sampling of a Gaussian in the interval $$\alpha \in [0, \pi/2]$$, which is then fed into a rejection sampling for $$\sin\alpha$$. The calculator should look something like
```
f_max_ = fmin(real_type{1}, 4 * sigma_alpha);

real_type cos_alpha{};
real_type sin_alpha{};
do
{
  real_type alpha{};
  do
  {
    alpha = std::fabs(sample_alpha_(rng));
  } while (alpha >= constants::pi / 2);
  sincos(alpha, &sin_alpha, &cos_alpha);
} while (RejectionSampler{sin_alpha, f_max_}(rng));

return ExitingDirectionSampler{cos_alpha, normal_}(rng);
```

Changing the initializations in Soon's plotting script to:
```
const Double_t cel_alpha_max = TMath::PiOver2();
const Double_t cel_f_max = std::fmin(1, 4 * sigma_alpha);
```
Gives pretty good agreement:
fixed_gaussian_sample_comparison

sethrj:
@hhollenb Should we include that fix on this branch, and we can both review the final work?

test/celeritas/optical/RoughnessCalculator.test.cc
```diff
@@ -149,11 +149,11 @@ TEST_F(RoughnessSamplerTest, gaussian)
     }
 
     static SampledHistogram const expected[] = {
-        {{0, 0, 0, 0, 5}, 21.7884},
-        {{0, 0, 0, 0, 5}, 21.9014},
-        {{0, 0, 0, 0, 5}, 21.9502},
-        {{0, 0, 0, 0.034, 4.966}, 20.192},
-        {{0.0105, 0.051, 0.235, 0.971, 3.7325}, 15.1376},
+        {{0, 0, 0, 0, 5}, 507.9478},
```
sethrj: @hhollenb This change results from not truncating the gaussian to 4 sigma: it means it takes on average 500 throws per sample (!) for smooth surfaces. I thought geant4 had that limit built in as well.

The smaller the sigma, the more the rejection loop costs, and the less of a difference in the facet slope it makes. Maybe we could do 5 sigma?

hhollenb: Yeah I saw that and I'm looking into it, I think I may have found my logic issue (I'm not great at rewriting rejection samplers with all the double negatives oops)

hhollenb:
To relitigate the `f_max_` issue, the old error you were seeing was that $$4 \sigma_\alpha$$ is sometimes less than $$\sin\alpha$$ for values of $$\alpha$$ near $$\pi/2$$. The `RejectionSampler` doesn't like that because it always accepts values of $$\alpha$$ above $$\sin^{-1} (4\sigma_\alpha)$$. Thus we can just add a check that breaks the loop (always accept) when $$\sin\alpha >= f_{max}$$ and otherwise does rejection sampling when $$\sin\alpha \in [0, f_{max}]$$ as required

test/celeritas/optical/RoughnessCalculator.test.cc
sethrj: I don't think it's an error in your code, I think it's just skipping an optimization that's surprisingly necessary based on how the math turns out.

sethrj:
@hhollenb The coverage on this branch is pretty low; is there a reasonable way to test execution of the models?

sethrj:
Actually from my work with #2184 it does looks like we're missing tests of most optical surface models, aside from importing the data. We can do that separately.

hhollenb:
Ok I'm getting a very weird issue with debug statements and the tests. The distributions appear to be different when debugging is turned off, so I added some more checks and found a possible bug with the `GaussianRoughnessSampler` where it samples a surface normal in the opposite direction to the original normal. The behavior is different with and without the `CELER_ASSERT` lines which is odd:

Without the `CELER_ASSERT(cos_alpha >= 0);` line:
```
[----------] 1 test from SurfacePhysicsIntegrationGaussianTest
[ RUN      ] SurfacePhysicsIntegrationGaussianTest.gaussian
info: Constructed 2 optical materials with 2 present in the geometry
status: Celeritas core setup complete
info: Skipping scintillation process: no scintillating materials are present
status: Celeritas optical setup complete
status: Initialized Celeritas optical state
/home/celeritas/src/src/corecel/sys/MultiExceptionHandler.cc:91: critical: Suppressed exception from parallel thread: /home/celeritas/src/src/celeritas/optical/surface/SurfacePhysicsTrackView.hh:240:
celeritas: precondition failed: dot_product(normal, this->global_normal()) >= 0
/home/celeritas/src/src/corecel/sys/MultiExceptionHandler.cc:118: warning: Suppressed 1 similar exceptions
unknown file: Failure
C++ exception with description "/home/celeritas/src/src/celeritas/optical/surface/SurfacePhysicsTrackView.hh:240:
celeritas: precondition failed: dot_product(normal, this->global_normal()) >= 0" thrown in the test body.

Writing diagnostic output because test failed
info: Wrote output to surfacephysicsintegrationgaussian-gaussian-1.out.json
[  FAILED  ] SurfacePhysicsIntegrationGaussianTest.gaussian (3 ms)

```

With the `CELER_ASSERT(cos_alpha >= 0);` line:
```
[----------] 1 test from SurfacePhysicsIntegrationGaussianTest
[ RUN      ] SurfacePhysicsIntegrationGaussianTest.gaussian
info: Constructed 2 optical materials with 2 present in the geometry
status: Celeritas core setup complete
info: Skipping scintillation process: no scintillating materials are present
status: Celeritas optical setup complete
status: Initialized Celeritas optical state
[       OK ] SurfacePhysicsIntegrationGaussianTest.gaussian (4 ms)
[----------] 1 test from SurfacePhysicsIntegrationGaussianTest (4 ms total)
```

Not sure why adding a check causes it to pass, maybe there's some weird inlining / compiler optimization going on that can't be done with an explicit usage before the return statement? You would've thought my quantum mechanics course covered how observing variables changed their state...

sethrj:
@hhollenb I think there may be a use-before-initialize error somewhere?
```
/__w/celeritas/celeritas/src/corecel/sys/MultiExceptionHandler.cc:91: critical: Suppressed exception from parallel thread: runtime error: failed to sample a microfacet direction into which {0, 1, 0} is entering (last normal sampled: {-nan, -nan, -nan})
/__w/celeritas/celeritas/src/celeritas/optical/surface/SurfacePhysicsUtils.hh:111: 'loop_guard > 0' failed
/__w/celeritas/celeritas/src/corecel/sys/MultiExceptionHandler.cc:118: warning: Suppressed 10 similar exceptions
unknown file: Failure
C++ exception with description "runtime error: failed to sample a microfacet direction into which {0, 1, 0} is entering (last normal sampled: {-nan, -nan, -nan})
/__w/celeritas/celeritas/src/celeritas/optical/surface/SurfacePhysicsUtils.hh:111: 'loop_guard > 0' failed" thrown in the test body.
```

Alas that we can't build with Geant4 in the asanlite build 😢

PR #1929, opened 2025-09-02
Improve version metadata for built-in packages and VecGeom

sethrj:
- Exports and saves versions from "vendored" packages (e.g. G4VG)
- Correctly sets the `VecGeom_NAV` variable (either 'index' or 'path') for VecGeom 1.x
- Print vecgeom versioning information at the beginning of tests to help with debugging

### DIFF
--- a/cmake/FindVecGeom.cmake
+++ b/cmake/FindVecGeom.cmake
@@ -79,4 +79,13 @@ if(VecGeom_FOUND AND VecGeom_CUDA_FOUND AND (TARGET VecGeom::vecgeomcuda OR TARG
   endif()
 endif()
 
+# Save VecGeom_NAV variable for diagnostics
+if(VecGeom_VERSION VERSION_LESS 2.0)
+  if(VECGEOM_USE_NAVINDEX)
+    set(VecGeom_NAV "index")
+  else()
+    set(VecGeom_NAV "path")
+  endif()
+endif()
+
 #-----------------------------------------------------------------------------#
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -14,7 +14,7 @@ macro(celer_fetch_external name version url sha)
   set(${name}_VERSION ${version} PARENT_SCOPE)
   set(${name}_VERSION_STRING "${version}+builtin" PARENT_SCOPE)
 
-  string(REGEX REPLACE "VERSION" "${version}" _url "${url}")
+  string(REGEX REPLACE "@VERSION@" "${version}" _url "${url}")
   FetchContent_Declare(
     "${name}"
     URL "${_url}"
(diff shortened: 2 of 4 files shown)


external/CMakeLists.txt
```diff
@@ -7,12 +7,17 @@
...
-# Download the package and add to a list
+# Download the package, save version, add to a list
 set(_celer_builtin_packages)
-macro(celer_fetch_external name url sha)
+macro(celer_fetch_external name version url sha)
+  # Save version and annotated string for corecel/Config.cc include
+  set(${name}_VERSION ${version} PARENT_SCOPE)
+  set(${name}_VERSION_STRING "${version}+builtin" PARENT_SCOPE)
+
+  string(REGEX REPLACE "VERSION" "${version}" _url "${url}")
```
sethrj: There might be a better way to do this... I couldn't think of a suitable substitution expression which made it clear that the function would update the URL as part of a template.

cmake/FindVecGeom.cmake
```diff
@@ -79,4 +79,13 @@ if(VecGeom_FOUND AND VecGeom_CUDA_FOUND AND (TARGET VecGeom::vecgeomcuda OR TARG
...
 endif()
 
+# Save VecGeom_NAV variable for diagnostics
+if(VecGeom_VERSION VERSION_LESS 2.0)
+  if(VECGEOM_USE_NAVINDEX)
+    set(VecGeom_NAV "index")
+  else()
+    set(VecGeom_NAV "path")
+  endif()
+endif()
```
pcanal, line 89:
```suggestion
  endif()
else()
  set(VecGeom_NAV "v2 index")
endif()
```
for consistency (not having `VecGeom_NAV` might other means either that this build is using v2 or that this mechanism failed) and it will help prevent confusing if an adventurous user is switching the same Celeritas build area between VecGeom v1 and v2 ...

external/CMakeLists.txt
pcanal: We could use the `configure_file` syntax:
```suggestion
  string(REGEX REPLACE "@VERSION@" "${version}" _url "${url}")
```
and the use would look like:
```
CLI11 2.5.0
    "https://github.com/CLIUtils/CLI11/archive/refs/tags/v@VERSION@.tar.gz"
```

test/geocel/vg/Vecgeom.test.cc
```diff
@@ -77,6 +78,10 @@ class VecgeomVgdmlTestBase : public VecgeomTestBaseImpl
   public:
     SPConstGeo build_geometry() const final
     {
+        using namespace celeritas::cmake;
+        cout << color_code('x') << "VecGeom v" << vecgeom_version << " ("
+             << vecgeom_options << ") VGDML" << color_code(' ') << endl;
```
pcanal:
```suggestion
             << vecgeom_options << ") using VGDML" << color_code(' ') << endl;
```
to follow the pattern of the other message.

pcanal (approved):
LGTM.

cmake/FindVecGeom.cmake
sethrj: Actually this is deliberately here because I was doing that. VecGeom 2 stores `VecGeom_NAV` as a cache variable, and it's still exported by v1 even though it's not defined anywhere in the CMakeLists, leading to an inconsistent diagnostic downstream in Celeritas. By forcibly providing the navigation type here we can consistently tell whether we're using the index, path, or tuple nav state regardless of the vecgeom version.

external/CMakeLists.txt
sethrj: Oh I like that. On it!

test/geocel/vg/Vecgeom.test.cc
sethrj: Yep that was a typo.

PR #1930, opened 2025-09-02
Impose minimum size to split leaf nodes during BIH construction

elliottbiondo:
This MR introduces a `min_split_size` parameter to BIH construction. This parameter determines the minimum number of bboxes on a given node in order to split the node, rather than create a leaf. Upcoming performance testing will determine a suitable value for this parameter. In addition to testing the `min_split_size` modifies tree construction in the proper way, test for the two traversal classes, `BIHIntersectingVolFinder` and  `BIHEnclosingVolFinder` are also modified to ensure that the tree configuration does not affect tracking.


src/orange/detail/BIHBuilder.hh
```diff
@@ -25,17 +25,20 @@ namespace detail
 /*!
  * Create a bounding interval hierarchy from supplied bounding boxes.
  *
- * This implementation matches the structure proposed in the original
- * paper \citep{wachter-bih-2006, https://doi.org/10.2312/EGWR/EGSR06/139-149}.
+ * This implementation matches the structure proposed in the original paper
+ * \citep{wachter-bih-2006, https://doi.org/10.2312/EGWR/EGSR06/139-149}.
+ * Construction is done recursively. At a given recursion depth, if at least
+ * \param max_split_size boumding boxes are present, partitioning is attempted.
```
sethrj:
```suggestion
 * \param max_split_size bounding boxes are present, partitioning is attempted.
```

src/orange/detail/BIHBuilder.hh
```diff
@@ -25,17 +25,20 @@ namespace detail
...
  * Partitioning is done on the basis of bounding box centers using the "longest
- * dimension" heuristic. All leaf nodes contain either a single volume id, or
- * multiple volume ids if the volumes have bounding boxes that share the same
- * center. A tree may consist of a single leaf node if the tree contains only 1
- * volume, or multiple non-partitionable volumes. In the event that all
- * bounding boxes are infinite, the tree will consist of a single empty leaf
- * node with all volumes in the stored inf_vols. This final case is useful in
- * the event that an ORANGE geometry is created via a method where volume
- * bounding boxes are not available.
+ * dimension" heuristic.  If more than \param max_split_size boumding boxes are
```
sethrj:
```suggestion
 * dimension" heuristic.  If more than \param max_split_size bounding boxes are
```

src/orange/detail/BIHBuilder.hh
```diff
@@ -25,17 +25,20 @@ namespace detail
...
+ * present, or if the bounding boxes are are non-partitionable (i.e., they all
+ * have the same center), a leaf node is created.
+ *
+ * Any bounding boxes that have at least one infinite dimension are stored in
+ * inf_vols. In the event that all bounding boxes are infinite, the tree will
```
sethrj, line 38: I think that propagating the universe boundary sizes meant we don't have to worry about infinities except in the global unit?

src/orange/detail/BIHBuilder.hh
```diff
@@ -53,11 +56,11 @@ class BIHBuilder
     //!@}
 
   public:
-    // Construct from a Storage object
-    explicit BIHBuilder(Storage* storage);
+    // Construct from a Storage object and a minimum split size.
+    explicit BIHBuilder(Storage* storage, size_type min_split_size = 2);
```
sethrj: Since there might be additional input parameters, it might be good to define an `Input` struct for them. Also, avoid `explicit` with default parameters, because it breaks some construction paradigms (not an issue here, but if you wanted to do BihBuilder a; a = {&storage, 3};` you'd get an error that the constructor needs to be explicit.)

test/orange/detail/BIHBuilder.test.cc
```diff
@@ -472,6 +482,193 @@ TEST_F(BIHBuilderTest, grid)
...
+
+    bboxes_.push_back({{1, 0, 0}, {2, 1, 100}});
+    bboxes_.push_back({{1, 1, 0}, {2, 2, 100}});
+    bboxes_.push_back({{1, 2, 0}, {2, 3, 100}});
+    bboxes_.push_back({{1, 3, 0}, {2, 4, 100}});
+
+    bboxes_.push_back({{2, 0, 0}, {3, 1, 100}});
+    bboxes_.push_back({{2, 1, 0}, {3, 2, 100}});
+    bboxes_.push_back({{2, 2, 0}, {3, 3, 100}});
+    bboxes_.push_back({{2, 3, 0}, {3, 4, 100}});
```
sethrj:
```suggestion
    bboxes_ = {
        FastBBox::from_infinite(),
        {{0, 0, 0}, {1, 1, 100}},
        {{0, 1, 0}, {1, 2, 100}},
        {{0, 2, 0}, {1, 3, 100}},
        {{0, 3, 0}, {1, 4, 100}},

        {{1, 0, 0}, {2, 1, 100}},
        {{1, 1, 0}, {2, 2, 100}},
        {{1, 2, 0}, {2, 3, 100}},
        {{1, 3, 0}, {2, 4, 100}},

        {{2, 0, 0}, {3, 1, 100}},
        {{2, 1, 0}, {3, 2, 100}},
        {{2, 2, 0}, {3, 3, 100}},
        {{2, 3, 0}, {3, 4, 100}},
    };
```

test/orange/detail/BIHEnclosingVolFinder.test.cc
```diff
@@ -47,90 +47,112 @@ class BIHEnclosingVolFinderTest : public Test
...
-    ref_storage_ = storage_;
-    BIHEnclosingVolFinder find_volume(bih_tree, ref_storage_);
+    auto run_test = [&](size_type min_split_size) {
+        bboxes_.push_back(FastBBox::from_infinite());
+
+        bboxes_.push_back({{0, 0, 0}, {1.6f, 1, 100}});
+        bboxes_.push_back({{1.2f, 0, 0}, {2.8f, 1, 100}});
+        bboxes_.push_back({{2.8f, 0, 0}, {5, 1, 100}});
+        bboxes_.push_back({{0, -1, 0}, {5, 0, 100}});
+        bboxes_.push_back({{0, -1, 0}, {5, 0, 100}});
```
sethrj: Same here (also it makes it clearer that you're expecing bboxes to be empty; otherwise if it weren't for the `move` on line 80 you'd be appending for each test)

And actually since the bboxes are consumed immediately, and the test harness doesn't use any of them, I think you should avoid making this guy member data, instead just declare `using VecFastBbox = BIHBuilder::VecBBox`

src/orange/detail/BIHBuilder.hh
```diff
@@ -76,14 +79,15 @@ class BIHBuilder
...
     //// DATA ////
-
     Temporaries temp_;
 
     CollectionBuilder<FastBBox> bboxes_;
     CollectionBuilder<LocalVolumeId> local_volume_ids_;
     CollectionBuilder<BIHInnerNode> inner_nodes_;
     CollectionBuilder<BIHLeafNode> leaf_nodes_;
 
+    size_type min_split_size_;
```
sethrj: And if your input struct is pretty simple you can just change this to `Input inp_` (or `Options opts_`) and reference the member data through that.

src/orange/detail/BIHBuilder.cc
```diff
@@ -125,10 +127,28 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
...
+        node.parent = parent;
+        node.vol_ids
+            = local_volume_ids_.insert_back(indices.begin(), indices.end());
+        CELER_EXPECT(node);
+        (*nodes)[current_index] = node;
+    };
+
+    if (indices.size() < min_split_size_)
+    {
+        // All bboxess fit on a single leaf; make it and exit early
```
sethrj:
```suggestion
        // All bboxes fit on a single leaf; make it and exit early
```

src/orange/detail/BIHBuilder.cc
```diff
@@ -125,10 +127,28 @@ void BIHBuilder::construct_tree(VecIndices const& indices,
     auto current_index = nodes->size();
     nodes->resize(nodes->size() + 1);
 
+    // Create a single leaf containing all bboxes
```
sethrj: I'd make it clearer the lambda is used only once per `construct_tree` call, rather than multiple leaves inside hte single function.

sethrj (approved):
A few minor comments. Thanks!

src/orange/detail/BIHBuilder.hh
elliottbiondo: Yep, looks like that is done in the constuctor of `IntersectSurfaceBuilder`

elliottbiondo: I think it's still worth supporting though right? Should I just update the comment?

elliottbiondo: I updated the comment to read:
```
 * This final case should not occur unless an ORANGE geometry is created via
 * a method where volume bounding boxes are not available.
 ```
 I can make the statement stronger; but I don't have a good enough sense off the top of my head how far up/down `IntersectSurfaceBuilder` is in the construction process, i.e., how unlikely would it be that we would have a construction pathway in which bboxes are not available.

elliottbiondo:

sethrj (approved):

PR #1933, opened 2025-09-03
Implement Mie scattering 

Rashika-Gupta:
This PR adds support for optical Mie scattering, including data import, model implementation, and unit tests in the following way :

1. Introduce `ImportMie` struct – defines forward/backward asymmetry parameters, ratio.
2. Extend `ImportOpticalMaterial` – add a `mie` field so each optical material can store Mie scattering parameters.
3. Update `ImportedMaterials` – allocate storage for `mie`.
4. Add `mie` to `ImportModelClass` enum – so Mie appears alongside absorption, Rayleigh, WLS, etc.
5. Support `G4OpMieHG` in `GeantImporter` – detect and import forward/backward g and forward ratio from Geant4 materials.
6. Extend `GeantOpticalModelImporter` – allow `IMC::mie` mapping to `MIEHG MFPs`.
7. Implement `MieData` struct – store per-material Mie parameters in `device/host collections`.
8. Add `MieInteractor `– physics sampler that applies  scattering using forward/backward weighted scattering parameters.
9. Introduce `MieExecutor` – per-track executor that sets up and runs the `MieInteractor`.
10. Implement `MieModel `– full model with constructor (building data), `build_mfps`, and step for applying scattering.
11. Integrate into `ModelImporter` and build system – register `Mie model_builder`. 
12. update `CMakeLists.txt` to compile.
13. Add unit tests – extend `OpticalMockTestBase` with mock `Mie data`
14. Add `Mie.test.cc` covering parameters, basic scattering, and MFP handling.


src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -155,6 +155,26 @@ struct ImportWavelengthShift
...
+/*!
+ * Store Mie scattering properties (Henyey–Greenstein model).
+ */
+struct ImportMie
+{
+    double forward_g{};  //!< Henyey–Greenstein "g" parameter for forward
+                         //!< scattering
+    double backward_g{};  //!< Henyey–Greenstein "g" parameter for backward
+                          //!< scattering
+    double forward_ratio{};  //!< Fraction of forward vs backward scattering
```
sethrj: Easier to read (and won't unintentionally causeROOT to ignore the variable)
```suggestion
    //! Henyey–Greenstein "g" parameter for forward scattering
    double forward_g{};
    //! Henyey–Greenstein "g" parameter for backward scattering
    double backward_g{};
    //! Fraction of forward vs backward scattering
    double forward_ratio{};
```

src/celeritas/optical/MieData.hh
```diff
@@ -0,0 +1,63 @@
...
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using OpticalMaterialItems = Collection<T, W, M, OptMatId>;
+
+    //// MEMBER DATA ////
+
+    OpticalMaterialItems<MieMaterialData> mie_record;
+
+    // Backend data
+    Items<real_type> reals;
```
sethrj: This is unused and I thikn unneeded

src/celeritas/optical/MieData.hh
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
sethrj, line 7: This file needs the copyright header; please use `scripts/dev/celeritas-gen.py` to create new files (for example, `celeritas-gen.py -n celeritas::optical MieData.hh`)

src/celeritas/optical/ModelImporter.cc
```diff
@@ -126,8 +134,17 @@ auto ModelImporter::build_wls() const -> ModelBuilder
...
     {
-        input.data.push_back(input_.import_material->wls(mid));
+        auto wls_data = input_.import_material->wls(mid);
+        if (wls_data)
+        {
+            input.data.push_back(std::move(wls_data));
+        }
+        else
+        {
+            return {};
```
sethrj: I think this logic implies that if any WLS data is missing, *all WLS will be ignored* (which is not correct). Maybe this is some leftover code that was taken care of in  #1963?

sethrj (changes requested):
Hi @Rashika-Gupta , normally we don't rebase once an MR is marked as "ready", but since this has several changes that have already been merged upstream and now conflict (mainly #1983) maybe we should? Since there are many commits on this branch it might be easier to `git merge upstream/develop` and fix the conflicts.

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+
+    UniformRealDist sample_phi(0, real_type(2 * constants::pi));
+    UniformRealDist sample_r(0, 1);
+    UniformRealDist sample_g(0, 1);
+
+    real_type u = sample_g(rng);
+    real_type r = sample_r(rng);
+
+    // Check if scattering should be forward scattering or backward scattering
+    bool is_forward = u < mie_params_.forward_ratio;
```
sethrj: Choosing between two options with a probability *p* is more cleanly expressed as a [https://celeritas-project.github.io/celeritas/user/implementation/core-physics/rng.html#_CPPv4N9celeritas21BernoulliDistributionE](bernoulli distribution):
```c++
bool is_forward = BernoulliDistribution{mie_params_.forward_ratio}(rng);
```

We also often put the distributions themselves as class data (important for when the distribution has a state, like in the normal distribution which samples two values per invocation), so you could have `BernoulliDistribution sample_forward_;` as class data, initialize it with `sample_forward_{params.forward_ratio}`, and sample here with `bool is_forward = sample_forward_(rng)`.

Also, due to the way GPUs work, it's usually very important to limit the *scope* of variables (especially real numbers which are usually 64 bits). This is also good practice in general (see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-scope ). Move your variables close to where you use them (direction, polarization, r, etc). (Copilot is useful for this kind of refactoring.) I know it'll look very different from Geant4 and will take some getting used to, but it ends up making a marked improvement in readability.

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+    real_type g = (is_forward ? mie_params_.forward_g : mie_params_.backward_g);
+
+    // Sample cos theta
+    real_type costheta
+        = (g != 0) ? (2. * r * (1. + g) * (1. + g) * (1. - g + g * r)
+                          / ((1. - g + 2. * g * r) * (1. - g + 2. * g * r))
+                      - 1.)  // will make it cleaner
+                   : (2 * r - 1);
```
sethrj: A few questions/comments (and a reminder that you maybe meant to go back to this)?
- When can `g` be zero and what's the significance of that?
- If that's unlikely but possible, does the expression simplify to 2r - 1 or is the expression wrong/nan? (I think the former, so this is a flawed attempt at microoptimization.) If that's the case then just write it as the first half of the expression
- Do not use `1.`, `2.` etc. because these are double-precision literals that "promote" single-precision arithmetic (if we're using real_type=float) which is bad for GPU.
- Use `ipow<2>(expr)` instead of `expr * expr`; it looks like you have at least two places here that use it.
- You might be able to use [https://celeritas-project.github.io/celeritas/dev/d9/d9b/classceleritas_1_1PolyEvaluator.html](PolyEvaluator)

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+    if (!is_forward)
+        costheta = -costheta;
+
+    real_type phi = sample_phi(rng);
+
+    // Creating new direction
+    new_dir = from_spherical(costheta, phi);
+
+    // Build polarization vector
+    SoftZero const soft_zero{SoftEqual{}.rel()};
```
sethrj: @amandalund updated the polarization rejection recently to use more consistent zero-checking, see #1972 .

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+        if (!BernoulliDistribution{0.5}(rng))
+        {
+            // Flip direction with 50% probability: there are two polarizations
+            // perpendicular to the new direction and the original polarization
+            new_pol = -new_pol;
+        }
+        // Accept with the probability of the scattered polarization overlap
+        // squared
+    } while (RejectionSampler{ipow<2>(
+        clamp<real_type>(dot_product(new_pol, inc_pol_), -1, 1))}(rng));
```
sethrj: This looks like a copy-paste error from Rayleigh; I don't think Mie has the same $`\sin^2 \theta`$ rejection.

src/celeritas/optical/model/MieExecutor.hh
```diff
@@ -0,0 +1,59 @@
+//---------------------------------*- C++
+//-*----------------------------------//
+// Copyright ...
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
sethrj: Formatting and copyright need fixing

src/celeritas/optical/model/MieExecutor.hh
```diff
@@ -0,0 +1,59 @@
...
+#include "corecel/Types.hh"
+#include "corecel/io/Logger.hh"
+#include "celeritas/geo/GeoFwd.hh"
+#include "celeritas/io/ImportOpticalMaterial.hh"
+#include "celeritas/optical/CoreTrackView.hh"
+#include "celeritas/optical/ImportedMaterials.hh"
+#include "celeritas/optical/Interaction.hh"
+#include "celeritas/optical/MieData.hh"
+#include "celeritas/optical/ParticleTrackView.hh"
+#include "celeritas/optical/interactor/MieInteractor.hh"
```
sethrj, line 19: Leave a blank line between the includes and the namespace at the start of the code

sethrj (commented):
Continuing the review... :)

src/celeritas/optical/interactor/MieInteractor.hh
Rashika-Gupta: > * When can `g` be zero and what's the significance of that?

Ideally it should not be equal to zero.

> * If that's unlikely but possible, does the expression simplify to 2r - 1 or is the expression wrong/nan? (I think the former, so this is a flawed attempt at microoptimization.) If that's the case then just write it as the first half of the expression

the above expression simplifies to $\cos \theta = 2r -1$ when $g = 0$

Addressed the comments, thanks !

test/celeritas/optical/Mie.test.cc
```diff
@@ -100,19 +101,19 @@ TEST_F(MieTest, mie_basic)
...
-        -0.999998838605416,
-        -0.99851580312691,
+        0.999904430863429,
+        -0.99959742953257,
+        -0.999999650643697,
+        0.996510957439599,
     };
 
-    EXPECT_EQ(40, rng_engine.count());
+    EXPECT_EQ(32, rng_engine.count());
```
sethrj: You may want to enclose the checks/expectations in an `if` statement checking for double-precision, which may be where your failures are coming from.

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,145 @@
...
+#include "celeritas/optical/MieData.hh"
+#include "celeritas/optical/ParticleTrackView.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample optical Mie scattering using the Henyey–Greenstein distribution.
```
sethrj, line 34: Was the PDF for cos theta accidentally deleted? 😕

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,145 @@
...
+    // Select forward or backward lobe
+    real_type g = (is_forward ? mie_params_.forward_g : mie_params_.backward_g);
+
+    using Linear = PolyEvaluator<real_type, 1>;
+
+    // Compute cos(theta) distribution for optical photons that undergo mie
+    // scattering
+    real_type costheta = 2 * r * ipow<2>((1 + g) / Linear(1 - g, 2 * g)(r))
+                             * Linear(1 - g, g)(r)
+                         - 1;
```
sethrj: Much more understandable :) I thought from the original equation that there were quadratic elements (-1 + 2 *r * .. + ....) but it looks like not in practice? If not then I was wrong to recommend the PolyEvaluator: it usually only makes the code clearer for second-order terms and higher.

src/celeritas/optical/model/MieModel.cc
```diff
@@ -0,0 +1,121 @@
...
+        CELER_LOG(debug)
+            << "mie model MFP not found for setting to infinity for "
+            << mat.get();
+        inp::Grid g;
+        // should change these at some point. the .x vals are coming from the
+        // opnovice example as of now
+        g.x = {1.56962, 6.19998};
+        g.y = {std::numeric_limits<real_type>::infinity(),
+               std::numeric_limits<real_type>::infinity()};
+        build(g);
```
sethrj: I assume the energy bounds are stray debugging? I think this would work:
```suggestion
        build(inp::Grid::from_constant(std::numeric_limits<double>::infinity()));
```

(or should we change the MfpBuilder to construct an infinite MFP over all energies instead of an invalid grid?)

test/celeritas/optical/Mie.test.cc
```diff
@@ -0,0 +1,141 @@
...
+
+    for (auto mat : range(OptMatId(this->num_optical_materials())))
+    {
+        model_->build_mfps(mat, builder);
+    }
+
+    EXPECT_TABLE_EQ(
+        this->import_model_by_class(ImportModelClass::mie).mfp_table,
+        storage(builder.grid_ids()));
+}
```
sethrj, line 142: Can you also add a "stress test" like Rayleigh.test.cc so that we can check the sample distribution itself?

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -172,6 +193,7 @@ struct ImportOpticalMaterial
     ImportOpticalRayleigh rayleigh;
     ImportWavelengthShift wls;
     ImportWavelengthShift wls2;
+    ImportMie mie;
```
sethrj, line 196: I think this may be causing the ROOT tests to fail somehow? Can you run `{make/ninja} update-root-test-data`?

sethrj (commented):
Thanks @Rashika-Gupta !

src/celeritas/optical/interactor/MieInteractor.hh
Rashika-Gupta: eventually it will be as there is a 'r' term in the beginning. $\cos\theta = \frac{2r(1+g)^2(1-g+gr)}{(1-g+2gr)^2}-1$

src/celeritas/optical/model/MieModel.cc
Rashika-Gupta: It is for debugging purpose. I thought it will be good to handle the `else` condition but was not sure of a cleaner way to get the `x` values- hence hardcoded.

src/celeritas/optical/interactor/MieInteractor.hh
Rashika-Gupta: oh yes! 😑
let me add that back

src/celeritas/io/ImportOpticalMaterial.hh
Rashika-Gupta: After running the above command it returns 
```cpp
status: Initializing Geant4 run manager
status: Initializing Geant4 geometry and physics list
info: Loading Geant4 geometry from GDML at /Users/r1i/Desktop/project/forked/celeritas/test/geocel/data/simple-cms.gdml
status: Building Geant4 physics tables
/Users/r1i/Desktop/project/forked/celeritas/src/celeritas/ext/GeantImporter.cc:561: warning: Optical materials were requested but none are present
info: Trimming data for testing
info: Creating ROOT file at /Users/r1i/Desktop/project/forked/celeritas/test/celeritas/data/simple-cms.root
ROOT: warning: TStreamerInfo::Build: celeritas::ImportOpticalMaterial: celeritas::ImportMie has no streamer or dictionary, data member "mie" will not be saved
```
to which i have added 
`#pragma link C++ class celeritas::ImportMie+;` in `celeritas/ext/RootInterfaceLinkDef.h` and on re-building, it does not give the ROOT warning anymore.

src/celeritas/optical/model/MieModel.cc
sethrj: I agree, just like the other models where there's "incomplete" cross sections. I think as long as `x` encompasses the energy range of the optical photons we care about, we'll be ok: hence the "constant" (which uses `[0, inf)`).

src/celeritas/io/ImportOpticalMaterial.hh
sethrj: Perfect!

sethrj:
@Rashika-Gupta one of the failures (with interpolator) is that the use of infinity/zero for the energy boundaries is breaking the mfp calculator. I'm going to submit a PR that fixes this by allowing an empty grid to signify an infinite MFP, so instead of `build_mfp(Grid::from_constant(...))` you can just `build_mfp()`.

src/celeritas/optical/interactor/MieInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+#include "celeritas/optical/ParticleTrackView.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample optical Mie scattering using the Henyey–Greenstein distribution
+ *\f[
```
sethrj:
```suggestion
 * Sample optical Mie scattering using the Henyey–Greenstein distribution.
 * 
 *
 *\f[
```

sethrj: Which reminds me, please add the interactor to the optical physics docs, and make the following change to replace the existing "models" with the interactors:
```diff
diff --git a/doc/implementation/optical-physics.rst b/doc/implementation/optical-physics.rst
index b83105f02..6e5d84010 100644
--- a/doc/implementation/optical-physics.rst
+++ b/doc/implementation/optical-physics.rst
@@ -87,8 +87,9 @@ Volumetric processes
 Like other particles, optical photons undergo stochastic interactions inside
 optical materials.

-.. doxygenclass:: celeritas::optical::AbsorptionModel
-.. doxygenclass:: celeritas::optical::RayleighModel
+.. doxygenclass:: celeritas::optical::AbsorptionInteractor
+.. doxygenclass:: celeritas::optical::RayleighInteractor
+
 .. doxygenclass:: celeritas::optical::RayleighMfpCalculator

 .. _surface_processes:
```
(and also please add the WLS interactors!)

src/celeritas/optical/model/MieModel.cc
```diff
@@ -0,0 +1,114 @@
...
+{
+    CELER_EXPECT(mat < imported_.num_materials());
+
+    if (auto const& mfp = imported_.mfp(mat))
+    {
+        build(mfp);
+    }
+    else
+    {
+        build(inp::Grid::from_constant(1e7));
```
sethrj: Verify consistency with input data: if the MFP are there, check g and ratio are in bounds using the records accessible by `this->host_ref().mie_data[mat]`.

sethrj (approved):
Just a few last changes, and we can merge!

src/celeritas/optical/model/MieModel.cc
```diff
@@ -0,0 +1,125 @@
...
+    if (auto const& mfp = imported_.mfp(mat))
+    {
+        if (!mie_data)
+        {
+            CELER_VALIDATE(mie_data,
+                           << "Mie parameters out of bounds for material "
+                           << ": forward_g=" << mie_data.forward_g
+                           << ", backward_g=" << mie_data.backward_g
+                           << ", forward_ratio=" << mie_data.forward_ratio);
+        }
```
sethrj:
```suggestion
        CELER_VALIDATE(mie_data,
                       << "Mie parameters out of bounds for material " << mat.get()
                       << ": forward_g=" << mie_data.forward_g
                       << ", backward_g=" << mie_data.backward_g
                       << ", forward_ratio=" << mie_data.forward_ratio);
        
```

sethrj (approved):
One last minor change. Thanks!

sethrj:
Sorry for not noticing that it didn't get merged! 😓 (Adding support merge queues is on the list...) Thanks @Rashika-Gupta ! 😄

PR #1934, opened 2025-09-04
Implement Geant4 geometry normals and refactor/add tests

sethrj:
This implements the `GeantGeoTrackView ::normal` function and adds a bunch of tests based on it, both regression (saving the dot product with the track direction) and consistency (checking the normal direction before and after a surface crossing).

To do in follow-on PRs:
- Fix ORANGE normals on transformed daughters
- Follow-up to #1926 which fails whenever the state is not a daughter


elliottbiondo:
@sethrj let me know when this is RTR (I keep seeing you push new commits)

sethrj:
@elliottbiondo It's ready now. What I usually do is mark it ready when it's passing on a couple of CI configurations and all the substantive changes are done, then I knock out the edge cases (older versions of geant4, single precision, etc.) that show up on the CI. (This time I think I had to add a couple of extra diagnostics to debug those failures.) And if something major crops up I'll mark it as draft again.

test/geocel/GenericGeoResults.cc
```diff
@@ -28,29 +29,61 @@ namespace test
...
+{
+    auto& dn = this->dot_normal;
+    return dn.size() == 1 && dn.front() == 666;
+}
+
+void GenericGeoTrackingResult::clear_boring_normals()
+{
+    auto& dn = this->dot_normal;
+    if (std::all_of(dn.begin(), dn.end(), [](real_type n) {
+            return soft_equal(n, real_type{1});
```
hhollenb, line 49: Shouldn't surface normals satisfy `dot(normal, dir) <= 0` by convention, so the boring normals be -1?

test/geocel/GeoTests.hh
```diff
@@ -338,8 +366,25 @@ void FourLevelsGeoTest::test_detailed_tracking(GeoTest* test)
...
+        }
         EXPECT_EQ("Shape2", test->volume_name(geo));
         geo.cross_boundary();
+        if (!check_normal)
+        {
+            // Don't check
+        }
+        else if (test->geometry_type() != "ORANGE")
+        {
+            EXPECT_VEC_SOFT_EQ((Real3{1, 0, 0}), geo.normal());
```
hhollenb, line 382: Is this a bug associated with the ORANGE transformed daughters?

hhollenb (commented):
Looks good! The surface normal convention is a bit confusing with all the pre-normal/post-normal, maybe we should have some common helper functions to check / enforce the convention like `is_entering_surface` in optical physics?

test/geocel/GenericGeoResults.cc
```diff
@@ -28,29 +29,61 @@ namespace test
 //---------------------------------------------------------------------------//
 // TRACKING RESULT
 //---------------------------------------------------------------------------//
+// Replace dot-normals with a sentinel value
+void GenericGeoTrackingResult::disable_surface_normal()
+{
+    this->dot_normal = {666};
```
elliottbiondo: Make the sentinel value max double and store in on the class as consexpr

elliottbiondo (approved):

test/geocel/GenericGeoResults.cc
sethrj: I already have `inf` as another sentinel; I figured this was ok since it's only used internally and within about 10 lines of context.

test/geocel/GeoTests.hh
sethrj: I think so. I suspect the error is that we're doing the transformation from the "level" rather than the "surface level" (which can be different). That's next on my list.

test/geocel/GenericGeoResults.cc
sethrj: I don't think we want to have a convention for the sign of the normal. First, not all use cases require it: I think plain old reflection is indifferent to the sign since it cancels out. Second, it changes the normal from a function of (surface, position) to one of (surface, position, direction) or worse, (surface, position, previous position). If the sign of the surface matters for our surface physics (and I'm not 100% sure it does) then we can normalize it inside that class based on the geometry-provided normal and direction.

test/geocel/GenericGeoResults.cc
sethrj: Will update to reduce the hailing of satan by 100%.

test/geocel/GenericGeoResults.cc
sethrj: Sorry, I forgot that you'd made the "geo normal" change in #1804 . I think directional "sign" is just going to be too arbitrary (it would be positive "inward" on the lower side of a cylinder, but positive "outward" on the upper side) to bother flipping unless we tie it to a track's history, which we should definitely not do. I think we'll need to keep track of the orientation separately in the surface physics.

sethrj: I've added documentation noting the surface normal sign is state- and implementation-dependent.

PR #1935, opened 2025-09-04
IWYU on some files

sethrj:
These changes were causing errors in some builds of downstream users' code with newer GCC/Clang.


pcanal (approved):
LGTM.

PR #1936, opened 2025-09-04
Fix edge case where imported lambda and lambda prime tables have overlapping range

amandalund:
This fixes a failure due to an edge case in Geant4 that @sethrj inadvertently ran into in #1925 (https://github.com/celeritas-project/celeritas/pull/1925#discussion_r2322178014). If `minKinEnergy == minKinEnergyPrim`, both the low- and high-energy lambda tables are built in Geant4, but the range of the lower table is redundant with the upper table. This causes the contiguous-increasing check in Celeritas to fail; to prevent this, we can simply ignore the lower table in this case.

Also fixes the spline setting for the Rayleigh lambda prime table in older Geant4 versions.

### DIFF
--- a/src/celeritas/ext/detail/GeantProcessImporter.cc
+++ b/src/celeritas/ext/detail/GeantProcessImporter.cc
@@ -178,12 +178,14 @@ void assign_table(G4PhysicsTable const* g4table,
         table->grids.emplace_back(import_physics_log_vector(*g4vector, units));
 #if G4VERSION_NUMBER < 1100
         // Hardcode whether the lambda table uses spline for older Geant4
-        // versions. Always use spline for lambda, energy loss, range, and msc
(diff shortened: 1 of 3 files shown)


sethrj (approved):
Generally speaking I think we should do fixups during import (external input -> internal input) and validation during construction (internal input -> runtime data). But since here is where we're actually constructing `inp::XsGrid` I think it's fine. Maybe revert 695db88fa13b3bae5531fdb566b25ad5a8cb02a1 so that this gets tested (on the 11.0 build at least?)

amandalund:
Yeah, I think we might want to move it there as we refactor the import and transition towards using `inp`. Right now because of how the import is structured (e.g. we still have separate lambda and lambda prime `ImportPhysicsTable`s) it's a little simpler to do it here.

PR #1937, opened 2025-09-04
Use environment variable values if set

sethrj:
This changes the use of many environment variables such as `CELER_KILL_OFFLOAD` to use `getenv_flag`, which compares the *value* of the variable, rather than the non-emptiness of it. With this change, `CELER_KILL_OFFLOAD=0` will leave the flag disabled rather than setting it to 'on'.

It also fixes the logic on the (mostly useless) `CELER_STRIP_SOURCEDIR` flag default and allows a few hardcoded values to be disabled when building with debug assertions.


pcanal (approved):
LGTM.

PR #1940, opened 2025-09-05
Fix weight reconstruction for GeantSd

sethrj:
There was one final missing propagation of flags/data in the long chain of code. Follow-on to #1853 .


Rashika-Gupta (approved):
Thanks!

PR #1942, opened 2025-09-08
Fix ORANGE surface normal calculation on boundary of daughter universes

sethrj:
This fixes a bug in the calculation of surface normals identified in #1934 . When on the boundary of a universe (except for the top-level one), the *surface* level is less than (shallower/more global compared to) the *volume* level. This updates the transform loop to apply only from the surface level to the global level.


sethrj:
@elliottbiondo This one is ready!

hhollenb (approved):
Looks like a pretty straightforward fix, good stuff.

PR #1943, opened 2025-09-08
Use JSON reference for CSG tree

sethrj:
I'm making some simplification improvements to the CSG tree simplification and wanted to be able to do a more detailed comparison than "string not equal". This rebaselines the CsgTree test to use `EXPECT_JSON_EQ` with the help of a Copilot-assisted script.

### DIFF
--- a/scripts/dev/update-reference-json.py
+++ b/scripts/dev/update-reference-json.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python
+# Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+"""
+Update reference JSON values in googletest C++ test files based on actual output.
+
+This script reads googletest output from stdin, parses failure messages to
+extract actual JSON values, and safely replaces the corresponding expected
+values in the source test files.
+
+(NOTE: this script was written using Claude 4.0 via Copilot; its goal is
+utility over style.)
+"""
+
+import sys
+import re
+import os
+
+
+def parse_gtest_output(output_text):
+    """Parse googletest output to extract file paths, line numbers, and actual JSON values."""
+    failures = []
+
+    # Pattern to match file path and line number
+    file_pattern = r'(.*\.cc):(\d+): Failure'
+
+    lines = output_text.split('\n')
+    i = 0
+
+    while i < len(lines):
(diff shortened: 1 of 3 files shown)


test/testdetail/TestMacrosImpl.cc
```diff
@@ -77,8 +77,14 @@ ::testing::AssertionResult IsJsonEq(char const*,
     auto result = compare(expected, actual);
     if (!result)
     {
+        // Delete newlines from replacement
+        std::string new_actual{actual};
```
pcanal:
```suggestion
        std::string singleline_output{actual};
```

esseivaju (approved):
Looks good! Somewhat related, we should close the loop on #1565 since we only need to update the data files generated from SCALE to the infix notation.

scripts/dev/update-reference-json.py
```diff
@@ -0,0 +1,174 @@
...
+        return 0
+
+    with open(file_path, "r", encoding="utf-8") as f:
+        lines = f.readlines()
+
+    updated_count = 0
+
+    # Process failures in reverse line order to avoid line number shifts
+    for failure in reversed(failures):
+        line_number = failure["line_number"] - 1  # Convert to 0-based index
```
pcanal: I guess we could assert here for `failures.file_path == file_path`.

pcanal (approved):
LGTM.

scripts/dev/update-reference-json.py
sethrj: Yeah this was a pair programming exercise with Copilot/claude; it's rough around the edges.

sethrj:
@esseivaju agreed, this tangent was from me trying to solve a problem where two hemisphere shapes don't get turned into a  sphere .

test/testdetail/TestMacrosImpl.cc
```diff
@@ -77,8 +77,13 @@ ::testing::AssertionResult IsJsonEq(char const*,
     auto result = compare(expected, actual);
     if (!result)
     {
+        // Delete newlines from replacement by copying and erasing \n
+        std::string oneline{actual};
+        oneline.erase(std::remove(oneline.begin(), oneline.end(), '\n'),
+                      oneline.end());
```
amandalund, line 83: Is this change needed, or could it be reverted?

PR #1944, opened 2025-09-08
Define EXPECT_NORMAL_EQUIV helper function

sethrj:
This adds a helper function to check for soft equivalence of two directions, allowing for opposite signs.


pcanal (approved):
LGTM.

elliottbiondo (approved):

PR #1946, opened 2025-09-09
Simplify CSG expressions with two opposite clauses

sethrj:
Using a new `find` function in CsgTree, the "join" simplification now can simplify `any(a, !a, ...)` to `true` or `alll(a, !a, ...)` to false. This allows elimination of the "background" cell for the testem3 daughter, for example. To simplify the calorimeter region (`all(..., !any(all(...), all(...), ...)))`) we need to be lift common sub-expressions from the `all` into the `any`. I think given the complexity of the expressions we need to use an external logic package; this could be future work.


esseivaju (approved):
Looks all good to me! Indeed having an external package to handle logic would be a nice improvement, are you thinking a refactor similar to g4vg, or third-party library?

sethrj:
I was thinking `z3` ; here's one proposed way of trying to adapt our tree (no idea how performant it is): https://chatgpt.com/share/68c1ed16-1084-800a-aa9a-2b2c0c141ba4

PR #1947, opened 2025-09-09
Fix initialization when starting on an internal surface

sethrj:
While trying to add a split shell around the lAr detector I found that ORANGE crashed when starting tracks at the origin. To make a long story short, we were using the `on_surface` local variable as a side effect return object for when we found a legitimate volume. However, if an earlier volume tried to initialize "on" a surface but wasn't in that volume after all, we wouldn't clear the on-surface flag before testing the next volume.

### DIFF
--- a/test/orange/orangeinp/UnitProto.test.cc
+++ b/test/orange/orangeinp/UnitProto.test.cc
@@ -564,7 +564,7 @@ TEST_F(InputBuilderTest, lar_split_detector)
             "full_shell",
             {{Sense::inside, middle_sphere}, {Sense::outside, inner_sphere}});
 
-        // Construct exterior, shell, shell halvs, interior
+        // Construct exterior, shell, shell halves, interior
         append_material(inp,
                         make_rdv("outer_region",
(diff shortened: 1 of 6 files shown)


test/orange/orangeinp/UnitProto.test.cc
```diff
@@ -532,6 +533,61 @@ TEST_F(InputBuilderTest, globalspheres)
...
+            "lower_shell",
+            {{Sense::inside, lower_half}, {Sense::outside, inner_sphere}});
+        auto upper_shell = make_rdv(
+            "upper_shell",
+            {{Sense::inside, upper_half}, {Sense::outside, inner_sphere}});
+        auto full_shell = make_rdv(
+            "full_shell",
+            {{Sense::inside, middle_sphere}, {Sense::outside, inner_sphere}});
+
+        // Construct exterior, shell, shell halvs, interior
```
elliottbiondo: halves

elliottbiondo (approved):

PR #1948, opened 2025-09-09
Add shell detectors around lAr sphere test problem

sethrj:
This adds some complexity to the lAr sphere problem for testing. This was originally in #1925 but then I found a bug which necessitates #1947.

I've deleted the ROOT importing test because the ROOT capability isn't currently useful due to the tightly coupled geometry metadata.


test/geocel/GeoTests.cc
```diff
@@ -366,6 +366,62 @@ void FourLevelsGeoTest::test_trace() const
...
+            "detshell_top_PV",
+            "world_PV",
+        };
+        ref.distances = {10, 10, 200, 10, 890};
+        ref.halfway_safeties = {5, 5, 100, 5, 445};
+        ref.bumps = {};
+        if (is_orange)
+        {
+            // TODO: at this exact point it ignores the spherical distance
+            ref.halfway_safeties[2] = result.halfway_safeties[2];
```
amandalund, line 400: Why is this?

amandalund (approved):
Looks good!

test/geocel/GeoTests.cc
sethrj: The ORANGE safety calculation currently calculates the gradient of the quadric and finds the distance to boundary along that direction. Unfortunately at the exact center of the sphere, the gradient is zero so there's no intersection.

Issue #1750, opened 2025-04-17
Define new detector params class for managing Geant4 SD callbacks


sethrj:
This will (eventually) replace some components of the `GeantSd` and `HitProcessor` classes and prepare them for optical detectors.  It should also replace part of the SimpleCalo class (basically could start as a refactoring of it). This should be used for #1746 and, more immediately, #1397.

Construction will be with list of canonical volume IDs and geo interface

DetectorParamsData:
- `Collection<DetectorId, W, M, VolumeId> detector_ids`

Class interface:
- `class DetectorParams : public ParamsDataInterface<DetectorParamsData>`
- Member function to return DetectorId -> VolumeId
- Lives in `celeritas/user`

Then:
- Add to optical core params
- CoreParamsData would have a `DetectorParamsData<Ownership::const_reference, M> detectors`

Future work:
- Add to global core params
- Add thread-local DetectorId -> G4SD* mapping
- Replace components in GeantSd

sethrj:
See #1764

sethrj:
Add detectors to the `inp::Model`:
- `Detector` struct 
-  DetectorId moves to geocel/Types
- GeantGeoParams defines `G4SD* id_to_geant(DetectorId)` and `geant_to_id` functions (use code from `GeantSd::setup_volumes` to map)
- Add tests to GeantGeo.test.cc (model testing is implemented in `GenericGeoResults.{hh,cc}`
- Update `SDParams` to take `inp::Detectors`
- Construct SDParams in `setup::model`
- Add SDParams to global `CoreParams::Input` and store in CoreParams
- Copy to optical core params via OpticalCollector
- Look up post-volume crossing in optical PostBoundaryExecutor, log and absorb if detector found




```c++
struct Detector
{
    Label label;
    std::vector<VolumeId> volumes;
};
```

PR #1949, opened 2025-09-09
Create placeholder SDs when loading GeantGeoParams from GDML

sethrj:
To support #1750 we want to be able to test that SDs get loaded from users (or GDML input). This adds a little code during GeantGeoParams setup that creates dummy SDs based on the aux data loaded from the GDML file. They can be replaced without consequence by the user or left in place.

### DIFF
--- a/src/geocel/GeantGeoParams.cc
+++ b/src/geocel/GeantGeoParams.cc
@@ -20,6 +20,7 @@
 #include <G4Transportation.hh>
 #include <G4TransportationManager.hh>
 #include <G4VPhysicalVolume.hh>
+#include <G4VSensitiveDetector.hh>
 #include <G4VSolid.hh>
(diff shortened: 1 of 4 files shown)


src/geocel/GeantGeoParams.cc
```diff
@@ -456,6 +470,8 @@ std::shared_ptr<GeantGeoParams> GeantGeoParams::from_tracking_manager()
  * This assumes that Celeritas is driving and will manage Geant4 logging
  * and exceptions. It saves the result to the global Celeritas Geant4 geometry
  * weak pointer \c global_geant_geo.
+ *
+ * It also loads sensitive detectors and assigns dummy
```
amandalund: 🐿️

amandalund (approved):

PR #1951, opened 2025-09-09
Fix sincos on device when math/Turn.hh is included

sethrj:
When a `sincos` in the `celeritas` namespace is available that has an incompatible signature with the `::sincos` defined by CUDA's headers, using `sincos` inside the celeritas namespace causes an error. This fixes the conflict by making sure that there's a compatible `sincos` that correctly overloads with the celeritas Turn one; it simply forwards to CUDA (but correctly overloads float/double values).

### DIFF
--- a/test/corecel/math/Algorithms.test.cu
+++ b/test/corecel/math/Algorithms.test.cu
@@ -9,6 +9,8 @@
 
 #include "corecel/DeviceRuntimeApi.hh"
 
+#include "corecel/Constants.hh"
+#include "corecel/math/Turn.hh"
 #include "corecel/sys/Device.hh"
 #include "corecel/sys/KernelParamCalculator.device.hh"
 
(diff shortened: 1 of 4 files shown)


test/corecel/math/Algorithms.test.cu
```diff
@@ -32,6 +34,9 @@ __global__ void alg_test_kernel(AlgorithmTestData data)
     if (tid.get() < inp.pi_frac.size())
     {
         sincospi(inp.pi_frac[tid], &out.sinpi[tid], &out.cospi[tid]);
+        sincospi(inp.pi_frac[tid], &out.sinpi[tid], &out.cospi[tid]);
```
amandalund: This line was duplicated.

amandalund (approved):
Thanks @sethrj!

hhollenb (approved):
Always enjoy trying to parse C macros. Looks good!

PR #1952, opened 2025-09-10
Add log code to print executable/source/dependency locations

sethrj:
This adds some example code that can be copied by developers/users to help ensure that a running example is using the expected source code/build directory.


example/offload-template/run-offload.cc
```diff
@@ -40,6 +71,26 @@ int main(int argc, char* argv[])
     // Initialize Celeritas
     auto& tmi = celeritas::TrackingManagerIntegration::Instance();
 
+    {
+        namespace fs = std::filesystem;
```
pcanal: Since this is an example, consider adding documentation, for example here, explaining that this relies on `SourceFiles.hh.in` being configured by CMake and resulting in the information being recorded, etc.

example/offload-template/run-offload.cc
```diff
@@ -40,6 +71,26 @@ int main(int argc, char* argv[])
...
+                        << shorter_path(celeritas::example::build_dir);
+        CELER_LOG(info)
+            << "Celeritas install: "
+            << shorter_path(celeritas::example::celeritas_install_dir);
+        CELER_LOG(info)
+            << "Geant4 install: "
+            << shorter_path(celeritas::example::geant4_install_dir);
+
+        CELER_LOG(info) << "Run manager type: "
+                        << TypeDemangler<G4RunManager>{}(*run_manager);
```
pcanal:
```suggestion
                        << celeritas::TypeDemangler<G4RunManager>{}(*run_manager);
```

pcanal (approved):
LGTM.

Rashika-Gupta (approved):
Thanks for including these changes. Looks good!

stognini:
Looks great. Agree with @pcanal about a short comment documenting that. Or even making the whole `{ ... }` as a free function (`void print_setup_info()`) which allows for documentation at the top of the free function. I don't have a strong opinion, so either case looks good to me.

stognini (approved):
Thanks!

PR #1953, opened 2025-09-10
Replace `ImportParticle` with `inp::Particle`

amandalund:
Small update that adds a `inp::Particle` struct to replace `ImportParticle` and `ParticleParams::ParticleInput`.

PR #1954, opened 2025-09-11
Define Geant particle view

sethrj:
This follows up on #1953 with a helper lass I wrote a few months ago to help with G4Cherenkov integration.

### DIFF
--- a/src/celeritas/ext/GeantParticleView.hh
+++ b/src/celeritas/ext/GeantParticleView.hh
@@ -1,0 +1,75 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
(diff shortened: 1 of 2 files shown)


src/celeritas/ext/GeantParticleView.hh
```diff
@@ -0,0 +1,86 @@
...
+ * Access invariant particle data from Geant4 with Celeritas units.
+ */
+class GeantParticleView
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Charge = units::ElementaryCharge;
+    using Energy = units::MevEnergy;
+    using Mass = units::MevMass;
```
amandalund: Should we be using `double` `Quantity`s for these?

amandalund (approved):
Cool, thanks!

src/celeritas/ext/GeantParticleView.hh
sethrj: Yep indeed, that's why single-precision was failing 🙃

PR #1955, opened 2025-09-11
Adds Detectors to model input and builds the detectors from G4 sensitive detectors

lebuller:
This PR creates a new `Detectors` struct in the model input, which contains a vector of `Detector` structs. The `Detector` struct contains a label for the detector and a list of volume IDs over which the detector is defined. The `Detectors` are constructed from existing G4 sensitive detectors in `GeantGeoParams`. Additionally, the `DetectorID` declaration was moved from `celeritas/types` to `geocel/types`


test/geocel/g4/GeantGeo.test.cc
```diff
@@ -388,6 +390,8 @@ TEST_F(LarSphereTest, model)
     };
     ref.volume_instance.volumes = {4, 3, 1, 2, 0};
     ref.world = "world";
+    ref.detector.labels = {"detshell"};
+    ref.detector.volumes = {{1}};
```
sethrj: This will probably need to be updated when merging upstream due to #1948 .

src/geocel/GeantGeoParams.cc
```diff
@@ -383,6 +383,65 @@ std::vector<inp::Surface> make_inp_surfaces(GeantGeoParams const& geo)
...
+    for (auto iv_id : range(ImplVolumeId{vol_labels.size()}))
+    {
+        auto const& label = vol_labels.at(iv_id);
+        if (label.empty())
+        {
+            // This volume isn't part of the world hierarchy
+            continue;
+        }
+
+        auto vol_id = geo.volume_id(iv_id);
```
sethrj: I think this should work:
```suggestion
        auto vol_id = geo.volume_id(iv_id);
        if (!vol_id)
        {
            // This volume isn't part of the world hierarchy
            continue;
        }
```

src/geocel/GeantGeoParams.cc
```diff
@@ -383,6 +383,65 @@ std::vector<inp::Surface> make_inp_surfaces(GeantGeoParams const& geo)
...
+            // match.
+            for (auto detector : result)
+            {
+                if (detector.label == sd_label)
+                {
+                    existing_sd = true;
+                    detector.volumes.push_back(vol_id);
+                    break;
+                }
+            }
```
sethrj: To avoid this O(N^2) loop, create a temporary `unordered_map<G4VSD*, std::vector<VolumeId>> detectors` beforehand, do `detectors[sd].append(vid);` in the volume loop, then after the volume loop convert into inp::Detector. Finally, sort the result based on `volumes.front()` to be reproducible.

test/geocel/g4/GeantGeo.test.cc
```diff
@@ -220,6 +220,8 @@ TEST_F(CmsEeBackDeeTest, model)
     };
     ref.volume_instance.volumes = {3, 2, 0, 1, 4, 5, 6};
     ref.world = "EEBackDee";
+    ref.detector.labels = {"ee_back_plate", "ee_s_ring"};
+    ref.detector.volumes = {{0}, {1}};
```
sethrj: I'm not sure this is right; the `EEBackPlate_refl` should also have an SD attached :\ the same is true for `sph_sd` below; see `MultiLevelTest.sd_creation` in GeantGeo.test.cc

src/geocel/GeantGeoParams.cc
```diff
@@ -383,6 +383,65 @@ std::vector<inp::Surface> make_inp_surfaces(GeantGeoParams const& geo)
...
+        // check if volume is part of a sensitive detector
+        if (sd)
+        {
+            Label sd_label = sd->GetName();
+            bool existing_sd = false;
+
+            // Loop over existing input detectors in the result vector and
+            // append volume id to input detector volume id list if labels
+            // match.
+            for (auto detector : result)
```
sethrj: This is why the extra volumes weren't getting appended: you're unintentionally creating copies that get modified and discarded 🙃 
```suggestion
            for (auto& detector : result)
```

sethrj (commented):
Thanks @lebuller ! Needs just a little work.

src/geocel/GeantGeoParams.cc
```diff
@@ -383,6 +383,56 @@ std::vector<inp::Surface> make_inp_surfaces(GeantGeoParams const& geo)
...
+            detector_map[sd].push_back(vol_id);
+        }
+    }
+
+    // Loop over detector_map and add detectors to result vector
+    for (auto const& sd_pair : detector_map)
+    {
+        inp::Detector detector;
+        detector.label = sd_pair.first->GetName();
+        detector.volumes = detector_map[sd_pair.first];
```
sethrj: A little simpler and avoids a vector copy:
```suggestion
    for (auto&& [sd, volumes] : detector_map)
    {
        inp::Detector detector;
        detector.label = sd_pair.first->GetName();
        detector.volumes = std::move(volumes);
```

src/geocel/GeantGeoParams.cc
```diff
@@ -383,6 +383,56 @@ std::vector<inp::Surface> make_inp_surfaces(GeantGeoParams const& geo)
...
+        if (!vol_id)
+        {
+            // This volume isn't part of the world hierarchy
+            continue;
+        }
+        auto& g4lv = *geo.id_to_geant(vol_id);
+
+        G4VSensitiveDetector const* sd = g4lv.GetSensitiveDetector();
+        // Add volume id to detector map if it is in a detector region
+        if (sd)
```
sethrj: One idiom we often use is to limit the scope of a variable by declaring it inside the `if` condition:
```suggestion
        // Add volume id to detector map if it is in a detector region
        if (auto const* sd = g4lv.GetSensitiveDetector())
```

sethrj (approved):
Excellent! Two minor style suggestions.

sethrj:
Thanks @lebuller !

PR #1957, opened 2025-09-11
Add a test of UI command modification

sethrj:
This adds a test of the SetupOptionsMessenger class and makes some improvements to the integration helpers and test classes.


src/accel/SharedParams.cc
```diff
@@ -154,7 +154,8 @@ std::mutex& updating_mutex()
 /*!
  * Whether celeritas is disabled, set to kill, or to be enabled.
  *
- * This gets the value from environment variables and
+ * Currently the momde is set by querying the environment variables \c
```
stognini:
```suggestion
 * Currently the mode is set by querying the environment variables \c
```

src/accel/IntegrationBase.cc
```diff
@@ -21,23 +21,31 @@ namespace celeritas
...
+ * This is only legal to call after \c SetOptions.
  */
-void IntegrationBase::SetOptions(SetupOptions&& opts)
+OffloadMode IntegrationBase::GetMode() const
 {
-    IntegrationSingleton::instance().setup_options(std::move(opts));
+    auto const& singleton = IntegrationSingleton::instance();
+    if (singleton.offloaded_particles().empty())
+    {
+        CELER_LOG(warning) << "GetMode cannot be called before SetOptions";
```
stognini: Using `IntegrationSingleton::offloaded_` as the main point to know if the `OffloadMode` is uninitialized feels a bit awkward. We do have a shared params default constructor with an uninitialized Offload mode once the singleton exists, right? If so, we could just fetch `IntegrationSingleton::instance().shared_params().mode()` and `CELER_LOG` accordingly.

sethrj: I'll move this inside the singleton as an implementation detail. `.mode()` doesn't work because initialization only happens at BeginOfRun, whereas here we need to know the run mode (which, although currently set by global environment variables, will soon be set using `inp` structures.)

sethrj: I see that I was accidentally using the member data `mode` rather than class/environment `GetMode`...

stognini (approved):
Thanks @sethrj !

PR #1959, opened 2025-09-12
Fix data selection in `SetupOptions` -> `FrameworkInput` conversion

amandalund:
Little fix for the conversion from `SetupOptions` to `FrameworkInput` where the optical selection was being overwritten if muons were present.


sethrj (approved):
Thanks. This is high on my list...

amandalund:
I saw you mentioned it, and then coincidentally ended up digging into it today. Definitely needed, and appreciate it!

PR #1961, opened 2025-09-14
Fix offloaded particles in celer-g4

amandalund:
Muons were no longer being offloaded through the celer-g4 app even when muon physics was enabled since we moved to a hardcoded list of offload particles. This adds a quick fix which checks whether muons are enabled in physics options and sets the offload particles to the "supported" particles if so. However, it seems like in the long term it might be better to detect the offload particles automatically; is there a reason we moved away from doing this?


sethrj:
Let's talk about this Monday: I was imagining the user would want to input offload particles (combo of EM/muon/optical) and then we pull in the corresponding processes.

sethrj (approved):

PR #1962, opened 2025-09-15
Use Geant4 rather than ROOT for MSC tests

sethrj:
Trying to reduce the reliance on internal ROOT data (more stable if ImportData doesn't change, more annoying if it does).


sethrj:
OK, I guess since MSC physics depends on the slowing down energy loss, which is very sensitive to Geant4 version number, we can close this. Once we get all the physics data into the `inp` pipeline and can export it without ROOT we should be able to run these tests reliably without ROOT *or* Geant4 (good for the "ultralite" runs to test on windows).

PR #1963, opened 2025-09-15
Skip wavelength shift model without data during optical model import

Rashika-Gupta:
This PR updates the optical model import logic to safely skip models that have no valid data instead of failing during construction. Currently, this behavior is implemented only for WLS but can be extended to other models in the future.

1. `ModelImporter` :  Return `std::nullopt` when a WLS/WLS2 material has no data, instead of attempting to build.
2. `PhysicsParams` : Add a guard to continue building remaining optical models when a builder is empty.


src/celeritas/optical/ModelImporter.cc
```diff
@@ -126,8 +131,17 @@ auto ModelImporter::build_wls() const -> ModelBuilder
...
     {
-        input.data.push_back(input_.import_material->wls(mid));
+        auto wls_data = input_.import_material->wls(mid);
+        if (wls_data)
+        {
+            input.data.push_back(std::move(wls_data));
+        }
+        else
+        {
+            return {};
```
sethrj: Doesn't this mean that if any model *lacks* data, WLS will be omitted? I think if you revert this change and do, after the loop:
```c++
if (!std::any_of(input.data.begin(), input.data.end(), Identity{})
{
    // None of the materials have WLS data
    return {};
}
```

src/celeritas/optical/ModelImporter.cc
```diff
@@ -145,7 +159,15 @@ auto ModelImporter::build_wls2() const -> ModelBuilder
...
-        input.data.push_back(input_.import_material->wls2(mid));
+        auto wls2_data = input_.import_material->wls2(mid);
+        if (wls2_data)
+        {
+            input.data.push_back(input_.import_material->wls2(mid));
+        }
+        else
+        {
+            return {};
+        }
```
sethrj: Same here.

sethrj (commented):
I think the optical model importer test is failing too: https://github.com/celeritas-project/celeritas/actions/runs/17744182357/job/50425253624?pr=1963

`bad_function_call` means there's a std::function (which `ModelBuilder` is aliased to) that is null, which means somewhere is missing an `if (builder)` or `CELER_ASSERT(builder)`.

src/celeritas/optical/ModelImporter.cc
Rashika-Gupta: you're right!

src/celeritas/optical/ModelImporter.cc
Rashika-Gupta: done

src/celeritas/optical/ModelImporter.cc
```diff
@@ -128,6 +133,11 @@ auto ModelImporter::build_wls() const -> ModelBuilder
     {
         input.data.push_back(input_.import_material->wls(mid));
     }
+    if (!std::any_of(input.data.begin(), input.data.end(), Identity{}))
```
sethrj, line 138: I wish we had an automated IWYU, but this needs the `<algorithm>` header to be included. Mind adding it? :)

sethrj (approved):
Looks good aside from the missing include (I know my accidental omission of `<algorithm>` in the past have caused build errors for other users). You can also merge in upstream/develop so that this will be ready to squash. Thanks!

src/celeritas/optical/ModelImporter.cc
Rashika-Gupta: done!

PR #1965, opened 2025-09-16
Fix propagation of weight from primary to track to secondary

sethrj:
There were a few assignments missing in step initialization. This adds a front-to-back test for the track weights, as well as a helper class to easily check hits inside the integration unit test.


amandalund (approved):
Thanks for the final fix!

Rashika-Gupta (approved):
Thank You !

PR #1966, opened 2025-09-17
Add documentation and use flag for CELER_USE_PROFILING

sethrj:
- Use an environment flag (true if perfetto is enabled, false otherwise; we can change this) for enabling profiling. I also tried to find a way to detect whether we're being run through nsys profiler but no, alas
- Add user documentation for the profiling classes
- Add an example to the offload app to show how to profile events
- Fix use of `string_view` when null-terminated string is needed
- Add log message when opening perfetto
- Add perfetto tracing to IntegrationTestBase


doc/usage/execution/profiling.rst
```diff
@@ -5,8 +5,8 @@
 
 .. _profiling:
 
-Profiling
-=========
+Performance profiling
```
pcanal, line 8:
```suggestion
CPU/GPU Performance profiling
```

example/offload-template/README.rst
```diff
@@ -36,6 +36,11 @@ Example classes
   master and worker threads, setting up a tracking manager under the hood.
   :cpp:class:`EndOfRunAction` clears data and finalizes Celeritas data.
 
+:cpp:class:`EventAction`
+  The event action is not necessary to run Celeritas, but its use of
+  :cpp:class:`celeritas::ScopedProfiling` demonstrates how to profile Geant4
+  and annotate GPU profiling results in a Geant4 context.
```
pcanal: Why the focus on 'not being necessary' since those are listed under "Example classes"?
```suggestion
  The event action use of
  :cpp:class:`celeritas::ScopedProfiling` demonstrates how to profile Geant4
  and annotate GPU profiling results in a Geant4 context.
```

sethrj: Fair enough. I was trying to contrast it with the "minimal necessary" aspect of this example.

doc/usage/execution/profiling.rst
sethrj: As opposed to memory profiling? Network profiling? 🙃 I think we'd want them all in this documentation page

src/corecel/sys/ScopedProfiling.cc
```diff
@@ -26,34 +26,44 @@ namespace celeritas
 /*!
  * Whether profiling is enabled.
  *
- * This is true only if the \c CELER_ENABLE_PROFILING environment variable is
- * set to a non-empty value. Profiling is never enabled if CUDA/HIP/Perfetto
- * isn't available.
+ * This is true by default when compiling with Perfetto enabled. The
```
pcanal: > This is true by default ...

humm ... pondering if this is the most intuitive choice.

sethrj: Yeah I could revert this. I really wanted to be able to detect whether Celeritas is being run through `nsys` so that the user wouldn't suprisingly get empty output. I think if they explicitly specify to build with Perfetto (off by default) that they're looking for profiling results "most of" the time. /shrug

test/accel/IntegrationTestBase.cc
```diff
@@ -155,19 +166,27 @@ class EventAction final : public G4UserEventAction
 class ActionInitialization final : public G4VUserActionInitialization
 {
   public:
-    explicit ActionInitialization(IntegrationTestBase* test) : test_{test} {}
+    explicit ActionInitialization(IntegrationTestBase* test) : test_{test}
+    {
+        if (!CELER_USE_DEVICE && use_profiling())
```
pcanal: That smell fishy ... Is it actually semantically:
```suggestion
        if (use_profiling() && CELERITAS_ACTUALLY_USE_PERFETTO)
```
?
or can it just be
```suggestion
        if (use_profiling())
```

example/offload-template/README.rst
pcanal:
```suggestion
  The event action is optional and its use of
  :cpp:class:`celeritas::ScopedProfiling` demonstrates how to profile Geant4
  and annotate GPU profiling results in a Geant4 context.
```

doc/usage/execution/profiling.rst
pcanal: > As opposed to memory profiling? Network profiling?

Yes (and I/O etc.)

>  I think we'd want them all in this documentation page

Fair enough, especially after scanning the whole file (rather than just the diffs :) )

src/corecel/sys/ScopedProfiling.cc
pcanal: > that they're looking for profiling results "most of" the time. 

That's a good point.  Similarly the `-fprofile` compiler flags leads to always on profiling.  But then conversely why/when would you want to disable it?

pcanal: So maybe the right comparison is the way the GPU profiling works (on only if run under nsys) in which case it would seem the default should be off. ie. profiling on only if:
```
nsys myexec
CELER_ENABLE_PROFILING=true myexec
```

sethrj: @pcanal Disabling is useful to measure the overhead of the profiling calls, I suppose? And for nsys it has to be `CELER_ENABLE_PROFILING=true nsys myexec` sadly. Maybe I'll just print a warning if it defaults (no profiling specified) when perfetto is explicitly compiled in.

pcanal: > Disabling is useful to measure the overhead of the profiling calls, I suppose?

Wouldn't the perfecto-disabled build be a better yardstick?

test/accel/IntegrationTestBase.cc
sethrj: Oh it should be perfetto, silly me...

sethrj:
@pcanal Ready for re-review 🙂

sethrj:
Sorry, accidentally force-pushed 😒

src/corecel/sys/Environment.cc
```diff
@@ -64,6 +65,9 @@ std::string const& getenv(std::string const& key)
  * specifying whether the default was used. The insertion result can be useful
  * for providing a diagnostic message to the user.
  *
+ * As with the general \c Environment class, the already-set values override
+ * whatever variables are in the system.
```
pcanal: Can you clarify a bit (what is 'the system', how can the value be already set?, etc.).
```suggestion
 * As with the general \c Environment class, the already-set values has priority over
 * whatever variables are in the system.
```

src/corecel/sys/Environment.cc
```diff
@@ -73,32 +77,49 @@ GetenvFlagResult getenv_flag(std::string const& key, bool default_val)
...
+        auto& env = environment();
+        if (auto iter = env.find(key); iter != env.end())
+        {
+            // Variable was already loaded internally
+            return iter->second;
+        }
+        else if (char const* sys_value = std::getenv(key.c_str()))
+        {
+            // Variable is set in the user environment
+            return sys_value;
```
pcanal, line 99: Do we mean to insert this value in environment?  If so where is the insert?

src/corecel/sys/ScopedProfiling.hh
```diff
@@ -38,25 +38,40 @@ struct ScopedProfilingInput
...
 /*!
- * RAII class for scoped profiling.
+ * Enable and annotate performance profiling during the lifetime of this class.
  *
- * Implementations should support multithreaded context where each thread have
- * one or more alive instance of this class.
+ * This RAII class annotates the profiling output so that, during its scope,
+ * events and timing are associated with the given name. For use cases inside
+ * separate begin/end functions of a class (often seen in Geant4), use
+ * \c std::optional to start and end the class lifetime.
```
pcanal, line 46:
```suggestion
 * \c std::optional to start and end the object lifetime.
```
You may(or may not) also want to add an explicit example.

src/corecel/sys/Environment.cc
sethrj: Crap, you're right, ... time to add another test

pcanal (approved):
LGTM.

PR #1968, opened 2025-09-17
Add push/flush profiling to accel interface

sethrj:
This complements #1966 by adding ScopedProfiling calls to LocalTransporter and a few other potential hot spots (optical tracking loop and RNG reseed). It would be good to know whether these extra calls lead to any extra measurable overhead, both with and without `use_profiling` enabled, for the different back ends. (I can compare this on wildstyle for the extreme case of having many primaries with very little work performed.)


pcanal (approved):
LGTM.

PR #1970, opened 2025-09-18
Improve optical setup and testing

sethrj:
- Use sentinel value instead of `enabled` for WLS physics options
- Add optical `accel` test
- Catch error if optical models are enabled but optical loop isn't to be built

And testing improvements:
- Make lAr dummy values less unrealistic
- Add error checking to accel test harness
- Add tracking action to count optical photons

To do:
- Fix "unknown process" error messages for Mie, OpBoundary
- Fix "Mismatch between ParticlesParams (size 4) and user-defined offload list (size 3)" warning
- Track down and fix performance issues for this test case (geometry initialization definitely isn't great, and maybe we can have an option to flip the loop structure for generation to make it more efficient for CPU)


test/accel/IntegrationTestBase.cc
```diff
@@ -386,13 +399,23 @@ auto LarSphereIntegrationMixin::make_sens_det(std::string const& sd_name)
  */
 void LarSphereIntegrationMixin::process_hit(G4Step const* step)
 {
-    ASSERT_TRUE(step);
-    ASSERT_TRUE(step->GetTrack());
+    if (!step || !step->GetTrack())
```
stognini, line 436: This seems off. If `step = nullptr` we assert that it is not a `nullptr`?
We either test that every `step*` is in a valid state, or if we are sure they will be, we can just proceed with the `EXPECT` calls, no?

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -174,8 +183,8 @@ struct GeantOpticalPhysicsOptions
         GeantOpticalPhysicsOptions opts;
         opts.cherenkov.enable = false;
         opts.scintillation.enable = false;
-        opts.wavelength_shifting.enable = false;
-        opts.wavelength_shifting2.enable = false;
+        opts.wavelength_shifting = WavelengthShiftingOptions::deactivated();
+        opts.wavelength_shifting2 = WavelengthShiftingOptions::deactivated();
```
amandalund, line 189: I know we've discussed this before (https://github.com/celeritas-project/celeritas/pull/1822#issuecomment-2963489854): I do think the original version makes it a little more transparent when WLS is enabled (`wavelength_shifting.enable = true` vs. `wavelength_shifting.time_profile = WlsTimeProfile::delta`) and is more consistent with the other options, but since we're likely going to refactor this to be more `inp`-like it doesn't really matter.

If we do keep this though we should check for `enable` in `from_json_deprecated` to support older version.

test/accel/IntegrationTestBase.cc
sethrj: The line below "Reduce testing overhead: google assertions allocate memory" is supposed to explain this: `ASSERT_TRUE` is apparently a little expensive, and since this is in "process_hit" (as much as once per step) it causes a noticeable slowdown. This way avoids the slowdown in the typical case, but correctly fails if something is very wrong.

stognini: Ohh I get it now. I did notice the comment, but did not process that it is checking "cheaply" and asserting when it is wrong.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -188,6 +202,213 @@ TEST_F(LarSphere, ui)
...
+ *
+ * \todo This is redundant with (but more "Geant4-like" than)
+ * \c GeantStepDiagnostic .
+ */
+class TrackingAction : public G4UserTrackingAction
+{
+  public:
+    void PreUserTrackingAction(G4Track const* t)
+    {
+        constexpr PDGNumber optical_photon{-pdg::gamma().unchecked_get()};
```
amandalund: Should we be supporting older Geant4 versions where the optical photon PDG was 0?

test/geocel/data/lar-sphere.gdml
```diff
@@ -2,13 +2,14 @@
...
     <matrix name="RS" coldim="1" values="1.0"/>
-    <matrix name="SY" coldim="1" values="50000.0"/>
+    <!-- Scintillation properties -->
+    <matrix name="SY" coldim="1" values="5000.0"/>
```
amandalund, line 9: This value was taken from the simpleLArTPC.gdml in the Geant4 CaTs example, so may or may not be unreasonable.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -188,6 +202,213 @@ TEST_F(LarSphere, ui)
...
+    {
+        std::size_t photons{0};
+        std::size_t electrons{0};
+        for (auto* tracking_action : tracking_)
+        {
+            photons += tracking_action->num_photons();
+            electrons += tracking_action->num_electrons();
+        }
+        CELER_LOG(info) << "Geant4 tracked a total of " << photons
+                        << " photons"
```
amandalund:
```suggestion
                        << " optical photons"
```
Just to be clear.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
```diff
@@ -83,20 +83,29 @@ constexpr bool operator==(ScintillationPhysicsOptions const& a,
 //! Optical wavelength shifting process options
 struct WavelengthShiftingOptions
 {
-    //! Enable the process
-    bool enable{true};
```
stognini: Seems like for a future PR we should make all the others consistent (ie removing bool enable, and update operator bool).

amandalund (approved):
Thanks for the test!

src/celeritas/ext/GeantImporter.cc
```diff
@@ -1160,11 +1160,16 @@ ImportOpticalParameters import_optical_parameters()
...
         {
             return WlsTimeProfile::exponential;
         }
-        CELER_ASSERT_UNREACHABLE();
+        else
+        {
+            CELER_LOG(error) << "unknown WLS time profile '" << time_profile
+                             << "': using exponential";
+            return WlsTimeProfile::exponential;
+        }
```
Rashika-Gupta: Just wondering why are we defaulting it to the exponential ?

stognini (approved):
Thanks!

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
sethrj: Ah I forgot that discussion. Did we have `OpticalPhysicsOptions::deactivated()` at the time? Having a separate "enable" option seemed redundant to me, and eliminating the redundancy did reveal that we were setting geant4 optical options even when we ourselves weren't using them.

test/accel/TrackingManagerIntegration.test.cc
sethrj: Yep good idea. I couldn't find where we were doing that elsewhere, but it's a little wrong (not version dependent) and implemented deep inside PhysicsParams:
```c++
bool ignore_particle(PDGNumber pdg)
{
    // PDG recommends 81-100 for internal MC pseudoparticles. The optical
    // physics is constructed separately from the rest of the physics: Geant4
    // uses 0 for the optical photon PDG before version 10.7 and -22 after.
    return (pdg.get() >= 81 && pdg.get() <= 100) || pdg.get() == 0
           || pdg.get() == -22;
}
```

So perhaps another utility `bool is_optical_photon(G4ParticleDefinition*)` that checks PDG if >= ... or checks name if old and PDG == 0? I wonder also if the geant4 model (currently geo params?) should be expanded to store G4PD* so that we can correlate directly to that from ParticleId.

test/geocel/data/lar-sphere.gdml
sethrj: Good to know. Still though, for a unit test, we don't need that many photons to ensure the code works.

Rashika-Gupta (approved):

src/celeritas/ext/GeantImporter.cc
sethrj: Good question. We probably should just assert that the unknown time profile isn't implemented in Celeritas.

test/accel/TrackingManagerIntegration.test.cc
amandalund: The other place is in the `GeantImporter`: https://github.com/celeritas-project/celeritas/blob/ac2926c0b03d70e937c0ad6bd777090e04af6439/src/celeritas/ext/GeantImporter.cc#L1040

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
amandalund: Looks like we did have it then. You mean setting the options in `SupportedOpticalPhysics`?

test/accel/TrackingManagerIntegration.test.cc
stognini: Is there any point where we need to check if PDG is valid?

`PDGNumber` does that via
```cpp
//! True if value is nonzero
explicit constexpr operator bool() const { return value_ != 0; }
```

and if we use PDG = 0 for optical photons for old G4 versions, I can see something breaking downstream.

amandalund: Good point @stognini: do we save the PDG as -22 in Celeritas for all Geant4 versions, or do we use whatever Geant has?

stognini: Looks like `GeantImporter` simply copies the PDG number, and `ParticleParams` just maps PDG to ID... so it seems we just use whatever we get from G4.

Using the antiphoton PDG for optical photons is a nice hack, since photons are their own antiparticles, but I don't know if we want to use the same solution in Celeritas.

For example, in this PR, in the `TrackingManagerIntegration` test, Seth avoided defining it by adding a negative sign here
```cpp
constexpr PDGNumber optical_photon{-pdg::gamma().unchecked_get()};
```
which to me was an indication that maybe he wanted to keep that out of our PDG list (i.e. not use the -22 hack from G4, and just have optical photons not be in the list, since they are not part of the official PDG numbering scheme).

sethrj: Yeah. It's not an antiphoton, and we don't want it in our particle parameters since it's not part of that transport loop. PDG number is metadata

stognini: In that case, I think making `-22` the default PDG for every G4 version in our import interface is the easiest. The `-22` simply becomes the trigger (or verification) before we build any params. Then the optical photons simply live in their own separate "world" (allocated memory, stepping loop, etc) in Celeritas. If we frequently need some comparison, such as that `optical_photon` value in the test, we can have a Geant4 free helper function that returns a bool (like `bool is_optical_photon(G4ParticleDefinition const& p)`).

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
sethrj: Yeah, I'm going to restore this bit but leave the 'deactivated' and `operator bool` so that we can change the implementation.

src/celeritas/ext/GeantOpticalPhysicsOptions.hh
sethrj: I'll revert the `enabled` flag change and just add the `deactivated`.

test/accel/TrackingManagerIntegration.test.cc
sethrj: I've added an is_optical_photon for the GeantParticleView.

test/accel/TrackingManagerIntegration.test.cc
sethrj: I mean, it is a unit test called `LarSphereOptical` 🙃

PR #1971, opened 2025-09-18
Warn user when running large problems on debug build

sethrj:
This adds a warning during "finalize" if the user runs more than a million steps on a non-release (or release with assertions) build. I almost started performance profiling a debug build 🙃


amandalund (approved):
Useful, thanks!

PR #1972, opened 2025-09-19
Use consistent tolerance when checking for orthogonality

amandalund:
Because we were using relative tolerance in the `soft_zero` check in the optical Rayleigh interactor but absolute tolerance elsewhere, tracks could pass the orthogonality test in the interactor but trigger assertions later. This adds a `is_soft_orthogonal` helper function to ensure our test for orthogonality is consistent everywhere.

### DIFF
--- a/test/celeritas/optical/OpticalCollector.test.cc
+++ b/test/celeritas/optical/OpticalCollector.test.cc
@@ -336,12 +336,12 @@ TEST_F(LArSphereOffloadTest, TEST_IF_CELER_DEVICE(device_distributions))
 
     if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)
     {
-        EXPECT_EQ(3562272,
+        EXPECT_EQ(381086,
                   result.cherenkov.total_num_photons
                       + result.scintillation.total_num_photons);
 
-        EXPECT_EQ(38699, result.cherenkov.total_num_photons);
-        EXPECT_EQ(82, result.cherenkov.num_photons.size());
+        EXPECT_EQ(38126, result.cherenkov.total_num_photons);
+        EXPECT_EQ(78, result.cherenkov.num_photons.size());
         static unsigned int const expected_cherenkov_num_photons[] = {
             337u,  504u,  1609u, 1582u, 1314u, 1466u, 1164u, 880u, 777u,
             1477u, 1251u, 433u,  398u,  1273u, 271u,  1301u, 282u, 1132u,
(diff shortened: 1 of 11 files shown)


test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -336,12 +336,12 @@ TEST_F(LArSphereOffloadTest, TEST_IF_CELER_DEVICE(device_distributions))
 
     if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)
     {
-        EXPECT_EQ(3562272,
+        EXPECT_EQ(381086,
```
sethrj: Sorry, even though I thought about updating these I completely forgot by the time I merged 😞

sethrj (approved):
I think I started https://github.com/sethrj/celeritas/commit/b86a41889bbbf35ca45b2229520fc9c0603fcd54  when I was working on some of this a few months ago... but I got too sidetracked trying to make it perfect and ended up deferring it. Thanks for the simpler solution!

sethrj:
@amandalund I also forgot that I failed to track down an issue where the optical loop is incompatible with the Geant4 navigator in MT mode (presumably due to some stupid thread local issue), and it's causing an intermittent failure (and I didn't realize it was intermittent...):
```
[1/2] info@TrackingManagerIntegration.test.cc:344: Ran 0 over 0 step iterations from 0 flushes
```
https://github.com/celeritas-project/celeritas/actions/runs/17863479776/job/50801038112

Perhaps some sort of race condition? Either way, can you please apply this patch:
```
From 82cb5d99f2900ad59517e68dda25f7495e2b1547 Mon Sep 17 00:00:00 2001
From: Seth R Johnson <johnsonsr@ornl.gov>
Date: Fri, 19 Sep 2025 12:42:12 -0400
Subject: [PATCH] Disable MT optical physics test when using Geant4 navigator

---
 test/accel/CMakeLists.txt | 10 ++++++++--
 1 file changed, 8 insertions(+), 2 deletions(-)

diff --git a/test/accel/CMakeLists.txt b/test/accel/CMakeLists.txt
index c5c96c50f..f234018ea 100644
--- a/test/accel/CMakeLists.txt
+++ b/test/accel/CMakeLists.txt
@@ -88,11 +88,17 @@ celeritas_add_integration_tests(
   RMTYPE serial mt
 )
 
-# Test with optical physics
+set(_optical_rm_type "serial")
+if(NOT (CELERITAS_CORE_GEO STREQUAL "Geant4"))
+  list(APPEND _optical_rm_type "mt")
+endif()
+
+# Test with optical physics, except that currently trying to use Geant4
+# navigation for the optical tracks wreaks havoc
 celeritas_add_integration_tests(
   TrackingManagerIntegration LarSphereOptical run
   OFFLOAD cpu gpu g4
-  RMTYPE serial mt
+  RMTYPE ${_optical_rm_type}
 )
 
-- 
2.51.0
```

Maybe also you could do that, and do your two other unrelated unit test fixes, on a quick new branch?

amandalund:
Ah I wasn't sure what was going on there... thanks for that! Will do!

test/celeritas/optical/OpticalCollector.test.cc
amandalund: No worries, I forgot about the device tests too :)

PR #1973, opened 2025-09-19
Fix a few testing issues

amandalund:
- Update `OpticalCollector` device tests for #1970 
- Disable optical integration test when using Geant4 navigation in MT mode (see https://github.com/celeritas-project/celeritas/pull/1972#issuecomment-3312913674)
- Avoid narrowing warning with single precision build (`test/orange/OrangeJson.test.cc:788:56: error: narrowing conversion of ‘+Inf’ from ‘double’ to ‘float’ [-Wnarrowing]`)

### DIFF
--- a/test/accel/CMakeLists.txt
+++ b/test/accel/CMakeLists.txt
@@ -88,11 +88,17 @@ celeritas_add_integration_tests(
   RMTYPE serial mt
 )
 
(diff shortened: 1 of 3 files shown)


sethrj (approved):
Thanks!

Issue #1964, opened 2025-09-15
Unify and simplify importing of particles/processes to offload


stognini:
# Issue
The existing `GeantImporter` uses a small set of `GeantImportDataSelection::Flags` for importing particles and physics. This leads to inconsistencies between `SetupOptions` / `FrameworkInput` and the actually imported data: we may only have one process and one particle (e.g. Rayleigh scattering for gammas), but `GeantImporter` will still likely follow the `::em_basic` data selection for importing particles and processes.

# Requirement
Celeritas should allow for user-defined physics and particles and maintain consistency with the data loaded into memory for offloading. I.e. not load unused particles/processes.

# Implementation
First step is to update or remove existing `DataSelection` flags. Since the physics should come from the registered Geant4 physics list, we could loop over the offloaded list of particles and import any available process. The outer loop in `::import_processes` is already over particles, and thus updating/refactoring `DataSelection`/`ParticleFilter` may work already.

Breaking down `GeantImporter` into separate helpers, one for `inp::Problem::Model` and one for `inp::Problem::Physics` may make more sense for code logic (but is probably another issue).

## Caveats
Geant4 does not allow the creation of a given process without the minimum required particles (primary + secondaries) used by said process. Celeritas during offload **could** allow for the production of said secondary and not physically transport it. E.g. Geant4 has a set of processes for gammas, and only bremsstrahlung for electrons. If the only particle in the offload list is the electron: can the electron still undergo bremsstrahlung and produce a gamma, which is not offloaded, by
- killing the gamma and depositing the energy locally; or
- sending the gamma back to Geant4 for transport?

Another option is to require the offload of secondaries produced, but import the process data based on `inp::Physics` (which would be equivalent and likely supersede `GeantPhysicsOptions`). In this case, reusing the example above:
- Geant4: would have gamma (and its processes), e- with bremsstrahlung.
- Celeritas: would also have both particles, but could enable only bremsstrahlung through its physics options. In this case the produced gamma would have no processes attached to it and only be raytraced out of the geometry.

sethrj:
Some follow-on thoughts.

# Guiding principles

I think we want first and foremost, **Enabling Celeritas maintains consistent physics with Geant4**. Less important but still worth considering, as you mentioned is, **Minimize overhead by *not* converting data that's unnecessary at runtime**.

# Requirements

I think in your description we're missing the desired data flow:
- [In standalone mode only: JSON input determines Geant4 physics list]
[...]

PR #1976, opened 2025-09-21
Use inp data for more physics

sethrj:
Being completely honest, I have utterly forgotten about the technical hurdle that required these initial changes to the codebase. I started this a week ago by trying to get `inp::FrameworkInput` directly into `accel` and found that the problem physics and import data weren't playing well with the file-based readers. So, I have ended up making a cascade of changes (and I can try to separate these out if that helps). 

For optical physics offload (cherenkov/scintillation processes):
- From celer-sim, Cherenkov/scintillation is added to G4 physics list via user input
- Presence of C/S processes determines flags in ImportData
- C/S flags in ImportData, in addition to surface physics, are copied to Problem
- Problem's C/S flags build correspponding offload actions

Additional changes to ImportData:
- Add fluorescence to ImportData to dictate AtomicRelaxation
- Added interpolation which gets passed to AtomicRelaxation as well
- Delete deprecated units

Finally I made a number of `inp` data structures:
- Implement `setup::physics_from` to load from G4, ROOT, external data files
- Move process-specific data (livermore, SB, MuPP, relaxation) to `inp` data
  structures
- Use bool operators instead of `optional` for inp
- SB/MuPP/etc. now take `inp` data rather than readers

Toward #1964, #1900


src/celeritas/inp/PhysicsModel.hh
```diff
@@ -6,18 +6,30 @@
...
  * Seltzer-Berger bremsstrahlung model.
- *
- * \todo Move \c sb_data from celeritas::ImportData here.
  */
 struct SeltzerBergerModel
 {
+    //! Differential cross sections [(log MeV, unitless) -> millibarn]
+    std::map<AtomicNumber, inp::TwodGrid> atomic_xs;
+
+    //! TODO: macro xs tables
```
amandalund: Should those go in the process input?

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -44,18 +61,69 @@ struct MuBremsModel
...
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Pair production from muons.
+ */
+struct MuProductionModel
+{
+    //! Grid for sampling table
+    MuPairProductionEnergyTransferTable muppet_table;
```
amandalund, line 104: You had to do it...

src/celeritas/inp/PhysicsProcess.hh
```diff
@@ -20,20 +23,56 @@ namespace inp
...
+    MuBremsModel mu;
+
+    //! Whether process has data and is to be used
+    explicit operator bool() const { return sb || rel || mu; }
 };
+
 //---------------------------------------------------------------------------//
 /*!
- * Construct a physics process for electron/positron pair production.
+ * Construct a physics process for pair production from gammas.
```
amandalund:
```suggestion
 * Construct a physics process for electron/positron pair production.
```

src/celeritas/inp/Physics.hh
```diff
@@ -51,55 +56,44 @@ struct EmPhysics
...
+    //! Add custom user processes
     ProcessBuilderMap user_processes;
 };
 
 //---------------------------------------------------------------------------//
 /*!
  * Optical physics processes, options, and surface definitions.
+ *
+ * \todo Processes that apply only to EM particles (cherenkov, scintillation)
+ * will live in EmPhysics
```
amandalund: Can't these processes apply to particles other than EM?

src/celeritas/em/process/BremsstrahlungProcess.cc
```diff
@@ -24,20 +24,20 @@ namespace celeritas
...
                 ImportProcessClass::e_brems,
                 {pdg::electron(), pdg::positron()})
-    , load_sb_(std::move(load_sb))
+    , sb_input_(std::move(sb_input))
     , options_(options)
 {
     CELER_EXPECT(particles_);
     CELER_EXPECT(materials_);
-    CELER_EXPECT(load_sb_);
+    CELER_EXPECT(sb_input_);
```
amandalund, line 40: Is the SB data always present even if the model is not used (i.e. only relativistic bremsstrahlung is enabled)?

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -44,18 +61,69 @@ struct MuBremsModel
...
+    {
+        return !atomic_number.empty() && grids.size() == atomic_number.size();
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Pair production from muons.
+ */
+struct MuProductionModel
```
amandalund: Maybe `MuPairProductionModel`, or `MuPairProdModel`? Maybe it's clear, but to me it kind of sounds like a model producing muons.

src/celeritas/inp/PhysicsModel.hh
```diff
@@ -44,18 +61,69 @@ struct MuBremsModel
 /*!
  * Bethe-Heitler relativistic pair production from gammas.
  */
-struct BetheHeitlerModel
+struct BetheHeitlerProductionModel
```
amandalund, line 64: `BetheHeitlerPairProdModel`, or just keep `BetheHeitlerModel`? (e.g. we also have `SeltzerBergerModel`, not `SeltzerBergerBremsModel`)

src/celeritas/inp/Physics.hh
```diff
@@ -35,9 +35,14 @@ namespace inp
 struct EmPhysics
 {
     //! Bremsstrahlung process
-    std::optional<BremsProcess> brems{std::in_place};
+    BremsProcess brems;
```
amandalund: Could we use full names for the processes, i.e. `BremsstrahlungProcess`? (We already do for the rest)

src/celeritas/em/model/LivermorePEModel.cc
```diff
@@ -65,7 +64,11 @@ LivermorePEModel::LivermorePEModel(ActionId id,
     for (auto el_id : range(ElementId{materials.num_elements()}))
     {
         AtomicNumber z = materials.get(el_id).atomic_number();
-        insert_element(load_data(z));
+        auto iter = input.atomic_xs.find(z);
+        CELER_VALIDATE(iter != input.atomic_xs.end(),
+                       << "missing SB atomic xs for Z=" << z.get());
```
amandalund:
```suggestion
                       << "missing Livermore atomic xs for Z=" << z.get());
```

src/celeritas/phys/ProcessBuilder.cc
```diff
@@ -165,40 +144,33 @@ auto ProcessBuilder::build_eioni() -> SPProcess
 //---------------------------------------------------------------------------//
 auto ProcessBuilder::build_ebrems() -> SPProcess
 {
+    CELER_EXPECT(data_.seltzer_berger);
```
amandalund, line 147: Same here, is it always present?

src/celeritas/setup/Import.cc
```diff
@@ -6,46 +6,149 @@
...
+    };
+
+    //// PHOTOELECTRIC ////
+
+    if (have_process(ImportProcessClass::e_brems))
+    {
+        inp::SeltzerBergerModel sb_model;
+        sb_model.atomic_xs = load_data(SeltzerBergerReader{});
+        imported.seltzer_berger = std::move(sb_model);
+    }
```
amandalund, line 120: Ok, this answers my earlier question about whether the SB data is always loaded for bremsstrahlung. If we did want to only load it if the SB model is enabled, we could change this to `has_model(imported, ImportModelClass::e_brems_sb)` (free function in ImportData.hh).

amandalund (approved):
Just a few minor questions and comments, thanks for this!

src/celeritas/inp/PhysicsModel.hh
sethrj: Yes indeed they should.

src/celeritas/inp/Physics.hh
sethrj: Good point. So I wonder what's the best way to group physics processes... strong/weak/EM forces at play? Energy range (i.e., for HEP EM offloading: lower than hadron creation, higher than optical photon creation)? A grid {input particle group, output particle group}?

src/celeritas/em/process/BremsstrahlungProcess.cc
sethrj: Previously we expected the reader to be there even if the unused... but yes I agree it should be unnecessary if using only rel brems.

src/celeritas/setup/Import.cc
sethrj: Yeah that would be good. The current way retains the behavior of the old way... and there's more refactoring to be done on that front 🙃

amandalund: I guess the old way with the reader we wouldn't actually load the data unless the model was constructed... but yeah let's not worry about it now.

sethrj: I thought the old way ended up using the import adapter and loading inside the GeantImporter first?

amandalund: Oh of course, that's right!

PR #1978, opened 2025-09-22
Loosen restrictions and catch errors in `accel` initialization order

sethrj:
This allows the integration singletons to be created before or after the run manager (and to correctly use the number of threads in the logger even after the thread count is changed). It also prevents the setup options/static "offload" lists from accidentally creating particles before it's legal, which triggered a `PART002` failure in `G4ParticleTable::CheckReadiness()`.


amandalund (approved):
Nice, thanks @sethrj!

PR #1980, opened 2025-09-22
Add SurfaceModelView to optical surface physics

hhollenb:
This is a pared-down version of #1928 without the models.

The main purpose of this PR is to add `SurfaceModelView` which encapsulates relevant information for a single surface physics crossing. This relies upon the cached track `SubsurfaceDirection` (which is determined at the end of each step by the interaction sub-step). The `SurfaceModelView` may then be used by the track slot executor to determine which kernels apply to a given track, and by the models directly to get the appropriate data for a given surface. It also hides some of the awkward indexing to get the optical materials on either side of the interface.

I also kept the `BuiltinSurfaceModelBuilder` since it just builds fake models at the moment, but can be readily switched to real models when they're added.


src/celeritas/optical/action/detail/PropagateExecutor.hh
```diff
@@ -32,6 +32,11 @@ struct PropagateExecutor
 //---------------------------------------------------------------------------//
 CELER_FUNCTION void PropagateExecutor::operator()(CoreTrackView& track)
 {
+    if (track.surface_physics().is_crossing_boundary())
```
sethrj: I feel like this could be easily forgotten if we add new kernels that *aren't* surface physics. I think it could be better as an additional restriction for the ConditionalTrackSlotExecutor? Maybe split `make_action_thread_executor` to `make_volume_action_thread_executor` and the new executor in your current PR?

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -310,72 +324,67 @@ SurfacePhysicsView::subsurface_position(SurfaceTrackPosition pos)
...
+    CELER_EXPECT(step != SurfacePhysicsOrder::size_);
+
+    auto phys_surface = this->subsurface_interface(dir);
+    CELER_ASSERT(phys_surface);
+
+    return SurfaceModelView{
+        SurfacePhysicsMapView{params_.model_maps[step], phys_surface},
+        this->subsurface_material(this->subsurface_position()),
+        this->subsurface_material(advance_subsurface_position_along(
+            this->subsurface_position(), dir))};
```
sethrj: Pull out auto pos = `this->subsurface_position()` local so you can reuse it (and check that it's valid).

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -468,11 +480,58 @@ CELER_FUNCTION U SurfacePhysicsView::to_record_index(
...
+    auto pos_range = range(SurfaceTrackPosition{this->num_positions()});
+
+    // In pre-volume
+    if (pos == pos_range.front())
+    {
+        return states_.pre_volume_material[track_id_];
+    }
+    // In post-volume
+    if (pos == pos_range.back())
+    {
```
sethrj: Move comments inside the `if` blocks when they only apply to that scope

src/celeritas/optical/surface/SurfaceModelView.hh
```diff
@@ -0,0 +1,110 @@
...
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Optical surface data for a model.
+ *
+ * Wraps common behavior for querying the surface model data for a given
+ * physics surface interface.
```
sethrj: I'm not super convinced this class makes the code clearer/easier/more efficient since it doesn't do much:
- It  stores pre-calculated [^1] optical mat IDs on the two sides of the subsurface interface
- It forwards two functions to an existing view class

I think the clarity could be improved by also storing the "substep" (SurfaceActionOrder) since the view applies only to that. The utility of it is not much more than having PhysicsView return a {pre, post} material ID for the subsurface being crossed.  I guess the main benefit of this class is to combine in one place the subsurface crossing parameters that might be needed in the `Sampler` design in (#1928)?

**Since it's the main purpose of this PR and the later ones are built on it, let's keep the class, but please add some more documentation about its scope and purpose**

[^1]: By the way, even though the two optical mat ID values are expensive to calculate but not often used together (and in fact for the calculators in place, not used at all),  inlining seems to eliminate the cost of calculating them.

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -8,12 +8,18 @@
...
 
 namespace celeritas
 {
 namespace optical
 {
+//---------------------------------------------------------------------------//
+// TYPE ALIASES
+//---------------------------------------------------------------------------//
+
+using SurfaceTrackPosition = OpaqueId<struct SurfaceTrackPosition_>;
```
sethrj: Can you move this, and the other enums in SurfaceData,  to `OpticalTypes`.hh?

I missed previously (due to the size of #1801) that `SubsurfacePosition` is an OpaqueId but isn't called `Id`. Since it's not indexing into anything, maybe it shouldn't be an ID (we can discuss)? And can you also document what exactly its valid ranges are (I guess 0 to num_interfaces + 1?)

test/celeritas/CMakeLists.txt
```diff
@@ -32,8 +32,10 @@ endif()
 if(CELERITAS_CORE_GEO STREQUAL "ORANGE")
   set(_core_geo_libs testcel_orange Celeritas::orange)
 elseif(CELERITAS_CORE_GEO STREQUAL "VecGeom")
+  set(_needs_orange DISABLE)
   set(_core_geo_libs testcel_geocel ${VecGeom_LIBRARIES})
 elseif(CELERITAS_CORE_GEO STREQUAL "Geant4")
+  set(_needs_orange DISABLE)
```
sethrj: I think these are no longer necessary :)

src/celeritas/optical/surface/SurfaceModelView.hh
```diff
@@ -0,0 +1,110 @@
...
+ */
+class SurfaceModelView
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using InternalSurfaceId = SurfacePhysicsMapView::InternalSurfaceId;
```
sethrj: Regarding nomenclature for this ... maybe instead of trying to overload "surface" we can just refer to it as `LocalDataId`? `SubModelId`?  Man there's just too many choices. Since it doesn't have to correspond with a single surface in theory, let's just rename all this to `SubModelId` with the documentation "Internal index of physics data within a single model" (in theory it could be used for surfaces, materials, energy ranges, isotopes, ...) and put it in `celeritas/Types.hh`.

sethrj (commented):
Thanks a bunch for moving this to a separate PR! Let's move forward with only a couple small changes.

src/celeritas/optical/surface/SurfaceModelView.hh
amandalund: I agree, I'm not quite sure I understand the usefulness of this class, though maybe I need to dig deeper into #1928 ... it seems like everything could be accessed directly from the `SurfacePhysicsView`.

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -310,72 +324,67 @@ SurfacePhysicsView::subsurface_position(SurfaceTrackPosition pos)
...
  */
-CELER_FUNCTION SurfacePhysicsMapView SurfacePhysicsView::surface_physics_map(
-    SurfacePhysicsOrder step, PhysSurfaceId surface) const
+CELER_FUNCTION SurfaceModelView
+SurfacePhysicsView::surface_model(SurfacePhysicsOrder step) const
 {
-    return SurfacePhysicsMapView{params_.model_maps[step], surface};
+    auto dir = this->traversal_direction();
+    CELER_EXPECT(this->is_crossing_boundary());
+    CELER_EXPECT(!this->is_exiting(dir));
```
amandalund: We could probably remove these checks since they're redundant (already checked in `traversal_direction()` above and `subsurface_interface(dir)` below).

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -310,72 +324,67 @@ SurfacePhysicsView::subsurface_position(SurfaceTrackPosition pos)
...
-    CELER_ASSERT(this->subsurface_position().get() > 0);
-
-    auto material_record_id
-        = this->to_record_index(this->subsurface_position() - 1,
-                                this->surface_record().subsurface_materials);
-    CELER_ASSERT(material_record_id < params_.subsurface_materials.size());
-
-    return params_.subsurface_materials[material_record_id];
+    CELER_EXPECT(this->is_crossing_boundary());
+    return states_.track_direction[track_id_];
```
amandalund: I wonder if we should be calling this class `SurfacePhysicsTrackView` rather than `SurfacePhysicsView` since it depends on dynamic track data?

amandalund (commented):
Great work as always @hhollenb!

src/celeritas/optical/surface/SurfaceModelView.hh
hhollenb: I think the purpose of the class got a bit altered over time with some of the changes over the past PR. Before we weren't caching the track traversal direction and instead calculating it from the geometry direction. This class was then storing that result and calculating the appropriate model for the given step, and the optical materials from that direction. Now that we have the cached track direction, a lot of that becomes trivial calls.

The other benefit was that it pared down the surface physics functions so that models had a much simpler interface to use (e.g. no traversal). Effectively just "here's what's needed to run on a single subsurface interface".

hhollenb:
Made some changes following the discussion last Friday. 

Added `SurfaceTraversalView` which entirely handles the one dimensional traversal logic. Doesn't handle anything about how the record is laid out in memory (surface orientation) nor any of the geometry direction or normals. The important functions are:

- `position` - current track position
- `direction` - current track direction
- `next_position` - next track position following current direction

Change the old `SurfacePhysicsView` to become `SurfacePhysicsTrackView`. The new `SurfacePhysicsView` handles persistent data about a surface, and is mainly responsible for translating between the surface record and the track position. This means it needs each track's orientation on the surface, so it also takes a track slot ID.

`SurfacePhysicsTrackView` is mainly just a wrapper mapping calls between the traversal view and the physics view for the current track. Models should readily interface with this view for the relevant physics, and is able to construct the traversal and physics views from the track view if necessary. 

Finally, added some helpers for some of the indexing logic. `IfDirectionEquals` is a functor to hide the underlying boolean representation of `SubsurfaceDirection` in the cases where an offset needs to be sometimes calculated (and does it without branching). `OrientedItemMap` is just a wrapper around `ItemMap` that supports reverse iterating.

src/celeritas/optical/surface/detail/BuiltinSurfaceModelBuilder.hh
```diff
@@ -0,0 +1,147 @@
...
+        num_surf_ += layer_map.size();
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a fake surface model.
+ *
+ * A temporary utility to build fake surface models that have not yet been
+ * implemented.
```
sethrj, line 130: The word "placeholder" might be more precise (just as part of documentation, don't rewrite, since this is going to go away as it's just a placeholder 😉)

src/celeritas/optical/surface/SurfaceTraversalView.hh
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * This is equivalent to the number of interstitial sub-surface materials, plus
+ * the pre-volume and post-volumes.
+ */
+CELER_FUNCTION SurfaceTrackPosition::size_type
+SurfaceTraversalView::num_positions() const
+{
+    return params_.surfaces[states_.surface[track_id_]]
+               .subsurface_materials.size()
+           + 2;
```
sethrj, line 199: I would add a `\todo` here about testing the performance impact of caching this variable locally in the class as part of the constructor (and eliminating the params_ pointer) since it's used a lot and needs an indirection.

src/celeritas/optical/Types.hh
```diff
@@ -20,6 +20,15 @@ using ScintParticleId = OpaqueId<struct ScintParticle_>;
 //! Opaque index to a scintillation spectrum
 using ParScintSpectrumId = OpaqueId<struct ParScintSpectrum>;
 
+namespace optical
+{
+
+//! Opaque index into subsurface track position, in range [0, num subsurface
+//! interfaces + 1]
+using SurfaceTrackPosition = OpaqueId<struct SurfaceTrackPosition_>;
```
sethrj, line 28: Did we consider the pros and cons of making this an integer instead of an ID? (I think if we keep it as an OpaqueId type we should name it `SomethingId`. The current name is also confusing: it's a sub-surface position (not Surface), and it's no more associated to a track than any of the other IDs  (not Track).

src/celeritas/optical/action/detail/PreStepExecutor.hh
```diff
@@ -55,6 +55,11 @@ CELER_FUNCTION void PreStepExecutor::operator()(CoreTrackView const& track)
                  || sim.status() == TrackStatus::alive);
     sim.status(TrackStatus::alive);
 
+    if (track.surface_physics().is_crossing_boundary())
+    {
+        return;
+    }
+
```
sethrj: I think it'd be a bit more clear to wrap the rest of the function in an `if(!is_crossing)` since the remaining few lines are about updating the volumetric/bulk interactions.

src/celeritas/optical/surface/SurfacePhysicsTrackView.hh
```diff
@@ -0,0 +1,268 @@
...
+    inline CELER_FUNCTION void facet_normal(Real3 const&);
+
+    // Construct a traversal view for this track
+    inline CELER_FUNCTION SurfaceTraversalView traversal() const;
+
+    // Construct a constant physics view
+    inline CELER_FUNCTION SurfacePhysicsView surface() const;
+
+    // Access scalar data for surface physics
+    inline CELER_FUNCTION SurfacePhysicsParamsScalars const& scalars() const;
```
sethrj, line 94: It's a little weird to have views of views? (I thought our "global" track views were a special case.) But I think it's OK? I think you just want to make sure to note in the traversal/physics view documentation that the classes may have a very short lifetime and be rebuilt frequently.

src/celeritas/optical/surface/SurfacePhysicsTrackView.hh
```diff
@@ -0,0 +1,268 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get next subsurface material.
+ */
+CELER_FUNCTION OptMatId SurfacePhysicsTrackView::next_material() const
+{
+    return this->surface().material(this->traversal().next_position());
+}
+
```
sethrj, line 180: I think there are too many "helper" member functions.  See [this question regarding Effective C++](https://stackoverflow.com/questions/5989734/effective-c-item-23-prefer-non-member-non-friend-functions-to-member-functions): try to make class interfaces relatively thin.

What about making it a free function `OptMatId get_material(TraversalView const& , SurfacePhysicsView const&)`? The data it uses becomes much clearer.

src/celeritas/optical/surface/SurfacePhysicsTrackView.hh
```diff
@@ -0,0 +1,268 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate and update traversal direction from track momentum.
+ */
+CELER_FUNCTION void
+SurfacePhysicsTrackView::update_traversal_direction(Real3 const& dir)
+{
+    CELER_EXPECT(is_soft_unit_vector(dir));
+    this->traversal().direction(static_cast<SubsurfaceDirection>(
+        is_entering_surface(dir, this->global_normal())));
```
sethrj: What about a free function `SubsurfaceDirection get_direction(Real3 const& dir, SurfacePhysicsTrackView const&)`? Or maybe that function should just return `SubsurfaceDirection` in the first place?

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -28,6 +29,21 @@ is_entering_surface(Real3 const& dir, Real3 const& normal)
...
+/*!
+ * Get the next track surface position in the given direction.
+ *
+ * Type-safe operation to ensure direction is only added in track-local frames.
+ * Uses unsigned underflow when moving reverse (dir = -1) while on a
+ * pre-surface (pos = 0) to wrap to an invalid position value.
+ */
+CELER_FORCEINLINE_FUNCTION SurfaceTrackPosition
+advance_subsurface_position_along(SurfaceTrackPosition pos,
+                                  SubsurfaceDirection dir)
```
sethrj: Didn't we use `next` in some place? I think it's less ambiguous than `advance` which sounds like it might be modifying an argument.

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -67,6 +83,71 @@ class EnteringSurfaceNormalSampler
...
+ */
+template<SubsurfaceDirection D, class IntType>
+struct IfDirectionEquals
+{
+    IntType value;
+
+    //! Return value if direction matches template parameter
+    inline CELER_FUNCTION IntType operator()(SubsurfaceDirection dir) const
+    {
+        return static_cast<IntType>(D == dir) * value;
```
sethrj: Multiplication isn't usually better than branching... and it's not very clear from the implementation, documentation, and usage what exactly IntType and value are supposed to be, or what the function does...

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -67,6 +83,71 @@ class EnteringSurfaceNormalSampler
...
+  public:
+    //! Construct from ItemMap and orientation
+    inline CELER_FUNCTION
+    OrientedItemMap(ItemMap<T, U> const& map, SubsurfaceDirection orientation)
+        : map_(map), orientation_(orientation)
+    {
+    }
+
+    // Map surface track position, taking orientation into account
+    inline CELER_FUNCTION U operator[](T pos) const
```
sethrj: Shouldn't `pos` be of type `SurfaceTrackPosition`?

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -67,6 +83,71 @@ class EnteringSurfaceNormalSampler
...
+//---------------------------------------------------------------------------//
+/*!
+ * Wrapper for an \c ItemMap that can be oriented.
+ *
+ * If the map is forward-oriented, then it behaves the same as an \c ItemMap.
+ * If the map is reverse-oriented, then it behaves the same as a reverse
+ * iterator over \c ItemMap.
+ */
+template<class T, class U>
+class OrientedItemMap
```
sethrj: This class is never meant to be stored (it has a const reference as member data), or indeed have much lifetime or reuse.  Previously it was a member function, but (hate to sound like a broken record) wouldn't it be better as a free function?

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -19,12 +17,11 @@ namespace optical
 {
 //---------------------------------------------------------------------------//
 /*!
- * Optical surface physics data for a track.
+ * Persistent optical surface physics data for a track.
```
sethrj: In the EM loop we have "step view" and "track view" for persistent track-dependent data that's ephemeral versus long-lived (more or less, anyway). ParticleView depends on particle ID, not track slot ID; and ParticleTrackView depends on track slot ID (and has state). 

I think we need to make this class a true "surface physics view" that encapsulates only params-related data for a single surface. State-related data moves to SurfacePhysicsTrackView. We can add a `SurfaceCrossingTrackView` (or whatever; some more "ephemeral" state) for properties that only matter when entering/interacting with a surface.

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -219,260 +100,60 @@ CELER_FUNCTION SubsurfaceDirection SurfacePhysicsView::orientation() const
...
-{
-    return this->surface_record().subsurface_materials.size() + 2;
-}
-
 //---------------------------------------------------------------------------//
 /*!
  * Return the subsurface material ID of the current track position.
  */
-CELER_FUNCTION OptMatId SurfacePhysicsView::subsurface_material() const
+CELER_FUNCTION OptMatId SurfacePhysicsView::material(SurfaceTrackPosition pos) const
```
sethrj: This is an excellent example of something that needs to be a member function (since it consumes private data like the pre/post volume material). Since it depends not only on the "surface" definition but also the *state* information, because we have to cache the pre/post material, this should be part of the CrossingView.

src/celeritas/optical/surface/SurfaceTraversalView.hh
```diff
@@ -0,0 +1,235 @@
...
+    inline CELER_FUNCTION bool in_post_volume() const;
+
+    // Whether track is currently moving off the surface
+    inline CELER_FUNCTION bool is_exiting() const;
+
+    // Whether track is in pre-/post-volume and direction is off the surface
+    inline CELER_FUNCTION bool is_exiting(SubsurfaceDirection) const;
+
+    // Position of the track in the surface crossing
+    inline CELER_FUNCTION SurfaceTrackPosition position() const;
```
sethrj: Our geo track views use pos/dir. Should we be consistent with that, or try to distinguish the radically different type by giving it a different name? Since the semantics are similar I think the former would be better.

src/celeritas/optical/surface/SurfaceTraversalView.hh
```diff
@@ -0,0 +1,235 @@
...
+    inline CELER_FUNCTION SurfaceTrackPosition::size_type num_positions() const;
+
+    // Current track traversal direction
+    inline CELER_FUNCTION SubsurfaceDirection direction() const;
+
+    // Set track traversal direction
+    inline CELER_FUNCTION void direction(SubsurfaceDirection);
+
+    // Cross subsurface interface in the given direction
+    inline CELER_FUNCTION void cross_interface(SubsurfaceDirection);
```
sethrj, line 76: Should we say `cross_subsurface` and rename the existing geometry method `cross_boundary` to `cross_surface`? (I think this would improve code clarity and we can do it later.)

src/celeritas/optical/surface/SurfaceTraversalView.hh
```diff
@@ -0,0 +1,235 @@
...
+#include "SurfacePhysicsData.hh"
+#include "SurfacePhysicsUtils.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Manage one-dimensional logic for traversing an optical surface.
```
sethrj, line 21: I'd add a few bullets about how the traversal proceeds over the surface crossing:
- *Always* starts at position 0 in the forward direction and does not relate to the surface data layout
- The position corresponds to the surface layers

An ASCII art drawing of the pre/post volume, traversal direction, etc. would be extremely helpful.

Might also be worth mentioning that the only params data needed are the number of subsurfaces in the surface being crossed.

src/celeritas/optical/surface/SurfaceTraversalView.hh
```diff
@@ -0,0 +1,235 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ */
+CELER_FUNCTION SurfaceTrackPosition SurfaceTraversalView::next_position() const
+{
+    return advance_subsurface_position_along(this->position(),
+                                             this->direction());
+}
```
sethrj: Another good candidate for a free function: `SurfaceTrackPosition  next_position(SurfaceTraversalView const&)` (and in fact rename `advance_subsurface_position_along`)

sethrj (commented):
Finally done with the review 😪 sorry it took so long, I know it's not easy to make a small PR when everything is so experimental/up in the air/being fleshed out as we go.

Happy as always to discuss over zoom, or open a smaller PR with a subset of changes (only if you really want to, since I think there would be a lot in here to disentangle), etc.

src/celeritas/optical/Types.hh
hhollenb: Calling it an ID would be confusing since it's really a position. Maybe `SubsurfacePosition`?

I'd prefer it to be strongly typed so I can read off a function signature and see it requires a `SubsurfacePosition` and not some kind of offset or mixing it with some other IDs or anything.

src/celeritas/optical/surface/SurfacePhysicsTrackView.hh
hhollenb: I thought that short lifetime and rebuilt frequently was assumed for all of the views? We construct sim / physics / geo views quite often and freely.

These are also mainly useful for testing right now, I think most other code will exclusively use the `SurfacePhysicsTrackView`? They're effectively two track views with independent responsibilities, and the track view is a wrapper coordinating their functionality for a single track.

src/celeritas/optical/surface/SurfacePhysicsTrackView.hh
hhollenb: I'm not sure how much that makes sense? For any model in a single physical surface crossing, it needs the current optical material and next optical material. If you have a `TraversalView` and `SurfacePhysicsView` then you can just write `surface.material(traversal.position())` and `surface.material(traversal.next_position())`, but this also requires you now handling two views, making sure they refer to the same underlying track/surface, and copying all that logic to every time it's needed in a model.

For a track view, it encapsulates precisely this coordination and the logic relevant for models. I'm worried that exposing too much of the underlying surface logic is going to bloat the core track view and models and make everything more tightly coupled to how we handle it.

sethrj (approved):
Awesome! This is truly great. On to the next thing!

PR #1981, opened 2025-09-23
Add iostream wrappers for Geant physics options

sethrj:
Tiny change pulled from #1962 that allows us to use stream operators to read and write IO options.


amandalund (approved):
Thanks!

PR #1983, opened 2025-09-23
Approximate two-point scintillation spectra as a gaussian

### LINKED ISSUE #1974: Scintillation spectrum variance calculation fails for two-point input
When importing scintillation spectra, `MomentCalculator` in`GeantImporter.cc` is used to estimate the mean and standard sigma   for  the photon wavelength distribution which then impacts the generation of photons
It returns `variance = 0` for case when there are only two points in the grid.
Could be reproduced in `calc_moments::PdfUtilsTest.cc` by setting a grid with 
```cpp
x = {2.034 , 4.136 };
f = {1.0, 1.0};
```

Rashika-Gupta:
The following PR:
1. Two-point scintillation spectra can trigger errors such as
```cpp
issued by : src/celeritas/optical/gen/detail/MatScintSpecInserter.hh:85
Celeritas runtime error: invalid lambda_sigma=0 (should be positive)
```
To address this, approximate the mean as the midpoint energy and set $\sigma = \frac{E_{max} - E_{min}}{2.355}$ 
2. Logs a warning to indicate that this approximation is applied since the input does not define a full distribution.


sethrj (commented):
Looking at this I'm a little confused about what we're reading from Geant4. @amandalund @whokion is the scintillation material property a *spectrum* or a *piecewise constant PDF* or a *piecewise linear PDF* (which I think is what this is assuming)?

This also needs a test. (I've forgotten my statistics but I feel like a line segment shouldn't have a variance of zero... and shouldn't it give different variance if it's constant versus linear, which this implementation doesn't distinguish between?)

I think also the special case should apply only for scintillation where we're using these to approximate a gaussian, not in the general case.

whokion:
In Geant4, the scintillation components (input material property vectors) are the relative intensities of the emission spectrum, which are piece-wise constant functions of the outgoing optical photon energy. These are internally
converted into piece-wise linear distributions for sampling (i.e., intensity vectors). Note that in version 11.3, the distributions are not normalized, but this will be normalized (i.e. c.d.f) in version 11.4.

amandalund:
> I've forgotten my statistics but I feel like a line segment shouldn't have a variance of zero... and shouldn't it give different variance if it's constant versus linear, which this implementation doesn't distinguish between?

The estimate of the mean/variance does approach that of a uniform distribution as the number of grid points increases: https://github.com/celeritas-project/celeritas/blob/5ec35c333b62cdeb8ef20c7e3c6cd4451b7df126/test/corecel/math/PdfUtils.test.cc#L62-L84
but yeah with just two points it's no good.

I agree about the constant vs. linear here, though I'm not sure it's worth worrying about since this is a workaround until we support the grid-based sampling and we're not likely to run into this in realistic applications.

Rashika-Gupta:
> This also needs a test. (I've forgotten my statistics but I feel like a line segment shouldn't have a variance of zero... and shouldn't it give different variance if it's constant versus linear, which this implementation doesn't distinguish between?)

Right now `MomentCalculator` can return `variance = 0` when there are only two sample points, since it uses a midpoint evaluation of a single trapezoid. That effectively collapses to a delta at the midpoint.  

The current fallback assumes a uniform distribution across `[emin, emax]`, motivated by the Geant4[opnovice](https://github.com/Geant4/geant4/blob/a499fb82e9d3310f83c34f52806127561bdaab40/examples/extended/optical/OpNovice/src/OpNoviceDetectorConstruction.cc#L102)
 where  
```cpp
G4double scintilFastArray[]{1.0, 1.0};
G4double energyArray[]{2.034 * eV, 4.136 * eV};
G4int lenArray = 2;
```
To avoid confusion can set the variance to sigma = (emax - emin)/2.355) as discussed in #1974 

> I think also the special case should apply only for scintillation where we're using these to approximate a gaussian, not in the general case.

I will make that change

sethrj:
Ah, the variance is *evaluated* at the midpoint. That's what I was missing. I feel like it wouldn't be difficult to replace the midpoint evaluation with a trapezoidal one? It should be almost trivial to analytically integrate `x^2 f(x)` for piecewise constant `f(x)`. The mean and variance shouldn't depend on the number of points if the shape of the PDF is identical... But that's a pretty unimportant diversion that we shouldn't invest further time into. @amandalund should we bother adding a test for the N=2 case with constant f(x) and N=2 with *non-constant* f(x)? Or are we going to delete this soon enough if/when we add grid sampling?

That aside, I think it's ok to do what's there now.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -255,6 +255,22 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
+                    {
+                        // This case is triggered when Geant4 provides only two
+                        // points for a scintillation component (so the
+                        // variance from MomentCalculator is zero). We
+                        // approximate the distribution as a Gaussian centered
+                        // at the midpoint with sigma ~ FWHM/2.355.
+                        double emin = grid.x.front();
+                        double emax = grid.x.back();
+                        comp.lambda_mean = (emax + emin) / 2;
+                        comp.lambda_sigma = (emax - emin) / 2.355;
```
sethrj: Even though you explained what 2.355 is (so it's no longer just a "magic" number!) it provides additional clarity to use as high-precision values as possible for mathematically derived constants. (If we didn't know this value to more than 4 digits of precision, as is the case for our "experimental constants" like the electron mass, that would be fine.) So please replace with `2.3548200450309493` (and you can even add a comment that this is  2 * sqrt(2 * log(2))).

sethrj (approved):
By the way, adding your CMakeUserPresets to the Celeritas repo is not only allowed but encouraged. If we didn't want them to be committed, we'd have a gitignore file in place. If you notice any other annoyances (such as having to constantly remove/ignore your cmake presets file) in your development workflow, please bring them up and we can figure out how to adapt to your needs because you're likely not the only one having that issue! 😄

amandalund (approved):
Thanks @Rashika-Gupta!

@sethrj I think it wouldn't hurt to add a couple simple tests for the two grid point case. That update to the moment calculator would definitely be a straightforward improvement, but yeah better to focus on more important things now... maybe some weekend or vacation. :)

PR #1984, opened 2025-09-23
Add Fresnel reflectivity, refraction, and UNIFIED reflection calculators

hhollenb:
~This is just a draft to outline the API of the calculators and the appropriate tests. Actual implementations will be added later when the PR is opened. Any feedback you have on the API / tests is welcome, but no rush!~

These are the calculators for analytically calculating surface interactions between dielectrics and metals. This code is entirely functional and doesn't depend on how models are being laid out in the other PRs. The dielectric-dielectric and dielectric-metal models should be readily composed of these calculators (Fresnel reflectivity --> Fresnel refraction / UNIFIED reflection). Writing the interactors will probably rely on the `SurfaceModelView`.


src/celeritas/optical/surface/model/SurfaceInteraction.hh
```diff
@@ -0,0 +1,77 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Representation of a photon's direction and polarization as a phasor.
+ */
+struct PhotonPhasor
+{
+    Real3 direction;
+    Real3 polarization;
```
sethrj: To reduce register pressure we may want to make these const references? (Instead of doing it now, add a to-do for testing that assumption: we should test whether the compiler is able to optimize out the local storage, whether the kernel occupancy actually changes, and whether there's any actual measurable performance improvement from it... I'm guessing the answer to the last is 'no')

src/celeritas/optical/surface/model/SurfaceInteraction.hh
```diff
@@ -0,0 +1,77 @@
...
+
+    //! Whether the phaser is a valid photon state
```
sethrj:
```suggestion
    //! Whether the phasor is a valid photon state
```

src/celeritas/optical/surface/model/SurfaceInteraction.hh
```diff
@@ -0,0 +1,77 @@
...
+    explicit CELER_FUNCTION operator bool() const
```
sethrj: This operation is expensive enough that I think it should be called something else (it should be used only for error checking, not for program logic).

src/celeritas/optical/surface/model/LambertianDistribution.hh
```diff
@@ -0,0 +1,66 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample a diffuse Lambertian distribution of reflected vectors about a
+ * normal.
+ *
+ * Lambertian reflectance is an approximation of a diffuse material where the
+ * apparent brightness is equal for observers at all angles. Reflected vectors
+ * follow Lambert's cosine law, which states the intensity of reflected light
+ * is proportional to the cosine of the reflection angle.
```
sethrj: Great documentation! An equation/reference could help it even more.

src/celeritas/optical/surface/model/LambertianDistribution.hh
```diff
@@ -0,0 +1,66 @@
...
+#include "celeritas/optical/Types.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Sample a diffuse Lambertian distribution of reflected vectors about a
+ * normal.
```
sethrj:
```suggestion
 * Sample a reflected vector from a diffuse Lambertian distribution.
```

src/celeritas/optical/surface/model/SurfaceInteraction.hh
```diff
@@ -0,0 +1,77 @@
...
+#include "corecel/math/ArrayUtils.hh"
+#include "corecel/math/SoftEqual.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Representation of a photon's direction and polarization as a phasor.
```
sethrj:
```suggestion
 * Representation of a direction and polarization as a phasor (phase vector).
```
Is this updated description right? And is this nomenclature commonly used in the field?

src/celeritas/optical/surface/model/UnifiedReflectionCalculator.hh
```diff
@@ -0,0 +1,181 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate geometric reflection of an incident vector about a normal.
+ */
+inline CELER_FUNCTION Real3 geometric_reflection(Real3 const& dir,
+                                                 Real3 const& normal)
+{
+    return dir - 2 * dot_product(dir, normal) * normal;
```
sethrj: This should be written using axpy/fma (using `a = -2 * dot_product(dir, normal)`), not using array operators (which are inefficient in their generation of temporaries).

src/celeritas/optical/surface/model/UnifiedReflectionCalculator.hh
```diff
@@ -0,0 +1,181 @@
...
+
+    // Calculate specular lobe reflection
+    inline CELER_FUNCTION PhotonPhasor specular_lobe() const;
+
+    // Calculate back-scattering reflection
+    inline CELER_FUNCTION PhotonPhasor back_scattering() const;
+
+    // Sample diffuse Lambertian reflection
+    template<class Engine>
+    inline CELER_FUNCTION PhotonPhasor lambertian_reflection(Engine& rng) const;
```
sethrj: Prefix these with `calc_` for operations that don't rely on random numbers but do a fair bit of work, and `sample_` for the ones that take the RNG

src/celeritas/optical/surface/model/UnifiedReflectionCalculator.hh
```diff
@@ -0,0 +1,181 @@
...
+ * The model specifies 4 different reflection modes:
+ *  1. Specular spike: geometric reflection about global normal
+ *  2. Specular lobe: geometric reflection about facet normal
+ *  3. Back-scattering: reversed photon direction and polarization
+ *  4. Diffuse Lambertian: reflection following Lambert's cosine law
+ *
+ * Only one reflection mode is selected based on the provided list of
+ * probabilities.
+ */
+class UnifiedReflectionCalculator
```
sethrj: Call it `Sampler` since it's choosing one instance from a distribution using an RNG.

sethrj (commented):
A few comments but I think this is good. One question to ask is whether we want the same interface for all the "samplers" of a kind, even if they don't do any sampling. I want to say the answer is yes: if we define `random/generators/DeltaDistribution.hh` it should damn well take an RNG (but never use it).

As an aside, not quite related to this PR, I think we might start moving away from a two-class "Executor + Interactor" model toward one model that has a constructor with all the data the class *actually* needs (good for testing and documentation) with a constructor that takes a higher-level class like the SurfacePhysicsView and uses a delegating constructor to load the data it needs.

src/celeritas/optical/surface/model/SurfaceInteraction.hh
hhollenb: I've seen it in Jackson E&M and Fowles' Optics, so it's definitely associated with classical EM plane waves. It should also be something readily understood coming from different fields just from it's mathematical definition: https://en.wikipedia.org/wiki/Phasor

hhollenb: Generally I want to pass around the term

$$\vec{E} e^{i\vec{k} \cdot \vec{r}}$$

which consists of orthogonal unit vectors, and wanted to try avoiding mixing them up in function arguments, especially with also the normals (more unit vectors) being passed around. I'm open to a better term, this is just the first one that came to mind

src/celeritas/optical/surface/model/SurfaceInteraction.hh
hhollenb: I was thinking of using this class to possibly handle intermediate calculations or results for some of the larger calculations, so I made them values instead of references. If we want to make them const refs then that would be better to do that now rather than later. It'd also complicate the current `UnifiedReflectionSampler` return values and `SurfaceInteraction` interface

hhollenb:
Did some reorganizing and drafted out the `DielectricInteractor`. This becomes pretty simple to handle reflectivity and both the dielectric-dielectric and dielectric-metal cases. I think this should have the best thread utilization on the GPU since all dielectric and metal interfaces will do the `FresnelCalculator` to begin with, then they will share the `UnifiedReflectionSampler` if relevant, and if not they'll do their appropriate refraction mode.

I'm thinking we may want to (in the future) change the `SurfacePhysicsOrder::reflectivity` to basically be the grid user override in Geant4 of absorption/reflection/refraction/transmission, but for a simple first pass this looks like a way we can get an interactor in the code with full physics.

hhollenb:
Quick overview of some of the changes:

- Decoupled the reflection calculators and samplers for easier testing.
- Added surface interaction applier and updated launch action calls in the dielectric models. Can probably make this cleaner?
- Changed surface crossing status from having a surface ID to being on the surface stepping action. This helps avoid edge cases where a track is exiting the surface but still on it, which means you can't look-up the next surface interface model. 
- Integration tests (including OpticalCollector and PrimaryGenerator in `optical`) now run through the surface stepping loop and the relevant models. The `PostSurfaceExecutor` still kills the tracks when they move off the surface.

sethrj:
@hhollenb Working on this next. Is the description up to date? Is there also any chance of quickly splitting of a subset of this closer to ~500 lines of code to review? 😅

hhollenb:
Ok I pulled out most of the framework code so now it's just the calculators and the tests. I might be able to split out some micro-PRs if that helps, but I'll probably need to change the order of which ones get reviewed.

I should also note that for the Fresnel reflectivity and refraction calculators, those will need to be rewritten when we want to support complex indices of refraction since those equations will no longer be valid. The UNIFIED reflection calculators should stay the same though.

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -57,6 +58,16 @@ calc_subsurface_direction(Real3 const& geo_dir, Real3 const& normal)
         is_entering_surface(geo_dir, normal));
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate geometric reflection of an incident vector about a normal.
+ */
+inline CELER_FUNCTION Real3 geometric_reflection(Real3 dir, Real3 const& normal)
```
sethrj: Maybe if we named it `reflected_from` it would imply "[return] reflected [direction] from [normal]"? And a `[[nodiscard]]` decorator will help ensure no one expects it to modify `dir`.

src/celeritas/optical/surface/model/DielectricInteractor.hh
```diff
@@ -0,0 +1,151 @@
...
+  public:
+    struct Builder
+    {
+        NativeCRef<DielectricData> dielectric_data;
+        NativeCRef<UnifiedReflectionData> unified_data;
+
+        // Build and sample interactor for a track
+        inline CELER_FUNCTION SurfaceInteraction
+        operator()(CoreTrackView const&) const;
+    };
```
sethrj, line 44: At first I didn't understand this struct but I see it's necessary to copy the data to GPU through the "parallel for" launch arguments. I completely forgot when I made my "let's just add another constructor argument" that we needed to pass additional data.

Here's another idea (and it might be more useful for us all to chat this over rather in slack/person than doing it over github comments): what if we have a `struct ExecutionData { ... };`, and a standardized constructor  `Interactor(CoreTrackView const&, ExecutionData const&)`? Then we might be able to have a helper function `apply_interaction_kernel<FooInteraction>(params, state, ExecutionData{...})`?

src/celeritas/optical/surface/model/FresnelCalculator.hh
```diff
@@ -0,0 +1,321 @@
...
+    , normal_(normal)
+    , relative_r_index_(relative_r_index)
+{
+    CELER_EXPECT(is_soft_unit_vector(direction_));
+    CELER_EXPECT(is_soft_unit_vector(polarization_));
+    CELER_EXPECT(is_soft_unit_vector(normal_));
+    CELER_EXPECT(relative_r_index > 0);
+    CELER_EXPECT(is_entering_surface(direction_, normal));
+
+    // \todo Check why the dot product is slightly greater than 1 sometimes
```
sethrj: @Rashika-Gupta has seen similar issues I think? It could be just from slightly nonnormalized directions. I think we can guarantee that the direction and normal are consistent, so I believe we can replace the `clamp` with a `max(..., real_type{1})`.

test/celeritas/optical/Fresnel.test.cc
```diff
@@ -0,0 +1,449 @@
...
+static LinearPolarization const TE{1, 0};
+static LinearPolarization const TM{0, 1};
+
+struct CoordinateAxes
+{
+    real_type rel_r_index;
+
+    Real3 n_hat;
+    Real3 s_hat;
+    Real3 p_hat;
```
sethrj: Even though this is a unit test I think these could use a little extra documentation. Cool idea for a class though! 

Maybe your basis functions could be represented better as a `SquareMatrix<real_type, 3>` (currently only in ORANGE)? For geometry reasons I was literally just thinking of adding an `is_soft_orthonormal` function which would work for this use case as well.

src/celeritas/optical/surface/model/ReflectionFormCalculator.hh
```diff
@@ -6,69 +6,173 @@
...
+#include "SurfaceInteraction.hh"
 
 namespace celeritas
 {
 namespace optical
 {
 //---------------------------------------------------------------------------//
 /*!
- * A view into UNIFIED model grids to calculate reflection mode probabilities.
+ * Calculator for UNIFIED reflection model.
```
sethrj, line 22: We should note that this is no longer about combining the weighted probabilities into a single angular distribution: it's a helper to sample the individual components which are *selected* by the reflection form grids.

src/celeritas/optical/surface/model/ReflectionFormSampler.hh
```diff
@@ -0,0 +1,136 @@
...
+        case ReflectionMode::specular_spike:
+            return calc_reflection_.calc_specular_spike();
+        case ReflectionMode::specular_lobe:
+            return calc_reflection_.calc_specular_lobe();
+        case ReflectionMode::backscatter:
+            return calc_reflection_.calc_backscatter();
+        case ReflectionMode::size_:
+            return calc_reflection_.sample_lambertian_reflection(rng);
+        default:
+            CELER_ASSERT_UNREACHABLE();
```
sethrj: Since `default` has the semantics of "everything else that's left`, and `size_` is more about "how many options do we have", I think it would be better to do:
```suggestion
        default:
            // The other probabilities are allowed to sum to less than unity: the remainder is diffuse
            return calc_reflection_.sample_lambertian_reflection(rng);
```
Does that produce compiler warnings/errors?

sethrj (commented):
I've gotten 90% of this reviewed but it looks great! Love the thorough Fresnel test.

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
hhollenb: I'd like to keep geometric / Euclidean in the name so it doesn't get confused with any other type of reflection. Maybe `geometric_reflected_from`?

src/celeritas/optical/surface/model/DielectricInteractor.hh
hhollenb: Agreed. I'd suggest pulling some of the other roughness / reflection models out of the old drafts to check their cases as well.

test/celeritas/optical/Fresnel.test.cc
hhollenb: If you have a matrix rep then you could also check the matricies are orthogonal $$ O^T O = 1 $$ which might have some nice BLAS-like matrix optimizations

src/celeritas/optical/surface/model/ReflectionFormSampler.hh
hhollenb: My thought is that since we're using `size_` implicitly to mean Lambertian reflection, this makes it explicit at least one place in the code and changes to the enum will get caught by the default.

Looks like the change doesn't produce any warnings / errors

src/celeritas/optical/surface/model/FresnelCalculator.hh
hhollenb: Doing some debugging, the hexfloat output of
```
cos_theta_ = -dot_product(direction_, normal_);
```
 is
```
cos_theta: 0x1.0000000000001p+0
```
wrapping the direction and normal with `make_unit_vector` gives the same value. I put in a check for it being the next floating point representation and added the `min` so we should see if it ever deviates more than that.

src/celeritas/optical/surface/model/DielectricInteractor.hh
sethrj: Since this isn't tested in the PR (but it's a good demonstration case) we could postpone the structural discussion until our next meeting and just get the fresnel stuff in?

sethrj (commented):
Still working but perhaps @whokion could check over the fresnel code?

src/celeritas/optical/surface/model/ReflectionFormSampler.hh
sethrj: Good point; how about in the enum we add an alias for the diffuse lobe, so that `size_` gets used correctly for the EnumArray/range, and you can use `diffuse_lobe` in the switch above?
```c++
...,
backscatter,
diffuse_lobe,
size_ = diffuse_lobe,
};
```

src/celeritas/optical/surface/model/DielectricInteractor.hh
hhollenb: I forgot to take out the temporary executor after splitting up the PRs. Now it's running in the integration tests at the very least.

src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
```diff
@@ -70,6 +70,13 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
     }
 
     OptMatId post_volume_material = track.material_record().material_id();
+    // If pre-volume and post-volume materials are the same, there's no surface
+    // interaction
+    if (pre_volume_material == post_volume_material)
```
sethrj: This assumes that the user doesn't have coatings (or optical surfaces that are supposed to represent coatings). For example, you could have two plastic scintillators bonded by an epoxy, and the required physics would show up as an interface surface.

There's also the question of, if users can arbitrarily change the default physics, what if they add a custom scattering kernel at the default boundary?

hhollenb: That's a good point, maybe the check can go into the dielectric interactor directly. Then it's handled as a physical surface crossing instead of a geometry surface crossing and would be correct for (admittedly the unusual case) of having two interstitial materials that are the same R index

hhollenb:
## Failing optical PrimaryGenerator test results

build-spack / vecgeom-g4 with g4 version $$\geq$$ 11.0
```
[ RUN      ] LArSpherePrimaryGeneratorTest.primary_generator
/home/runner/work/celeritas/celeritas/src/celeritas/ext/GeantImporter.cc:1038: error: Cannot export unknown process 'OpBoundary' (RTTI: G4OpBoundaryProcess)
/home/runner/work/celeritas/celeritas/test/celeritas/optical/PrimaryGenerator.test.cc:166: Failure
  106584
  result.steps
    Which is: 106552
```

build-docker / vecgeom-{debug,reldeb}-rocky-cuda
```
[ RUN      ] LArSpherePrimaryGeneratorTest.primary_generator
/__w/celeritas/celeritas/src/celeritas/ext/GeantImporter.cc:1038: error: Cannot export unknown process 'OpBoundary' (RTTI: G4OpBoundaryProcess)
/__w/celeritas/celeritas/test/celeritas/optical/PrimaryGenerator.test.cc:166: Failure
  106584
  result.steps
    Which is: 106838

/__w/celeritas/celeritas/test/celeritas/optical/PrimaryGenerator.test.cc:167: Failure
  36
  result.step_iters
    Which is: 34
```

These are after adding the same optical-volume check. Note that the `lar-sphere` GDML file has a sphere of LAr, surrounded by two shells of LAr which are labelled as sensitive detectors, and then outside is no optical material. Adding a check in dielectric interactor to see if the post material is invalid.

hhollenb:
Looks like the error is with the vecgeom normal calculations. In an ORANGE + G4 11.0.2 build, I produced some debugging output:
```
info:     pos: {88.4732,38.0064,26.9818}, dir: {-0.929895,-0.260691,-0.25949}, normal: {0.884732,0.380064,0.269818}
info:  Track 2732: 4 --> 2
info:     pos: {-22.057,88.3274,41.3735}, dir: {-0.367865,0.910641,-0.18817}, normal: {0.22057,-0.883274,-0.413735}
info:  Track 3651: 4 --> 2
info:     pos: {36.9625,42.0776,-82.8447}, dir: {0.811656,0.474604,-0.340537}, normal: {-0.369625,-0.420776,0.828447}
info:  Track 2030: refracted (1)
info:  Track 2732: refracted (1)
info:  Track 3651: refracted (1)
info:  Track 2030: moved to 4
info:  Track 2732: moved to 2
info:  Track 3651: moved to 2
info:  Track 3290: 3 --> 4
info:     pos: {-0.573875,-22.9944,-97.3187}, dir: {-0.527635,-0.34029,0.778334}, normal: {-0.00573875,-0.229944,-0.973187}
info:  Track 3290: refracted (1)
info:  Track 3290: moved to 4
info:  Track 3290: 4 --> 3
info:     pos: {-0.982981,-22.9121,-97.3348}, dir: {-0.274585,0.505989,-0.817666}, normal: {0.00982981,0.229121,0.973348}
info:  Track 3917: 4 --> 3
info:     pos: {24.5657,-85.4098,45.8443}, dir: {0.551915,-0.830253,0.0779121}, normal: {-0.245657,0.854098,-0.458443}
info:  Track 3290: refracted (1)
info:  Track 3917: refracted (1)
info:  Track 3290: moved to 3
info:  Track 3917: moved to 3
```
You can see the normal vector is a unit vector of the position. While in a vecgeom + g4 build I got
```
info:  Track 238: 4 --> 2
info:     pos: {-73.2143,36.1631,-57.7225}, dir: {-0.764737,-0.209753,-0.609246}, normal: {0.764737,0.209753,0.609246}
info:  Track 1395: 4 --> 2
info:     pos: {-13.6099,99.0428,2.30125}, dir: {0.102747,0.802923,-0.587161}, normal: {-0.102747,-0.802923,0.587161}
info:  Track 1994: 4 --> 3
info:     pos: {65.1013,-32.4723,68.6103}, dir: {0.904736,0.00747914,0.425907}, normal: {-0.904736,-0.00747914,-0.425907}
info:  Track 2296: 4 --> 3
info:     pos: {-37.2304,-33.3393,-86.6163}, dir: {0.346089,-0.682314,-0.643948}, normal: {-0.346089,0.682314,0.643948}
info:  Track 3388: 4 --> 3
info:     pos: {74.6755,-18.8631,-63.7789}, dir: {0.887871,0.383771,0.253782}, normal: {-0.887871,-0.383771,-0.253782}
info:  Track 238: refracted (1)
info:  Track 1395: refracted (1)
info:  Track 1994: refracted (1)
info:  Track 2296: refracted (1)
info:  Track 3388: refracted (1)
info:  Track 238: moved to 2
info:  Track 1395: moved to 2
info:  Track 1994: moved to 3
info:  Track 2296: moved to 3
info:  Track 3388: moved to 3
info:  Track 538: 2 --> 4
info:     pos: {17.7539,40.7805,-89.5642}, dir: {-0.114478,0.341311,0.932953}, normal: {0.114478,-0.341311,-0.932953}
info:  Track 538: refracted (1)
info:  Track 538: moved to 4
info:  Track 1233: 4 --> 2
info:     pos: {68.4585,65.5968,31.7882}, dir: {0.805776,0.539502,0.244257}, normal: {-0.805776,-0.539502,-0.244257}
info:  Track 2564: 4 --> 2
info:     pos: {-64.265,23.2562,-73.0011}, dir: {-0.869784,-0.282591,0.404498}, normal: {0.869784,0.282591,-0.404498}
info:  Track 3003: 4 --> 2
info:     pos: {-56.3338,32.266,76.0618}, dir: {-0.956893,-0.086279,0.27733}, normal: {0.956893,0.086279,-0.27733}
info:  Track 1233: refracted (1)
info:  Track 2564: refracted (1)
info:  Track 3003: refracted (1)
info:  Track 1233: moved to 2
info:  Track 2564: moved to 2
info:  Track 3003: moved to 2
info:  Track 1241: 4 --> 2
info:     pos: {85.8119,0.43535,51.3433}, dir: {0.937327,0.308698,0.161626}, normal: {-0.937327,-0.308698,-0.161626}
info:  Track 1241: refracted (1)
info:  Track 1241: moved to 2
```
where the normal vector is just the direction, which is what the current vecgeom normal is: https://github.com/celeritas-project/celeritas/blob/baa1f6acce1868037804342a6bd2891644e41365/src/geocel/vg/VecgeomTrackView.hh#L398-L402

We'll have to wait for #1926 before fully integrating the code in the tests. For now I'll revert back to using the dummy executor.

sethrj:
@hhollenb Rather than wait for vecgeom to work, could you disable the test when vecgeom the 'core' geometry, and add a `CELER_NOT_IMPLEMENTED` to the surface physics params explaining the limitation (with a link to #1926)?

src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
```diff
@@ -52,6 +52,10 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
                && sim.status() == TrackStatus::alive;
     }());
 
+#if CELERITAS_CORE_GEO == CELERITAS_CORE_GEO_VECGEOM
+    // \todo VecGeom surface normals are not currently supported, see PR #1926
+    track.sim().status(TrackStatus::killed);
+#else
```
amandalund: We should probably emit a warning somewhere if we're running an optical problem with VecGeom. Would it be difficult to still allow the track to propagate through the geometry/undergo volumetric processes but ignore any surface physics?

sethrj: Yes, I was thinking that in the constructor of the boundary *action class constructor*, we call `CELER_NOT_IMPLEMENTED(...)` with a link to the issue. That would disable optical physics entirely in practice, which is what we basically have to do.

hhollenb: Looking at some of the options to handle this:

1. Putting the absorption here allows volumetric processes and internal traversal but not passing through a boundary (this was the behavior before this PR). Can add a warning / error message to state VecGeom isn't currently supported.
2. Throwing in the class constructor means `SurfacePhysicsParams` will fail to build and all VecGeom tests involving surface physics will error out and need to be disabled / masked somehow.
3. Can not build surface physics, but then we will need to add some more behavior as an alternative for when a VecGeom track reaches a boundary.
4. Add a trivial surface interactor that doesn't depend on surface normals. I've started working on this but adding custom models isn't straightforward for surface physics right now.

Of course open to other ideas. I think 1 touches the least amount of code

amandalund: Ah, I forgot we were already killing the track in the post-boundary action. I'd lean toward still being able to run with VecGeom if possible, though it is silly to kill the track at the first boundary. A trivial surface interactor is a nice idea (especially if implementing the normal in VecGeom is going to take a while) but probably not worth the effort right now if it's not straightforward to add.

sethrj: Aha, good delineation of the consequences @hhollenb . I think as we transition to actually testing the whole optical transport loop, it's better to have a prompt detailed failure rather than silent wrong results after the run is complete. So I think it might be best to choose #2 and disable optical testing for vecgeom builds because, well, optical physics needs surface normals except in the infinite case. I can add this to your branch if it's onerous (but indeed I think we had it in place at one point).

hhollenb: Ok so I did some more testing, and if you replace the `InitBoundaryExecutor` with just crossing the boundary (no scattering, normal, etc), then the discrepancy still appears:
```
  CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
  {
       CELER_EXPECT([track] {
           auto sim = track.sim();
           return sim.post_step_action()
                      == track.surface_physics().scalars().init_boundary_action
                  && sim.status() == TrackStatus::alive;
       }());
   
       auto geo = track.geometry();
       CELER_EXPECT(geo.is_on_boundary());
  
       // Move the particle across the boundary
       geo.cross_boundary();
       if (CELER_UNLIKELY(geo.failed()))
       {   
           track.apply_errored();
           return;
       }   
  
       if (!track.material_record().material_id())
       {   
           track.sim().status(TrackStatus::killed);
           return;
       }   
  }
```
Getting a different number of steps for just g4 11.0.2 and the vecgeom+g4@11.0.2 builds, same as above. So it's not a vecgeom normal issue and not a surface physics specific issue I think?

(just to be sure, added `CELER_NOT_IMPLEMENTED` to the VecGeom normal function and it doesn't throw any errors)

amandalund: Another idea (though out of scope for this PR): currently I think disabling the boundary process in `GeantOpticalPhysicsOptions` will disable the process in Geant4 but have no effect in Celeritas. We could support that option by adding another boundary executor that just crosses the boundary/kills the track if it's not in an optical material and build that boundary action instead of the surface physics ones if the boundary option is disabled.

whokion:
> Still working but perhaps @whokion could check over the fresnel code?

The Fresnel code (FresnelCalculator) looks good to me - may add relevant references (as the convention for the direction of reflected **E** and **H** fields for TM may be different).

src/celeritas/optical/surface/model/DielectricInteractor.hh
```diff
@@ -0,0 +1,166 @@
...
+    // Can't do analytic interaction if there's no post-volume optical material
+    if (!s_phys.next_material())
+    {
+        return SurfaceInteraction::from_absorption();
+    }
+
+    // No interaction if optical materials are identical
+    if (s_phys.material() == s_phys.next_material())
+    {
+        return SurfaceInteraction{SurfaceInteraction::Action::refracted,
```
whokion: Hard to imagine a realistic use case where two identical materials are separated by an optical boundary, even though nothing technically prevents defining such a surface. That said, in this case, should  the action flag be transmitted (i.e., with no change in direction or polarization)?

src/celeritas/optical/surface/model/ReflectionFormCalculator.hh
```diff
@@ -6,69 +6,170 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate back-scattering reflection.
+ *
+ * The photon direction and polarization are reversed.
+ */
+CELER_FUNCTION SurfaceInteraction ReflectionFormCalculator::calc_backscatter() const
+{
+    SurfaceInteraction result;
+    result.action = SurfaceInteraction::Action::reflected;
```
whokion, line 139: The inclusive `Action::reflected` could be expanded to cover more specific cases (e.g., backscattering, Lambertian reflection, etc.), which may be useful for validation — particularly against Geant4 results.

src/celeritas/optical/surface/model/FresnelCalculator.hh
```diff
@@ -0,0 +1,326 @@
...
+/*!
+ * Helper class for calculating quantities for Fresnel equations.
+ *
+ * Incident waves are decomposed into transverse-electric (TE) and
+ * transverse-magnetic (TM) modes. The interaction plane is defined as the span
+ * of the photon direction and the surface normal. The TE (TM) mode has the
+ * electric (magnetic) field polarization transverse to the interaction plane.
+ * If the direction is parallel to the surface normal, then interaction plane
+ * is degenerate and the incident photon is defined to be entirely in the TE
+ * mode.
```
whokion:
```suggestion
 * Incident waves are decomposed into transverse-electric (TE) and
 * transverse-magnetic (TM) polarization. The interaction plane is defined as the span
 * of the photon direction and the surface normal. The TE (TM) polarization has the
 * electric (magnetic) field transverse to the interaction plane.
 * If the direction is parallel to the surface normal, then interaction plane
 * is degenerate and the incident photon is defined to be entirely in the TE
 * polarization.
```

whokion (commented):
The implementation and refactoring differ significantly from Geant4. A systematic validation against Geant4 should be added, perhaps in follow-up MRs.

src/celeritas/optical/surface/model/FresnelCalculator.hh
```diff
@@ -0,0 +1,326 @@
...
+    return this->reflectivity_ratio(cosine_ratio_ * relative_r_index_);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate reflectivity coefficient of the TM component.
+ */
+CELER_FUNCTION real_type FresnelCalculator::calc_reflectivity_tm() const
+{
+    return this->reflectivity_ratio(cosine_ratio_ / relative_r_index_);
```
whokion, line 287:
```suggestion
    return -this->reflectivity_ratio(cosine_ratio_ / relative_r_index_);
```
It does not affect the final reflectivity, but it seems a negative sign is missing here — at least add a note to clarify if the sign is dropped intentionally.

src/celeritas/optical/surface/model/DielectricInteractor.hh
hhollenb: We had this case come up in the lar-sphere tests. Added a transmitted flag like for the core physics interactions.

src/celeritas/optical/surface/model/ReflectionFormCalculator.hh
hhollenb: Expanding it would make the `SurfaceInteractionApplier` logic become very complicated. We could probably get the same validation from using surfaces with just spike / lobe / back-scatter / Lambertian to run isolated tests.

src/celeritas/optical/surface/model/FresnelCalculator.hh
hhollenb: I went through Fowles modern optics to check my calculations and added some equation references. Based on those equations the minus sign error I have is in the TE mode. While it doesn't affect the reflectivity, it does affect the refracted polarization of mixed TE/TM waves.

doc/_static/zotero.bib
```diff
@@ -1090,6 +1090,15 @@ @article{fiorini-definingcyclotronbased-2018
...
 }
 
+@book{fowles-modern-optics,
+  title = {Introduction to Modern Optics},
+  author = {Fowles, Grant R.},
+  year = {1975},
+  edition = {Second},
+  publisher = {Holt, Rinehart and Winston, Inc.},
+  isbn = {0-03-089404-2}
+}
```
sethrj, line 1108: I've added this to the actual [Zotero group database](https://www.zotero.org/groups/2380941) (sorry I didn't notice before now you weren't a member!) ; could you please replace this citation key with `fowles-optics-1975`?

src/celeritas/optical/CoreTrackView.hh
```diff
@@ -57,6 +57,9 @@ class CoreTrackView
     // Return a material view (using an existing geo view)
     inline CELER_FUNCTION MaterialView material_record(GeoTrackView const&) const;
 
+    // Return a material view for a specific optical material
+    inline CELER_FUNCTION MaterialView material_record(OptMatId) const;
```
sethrj, line 61: Future work: rename all this to `material`... it's an artifact of the way we did the material view

src/celeritas/optical/surface/model/DielectricInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+ * equations and sampled to determine if the photon will reflect or refract. If
+ * it reflects, then the UNIFIED model is used to handle the different forms of
+ * reflection. If it refracts, then dielectric-dielectric interfaces will use
+ * Snell's law to determine the refracted wave direction and polarization. For
+ * dielectric-metal interfaces, refracted waves are just absorbed.
+ */
+class DielectricInteractor
+{
+  public:
+    struct Builder
```
sethrj: I think for this to be called "Interactor::Builder" it would need to return an `Interactor` without changing its state. We can brainstorm different designs later, but for now it should be easy and obvious enough to rename it `Executor`.

doc/implementation/optical-physics.rst
```diff
@@ -115,3 +115,9 @@ distribution" to account for the roughness of the surface.
 
 .. doxygenclass:: celeritas::optical::SmearRoughnessSampler
 .. doxygenclass:: celeritas::optical::GaussianRoughnessSampler
+
+Interactions are sampled from models describing the distributions of absorption,
+reflection, and refraction on the surface.
+
+.. doxygenclass:: celeritas::optical::FresnelCalculator
+.. doxygenclass:: celeritas::optical::ReflectionFormCalculator
```
sethrj, line 124: Also add DielectricInteractor here please! :)

src/celeritas/optical/surface/model/DielectricInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+/*!
+ * Sample optical interactions for dielectric-dielectric and dielectric-metal
+ * interfaces.
+ *
+ * For both interfaces, the reflectivity is first calculated from Fresnel
+ * equations and sampled to determine if the photon will reflect or refract. If
+ * it reflects, then the UNIFIED model is used to handle the different forms of
+ * reflection. If it refracts, then dielectric-dielectric interfaces will use
+ * Snell's law to determine the refracted wave direction and polarization. For
+ * dielectric-metal interfaces, refracted waves are just absorbed.
```
sethrj: Mention the corresponding sampler classes so it'll be hyperlinked in the documentation

src/celeritas/optical/surface/model/DielectricInteractor.hh
```diff
@@ -0,0 +1,164 @@
...
+
+    // No interaction if optical materials are identical
+    if (s_phys.material() == s_phys.next_material())
+    {
+        return SurfaceInteraction::from_transmission();
+    }
+
+    auto rng = track.rng();
+    auto sub_model_id = s_phys.interface(SurfacePhysicsOrder::interaction)
+                            .internal_surface_id();
```
sethrj, line 95:
```c++
CELER_ASSERT(sub_model_id < dielectric_data.interface.size());
```

sethrj (approved):
Looks good, sorry for the delay in reviewing !

doc/_static/zotero.bib
sethrj: The reference should be defined on https://github.com/celeritas-project/celeritas/pull/2052

sethrj:
@hhollenb @pcanal I'm going to ignore the coverage drop and merge since we don't have a suitable way to test the model — and its purpose is for testing anyway.

sethrj:
Oops I ignored the coverage change, I think later we can revisit the coverage issues (unused core track view methods).

PR #215, opened 2021-04-14
Use CutoffView in atomic relaxation

amandalund:
Now that we have production cuts, use `CutoffView` to get the electron and photon energy cutoffs for atomic relaxation. I made a minor change to the `CutoffView` interface so it now constructs the view with the material ID and accepts the particle ID as an argument in the `energy()` and `range()` methods -- this way we don't have to construct a separate view for each particle type. I also moved the structs for the raw imported EADL transition data from the `AtomicRelaxationParams` class to the io/ directory.

Right now on emmet I'm getting the same nvlink error that we saw in #118,
```
nvlink error   : Size doesn't match for '_ZN9celeritas9SecondaryC1Ev$146' in 'src/CMakeFiles/celeritas.dir/physics/em/detail/MollerBhabha.cu.o', first specified in 'src/CMakeFiles/celeritas.dir/physics/em/detail/BetheHeitler.cu.o'
nvlink fatal   : merge_elf failed
```
which I don't have a workaround for yet.

PR #118, opened 2021-01-26
Workaround nvlink bug

sethrj:
@amandalund @pcanal

---

On emmet we saw the following error (which didn't appear on the CI, which runs CUDA 11.1):
```
nvlink error   : Size doesn't match for '_ZN9celeritas9SecondaryC1Ev$53' in 'CMakeFiles/celeritas.dir/physics/em/detail/EPlusGG.cu.o', first specified in 'CMakeFiles/celeritas.dir/physics/em/detail/BetheHeitler.cu.o'
nvlink fatal   : merge_elf failed
```

A bug report for LLVM hinted at a bug in `nvlink` regarding weak symbols: https://bugs.llvm.org/show_bug.cgi?id=40893 . Using the nvcc `-keep` flag led to finding the Secondary's constructor was defined as a weak symbol, and internally it stores two values (for the initialization values of def_id and energy), which are defined internally in BetheHeitler as
```
.weak .global .align 4 .b8 _ZN9celeritas9SecondaryC1Ev$53[4] = {255, 255, 255, 255};
.weak .global .align 8 .b8 _ZN9celeritas9SecondaryC1Ev$54[8];
```

However, the constructor's private variables show up with a different suffix in the EPlusGG kernel:
```
.weak .global .align 4 .b8 _ZN9celeritas9SecondaryC1Ev$52[4] = {255, 255, 255, 255};
.weak .global .align 8 .b8 _ZN9celeritas9SecondaryC1Ev$53[8];
```
so the suffixes are off by one and end up causing a collision.

I'm not sure how the suffixes are created, but changing the include order changes their value, so rearranging the includes causes the collision to disappear.

PR #1985, opened 2025-09-24
Remove `weight` field from `Secondary`

amandalund:
EDIT: Workaround didn't end up working, so this just removes the `weight` field from `Secondary` since it's unnecessary.

Long ago we kept running into an nvlink error (#118 #215 https://github.com/celeritas-project/celeritas/pull/257#issuecomment-868994741) and trying various workarounds to avoid it. What seemed to finally resolve it was removing member initialization from `Secondary`: https://github.com/celeritas-project/celeritas/pull/257#issuecomment-870251676.

I just started hitting a similar error on a branch I'm working on:
```console
nvlink error   : Size doesn't match for '_ZN9celeritas14StackAllocatorINS_9SecondaryEEclEj$188' in '/home/alund/celeritas_project/celeritas/build-reldeb/lib64/libceleritas_static.a:MollerBhabhaModel.cu.o', first specified in '/home/alund/celeritas_project/celeritas/build-reldeb/lib64/libceleritas_static.a:MuDecayChannel.cu.o' (target: sm_70)
nvlink fatal   : merge_elf failed (target: sm_70)
```
Removing initialization of the recently-added `weight` field in `Secondary` again appears to fix it. I _think_ we correctly set the weight elsewhere and don't need to initialize it to one here, but wanted to double check since we've had to fix this in several places already.

### DIFF
--- a/src/celeritas/phys/InteractionApplier.hh
+++ b/src/celeritas/phys/InteractionApplier.hh
@@ -148,13 +148,12 @@ InteractionApplierBaseImpl<F>::operator()(celeritas::CoreTrackView const& track)
         // Kill secondaries with energies below the production cut
         for (auto& secondary : result.secondaries)
         {
-            secondary.weight = sim.weight();
             if (cutoff.apply(secondary))
             {
                 // Secondary is an electron, positron or gamma with energy
                 // below the production cut -- deposit the energy locally
                 // and clear the secondary
-                deposition += secondary.energy.value() * secondary.weight;
+                deposition += secondary.energy.value() * sim.weight();
                 auto sec_par = track.particle_record(secondary.particle_id);
                 if (sec_par.is_antiparticle())
                 {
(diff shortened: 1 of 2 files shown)


sethrj (commented):
Very odd. Don't we need to initialize it elsewhere now? Would an empty constructor that initializes it work? Does this only show up when using RDC (VecGeom)?

amandalund:
I didn't see it with an orange build (but that doesn't necessarily mean we couldn't). Looking at our old discussions, we'd see it appear or disappear by things like removing an assertion or changing the order of headers.

Actually, do we even need to store the weight in the secondary? We never set it there in the interactors and set it in the track initializer directly from the parent track.

sethrj:
Oh, this is the *secondary* ... you're right, we shouldn't need it there at all. Since the Secondary is one of the few things created with the StackAllocator, perhaps there's something wrong/buggy with placement new in NVCC? :\

amandalund:
Well, removing the weight made the error reappear, so I guess this isn't a workaround :(. Rebasing that branch on develop made it go away again, so I'm just going to make this PR about removing the weight and defer worrying about the nvlink error until I'm forced to.

sethrj (approved):

PR #1986, opened 2025-09-24
Add `CELER_` prefix to G4 material property names specific to Celeritas

sethrj:
Following up on our discussion in #1974, this adds a `CELER_` prefix to Celeritas-specific material properties for optical physics. It still supports the existing ones for backward compatibility and gives a warning. It adds documentation for the new parameters.

### DIFF
--- a/app/data/lar-sphere.gdml
+++ b/app/data/lar-sphere.gdml
@@ -16,8 +16,6 @@
     <matrix name="ISY" coldim="2" values="1.0e-6 2000 6 4000"/>
     <matrix name="ESY" coldim="2" values="1.0e-6 3750 6 5000"/>
     <matrix name="ESY1" coldim="1" values="4000"/>
-    <matrix name="ELM1" coldim="1" values="1e-4"/>
-    <matrix name="ELS1" coldim="1" values="1e-5"/>
     <matrix name="ERT1" coldim="1" values="15"/>
     <matrix name="ETC1" coldim="1" values="5"/>
     <!-- End of particle scintillation data -->
@@ -106,8 +104,6 @@
       <property name="ELECTRONSCINTILLATIONYIELD1" ref="ESY1"/>
       <property name="ELECTRONSCINTILLATIONRISETIME1" ref="ERT1"/>
       <property name="ELECTRONSCINTILLATIONTIMECONSTANT1" ref="ETC1"/>
-      <property name="CELER_ELECTRONSCINTILLATIONLAMBDAMEAN1" ref="ELM1"/>
-      <property name="CELER_ELECTRONSCINTILLATIONLAMBDASIGMA1" ref="ELS1"/>
       <property name="RAYLEIGH" ref="RAY"/>
       <property name="ISOTHERMAL_COMPRESSIBILITY" ref="IC"/>
       <property name="ABSLENGTH" ref="ABS"/>
--- a/app/data/lar-sphere.gdml
+++ b/app/data/lar-sphere.gdml
@@ -92,10 +92,10 @@
       <property name="SCINTILLATIONTIMECONSTANT1" ref="TC1"/>
       <property name="SCINTILLATIONTIMECONSTANT2" ref="TC2"/>
       <property name="SCINTILLATIONTIMECONSTANT3" ref="TC3"/>
-      <property name="SCINTILLATIONLAMBDAMEAN1" ref="LM1"/>
-      <property name="SCINTILLATIONLAMBDAMEAN2" ref="LM2"/>
-      <property name="SCINTILLATIONLAMBDASIGMA1" ref="LS1"/>
-      <property name="SCINTILLATIONLAMBDASIGMA2" ref="LS2"/>
+      <property name="CELER_SCINTILLATIONLAMBDAMEAN1" ref="LM1"/>
+      <property name="CELER_SCINTILLATIONLAMBDAMEAN2" ref="LM2"/>
+      <property name="CELER_SCINTILLATIONLAMBDASIGMA1" ref="LS1"/>
+      <property name="CELER_SCINTILLATIONLAMBDASIGMA2" ref="LS2"/>
       <property name="SCINTILLATIONCOMPONENT3" ref="SC3"/>
       <property name="PROTONSCINTILLATIONYIELD" ref="PSY"/>
       <property name="DEUTERONSCINTILLATIONYIELD" ref="DSY"/>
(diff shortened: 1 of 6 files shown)


doc/implementation/optical-physics.rst
```diff
@@ -61,15 +61,37 @@ tracking loop:
...
+:cpp:class:`G4MaterialPropertiesTable` that has a ``RINDEX`` material property.
+(It also provides a special case for water if no material table is associated,
+allowing Rayleigh scattering by default by providing an isothermal
+compressibility and assuming STP.)
+All physical materials that use the geometry material or property table share
+the same optical material.
+
+Celeritas translates many Geant4 material properties into its internal physics
+input parameters. It also allows material-specific user configuration of
+Celeritas-only physics. The following table
```
amandalund: Too many rodents in your office.

amandalund (approved):
Thanks @sethrj!

app/data/lar-sphere.gdml
```diff
@@ -106,8 +106,8 @@
       <property name="ELECTRONSCINTILLATIONYIELD1" ref="ESY1"/>
       <property name="ELECTRONSCINTILLATIONRISETIME1" ref="ERT1"/>
       <property name="ELECTRONSCINTILLATIONTIMECONSTANT1" ref="ETC1"/>
-      <property name="ELECTRONSCINTILLATIONLAMBDAMEAN1" ref="ELM1"/>
-      <property name="ELECTRONSCINTILLATIONLAMBDASIGMA1" ref="ELS1"/>
+      <property name="CELER_ELECTRONSCINTILLATIONLAMBDAMEAN1" ref="ELM1"/>
+      <property name="CELER_ELECTRONSCINTILLATIONLAMBDASIGMA1" ref="ELS1"/>
```
whokion: We don't support the lambda mean[] and sigma[] for specific particle type (so, this will not be used anywhere unless we extend this feature).

doc/implementation/optical-physics.rst
```diff
@@ -61,15 +61,37 @@ tracking loop:
 Optical materials
 =================
 
-Each "physics material" (see :cpp:class:`celeritas::ImportPhysMaterial`) can
-have an associated "optical material." When importing from Geant4, each optical
-material corresponds to a single "geometry material" (see
-:cpp:class:`celeritas::ImportGeoMaterial`) that has a ``RINDEX`` material
-property, and all physical materials that use the geometry material share
-the same optical material.
+Each "physics material" (i.e., a combination of material and physics options) can
```
whokion, line 64: Could you clarify the physics options in more detail? It’s not clear what you mean by “physics material” after all and one may be having trouble interpreting it.

doc/implementation/optical-physics.rst
sethrj: True that...

sethrj:
Oops accidental remote branch delete...

doc/implementation/optical-physics.rst
sethrj: Well, I guess I just paused to write the table, and never got back to the reference 🙃

doc/implementation/optical-physics.rst
```diff
@@ -113,26 +135,15 @@ Surface processes
...
-1. The interface surface from P0 to P1
-2. If L1 is the child of L0 (crossing into an "enclosed" volume), then the
-   boundary surface of L1
-3. The boundary surface of L0
-4. The boundary surface of L1
-5. The volumetric material properties of L0 and L1
-
-.. todo:: Once surface models are implemented, move the
-   precedence above into SurfacePhysics documentation.
+specified largely by user-provided material properties.
```
whokion:
```suggestion
specified largely by user-provided surface types.
```

doc/implementation/optical-physics.rst
```diff
@@ -113,26 +135,15 @@ Surface processes
...
+Users can define "boundary" and "interface" surfaces representing,
+respectively, the entire surface of a volume (including interior surfaces
+created implicitly by child volumes) and the common face between two adjacent
```
whokion: I do not understand `interior surfaces created implicitly by child volumes` - what are they?

test/geocel/data/lar-sphere.gdml
```diff
@@ -110,8 +110,8 @@
       <property name="ELECTRONSCINTILLATIONYIELD1" ref="ESY1"/>
       <property name="ELECTRONSCINTILLATIONRISETIME1" ref="ERT1"/>
       <property name="ELECTRONSCINTILLATIONTIMECONSTANT1" ref="ETC1"/>
-      <property name="ELECTRONSCINTILLATIONLAMBDAMEAN1" ref="ELM1"/>
-      <property name="ELECTRONSCINTILLATIONLAMBDASIGMA1" ref="ELS1"/>
+      <property name="CELER_ELECTRONSCINTILLATIONLAMBDAMEAN1" ref="ELM1"/>
+      <property name="CELER_ELECTRONSCINTILLATIONLAMBDASIGMA1" ref="ELS1"/>
```
whokion: same comment here!

whokion (approved):
Good extension for distinguishable celeritas material properties!

app/data/lar-sphere.gdml
sethrj: These were added in #1151 to test the data import even before the capability was there... I'll delete it.

doc/implementation/optical-physics.rst
sethrj: Suppose we have a logical volume `A` that's a sphere: it has a single surface, the sphere shell. But then you add a couple of tiny daughter spheres `B` and `C`: without changing `A`'s solid/shape you've added two more tiny spherical shell surfaces inside of it. This "new surface by subtraction" underpins the weird skin surface priority behavior in the OpBoundaryProcess. I'm not sure how to better explain it...

whokion: From my (perhaps somewhat naive) understanding, there is only one physical surface interface, whether defined as a skin or a border logical surface. If this implicit surface behaves differently in OpBoundaryProcess, it seems to introduce a new concept beyond what Geant4 currently supports—rather than, for example, modeling a thin spherical shell (a volume with two distinct surfaces, but still a single logical skin or border depending on the surrounding materials) in place of the interior surface. If this new concept (which still seems rather vague to me) is important, it would be helpful to describe it more explicitly in the documentation to avoid confusion or ambiguity.

doc/implementation/optical-physics.rst
sethrj: I didn't even notice there's no documentation on the different 'material IDs'... thanks!

sethrj: I'm just going to remove this confusing bit for now because I've gotten massively sidetracked trying to document this better...

doc/implementation/optical-physics.rst
sethrj: I've updated to make this clearer; I get the confusion between a conceptual (topological) surface and a program definition of "surface".

sethrj:
Thanks for the review @whokion and @amandalund !

PR #1987, opened 2025-09-24
Support G4MultiUnion in ORANGE

elliottbiondo:
This MR adds support for `G4MultiUnion` within `g4org::SolidConverter`. Whereas, Geant4 uses a voxelization approach for acceleration, this MR uses the naive approach of simply `or`-ing all constituent volumes. A more general approach for sub-volume-level acceleration should be use for all volumes, not just converted multiunions.

### DIFF
--- a/src/orange/g4org/SolidConverter.cc
+++ b/src/orange/g4org/SolidConverter.cc
@@ -25,6 +25,7 @@
 #include <G4GenericTrap.hh>
 #include <G4Hype.hh>
 #include <G4IntersectionSolid.hh>
+#include <G4MultiUnion.hh>
 #include <G4Orb.hh>
 #include <G4Para.hh>
 #include <G4Paraboloid.hh>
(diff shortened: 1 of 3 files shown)


elliottbiondo:
@sethrj RTR

src/orange/g4org/SolidConverter.cc
```diff
@@ -613,6 +615,25 @@ auto SolidConverter::intersectionsolid(arg_type solid_base) -> result_type
...
+{
+    auto const& mu = static_cast<G4MultiUnion const&>(solid_base);
+    auto n = mu.GetNumberOfSolids();
+    std::vector<result_type> vols(n);
+
+    for (auto i : range(n))
+    {
+        auto vol = (*this)(*(mu.GetSolid(i)));
+        vols[i] = std::make_shared<Transformed>(
+            std::move(vol), transform_(mu.GetTransformation(i)));
```
sethrj, line 630: I was going to say use `Transformed::or_object` in case it's an identity, but it looks we don't check for the special cases of no-rotation and identity transformations. That won't affect the final geometry but could be nice for simplifying debug output (so we shouldn't worry about it).

sethrj (approved):
Perfect!

sethrj:
For what it's worth, `static_cast` *should* work there since it's not got virtual inheritance, and we know a priori from the function call that it's got that type. But I think clang-tidy is agressive about some of these thing,s, and it's not going to hurt us performance-wise since this is not a hot spot.

PR #1988, opened 2025-09-25
Abandon sphinx-mermaid and add celerstruct directive

sethrj:
- Replace the sphinx mermaid (which is rather limited and doesn't allow creating a referenceable figure) with a direct invocation of the mermaid generation app from `npm` through cmake
- Add a `celerstruct` directive that's a shortcut for `doxygenstruct` listing members and preventing class links that cause warnings

This PR should have no change to the content of the documentation; only how it's generated.


src/geocel/vg/VecgeomParams.hh
```diff
@@ -76,8 +76,7 @@ class VecgeomParams final : public GeoParamsInterface,
 
     //!@{
     //! \name Static constructor helpers
-    //! \todo: move these to a "model" abstraction that loads/emits geometry,
-    //! materials, volumes?
+    //! \todo move these to a "model" abstraction
```
pcanal:
```suggestion
    //! \todo Move these to a "model" abstraction
```
To follow the casing used in OrangeParams.hh.

.github/workflows/doc.yml
```diff
@@ -71,6 +71,8 @@ jobs:
             sudo apt-get -y update
             sudo apt-get -y install \
                 cmake graphviz ninja-build doxygen nlohmann-json3-dev
+            sudo npm install -g @mermaid-js/mermaid-cli
+            sudo npm -g rebuild puppeteer
```
pcanal:
```suggestion
            sudo npm -g rebuild puppeteer
            sudo npx puppeteer browsers install chrome-headless-shell
```
or something like that ... the doc fails now with:
```
Error: Could not find Chrome (ver. 131.0.6778.204). This can occur if either
 1. you did not perform an installation before running the script (e.g. `npx puppeteer browsers install chrome-headless-shell`) or
 2. your cache path is incorrectly configured (which is: /home/runner/.cache/puppeteer).
For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration.
```

pcanal (approved):
It looks good (except for the missing dependency ...)

.github/workflows/doc.yml
sethrj: Yeah... trying to figure this one out, didn't realize it was going to be so complicated. Apparently `mmdc` requires enough styling to require a local copy of chromium to work...

PR #1989, opened 2025-09-26
Fix setting CUDA heapSize through G4UI

esseivaju:
Correctly call `set_cuda_heap_size` when specifying the heap size through Geant4 UI commands / SetupOptions.


sethrj (approved):
Oops! Thanks @esseivaju .

PR #1990, opened 2025-09-26
Improve some nomenclature

sethrj:
- Use enable_ instead of allow_ for stream/signal variables, document them, use getenv_flag
- use_profiling -> enable_profiling
- Make some comments for to-do about nuclides
- Use "reset" instead of "clear" when it means "set to a default state" rather than "remove all elements from"
- Replace ICRT (inline constexpr real_type) with ICC (inline constexpr Constant)
- Replace Printable -> Streamable
- Rename lingering uses of  ExplicitAction with StepAction

See #721


pcanal:
> Use "reset" instead of "clear" when it means

Are there missing words?

src/celeritas/mat/MaterialParams.hh
```diff
@@ -54,7 +54,7 @@ class MaterialParams final : public ParamsDataInterface<MaterialParamsData>
     using SpanConstIsotopeId = Span<IsotopeId const>;
     //!@}
 
-    //! Define an element's isotope input data
+    //! Define a single nuclide's data
     struct IsotopeInput
```
pcanal, line 58: Should we also change the class name?
```suggestion
    //! Define a single nuclide's data
struct NuclideInput
```

src/celeritas/Constants.hh
```diff
@@ -57,53 +57,53 @@ namespace constants
 {
 //---------------------------------------------------------------------------//
 
-#define CELER_ICRT inline constexpr Constant
+#define CELER_ICC inline constexpr Constant
```
pcanal, line 60: (minor).  Should there be some hint in the name that this is a local symbol (i.e. 'valid with this value only this header')?

src/celeritas/mat/MaterialParams.hh
sethrj: That is on the to-do list, but it seems to big a change for this pr

src/corecel/io/ScopedStreamRedirect.hh
```diff
@@ -35,7 +35,7 @@ class ScopedStreamRedirect
 {
   public:
     // Whether stream redirection is enabled
-    static bool allow_redirect();
+    static bool enable_redirect();
```
pcanal: Both the old and new spelling conflict (for me) with the documentation.
They both sounds to me as "please turn-on/enable redirection starting now (if not already on)'
as opposed to asking the question "is it already enabled?"
```suggestion
    static bool redirect_is_enabled();
```
The comment applies equally to `enable_profiling` and `enable_signals`.

sethrj: Yeah, I was trying to be consistent with the environment variable naming, but I agree `foo_enabled` is clearer. How about make them all `static bool enabled()`?

src/celeritas/Constants.hh
sethrj: Is the `#undef` a few lines later and the lack of Doxygen a good enough indicator? We actually have quite a few "local" macros in the library:
```console
$ git grep -h '#undef' src | sort -u
#undef ASU_SAMPLE_ELOSS
#undef CELER_DEFINE_ARRAY_ARITHM
#undef CELER_DEFINE_ARRAY_ASSIGN
#undef CELER_DEFINE_ATOMICNUMBER_CMP
#undef CELER_DEFINE_CONSTANT_CMP
#undef CELER_DEFINE_CONSTANT_OP
#undef CELER_DEFINE_MPITYPE
#undef CELER_DEFINE_OBSPTR_CMP
#undef CELER_DEFINE_OPAQUEID_CMP
#undef CELER_DEFINE_PDGNUMBER
#undef CELER_DEFINE_QUANTITY_CMP
#undef CELER_DEFINE_VERSION_CMP
#undef CELER_DIO_PAIR
#undef CELER_ICC
#undef CELER_ICC_
#undef CELER_US_VISIT_CASE
#undef CO_ADD_CFG
#undef CO_ADD_COND_VERS
#undef CO_ADD_OPT
#undef CP_VALIDATE_INPUT
#undef CSB_INSTANTIATE
#undef DS_ASSIGN
#undef DS_COPY_IF_SELECTED
#undef FDO_INPUT
#undef GBPO_LOAD_OPTION
#undef GCPO_LOAD_OPTION
#undef GMPO_LOAD_OPTION
#undef GOPO_LOAD_OPTION
#undef GPO_LOAD_OPTION
#undef GSPL_OSF_PAIR
#undef GSPL_OSM_PAIR
#undef GSPL_ST_PAIR
#undef GSPL_VALIDATE_PROB
#undef GSPO_LOAD_OPTION
#undef HP_CLEAR_STEP_POINT
#undef HP_SET
#undef HP_SETUP_POINT
#undef IM_LOAD_OPTION
#undef IM_LOAD_REQUIRED
#undef KCE_INSERT_IF_VALID
#undef LSI_INSTANTIATE
#undef MRIO_MEM_OUT
#undef OPO_SAVE_BIH_SIZE
#undef OPO_SAVE_SCALAR
#undef OPO_SAVE_SIZE
#undef ORANGE_INSTANTIATE_SHAPE_STREAM
#undef ORANGE_ST_VISIT_CASE
#undef ORANGE_SURFACE_TRAITS
#undef ORANGE_TRANSFORM_TRAITS
#undef ORANGE_TT_VISIT_CASE
#undef ORANGE_UNIV_TRAITS
#undef ORANGE_UT_VISIT_CASE
#undef PPO_SAVE_OPTION
#undef PPO_SAVE_SIZE
#undef RSW_CREATE_BRANCH
#undef RSW_STORE
#undef RZFI_LOAD
#undef SC_TYPE_FUNC
#undef SCCEF_
#undef SD_RESIZE_IF_SELECTED
#undef SGL_SET_IF_SELECTED
#undef SRFI_LOAD_OPTION
#undef SRFI_SAVE_OPTION
#undef TC_SET_PAYLOAD
```

pcanal: > Is the #undef a few lines later and the lack of Doxygen a good enough indicator?

Essentially no, especially for the variable names that are reused and have potentially different content or meaning.  In practice you are right hence my rating of 'minor' for this.

src/corecel/io/ScopedStreamRedirect.hh
pcanal: Do you mean:
```
if (this->enabled())
```
or
```
if (this->redirect_enabled())
```
?

sethrj: Since it's a static method, it would be `ScopedProfiling::enabled()` etc.

sethrj: We also have a global `bool celeritas::use_color()` that's based on the `CELER_COLOR` flag...

pcanal: > Since it's a static method, it would be ScopedProfiling::enabled() etc.

Used this way, it works well.  Used (as it is as the moment in some places) as `this->enabled()` this would be okay-ish but likely suboptimal/sometime-confusing.

> We also have a global `bool celeritas::use_color()` that's based on the CELER_COLOR flag...

I think that form also works as it can not be read as trying to enable the use of color but rather as "use colors if we are allowed".

sethrj: `this->enabled()` won't work for static methods.

pcanal: I agree that it is something that does not 'work' for us but it (for better or worse) compiles: https://godbolt.org/z/G87xq5a5f

sethrj: Weird, I know I encountered an error... perhaps I was trying to call a static method from within another static method? That does indeed fail: https://godbolt.org/z/4x8h1sozh

pcanal: Ah :) ... that is 'just' the `this` keyword not being defined in a static function.  The version that compiles (https://godbolt.org/z/jYrxbKqYq) is as unclear of what is 'enabled' :)

PR #1991, opened 2025-09-27
Support decay in stepping loop

amandalund:
This adds initial support for decay in the stepping loop. Muon decay is currently the only decay channel supported, but adding additional channels should be straightforward now. 

To easily support the different decay channel types (which may not have fixed daughter particle types, or may have an arbitrary number of daughters), this moves from using channel-specific data (e.g. `MuDecayData`) to the generic `DecayChannelData` and `DecayTableData`. Daughter particle IDs are passed to the interactor and are expected to be in a specific order (this should be validated in the `DecayChannel` constructor).

Changes include:
  - Add decay input structs
  - Import decay tables from Geant4 for particles that have the decay process enabled
  - Rename `Process` to `InteractionProcess` and add a new `DecayProcess` (used for all particle types that decay)
  - Add `DecayChannel` which extends from `CoreStepActionInterface`,  responsible for validating decay table input and launching kernels for interactors
  - Add `MuDecayChannel` derived class
  - Refactor `MuDecayInteractor` to use generic decay data
  - Store decay tables and mapping from channel ID to action ID in physics data
  - Include decay as a step limiter, calculating cross section from decay constant
  - Sample decay process from among other discrete processes in `select_discrete_interaction`
  - If decay is selected, sample decay channel according to branching ratios

I haven't added any tests yet (only checked that running celer-g4 on a simple problem with muon decay imports the data and interacts by decay without any failures), but otherwise this should be ready for a first look.

See #1375


src/celeritas/phys/ParticleTrackView.hh
```diff
@@ -405,5 +422,18 @@ CELER_FUNCTION auto ParticleTrackView::momentum() const -> Momentum
...
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Decay length \f$ d = \beta \gamma c \tau \f$ [len].
+ */
+CELER_FUNCTION real_type ParticleTrackView::decay_length() const
+{
+    CELER_EXPECT(this->mass() > zero_quantity());
+    CELER_EXPECT(this->decay_constant() != constants::stable_decay_constant);
```
sethrj, line 432: Since momentum is always nonzero, I think we just get a (correct) value of infinity if either or both of these is zero.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -929,6 +954,50 @@ std::vector<ImportRegion> import_regions()
...
+
+        inp::DecayChannel channel;
+        channel.type = to_decay_channel_type(g4_channel);
+        channel.branching_ratio = g4_channel->GetBR();
+
+        // Exclude the neutrino daughters for muon decay
+        //! \todo Remove once neutrinos are included in \c MuDecayInteractor
+        size_type num_daughters = channel.type == DecayChannelType::muon
+                                      ? 1
+                                      : g4_channel->GetNumberOfDaughters();
```
sethrj, line 981: Should we warn about missing neutrinos? (And is there another process in which we're neglecting neutrinos as well, or am I thinking of the muon decay that Stefano already put in?) Also, are we sure the neutrino is always listed last? Maybe it'd be better test the PDG inside the loop and warn+skip if neutrino.

I think it's the right thing to omit them on the "input" side rather than pull them in and ignore them at the process level, but maybe it'd be a nice feature to be able to store a more complete physics as input have have the PhysicsParams class trim out deactivated/unsupported features, e.g. removing hadronic physics if hadron particles are disabled?

src/celeritas/decay/DecayData.hh
```diff
@@ -0,0 +1,54 @@
...
+    //! Whether the data is assigned
+    explicit CELER_FUNCTION operator bool() const
+    {
+        return !daughters.empty();
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Decay channels for a particle type.
```
sethrj, line 36: Add a note that these are stored by the `PhysicsParamsData` and used in the `ProcessGroup` (processes that apply to a particular particle type).

src/celeritas/inp/Physics.hh
```diff
@@ -95,6 +96,32 @@ struct OpticalPhysics
...
+    std::vector<PDGNumber> daughters;
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Decay processes and options.
+ */
+struct DecayPhysics
+{
+    using DecayTable = std::vector<DecayChannel>;
```
sethrj, line 116: I think `VecDecayChannel` since it's not really a table: there's no [row, column] access, and each vector of particle decay channels has additional data.

Does the order of this vector matter, or should it be a `set<DecayChannel>`?

src/celeritas/decay/DecayProcess.hh
```diff
@@ -0,0 +1,64 @@
...
+#include "celeritas/phys/ImportedProcessAdapter.hh"
+#include "celeritas/phys/Process.hh"
+
+namespace celeritas
+{
+class ParticleParams;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Process for decay.
```
sethrj, line 23: A little more documentation here would be good: this single class manages all possible particle decay channels, each of which is analogous to a model for the action process.

src/celeritas/inp/Physics.hh
```diff
@@ -95,6 +96,32 @@ struct OpticalPhysics
...
+//---------------------------------------------------------------------------//
+/*!
+ * Branching ratio and daughters for a decay channel.
+ */
+struct DecayChannel
+{
+    DecayChannelType type{DecayChannelType::size_};
+    double branching_ratio{};
+    std::vector<PDGNumber> daughters;
+};
```
sethrj, line 108: I'm not 100% sure I understand the input here. For muon decay, don't we know what the daughter types are? (Is this like specifying the secondary particle types for compton scattering?) Will other decay types need more data than just the branching ratios?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -171,6 +175,8 @@ struct ProcessFilter
                 return (which & DataSelection::em);
             case G4ProcessType::fOptical:
                 return (which & DataSelection::optical);
+            case G4ProcessType::fDecay:
+                return (which & (DataSelection::em | DataSelection::hadron));
```
sethrj, line 179: Could you add a test to the GeantImporter to see what decay models it pulls out?

src/celeritas/decay/DecayProcess.cc
```diff
@@ -0,0 +1,90 @@
...
+            // Identify the unique channel types
+            auto [iter, inserted] = types.insert(channel.type);
+            if (inserted)
+            {
+                // Construct an action for each channel
+                switch (*iter)
+                {
+                    case DecayChannelType::muon:
+                        result.push_back(std::make_shared<MuDecayChannel>(
+                            *start_id++, input_));
```
sethrj, line 52: Does this mean that for muons we'd create two separate step actions, one for each particle type?

sethrj (commented):
I'm still working through the PR. It looks like currently we have one step action per decay channel per particle type? I think we'd be better off treating the decay physics input like a collection of models starting with those we support (as we do with the new optical surface physics), rather than the old ImportX style of making everything generic with an enum to tie it together. Each model would provide a set of particles it applies to.

So for muon decay, we might have `optional<MuDecayChannel>` in the `DecayPhysics`. We wouldn't need `daughters` since each decay channel knows what it can decay and how the particle decays. (If we really wanted to allow more granularity, we could have the MuDecayChannel input take a set of parent particles, which normally would be {mu-,mu+} or empty.) A more complex/general decay channel such as a three-body decay (subset of G4PhaseSpaceDecayChannel functionality) would have a different set of data underneath it: a vector of {particle -> daughters} etc. User-specified decay channel input would be a generic function that returns a vector of decay channels (or something).

As for branching ratios, it might make sense to give them as an unnormalized value or as a decay constant (1/time), which would get normalized among all the decay channels for a particular particle type. For the current capabilities, we only have one decay channel so the ratio doesn't even matter yet.

The DecayProcess would then construct the different decay channels, categorize them by particle ID, and determine branching ratios by relative decay constants/values. The decay channel itself would change (I think?) so that it might contain data from multiple particle types rather than a single one. For the typical case of particle/antiparticle (muon, tau) this is sensible because the only difference in the physics is the daughter particle IDs (I think?).

I'm going to keep reviewing, and I know that this suggestion (assuming I'm not missing something important) is a major change, so we don't have to implement it as part of this PR. But if we do want to go in that direction, we could at least delete aspects that aren't yet needed and will be deleted in the future (I'm thinking specifically the decay channel enum).

whokion:
@amandalund I am mostly off this week, but I can catch up on this MR later if a second review is needed.

amandalund:
Thanks @sethrj for the comments! I'll look over the rest soon, but just to clarify it's one step action per _channel_, not per channel per particle type (I don't think we'd want that!)

@whokion no worries, enjoy the time off! If you have a chance when you're back I'd definitely appreciate your review to make sure I've got the physics right.

src/celeritas/phys/PhysicsStepUtils.hh
```diff
@@ -116,7 +116,13 @@ calc_physics_step_limit(MaterialTrackView const& material,
     for (auto ppid : range(ParticleProcessId{physics.num_particle_processes()}))
     {
         real_type process_xs = 0;
-        if (auto const& process = physics.integral_xs_process(ppid))
+        if (physics.process(ppid) == physics.scalars().decay)
+        {
+            // Calculate the macroscopic cross section as the inverse of the
+            // particle's decay length
+            process_xs = 1 / particle.decay_length();
```
whokion, line 123: I assume this is a technical tweak treating decay as part of a macroscopic interaction. In reality, decay is independent of particle–matter interactions and is an intrinsic process governed solely by the particle’s proper time. A better approach would be to let the interaction process and the decay process compete independently, with the final proposed step length whichever occurs first. I understood this was part of the motivation for introducing `InteractionProcess` and `DecayProcess` separately (along with other variations). Mixing these two intrinsically different processes through a material-dependent macroscopic cross section is somewhat physically unnatural, since the proper time is independent of the material.

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -643,6 +686,21 @@ CELER_FUNCTION ActionId PhysicsTrackView::model_to_action(ModelId model) const
     return ActionId{model.unchecked_get() + params_.scalars.model_to_action};
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Convert a sampled decay channel ID into a simulation action ID.
+ *
+ * Unlike the model-to-action ID conversion, there isn't a one-to-one
+ * correspondence between the channel ID and the action ID. Multiple channel
+ * IDs can map to a single channel action.
```
whokion, line 695: Could you clarify how multiple channel IDs map to a single channel action? Does this imply that the action is only depending on the multiplicity of daughters?

whokion (commented):
Overall, this looks reasonable to me, although the current implementation is somewhat over-engineered if we’re only supporting muon decay for now - of course, there is nothing wrong with being general from the start. Separately from this MR (perhaps by @stognini), we should add all daughters of the muon decay if we claim full support for it (in the context of the neutrino physics).

src/celeritas/phys/PhysicsStepUtils.hh
sethrj: From a computational perspective I think it's better to combine the probabilities of a change in particle state into a single random number sample. But you do bring up a very good point about length versus time: **is there any way this representation will interfere with assumptions about changing speed over the step length**? Does Geant4's decay process factor in an average speed, or does it assume the speed is constant with the starting speed? Can we improve up on that? (Or must we avoid improvements to better agree with existing results?)

whokion: In Geant4, the particle velocity is assumed to remain constant during decay-in-flight. This approximation is generally valid, particularly for the muon weak decay, since the mean decay length of a high-energy muon is relatively long (on the order of $\sim 2.2 \mu {\rm sec} \cdot c$). As a result, a muon rarely decays inside dense materials, where particle-material interactions would typically limit the step length first. If the decay occurs quickly, the associated energy loss over that step is to be negligible—especially for muons, which lose only minimum ionizing energy—since it is much smaller than their (eloss) range in this case. Thus, muon decay is primarily relevant in sparse media, where the muon velocity can be treated as approximately constant until decay.  The same assumption also holds for short-lived particles, of which lifetimes are so short that they decay essentially immediately in flight, before significant energy loss or material interactions occur. 

In principle, your point remains valid in situations where particle decay competes with material interactions on comparable length scales. However, such cases are not typical for common unstable particles in HEP detector simulations and, when they do arise, of course, should be handled with additional care even though I have not seen any good example.

sethrj: The application for muon decay being funded is muon catalyzed fusion which uses cold muons. So this may be an important effect we can model more accurately than geant4?

whokion: Not sure—probably @stognini can clarify. My assumption is that it would be undesirable for muons to produce Michel electrons during the catalyzed fusion process.

src/celeritas/phys/detail/DecayTableInserter.cc
```diff
@@ -0,0 +1,130 @@
...
+        accum_br += ch_inp.branching_ratio;
+    }
+    result.channel_ids
+        = channel_ids_.insert_back(channel_ids.begin(), channel_ids.end());
+
+    if (!(soft_equal<double>(1, accum_br)))
+    {
+        CELER_LOG(warning)
+            << "branching ratios for decay channels should sum to 1 "
+               "but instead sum tp "
```
stognini, line 108:
```suggestion
               "but instead sum to "
```

src/celeritas/inp/Physics.hh
stognini: I think I see where @amandalund is coming from with this name, since the PDG book has decay tables where the 1st column is the decay channel (so daughters here) and the 2nd column is the branching ratio. But `Vec` is definitely better in the code.

About the vector ordering: sorting channels by BR may be more efficient when selecting them. Specially in cases like pi0 -> 2gamma, where the BR is ~99%.

PR #1992, opened 2025-09-28
Update CMake requirements

sethrj:
I've installed macOS 26 and a completely fresh toolchain, including CMake 4.1.1, and verified everything works.

- Added a check for CXX_STANDARD compatibility with ROOT
- Updated `cmake_minimum_required` for all scripts
- Reorganized spack.yaml to make dependency requirements clearer


example/geant4/CMakeLists.txt
```diff
@@ -3,11 +3,10 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 #-----------------------------------------------------------------------------#
 
-cmake_minimum_required(VERSION 3.18)
+cmake_minimum_required(VERSION 3.18...4.1)
 project(CeleritasAccelExample VERSION 0.0.1 LANGUAGES CXX)
-cmake_policy(VERSION 3.12...3.30)
```
pcanal, line 8: Why is this no longer needed?

scripts/spack.yaml
```diff
@@ -1,37 +1,46 @@
...
+    - git-lfs
     - py-breathe
     - py-furo
-    - py-pre-commit
     - py-sphinx
     - py-sphinxcontrib-bibtex
     - py-sphinxcontrib-mermaid
-    - "root@6.28:"
-    - "vecgeom@1.2.10:1 +gdml"
+    # ... plus experimental libraries
```
pcanal: I hope `mpi` is not still experimental but rather using it in Celeritas is :)

scripts/spack.yaml
```diff
@@ -1,37 +1,46 @@
...
+    - mpi
   view: true
   concretizer:
     unify: true
   packages:
     root:
       # Note: ~gsl and ~math are removed because dd4hep requires them
```
pcanal: What is 'removed' referring to (from where/when)?

pcanal (approved):
LGTM

scripts/spack.yaml
sethrj: Indeed

example/geant4/CMakeLists.txt
sethrj: The `3.18...4.1` is the recommended way of stating the "known working" versions. It automatically sets the policy: see https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

scripts/spack.yaml
sethrj: I'd listed them in the recommended variants but removed them, you're right it's confusing.

PR #1993, opened 2025-09-29
Allow use of VecGeom solid model for surface-enabled build

sethrj:
It turns out that installing VecGeom with `VECGEOM_USE_SURF` does not *require* the use of the surface navigation. For the ATLAS hackathon, the installation of vecgeom has this option enabled, but the surface geometry doesn't yet work with the ATLAS model due to overlaps (according to @SeverinDiederichs).

A new conditional configuration option `CELERITAS_VecGeom_SURFACE` has been added to allow an override of `VECGEOM_USE_SURF`. I've tested all allowable combinations with vecgeom 2.0+surf and vecgeom 1.x.

---

@mrguilima This is undoubtedly going to add many more conflicts to your #1720. I recommend that, on your branch, you do the following:
- Replace all `defined(VECGEOM_USE_SURF)` -> `CELERITAS_VECGEOM_SURFACE`
- Replace all `#ifdef VECGEOM_USE_SURF` -> `#if CELERITAS_VECGEOM_SURFACE`
- Add the following to the top of the `vg/detail/VecgeomVersion.hh` file:
```c++
#ifdef VECGEOM_USE_SURF
#define CELERITAS_VECGEOM_SURFACE 1
#else
#define CELERITAS_VECGEOM_SURFACE 0
#endif
```

This should allow Git to resolve more of the conflicts automatically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -486,6 +486,12 @@ if(CELERITAS_USE_VecGeom)
     # Enforce minimum version
     celeritas_find_or_builtin_package(G4VG 1.0.3)
   endif()
+
+  # Default to using surface navigation if it's available
+  cmake_dependent_option(CELERITAS_VecGeom_SURFACE
+    "Use VecGeom surface navigation" "ON"
+    "VecGeom_USE_SURF" OFF
+  )
 endif()
 
 if(CELERITAS_BUILD_DOCS)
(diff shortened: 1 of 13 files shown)


drbenmorgan (approved):
I can only comment from the build perspective, but this seems to work fine for the ATLAS case where we picked this up. It builds fine (setting the `CELERITAS_VecGeom_SURFACE` to `OFF` in the build of the externals) against Vecgeom 2.0.0-rc7. @esseivaju has tested a build on Perlmutter and that runs fine so far. 

In the externals, we're just applying this as a patch for now, so it's fine if @mrguilima needs more time to rebase/work on the co-working changes.

sethrj:
Thanks @drbenmorgan ! I think we can merge this now, and I will work with @mrguilima to update his branch which is well overdue for merging at least in part.

PR #1994, opened 2025-09-30
Split corecel docs into multiple pages

sethrj:
This breaks up the overly long `codecel` user documentation page into a few individual ones, which makes it easier to read online. It also adds a few of the solver types to the math documentation.


src/corecel/math/TridiagonalSolver.hh
```diff
@@ -21,6 +21,8 @@ namespace celeritas
  *
  * This is a simplified form of Gaussian elimination that can solve a
  * tridiagonal system \f$ \mathbf{T} \mathbf{x} = \mathbf{b} \f$ in O(n) time.
+ *
+ * This class is to be used \em only during setup.
```
pcanal: Maybe add a hint why? (Just being 'time consuming' wouldn't necessarily require such a 'tight' restriction).

pcanal (approved):
LGTM.

src/corecel/math/TridiagonalSolver.hh
sethrj: It requires `std::vector` and has its code in a .cc file...

pcanal: I see.  Fair enough (by `setup` we here mean `cpu-only` code; ).
```suggestion
 * This class is to be used \em only during setup (as it can only run on CPU).
```

PR #1995, opened 2025-09-30
Fix option to kill tracks after user-specified number of steps

amandalund:
Tracks weren't being killed when `max_steps` was specified as the options wasn't being propagated to the `SimParams`.

### DIFF
--- a/src/celeritas/track/SimParams.hh
+++ b/src/celeritas/track/SimParams.hh
@@ -42,6 +42,7 @@ class SimParams final : public ParamsDataInterface<SimParamsData>
         // Construct with imported data and max field substeps
         static Input from_import(ImportData const&,
                                  SPConstParticles,
+                                 size_type max_steps,
                                  size_type max_field_substeps);
 
         //// DATA ////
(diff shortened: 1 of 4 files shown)


src/celeritas/track/SimParams.hh
```diff
@@ -42,6 +42,7 @@ class SimParams final : public ParamsDataInterface<SimParamsData>
         // Construct with imported data and max field substeps
         static Input from_import(ImportData const&,
                                  SPConstParticles,
+                                 size_type max_steps,
```
sethrj: Maybe we should pass `inp::TrackingLimits const&` instead of this and field substeps?

sethrj (approved):
In the interests of expedience for the hackathon please feel free to postpone (or add "to-do" somewhere like a github issue).

src/celeritas/track/SimParams.hh
amandalund: Done, and updated the `TrackingLimits` max field substeps to match the driver options.

sethrj (approved):
Thanks @amandalund !

PR #1996, opened 2025-09-30
Move loading of celer-g4 setup options to after run manager construction

amandalund:
`GlobalSetup::ReadInput()` needs `SharedParams::supported_offload_particles()`, which expects the run manager to already be constructed:

```console
app/celer-g4/celer-g4.cc:271: critical: While running cms-hgcal-g4.inp.json: /home/alund/celeritas_project/celeritas/src/accel/SharedParams.cc:216:
celeritas: precondition failed: G4RunManager::GetRunManager()
```


stognini (approved):
Thanks @amandalund !

app/celer-g4/celer-g4.cc
```diff
@@ -107,8 +107,6 @@ void run(std::string_view filename, std::shared_ptr<SharedParams> params)
 
     // Construct global setup singleton and make options available to UI
     auto& setup = *GlobalSetup::Instance();
-    // Read user input
-    setup.ReadInput(std::string(filename));
```
sethrj, line 111: Removing this breaks the tracing file definition (immediately below). Also since we hit a debug error in "user code" by doing the check out of order, we should probably upgrade that warning to a validate.

The real fix I think is to change the "supported offload particles" to a `variant<PDG,G4PD*,particle_group>` related to the `inp` that we've talked about constructing.

amandalund: Ah yes I missed the tracing file...

sethrj: If @esseivaju is ok with it, you could just move the tracing session after the run manager construction?

esseivaju: Yeah that should be fine, I don't think it's too important to capture the run manager instantiation

sethrj (commented):
Looking at *why* this could've been triggered... it looks like it's only when `input_.physics_options.muon` so not related to any of the ATLAS work?

amandalund:
Nope, just something I ran into a couple days ago when testing decay.

sethrj (approved):

PR #1997, opened 2025-10-01
Add covfie device test

esseivaju:
Running into a crash in Athena when enabling covfie field interpolation, so I'm adding a device test for `CartMapField`

### DIFF
--- a/test/celeritas/CMakeLists.txt
+++ b/test/celeritas/CMakeLists.txt
@@ -235,7 +235,7 @@ else()
   set(_fieldprop_filter)
 endif()
 
-celeritas_add_test(field/Fields.test.cc
+celeritas_add_device_test(field/Fields
   LINK_LIBRARIES Celeritas::Extcovfie
 )
 celeritas_add_test(field/Integrators.test.cc ${_fixme_cgs})
--- a/test/celeritas/field/Fields.test.cu
+++ b/test/celeritas/field/Fields.test.cu
@@ -40,7 +40,7 @@ __global__ void field_test_kernel(unsigned int const size,
                                   AxisGrid<real_type> x_grid,
                                   AxisGrid<real_type> y_grid,
                                   AxisGrid<real_type> z_grid,
-                                  Real3 n_samples,
+                                  Array<size_type, 3> n_samples,
                                   real_type* field_values)
 {
     auto tid = TrackSlotId{KernelParamCalculator::thread_id().unchecked_get()};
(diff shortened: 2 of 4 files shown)


test/celeritas/CMakeLists.txt
```diff
@@ -235,7 +235,7 @@ else()
   set(_fieldprop_filter)
 endif()
 
-celeritas_add_test(field/Fields.test.cc
+celeritas_add_device_test(field/Fields
```
sethrj, line 238: Should this need a "DISABLED" if not `CELERITAS_USE_covfie`?

test/celeritas/field/Fields.test.cu
```diff
@@ -0,0 +1,115 @@
...
+        for (size_type iy = 0; iy < ny_samples; ++iy)
+        {
+            real_type y = y_grid.min
+                          + iy * (y_grid.max - y_grid.min) / (ny_samples - 1);
+            y = celeritas::min(y, y_grid.max - 1);
+            for (size_type iz = 0; iz < nz_samples; ++iz)
+            {
+                real_type z = z_grid.min
+                              + iz * (z_grid.max - z_grid.min)
+                                    / (nz_samples - 1);
```
sethrj: I think `Interpolator` would be helpful here

sethrj (approved):
Good enough for government work!

test/celeritas/field/Fields.test.hh
```diff
@@ -0,0 +1,25 @@
...
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+// TESTING INTERFACE
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
+//! Run on device and return results
+void field_test(CartMapFieldInput&, Span<real_type>&, Real3&);
```
sethrj: Also needs a `CELER_NOT_CONFIGURED` inline definition when not using cuda/hip

test/celeritas/CMakeLists.txt
esseivaju: Can we disable only the device test?

sethrj: Oh sorry I forgot this is the whole field test, not just for covfie. It should be fine.

sethrj: Oh yeah and we have
```c++
#if !CELERITAS_USE_COVFIE
#    define CartMapFieldTest DISABLED_CartMapFieldTest
#endif
```
so this does what we expect.

PR #1998, opened 2025-10-01
Simplify perfetto integration

sethrj:
This removes the seemingly redundant `start` method from the perfetto tracing session (the construction of the session should just be delayed to where `start` is called). The global `flush_tracing` is replaced with `PerfettoSession::flush`, and log messages are used to inform about what's happening inside perfetto.


esseivaju (approved):
Looks all good to me once CI is fixed!

PR #1999, opened 2025-10-01
Improve autodetection of ANSI color support

sethrj:
This fixes the `use_color` result to be more consistent with other command-line utilities
- Previously piping to a file from within the shell would have `use_color=true` because `TERM` is set.
- GTEST_COLOR now shows up in `celeritas::environment()` if it's used (i.e., if it exists and CELER_COLOR doesn't)
- Now 1/0 or boolean-like values are respected since we're using `getenv_flag`.


pcanal (approved):
LGTM

PR #2000, opened 2025-10-02
Check vecgeom navindex table for consistency

sethrj:
I'll test this on a CUDA build in a couple of minutes but wanted to get it to y'all before your return from lunch :)


src/geocel/vg/detail/VecgeomVersion.hh
```diff
@@ -15,11 +15,11 @@
 #include <VecGeom/base/Version.h>
 #include <VecGeom/management/ABBoxManager.h>
 
-#if VECGEOM_VERSION < 0x020000 && CELERITAS_VECGEOM_SURFACE
+#define VECGEOM_V2 0x020000
```
mrguilima: How about a boolean VECGEOM_V2 being True if VECGEOM_VERSION >= 0x020000 otherwise False?

sethrj: Yeah, good idea. I think previously I'd had a workaround to support the "1.x with surfaces"...

esseivaju (approved):
Tried in Athena, check looks good!

PR #2001, opened 2025-10-02
Add Ranluxpp Random Number Generator

davidsgr:
Adds the new Ranluxpp random number engine to the Celeritas codebase conformant with the interface required for the generators.

Towards #2124


src/corecel/random/data/detail/RanluxppHelpers.hh
```diff
@@ -0,0 +1,217 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
```
sethrj: Since these are derived from others' (copyrighted) code, we need to make sure that the copyright reflects this. See https://github.com/celeritas-project/celeritas/tree/develop/src/geocel/vg/detail/BVHNavigator.hh and https://github.com/celeritas-project/celeritas/tree/develop/src/corecel/math/detail/AlgorithmsImpl.hh , and please update `doc/licenses/additional.rst`.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,212 @@
...
+    //! Pickle the state into the given array
+    CELER_FUNCTION void saveState(RanluxppArray9& state) const
+    {
+        std::copy_n(state.cbegin(), 9, state.begin());
+    }
+
+    //! Perform XOR operation on state
+    CELER_FUNCTION void xorState(RanluxppArray9 const& other_state)
+    {
+        std::transform(other_state.cbegin(),
```
sethrj: This won't work on device, so you'll probably want to write as a loop over indices.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,212 @@
...
+ */
+struct RanluxppRngState
+{
+    //// DATA ////
+    RanluxppArray9 state;
+    unsigned int carry;
+    int position;
+
+    //! Pickle the state into the given array
+    CELER_FUNCTION void saveState(RanluxppArray9& state) const
```
sethrj: With few exceptions, state objects shouldn't perform computations/operations, especially since their data is fully exposed. Here, you could delete the method and write `state.saveState(arr)` as `arr = state.state`. Anything that has access to the state itself has permission/knowledge of how to play with its internals.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,212 @@
...
+        RanluxppArray9 lcg;
+        celeritas::detail::toLCG(state, carry, lcg);
+        celeritas::detail::mulmod(kA, lcg);
+        celeritas::detail::toRanlux(lcg, state, carry);
+        position = 0;
+    }
+
+    // Skip 'n' random numbers without generating them
+    CELER_FUNCTION inline void
+    skip(RanluxppUInt n, int kMaxPos, RanluxppArray9 const& kA_2048, int offset);
```
sethrj: These should be functions on the engine/view rather than the state, especially since it combines "params data" (the kA_2048 constant) with "state data" (the local state) and modifies the state itself.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,212 @@
...
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize the state with the given seed
+ */
+// Initialize the state with the given seed
+CELER_FUNCTION void
+RanluxppRngState::initialize(RanluxppUInt seed, RanluxppArray9 const& kA_2048)
```
sethrj: See `XorwowRngEngine::operator=(Initializer_t const& init)`: this also requires both the params and state data and should be written an assignment operator, where the RHS is an "initialization" object that contains the seed.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,167 @@
...
+#include "corecel/random/data/detail/RanluxppMulMod.hh"
+#include "corecel/sys/ThreadId.hh"
+
+namespace celeritas
+{
+
+//---------------------------------------------------------------------------//
+/*!
+ * Implementation of the RanluxPP RNG Engine
+ *
```
sethrj: This is the place to document the key algorithms of the RNG and to provide a citation.

Could you please apply this patch:
```diff
diff --git a/doc/_static/zotero.bib b/doc/_static/zotero.bib
index 75087cd97..4300810af 100644
--- a/doc/_static/zotero.bib
+++ b/doc/_static/zotero.bib
@@ -1287,7 +1287,7 @@
   doi = {10.1051/epjconf/202429503018}
 }

-@article{hahnfeld-portableimplementation-2021,
+@article{hahnfeld-ranlux-2021,
   title = {A {{Portable Implementation}} of {{RANLUX}}++},
   author = {Hahnfeld, Jonas and Moneta, Lorenzo},
   editor = {Biscarat, C. and Campana, S. and Hegner, B. and Roiser, S. and Rovelli, C.I. and Stewart, G.A.},
```
and add a citation to the doxygen:
```
\citet{hahnfeld-ranlux-2021, https://doi.org/10.1051/epjconf/202125103008}
```

Finally, remember that the "brief" documentation on the first line needs a trailing period for Doxygen to parse correctly (since we're using the AUTODOC option and not explicitly putting `\brief`).

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,167 @@
...
+
+  private:
+    /// DATA ///
+    ParamsRef const& params_;
+    RanluxppRngState* state_;
+};
+
+//---------------------------------------------------------------------------//
+
+class RanluxppRngEngine final : public RanluxppRngEngineImpl<48>
```
sethrj: Instead of subclassing, could you use `if constexpr`? Or since it looks like 48 is the only value used in practice, make it a `static constexpr int offset = 48;`

src/corecel/random/params/RanluxppRngParams.hh
```diff
@@ -0,0 +1,44 @@
...
+#include "corecel/data/CollectionMirror.hh"
+#include "corecel/data/ParamsDataInterface.hh"
+#include "corecel/random/data/RanluxppRngData.hh"
+#include "corecel/random/data/RanluxppTypes.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Shared ata for Ranluxpp pseudo-random number generator
```
sethrj:
```suggestion
 * Shared data for Ranluxpp pseudo-random number generator.
```

test/corecel/random/RanluxppRngEngine.test.cc
```diff
@@ -0,0 +1,236 @@
...
+    //(9 * sizeof(RanluxppUInt) + sizeof(unsigned int) + sizeof(int)) * 8);
+    ASSERT_EQ(flattened.size() * sizeof(uint_t),
+              state_ref.size() * sizeof(RanluxppRngState));
+    ASSERT_TRUE(std::is_standard_layout_v<RanluxppRngState>);
+    ASSERT_EQ(0, offsetof(RanluxppRngState, state));
+    ASSERT_EQ(9 * sizeof(RanluxppUInt), offsetof(RanluxppRngState, carry));
+    ASSERT_EQ(9 * sizeof(RanluxppUInt) + sizeof(unsigned int),
+              offsetof(RanluxppRngState, position));
+    std::copy_n(
+        &state_ref.begin()->state[0], flattened.size(), flattened.begin());
```
sethrj: Unfortunately this isn't legal for the Ranlux state; it's a shortcut I took for the Xorwow since it's several identically-sized integers. Ranlux has 64-bit and 32-bit mixed together. Try:

```suggestion
    // Check that initial states are reproducibly random
    std::vector<RanluxppUInt> flattened;
    for (RanluxppRngState const& s : state_ref)
    {
        flattened.insert(flattened.end(), s.state.begin(), s.state.end());
        flattened.push_back(s.carry);
        flattened.push_back(s.position);
    }
```

sethrj (changes requested):
Great initial PR @davidsgr , especially since you're translating code from one totally new codebase to another totally new codebase.

I'll dig into the implementation more once the bit of suggested refactoring is done, but we should decide:
- Are we making a "thin wrapper" around an existing codebase (my original thought that proved impractical)? Or,
- Are we reimplementing an existing algorithm using Celeritas paradigms (like state+params)?

The real answer is somewhere between of course. I think though that the data/portability constraints make it more of the second. And so in that case:
- Let's combine the three detail headers into a single `RanluxppImpl.hh` file.
- Let's update the function names so that they're consistent with the celeritas style.
- Let's continue to separate the raw state from the compute methods (making it diverge further from the Copcore implementation).

The function name refactoring is the most potentially onerous, but I've had success using Copilot in that sort of thing: if you want to do a little pair programming early next week to implement these changes let's see how fast we can knock it out.

src/corecel/random/data/detail/RanluxppHelpers.hh
davidsgr: According to the comments in AdePT, the Ranluxpp engine is copyrighted by CERN, but under an LGPL license, while the rest of the AdePT code is under Apache.  Do you want me to take a stab at updating `additional.rst` to reflect this?

sethrj: Only if you want: it's no trouble for me to modify the additional.rst and push to your branch.

sethrj: (Apparently the licensing difference is that this was a modification of an extracted file from [ROOT](https://github.com/root-project/root)

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Fixed.

src/corecel/random/data/detail/RanluxppHelpers.hh
davidsgr: Ok, I took a stab at updating `additional.rst`

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Do I need to create an `Initializer` object?  I think if I implement this in RanluxppEngEngine, I only need to take the integer seed

src/corecel/random/data/RanluxppRngData.hh
davidsgr: These weren't even being used, so I just removed them

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Moved all of this into `RanluxppRngEngine`

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Changed.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Fixed.

src/corecel/random/params/RanluxppRngParams.hh
davidsgr: Fixed.

test/corecel/random/RanluxppRngEngine.test.cc
davidsgr: Updated test.

sethrj:
@davidsgr I see an IWYU error causing a build failure: https://github.com/celeritas-project/celeritas/actions/runs/18719387226/job/53387353724?pr=2001

missing `<algorithm>` I think. Could you ensure all the CI tests pass while I review this morning? Thanks!

sethrj:
@davidsgr Since the last review it looks like you've moved a lot of stuff in to `.cc` files: this unfortunately *will not work* in Celeritas. All code that's used inside a kernel has to be inlined. Setup code should be in .cc files where it's not needed for performance, but anything runtime has to be available to the downstream `.cu` kernel.   To fix this, you'll want to make sure your class definitions declare all their methods as `inline`,  then move all the code from the `.cc` file to the bottom of the `.hh` file.

Something to keep an eye out for: if you see a `CELER_FUNCTION` annotation in a `.cc` file, that's an error since it's declaring the function `__host__ __device__` which is only meaningful for CUDA/HIP.

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,130 @@
...
+#include "corecel/data/CollectionBuilder.hh"
+#include "corecel/random/data/RanluxppTypes.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize the state with the given seed.
+ */
+void initialize_state(RanluxppRngState& state,
```
sethrj: Is there a reason we're initializing states only on the CPU? This looks like it could be done in parallel on GPU.

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,130 @@
...
+{
+    CELER_EXPECT(params);
+
+    // Generate well-distributed seed numbers, including StreamId so that each
+    // stream has different starting contribution
+    std::vector<std::seed_seq::result_type> host_seeds(1, params.seed);
+    if (stream != StreamId{0})
+    {
+        host_seeds.push_back(stream.get());
+    }
```
sethrj: BTW this was only needed for reproducibility with results before we had streams; in general we should just push the stream ID.

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,130 @@
...
+ */
+void initialize_state(RanluxppRngState& state,
+                      RanluxppUInt seed,
+                      HostCRef<RanluxppRngParamsData> const& params)
+{
+    // Skip 2 ** 96 states
+    RanluxppArray9 a_seed = celeritas::detail::compute_power_modulus(
+        params.kA_2048, RanluxppUInt(1) << 48);
+    a_seed = celeritas::detail::compute_power_modulus(a_seed,
+                                                      RanluxppUInt(1) << 48);
```
sethrj: I think this state could be precomputed and stored in params, right? It's constant wrt the input seed.

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,130 @@
...
+
+    // Skip another s states.
+    a_seed = celeritas::detail::compute_power_modulus(a_seed, seed);
+    RanluxppArray9 lcg = {1, 0, 0, 0, 0, 0, 0, 0, 0};
+    lcg = celeritas::detail::compute_mod_multiply(a_seed, lcg);
+
+    // Set state and carry variable
+    state.state = celeritas::detail::to_ranlux(lcg, state.carry);
```
sethrj: Where is state.carry initialized?

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,130 @@
...
+
+    // Initialize state from the random seed
+    for (RanluxppRngState& state : states)
+    {
+        // Sample RNG to get seed for initialization
+        RanluxppUInt s = sample_uniform_int(rng);
+
+        // Initialize the state with the given seed
+        initialize_state(state, s, params);
+    }
```
sethrj: From the way that adept constructs  the seeds:
```cpp
track.rngState.SetSeed(initialSeed * event + startTrack + thread_id);
```
or for their asynchronous code
```cpp
__device__ inline uint64_t GenerateSeedFromTrackInfo(const AsyncAdePT::TrackDataWithIDs &trackInfo, uint64_t baseSeed)
{
  uint64_t input[12] = {baseSeed,
                        trackInfo.eventId,
                        trackInfo.trackId,
                        trackInfo.stepCounter,
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.eKin)),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.globalTime)),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.position[0])),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.position[1])),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.position[2])),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.direction[0])),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.direction[1])),
                        static_cast<uint64_t>(__double_as_longlong(trackInfo.direction[2]))};
  return fnv1a_hash64(input, 12);
}
```

Based on the `initialize_state` which does a *lot* of shuffling to the I'm not sure we need to be using truly random numbers for the seed. Maybe we can do this on GPU with just the event ID, stream ID, and track slot ID? (But honestly I don't know if they're doing it right: hashing into 64 bits means that best case you'd have a collision in 2^32, and there's way way more than 4B EM secondaries coming out of a production run.)

Also with @amandalund 's skip-ahead stuff with XORWOW we might be able to do something similar with xorwow. We should however try to find a way to distribute the samples over the entire RNG state space, and 64 bits obviously isn't enough to do that.

src/corecel/random/params/RanluxppRngParams.cc
```diff
@@ -0,0 +1,52 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Get the Ka polynomial.
+ */
+RanluxppArray9 const& RanluxppRngParams::getKa() const
+{
+    static RanluxppArray9 const k_array = {
+        0xed7faa90747aaad9,
```
sethrj: These might need `ull` suffixes to avoid truncation on some platforms.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,120 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Persistent data for the Ranluxpp random number generator.
+ */
+template<Ownership W, MemSpace M>
+struct RanluxppRngParamsData
+{
+    //// DATA ////
+    RanluxppArray9 kA_2048;
+    int kMaxPos;
```
sethrj: The use of `k` meaning "constant" is the old Hungarian notation and not something we do in Celeritas (and besides they're no longer constant. The names are also not very meaningful, so I would be happy to see them changed.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,120 @@
...
+    RanluxppUInt seed = 314159265;
```
sethrj: Also default-initialize this and the "max pos" to zero; they should be overwritten whenever the RNG params class is constructed.

src/corecel/random/params/RanluxppRngParams.hh
```diff
@@ -0,0 +1,47 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+//---------------------------------------------------------------------------//
+/*!
+ * \file corecel/random/params/RanluxppRngParams.hh
+ *
+ * Original source:
+ * https://github.com/apt-sim/AdePT/blob/master/include/AdePT/copcore/Ranluxpp.h
+ */
```
sethrj: This file definitely isn't derived from CERN... I don't want to worry about the philosophical/legal meaning of "derivative" and who owns copyright of what code, so I'm not going to insist that these all be changed, but I feel as a guideline that if it looks like their original code (modulo formatting) then it should have their copyright, but if there are enough structural/implementation differences to make it look completely different then it gets a celeritas copyright with a 'derivative' note. I am not remotely an expert in open source copyright law so I don't know what the right thing to do or the best practice is if we want to copy and substantially modify someone else's code, because then it's a mix of their work (their copyright) and our work (our copyright I think).

sethrj (commented):
Still working through this. Sorry if some of the comments seem like flip flopping: do you want to get together and go over some of this in person?

davidsgr:
> Still working through this. Sorry if some of the comments seem like flip flopping: do you want to get together and go over some of this in person?

Yeah, we can go over this.  I'm free all day today.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: It's actually initialized right there.  `state.carry` is an out-parameter for `to_ranlux`

src/corecel/random/params/RanluxppRngParams.hh
davidsgr: Ok, I'll leave this alone then.  Basically, the algorithm is theirs, but I've made changes throughout.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Fixed.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Fixed.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Fixed.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Clarified.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Now initialized with user-provided seed and trackid

src/corecel/random/params/RanluxppRngParams.cc
davidsgr: Fixed.

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Improved names

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Done.

davidsgr:
@sethrj   I think this is ready for another review.  The remaining CI failures do not appear to have anything to do with my PR.

test/corecel/random/RanluxppRngEngine.test.cc
```diff
@@ -0,0 +1,340 @@
...
+        = states.ref().state[AllItems<RanluxppRngState>{}];
+    std::vector<uint_t> flattened;
+    for (RanluxppRngState const& s : state_ref)
+    {
+        flattened.insert(
+            flattened.end(), s.value.number.begin(), s.value.number.end());
+        flattened.push_back(s.value.carry);
+        flattened.push_back(s.position);
+    }
+    static uint_t const expected_flattened[] = {
```
sethrj: The xorwow version of this test demonstrates that the RNG states on the second stream are independent to those on the first. (That's another instance of the test not documenting what it was doing, sorry.) In this test, the results are identical.

In Geant4 applications I don't think we use this way of initialization; we use the `RngReseedExecutor` via `reseed_rng`. So maybe it doesn't matter and we should delete this test 🤔

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    inline CELER_FUNCTION RanluxppUInt nextRandomBits();
+
+    //! Generate a uniformly random 64-bit integer by concatenating two
+    //! 32-bit words.
+    CELER_FUNCTION RanluxppUInt intRndm64()
+    {
+        // draw two 48-bit words, but take only their low 32 bits each
+        RanluxppUInt lo = this->nextRandomBits() & 0xFFFFFFFFu;
+        RanluxppUInt hi = this->nextRandomBits() & 0xFFFFFFFFu;
+        return (lo << 32) | hi;
```
sethrj: Just inline this into `operator()`

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    static CELER_CONSTEXPR_FUNCTION result_type min() { return 0u; }
+
+    //! Highest value potentially generated.
+    static CELER_CONSTEXPR_FUNCTION result_type max()
+    {
+        return celeritas::numeric_limits<RanluxppUInt>::max();
+    }
+
+    //! Initialize state with the given seed.
+    inline CELER_FUNCTION RanluxppRngEngine& operator=(RanluxppInitializer init)
```
sethrj: This function is long enough to interfere with "reading" the class definition, so it should be moved to a separate inline definition. I think the same is true for `advance` and maybe even `intRndm64`: honestly I would be happy with a single line being the threshold.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    {
+        // Just skip the next few entries in the currently
+        // available bits.
+        state_->position += n * offset_;
+        CELER_ASSERT(state_->position <= params_.max_position);
+        return;
+    }
+
+    n -= left;
+    // Need to advance and possibly skip over blocks.
```
sethrj: If you understand what "blocks" are, some discussion of that and the RNG structure would be helpful to add to the class docs.

src/corecel/random/data/RanluxppRngData.cc
```diff
@@ -0,0 +1,68 @@
...
+            size_type size)
+{
+    CELER_EXPECT(params);
+    CELER_EXPECT(stream);
+    CELER_EXPECT(size > 0);
+    CELER_EXPECT(M == MemSpace::host || celeritas::device());
+
+    // Move params to device
+    RanluxppRngParamsData<Ownership::value, M> p;
+    p = params;
```
sethrj: Update the comment to make it clearer that we're simply creating a temporary "native" copy of the params so that we can call `ranlux_state_init`.

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,118 @@
...
+    //! Ranluxpp state number and carry bit
+    RanluxppNumber value;
+    //! Current position in the state.
+    int position;
+};
+
+struct RanluxppInitializer
+{
+    //! Thread-local id.
+    RanluxppUInt thread_local_id;
```
sethrj: The documentation here is redundant with the variable name (and confusingly it's not an ID, it's an int); we should make it clear that this a "seed offset"  that needs to be different for each thread.

src/corecel/random/params/RanluxppRngParams.cc
sethrj: I don't see the fixx pushed here. Also since this function is only used by line 27, you might as well move it there:
```c++
host_data.seed_2048 = {0xed7faa90747aaad9ull, ...};
```

src/corecel/random/engine/detail/RanluxppImpl.hh
```diff
@@ -0,0 +1,656 @@
...
+    }
+    c += carry;
+
+    // c = floor(r / 2 ** 576) has been computed along the way via the carry
+    // flags. Now if c = 0 and the value currently stored in r is greater or
+    // equal to m, we need cbar = 1 and subtract m, otherwise cbar = c. The
+    // value currently in r is greater or equal to m, if and only if one of
+    // the last 240 bits is set and the upper bits are all set.
+    bool greater_m = r[0] | r[1] | r[2] | (r[3] & 0x0000ffffffffffff);
+    greater_m &= (r[3] >> 48) == 0xffff;
```
sethrj, line 268: This is a bitwise (masking) operation, rather than a logical in-place `x ← x && rhs`. I think it should be

```suggestion
    greater_m = greater_m && ((r[3] >> 48) == 0xffff);
```

src/corecel/random/engine/detail/RanluxppImpl.hh
```diff
@@ -0,0 +1,656 @@
...
+    for (int i : celeritas::range(9))
+    {
+        RanluxppUInt r_i = r[i];
+        r_i = sub_overflow(r_i, carry, carry);
+
+        RanluxppUInt t1_i = upper[i];
+        r_i = sub_carry(r_i, t1_i, carry);
+        r[i] = r_i;
+    }
+    int64_t c = -(static_cast<int64_t>(carry));
```
sethrj, line 189: I don't think the return result has to be anywhere close to 64 bits... since each `overflow` call sets the result to zero or 1, and `carry` simply increments up to one, `c` and `carry` are limited to the number of calls to add/sub, which are quite small. As the code below says, `c` is the carried/twiddled value of `r` over `2**576`.

src/corecel/random/engine/detail/RanluxppImpl.hh
```diff
@@ -0,0 +1,656 @@
...
+    //              bits 576 to 240 are zero and the rest is set.
+    // Note that all bits except the last are exactly complimentary (unless c =
+    // 0) and the last byte is conveniently represented by c already. Now
+    // construct the three bit patterns from c, their names correspond to the
+    // assembly implementation by Alexei Sibidanov.
+
+    // c = 0 -> t0 = 0; c = 1 -> t0 = 0; c = -1 -> all bits set (sign
+    // extension) (The assembly implementation shifts by 63, which gives the
+    // same result.)
+    int64_t t0 = c >> 1;
```
sethrj, line 466: OK, *this* is where `c` needs to be 64 bits.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    // Return the next random bits, generate a new block if necessary
+    inline CELER_FUNCTION RanluxppUInt nextRandomBits();
+
+    //! Generate a uniformly random 64-bit integer by concatenating two
+    //! 32-bit words.
+    CELER_FUNCTION RanluxppUInt intRndm64()
+    {
+        // draw two 48-bit words, but take only their low 32 bits each
+        RanluxppUInt lo = this->nextRandomBits() & 0xFFFFFFFFu;
+        RanluxppUInt hi = this->nextRandomBits() & 0xFFFFFFFFu;
```
sethrj:
```suggestion
        RanluxppUInt lo = this->nextRandomBits() & 0xffffffffu;
        RanluxppUInt hi = this->nextRandomBits() & 0xffffffffu;
```
for consistency with the rest of the code

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,118 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
```
sethrj: After consulting with ChatGPT I think the files that are *just* our interfaces (so the *Data.*, *Types.hh, *Params.hh. *StateInit*) can have our pure Celeritas copyright.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
```
sethrj: After consulting with ChatGPT let's go for this format in the files that are recognizably derivative of the CERN code.
```suggestion
// SPDX-FileCopyrightText: 2020 CERN
// SPDX-FileCopyrightText: 2025 Celeritas contributors
// SPDX-License-Identifier: (Apache-2.0 OR MIT)
```

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    CELER_ASSERT(remaining >= 0);
+    state_->position = remaining * offset_;
+    CELER_ASSERT(state_->position <= params_.max_position);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return the next random bits, generate a new block if necessary.
+ */
+CELER_FUNCTION RanluxppUInt RanluxppRngEngine::nextRandomBits()
```
sethrj: This, `numBits`, and a couple others should have their symbols updated to match our `snake_case` usage.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+
+    RanluxppArray9 lcg = celeritas::detail::to_lcg(state_->value);
+    lcg = celeritas::detail::compute_mod_multiply(a_skip, lcg);
+    state_->value = celeritas::detail::to_ranlux(lcg);
+
+    // Potentially skip numbers in the freshly generated block.
+    int remaining = n - skip * nPerState;
+    CELER_ASSERT(remaining >= 0);
+    state_->position = remaining * offset_;
+    CELER_ASSERT(state_->position <= params_.max_position);
```
sethrj:
```suggestion
    CELER_ENSURE(state_->position <= params_.max_position);
```

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+//---------------------------------------------------------------------------//
+// INLINE FUNCTIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Skip `n` random numbers without generating them.
+ */
+CELER_FUNCTION void RanluxppRngEngine::skip(RanluxppUInt n)
+{
+    CELER_ASSERT(n > 0);
+    CELER_ASSERT(params_.max_position > 0);
```
sethrj:
```suggestion
    CELER_EXPECT(n > 0);
    CELER_EXPECT(params_.max_position > 0);
```

Our contract macros expect/assert/ensuer match the [names chosen for C++20](https://omegaup.com/docs/cpp/en/cpp/language/attributes/contract.html), although ours are imperative rather than indicative because they are macro commands rather than attributes.

sethrj (approved):
Looks great, thanks for slogging through the refactors and initailization!

sethrj:
@davidsgr Did you see the comments above? Not sure if you're waiting on another review and there are several comments unresolved. I'll merge once they're addressed. Regarding the cmake, have you tried building and running all tests with the core RNG set to that? (I expect some of the tests will fail despite our being careful to guard against our assumption about the core RNG.)

davidsgr:
> @davidsgr Did you see the comments above? Not sure if you're waiting on another review and there are several comments unresolved. I'll merge once they're addressed. Regarding the cmake, have you tried building and running all tests with the core RNG set to that? (I expect some of the tests will fail despite our being careful to guard against our assumption about the core RNG.)

I'll start working through these comments.  I'm still getting used to the GitHub interface.  Not sure why I didn't see these before.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    {
+        return celeritas::numeric_limits<RanluxppUInt>::max();
+    }
+
+    //! Initialize state with the given seed.
+    inline CELER_FUNCTION RanluxppRngEngine& operator=(RanluxppInitializer init)
+    {
+        // Skip forward (params_.seed + init.thread_local_id) states
+        RanluxppArray9 new_a_seed = celeritas::detail::compute_power_modulus(
+            params_.seed_state, params_.seed + init.thread_local_id);
```
sethrj: Since the seed value is only used once, and the thread-local IDs are *always* contiguous and non-repeating, and we're using unsigned ints that wrap around naturally, adding the seed to the local ID is OK.  👍 

In the future we may want to hash the seed into other data to produce a "local seed":
https://github.com/apt-sim/AdePT/blob/3f9b7759ae44c89a66309f24ea33bf6a633c6d8a/include/AdePT/core/AsyncAdePTTransport.cuh#L159-L154

src/corecel/random/params/RanluxppRngParams.cc
```diff
@@ -0,0 +1,50 @@
...
+/*!
+ * Construct with a seed.
+ */
+RanluxppRngParams::RanluxppRngParams(RanluxppUInt seed)
+{
+    HostVal<RanluxppRngParamsData> host_data;
+
+    // Save basic data
+    host_data.seed = seed;
+    host_data.max_position = 9 * 64;
```
sethrj: I understand better what's going on in the RNG; should value needs to be constexpr inside the ParamsData (and could be written `sizeof(RanluxppArray9)*8` or `RanluxppArray9{}.size() * 64`): it's just the number of bits in `RanluxppArray9`.  The constexpr will allow the compiler to better optimize the "Advance" arithmetic.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+        // Just skip the next few entries in the currently
+        // available bits.
+        state_->position += n * offset_;
+        CELER_ASSERT(state_->position <= params_.max_position);
+        return;
+    }
+
+    n -= left;
+    // Need to advance and possibly skip over blocks.
+    int nPerState = params_.max_position / offset_;
```
sethrj: This will also be constexpr (# bits per state /  # bits per sample).

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    int skip = n / nPerState;
+
+    RanluxppArray9 a_skip = celeritas::detail::compute_power_modulus(
+        params_.state_2048, skip + 1);
+
+    RanluxppArray9 lcg = celeritas::detail::to_lcg(state_->value);
+    lcg = celeritas::detail::compute_mod_multiply(a_skip, lcg);
+    state_->value = celeritas::detail::to_ranlux(lcg);
+
+    // Potentially skip numbers in the freshly generated block.
```
sethrj, line 165: OK so a "block" is just the RNG state.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    int remaining = n - skip * nPerState;
+    CELER_ASSERT(remaining >= 0);
+    state_->position = remaining * offset_;
```
sethrj, line 168: I think it would make a lot more sense to write position as "index of the 48-bit slice" rather than "starting bit position". That would change all `position += offset_` to `++position`, `remaining * offset_` to `remaining`, `(params_.max_position - state_->position) / offset_` to `params_.max_position - state_->position`.

I'd also rename `max_position` to `num_chunks` or `num_words` or something, using chunks/words to indicate the unique 48-bit slice of the state.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+    CELER_ASSERT(state_->position <= params_.max_position);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return the next random bits, generate a new block if necessary.
+ */
+CELER_FUNCTION RanluxppUInt RanluxppRngEngine::nextRandomBits()
+{
+    if (state_->position + offset_ > params_.max_position)
```
sethrj, line 178: This would also become more clear:

```suggestion
    if (state_->word_index == params_.num_words)
```

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -0,0 +1,191 @@
...
+
+    RanluxppUInt bits = state_->value.number[idx] >> offset;
+    if (numBits < offset_)
+    {
+        bits |= state_->value.number[idx + 1] << numBits;
+    }
+    bits &= ((RanluxppUInt(1) << offset_) - 1);
+
+    state_->position += offset_;
+    CELER_ASSERT(state_->position <= params_.max_position);
```
sethrj: So this is just a fancy way of slicing out a 48-bit integer out of a large array of integers. I think a helper class to "view" the bit state could be helpful (but maybe redundant now that it's all working).

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -0,0 +1,118 @@
...
+ * Individual RNG state for Ranluxpp.
+ */
+struct RanluxppRngState
+{
+    //// DATA ////
+    //! Ranluxpp state number and carry bit
+    RanluxppNumber value;
+    //! Current position in the state.
+    int position;
+};
```
sethrj, line 77: One performance optimization we could try later will be to expand this into two arrays in  RanluxppRngStateData: `StateItems<RanluxppNumber>` and `Stateitems<very_short_int>` (where numeric_limits<very_short_int>() < 48, i.e. we could use an unsigned char if we wanted).

sethrj (commented):
@davidsgr I finally understand what this RNG is doing 🤪 it's basically a fancy LCG.

sethrj:
@davidsgr I took the liberty of adding some documentation (and moving the Impl file to engine/detail where it more closely belongs). Please `git pull --rebase` on your local branch!

sethrj:
Future performance improvement work for the RNG (if it ends up being expensive) should focus on register occupancy, since the current implementation has a lot of large temporary variables:
- In-place conversion between ranlux and lcg spaces
- Temporary async allocation of scratch space for the large multiplies
- Add a LCG impl view that references the global memory state, and stores temporary values like the carry int as register

src/corecel/random/params/RanluxppRngParams.cc
davidsgr: Now fixed.

test/corecel/random/RanluxppRngEngine.test.cc
davidsgr: Agreed.  I've deleted the test.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Done.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Done.

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Done.

src/corecel/random/data/RanluxppRngData.cc
davidsgr: Done.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Done.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Fixed.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Fixed.

src/corecel/random/params/RanluxppRngParams.cc
davidsgr: Done.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: Done.

src/corecel/random/engine/RanluxppRngEngine.hh
davidsgr: I believe the `position` refers to the index into the 576-bit state.  It (currently) advances by 48 bits at a time every time a sample is drawn.

I kinda like it this way because we could change the amount of data drawn from the state without making many changes to the algorithm.  I'll change it if you want, let me know.

sethrj (approved):
Excellent work @davidsgr ! This was fun.

PR #2004, opened 2025-10-03
Add json event reader and writer

amandalund:
This adds a lightweight json reader/writer for offloaded primaries, since HepMC3 isn't really designed for this type of output.


app/celer-sim/RunnerInput.cc
```diff
@@ -293,11 +294,14 @@ inp::StandaloneInput to_input(RunnerInput const& ri)
             [](inp::PrimaryGenerator const& pg) { return pg.num_events; },
             [](inp::SampleFileEvents const& sfe) { return sfe.num_events; },
             [](inp::ReadFileEvents const& rfe) {
-                if (ends_with(rfe.event_file, ".root"))
+                if (ends_with(rfe.event_file, ".json"))
```
sethrj: These are actually `jsonl` files (one event per line) so the extension should also be `jsonl`.

sethrj (approved):
This looks good, aside from the quibble about the format. Since the number of primaries may be large, it could improve readability to have one jsonl file per event, one line per primary (but for the sake of expediency we needn't worry about it.)

PR #2005, opened 2025-10-03
Use empty MFP grids to have zero cross section for model/material

sethrj:
This follow-on to #1913 uses empty grids, rather than grids with hardcoded infinities, to signify that a model does not apply in a material. (I.e., the MFP is infinite, or the interaction probability is zero.)


src/celeritas/optical/PhysicsTrackView.hh
```diff
@@ -214,59 +260,25 @@ CELER_FUNCTION ActionId PhysicsTrackView::discrete_action() const
...
 
+//---------------------------------------------------------------------------//
+// PRIVATE HELPER FUNCTIONS
 //---------------------------------------------------------------------------//
 /*!
  * Get the MFP grid ID for the given model.
  *
  * The grid corresponds to the optical material this track view was
- * constructed with.
+ * constructed with and is a simple 2D inedxing.
```
hhollenb: inedxing --> indexing

src/celeritas/optical/PhysicsTrackView.hh
```diff
@@ -167,6 +171,48 @@ CELER_FUNCTION bool PhysicsTrackView::has_interaction_mfp() const
...
+    CELER_EXPECT(model < this->num_models());
+
+    auto const& grid = params_.grids[this->mfp_grid(model)];
+    if (!grid)
+    {
+        // Model does not apply: cross section is zero
+        return 0;
+    }
+    NonuniformGridCalculator calc{grid, params_.reals};
+    real_type result = calc(value_as<Energy>(energy));
```
hhollenb, line 195: Did the grid change from MFP to XS? If so, then we should be renaming the `mfp_grid` functions and give some warnings about MFP vs XS grid interpolations. Otherwise, we probably want `1 / result`, right?

src/celeritas/optical/model/RayleighModel.cc
```diff
@@ -98,25 +97,8 @@ void RayleighModel::build_mfps(OptMatId mat, MfpBuilder& build) const
...
-            g.x = {energies.front(), energies.back()};
-            g.y = {std::numeric_limits<real_type>::infinity(),
-                   std::numeric_limits<real_type>::infinity()};
-            CELER_LOG(debug) << "Material " << mat.get()
-                             << " has no MFP data: setting mfp to infinity";
-
-            build(g);
-        }
+        // Build a grid with infinite MFP that prevents selection of the model
+        build();
```
hhollenb, line 101: Because of how invalid grids are handled in the `NonuniformGridInserter`, we can probably just write something like
```
if (input_ && input_.imported_materials->rayleigh(mat))
{
    // MFPS calculated from user properties
    ...
}
else
{
    // Invalid imported grids are inserted as empty grids that prevent model selection
    build(imported_.mfp(mat));
}
```

hhollenb (commented):
Definitely like removing the hard-coded infinities! I think there's just some minor things to sort out which places are using XS and which are using MFP. Maybe one of these days I'll get around to writing strongly typed grids...

src/celeritas/optical/PhysicsTrackView.hh
sethrj: Ack I meant to do that! Good catch.

src/celeritas/optical/model/RayleighModel.cc
Rashika-Gupta: I do remember trying to implement in the similar way using `NonuniformGridInserter` but was not successful ... :(

sethrj: I think we can do this with:
```c++
    if (mfp || !(input_.imported_materials
                 && input_.imported_materials->rayleigh(mat)))
    {
        // User explicitly provided Rayleigh MFP, or user properties are
        // unavailable
        build(mfp);
    }
    else
    {
        // MFPs can be calculated from user given properties
```
but I think it's much clearer to have the explicit second call.

Rashika-Gupta (approved):
looks good to me :)

PR #2006, opened 2025-10-03
Add polished surface roughness model

hhollenb:
Add the trivial perfectly polished surface roughness model. This is the default Geant4 behavior, and important for making sure facet normals are well defined for the reflectivity and interaction models. This is the simplest implementation, and can be refactored later when other builtin roughness models are added.


sethrj (approved):
Very polished pull request! 😉 In a follow-up, maybe once we have one of each model type, we can add documentation to the user manual.

PR #2007, opened 2025-10-03
Fix build issues with AMD ROCm

elliottbiondo:
I uncovered a few minor issues when compiling with the ROCm/Clang toolchain. With these fixes I can successfully run `celer-sim` on an MI300X on JLSE.

### DIFF
--- a/scripts/cmake-presets/jlse.json
+++ b/scripts/cmake-presets/jlse.json
@@ -37,8 +37,22 @@
     {
       "name": "release",
       "displayName": "Build with optimizations and disable debug assertions",
-      "inherits": [".base", ".ndebug"],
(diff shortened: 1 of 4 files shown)


elliottbiondo:
@sethrj RTR

sethrj (commented):
Could you please add:
1. the necessary `script/{env,cmake-presets}` files to build on the system,
2. documentation of which compiler versions failed (since we do have a ROCm toolchain as part of the CI)

Thanks!

elliottbiondo:
I updated the JSLE preset script for AMD. If I find out that there is shared space where I can put up a build, I will do so, and provide corresponding env files that point to the correct spack.

sethrj (approved):

PR #2008, opened 2025-10-03
Refactor dielectric model

hhollenb:
Dielectric-dielectric and dielectric-metal are readily combined since they differ just in how they handle refracted waves. Combined their inputs into the reflection form and a flag to handle if the interface is metal or not. Added a simple draft of the model and data which will use #1984 but currently doesn't do anything.

This touches a lot of files with pretty small changes, so putting it here before adding in `SurfaceInteractionApplier` logic.


src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -350,20 +350,28 @@ void GeantSurfacePhysicsLoader::insert_interaction(
     GeantSurfacePhysicsHelper const& helper, inp::ReflectionForm&& rf)
 {
     auto& interaction = models_.interaction;
+    CELER_LOG(info) << "GSPH dielectric size: "
+                    << interaction.dielectric.size();
```
sethrj: Is this a debug statement?

src/celeritas/inp/SurfacePhysics.hh
```diff
@@ -148,6 +148,31 @@ struct ReflectionForm
...
+/*!
+ * Analytic interactions between dielectric-dielectric and dielectric-metal
+ * interfaces.
+ */
+struct DielectricInteraction
+{
+    ReflectionForm reflection;
+
+    //! Whether the interface is dielectric-dielectric or dielectric-metal
+    bool is_metal;
```
sethrj: Add a default initializer for safety.

src/celeritas/optical/surface/model/DielectricInteractionData.hh
```diff
@@ -0,0 +1,214 @@
...
+namespace optical
+{
+//---------------------------------------------------------------------------//
+// TYPE ALIASES
+//---------------------------------------------------------------------------//
+// Dielectric interface type
+enum class DielectricInterface : bool
+{
+    metal = true,
+    dielectric = false,
```
sethrj: There shouldn't be any need to explicitly enumerate since it's an `enum class` (especially since it's "interface type" rather than IsDielectricMetal). And generally there should also be a good reason to switch the order from the default 😉 
```suggestion
    metal,
    dielectric,
```

src/celeritas/optical/surface/model/DielectricInteractionData.hh
```diff
@@ -0,0 +1,214 @@
...
+        reals = other.reals;
+        return *this;
+    }
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * A view into UNIFIED model grids to calculate reflection mode probabilities.
+ */
+class UnifiedReflectionView
```
sethrj: Could you put this into a separate header file? Also, if you made energy a construction argument and operated on the enum value one at a time, then it could be passed directly into the Selector class. (I'll add a test to make sure that Selector works with enum arrays.) That would work well with changing the data to be an enum array.

This would reduce register pressure and simplify the sampling algorithm.

The class name in this case could be ReflectionFormProbabilityCalculator (or some simplified version thereof).

src/celeritas/optical/surface/model/DielectricInteractionData.hh
```diff
@@ -0,0 +1,214 @@
...
+    metal = true,
+    dielectric = false,
+};
+
+// Supported reflection modes in UNIFIED model
+enum class UnifiedReflectionMode
+{
+    specular_spike,
+    specular_lobe,
+    back_scattering,
```
sethrj:
```suggestion
    backscattering,
```

src/celeritas/optical/surface/model/DielectricInteractionData.hh
```diff
@@ -0,0 +1,214 @@
...
+    //! \name Type aliases
+    using SurfaceGrids = Collection<NonuniformGridRecord, W, M, SubModelId>;
+
+    template<class T>
+    using Items = Collection<T, W, M>;
+    //!@}
+
+    SurfaceGrids specular_spike;
+    SurfaceGrids specular_lobe;
+    SurfaceGrids back_scattering;
```
sethrj: You could make this an EnumArray of collections and just leave the "diffuse" entry empty.

src/celeritas/optical/surface/model/DielectricInteractionData.hh
```diff
@@ -0,0 +1,214 @@
...
+    probs[UnifiedReflectionMode::back_scattering]
+        = this->calc_probability(data_.back_scattering[surface_], energy);
+
+    probs[UnifiedReflectionMode::diffuse_lambertian]
+        = 1
+          - (probs[UnifiedReflectionMode::specular_spike]
+             + probs[UnifiedReflectionMode::specular_lobe]
+             + probs[UnifiedReflectionMode::back_scattering]);
+
+    CELER_ENSURE(probs[UnifiedReflectionMode::diffuse_lambertian] >= 0);
```
sethrj: This reminds me, we should add a to-do about ensuring these models are on the same grid

src/celeritas/optical/surface/model/DielectricInteractionExecutor.hh
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include "DielectricInteractionData.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
```
sethrj, line 16: Add a placeholder indicator of some kind.

src/celeritas/optical/surface/model/DielectricInteractionModel.cc
```diff
@@ -0,0 +1,93 @@
...
+    SurfaceModelId id, std::map<PhysSurfaceId, InputT> const& layer_map)
+    : SurfaceModel(id, "interaction-dielectric")
+{
+    HostVal<DielectricData> dielectric_data;
+    HostVal<UnifiedReflectionData> reflection_data;
+
+    auto build_dielectric = make_builder(&dielectric_data.interface);
+    auto build_spec_spike = make_builder(&reflection_data.specular_spike);
+    auto build_spec_lobe = make_builder(&reflection_data.specular_lobe);
+    auto build_back_scatter = make_builder(&reflection_data.back_scattering);
```
sethrj: I need to officially deprecate `make_builder` 😬 and also I think it makes the constructors clearer if you name them the same as your data members, so you would end up with `interface.push_back(static_cast<DielectricInterface>(input.is_metal))`. (Which reminds me, instead of static casting to an interface there, you should rename the class or use a helper function that sits next to the enum.)

src/celeritas/optical/surface/model/DielectricInteractionModel.hh
```diff
@@ -0,0 +1,52 @@
...
+#include "celeritas/optical/surface/SurfaceModel.hh"
+
+#include "DielectricInteractionData.hh"
+
+namespace celeritas
+{
+namespace optical
+{
+//---------------------------------------------------------------------------//
+/*!
```
sethrj, line 20: Add basic documentation here. (The class is almost complete.)

sethrj (approved):
Nice! Please take the comments about the enum array as a suggestion; I think it's a good idea but would be easy to postpone as an "optimization".

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
hhollenb: yep it was, thanks for catching it

src/celeritas/optical/surface/model/DielectricInteractionData.hh
hhollenb: Do they need to be on the same grid so long as they add up to a total probability of 1?

sethrj: It would make sense to create a unified grid so we can just do the find+interpolate once, and we could make probabilities a setup time check that way too.

src/celeritas/optical/surface/model/DielectricInteractionData.hh
hhollenb: Maybe an implicit selector class would help here? Having the enum just be
```
enum class UnifiedReflectionMode
{
    specular_spike,
    specular_lobe,
    backscattering,
    size_
};
```
removes a lot of the exceptional behavior of the `diffuse_lambertian` case, which is only particularly relevant in the selector.

sethrj: Yeah, actually I'm working on that right now :)

src/celeritas/ext/detail/GeantSurfacePhysicsLoader.cc
```diff
@@ -174,22 +174,21 @@ bool is_probability(inp::Grid const& grid)
 inp::ReflectionForm
 load_unified_refl_form(GeantSurfacePhysicsHelper const& helper)
 {
-    inp::ReflectionForm refl_form;
-    helper.get_property(&refl_form.specular_lobe, "SPECULARLOBECONSTANT");
-    helper.get_property(&refl_form.specular_spike, "SPECULARSPIKECONSTANT");
-    helper.get_property(&refl_form.backscatter, "BACKSCATTERCONSTANT");
-    CELER_VALIDATE(refl_form, << "missing UNIFIED model reflection form grids");
+    static EnumArray<optical::ReflectionMode, char const*> labels{
+        "SPECULARSPIKECONSTANT", "SPECULARLOBECONSTANT", "BACKSCATTERCONSTANT"};
```
sethrj, line 178: Love it!

sethrj (approved):

PR #2009, opened 2025-10-05
Support unnormalized selection for a discrete PDF

sethrj:
This allows the `Selector` class, which samples on the fly from a discrete distribution, to allow a "remainder" (past-the-end) value to be sampled. If you specify a total probability of 1 with components {0.3, 0.2, 0.4}, there's a 10% chance the selector will return index 3. See https://github.com/celeritas-project/celeritas/pull/2008#discussion_r2404163467


src/corecel/random/distribution/Selector.hh
```diff
@@ -43,103 +37,26 @@ namespace celeritas
...
+make_unnormalized_selector(F&& func, T size, real_type total)
 {
-    real_type accum = 0;
-    for (IterT iter{}; iter != last_; ++iter)
-    {
-        accum += eval_(*iter);
-    }
-    return accum;
+    using NormTag = detail::UnnormalizedSelectorTag;
+    return {celeritas::forward<F>(func), size, total, NormTag{}};
```
hhollenb: Since `make_selector` and `make_unnormalized_selector` are just these specific forwards to the Selector constructor, it might make sense to put the normalized / unnormalized specific construction logic here and pass just the final `(eval, last, total)` to the Selector.

This would simplify some of the enum / tag dispatch / constexpr in the constructors.

src/corecel/random/distribution/detail/SelectorImpl.hh
```diff
@@ -0,0 +1,130 @@
...
+                                        value_type size,
+                                        real_type total,
+                                        SelNormTag<N>)
+    : eval_{celeritas::forward<F>(eval)}, last_{size}, total_{total}
+{
+    CELER_EXPECT(last_ != IterT{});
+    CELER_EXPECT(total_ > 0);
+    if constexpr (CELERITAS_DEBUG)
+    {
+        real_type debug_total = 0;
```
hhollenb: Can make this a free function in the detail namespace, and then check soft_equal in the normalized or <= in the unnormalized `make_selectors`

hhollenb (approved):
Thanks for getting to this so quick! I think the tag dispatching is a bit overkill and added a quick suggestion (I did spend an embarrassing amount of time trying to see if the constructor could just be explicitly templated but it can't).

Since it's all behind two functions and the templates aren't public, I'm happy whichever method you prefer.

src/corecel/random/distribution/Selector.hh
sethrj: Oh, that is a much better idea...

sethrj: But then again, the decrementing of the `last_` iterator is a subtle implementation detail that should probably stay hidden. I could also just pass the normalization as a bool and hope the compiler optimizes it in the usual case.

hhollenb: I was thinking have the Selector constructor itself not do any logic and accept handle `last` as the last iterator that can be returned by Selector (and the documentation states it won't evaluate `last`). Then Selector won't have any normalization dependence.
```
template<class F, class T>
CELER_FUNCTION Selector<F,T>::Selector(F&& eval, IterT last, real_type total)
    : eval_{celeritas::forward<F>(eval)}, last_{last}, total_{total}
{
    CELER_EXPECT(total_ > 0);
}
```
and then in the `make_selector` functions you can put some of the checks and normalization dependent logic:
```
template<class F, class T>
CELER_FUNCTION Selector<F,T> make_selector(F&& func, T size, real_type total = 1)
{
    Selector<F,T>::IterT last{size};
    // normalized debug checks...
    // Don't accumulate last value...
    --last;
    return {celeritas::forward<F>(func), last, total);
}

template<class F, class T>
CELER_FUNCTION Selector<F,T> make_unnormalized_selector(F&& func, T size, real_type total = 1)
{
    Selector<F,T>::IterT last{size};
    // unnormalized debug checks...
    return {celeritas::forward<F>(func), last, total);
}
```
I think with some forward declarations you can also make the Selector constructor protected / private and the make_selector functions friends, so outside code has to go through the safety checks of the normalized / unnormalized code.

sethrj: I would only want that if Selector is a `detail` function, and (see the latest update for how I refactored ElementSelector) I don't think we want that. I really think it's OK to have the boolean argument.

src/celeritas/em/interactor/AtomicRelaxation.hh
```diff
@@ -143,17 +141,25 @@ AtomicRelaxation::operator()(Engine& rng)
...
 
-        // Sample a transition (TODO: refactor to use Selector but with
-        // "remainder")
+        // Sample a transition using shell probabilities
         AtomicRelaxSubshell const& shell = shells[vacancy_id.get()];
-        TransitionId const trans_id = this->sample_transition(shell, rng);
-
-        if (!trans_id)
+        auto transitions = shared_.transitions[shell.transitions];
+        TransitionId const trans_id = make_unnormalized_selector(
```
hhollenb, line 147: It's a bit weird to keep wrapping and unwrapping the index in a type safe ID but never use that type safety anywhere. Not something to address in this PR but maybe something to address later.

hhollenb (approved):
Did another once over since more files got changed, looks good!

src/celeritas/em/interactor/AtomicRelaxation.hh
```diff
@@ -143,17 +141,25 @@ AtomicRelaxation::operator()(Engine& rng)
         if (vacancy_id.get() >= shells.size())
             continue;
 
-        // Sample a transition (TODO: refactor to use Selector but with
-        // "remainder")
```
amandalund, line 147: We also have a TODO for this in the photoelectric interactor, not sure if that's a straightforward update too?

sethrj: That one has two different use cases: one where it's accumulating cross sections as PDF values, and one where it's searching an already constructed CDF. I think it's not worth playing with right now since nothing else is touching (or in practice using) that code...

src/celeritas/em/interactor/AtomicRelaxation.hh
sethrj: Yeah to do this correctly, `ItemRange<AtomicRelaxTransition> transitions;` should really be `ItemMap<TransitionId, AtomicRelaxTransition> transitions;` and to make things easier I think we can define a new helper function on Collection:
```c++
    template<class U>
    CELER_FORCEINLINE_FUNCTION Collection<U, Ownership::const_reference, M>
    operator[](ItemMap<T, U>) const;
```

PR #2010, opened 2025-10-06
Add Ferrari-Cardano algorithm and general Quartic Solver test suite

osanstrong:
This pull request adds FerrariSolver, a class implementing the Ferrari-Cardano method to solve quartic polynomials of the form `ax⁴ + bx³ + cx² + dx + e = 0`, and QuarticSolver.test, a class with tests for quartic solvers. 

The Ferrari-Cardano method solves a related cubic equation to find a factor which breaks the quartic equation into two quadratic equations, which are then solved for the final roots. 

The QuarticSolver test file contains a brief abstract framework for testing quartic functions, and a specific subclass FerrariSolverTest which plugs in FerrariSolver for those tests. Most of the tests can then be easily reused to test an alternative quartic solver, with a different subclass.


src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    real_type r = 3 * qb2 * qb2 - ca_ * qb2 + da_ * qb - (e * a_inv_);
+
+    // Final roots to return
+    Intersections roots(no_intersection(),
+                        no_intersection(),
+                        no_intersection(),
+                        no_intersection());
+
+    // Edge case: equation is biquadratic TODO: need biquadratic tolerance
+    if (std::abs(q) <= 0)
```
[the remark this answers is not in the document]
sethrj: Ignore Copilot, but I suggest storing a `celeritas::SoftZero soft_zero_` object and perhaps initialize it with `Tolerance<real_type>::sqrt_quadratic()`?

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    real_type third_b = b * third;
+    real_type ninth_b2 = third_b * third_b;
+
+    // Intermediate values
+    real_type f = third * c - ninth_b2;
+    real_type g = third_b * (2 * ninth_b2 - c) + d;
+    real_type h = 0.25 * g * g + f * f * f;
+
+    // degenerate case, NEEDS TOLERANCE FOR ZERO CHECK
+    if (f == 0 && g == 0 && h == 0)
```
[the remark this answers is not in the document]
sethrj: Same: use `soft_zero_`

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    -> Roots2
+{
+    real_type qb2 = ipow<2>(hb);
+    if (qb2 > c)
+    {
+        // Two real roots
+        real_type ht = std::sqrt(qb2 - c);
+        return Roots2(-hb - ht, -hb + ht);
+    }
+    else if (qb2 == c)
```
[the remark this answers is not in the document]
sethrj: If this is an absolute comparison, you can use `soft_zero_(qb2 - c)`, otherwise you can use SoftEqual. See https://celeritas-project.github.io/celeritas/user/implementation/corecel/numerics.html#soft-equivalence .

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+ * The quartic equation \f[
+ * a x^4 + b x^3 + c x^2 + d x + e = 0
+ * \f]
+ * has four solutions mathematically, but we only require solutions which are
+ * both real and positive. This equation is also subject to multiple cases of
+ * catastrophic precision-limitation-based error both fundamentally and as a
+ * consequence of the particular algorithm chosen. This solver implements the
+ * Ferrari-Cardano method, which is well-established and simple, but more
+ * prone to numerical error than contemporary methods to be explored such as
+ * Algorithm 1010.
```
elliottbiondo:
```suggestion
 * Algorithm 1010 \citet{orellana-alg1010-2020, https://doi.org/10.1145/3386241}.
```
I added the Algorithm 1010 paper to the Celeritas Zotero library. Cite it here like so.

What's your favorite reference for Ferrari-Cardano? I'll add that as well and give you the citation line. Choose one that someone who wants implement the method can reference, i.e., probably not the original 1500's reference!

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Find positive, real, non-zero roots for quartic functions using the
+ * Ferrari-Cardano method.
+ *
+ * The quartic equation \f[
```
elliottbiondo:
```suggestion
 * The quartic equation
  * \f[
   a x^4 + b x^3 + c x^2 + d x + e = 0
 * \f]
```

elliottbiondo: We typically format equations like so

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+ * \f]
+ * has four solutions mathematically, but we only require solutions which are
+ * both real and positive. This equation is also subject to multiple cases of
+ * catastrophic precision-limitation-based error both fundamentally and as a
+ * consequence of the particular algorithm chosen. This solver implements the
+ * Ferrari-Cardano method, which is well-established and simple, but more
+ * prone to numerical error than contemporary methods to be explored such as
+ * Algorithm 1010.
+ *
+ * \return An Intersections array where each item is either a positive valid
```
elliottbiondo: N.B., @sethrj although this style of doc string are atypical in our codebase, this is how it's done in `QuadraticSolver`.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+ * intersection or the sentinel result \c no_intersection().
+ */
+class FerrariSolver
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using Intersections = Array<real_type, 4>;
+    using Roots2 = Array<real_type, 2>;
```
elliottbiondo: @sethrj these are the same types used by `InvoluteSolver`;  I am surprised you didn't want `Real4` and `Real2`?

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    // Final roots to return
+    Intersections roots(no_intersection(),
+                        no_intersection(),
+                        no_intersection(),
+                        no_intersection());
+
+    // Edge case: equation is biquadratic TODO: need biquadratic tolerance
+    if (std::abs(q) <= 0)
+    {
+        // QuadraticSolver solve(1, -2*p);
```
elliottbiondo: Make this block a private helper method

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+        place_root_sorted(roots, roots01[1] - qb);
+        place_root_sorted(roots, roots23[0] - qb);
+        place_root_sorted(roots, roots23[1] - qb);
+    }
+    return roots;
+}
+
+CELER_FUNCTION auto FerrariSolver::operator()() const -> Intersections
+{
+    // TODO: Either find an optimization to make here, or remove this entirely
```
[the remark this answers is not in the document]
elliottbiondo: I would go ahead on remove this. This is suppose to be the simple, baseline method. We can let Algorithm 1010 handle any optimziations.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    inline CELER_FUNCTION Intersections operator()(real_type e) const;
+
+    // Solve degenerate / known surface case (e = 0)
+    inline CELER_FUNCTION Intersections operator()() const;
+
+    // Find dominant root of normalized cubic
+    static inline CELER_FUNCTION real_type
+    dominant_root_normalized_cubic(real_type b, real_type c, real_type d);
+
+    // Find real quadratic roots TODO: is there a way to use existing one?
```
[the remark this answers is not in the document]
elliottbiondo: Remove todo: we can't combine this with the existing quadratic solver because we don't want to introduce any additional logic: the exist quadratic solver is a "tight inner loop" within the transport solve, and we shouldn't let support for torodial surfaces slow it down.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+                  real_type d,
+                  real_type e,
+                  SurfaceState on_surface);
+
+  public:
+    // Construct w/ a, b, c, d
+    inline CELER_FUNCTION
+    FerrariSolver(real_type a, real_type b, real_type c, real_type d);
+
+    // Solver fully general case
```
elliottbiondo: For all doc lines like this, as well as the commit message, use the infinitive form of the verb: "Solve in the fully general case"

(Which is also the imperative, if you want to look at it that way)

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+                  real_type b,
+                  real_type c,
+                  real_type d,
+                  real_type e,
+                  SurfaceState on_surface);
+
+  public:
+    // Construct w/ a, b, c, d
+    inline CELER_FUNCTION
+    FerrariSolver(real_type a, real_type b, real_type c, real_type d);
```
elliottbiondo: This is consistent with `QuadraticSolver`, and I think this is fine. The other option would be:

```FerrariSolver(Real3 abc, Real3 cde)```

with a single `operator()` method.

File: src/orange/surf/detail/FerrariSolver.hh [hunk 13]
```diff
@@ -0,0 +1,324 @@
...
+ */
+CELER_FUNCTION
+FerrariSolver::FerrariSolver(real_type a, real_type b, real_type c, real_type d)
+    : a_inv_(1 / a), ba_(b * a_inv_), ca_(c * a_inv_), da_(d * a_inv_)
+{
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Find all positive roots of x^4 + (b/a)x^3 + (c/a)x^2 + (d/a)x + (e/a) = 0
```
elliottbiondo: For declarations, doc strings should not end in periods (you have this correct). For definitions (i.e., like this one) you should end with a period.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    real_type qb2 = qb * qb;
+
+    // Incomplete quartic
+    real_type p = 3 * qb2 - 0.5 * ca_;
+    real_type q = 4 * qb * qb2 - ca_ * qb + 0.5 * da_;
+    real_type r = 3 * qb2 * qb2 - ca_ * qb2 + da_ * qb - (e * a_inv_);
+
+    // Final roots to return
+    Intersections roots(no_intersection(),
+                        no_intersection(),
```
elliottbiondo: use:
```
roots.fill(no_intersection())
```

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    FerrariSolver(real_type a, real_type b, real_type c, real_type d);
+
+    // Solver fully general case
+    inline CELER_FUNCTION Intersections operator()(real_type e) const;
+
+    // Solve degenerate / known surface case (e = 0)
+    inline CELER_FUNCTION Intersections operator()() const;
+
+    // Find dominant root of normalized cubic
+    static inline CELER_FUNCTION real_type
```
elliottbiondo: These two helper functions should not be static: they should be private helper functions.

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,308 @@
...
+};
+
+//---------------------------------------------------------------------------//
+/*
+ * Test cases with all non-zero roots, i.e., the ray does not start on or close
+ * to the surface
+ */
+TEST_F(FerrariSolverTest, no_roots)
+{
+    // x**4 + 2*x**3 - 2.999998*x**2 - 3.999998*x + 4.000005000001
```
elliottbiondo: Since comments are meant to be human-readable, use `^`

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+    //// UTIL ////
+    // Place real root in an ascending list
+    static inline CELER_FUNCTION void
+    place_root_sorted(Intersections& roots, real_type new_root);
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
```
elliottbiondo, line 117: Format definitions doc strings like:

```
/*!
 * Brief description, try to stick to one line, period at end.
 * 
 * As many additional lines as needed, formatted as a paragraph, wrapped
 * to 80 characters.
 */

File: src/orange/surf/detail/FerrariSolver.hh [hunk 18]
```diff
@@ -0,0 +1,324 @@
...
+CELER_FUNCTION auto FerrariSolver::operator()(real_type e) const
+    -> Intersections
+{
+    real_type qb = 0.25 * ba_;
+    real_type qb2 = qb * qb;
+
+    // Incomplete quartic
+    real_type p = 3 * qb2 - 0.5 * ca_;
+    real_type q = 4 * qb * qb2 - ca_ * qb + 0.5 * da_;
+    real_type r = 3 * qb2 * qb2 - ca_ * qb2 + da_ * qb - (e * a_inv_);
```
elliottbiondo: To square, use `ipow<2>` in `corecel/math/Algorithms.hh`. 

This function uses template recursion, so the compiler converts:

```
ipow<N> (x)
```
to bitcode equivalent to:
```
x*x*x* ... *x
```
This provides the readability of `std::pow` with the performance of pure multiplication.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,324 @@
...
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Utility function which places the given real root into an intersection list
+ * in increasing order.
+ *
+ */
+CELER_FUNCTION void
+FerrariSolver::place_root_sorted(Intersections& roots, real_type new_root)
```
elliottbiondo: Several things:

1. In the worse case, this requires something like 10 assignments. For readability, maintainability, and probably performance you are better off sorting at the end with `sort` in `corcel/math/Algorithms.hh`

2. For future reference, use `range` in `corece/cont/Range.hh` for your for loops, e.g.,

```
for (auto i : range(4))
{
   ...
}
```

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,308 @@
...
+        {
+            return solve_quartic(e);
+        }
+    }
+
+    void
+    expect_dominant_cubic_root(real_type const& expected, Coeffs3 const& bcd)
+    {
+        FS solve_quartic(1, 1, 1, 1);
+        real_type bigroot = solve_quartic.dominant_root_normalized_cubic(
```
elliottbiondo: It's generally not a good idea to make functions public for the purpose of testing. Can you refactor these tests such that they tests only the final answers, i.e. the roots?

test/corecel/math/QuarticSolver.test.cc
```diff
@@ -0,0 +1,308 @@
...
+//---------------------------------------------------------------------------//
+/*
+ * Test cases with all non-zero roots, i.e., the ray does not start on or close
+ * to the surface
+ */
+TEST_F(FerrariSolverTest, no_roots)
+{
+    // x**4 + 2*x**3 - 2.999998*x**2 - 3.999998*x + 4.000005000001
+    // Four complex roots 1+-0.001i, -2+-0.001i
+    {
```
elliottbiondo, line 68: When you add Alg1010 you will have to copy and paste all these cases in to a new file, which is not ideal. However, that is how it's done for the involute stuff, so I think this is fine for now.

elliottbiondo (changes requested):
Great work @osanstrong! You did an awesome job matching converting this to C++ and matching the style of rest of the codebase. Most of my comments are pretty minor. This big thing is to shore up your "soft" comparisons using a supplied tolerance, as Seth said.

scripts/cmake-presets/.json
```diff
@@ -0,0 +1,55 @@
+{
```
elliottbiondo, line 1: If this file is still giving you trouble, let me know and I will try to delete it on your branch.

src/orange/surf/detail/FerrariSolver.hh
sethrj: Sloppy code review on my part for the `Roots2` (we should change it); the `Intersections` is because the distance-to-boundary code uses that type alias. If it's strictly for returning intersection distances then it's ok to name it so.

src/orange/surf/detail/FerrariSolver.hh
sethrj: (And not just pure multiplication, because ipow<4>(x) is `(x*x)*(x*x)`, which is optimized to two multiplications; strict C++/IEEE compliance means that `x * x * x * x` is `x * (x * (x * x))`, which is three multiplications and may be different in the last binary digit.)

test/orange/surf/detail/QuarticSolver.test.cc
sethrj: If your class needs to expose bits of itself for testing purposes, it probably means it's too complicated, and you may want to break a piece off into a helper class.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: @osanstrong can you change `Roots2` to `Real2` here and also in `InvoluteSolver`?

test/corecel/math/QuarticSolver.test.cc
sethrj: We could include all the solvers in a single test, and use `TYPED_TEST_SUITE` to run multiple (see `corecel/math/SoftEqual.test.cc`, `orange/univ/detail/SenseCalculator.test.cc`).

src/orange/surf/detail/FerrariSolver.hh
sethrj: It's some of the older code we have (more or less copied from what I wrote in SCALE). I think it's good to describe the return value but, perhaps since return/parameter values often suffer from code rot, we have tended to omit them.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: Oh shoot, I didn't even realize it was recursing in that fashion. Good to know.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: So just changing the name of the type alias right?

src/orange/surf/detail/FerrariSolver.hh
osanstrong: 1. ) Cool, I'll go ahead and swap that out; didn't notice the sorting algorithms the first time
2. ) 👍; Is there a reason to declare the iterator `auto i` instead of `int i`?

elliottbiondo: `range` returns `size_type` so `size_type i : range(4)` is fine too. Since `range` is so pervasive, its easier to just write `auto`, but it does matter too much.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: Yep!

scripts/cmake-presets/.json
osanstrong: Sure, I have no objections to deleting it! I wasn't expecting trouble getting rid of it so much as I wanted to make sure it wasn't actually expected/necessary.

I was gonna delete it as one extra commit when I went to do all the other review concerns, but if you'd rather do it on your end that's also fine.

test/corecel/math/QuarticSolver.test.cc
osanstrong: I actually like the idea of `TYPED_TEST_SUITE`, I'll take a look at those to refactor if it seems appropriate here

scripts/cmake-presets/.json
elliottbiondo: Oh gotcha, yeah you can just delete it as you address these other changes.

src/orange/surf/detail/FerrariSolver.hh
sethrj: @elliottbiondo not quite right: the `range` return result is based on argument deduction. `4` is an `int`, so it the resulting type is `int`, and writing `size_type i : range(4)` will static-cast the int to an (unsigned) size_type at every iteration. This primarily matters when you mix iteration with index access; you can accidentally end up with truncation or signed/unsigned comparison.

scripts/cmake-presets/.json
osanstrong: After trying again it seems like this is definitely a file created when I ran the build script the first time
image
But it seems weird that it would make one just named `.json` 
image
Replacing `.json` with one which it was expecting (`<sysname>.json`, like the other config files) stops the issue, but it seems like for some reason `CMakeUserPresets.json` links to `scripts/cmake-presets/.json` on my machine:
image
which had been causing it to default to making that file. 

For now I think I'm gonna get rid of the `scripts/cmake-presets/.json` from the branch and keep my cmake config untracked, unless you think it's important to have something like that on the main project for just a personal device.

elliottbiondo: Yep that sounds good. No need for a personal .json to be in the main repo: it's a lot more useful for using shared machines, like for example if a bunch of different people are building on Frontier.

sethrj: Somehow it's not seeing a hostname, and the build script needs a fix for that.

Core developers get the privilege of keeping their presets in the repo ;) I think it's also useful for demonstrating different use cases/build environments for Celeritas.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: I'd probably go with the source I used in the report, the Polyanin&Manzhirov mathematical handbook:
```
@incollection{polyanin_ferrari_2007,
    title = {Polynomials and {Algebraic} {Equations}: {Ferrari}'s {Solution}},
    isbn = {978-1-58488-502-3},
    language = {en},
    booktitle = {Handbook of {Mathematics} for {Engineers} and {Scientists}},
    publisher = {Chapman \& Hall/CRC, Taylor \& Francis},
    author = {{Andrei D Polyanin} and {Alexander V Manzhirov}},
    year = {2007},
    pages = {161},
}
```

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Is there a different documentation style (and/or function with such a style) which would be more preferred for this then? This was also how `InvoluteSolver` documented the functor, but as per the comment on type aliases that might have also just gone unchecked.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: With the unique indentation and everything?

osanstrong: Also just now looking, I'm noticing some of the existing files put a `*` in front of the equation content line, while others don't. E.g. `src/celeritas/decay/interactor/MuDecayInteractor.hh` adds them, while `src/celeritas/em/distribution/MuAngularDistribution.hh` avoids them like your example just now. Are there contexts these would be different, or should they all be one way?

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Would it be preferable to define `qb3` and `qb4` like this for readability, or not since those would be introducing local variables only used once?

sethrj: It's generally better to just write `ipow<2>(qb)`, `ipow<3>(qb)` and `ipow<4>(qb)` directly when they're needed. Because you're operating on local variables in an inlined function, the compiler will be able to optimize them.

However, the way you're using it is as a polynomial evaluation, for which we have an efficient helper function in `corecel/math/PolyEvaluator.hh`.
Your version
```c++
real_type eval(real_type a, real_type b, real_type c, real_type d, real_type x)
{
    real_type qb  = 0.25 * b;
    real_type qb2 = qb * qb;
    return 3 * qb2 * qb2 - c * qb2 + d * qb - (x * a);
}
```
can be written much more cleanly:
```c++
    PolyEvaluator eval{-(x * a), d, -c, 0, 3};
    return eval(real_type{0.25} * b);
```
which also highlights that it's a quartic calculation in `ba_`, *not* in the input `x` (which seems strange to me, because aside from the scalar component, the evaluated value is invariant after construction).

Also note that `real_type qb = 0.25 * b` is implicitly
```c++
auto qb = static_cast<real_type>(0.25 + static_cast<double>(b));
```
which, for real_type=float, does a promotion then double-precision arithmetic then truncation. You may want to define a local `constexpr real_type half{0.5};`.

sethrj:
@osanstrong I know you're allergic to LLM bots but the kind of refactoring to replace `0.25` -> `constexpr real_type quarter{0.25};` and `q2 * q2 -> ipow<2>(q2)` is an ideal use case of copilot to save some drudgery. If you've got tests that pass, I would recommend trying it otu.

test/orange/surf/detail/QuarticSolver.test.cc
osanstrong: After a bit of tinkering, I found an extra quartic test case to add to confirm that all three paths of the helper function are taken in at least 1 test; I think that should be sufficient to catch if something breaks.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: This is fine as-is

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Should be addressed, but leaving this open to double check that I covered the polyeval and type casting correctly

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Our of curiosity, why this particular grouping? I had the impression putting `e` in `operator(e)`/`operator()` was a matter of declaring the particle was on the surface, but what's so special about putting abcde in two groups and picking `c` to be in both?

elliottbiondo: Sorry, this was my mistake. Take a look at `SimpleQuadratic`. The equation is:

```
ax^2 + by^2 + cz^2 + dx + ey + fz + g = 0
```

and the constructor takes

```
SimpleQuadric(Real3 const& abc, Real3 const& def, real_type g);
```

where `abc` and `def` are the second- and first-order terms, respectively, each set conveniently fitting into a `Real3`.

Looking at the quartic:

```
a x^4 + b x^3 + c x^2 + d x + e = 0
```

it can't be really split into `Real3`s too nicely (my mistake!). My new revised suggestion is:

```
class FerrariSolver
{
public:

    //! Input struct containing polynomial coefficents
    struct Coefficients
    {
         real_type a;
         real_type b;
         real_type c;
         real_type d;
         real_type d;
    }

    // Construct from coeffiecents
    inline CELER_FUNCTION  FerrariSoliver(Coefficients const& coeffs);

     // Solve for the roots
     inline CELER_FUNCTION Intersections operator()() const;
}
```
Basically, anytime you have more the 3-4 arguments, its usually preferred to make a input struct. Additionally, the only reason to supply `e` in the operator is if there is a use case where we set up `a`, `b`, `c`, and `d`, and the solve for the roots using multiple `e`s. I am not sure if there is?

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: Oops, the indentation was slightly screwed up in my example.

Yes, astute observation. Both styles are valid doxygen. I'd say omit them because I think that is what is done more often in ORANGE.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: Great, added it to the group library. You can now cite it using:

```
\citet{polyanin-ferrari-2007, https://doi.org/10.1201/9781420010510}
```

elliottbiondo:
@osanstrong this is shaping up nicely. Ping us whenever you are ready for us to take another look.

src/orange/surf/detail/FerrariSolver.hh
sethrj: Instead of a struct just use an `Array<real_type, 5> coeff`, so `coeff[0]` is the zeroth order terms, etc. You can locally expand it (maybe `auto [a, b, c, d, e] = coeff;` will work) for convenience.

And also I think that the QuadraticSolver's design may have also been deceiving; it tries to differentiate a major branch in the logic due to being degenerate or not. I think a better design would've been to have the "solver options/configuration" (aka min_a, tolerance) as the constructor arguments, and `operator()` takes the parameters.

So for your case, I would construct with a `real_type tolerance` that you use to build class data `SoftZero soft_zero_`. The `operator()` then would take `Array<real_type, 5>`.

sethrj: You could have an overload for `Array<real_type, 4>` for "on surface", but since you just treat "on surface" as "e=0" I don't see a big advantage in that case.

src/orange/surf/detail/FerrariSolver.hh
sethrj: Looks great to me!

src/orange/surf/detail/FerrariSolver.hh
osanstrong: I think I prefer this; I just reorganized to the functor being constructed with a tolerance and passing all of `abcde` as the operator argument. 

Though this does raise a couple of questions: 

- For the method manually specifying surface state, would it be better to have the method make a whole new array just to replace `e` with `0` so the original parameter is not modified (currently implemented but idk if this is bad for performance), or modify the original to avoid initializing a new array? 
```
// Clone
CELER_FUNCTION auto FerrariSolver::solve_general(Real5 abcde,
                                                 SurfaceState on_surface,
                                                 real_type tolerance)
    -> Intersections
{
    FerrariSolver solve(tolerance);
    auto selected_abcde = abcde;
    if (on_surface == SurfaceState::on)
    {
        auto [a, b, c, d, e] = abcde;
        selected_abcde = {a, b, c, d, 0};
    }
    return solve(selected_abcde);
}
```
```
// Modify
CELER_FUNCTION auto FerrariSolver::solve_general(Real5 abcde,
                                                 SurfaceState on_surface,
                                                 real_type tolerance)
    -> Intersections
{
    FerrariSolver solve(tolerance);
    auto selected_abcde = abcde;
    if (on_surface == SurfaceState::on)
    {
        auto [a, b, c, d, e] = abcde;
        selected_abcde = {a, b, c, d, 0};
    }
    return solve(selected_abcde);
}
```
- For that same method, it also uses a specified tolerance but this leads to both the method and the original constructor having a default tolerance argument; Is there a preferable way to to make this less redundant beyond just making a constant `default_quartic_tolerance` in the file? Or better to leave tinkering with this until actually implementing the Toroid surface where it'd actually be used?
```
  public:
    //!@{
    //! \name Type aliases
    using Intersections = Array<real_type, 4>;
    using Real2 = Array<real_type, 2>;
    using Real5 = Array<real_type, 5>;
    //!@}

    // Solve with manually specified surface state
    static inline CELER_FUNCTION Intersections solve_general(
        Real5 abcde,
        SurfaceState on_surface,
        real_type tolerance = Tolerance<real_type>::sqrt_quadratic());

  public:
    // Construct w/ given tolerance. Uses quadratic tolerance by default.
    inline CELER_FUNCTION
    FerrariSolver(real_type tolerance = Tolerance<real_type>::sqrt_quadratic());
```

osanstrong:
@elliottbiondo @sethrj Sorry for the three-week delay (caught up with a glut of uni work);  I think this should address the last couple comments from back then, with a bit of clarification on one of them that I put in the original conversation (About refactoring arguments for the functor constructor/operator)

osanstrong:
Also is there a good way to replicate the conditions of the automatic pull request test locally? The automatic test hosted on GitHub is catching a unit test fail that I think is relevant (finding a near-zero root instead of ignoring it) but I can't get that test to fail when I run it locally. 
The output says it's running `ctest --preset=${CMAKE_PRESET}-unit` but when I try locally there aren't any acceptable presets ending in `-unit`.

src/orange/surf/detail/FerrariSolver.hh
elliottbiondo: 1)

> would it be better to have the method make a whole new array just to replace e with 0 so the original parameter is not modified (currently implemented but idk if this is bad for performance), or modify the original to avoid initializing a new array?

Your "clone" and "modify" code blocks are the same, but I think I know what you are asking. I think in this case it's fine to do this:
                                                                                                                                                                               
```
CELER_FUNCTION auto FerrariSolver::solve_general(
               Real5 coeffs,
               SurfaceState on_surface,
               real_type tolerance)
    -> Intersections
{
    FerrariSolver solve(tolerance);
    if (on_surface == SurfaceState::on)
    {
         coeffs.e = 0
    }
    return solve(coeffs);
}
```

Basically, upon passing `coeffs` to `solve_general` a copy is made, and any changes that are made to `coeffs` are done on the local copy only. If `coeffs` was a huge object, and we didn't want to make a copy, we could pass by reference via `Real5& coeffs`. The issue with this that any changes made to `coeffs` would also be reflected in the caller's `coeffs`. This is called a "side effect" and is generally not a good idea. 

2) It's not necessary for these functions to take default tolerances. That could actually introduces bugs, as we need the tolerances to be consistent throughout the code.

elliottbiondo:
The automatic tests are the CI (continuous-integration) tests. The one that is failing for you is `noble-arm-clang-18-c++20` which means:

noble : Ubuntu 24.04, "Noble Numbat"
arm : processor using ARM instruction set, as opposed to the more standard x86 instruction set you see on Intel/AMD processors
clang-18: compiler being used
c++20: the language standard being used

So unfortunately there is no way to test locally, unless you have an ARM machine with Ubuntu and etc. I would say that this is not a common CI failure, but perhaps it's not surprising giving that you are specifically test floating point edge cases. My theory is that whatever sequence of ARM instructions being done is different enough from their x86 counterparts that it effects the floating point precision enough to flip the result of your test. I think the only thing you can do is change the test to be further from this flip point.

sethrj:
@osanstrong @elliottbiondo When an ARM test fails when the x86_64 build succeeds, it's almost always in some complex numerical function (as is the case here) and indicates a numerical instability in the calculations due to floating point arithmetic.

This is going to be hard to replicate locally: your best bet is to compile a RelWithDebInfo using `-O3 -march=native`, which will enable additional optimizations and advanced (sometimes "fused") instructions specific to your CPU. Those instructions can result in changes of a few ULP, which can add up and/or result in large differences when catastrophic floating point cancellation occurs.

Another option to replicate similar errors (which you'll see once the 'fast' ARM build passes) is to build with `CELERITAS_REAL_TYPE=float`, which will make cancellation errors even more apparent.

Once you're satisfied that there are no errors in the code (or obvious changes that could result in less cancellation), you can update the test to allow the different results, as we've done for the general quadric test:
```c++
    if constexpr (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)
    {
        EXPECT_SOFT_EQ(1e-16, distances[0]);
        EXPECT_SOFT_EQ(6.0, distances[1]);
    }
    else if (distances[1] == no_intersection())
    {
        // x86 hardware
        EXPECT_SOFT_EQ(6.0f, distances[0]);
    }
    else
    {
        // Apple Silicon
        EXPECT_SOFT_EQ(1e-7f, distances[0]);
        EXPECT_SOFT_EQ(6.0f, distances[1]);
    }
```

elliottbiondo:
@osanstrong @sethrj I think compiling with `CELERITAS_REAL_TYPE=float` is going to be the best bet. Your tolerances are going to have to be loose enough that the code passes using `float` as the `real_type`, and I would think that would require looser tolerances than any discrepancy caused by differences of instruction set or optimizations flags.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/surf/detail/FerrariSolver.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <cmath>
+#include <iostream>
```
elliottbiondo: While useful during debugging, please remove now that we are getting close to merging this.

osanstrong:
Cool! Compiling now with these:
`scripts/cmake-presets/OwenS360.json`:
```
{
  "version": 3,
  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0},
  "configurePresets": [
    {
      "name": "dev",
      "displayName": "dev: default development options",
      "inherits": [".debug", "default"],
      "cacheVariables": {
        "CMAKE_EXPORT_COMPILE_COMMANDS": {"type": "BOOL", "value": "ON"},
        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install",
        "CMAKE_BUILD_TYPE": {"type": "STRING", "value": "RelWithDebInfo"},
        "CELERITAS_REAL_TYPE": "float",
        "CELERITAS_USE_VecGeom": {"type": "BOOL", "value": "OFF"}
      }
    },
   ...
```
`scripts/build.sh`:
```
...
# Configure, build, and test
log info "Configuring with verbosity"
cmake --preset=${CMAKE_PRESET} --log-level=VERBOSE "$@"
log info "Building"
if cmake --build --preset=${CMAKE_PRESET} -O3 -march=native; then
...
```
Also, when using the build script, is there a recommended way to pass arguments to the build call of `cmake`? It feels like I'm missing a better way than either modifying the script to insert `-O3 -march=native` (what I did just now to check) or manually doing each step 

Update: these did make 3 of the tests fail, i'll let you know how it goes

osanstrong:
So:
- Reducing build precision to float from double does indeed cause some of unit tests right on the edge of having a real root to fail, and I could probably change these to not be right on the edge (as suggested), since I guess that's not really the focus of a unit test
- BUT, now that I check back on the github CI test that failed: that test was a surface root test. All that happens for a surface root is that it found a root slightly off from zero, which I can't reliably just move away from (unless there's a deeper connection I'm missing). Having a zero root is the point of the test. I'd just be randomly picking a different polynomial to test and hoping it doesn't run into the same precision issue.
- Is this something we should actually design around? Should the solver itself have a tolerance for what counts as zero or negative root? e.g. from
```c++
/*!
 * Attempt to put a value into the given list at given index, returning where
 * to place the next item.
 *
 * If the given value is no_intersection() or is not positive, does not place
 * the root, and returns the same index for the next one.
 */
CELER_FUNCTION int FerrariSolver::place_root(Intersections& roots,
                                             real_type new_root,
                                             int free_index)
{
    if (!(new_root == no_intersection() || new_root <= 0))
    {
        roots[free_index] = new_root;
        free_index += 1;
    }
    return free_index;
}
```
to
```c++
CELER_FUNCTION int FerrariSolver::place_root(Intersections& roots,
                                             real_type new_root,
                                             int free_index)
{
    if (!(new_root == no_intersection() || new_root <= 0 || _soft_zero(new_root)))
    {
        roots[free_index] = new_root;
        free_index += 1;
    }
    return free_index;
}
```
 Or is that something where the toroid code or broader program would just keep a tolerance for "this intersection was too close to the surface"
- It then also occurs to me that surface tests aren't unit tests in this case; the surface tests go down the same paths as the others, only introducing numerical precision errors like this. It might just not make sense to have them as unit tests, in contrast to the quadratic solver tests:
- The quadratic solver _does_ have a specific code path for surface solutions, because the non-zero roots of `ax² + bx = 0` are just the roots of `ax + b = 0`:
```c++
/*!
 * Solve degenerate case (known to be "on" surface).
 *
 * Since x = 0 is a root, then c = 0 and x = -b is the other root. This will be
 * inaccurate if a particle is logically on a surface but not physically on it.
 */
CELER_FUNCTION auto QuadraticSolver::operator()() const -> Intersections
{
    Intersections result{-2 * hba_, no_intersection()};

    if (result[0] <= 0)
    {
        result[0] = no_intersection();
    }

    return result;
}
```
- Should I just discard the surface tests as being stability tests instead of unit tests, or do we think it could be worth it for the solver to actually include a special case for `e = 0` .: `solve_quartic(ax⁴ + bx³ + cx² + dx = 0) -> solve_cubic(ax³ + bx² + cx + d = 0)`? If I rewrite the cubic solver already in `FerrariSolver` to be the full Cardano method (instead of the short-circuited version that only finds the dominant root), and rerun the Ferrari-Cardano numerical stability tests on the Python prototype for the analysis, that kills two birds with one stone:
    - Surface Tests: the surface tests, which are right now just stability tests, are now explicitly unit tests because this would mean there's a corresponding code path to test
    - Reproducibility: the version of the Ferrari-Cardano method is more accurate to the original textbook definition, instead of  `NKrvavrica`'s short-circuit of Cardano that only returns largest root (which actually shoot I forgot about referencing that; for the full published transaction I think I still need to add an aside to the Ferrari-Cardano explanation to also cite his specific implementation, or at least if I keep the short-circuited version)

TL;DR: I think I have three options:
- For either the surface cases or for all cases, include a tolerance so roots really close to zero are discarded
- Replace the short-circuited Cardano with the full Cardano method, and if `e==0` (or manually specified to be on-surface), just solve `ax³+bx²+cx+d=0` w/ Cardano method (the closest analogy to what `QuadraticSolver` does)
- Do nothing for now; treat this as a numerical stability issue and pick a different polynomial that doesn't happen to fail the CI unit tests. In fact possibly remove the surface tests, because in hindsight these aren't actually testing a different code path.

I think 2 probably requires rerunning the python analysis (which shouldn't take too long anyways), and 1 might get away without it (because this is just picking a tolerance to ignore bad results; the python analysis is how quickly run into those bad results in the first place)

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+    -> Intersections
+{
+    FerrariSolver solve(tolerance);
+    auto selected_abcde = abcde;
+    if (on_surface == SurfaceState::on)
+    {
+        auto [a, b, c, d, e] = abcde;
+        selected_abcde = {a, b, c, d, 0};
+    }
+    return solve(selected_abcde);
```
sethrj: Since `abcde` is a local variable you can simply do
```suggestion
    if (on_surface == SurfaceState::on)
    {
        abcde[4] = 0;
    }
    return solve(abcde);
```

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+    inline CELER_FUNCTION
+    FerrariSolver(real_type tolerance = Tolerance<real_type>::sqrt_quadratic());
+
+    // Solver for fully general case
+    inline CELER_FUNCTION Intersections operator()(Real5 abcde) const;
+
+  private:
+    //// DATA ////
+    // Soft zero for biquadratic and degenerate cubic detection
+    static inline SoftZero<real_type> const soft_zero_;
```
sethrj: Should be member data, not static
```suggestion
    SoftZero<real_type> const soft_zero_;
```

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Construct a solver instance with a specified tolerance for degenerate cases,
+ * such as the particle starting on the surface.
+ */
+CELER_FUNCTION
+FerrariSolver::FerrariSolver(real_type tolerance)
+{
+    SoftZero soft_zero_{tolerance};
```
sethrj: To construct as member data:
```suggestion
FerrariSolver::FerrariSolver(real_type tolerance)
    : soft_zero_{tolerance}
{
```

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+ * Find all positive roots of the polynomial with given a, b, c, d, e.
+ *
+ * The polynomial takes the form:
+ * \f[
+   ax^4 + bx^3 + cx^2 + dx + e = 0.
+ *\f]
+ * Where the given array abcde corresponds to {a, b, c, d, e}.
+ * Replaces negative or complex roots with sentry value no_intersection().
+ */
+CELER_FUNCTION auto FerrariSolver::operator()(Real5 abcde) const
```
sethrj: Since you immediately unpack these (and the coefficients aren't modified like the "general" case), you could/should use a const reference here. Since it's all inline, the compiler *should* optimize it to the same thing, though.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+        p, r, p * r - half * q * q);
+
+    real_type s2 = 2 * p + 2 * z0;
+    if (s2 >= 0)
+    {
+        real_type s = std::sqrt(s2);
+        real_type t;
+        if (soft_zero_(s))
+        {
+            t = z0 * z0 + r;
```
sethrj:
```suggestion
            t = ipow<2>(z0) + r;
```

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,219 @@
...
+
+    void expect_softeq_list(Intersections const& expected,
+                            Intersections const& actual)
+    {
+        ASSERT_EQ(expected.size(), actual.size());
+        for (auto i : range(expected.size()))
+        {
+            EXPECT_SOFT_EQ(expected[i], actual[i]);
+        }
+    }
```
sethrj: Replace this call with `EXPECT_VEC_SOFT_EQ(expected, actual);` (see https://celeritas-project.github.io/celeritas/user/development/testing.html#c.EXPECT_VEC_SOFT_EQ ) and you'll get clearer unit test failure messages too.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,347 @@
...
+    using Intersections = Array<real_type, 4>;
+    using Real2 = Array<real_type, 2>;
+    using Real5 = Array<real_type, 5>;
+    //!@}
+
+    // Solve with manually specified surface state
+    static inline CELER_FUNCTION Intersections solve_general(
+        Real5 abcde,
+        SurfaceState on_surface,
+        real_type tolerance = Tolerance<real_type>::sqrt_quadratic());
```
sethrj: I think with the updated constructor (that has a tolerance) this should be a regular member function, not a static function:

```suggestion
    // Solve with manually specified surface state
   inline CELER_FUNCTION Intersections operator()(
        Real5 abcde,
        SurfaceState on_surface);
```
so if you wanted to change the tolerance you'd write:
```c++
FerrariSolver solve{0.123};
solve({a, b, c, d, e}, SurfaceState::off);
```
etc.

sethrj:
Hi @osanstrong , you've got some thorough thoughts on a deep problem.
- The "broader program" never keeps track of whether intersections are too small, but as you know, we do keep track of whether we're logically *on* the surface of the toroid (i.e., maximum of three other roots, even if there's a slight numerical difference between the position and the actual surface of the toroid), as you have for the on-surface case.
- We do *not* have any way to track whether we're on a "double root" (e.g., tangent to the toroid) so can at best do a special case for cubic but not quadratic.
- We *do* prevent zero distances because that will lead to infinite loops, but that's covered by your `place_root` function.
- It is allowable to have small distances so I don't think we want to add a soft-zero to `place_root`.

> Should I just discard the surface tests as being stability tests instead of unit tests, or do we think it could be worth it for the solver to actually include a special case for e = 0 .: solve_quartic(ax⁴ + bx³ + cx² + dx = 0) -> solve_cubic(ax³ + bx² + cx + d = 0

I think it is very important at this level to check and document the behavior of the solver when we're *logically on a surface*, but we don't have to necessarily test for the case "e = 0". For physical points near the surface we're going to assume there's always some fuzziness about every one of those coefficients, so from the bigger picture we don't really need to have a consistent behavior for `e = 0`. (And there's *always* going to be a sharp discontinuity in the solver behavior, whether it's at exactly `e = 0`, or exactly  `e = ±ε`.)

In other words, a special case for the cubic solve is indeed probably the way to go. You would add a `Intersections operator()(Real4 const&);`, and instead of the `on_surface == SurfaceState::on` setting e=0 and running the full solver, it would just return `(*this)(Real4{a,b,c,d})`.

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,219 @@
...
+
+    void expect_surface_roots_from_coeffs(
+        std::initializer_list<real_type> const& expected, Coeffs4 const& abcd)
+    {
+        auto [a, b, c, d] = abcd;
+        auto x = this->get_roots({a, b, c, d, 0});
```
sethrj: This also suggests you should have a `operator()(Real4 const&)` for your solver.

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,219 @@
...
+ * to the surface
+ */
+
+TYPED_TEST(QuarticSolverTest, no_roots)
+{
+    // x^4 + 2*x^3 - 2.999998*x^2 - 3.999998*x + 4.000005000001
+    // Four complex roots 1+-0.001i, -2+-0.001i
+    {
+        this->expect_nonsurface_roots_from_coeffs(
+            {}, {1, 2, -2.999998, -3.999998, 4.000005000001});
```
sethrj: I recommend changing `fill_inf` to a free function `make_roots`, declaring your solver locally, and writing all your tests as:
```c++
TypeParam solve{}; // Solver with default tolerance

// Four complex roots...
EXPECT_VEC_EQ(make_roots({}), 
        solve({1, 2, -2.999998, -3.999998, 4.000005000001}));

// Two negative real roots...
EXPECT_VEC_EQ(make_roots({}), ...);

// etc
```

Your output will be clearer and your test smaller.

sethrj (commented):
Top-notch work! A few comments based on the discussion.

osanstrong:
When I was reviewing the Cardano method and other options for cubic solvers, I found a more comprehensive description of what I think [1728 systems](https://www.1728.org/cubic2.htm) (cubic implementation I referenced through NKvravica) used, which I think is pretty similar what Algorithm 1010 used. I'm just kicking myself for not looking closely enough at the cubics when I first did lit review over the summer, because I should have just used the comprehensive one from the start.

The [Numerical Recipes method](https://faculty.kfupm.edu.sa/phys/aanaqvi/Numerical%20Recipes-The%20Art%20of%20Scientific%20Computing%203rd%20Edition%20(Press%20et%20al).pdf) (Press et. al, 2007), seemingly used by 1728 systems but [poorly documented](https://www.1728.org/cubic2.htm), and definitely used by Algorithm 1010 and the "[PRACTICAL METHOD](https://quarticequations.com/Cubic.pdf)" I found posted this month, is a combination of the Cardano method and the trigonometric method, as opposed to the straight Cardano method. Specifically, it avoids using complex numbers by calculating the three real roots case using trigonometric functions, which is why it used trig functions and not a complex library.

For one thing, this makes our previous description of "Ferrari-Cardano method" incorrect, because it's definitely not the simple Cardano method being plugged in here. It's also a little annoying because it means the cubic solver I used in Ferrari vs Algorithm 1010 was slightly different but could have just been the same if I was paying attention.
I'm going to start by implementing the full numerical recipes method and changing the description to clarify that it's something different than the original Cardano method (Ferrari-NR?), and this is going to be something I have to change for the ANS submission. 

In the meantime, since we still only need real roots, the main decision is whether to use complex numbers (Cardano) or to use trigonometric functions (the Numeric Recipes optimization in Algorithm 1010, and similar to the Ferrari implementation I referenced). 

Is it worth comparing them / Is there one of these two (complex math or trigonometric functions) that's a huge issue for GPU implementations? I've experienced trig functions being an issue for GPU performance when I was going amateur OpenGL stuff, but it a) seems like trig functions are used elsewhere in celeritas and b) this isn't that much extra code complexity, which seemed like the bigger concern bc of registry usage. Essentially:
```
# Numerical Recipes, described to be more robust but uses trig functions
    if h <= 0:
        j = math.sqrt(-f) 
        k = math.acos(-0.5*g / (j*j*j))
        m = math.cos(third*k)
        n = sqr3 * math.sin(third*k)
        r1 = 2*j*m - a13
        r2 = -j * (m + n) - a13
        r3 = -j * (m - n) - a13 # Is this enough temporary variables to warrant analyzing how necessary it is? Vs. just using it 
        return r1, r2, r3

    else: 
        sqrt_h = cmath.sqrt(h)
        S = cubic_root(-0.5*g + sqrt_h)
        U = cubic_root(-0.5*g - sqrt_h)
        S_plus_U = S + U
        r1 = S_plus_U - a13
        return r1
```
```
# Cardano, the original 
        sqrt_h = cmath.sqrt(h)
        S = cubic_root(-0.5*g + sqrt_h)
        U = cubic_root(-0.5*g - sqrt_h)
        S_plus_U = S + U
        r1 = S_plus_U - a13
        if (h > 0): # Only one real root
             return r1
        S_minus_U = S - U
        r2 = -0.5*S_plus_U - a13 + S_minus_U*sqr3*0.5j # These cancel out to real numbers
        r3 = -0.5*S_plus_U - a13 - S_minus_U*sqr3*0.5j # "
        return real(r1), real(r2), real(r3)
```

So I'm curious for opinions if it's worth testing if we need the precision improvements, or if that'd just be [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) since this isn't that big of a difference compared to the quartic algorithm complexity.

osanstrong:
There's also the new Practical Method, which isn't published in a journal as near as I can tell, which seems to be about the same complexity as in Numerical Recipes with one extra if statement:
image

osanstrong:
Update I think I really just want to switch entirely to Practical Algorithms, and redo the precision loss analysis if necessary, because not properly lit-reviewing the cubic solver used by the Ferrari implementation I referenced has been a never ending source of citation issues.

I.e. switch to using [Numerical Recipes' algorithm](https://iate.oac.uncor.edu/~mario/materia/nr/numrec/f5-6.pdf) from the one I copied over the summer, which looked very similar but the citation just goes to a poorly documented [calculator website](https://www.1728.org/cubic2.htm)

And then from there if necessary I can compare the NR cubic (which would be shared between ferrari and Alg1010) to the original Cardano or to the slightly different/newer [Practical Method](https://quarticequations.com/Cubic.pdf).


BibLaTeX citation for the Numerical Recipes cubic algorithm, if it looks small enough to not analyze thoroughly:
```
@incollection{press_56_2007,
	location = {Cambridge},
	edition = {2. ed., repr., corr. to software version 2.10},
	title = {5.6 Quadratic and Cubic Equations},
	isbn = {978-0-521-43064-7},
	series = {Fortran numerical recipes},
	number = {1},
	booktitle = {Numerical recipes in Fortran 77: the art of scientific computing},
	publisher = {Cambridge University Press},
	editor = {Press, William H.},
	date = {2007},
	langid = {english},
	file = {PDF:C\:\\Users\\Owen\\Zotero\\storage\\CB6KJZAC\\Press - 2007 - Numerical recipes in Fortran 77 the art of scientific computing.pdf:application/pdf},
}

```

test/orange/surf/detail/QuarticSolver.test.cc
osanstrong: Not a bad idea, but with this implemented it makes passing an initializer list ambiguous between the two function overloads, so the tests have to specify whether it's an array of 4 or 5:
```
TYPED_TEST(QuarticSolverTest, four_roots)
{
    TypeParam solve{};
    // x^4 - 10*x^3 + 35*x^2 - 50*x + 24
    // Four roots at 1, 2, 3, 4
    EXPECT_VEC_SOFT_EQ(
        make_roots({1.0, 2.0, 3.0, 4.0}),
        solve(Real5{1, -10, 35, -50, 24})
    );
}
```
I actually think this is fine, and if anything might be a little clearer bc it keeps it specified whether it's putting in 4 or 5 coefficients.

sethrj:
@osanstrong you may want to start building locally with `-Werror -Wall -Wextra` (incidentally, I fixed the error in the `build.sh` script that kept it from making the correct preset filename for your machine). The [CI is failing](https://github.com/celeritas-project/celeritas/actions/runs/19781254003/job/56681727399?pr=2010):
```
In file included from /home/runner/work/celeritas/celeritas/test/orange/surf/detail/QuarticSolver.test.cc:10:
/home/runner/work/celeritas/celeritas/src/orange/surf/detail/FerrariSolver.hh:302:25: error: unused variable 'half' [-Werror,-Wunused-variable]
    constexpr real_type half = real_type{0.5};
                        ^
/home/runner/work/celeritas/celeritas/src/orange/surf/detail/FerrariSolver.hh:396:15: error: unused variable 'z0' [-Werror,-Wunused-variable]
    real_type z0 = FerrariSolver::dominant_root_normalized_cubic(b, c, d);
              ^
/home/runner/work/celeritas/celeritas/src/orange/surf/detail/FerrariSolver.hh:401:15: error: unused variable 'sqrt3' [-Werror,-Wunused-variable]
    real_type sqrt3 = std::sqrt(3);
              ^
```

(Also, for the last one, you should be using `sqrt_three` from `corecel/Constants.hh`, which is compile-time as opposed to invoking the math library on every call (!).)

Thanks again for your hard work on this!

elliottbiondo:
@osanstrong sounds like there are some strategy decisions to be made. May we should meet on this? I am free all week.

To me, the original plan to test the two extremes makes sense, i.e., the simplest/quickest possible method, Ferrari-Cardano, versus the most robust, Algorithm 1010. Since you found working precision to be the dominant factor, it seems like using quad precision for an isolated quartic order root finding step with the least amount of floating point operations is a compelling option.

With regards to trig functions vs. complex value keep in mind that this doing trig functions under the hood:

```
        S = cubic_root(-0.5*g + sqrt_h)
        U = cubic_root(-0.5*g - sqrt_h)
 ``` 

in the case where the argument is imaginary.

If we decide to switch to NR, it might make sense to make that a second MR.

**Comment by elliottbiondo** [comment]
For posterity, @osanstrong and I decided that it's better to switch to using the Numerical Recipes cubic solver because 1) it's more citable, 2) it's whats used with Algorithm 1010, 3) it's pretty similar to what we originally had, so it's not a big change.

Owen has already updated this MR accordingly.

@osanstrong your new citation key is `press-numericalrecipes-2007`

File: src/orange/surf/detail/FerrariSolver.hh [hunk 33]
```diff
@@ -0,0 +1,396 @@
...
+#include "orange/surf/detail/QuadraticSolver.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Find positive, real, non-zero roots for quartic functions using the
+ * Ferrari-Cardano method\citet{polyanin-ferrari-2007,
```
elliottbiondo: Update terminology appropriately here and throughout

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+
+    // Solver for surface case
+    inline CELER_FUNCTION Intersections operator()(Real4 const& abcd) const;
+
+  private:
+    //// DATA ////
+    // Soft zero for biquadratic and degenerate cubic detection
+    SoftZero<real_type> const soft_zero_;
+
+    //// UTIL ////
```
elliottbiondo: We usually call these `//// HELPER FUNCTIONS ////`

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+    // Find real quadratic roots
+    inline CELER_FUNCTION Real2
+    real_roots_normalized_quadratic(real_type b, real_type c) const;
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+
+//---------------------------------------------------------------------------//
```
elliottbiondo:
```suggestion
//---------------------------------------------------------------------------//
```

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+/*!
+ * Solve for the real roots of a cubic function.
+ *
+ * Specifically, the cubic function
+ * \f[
+   a x^3 + b x^2 + c x + d
+ * \f]
+ * where a is assumed to already be 1, and is not provided to the
+ * function.
+ * Uses the Numerical Recipes cubic algorithm, a combination of Cardano and
```
elliottbiondo: Add citation

test/orange/surf/detail/QuarticSolver.test.cc
```diff
@@ -0,0 +1,189 @@
...
+//---------------------------------------------------------------------------//
+
+using QuarticSolvers = ::testing::Types<FerrariSolver>;
+TYPED_TEST_SUITE(QuarticSolverTest, QuarticSolvers, );
+
+/*
+ * Test cases with all non-zero roots, i.e., the ray does not start on or close
+ * to the surface
+ */
+
```
elliottbiondo:
```suggestion
using QuarticSolvers = ::testing::Types<FerrariSolver>;
TYPED_TEST_SUITE(QuarticSolverTest, QuarticSolvers, );

//---------------------------------------------------------------------------//
/*
 * Test cases with all non-zero roots, i.e., the ray does not start on or close
 * to the surface
 */
```

elliottbiondo (dismissed):
I think this is basically ready to merge: just a few minor comments

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+ * Allows user to specify if the operation takes place on the surface, in which
+ * case it will solve as a cubic to avoid the root at 0.
+ */
+CELER_FUNCTION auto
+FerrariSolver::operator()(Real5 const& abcde, SurfaceState on_surface) const
+    -> Intersections
+{
+    // Normalize coefficients
+    auto [a, b, c, d, e] = abcde;
+    real_type ba = b / a, ca = c / a, da = d / a, ea = e / a;
```
sethrj: Add a `CELER_EXPECT(abcde[0] != 0);` at the top of the function: in the degenerate case you're expecting `e=0` not `a=0`. Also I'm concerned about how this normalization (which is new I guess?) affects your soft_zero tests. For a circular torus, $`a = R^{-4} `$ right? Our absolute tolerance has units of length, not `[length^4]`.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+                            no_intersection(),
+                            no_intersection());
+
+        int idx = 0;
+        idx = place_root(roots, z0, idx);
+        idx = place_root(roots, z1, idx);
+        idx = place_root(roots, z2, idx);
+        sort(&roots[0], &roots[idx]);
+        return roots;
+    }
```
sethrj: This "special case" should live in the special case operator `operator()(Real4 const& abcd)`. The general case should  dispatch to the specific, rather than the specific generalizing the problem and then simplifying again.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+                            no_intersection(),
+                            no_intersection(),
+                            no_intersection());
+        int idx = 0;
+        idx = place_root(roots, r0 - qb, idx);
+        idx = place_root(roots, r1 - qb, idx);
+        idx = place_root(roots, r2 - qb, idx);
+        idx = place_root(roots, r3 - qb, idx);
+
+        sort(&roots[0], &roots[idx]);
```
sethrj: In a subsequent PR let's revisit the "no intersection" in ORANGE; I think it might make sense to treat any nonpositive value as "no intersection", and we'll be able to avoid all these constants and many other tests against infinity. We'd define a helper functor for sorting.

And in practice, sorting isn't even needed! We either do a "sort" externally in the complex intersect case, or a "min element" in the simple intersect case. **So I would for this PR remove the `sort`** (but leave as a postprocessing step in your unit test to reduce the amount of changes you have to make)

Additionally, during typical transport, we've already calculated a "maximum distance" the particle can travel (due to a future collision site, or step limiters, or field propagation) and so we already ignore *all intersections beyond that value*: see `detail::CalcIntersections`. So *if we were* to consider negatives as "invalid", then you could literally just return `{r0 - qb, r1 -qb, ...}`.

src/orange/surf/detail/FerrariSolver.hh
```diff
@@ -0,0 +1,396 @@
...
+ *\f]
+ * Where the given array abcde corresponds to {a, b, c, d, e}.
+ * Replaces negative or complex roots with sentry value no_intersection().
+ *
+ * Allows user to specify if the operation takes place on the surface, in which
+ * case it will solve as a cubic to avoid the root at 0.
+ */
+CELER_FUNCTION auto
+FerrariSolver::operator()(Real5 const& abcde, SurfaceState on_surface) const
+    -> Intersections
```
sethrj: Based on your test cases and some of the earlier discussion, I think you should just eliminate the `on_surface` input entirely and rely on the caller managing conversion between "are we on a surface" and "solve a fourth or third order equation".

sethrj (commented):
Looks great Owen! Just a few final comments based on the changes.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Actually in our specific case (the formula we chose for toroids) it's redundant anyways because a already starts at 1:
image
Although I have been meaning to do a more thorough review of different people's ray-toroid polynomials at some point, given that we know it can be an issue.

osanstrong: FWIW the normalization has always been there.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Re: sorting roots, Sure, it's good to know that the sorting isn't needed. Don't worry, I would have sorted for the unit tests at some point anyways, over trying to guarantee the specific order it spits out four different roots in.
Re: negative roots, Do any of our trackers reject negative roots, or is this more of a hypothetical? When I checked through (though I admit I don't know celeritas enough to be certain), I only saw the "maximum distance" rejection you were talking about.

Removing the sorting for now, but I'm gonna leave `place_root` as it is for the moment until I confirm whether or not it still needs to check that roots are positive.

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Fair enough, that pattern was only following how `QuadraticSolver` handles it; i.e. in `GeneralQuadric`:
```
/*!
 * Calculate all possible straight-line intersections with this surface.
 */
CELER_FUNCTION auto
GeneralQuadric::calc_intersections(Real3 const& pos,
                                   Real3 const& dir,
                                   SurfaceState on_surface) const
    -> Intersections
{
    real_type const x = pos[0];
    real_type const y = pos[1];
    real_type const z = pos[2];
    real_type const u = dir[0];
    real_type const v = dir[1];
    real_type const w = dir[2];

    // Quadratic values
    real_type a = (a_ * u + d_ * v) * u + (b_ * v + e_ * w) * v
                  + (c_ * w + f_ * u) * w;
    real_type b = (2 * a_ * x + d_ * y + f_ * z + g_) * u
                  + (2 * b_ * y + d_ * x + e_ * z + h_) * v
                  + (2 * c_ * z + e_ * y + f_ * x + i_) * w;
    real_type c = ((a_ * x + d_ * y + g_) * x + (b_ * y + e_ * z + h_) * y
                   + (c_ * z + f_ * x + i_) * z + j_);

    return detail::QuadraticSolver::solve_general(a, b / 2, c, on_surface);
}
```

src/orange/surf/detail/FerrariSolver.hh
osanstrong: Right now I only have the placeholder `\citet{}`, let me know when the Numerical Recipes entry is in and I have a key I can put there.

src/orange/surf/detail/FerrariSolver.hh
sethrj: Yes, exactly, replacing "no intersection" with "is positive" is future work. Thanks!

src/orange/surf/detail/FerrariSolver.hh
sethrj: Please use `\citet{press-nr-1992, https://dl.acm.org/doi/10.5555/1403886}` which is the C version of the book. I can send you a PDF of the relevant content if you want, but it looks pretty much the same (minus the commentary about FORTRAN missing some functions.)

sethrj:
@elliottbiondo I think this is ready to merge!

sethrj:
@osanstrong @elliottbiondo Since we've made the solver more generic, I've moved it to its rightful place in `corecel/math`. I also made a couple of small adjustments to the documentation so that the key capabilities appear in the user docs. Thanks again!

sethrj:
@elliottbiondo @osanstrong Please take a quick look at my changes and then we can merge:

https://github.com/celeritas-project/celeritas/pull/2010/files/e487c26f380d3e4d1a79bc285bc352a1a8225263..58bf4c523865dfa70f1a2a6e5b4f403ac51d9cbe

elliottbiondo (approved):
These changes look good to me. Let's wait for @osanstrong to take a look before merging.

sethrj:
@elliottbiondo I'm going to merge this unless @osanstrong says otherwise. Thanks guys! Next step, basic torus support?

osanstrong:
Looks good to me! (sorry for not getting back to you faster) Only question I had atp was if you think it's generic enough to be in corecel, should it also return negative roots and let the torus code ignore them? Thanks for the last edits!

sethrj:
I thought about that, and if we ever need nonpositive roots, we can easily make it more generic by constructing with a "filter" template argument.

PR #2011, opened 2025-10-06
Add profiling capabilities to `celer-geo`

sethrj:
- Adds profiling session to `celer-geo` and an option to pass a perfetto output filename
- Export the environment and version from `celer-geo`
- Add a few carefully chosen scoped profiling for geometry conversion and setup


esseivaju (approved):
LGTM

PR #2013, opened 2025-10-07
Code coverage

pcanal:
This add gathering of code line coverage by the test run during the CI.

Currently 2 configuration have been added (one Spack and one Docker) for code coverage:

Docker: code-coverage-vecgeom-reldeb-rocky-cuda
Spack: vecgeom-code-coverage-g4@11.3

A preview of the result can be seen at: https://app.codecov.io/github/pcanal/celeritas/tree/code_coverage 
Actual result will be at: https://app.codecov.io/github/celeritas-project/celeritas/tree/code_coverage


scripts/gcov_for_clang.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+if [ -z ${LLVM_COV+x} ];
```
sethrj: What's `+x`? And isn't `-z "${LLVM_COV}"` better (and still POSIX compliant)?

scripts/gcov_for_clang.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
sethrj: Is this a file you wrote yourself? If so it should have the celeritas copyright at the top (I should've put those for the workflows as well). Also is it bash only or strict POSIX (preferable)?

CMakeLists.txt
```diff
@@ -215,6 +218,18 @@ if(CMAKE_VERSION VERSION_LESS 3.27 AND CELERITAS_USE_covfie)
...
+# Enable test coverage
+#----------------------------------------------------------------------------#
+if(CELERITAS_CODE_COVERAGE)
+  set(GCC_COVERAGE_COMPILE_FLAGS "-fprofile-arcs -ftest-coverage")
+  set(GCC_COVERAGE_LINK_FLAGS    "-fprofile-arcs")
+  set(CMAKE_CXX_FLAGS            "${CMAKE_CXX_FLAGS} ${GCC_COVERAGE_COMPILE_FLAGS}")
+  set(CMAKE_EXE_LINKER_FLAGS     "${CMAKE_EXE_LINKER_FLAGS} ${GCC_COVERAGE_LINK_FLAGS}")
+  set(CMAKE_SHARED_LINKER_FLAGS  "${CMAKE_SHARED_LINKER_FLAGS} ${GCC_COVERAGE_LINK_FLAGS}")
+  set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "${CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS} ${GCC_COVERAGE_COMPILE_FLAGS}")
+endif()
```
sethrj: I don't like having custom compiler-dependent flags tied to a specific option in our main cmake, *especially* since it's CI-specific code (no user will ever run it). What about adding the flags to CMakePresets? Or if the "multiple link flags get X, multiple compile flags get Y" paradigm is useful, perhaps defining string `CELERITAS_LINKER_FLAGS` and `CELERITAS_CXX_FLAGS` that get added globally? And even this is vaguely horrifying to me because since they're directory properties, everything *underneath* celeritas (including, e.g. CLI11) is also getting these flags. So instead add those flags to the `celeritas::BuildFlags` target in `src/CMakeLists.txt`?

.github/workflows/build-spack.yml
```diff
@@ -172,6 +182,12 @@ jobs:
           echo "${SPACK_VIEW}/bin" >> $GITHUB_PATH
           echo "CMAKE_PREFIX_PATH=${SPACK_VIEW}:${CMAKE_PREFIX_PATH}" >> $GITHUB_ENV
           spack env activate . --sh > "${SPACK_VIEW}/rc"
+      - name: Install cov packages
+        if: ${{matrix.special == 'code-coverage'}}
+        run: |
+          sudo apt-get -q -y update
+          sudo apt-get -q -y install lcov
+          sudo pip3 install gcovr
```
sethrj: Can/should we add these to the spack environment instead of a separate dependency? Also since these dependencies are only needed by the `generate-coverage` action, why isn't the dependency setup inside of that action?

.github/workflows/build-docker.yml
```diff
@@ -99,6 +114,14 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+      - name: Generate Coverage Report and Upload to Codecov
+        uses: ./.github/actions/generate-coverage
+        if: ${{ matrix.special == 'code-coverage' && !cancelled() && steps.test.outcome == 'success' }}
+        with:
+          core_count: '4'
```
sethrj: We use `$(nproc)` in a couple of places. Is the core count guaranteed to be 4? Should that be a workflow option/variable/etc?

scripts/gcov_for_clang.sh
```diff
@@ -0,0 +1,39 @@
...
+  if command -v llvm-cov >/dev/null 2>&1; then
+    LLVM_COV=$(command -v llvm-cov)
+  else
+    if [ -z ${CXX+x} ];
+    then
+      CXX=clang++
+    fi
+    CLANG_MAJOR=$(${CXX} --version | grep -oE 'clang version [0-9]+' | grep -oE '[0-9]+')
+    LLVM_COV=llvm-cov-${CLANG_MAJOR}
+  fi
```
sethrj, line 39: Are all these necessary if we're just running them from the CI?

.github/actions/generate-coverage/action.yml
```diff
@@ -0,0 +1,56 @@
+name: 'Generate Coverage Report'
```
sethrj, line 1: Did you create this action yourself or derive it from another project?

sethrj (commented):
Thanks @pcanal ! A couple of questions:

1. Do you envision anyone aside from the CI ever using this capability? If not, you can probably eliminate 90% of the wrapper scripts and possibly most of the action as well.
2. Does the coverage not show header files? Browsing the source tree I see quite a few missing...
3. Since coverage can't actually run any of the CUDA tests, perhaps it's undesirable to have it enabled for the docker+cuda case?
4. From what I can guess from the highlighted "not tested" lines, it doesn't like the fact that most of our assertion checking (`CELER_EXPECT` etc) doesn't hit the exceptional cases. Should we set `CELERITAS_DEBUG=OFF` in the code coverage cmake presets?
5. I don't understand some of the negative coverage results like src/orange/univ/detail/LazySenseCalculator.hh , which is definitely tested...

Overall this is a really nice capability that'll help us find dead code. Much appreciated!

.github/actions/generate-coverage/action.yml
pcanal: The core (line 35-47) have been tested with the ROOT CI.  The rest is a combination of hand and Claude-helped refactorings.  Support for clang build was annoying and required input from multiple sources.

.github/workflows/build-docker.yml
pcanal: I can't find the official documentation for `$(nproc)`  ...

.github/workflows/build-spack.yml
pcanal: > Can/should we add these to the spack environment instead of a separate dependency? 

Yes.  This steps should probably be part of the Docker image creation and the Spack environment (although this is really only needed for the upload part so hopefully we can do it without really affecting interactive use).  

I propose to do these as part of a distinct PR.

scripts/gcov_for_clang.sh
pcanal: I wrote the first and last line.  "Claude" provided the rest.   I am not sure if it is POSIX compliant.

scripts/gcov_for_clang.sh
pcanal: The difference is that `-z` only detect that the expansion is empty while `-z..+x` detect that the variable is not set.

See the first answer to https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash?page=1&tab=scoredesc#tab-top
```
if [ -z ${var+x} ]; then echo "var is unset"; else echo "var is set to '$var'"; fi

where ${var+x} is a [parameter expansion](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) which evaluates to nothing if var is unset, and substitutes the string x otherwise.
```

scripts/gcov_for_clang.sh
pcanal: Necessary, probably not (in the CI `LLVM_COV` would always be set).  But helpful for debugging (if the version of `llvm` is updated but `LLVM_COV` is forgotten to be updated).

It could also be used locally to generate the coverage report (a human readable xml file) locally.

CMakeLists.txt
pcanal: > I don't like having custom compiler-dependent flags tied to a specific option in our main cmake,

Fair enough.

> especially since it's CI-specific code (no user will ever run it)

User indeed no.  Developer might run it locally in some (rare) cases.

.github/workflows/build-docker.yml
sethrj: https://man7.org/linux/man-pages/man1/nproc.1.html , but  I've just added it as a repository-level variable `CELER_GHA_NUM_PROCS=4` that you should be able to access with `${{vars.CELER_GHA_NUM_PROCS}}`

CMakeLists.txt
pcanal: > everything underneath celeritas (including, e.g. CLI11) is also getting these flags.

Indeed that is not the intent (although it probably does not hurt much).

pcanal: > So instead add those flags to the celeritas::BuildFlags target in src/CMakeLists.txt?

This is likely appropriate (there are other compiler specific flags there) as long as those flags are then applied to all (Celeritas) libraries and all the test and users apps (that are run as part of the ~CI~ test suite).

.github/workflows/build-docker.yml
pcanal: '4' is my case was an arbitrary number ( `> 1 and < 'likely' number of core' ) and by making it a parameters, I was intending to make it customizable (either when we have knowledge of the number of cores programatically or not).

pcanal: > [man7.org/linux/man-pages/man1/nproc.1.html](https://man7.org/linux/man-pages/man1/nproc.1.html)

I would guess that it is only usable within the part of the action that is run within a shell, isn't it?

pcanal: which actually is good enough here.

pcanal:
>     Do you envision anyone aside from the CI ever using this capability? If not, you can probably eliminate 90% of the wrapper scripts and possibly most of the action as well.

If we were using it only from the CI, I can see the wrapper script returning to be trivial.  I don't see any real simplification of the action.

I do envision some developers running it locally.  For example to debug why the developer understanding of the code coverage differs from the reported code coverage.

>    Does the coverage not show header files? Browsing the source tree I see quite a few missing...

It should show both header and source files.  If source files are not shown they are either 'not run at all' *or* not compile with the right flags *or* somehow filter out by the reporter and/or uploader. 

>    Since coverage can't actually run any of the CUDA tests, perhaps it's undesirable to have it enabled for the docker+cuda case?

That a fair point. I'll change that.

>     From what I can guess from the highlighted "not tested" lines, it doesn't like the fact that most of our assertion checking (CELER_EXPECT etc) doesn't hit the exceptional cases. Should we set CELERITAS_DEBUG=OFF in the code coverage cmake presets?

Indeed the exceptional case/branches that we are not testing, would be reported as not tested ( and/or we might be hitting bug in code coverage and/or gcov).   For lines of code that we known we can't and do not want to test there are annotation to tell it to ignore them.
```
// GCOVR_EXCL_START
...
// GCOVR_EXCL_STOP
```

>    I don't understand some of the negative coverage results like src/orange/univ/detail/LazySenseCalculator.hh , which is definitely tested...

I can investigate.  But first question, should it hold off the PR or be done later?

> Overall this is a really nice capability that'll help us find dead code. Much appreciated!

Thanks :)

.github/workflows/build-docker.yml
pcanal: For now, moved to use `nproc`

scripts/gcov_for_clang.sh
sethrj: But it is a bash extension

pcanal: Which should be fine as long as `bash` is installed on the system, shouldn't it be?

sethrj: Hopefully? I just know the macos version of bash has been stuck at 3.x forever since 4.x is GNU and not BSD-compatible. It's possible it'll be removed someday, and it's easy enough to just use the standards compliant version. The (rather opinionated) stack overflow link primarily is saying this lets you distinguish between 'empty' and 'not set'. Shell scripting is confusing enough that I think they *should* be treated the same way. 

In sum:
- `${foo+x}` is a bash extension
- it uses a meaningless string `x` as a cheat that can be misread (as I did) as a flag with significance (we could do `${var+pcanal}` the same result)
- if I had `LLVM_COV` set in my environment for some reason it and wanted to temporarily disable it when running the script I would use `LLVM_COV= ./gcov_for_clang.sh`, but that defines an empty variable rather than unset

Are these a strong enough argument for just using `-z "${foo}"`?

scripts/gcov_for_clang.sh
sethrj: It just seems fragile and edge-case enough that it would be better to emit an error and tell the user to set LLVM_COV themselves. Did you test each of these `else` cases?

scripts/gcov_for_clang.sh
```diff
@@ -0,0 +1,73 @@
...
+      done
+    fi
+  done
+  if [ "$found_variants" = false ]; then
+    echo "  No llvm-cov variants found" >&2
+  fi
+  exit 1
+fi
+
+exec ${LLVM_COV}  gcov "$@"
```
sethrj:
```suggestion
exec "${LLVM_COV}" gcov "$@"
```
in case the user has llvm-cov in `/Volumes/My Stuff/bin/llvm-cov` 🙃

scripts/gcov_for_clang.sh
```diff
@@ -0,0 +1,73 @@
...
+          found_variants=true
+        fi
+      done
+    fi
+  done
+  if [ "$found_variants" = false ]; then
+    echo "  No llvm-cov variants found" >&2
+  fi
+  exit 1
+fi
```
sethrj: This is seriously too much for a wrapper script. We can't do this for every command that might have an alternate name. Delete.

.github/workflows/build-spack.yml
sethrj: Fair enough. Can you move the step to `Generate Coverage Report` since it's not needed till then? (or even better would be if this was part of the generate-coverage action...)

scripts/gcov_for_clang.sh
sethrj: Gotcha. I mostly wanted to ensure we had the necessary attribution/copyright.

src/CMakeLists.txt
```diff
@@ -50,6 +50,20 @@ celeritas_target_include_directories(BuildFlags
...
+#----------------------------------------------------------------------------#
+if(CELERITAS_CODE_COVERAGE)
+  set(GCC_COVERAGE_COMPILE_FLAGS "-fprofile-arcs;-ftest-coverage")
+  set(GCC_COVERAGE_LINK_FLAGS    "-fprofile-arcs")
+  target_compile_options(BuildFlags
+    INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:${GCC_COVERAGE_COMPILE_FLAGS}>"
+  )
+  target_link_options(BuildFlags
+    INTERFACE "${GCC_COVERAGE_LINK_FLAGS}"
+  )
```
sethrj: Is this GCC specific? Does it wrk for LLVM too? Shouldn't it be `CELERITAS_GCC_CODE_COVERAGE` if it's compiler specific?

Couldn't we just have:
```suggestion
if(DEFINED CELERITAS_COMPILE_FLAGS)
  target_compile_options(BuildFlags
    INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:${CELERITAS_COMPILE_FLAGS}>"
  )
endif()
if(DEFINED CELERITAS_LINK_FLAGS)
  target_link_options(BuildFlags
    INTERFACE "${CELERITAS_LINK_FLAGS}"
  )
endif()
```

and add `{"CELERITAS_COMPILE_FLAGS": "-fprofile-arcs;-ftest-coverage"},
{"CELERITAS_LINK_FLAGS": "-fprofile-arcs"}` to the cmake presets for the jobs that run it?

These just seem too specific to work in all cases for all compilers (past and future).

.github/workflows/build-spack.yml
pcanal: I can move them (athough they are meant to disappear).  I will keep them in their current file as they are platform dependent.

scripts/gcov_for_clang.sh
pcanal: Finding the name/location of llvm-cov was annoying enough that I felt the need to add this.  That said, I don't feel strongly about it and more importantly I don't really plan to routinely/on-going testing it, so I will remove/simplify it.

scripts/gcov_for_clang.sh
pcanal: Fair enough.

src/CMakeLists.txt
pcanal: > Is this GCC specific? Does it wrk for LLVM too?

I think it originated with GCC but is indeed working with Clang.

pcanal: >  to the cmake presets for the jobs that run it?

This is a philosophical question.  Should coverage be enable only for the CI (and/or subset of preset) or should you be able to run it locally?  You seem lean for the former, I strongly recommend/want the later.

sethrj (approved):
Thanks again @pcanal !

Issue #1945, opened 2025-09-08
Geant4 OpNovice Integration


Rashika-Gupta:
This issue tracks the integration of  Geant4 [OpNovice example](https://github.com/Geant4/geant4/tree/a499fb82e9d3310f83c34f52806127561bdaab40/examples/extended/optical/OpNovice) with Celeritas to validate and support the optical physics processes.

- [x] Updated optical material validation to permit constant or non-decreasing refractive index values (#1904)
- [x] Enabled optical photons and processes in `SetupOptions`, added optical capacity, and handled Rayleigh scattering for water or materials without MFP tables by defaulting to infinity (#1913)
- [x] Skip importing WLS optical model altogether when there are no model properties are present. - (#1963 )

- [x] Mie scattering in celertias #1933 
- [x] Add and validate OpNovice example using Celeritas test harness #2016 
- [x] Demonstrate Celeritas integration into G4 OpNovice example
  - [ ] Step count comparison
  - [x] Visualization of tracks: celeritas vs g4

sethrj:
We showed today we can export a GDML with all the optical properties, so let's also add as a unit test in `accel`
- Copy the output to `test/geocel/data/op-novice.gdml`, add the disclaimer from `solids.gdml` since it's exported from Geant4
- Add a new `OpNoviceIntegrationMixin` in `test/accel/IntegrationTestBase.{hh,cc}` that enables optical physics
- Add a new `OpNovice` test that derives from it in `TrackingMangerIntegration.test.cc`
- Set up the test torun (like is done with LarSphereOptical) using the CMakeLists

sethrj:
@Rashika-Gupta I forgot that for achievements like this I'd like to have some visuals to communicate the milestone: once #1984 is merged and we can actually get refraction through the water, could you and @stognini generate a visualization of tracks in the model?

sethrj:
@Rashika-Gupta and @stognini hacked in a track step output and gave us some visuals:

<img width="1405" height="1257" alt="Image" src="https://github.com/user-attachments/assets/25d2500b-8ca4-435c-966c-ef94b97df6e5" />

PR #2016, opened 2025-10-07
Add OpNovice GDML optical geometry and integration test

Rashika-Gupta:
This PR adds a new integration test for the `OpNovice` geometry with full optical properties exported from Geant4 as GDML implementing the following #1945 in the following way : 
1. Added `op-novice.gdml`.
2. Implemented `OpNoviceIntegrationMixin` to test optical physics - photon generation.
3. Added a new `OpNovice` test in `TrackingManagerIntegration`.
4. Updated `CMakeLists.txt` to build and run the new test.


test/accel/IntegrationTestBase.cc
```diff
@@ -497,6 +497,50 @@ auto TestEm3IntegrationMixin::make_sens_det(std::string const& sd_name)
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Create physics list: default is EM only using make_physics_input.
+ */
+auto OpNoviceIntegrationMixin::make_physics_input() const -> PhysicsInput
+{
+    PhysicsInput result = Base::make_physics_input();
+    result.em_bins_per_decade = 5;
+    return result;
```
sethrj, line 517: I think we want OpNovice to *always* have optical physics, so you'll need to copy/adapt the code from `LarSphereOptical::make_physics_input` (from `TrackingManagerIntegration`) here and in `make_setup_options`.

test/accel/CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ foreach(_basename IN ITEMS UserActionIntegration TrackingManagerIntegration)
     OFFLOAD cpu gpu
     RMTYPE serial
   )
+  # Test OpNovice
+  celeritas_add_integration_tests(${_basename} OpNovice run)
```
sethrj: This line adds a bunch of different combinations, which is useful for testing primary functionality (GPU compatibility, thread safety) but not necessary for running the same type of problem that's already got a test. Replace with:

```suggestion
  # Test OpNovice
  celeritas_add_integration_tests(
    ${_basename} OpNoviceOptical run
    OFFLOAD cpu gpu g4
    RMTYPE ${_optical_rm_type}
  )
```
after moving the `_optical_rm_type` code above this loop.

test/accel/CMakeLists.txt
```diff
@@ -100,5 +107,11 @@ celeritas_add_integration_tests(
   OFFLOAD cpu gpu g4
   RMTYPE ${_optical_rm_type}
 )
+celeritas_add_integration_tests(
+  TrackingManagerIntegration OpNoviceOptical run
+  OFFLOAD cpu gpu g4
+  RMTYPE ${_optical_rm_type}
+)
```
sethrj: Since the `OpNoviceOptical.run` tests are being added in the loop above, this is redundant.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -409,6 +415,273 @@ TEST_F(LarSphereOptical, run)
...
+            {
+                G4PrimaryVertex* vtx = event->GetPrimaryVertex(i);
+                for (auto j : range(vtx->GetNumberOfParticle()))
+                {
+                    G4PrimaryParticle* p = vtx->GetPrimary(j);
+                    p->SetWeight(10.0);
+                }
+            }
+        }
+    }
```
sethrj: There's no need to adjust the weight of the particles; this was done once to test that weight gets correctly sent to the hit objects.

sethrj (commented):
I hope I'm not reviewing this too early (I saw the review request but didn't notice it's still draft). It looks like there's a lot of unnecessarily duplicated code in the OpNovice test. Keep in mind:
1. We only want to test the problem with optical physics (unlike the LarSphere problem which I used to test some detector mechanics).
2. Eliminate tests/code that are the same as the other tests. For example, we already tested that the weights are being set correctly for EM particles, so there's no need to change the primary weights. (We *might* want to do something similar if we wanted to test that the weights are getting sent from the EM loop to the optical loop.) We don't need to test the UI commands unless it's something new.
3. Remember the primary purpose of the OpNovice test is to make sure that the physics is being set up and built correctly, so we need to find a way to check the constructed Celeritas physics and setup, and make sure that optical photons are being run.

Rashika-Gupta:
Addressed the comments. 
Also, increased the number of events to reproduce the failure at
`precondition failed: wavelength > 0`, which matches the earlier error I encountered.
```cpp
G4WT1 > [2/2]/forked/celeritas/src/accel/LocalTransporter.cc:386: Executing the first Celeritas stepping loop
/forked/celeritas/test/accel/IntegrationTestBase.cc:123: Failure
Failed
GeantExceptionHandler caught runtime error (1,celer.event.flush): from forked/celeritas/src/celeritas/optical/detail/OpticalUtils.hh:27: precondition failed: wavelength > 0
```

Rashika-Gupta:
> Addressed the comments. Also, increased the number of events to reproduce the failure at `precondition failed: wavelength > 0`, which matches the earlier error I encountered.
> 
> ```c++
> G4WT1 > [2/2]/forked/celeritas/src/accel/LocalTransporter.cc:386: Executing the first Celeritas stepping loop
> /forked/celeritas/test/accel/IntegrationTestBase.cc:123: Failure
> Failed
> GeantExceptionHandler caught runtime error (1,celer.event.flush): from forked/celeritas/src/celeritas/optical/detail/OpticalUtils.hh:27: precondition failed: wavelength > 0
> ```

This  occurs in  `wavelength_to_energy()` called from [`ScintillationGenerator.hh`](https://github.com/celeritas-project/celeritas/blob/b28f327bc9fb76c48250d13a2e9337ad641575e9/src/celeritas/optical/gen/ScintillationGenerator.hh#L162C3-L167C71)
```cpp
 sample_lambda_
        = NormalDistribution{component.lambda_mean, component.lambda_sigma};
    ExponentialDist sample_time(real_type{1} / component.fall_time);

    TrackInitializer photon;
    photon.energy = detail::wavelength_to_energy(sample_lambda_(rng));

```
For the specific instance
`(celeritas::NormalDistribution<>::result_type) wavelength = -3.8856919072923377E-9` 
`(celeritas::real_type) component.lambda_mean = 3.0849999999999999E-7` and `(celeritas::real_type) component.lambda_sigma = 8.9263721210270798E-8`

A possible fix could be 
```cpp
auto wavelength = sample_lambda_(rng);
if (wavelength <= 0) {
    CELER_LOG(warning) << "Invalid wavelength " << wavelength
                             << " — clamping to mean";
    wavelength = component.lambda_mean;
}
```

sethrj:
Thanks @Rashika-Gupta . Instead of warning and replacing with mean (i.e., lumping all of the PDF that's <= 0 into a delta function) I think it'd be better to resample so that we're truncating and rescaling the normal PDF:
```c++
real_type wavelength;
do {
wavelength = sample_lambda_(rng);
} while (wavelength <= 0);
```

Of course, the *correct* solution is to sample from the user PDF not the gaussian....

Rashika-Gupta:
> Thanks @Rashika-Gupta . Instead of warning and replacing with mean (i.e., lumping all of the PDF that's <= 0 into a delta function) I think it'd be better to resample so that we're truncating and rescaling the normal PDF:
> 
> ```c++
> real_type wavelength;
> do {
> wavelength = sample_lambda_(rng);
> } while (wavelength <= 0);
> ```
> 
> Of course, the _correct_ solution is to sample from the user PDF not the gaussian....

yes, would have preferred doing this but, did not want to mess with the existing code-base.

sethrj:
"Messing with" (i.e. "fixing") the code base is what these integration-level tests are supposed to drive us toward 😄 The code that existed before you isn't sacred, so it's always OK to bring up changes you think may be needed to get a problem to run.

src/celeritas/optical/gen/ScintillationGenerator.hh
```diff
@@ -161,10 +161,16 @@ CELER_FUNCTION TrackInitializer ScintillationGenerator::operator()(Generator& rn
     // "spare" value that the wavelength sampler might have stored
     sample_lambda_
         = NormalDistribution{component.lambda_mean, component.lambda_sigma};
+    real_type wavelength;
+    do
+    {
+        wavelength = sample_lambda_(rng);
+    } while (CELER_UNLIKELY(wavelength <= 0));
```
sethrj, line 170: It would be good to add a comment why this do loop is here (rejecting in the case of very large sigma and/or very small lambda)... and also to add an assertion after we get the component that `component.lambda_mean` is positive

test/accel/UserActionIntegration.test.cc
```diff
@@ -107,6 +107,26 @@ TEST_F(TestEm3, run)
...
+    UAI::Instance().SetOptions(this->make_setup_options());
+
+    cout << "initializing" << endl;
+    rm.Initialize();
+    cout << "beam on" << endl;
+
+    rm.BeamOn(3);
+    cout << "initial run done" << endl;
+    rm.BeamOn(1);
+    cout << "second run done" << endl;
```
sethrj: Most of our tests should probably have just a single run to make things quick: we want to make sure at least one of the tests has multiple runs, but I think that's covered.

test/accel/IntegrationTestBase.cc
```diff
@@ -497,6 +497,76 @@ auto TestEm3IntegrationMixin::make_sens_det(std::string const& sd_name)
...
+    EXPECT_TRUE(optical);
+    EXPECT_TRUE(optical.scintillation);
+    EXPECT_TRUE(optical.cherenkov);
+    EXPECT_TRUE(optical.mie_scattering);
+    EXPECT_TRUE(optical.rayleigh_scattering);
+
+    // Disable WLS (reemission not yet supported)
+    using WLSO = WavelengthShiftingOptions;
+    optical.wavelength_shifting = WLSO::deactivated();
+    optical.wavelength_shifting2 = WLSO::deactivated();
```
sethrj: Do we need to disable WLS? I don't see any related data in the GDML input, so shouldn't Celeritas just ignore the models?

sethrj (approved):
Great work!

test/accel/IntegrationTestBase.cc
amandalund: Probably not, though doesn't hurt to be explicit. If we used a GDML input with the data, would that enable it even though it's not fully supported?

sethrj: Leaving it enabled is closer to the behavior of a user standalone application. @Rashika-Gupta added code that *should* disable the corresponding Celeritas optical physics if no materials have MFPs for it, so leaving it enabled is also a good check of that I think.

test/accel/UserActionIntegration.test.cc
Rashika-Gupta: Originally, when the number of events was set to 1 or 2, the issue of generating photons with wavelength < 0 didn’t appear, likely due to limited statistics. Since that problem is now resolved, it should be fine to reduce the number of events again.

test/accel/IntegrationTestBase.cc
Rashika-Gupta: Yes, I agree with you Seth that section of code (commit fb1fcf529 from PR #1933 ) was added specifically to handle cases where WLS MFPs were being created unintentionally in OpNovice example.

amandalund: That's fair, so the OpNovice example creates the WLS process but then doesn't have any data for it?  I guess unlike other physics Geant4 might construct all the optical processes but then only "use" the ones that there's data for?

sethrj: That is my understanding yes. I think Geant4's optical physics constructor builds all the processes but they're just no-ops for the most part.

PR #2017, opened 2025-10-07
Add JLSE env file

elliottbiondo:
This MR adds an env file for JLSE, leveraging a spack install on the Celeritas project space. The env file spactivates environments for H100 and MI250X nodes according to the host name. I have verified I can build and run celer-sim on both node types.

### DIFF
--- a/scripts/env/jlse.sh
+++ b/scripts/env/jlse.sh
@@ -5,17 +5,20 @@ export PROJ=/vast/projects/celeritas
 export SPACKROOT=$PROJ/spack
 . $SPACKROOT/share/spack/setup-env.sh
 
-qsub-gpu() {
-    # Usage: qsub-gpu [mi300x|h100] <qsub args>
-    [[ $(hostname) =~ ^jlselogin[0-9]+ ]] || {
-        echo "Error: must be on a jlselogin* host to submit jobs."
-        return 1
-    }
+qsub_gpu() {
+    # Usage: qsub_gpu [mi300x|h100] <qsub args>
 
-    case "$1" in
-        mi300x) queue="gpu_amd_mi300x" ;;
-        h100)   queue="gpu_h100" ;;
-        *)      echo "Error: unknown GPU type '$1'"; return 1 ;;
+    # Verify we are on a login node
+    h=$(uname -n 2>/dev/null || hostname 2>/dev/null || printf '')
+    if ! expr "$h" : 'jlselogin[0-9][0-9]*$' >/dev/null 2>&1; then
+        printf '%s\n' 'Error: must be on a login node to submit jobs.' >&2
+        return 1
+    fi
+
+    case ${1-} in
+        mi300x) queue=gpu_amd_mi300x ;;
+        h100)   queue=gpu_h100 ;;
+        *)      printf "Error: unknown GPU type '%s'\n" "${1-}" >&2; return 1 ;;
     esac
 
     shift
--- a/scripts/env/jlse.sh
+++ b/scripts/env/jlse.sh
@@ -2,14 +2,25 @@
 
 module use /soft/modulefiles
 export PROJ=/vast/projects/celeritas
-export SPACKROOT=/vast/projects/celeritas/spack
+export SPACKROOT=$PROJ/spack
 . $SPACKROOT/share/spack/setup-env.sh
 
-# Interactive job on an MI30
-alias mi300x="qsub -I -t 60 -n 1 -q gpu_amd_mi300x"
+qsub-gpu() {
+    # Usage: qsub-gpu [mi300x|h100] <qsub args>
+    [[ $(hostname) =~ ^jlselogin[0-9]+ ]] || {
+        echo "Error: must be on a jlselogin* host to submit jobs."
+        return 1
+    }
 
-# Interactive job on an H100
-alias h100="qsub -I -t 60 -n 1 -q gpu_h100"
+    case "$1" in
+        mi300x) queue="gpu_amd_mi300x" ;;
+        h100)   queue="gpu_h100" ;;
+        *)      echo "Error: unknown GPU type '$1'"; return 1 ;;
+    esac
+
+    shift
+    exec qsub -q "$queue" "$@"
+}
 
 # Load Nvidia/AMD environment if on a compute node
 if [[ "$(hostname -s)" == hopper* ]]; then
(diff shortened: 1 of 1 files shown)


elliottbiondo:
@sethrj small one RTR

scripts/env/jlse.sh
```diff
@@ -0,0 +1,29 @@
...
+module use /soft/modulefiles
+export PROJ=/vast/projects/celeritas
+export SPACKROOT=/vast/projects/celeritas/spack
+. $SPACKROOT/share/spack/setup-env.sh
+
+# Interactive job on an MI30
+alias mi300x="qsub -I -t 60 -n 1 -q gpu_amd_mi300x"
+
+# Interactive job on an H100
+alias h100="qsub -I -t 60 -n 1 -q gpu_h100"
```
sethrj: Maybe write as a function:
```c++
qsub-gpu() {
    # Usage: qsub-gpu [mi300x|h100|...]
    case "$1" in
        mi300x) queue=gpu_amd_$1 ;;
        *) queue=gpu_$1 ;;
    esac
    exec qsub -q "$queue" "$@"
}
```
and you can `alias mi300x=qsub-gpu mi300x -I -t 60 -n 1` in your bashrc? 

It's great to have these GPU names here as documentation but everyone's going to want their own aliases.

Also maybe these should be in an `else` block since the jobs should be submitted only from the host `jlse*` node?

scripts/env/jlse.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh -e
+
+module use /soft/modulefiles
+export PROJ=/vast/projects/celeritas
+export SPACKROOT=/vast/projects/celeritas/spack
```
sethrj:
```suggestion
export SPACKROOT=$PROJ/spack
```

sethrj (approved):
Thanks Elliott for getting this set up!

scripts/env/jlse.sh
elliottbiondo: Yeah these lines were just meant to be an example. I expanded it out to a fully function, though it kinda seems like overkill.

elliottbiondo:
@sethrj RFAL

scripts/env/jlse.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh -e
+
+module use /soft/modulefiles
+export PROJ=/vast/projects/celeritas
+export SPACKROOT=$PROJ/spack
+. $SPACKROOT/share/spack/setup-env.sh
+
+qsub-gpu() {
+    # Usage: qsub-gpu [mi300x|h100] <qsub args>
+    [[ $(hostname) =~ ^jlselogin[0-9]+ ]] || {
```
sethrj: This is no longer POSIX 😢

scripts/env/jlse.sh
```diff
@@ -0,0 +1,40 @@
...
+        h100)   queue="gpu_h100" ;;
+        *)      echo "Error: unknown GPU type '$1'"; return 1 ;;
+    esac
+
+    shift
+    exec qsub -q "$queue" "$@"
+}
+
+# Load Nvidia/AMD environment if on a compute node
+if [[ "$(hostname -s)" == hopper* ]]; then
```
sethrj, line 29: Again, not POSIX. It does indeed matter because I have seen weird systems use zsh/ksh/tsch and fail unexpectedly on these kind of scripts. You can use a `case` statement instead and/or ask copilot/chatgpt to help.

elliottbiondo: Yeah JLSE does allow for zsh and tsch

elliottbiondo:
@sethrj this is now be POSIX. Note that hyphens are not allow in function names per:

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235

scripts/env/jlse.sh
```diff
@@ -5,17 +5,20 @@ export PROJ=/vast/projects/celeritas
 export SPACKROOT=$PROJ/spack
 . $SPACKROOT/share/spack/setup-env.sh
 
-qsub-gpu() {
-    # Usage: qsub-gpu [mi300x|h100] <qsub args>
-    [[ $(hostname) =~ ^jlselogin[0-9]+ ]] || {
-        echo "Error: must be on a jlselogin* host to submit jobs."
+qsub_gpu() {
```
sethrj, line 8: From what I see in my homebrew, dashes are more common than underscores by a factor of 2 ;)

scripts/env/jlse.sh
sethrj: But what I'm saying is that `[[` can cause problems...

sethrj (approved):
Sorry for spending more time on this, I'll just merge as is. If I hit any problems when trying to use it I can just open a follow-on PR.

scripts/env/jlse.sh
elliottbiondo: It's possible that dashes aren't POSIX, but also don't break other shells?

scripts/env/jlse.sh
elliottbiondo: Sorry, this was my oversight. I misread the location of this comment: I though I just needed to update the function 🤦

scripts/env/jlse.sh
sethrj: The dashes are just a stylistic choice, not a standards compliance issue ;) I personally don't like having to hit the shift key mid-command 😛

elliottbiondo: (I don't think you saw the link I posted above)

A `Name` is defined [here](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235
) as 

```
"... a word consisting solely of underscores, digits, and alphabetics from the portable character set."
```

noting that [here](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_05), a function name is:

```
The function is named fname; the application shall ensure that it is a name (see XBD Name)
```

So strictly speaking I don't think hyphens are allowed, but in practice it sounds like it's not an issue.

elliottbiondo: (I accidentally typed "sorry" instead of "so" above before editing which was NOT the tone I was going for :flushed:)

sethrj: oh you're right, I only saw the discussion reply, not the issue comment. Thanks for pointing out the posix compatibility rule!

PR #2018, opened 2025-10-08
Make ORANGE conversion/construction options configurable

sethrj:
This adds a `g4org::Options` struct that contains all the configuration options for converting a Geant4 geometry. It currently is set through unit tests *or* using a JSON file with the `G4ORG_OPTIONS` environment variable.
I intend to use it to gather statistics for the upcoming geant4 collaboration meeting about ORANGE construction, and to better understand how badly the union inlining is slowing down our construction.

### DIFF
--- a/src/orange/orangeinp/UnitProto.hh
+++ b/src/orange/orangeinp/UnitProto.hh
@@ -125,13 +125,6 @@ class UnitProto : public ProtoInterface
         explicit inline operator bool() const;
     };
 
-    //! Optional transformations to apply when building a CsgUnit.
-    enum class Simplification
-    {
-        none,  //!< No simplification
-        infix,  //!< CsgTree suitable for infix logic evaluation
-    };
-
     //! Required input data to create a unit proto
     struct Input
     {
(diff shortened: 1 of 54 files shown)


src/orange/orangeinp/UnitProto.hh
```diff
@@ -125,6 +125,13 @@ class UnitProto : public ProtoInterface
         explicit inline operator bool() const;
     };
 
+    //! Optional transformations to apply when building a CsgUnit.
+    enum class Simplify
```
elliottbiondo: This would be better as a noun: `Simplification`

elliottbiondo (approved):

src/orange/orangeinp/UnitProto.hh
sethrj: Actually `transformation` might be even better since applying DeMorgan's law might increase the number of tokens.

sethrj: CSG transformation?

sethrj: I'm just gonna make it another boolean flag. We can make a `vector<csgtranform>` later if we really need a combination/repeating.

elliottbiondo: The only issue is seeming this "in the wild" and confusing it with a `Transformation`, i.e. translation + rotation. I think `LogicTransformation` differentiates this better than `CSGTranformation`, but probably not a big deal.

sethrj: Yeah agreed. I just changed it to a flag.

PR #2019, opened 2025-10-08
Fix unsigned wraparound for `SurfaceTrackPosition`

amandalund:
The subsurface position (an `OpaqueId` type) intentionally relies on unsigned wraparound when moving backward from a pre-surface to an invalid position. However, `OpaqueId`'s `operator+` and `operator-` expect the value to not move below zero, leading to assertion failures when updating the subsurface position:
```console
corecel/OpaqueId.hh:242:
celeritas: internal assertion failed: offset >= 0 || static_cast<T>(-offset) <= id.unchecked_get()
```
This PR updates the position calculation in a few places where this assertion could be triggered and adds some basic `SurfacePhysicsUtils` tests. Currently the subsurface position remains an `OpaqueId`, though I know we've discussed whether that's the most appropriate type for it.

This was showing up in the `TrackingManagerIntegration` LAr sphere optical GPU tests (failing since #2006). Heads up that after this I'm still seeing another assertion failure there:
```console
celeritas/optical/surface/SurfacePhysicsView.hh:143:
celeritas: internal assertion failed: interface_pos < this->surface_record().subsurface_interfaces.size()
```


scripts/cmake-presets/jlse.json
```diff
@@ -37,13 +37,13 @@
...
       "name": "release",
       "displayName": "Build with optimizations and disable debug assertions",
-      "inherits": [".base", ".release"],
+      "inherits": [".base", ".ndebug"],
       "binaryDir": "${sourceDir}/build-release"
     },
     {
       "name": "release-amd",
       "displayName": "Release build for an AMD device",
-      "inherits": [".base", ".release-amd"],
```
amandalund, line 46: @elliottbiondo did you define this one locally? I couldn't find it in the base or jlse cmake presets.

hhollenb:
I haven't seen the first assertion failure yet, but I have seen the `SurfacePhysicsView` failure which I believe is due to the tracks exiting the surface (meaning their next surface model is undefined) but still checking to see if a model applies to them. Can you try the following patch to see if it fixes it?
```
diff --git a/src/celeritas/optical/action/TrackSlotExecutor.hh b/src/celeritas/optical/action/TrackSlotExecutor.hh
index 5250dbd04..f5306edf2 100644
--- a/src/celeritas/optical/action/TrackSlotExecutor.hh
+++ b/src/celeritas/optical/action/TrackSlotExecutor.hh
@@ -165,7 +165,8 @@ struct IsSurfaceModelEqual
     CELER_FUNCTION bool operator()(CoreTrackView const& track) const
     {
         auto s_phys = track.surface_physics();
-        return s_phys.is_crossing_boundary()
+        return IsStepActionEqual{s_phys.scalars().surface_stepping_action}(
+                   track)
                && s_phys.interface(step).surface_model_id() == model;
     }
 };
```

hhollenb (approved):
Definitely good to have some extra checks and tests, thanks!

src/celeritas/optical/surface/SurfacePhysicsUtils.hh
```diff
@@ -40,7 +40,7 @@ is_entering_surface(Real3 const& dir, Real3 const& normal)
 CELER_FORCEINLINE_FUNCTION SurfaceTrackPosition
 next_subsurface_position(SurfaceTrackPosition pos, SubsurfaceDirection dir)
 {
-    return pos + to_signed_offset(dir);
+    return SurfaceTrackPosition{pos.get() + to_signed_offset(dir)};
```
sethrj: Two suggested changes:
- Assert the position is valid to begin with (and not off the end) since `.get` just shows a generic OpaqueId assertion
- Be explicit about the casting from signed to unsigned int (especially since if we use mixed integer lengths, we could end up adding signed int to unsigned short/long and overflowing incorrectly)
 
```suggestion
    CELER_EXPECT(pos);
    return SurfaceTrackPosition{
        pos.unchecked_get()
        + static_cast<SurfaceTrackPosition::size_type>(to_signed_offset(dir))};
```

src/celeritas/optical/surface/SurfacePhysicsView.hh
```diff
@@ -159,9 +164,11 @@ template<class T>
 CELER_FUNCTION T SurfacePhysicsView::oriented_map(
     ItemMap<SurfaceTrackPosition, T> const& map, SurfaceTrackPosition pos) const
 {
-    auto index = orientation_ == SubsurfaceDirection::reverse
-                     ? SurfaceTrackPosition{(map.size() - 1) - pos.get()}
-                     : pos;
+    CELER_EXPECT(pos);
+    auto index
+        = orientation_ == SubsurfaceDirection::reverse
+              ? SurfaceTrackPosition{map.size() - 1 - pos.unchecked_get()}
```
sethrj, line 173: I think some static casts here as well, and also an assertion that map isn't empty?

sethrj (approved):
Thanks for the fix and tests!

scripts/cmake-presets/jlse.json
elliottbiondo: No, these were mistakes on my part, thanks for fixing!

Just to clarify, you aren't using the `build.sh` script for these, but soft-linking to `CMakePresets.json` and building via `cmake`?

amandalund: Correct!

amandalund:
@hhollenb thanks, that patch did fix the second failure! I'll push that here as well.

PR #2020, opened 2025-10-08
Add SurfaceInteractionApplier and implement PostBoundaryAction

hhollenb:
Add `SurfaceInteraction` and `SurfaceInteractionApplier` to optical surface physics. The applier will be used by the interaction step to apply the `SurfaceInteraction` result to the track.

The `PostBoundaryExecutor` is also implemented since the `SurfaceInteractionApplier` may change tracks to the `is_exiting()` state and off the surface. This executor will correctly update the geometry volume and reset the surface physics state so the photon can continue propagating. While a photon is in the `is_exiting()` state, the next surface model is undefined so the conditional track executor is updated to use the `surface_stepping_action` to tell if it should do a surface physics step instead of just having a valid surface ID.

### DIFF
--- a/src/celeritas/optical/action/TrackSlotExecutor.hh
+++ b/src/celeritas/optical/action/TrackSlotExecutor.hh
@@ -155,6 +155,11 @@ struct AppliesValidVolumetric
 /*!
  * Whether the track is undergoing a surface crossing and it matches the given
  * step and model.
+ *
+ * After the interaction step, a track may be on a surface but exiting so it
+ * does not have a valid next surface model. The surface stepping action is
+ * instead used, since tracks exiting the surface will have the post-boundary
+ * action set instead.
  */
 struct IsSurfaceModelEqual
 {
--- a/src/celeritas/optical/action/TrackSlotExecutor.hh
+++ b/src/celeritas/optical/action/TrackSlotExecutor.hh
@@ -165,7 +165,8 @@ struct IsSurfaceModelEqual
     CELER_FUNCTION bool operator()(CoreTrackView const& track) const
     {
         auto s_phys = track.surface_physics();
-        return s_phys.is_crossing_boundary()
+        return IsStepActionEqual{s_phys.scalars().surface_stepping_action}(
+                   track)
                && s_phys.interface(step).surface_model_id() == model;
     }
 };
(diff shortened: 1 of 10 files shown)


src/celeritas/optical/action/TrackSlotExecutor.hh
```diff
@@ -165,7 +165,8 @@ struct IsSurfaceModelEqual
     CELER_FUNCTION bool operator()(CoreTrackView const& track) const
     {
         auto s_phys = track.surface_physics();
-        return s_phys.is_crossing_boundary()
+        return IsStepActionEqual{s_phys.scalars().surface_stepping_action}(
```
sethrj, line 173: Can you add a note to the class documentation here about the subtle difference when exiting a surface that you mentioned in the PR description?

src/celeritas/optical/surface/model/SurfaceInteractionApplier.hh
```diff
@@ -0,0 +1,126 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Wrap a surface interaction executor and apply it to a track.
+ *
+ * The functor \c F must take a \c CoreTrackview and return a \c
+ * SurfaceInteraction.
+ */
+template<class F>
+struct SurfaceInteractionApplierBaseImpl
```
sethrj: We haven't done *any* launch parameter tweaking in the two years since we've had this capability. For simplicity let's remove the BaseImpl stuff and just have `SurfaceInteractionApplier` that assumes no kernel stuff.

sethrj (approved):
Thanks a bunch for splitting this out @hhollenb ! One thought is that we don't have great ways yet to test these, so maybe we need something like the `AlongStepTestBase` to set up and run single kernels and check the outcome.

hhollenb:
Yeah testing the kernels requires a hefty bit of machinery since the whole core params/state need to be setup for the kernels. I can start drafting that out now for a different PR.

sethrj:
Let's wait for kernel-level testing until @amandalund has the standalone loop work ready, which will probably simplify it. I think setting up the testing structure will benefit from you, me, and @amandalund putting our heads together to try to limit the amount of new code for testing.

PR #2023, opened 2025-10-08
Add regions to model input

amandalund:
As part of #983, this adds regions to `inp::Model` and loads the regions in the `GeantGeoParams`. It also removes the old `ImportRegion` and adds some very basic testing.


src/geocel/GeantGeoParams.cc
```diff
@@ -432,6 +478,75 @@ std::vector<inp::Detector> make_inp_detectors(GeantGeoParams const& geo)
...
+            auto const* g4reg = g4lv.GetRegion();
+            CELER_ASSERT(g4reg);
+            region_map[g4reg].insert(vol_id);
+        }
+        else
+        {
+            // Some of the volume data could override the region data: create a
+            // new Celeritas region containing only this volume
+            inp::Region region;
+            region.label = g4lv.GetName();
```
sethrj: I think we should try to incorporate the G4 region name into the label: so if volume `bob` is in region `scientists` but has a couple of extra properties, its new celeritas region would be `scientists@bob` so we can use the multilabel to select "all volumes in the `scientists` region" and still get `bob`.

src/geocel/GeantGeoParams.cc
```diff
@@ -432,6 +478,75 @@ std::vector<inp::Detector> make_inp_detectors(GeantGeoParams const& geo)
...
+            result.push_back(region);
+        }
+    }
+    CELER_LOG(debug) << "Loaded " << region_map.size() << " regions out of "
+                     << G4RegionStore::GetInstance()->size()
+                     << " Geant4 regions and created " << result.size()
+                     << " new regions";
+
+    // Add regions to result vector
+    for (auto&& [g4reg, volumes] : region_map)
```
sethrj, line 541: We'll need to sort the result as with the detectors since `region_map` is unordered (and has pointer keys).

sethrj (approved):
Awesome!

src/geocel/GeantGeoParams.cc
sethrj: @amandalund Just noticed you updated: using a `map` isn't sufficient since the keys are pointers. We can do the same thing as @lebuller did for detectors:
```c++
    std::sort(result.begin(), result.end(), [](auto& left, auto& right) {
        return left.volumes.front() < right.volumes.front();
    });
```

amandalund: Indeed! 🤦

src/geocel/GeantGeoParams.cc
```diff
@@ -525,7 +525,9 @@ std::vector<inp::Region> make_inp_regions(GeantGeoParams const& geo)
             // Some of the volume data could override the region data: create a
             // new Celeritas region containing only this volume
             inp::Region region;
-            region.label = g4lv.GetName();
+            region.label
+                = {g4lv.GetRegion() ? g4lv.GetRegion()->GetName() : "null",
```
sethrj, line 529: I think Geant4 has a `DefaultRegionForTheWorld` , maybe we should use that instead of null 😂 j/k

test/geocel/g4/GeantGeo.test.cc
```diff
@@ -451,8 +451,8 @@ TEST_F(MultiLevelTest, model)
     ref.world = "world";
     ref.detector.labels = {"sph_sd"};
     ref.detector.volumes = {{0, 5}};
-    ref.region.labels = {"box_region", "tri_region", "sph_region"};
-    ref.region.volumes = {{2, 4}, {1, 6}, {0, 5}};
+    ref.region.labels = {"sph_region", "tri_region", "box_region"};
+    ref.region.volumes = {{0, 5}, {1, 6}, {2, 4}};
```
sethrj, line 455: I guess there aren't any tests that adjust the physics for a specific LV? It might be good to test that (both for a volume that's in the default/world region, and for one that's in an existing region).

sethrj (approved):

test/geocel/g4/GeantGeo.test.cc
amandalund: Definitely agree, but I'm not sure if we can do that through the GDML...

sethrj: Maybe the gdml reader will show whether it attaches custom physics parameters? Do volume specific cutoffs invoke this?

src/geocel/GeantGeoParams.cc
amandalund: Each volume should always return a valid region though?

test/geocel/g4/GeantGeo.test.cc
amandalund: I don't think the Geant4 GDML reader uses the volume-specific auxiliary data at all. For region aux data it will read and set production cuts and user limits.

PR #2024, opened 2025-10-09
Improve warning messages during ORANGE simplification

sethrj:
- Move `simplify_joins` into a private function in UnitProto since it's not used or tested anywhere else
- Add more detailed warnings about unexpected nodes disappearing now that we have the proto as context
- Delete an unused method in CsgUnitBuilder
- Normalize some test reference strings so they fit on one line


esseivaju (approved):
LGTM

PR #2025, opened 2025-10-09
Add optical physics documentation

sethrj:
- Add documentation about ID-to-type mapping between Celeritas and Geant4 (requested by @whokion )
- Combine problem setup into a single doc page
- Update zotero and convert a few inline references into citations
- Improve surface physics and model documentation


amandalund (approved):
Thanks for the updates!

hhollenb (approved):
Looks good as well, especially the updates to the optical surface physics.

PR #2026, opened 2025-10-09
Prevent build.sh exiting if pre-commit not found at start

drbenmorgan:
Starting a fresh build locally, I found that running the recommended `build.sh` script did nothing more than:

```
$ ./scripts/build.sh dev
Determined system name: Mac (override with LMOD_SYSTEM_NAME)
```

I triaged this to the line that sticks the path of `pre-commit` into `OLD_PRE_COMMIT`. Since I _didn't_ have `pre-commit` installed this just failed and exited silently. The change here "protects" that detection in the same way as the CMake check, though I'm not sure it's truly needed as the `OLD_PRE_COMMIT` variable doesn't appear to be used elsewhere.

I still get an error at the end of the script when it tries to install the pre-commit hook, but that tells me what to do!

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -129,7 +129,7 @@ SYSTEM_NAME=$(fancy_hostname)
 
 # Check whether cmake changes from environment
 OLD_CMAKE=$(which cmake 2>/dev/null || echo "")
-OLD_PRE_COMMIT=$(which pre-commit 2>/dev/null)
+OLD_PRE_COMMIT=$(which pre-commit 2>/dev/null || echo "")
 
 # Load environment paths
 _env_script="scripts/env/${SYSTEM_NAME}.sh"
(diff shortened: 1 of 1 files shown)


sethrj (commented):
Thanks @drbenmorgan ! I think I meant to put a similar warning to the "old CMake" comparison; would that be easy to add? Also looking through the script (and talking with @elliottbiondo ) there are two other small things:
1. I think `install_precommit_if_git` should return 0 if git isn't available, rather than stopping there?
2. In some circumstances it seems that the hostname can return an empty string, leading to the creation of a bogus file `cmake-presets/.json`. Could you add guards for that where it's used? (You should be able to test the behavior with `HOSTNAME= ./build.sh ...`)

drbenmorgan:
I've added reporting of OLD/NEW precommit and a check for `SYSTEM_NAME`, but I wasn't quite sure about the check for `git`. The git executable is also used in `ln_presets` so should we put a non-fatal check there as well?

sethrj (approved):
Thanks much @drbenmorgan !

PR #2028, opened 2025-10-10
Finish making JLSE env script POSIX compliant

elliottbiondo:
This MR fixes a POSIX compliance issue I erroneously overlooked in #2017.


elliottbiondo:
@sethrj should be quick

sethrj (approved):

PR #2034, opened 2025-10-10
Use a single generator action and distribution buffer for Cherenkov and scintillation

amandalund:
There isn't any need to have separate distribution buffers and generator actions for scintillation and Cherenkov photon generation. This adds a `GeneratorType` field to the distribution struct and removes the templating on `GeneratorType` from the action. We still have separate offload actions for the two processes, but those will now write to the same buffer and the generation will be done in a single kernel launch. This should simplify the integration with Geant4 a bit.


src/celeritas/optical/gen/detail/GeneratorExecutor.hh
```diff
@@ -105,10 +96,19 @@ CELER_FUNCTION void GeneratorExecutor<G>::operator()(TrackSlotId tid) const
                               return state->init.vacancies[idx];
                           }()};
 
-    // Generate one primary from the distribution
+    // Generate one track from the distribution
     auto rng = track.rng();
     MaterialView opt_mat{material, dist.material};
-    vacancy = Generator(opt_mat, shared, dist)(rng);
+    if (dist.type == GeneratorType::cherenkov)
```
sethrj, line 103: So we're now branching inside of a single kernel rather than running separate kernels. Is there any measurable performance difference?

sethrj (approved):
Although this simplifies the built-in use case, it removes the generalization that would allow us to add primaries or other customizable distributions (e.g. uniform isotropic with gaussian energy). How soon do we need those cases? Is the plan to build them into new branches, or to add a new generalization, or to just duplicate+modify the generator code?

amandalund:
We do support optical primary generation with (eventually, since they're currently hardcoded) customizable distributions since #1858. The `GeneratorBase` class extracts most of the common code, but the logic for the primary generation is different enough from the scintillation/Cherenkov generation that we already have separate actions for the two.

src/celeritas/optical/gen/detail/GeneratorExecutor.hh
amandalund: Looking into this, but I'm currently seeing consistently ~30% more step iterations on this branch than develop and I still don't understand why (and if something's wrong now or was before or both or neither) :(

amandalund: Digging into this a little:
  - I don't think there's anything wrong in either case. The different number of step iterations seems to be a consequence of the particular model. In this (physically unrealistic) model, while the majority of the photons generated were from scintillation, almost none of those had an energy in the range where Rayleigh scattering applied. As a result, scintillation photons would only survive a single step iteration, while Cherenkov photons would take several steps before being killed. In the original implementation, all Cherenkov photons would be generated and tracked first, followed by all scintillation photons. Here they're mixed, and the mixing leads to more overall step iterations.
  - With the updated LAr sphere geometry used in our unit tests I was consistently seeing the opposite (more step iterations on develop), and did see scintillation photons undergoing Rayleigh scattering.
  - Turning off volumetric processes so we're just stressing the generation, I saw a 5-10% slowdown with this change. That may become negligible as we start running with more realistic physics and geometries, but for now I don't think it's worth the slight inconvenience of pushing to/copying over two buffers of distribution data in an integrated application, so I think I'll just close this.

sethrj: Your call! I didn't make the connection that the generators were intermingling the photons they create (and actually I'm still not 100% sure that is the case): is it possible to still have this simplified design but ensure that all the scintillation generation is contiguous, then all the cherenkov?

amandalund: Yeah, with the two generators/buffers the generator actions would be called in the same order each step, so Cherenkov would generate tracks in all the vacancies (leaving scintillation with nothing to do) until its buffer was depleted, and then scintillation would generate.

With this change the single generator will create both types from the one buffer, with the amount of mixing depending on things like the number of photons in each distribution and how mixed the distributions themselves are (maybe less if we're creating them in the Celeritas EM loop, more if we're pushing them at each Geant4 step).

We could sort the distributions of course (maybe not too expensive if it's just once per flush?) but I'll think if there are other ways.

amandalund: Ok, I finally figured this out: I _did_ have a bug in the optical photon generation, which was causing us to repeatedly generate photons from the same distributions. Going to fix that in a new branch and then reevaluate this.

amandalund: With the fix in #2053 I no longer see any difference in performance.

sethrj: Awesome! Nice work. Feel free to resolve this and merge whenever you're ready!

amandalund: Thanks for asking about performance, I wouldn't have found it otherwise!

PR #2035, opened 2025-10-11
Fix codecov deployment from master

sethrj:
Current PRs are getting invalid coverage reviews (after #2013) because the develop base branch could not push to codecov. After much gnashing of teeth I found this was because the organization secrets were not being propagated through the push action. Besides fixing this I made changes:
- Use sha for the codecov action as best practice
- Use a single "special" word `codecov` instead of `code-coverage` for action output
- Use bool for true/false instead of string
- Install dependencies as part of the action that uses it
- Simplify and move the gcov script to `ci/` which is its primary use (you can still use it as a developer, but it is primarily a CI script)


.github/workflows/push.yml
```diff
@@ -17,8 +17,10 @@ jobs:
     uses: ./.github/workflows/build-fast.yml
   build-docker:
     uses: ./.github/workflows/build-docker.yml
+    secrets: inherit
```
pcanal, line 20: Thanks for adding those in the push and deploy.  Sorry I missed them.

PR #2037, opened 2025-10-12
Add copyrights and consistent special name to workflow

sethrj:
- Add copyrights to the workflow headers
- Use `tidy` instead of `clang-tidy` to remove apparent separator
- Configure with verbosity
- Always build covfie, even when main code is c++17 (since we should be able to with cmake)


sethrj:
Ping @pcanal

.github/workflows/build-spack.yml
```diff
@@ -131,9 +138,6 @@ jobs:
                 && (env.CXXSTD == '20')}}; then
             spack -e . add root
           fi
-          if ${{env.CXXSTD == '20'}}; then
```
pcanal, line 134: For my information, does that mean that `covfie` is no longer spack-installed or that it is done in another way?

scripts/ci/spack.yaml
```diff
@@ -7,6 +7,7 @@ spack:
   - ccache
   - cli11 +pic ~ipo
   - cmake
+  - covfie
```
pcanal, line 10: This is the answer to https://github.com/celeritas-project/celeritas/pull/2037/files#r2433716170

.github/workflows/build-spack.yml
pcanal: It is now always install via the spack.yaml: https://github.com/celeritas-project/celeritas/pull/2037/files#r2433719877

pcanal (approved):
Looks good to me.  (Minor: I find 'tidy' less clear than 'clang-tidy')

PR #2038, opened 2025-10-13
Debug optical boundary crossing

### LINKED ISSUE #2041: Step counter inconsistency in `optical::PrimaryGenerator` tests
When implementing a trivial boundary crossing, the step counters in `optical::PrimaryGenerator` tests fail based on the build configuration. Vecgeom + g4 for versions above 11.0 will have slightly less steps than other configurations.

I've isolated some of the behaviors in #2038 so it's not related to the VecGeom normals, volumetric physics, or the current surface physics implementations. The discrepancy seems to be dependent on the number of primaries. Around 12000 and less tracks, the discrepancy disappears, but it returns for above around 14000 tracks.

Notably this uses the `lar-sphere` geometry, which has the following volumes:

1. world volume (vacuum)
2. detector shell (vacuum; sphere shell surrounding the other spheres)
3. detector shell top (LAr; semi-spherical shell that's an SD region)
4. detector shell bot (LAr; semi-spherical shell that's an SD region)
5. LAr sphere (LAr; uniform sphere inside all the other shells)

Photon primaries are generated isotropically in the center of the interior sphere. They are perfectly transmitted across boundaries but are killed if the post volume is not an optical material. In this case they transmit between volumes 3,4, and 5, and are killed upon reaching volume 2.

hhollenb:
While working on implementing optical boundary crossing, I've encountered some failing tests for the optical `PrimaryGenerator` based on vecgeom + g4 versus orange + geant4. The vecgeom builds have slightly less step counters, and sometimes different step iterators than the orange + geant4 version.

This branch isolates some of the recent changes (e.g. vecgeom normals, optical surface physics) so that it's just moving photons across a surface with no other physics involved.

Closes #2041

### DIFF
--- a/src/celeritas/optical/Transporter.cc
+++ b/src/celeritas/optical/Transporter.cc
@@ -77,6 +77,9 @@ void Transporter::transport_impl(CoreState<M>& state) const
             action->step(*params_, state);
         }
 
+        CELER_LOG(info) << " step_iter " << num_step_iters << ": "
+                        << counters.num_active << " active tracks";
+
         num_steps += counters.num_active;
         if (CELER_UNLIKELY(++num_step_iters == max_step_iters_))
         {
--- a/src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
+++ b/src/celeritas/optical/surface/detail/InitBoundaryExecutor.hh
@@ -55,6 +55,22 @@ CELER_FUNCTION void InitBoundaryExecutor::operator()(CoreTrackView& track) const
     auto geo = track.geometry();
     CELER_EXPECT(geo.is_on_boundary());
 
+    // Move the particle across the boundary
+    geo.cross_boundary();
+    if (CELER_UNLIKELY(geo.failed()))
+    {
+        track.apply_errored();
+        return;
+    }
+
+    // If post volume is not an optical material, kill the track.
+    if (!track.material_record().material_id())
+    {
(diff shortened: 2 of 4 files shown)


amandalund:
Were the differences you were seeing in the docker build? The primary generator test results here look consistent, though for orange I think that test was only run with the single precision build, which doesn't check the number of steps or step iterations.

hhollenb:
I'm seeing build errors in some of the build-spack tests: https://github.com/celeritas-project/celeritas/actions/runs/18468836404/job/52618665145?pr=2038#step:17:599

I've been comparing against some of the build-docker test failures as well, which build with orange and don't have these failures. I can move the PR out of drafts to run those as well

amandalund:
I think that test is failing in the same way at least in the spack build tests (we "continue on error" for vecgeom-g4 with Geant4 versions less than 11.0, so unfortunately the tests with earlier Geant4 versions were failing as well even though it looks like they pass).

hhollenb:
Gotcha, thanks for the heads up that's helpful to know!

Doing some more testing I found that the optical models were getting override in the LArSphereBase test, and after some searching it looks like only the Rayleigh model is causing the discrepancy. I know the inputs for that model got refactored a bit so I'll check that code next.

sethrj:
@hhollenb I'm closing this since Vecgeom errors are due to #1921 not being fixed. We can revisit at that point :)

PR #2040, opened 2025-10-13
Improve perfetto logger/failure integration

sethrj:
This dispatches the Perfetto log messages to the Celeritas logger. It also throws runtime errors if initialization fails rather than just warning: if the user wants profiling, we should give them profiling.

Finally it adjusts the output so that `diagnostic` messages (e.g., startup messages from perfetto and GDML loading) write their source file path.


esseivaju (approved):
Good addition, thanks!

PR #2044, opened 2025-10-13
Add orange builds to code coverage

pcanal:


sethrj:
@pcanal you may not have seen #2035 which I had to do over the weekend to fix the broken master. I shortened the `code-coverage` special to `codecov`; a similar shortening of `clang-tidy` to `tidy` is awaiting your review on https://github.com/celeritas-project/celeritas/pull/2037 .

.github/.codecov.yml
```diff
@@ -0,0 +1,44 @@
+# Codecov configuration
+# See https://docs.codecov.com/docs/codecov-yaml
```
sethrj, line 2: Is it possible to not have this file at the top level of the repository? Can we at least make it hidden? Or can we keep in `scripts/ci` and copy back to the root directory when running CI (or does codecov.org require it at top level in the source repo)?

.github/.codecov.yml
```diff
@@ -0,0 +1,44 @@
+# Codecov configuration
```
sethrj, line 1: Also please add a copyright header

.github/workflows/build-spack.yml
```diff
@@ -77,6 +77,10 @@ jobs:
             geometry: "vecgeom"
             special: "codecov"
             geant: "11.3"
+          - build_type: "reldeb"
+            geometry: "orange"
+            special: "code-coverage"
```
sethrj:
```suggestion
            special: "codecov"
```

.github/workflows/build-docker.yml
```diff
@@ -42,6 +42,11 @@ jobs:
             special: "codecov"
             geant: "11.3"
             image: "rocky-cuda"
+          - buildtype: "reldeb"
+            geometry: "orange"
+            special: "code-coverage"
```
sethrj:
```suggestion
            special: "codecov"
```

sethrj (commented):
Can you also make sure the examples count toward coverage?

Also as a follow-on do you think it's feasible to turn the spack/docker builds into another "action" like you did with codecov, and then we can have finer granularity so we can only build the expensive codecov checks after the earlier builds pass and not building as draft?

.github/.codecov.yml
pcanal: For a config file that is just (mostly) a copy/paste of the default ?  (i.e. what is the intellectual property there?)  [I will add it, but this is why I did not think of it]

.github/workflows/build-docker.yml
pcanal: Humm .. I missed https://github.com/celeritas-project/celeritas/pull/2035 ....

.github/.codecov.yml
sethrj: Looks like `.codecov.yml` is also valid, can we do that?

.github/.codecov.yml
sethrj: Let's see what other big+responsible projects do... spack doesn't, doesn't look like ROOT does with these minor files.  I do like having a header/footer in files... old habits? We can skip it.

.github/.codecov.yml
pcanal: We can also move it to the .github repository ...

sethrj:
@pcanal To reduce the turnaround time for your codecov testing I recommend deleting all the other jobs, then revering them later like so: https://github.com/celeritas-project/celeritas/pull/2035/commits/6fcbba242b937ee9c617c4d8c6b9bee8a3f64cde

Can you also please update the coverage to either use the results from the "test examples" job, or not run that part at all?

pcanal:
>  then we can have finer granularity so we can only build the expensive codecov checks after the earlier builds pass and not building as draft?

If I did not notice them (albeit they should be) to be much longer than other build.  Either way, if/since they are long they should run in parallel with the others rather than after to reduce turn around, shouldn't they?

pcanal:
> Can you also please update the coverage to either use the results from the "test examples" job

Yes, I am looking into that.

sethrj:
Unfortunately there are enough jobs that our "run pool" is getting full and new jobs end up requiring additional time :(

sethrj:
(Plus it blocks ohter PRs from running)

pcanal:
Temporarily Closing ... work is proceeding directly on fork.

Issue #2029, opened 2025-10-10
Unify Celeritas problem setup and construction: `inp` namespace


sethrj:
Creating a single setup mechanism for Celeritas, rather than the combination of ad-hoc runners that has developed over the last few years, will allow us to:
- Massively reduce duplicate code
- Create and destroy states in parallel so we can do gather/scatter operations (#1204)
- More tightly control how the parts of the code work to work together, so we don't have to (e.g.) assume users can add actions anytime before creating a state
- Better expose physics extension hooks to users
- Generate input definitions for front ends (#1204)

sethrj:
See #1562, #1601, #1632, #1735, #1739, #1862, #1953, #1976

PR #2045, opened 2025-10-13
Add options for constructing an ORANGE geometry

elliottbiondo:
This MR adds a `ConstructionOptions` struct in order to facilitate the study of performance-sensitive parameters that govern ORANGE geometry construction.

### DIFF
--- a/doc/usage/execution/environment.rst
+++ b/doc/usage/execution/environment.rst
@@ -18,36 +18,36 @@ has a default value but can be overridden with any boolean value.
 
 .. table:: Environment variables used by Celeritas.
 
- ======================== ========= ==========================================
- Variable                 Component Brief description
- ======================== ========= ==========================================
- CELER_BIH_MIN_SPLIT_SIZE orange    Set BIH ``min_split_size``
- CELER_COLOR              corecel   Flag: log with ANSI colors
- CELER_DEBUG_DEVICE       corecel   Flag: check device ID consistency
- CELER_DISABLE_DEVICE     corecel   Flag: disable CUDA/HIP support
- CELER_DISABLE_PARALLEL   corecel   Flag: disable MPI support
- CELER_DISABLE_REDIRECT   corecel   Flag: disable stream->logger redirection
- CELER_DISABLE_ROOT       corecel   Flag: disable ROOT I/O calls
- CELER_DISABLE_SIGNALS    corecel   Flag: disable signal handling
- CELER_DEVICE_ASYNC       corecel   Flag: allocate memory asynchronously
- CELER_ENABLE_PROFILING   corecel   Flag: use NVTX/ROCTX profiling [#pr]_
- CELER_LOG                corecel   Set "global" logger verbosity [#lg]_
- CELER_LOG_LOCAL          corecel   Set "local" logger verbosity [#lg]_
- CELER_MEMPOOL... [#mp]_  corecel   Set ``cudaMemPoolAttrReleaseThreshold``
- CELER_PERFETT... [#bs]_  corecel   Set in-process tracing buffer size
- CELER_PROFILE_DEVICE     corecel   Flag: record kernel launch counts
- CUDA_HEAP_SIZE           geocel    Set ``cudaLimitMallocHeapSize`` (VG)
- CUDA_STACK_SIZE          geocel    Set ``cudaLimitStackSize`` for VecGeom
- G4ORG_OPTIONS            orange    JSON filename for G4-to-ORANGE conversion
- G4VG_COMPARE_VOLUMES     geocel    Check G4VG volume capacity when converting
- HEPMC3_VERBOSE           celeritas HepMC3 debug level integer
- VECGEOM_VERBOSE          celeritas VecGeom CUDA verbosity integer
- CELER_DISABLE            accel     Flag: disable Celeritas offloading entirely
- CELER_LOG_ALL_LOCAL      accel     Flag: log more messages from all threads
- CELER_KILL_OFFLOAD       accel     Flag: kill offloaded tracks [#ko]_
- CELER_NONFATAL_FLUSH     accel     Flag: print and continue on failure [#nf]_
- CELER_STRIP_SOURCEDIR    accel     Flag: clean exception output
- ======================== ========= ==========================================
+ ========================= ========= ==========================================
+ Variable                  Component Brief description
+ ========================= ========= ==========================================
+ CELER_COLOR               corecel   Flag: log with ANSI colors
+ CELER_DEBUG_DEVICE        corecel   Flag: check device ID consistency
+ CELER_DISABLE_DEVICE      corecel   Flag: disable CUDA/HIP support
+ CELER_DISABLE_PARALLEL    corecel   Flag: disable MPI support
+ CELER_DISABLE_REDIRECT    corecel   Flag: disable stream->logger redirection
+ CELER_DISABLE_ROOT        corecel   Flag: disable ROOT I/O calls
+ CELER_DISABLE_SIGNALS     corecel   Flag: disable signal handling
+ CELER_DEVICE_ASYNC        corecel   Flag: allocate memory asynchronously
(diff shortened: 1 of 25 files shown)


elliottbiondo:
@sethrj can you take a preliminary look at this? We already have  `g4org::Options` -> `InputBuilder::Options`, so the addition of `OrangeInput::ConstructionOptions` feels suspicious.

src/orange/OrangeInput.hh
```diff
@@ -159,6 +161,19 @@ struct RectArrayInput
...
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Options that govern the geometry construction process.
+ */
+struct ConstructionOptions
+{
+    //! Options for Bounding Interval Hierarchy (BIH) construction
+    detail::BIHBuilder::Options bih_options;
```
sethrj: It might make it more messy in the short term, but I think the preferred pattern will look more like the newer celeritas code: `inp/Foo.hh` has the setup options for `Foo`. So we'd have `orange/inp/Bih.hh` which would have `struct inp::Bih` and that would replace `BihBuilder::Options`. We'd likewise move `orange/gorg/Options.hh` to `orange/inp/Converter.hh`. Finally we'd also move `OrangeInput` to `orange/inp/Model.hh` or similar.

src/orange/detail/BIHBuilder.hh
```diff
@@ -51,11 +51,14 @@ class BIHBuilder
     //!@{
     //! \name Type aliases
 
-    //! Input parameters
-    struct Input
+    //! Options parameters
+    struct Options
```
sethrj: if we move the struct, this'll become `using Input = inp::BihBuilder;` as is done in the other update celeritas classes.

src/orange/OrangeInput.hh
elliottbiondo: FWIW, I think `Options` makes more sense, lest we overload the word "input"; we already have an `orangeinp` folder. Likewise, below, I think semantically the input to the `BIHBuilder` are the bounding boxes, whereas the `min_split_size` is an option specifying how the process is done.

elliottbiondo: The `orange/inp` folder doesn't currently exist. I can make, but I am still not clear on how InputBuilder::Options fits in (seems like that should be defined at the `OrangeInput` level).

I also, I think `g4org::Options` should consist of: 1) a high-level struct will all general OrangeOptions, that is also used for a non-G4 workflow to specify all ORANGE options in one place, and 2) additional G4-specific options.

sethrj: The weirdness/confusion is because we've been inconsistent in the past and are transitioning toward having an `inp` namespace that has *all* of the input structs that were formerly called Input/Options. This namespace is to be the primary interface with the outside parts of the code: see #2029.

In a hypothetical/future refactorization that I am talking about:
- `celeritas::OrangeInput` -> `celeritas::inp::Orange`
- `celeritas::detail::BIHBuilder::Options` -> `celeritas::inp::BihBuilder`
- `celeritas::g4org::Options` -> `celeritas::inp::Converter` (?)
- `celeritas::orangeinp` -> `celeritas::setup` (?)
- `celeritas::g4org` -> `celeritas::setup::g4org` (?)

I don't know what's the right path between that future and our present, but I would at least like:
- BIH builder doesn't include half of orange transitively via the OrangeInput (since VariantSurfaces)
- OrangeInput doesn't include and rely on a detail namespace
- Mlutiple layers of nested options with only one member per struct

elliottbiondo: Not sure what you mean by these two:

> BIH builder doesn't include half of orange transitively via the OrangeInput (since VariantSurfaces)

> Mlutiple layers of nested options with only one member per struct

elliottbiondo:
@sethrj I am finally getting around to fixing this MR since I want to do the performance tests this week.

I moved the options struct to `inp/` and added a `CELER_BIH_MIN_SPLIT_SIZE` environment variable. This is ready for another look.

sethrj:
@elliottbiondo Thanks for picking this back up! Just mark it as "ready" to ask me for another review :)

doc/usage/execution/environment.rst
```diff
@@ -18,35 +18,36 @@ has a default value but can be overridden with any boolean value.
...
- CELER_DISABLE           accel     Flag: disable Celeritas offloading entirely
- CELER_LOG_ALL_LOCAL     accel     Flag: log more messages from all threads
- CELER_KILL_OFFLOAD      accel     Flag: kill offloaded tracks [#ko]_
- CELER_NONFATAL_FLUSH    accel     Flag: print and continue on failure [#nf]_
- CELER_STRIP_SOURCEDIR   accel     Flag: clean exception output
- ======================= ========= ==========================================
+ ======================== ========= ==========================================
+ Variable                 Component Brief description
+ ======================== ========= ==========================================
+ CELER_BIH_MIN_SPLIT_SIZE orange    Set BIH ``min_split_size``
```
sethrj: Rename `ORANGE_`... and since this is a temporary hack (we're going to add more options via a cleaner interface) don't bother putting it in the environment docs. No one will know what it means...

src/orange/detail/UnitInserter.hh
```diff
@@ -39,7 +39,9 @@ class UnitInserter
 
   public:
     // Construct from full parameter data
-    UnitInserter(UniverseInserter* insert_universe, Data* orange_data);
+    UnitInserter(UniverseInserter* insert_universe,
+                 Data* orange_data,
+                 ConstructionOptions* opts);
```
sethrj: Why are the options passed by mutable pointer?

src/orange/inp/Bih.hh
```diff
@@ -0,0 +1,28 @@
...
+namespace inp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Construction options for BIH tree.
+ */
+struct BIHBuilder
+{
+    //! Minimum number of bboxes needed to trigger a partitioning attempt
+    size_type min_split_size = 2;
```
sethrj, line 22: Missing include `corecel/Types.hh`

src/orange/orangeinp/InputBuilder.cc
```diff
@@ -13,6 +13,7 @@
 #include "corecel/io/JsonPimpl.hh"
 #include "corecel/io/Logger.hh"
 #include "corecel/io/ScopedTimeLog.hh"
+#include "corecel/sys/Environment.cc"
```
sethrj:
```suggestion
#include "corecel/sys/Environment.hh"
```

sethrj (approved):
Needs a couple of fixes but I think this is the right way to do it for the short term. Thanks!

src/orange/OrangeInput.hh
sethrj: - If `BIHBuilder` included `OrangeInput.hh` then it would end up pulling in a bunch of "setup-time" files. I guess it doesn't really matter since those are just included by the other setup code: when I wrote that I was thinking it was *runtime* (device) code that included it.
- I think I meant *avoiding* the second one? e.g.: avoid `struct BarOpts { int value; }; struct FooOpts { BarOpts bar_opts; }`

elliottbiondo: That all makes sense. I was confused because BIHBuilder did not include OrangeInput.hh, but you were speaking hypothetically. And yes, at some point maybe an options struct in `inp/orange.hh` can eliminate the `inp/Bih.hh` file.

elliottbiondo:
@sethrj RFAL

sethrj:
Note that we *very rarely* use const pointers; only as an "optional reference". Use const references in most cases. The reason for mutable pointers was following an older version of the Google style guide, which points out that passing by reference "hides" whether the argument is going to be modified or not. (This is I think one big shortcomings of C++ : sometimes you want signatures to be identical between mutable/immutable, but passing mutable arguments makes it hard to see from the caller's point of view what data is being read vs written.)

PR #2048, opened 2025-10-14
Fix polysolid with zero-radius inner segment

sethrj:
The `alice-pipe-its` model has a polysolid that has an inner exclusion but consecutive zero inner radii for parts (indicating no  exclusion in those segments). This adds a couple of helpful assertions and skips building the inner zones when not applicable.


elliottbiondo:
Simple enough! I suspect we will continue to see these sorts corner cases as we explain our library of models.

elliottbiondo (approved):

PR #2050, opened 2025-10-15
Fix Mie scattering execution on device

amandalund:
Tests with Mie scattering enabled were failing on device because the reference to the device data wasn't set in the executor.


sethrj (approved):
Thanks!

Rashika-Gupta (approved):
Thanks, Amanda

Issue #2046, opened 2025-10-14
Fix and implement last couple of missing ORANGE shape conversions


sethrj:
#1167 didn't quite get everything.
- CMS, LHCb, and ALICE all use "cut tubes"
- CMS and ALICE both have polyhedra with `numsides="1"` and inner radii, which is different from what we originally thought
- CMS and ATLAS both use polyhedra with different lo/hi radii
- I forgot about hyperbolic tube (but it's not used in any of our geometry)

Example cut tube from CMS run 3:
```xml
    <cutTube aunit="deg" deltaphi="7.90680999999999" highX="0" highY="0" highZ="1" lowX="0.48599950039277" lowY="0.00835999140593325" lowZ="-0.873919101611625" lunit="mm" name="pixfwdInnerDiskZplus_PixelForwardInnerDiskOuterRing_seg_10x7f7110ba4900" rmax="117.35" rmin="114.85" startphi="87.03229" z="15.5"/>
```

Example numsides=1 polyhedra from CMS run 4:
```xml
    <polyhedra aunit="deg" deltaphi="20" lunit="mm" name="HEC10x7f1fffce6500" numsides="1" startphi="350">
      <zplane rmax="1866.5" rmin="1775" z="3242"/>
      <zplane rmax="1866.5" rmin="1775" z="3347.8"/>
      <zplane rmax="1927.4" rmin="1775" z="3347.8"/>
      <zplane rmax="1927.4" rmin="1775" z="3436.4"/>
      <zplane rmax="1927.4" rmin="1838.8" z="3436.4"/>
      <zplane rmax="1927.4" rmin="1838.8" z="3770.42"/>
      <zplane rmax="1987.89" rmin="1838.8" z="3816.02"/>
      <zplane rmax="2876.5" rmin="2770.7" z="4462.99"/>
      <zplane rmax="2876.5" rmin="2813.42" z="4493.47"/>
      <zplane rmax="2876.5" rmin="2813.42" z="5541"/>
    </polyhedra>
```

Example "different rmin/rmax" polyhedra:
```xml
    <polyhedra aunit="deg" deltaphi="360" lunit="mm" name="HGCalHEAbsorber110x7f1fff5a7880" numsides="36" startphi="355">
      <zplane rmax="1984.08417370622" rmin="348.6" z="-20.75"/>
      <zplane rmax="2036.42657691209" rmin="348.6" z="20.7400000000002"/>
[...]

PR #2051, opened 2025-10-16
Add support for G4CutTubs

elliottbiondo:
This MR adds `CutCylinder` to `IntersectRegion` and corresponding support for `G4CutTubs` in `SolidConverter`.

Advances #2046


elliottbiondo:
@sethrj RTR

src/corecel/math/ArrayUtils.hh
```diff
@@ -198,6 +204,33 @@ is_soft_orthogonal(Array<T, N> const& x, Array<T, N> const& y)
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Check whether two vectors are approximately collinear.
+ *
+ * This is done by checking if the Gram determinant is soft zero:
+ * \f[
+ * (x \cdot x) (y \cdot y) - (x \cdot y)^2 \approx 0
+ * \f]
+ * This method is advantageous because it doesn't require normalizing inputs.
```
sethrj: Suppose we have (ignoring FP arithmetic for now) $`y = x + \epsilon e`$, where both *x* and *e* are unit vectors, and $` \epsilon `$ is the magnitude of the non-collinearity. (Moving a unit distance along each vectors will result in a track being an epsilon distance away of where it should be.) Then the determinant is $`2 c \epsilon - (\epsilon c)^2 `$ where $`0 \le c = e \cdot x  \le 1`$.

If I did the math correctly then I think the tolerance for soft zero here is slightly conservative (we should use 2 * rel), but I was afraid it was off by a power of epsilon (which it's not).

src/corecel/math/ArrayUtils.hh
```diff
@@ -198,6 +204,33 @@ is_soft_orthogonal(Array<T, N> const& x, Array<T, N> const& y)
...
+ */
+template<class T, size_type N>
+inline CELER_FUNCTION bool
+is_soft_collinear(Array<T, N> const& x, Array<T, N> const& y)
+{
+    SoftZero const soft_zero{SoftEqual{}.rel()};
+    T const xxyy = dot_product(x, x) * dot_product(y, y);
+
+    if (soft_zero(xxyy))
```
sethrj: If `x` and `y` aren't unit vectors, the soft zero test isn't meaningful (or at least, it has the wrong scale): I think there should be an implicit documented precondition that these are both unit vectors, and a CELER_EXPECT(!soft_zero(xxyy));` here.

(In the doxygen, use `\pre Both x and y must be approximately unit vectors`)

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -236,6 +237,81 @@ void Cone::output(JsonPimpl* j) const
...
+                   << "top cutting plane normal is not pointing up: "
+                   << top_normal_[Z]);
+    CELER_VALIDATE(is_soft_unit_vector(bot_normal_),
+                   << "bottom cutting plane normal is not a unit vector");
+    CELER_VALIDATE(is_soft_unit_vector(top_normal_),
+                   << "top cutting plane normal is not a unit vector");
+    CELER_VALIDATE(!soft_equal<real_type>(-1, bot_normal_[Z])
+                       || !soft_equal<real_type>(1, top_normal_[Z]),
+                   << "both planes are perpendicular to Z (use cylinder "
+                      "instead)");
```
sethrj: I think we should probably take out this and similar conditions in the code: we can't test soft equality correctly at region build time, and more importantly we don't have extra code in the solid converter to do this fallback.

The original idea behind preventing multiple regions from having the same representation was to do CSG simplification /deduplication before we even start making surfaces, but it's not implemented and I'm not sure it's worth it (there are better+smarter+more general ways to do improve construction).

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -236,6 +237,81 @@ void Cone::output(JsonPimpl* j) const
...
+//---------------------------------------------------------------------------//
+/*!
+ * Whether this encloses another cut cylinder
+ */
+bool CutCylinder::encloses(CutCylinder const& other) const
+{
+    // Check that cylinders have the same cut planes. Since the normal vectors
+    // have unit magnitude, a check for collinearity is sufficient. When
+    // cylinders have different cut planes, testing for enclosure becomes
+    // challenging.
```
sethrj, line 279: Plus the intended use case for this is always as a "solid" where we have identical normals :)

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -138,6 +138,56 @@ class Cone final : public IntersectRegionInterface
...
+ * axis; a Cylinder should be used in this case.
+ */
+class CutCylinder final : public IntersectRegionInterface
+{
+  public:
+    // Construct with radius, half-height, and bottom/top cut plane normals
+    CutCylinder(real_type radius,
+                real_type halfheight,
+                Real3 bottom_normal,
+                Real3 top_normal);
```
sethrj: const reference for "large" objects that can't be moved

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -138,6 +138,56 @@ class Cone final : public IntersectRegionInterface
...
+    real_type radius() const { return radius_; }
+
+    //! Half-height along Z
+    real_type halfheight() const { return hh_; }
+
+    //! Outward-facing normal of the bottom cuting plane
+    Real3 bottom_normal() const { return bot_normal_; }
+
+    //! Outward-facing normal of the top cuting plane
+    Real3 top_normal() const { return top_normal_; }
```
sethrj: Here as well

src/orange/orangeinp/IntersectRegion.hh
```diff
@@ -138,6 +138,56 @@ class Cone final : public IntersectRegionInterface
...
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * A *z*-aligned cylinder centered on the origin, with top/bottom cuts.
+ *
+ * The shape is defined with a radius, half-height, and the outward-facing
+ * normals of the cutting planes, passing through (0, 0, +/-hh). This class
+ * does not support the case where both cut planes are perpendicular to the z
+ * axis; a Cylinder should be used in this case.
```
sethrj: Could you use `\f$(0, 0, \pm  \mathrm{hh})\f$` , italicize `z`, and add this to `doc/implementation/geometry/orange.rst`?

test/corecel/math/ArrayUtils.test.cc
```diff
@@ -102,6 +102,20 @@ TEST(ArrayUtilsTest, is_soft_orthogonal)
...
+{
+    EXPECT_TRUE(is_soft_collinear(Dbl3{2, 0, 0}, Dbl3{1, 0, 0}));
+    EXPECT_TRUE(is_soft_collinear(Dbl3{0, 1, 0}, Dbl3{0, -3, 0}));
+    EXPECT_TRUE(is_soft_collinear(Dbl3{0, -5, -5}, Dbl3{0, 1, 1}));
+    EXPECT_TRUE(is_soft_collinear(Dbl3{1e-13, 1, 0}, Dbl3{0, 1, 0}));
+
+    EXPECT_FALSE(is_soft_collinear(Dbl3{2, 1, 3}, Dbl3{1, 0, 0}));
+    EXPECT_FALSE(is_soft_collinear(Dbl3{1e-6, 1, 0}, Dbl3{0, 1, 0}));
+    EXPECT_FALSE(is_soft_collinear(Dbl3{1e-6, 1, 0}, Dbl3{0, 1, 0}));
+    EXPECT_FALSE(is_soft_collinear(Dbl3{0, 0, 0}, Dbl3{1, 0, 0}));
```
sethrj: Some of these will become expect_throw. Also it might be good to test `{1 +/- eps, 0, 0}, {1,0,0}`.

sethrj (approved):
Great! Thank you.

src/corecel/math/ArrayUtils.hh
```diff
@@ -198,6 +204,33 @@ is_soft_orthogonal(Array<T, N> const& x, Array<T, N> const& y)
...
+ * \f[
+ * (x \cdot x) (y \cdot y) - (x \cdot y)^2 \approx 0
+ * \f]
+ * This method is advantageous because it doesn't require normalizing inputs.
+ */
+template<class T, size_type N>
+inline CELER_FUNCTION bool
+is_soft_collinear(Array<T, N> const& x, Array<T, N> const& y)
+{
+    SoftZero const soft_zero{SoftEqual{}.rel()};
```
sethrj: `SoftEqual` needs to explicitly use type `<T>`, not the default `real_type`, for the single-precision build to pass. The construction generator I think should work automatically:
```suggestion
    SoftZero const soft_zero{SoftEqual<T>{}.rel()};
```

src/orange/orangeinp/IntersectRegion.cc
elliottbiondo: Yes, it only matters for inner/outer type solid construction but I wouldn't want this to cause issues down the road if `encloses` ends up getting used for something else.

src/corecel/math/ArrayUtils.hh
elliottbiondo: There is no issue with scale: `xxyy` is zero if and only if one of the supplied vectors is (0, 0, 0).

If we assume the unit vectors are normalized, the function would just `return soft_eq(1, dot_product(x, y))`. I wrote this function with support for non-normalized vectors for consistency with `is_soft_orthogonal`.

elliottbiondo: (and I didn't want to use make_unit_vector because that involves a square root)

sethrj: The *tolerance* of being near zero depends on their scale right? If they both have magnitude 0.01 and are collinear then won't your check say "they're soft zero therefore false" because the magnitude of them gets multiplied in 4 times?

Exact zero is easy, tolerance is hard.

src/corecel/math/ArrayUtils.hh
elliottbiondo: Isn't it just this?

Screenshot 2025-10-16 at 3 05 57 PM

This has a squared epsilon term, but then again, so does `is_soft_orthongonal` due to just the dot product.

src/corecel/math/ArrayUtils.hh
elliottbiondo: Good point. This should really just be checking for non-soft (0, 0, 0), such that we don't return `true` in that case.

src/corecel/math/ArrayUtils.hh
sethrj: Uh oh so I did make a mistake. But the problem isn't that there *is* an epsilon squared term, the problem is that it's the *leading order term*. The second-order term in `is_soft_ortho` can be neglected since there's a first-order term. But here it can't; and you need a square root of the value (or a square of the epsilon) to do the correct magnitude comparison.

elliottbiondo: Ah yes, `eps > eps^2`

elliottbiondo:
@sethrj RFAL

elliottbiondo:
@sethrj okay I am now expecting vectors to be normalized going in. This should be good pending CI.

PR #2053, opened 2025-10-17
Fix calculation of distribution index in optical photon generation

amandalund:
When calculating the index of the distribution that a thread should generate photons from in the scintillation/Cherenkov generator, we weren't adding the offset from the start of the buffer. This meant that while we were generating the correct number of photons, we were repeatedly generating them from the distributions at the front of the buffer. This PR fixes that calculation and expands one of the tests to more accurately mirror the generation. Fortunately this didn't affect any of the performance results.


sethrj (approved):
Good find, good test! Thank you!

PR #2054, opened 2025-10-17
Fix G4Polyhedra with azimuthal truncation

elliottbiondo:
This MR adds support `G4Polyhedra`  with a `StackedExtrudedPolygon` approach per #2046.


elliottbiondo:
@sethrj this is a little messy, but it works. RTR.

elliottbiondo:
Oh, I see in the issue that you suggested I still go through `PolyPrism`. I didn't think of doing it that way, but also don't think there would be much benefit.

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -738,6 +738,7 @@ TEST_F(SolidConverterTest, polyhedra)
     static double const no_rmin[] = {0, 0, 0, 0};
     static double const rmin[] = {20, 4, 17, 4};
     static double const rmax[] = {30, 10, 20, 5};
+    static real_type const sqrt2 = std::sqrt(2);
```
sethrj: use `constants::sqrt_two`, which stays double precision unless you demote it

src/orange/g4org/SolidConverter.cc
```diff
@@ -716,46 +716,102 @@ auto SolidConverter::polycone(arg_type solid_base) -> result_type
...
+    using VecReal3 = StackedExtrudedPolygon::VecReal3;
+
     auto const& solid = dynamic_cast<G4Polyhedra const&>(solid_base);
     auto const& params = *solid.GetOriginalParameters();
+    size_type num_sides = params.numSide;
+    size_type num_z = params.Num_z_planes;
+    real_type start_angle = params.Start_angle;
+    real_type open_angle = params.Opening_angle;
+    bool is_two_pi
+        = soft_equal(open_angle, static_cast<real_type>(2 * constants::pi));
```
sethrj: Use `enclosed_azi_from_poly(solid)`, which has all the proper logic for edge cases and rounding (and also normalizes the result so stop > start). Then you can use `if (azi)` and `azi.start()` etc.

src/orange/g4org/SolidConverter.cc
```diff
@@ -716,46 +716,102 @@ auto SolidConverter::polycone(arg_type solid_base) -> result_type
...
+
+    CELER_VALIDATE(num_sides >= 1, << "must have at least one side");
+    CELER_VALIDATE(num_z >= 2, << "must have at least two z planes");
+
+    // If there is only 1 side, the "opening angle" must be less than pi. If
+    // there are only two sides, the opening angle must be less than 2pi.
+    CELER_VALIDATE(
+        num_sides > 2 || (!is_two_pi && open_angle < num_sides * constants::pi),
+        << "invalid number of sizes for the opening angle");
```
sethrj:
```suggestion
        num_sides > 2 || !azi || (azi.stop() - azi.start() > Turn{real_type{0.5} * num_sides})`,
        << "invalid number of sizes for the opening angle");
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -716,46 +716,102 @@ auto SolidConverter::polycone(arg_type solid_base) -> result_type
...
+    VecReal2 polygon;
+    real_type step = open_angle / params.numSide;
+    for (auto i : range(num_sides))
     {
-        zs[i] = scale_(params.Z_values[i]);
-        rmin[i] = scale_(params.Rmin[i]) * radius_factor;
-        rmax[i] = scale_(params.Rmax[i]) * radius_factor;
+        polygon.push_back(make_point(start_angle + i * step));
+    }
+    if (!is_two_pi)
```
sethrj:
```suggestion
    if (azi)
```

src/orange/g4org/SolidConverter.cc
```diff
@@ -716,46 +716,102 @@ auto SolidConverter::polycone(arg_type solid_base) -> result_type
...
+    // define the polygon and all subsequent scaling
+    real_type r0 = rmax[0];
 
-    std::vector<real_type> zs(params.Num_z_planes);
-    std::vector<real_type> rmin(zs.size());
-    std::vector<real_type> rmax(zs.size());
-    for (auto i : range(zs.size()))
+    // Create counterclockwise polygon on basis of r0;
+    auto make_point = [&r0](real_type theta) {
+        return Real2{r0 * std::cos(theta), r0 * std::sin(theta)};
```
sethrj: `sincos` works for `Turn` (and there's also plain sin/cos overloads).

src/orange/g4org/SolidConverter.cc
```diff
@@ -716,46 +716,102 @@ auto SolidConverter::polycone(arg_type solid_base) -> result_type
...
+        auto outer = StackedExtrudedPolygon::or_solid(base_name + "_outer",
+                                                      std::move(polygon),
+                                                      std::move(polyline),
+                                                      make_scaling(rmax));
+        auto inner = StackedExtrudedPolygon::or_solid(base_name + "_inner",
+                                                      std::move(inner_polygon),
+                                                      std::move(inner_polyline),
+                                                      make_scaling(rmin));
+
+        return make_subtraction(std::move(base_name), outer, inner);
```
sethrj, line 811: This works but is going to make a very messy solid (lots of duplicate Z planes etc). I wonder if it would be worth giving the SEP another argument? But that will be an optimization so we can do that in a follow-on PR (just add a to-do here). Also please use `.int` (interior) rather than `_outer` and `.exc`(excluded) rather than `_inner` for your adjusted names above for consistency with Solid/PolySolid.

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -847,39 +876,40 @@ TEST_F(SolidConverterTest, subtractionsolid)
...
+//     this->build_and_test(
+//         reflx,
+//         R"json({"_type":"transformed","daughter":{"_type":"transformed","daughter":{"_type":"shape","interior":{"_type":"prism","apothem":0.9999999999999999,"halfheight":2.0,"num_sides":3,"orientation":0.25},"label":"tri"},"transform":{"_type":"translation","data":[0.0,0.0,3.0]}},"transform":{"_type":"transformation","data":[1.0,0.0,0.0,0.0,1.0,0.0,0.0,0.0,1.0,0.0,0.0,0.0]}})json",
+//         {
+//             {0, 0.99, 1.1},
+//             {0, -0.99, 5.1},
+//             {0, 1.01, 1.1},
+//             {0, -1.01, 5.1},
+//         });
+// }
```
sethrj: Uncomment and fix; and I recommend making a habit of `#if 0`/`#endif` rather than individual line commenting. Most editors even render disabled `#if 0` blocks as a comment.

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -734,47 +734,76 @@ TEST_F(SolidConverterTest, polycone)
...
+    static double const no_rmin[] = {0, 0, 0, 0};
+    static double const rmin[] = {20, 4, 17, 4};
+    static double const rmax[] = {30, 10, 20, 5};
+    static real_type const sqrt2 = std::sqrt(2);
 
-    // Flat-top hexagon
+    // Full diamond shape, no interior
     this->build_and_test(
         G4Polyhedra(
-            "HGCalEEAbs", 330 * deg, 360 * deg, 6, std::size(z), z, rmin, rmax),
```
sethrj, line 744: Can you please keep but update these original tests, and add your new ones after? The HGCal one is an actual use case that we need to check, and I think I verified the full 360-degree polyhedra as well (see `test/geocel/data/polyhedra.gdml`):
polyhedra

sethrj (changes requested):

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -734,47 +734,76 @@ TEST_F(SolidConverterTest, polycone)
...
+            "polyhedra", 0 * deg, 180 * deg, 2, std::size(z), z, rmin, rmax),
+        R"json({"_type":"all","daughters":[{"_type":"stackedextrudedpolygon","polygon":[[4.242640687119285,0.0],[2.5978681687064796e-16,4.242640687119285],[-4.242640687119285,5.195736337412959e-16],[0.0,0.0]],"polyline":[[0.0,0.0,-1.0],[0.0,0.0,0.0],[0.0,0.0,2.0],[0.0,0.0,2.5]],"scaling":[1.0,0.3333333333333333,0.6666666666666666,0.16666666666666666]},{"_type":"negated","daughter":{"_type":"stackedextrudedpolygon","polygon":[[4.242640687119285,0.0],[2.5978681687064796e-16,4.242640687119285],[-4.242640687119285,5.195736337412959e-16],[0.0,0.0]],"polyline":[[0.0,0.0,-1.0],[0.0,0.0,0.0],[0.0,0.0,2.0],[0.0,0.0,2.5]],"scaling":[0.6666666666666666,0.13333333333333336,0.5666666666666668,0.13333333333333336]},"label":""}],"label":"polyhedra"})json",
+        {
+            {19, 1, -1},
+            {25, 1, -1},
+            {31, 1, -1},
+            {0, 31, -1},
+            {2, 29, -1},
+        });
```
sethrj: Can you also please add the actual CMS shapes described in https://github.com/celeritas-project/celeritas/issues/2046 ? That'll help verification even if you don't put manual points (since the test also does random sampling, 4k points by default).

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: Just doing `using constant::sqrt_two` causing problems converting `Constant` -> `double`. Instead I am doing:

```
auto sqrt_two = double{constants::sqrt_two};
```

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: Not sure what you are getting at with the polyhedra.gdml. You're mentioning it just because those the tri's in their are what is tested in the original tests?

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: Alas, the z-planes for `HEC10x7f1fffce6500` are not strictly monotonically increasing: there are repeated z-planes! We don't support with `StackedExtrudedSolid` because it would require making cones with zero height.

For now the terrible hack of simply bumped these planes by 1e-12, results in passing test with 4k points checked.

I think I can add support for this with `StackedExtrudedPolygon` but I will have to look.

elliottbiondo: Ughh, the other CMS one is even worse. We hit:
```
167: C++ exception with description "/Users/veb/opt/celeritas/celeritas/src/orange/orangeinp/detail/CsgUnitBuilder.cc:91:
167: celeritas: internal assertion failed: static_cast<bool>(bzone.interior) == static_cast<bool>(existing.bounds.interior)" thrown in the test body.
```
This will take some tinkering.

src/orange/g4org/SolidConverter.cc
elliottbiondo: The `azi.start()` isn't particularly helpful in this case, as we *don't* want to normalize the start/stop. The start determines where the vertices of the polygon are, so even if you've got a 2pi situation you still need to know the original start. I will use it instead of my is_two_pi functor though.

src/orange/g4org/SolidConverter.cc
elliottbiondo: I am sure I could refactor this with `Turn`, but since we can't use `EnclosedAzi` for the reason I stated above, I don't think it's worth it: how I have it now is pretty clear.

src/orange/g4org/SolidConverter.cc
elliottbiondo: You mean another argument for an interior shape? That's possible, but won't deduplication handle it anyway?

(Fixed the .int and .exc, but they don't get used for a subtraction solid anyway: it turns into "daughter")

elliottbiondo:
@sethrj this is probably ready for another look. I still need to fix the aforementioned CMS issues, but that will involves changes that might be better suited for another MR.

test/orange/g4org/SolidConverter.test.cc
sethrj: Yes, sorry. The polyhedra GDML solids (which have been verified with the ray tracer) are constructed as Geant4 solids in the tests that were deleted. So keeping them ensures we have a "validated" test case.

src/orange/g4org/SolidConverter.cc
sethrj: Yes, another argument for an interior shape. Deduplication currently simplifies surfaces, but CSG tree simplification is a separate matter. We do a few trivial simplifications there, but I'm pretty sure constructing a CSG tree with minimal nodes is an NP hard problem.

src/orange/g4org/SolidConverter.cc
sethrj: Ah, I forgot that although the start angle is kept in the regular constructor, it's *discarded* if the enclosure is a full circle.  👍

sethrj (commented):
Based on the failing tests due to numerical imprecision, and the remaining issues with CMS, I think this is only halfway the right approach. The `PolyPrism::build` method can be completely overridden, and I think it would make more sense to use the convex hull approach with the PolySegments class, moving the logic for the polygon calculation inside there. (I'd also change `orientation` to just be a `Turn start_angle`.) You might be able to trivially factor out the within-level polygon construction from StackedExtrudedPolygon so that it could be used there?

test/orange/g4org/SolidConverter.test.cc
sethrj: Can you elaborate on the problems? The `Constant` is designed so that:
```
op(Constant, double) -> double
op(Constant, float) -> float
op(Constant, int) -> Constant
op(Constant, Constant) -> Constant
```
where `op` is any math operation (and should be symmetric).

elliottbiondo:
In terms of CMS: the consecutive zero thing is actually an easy fix I think. I don't know the cause of the error for the second shape: it's a pretty weird shape. By floating point errors you mean within the .json comparison, correct? I am not sure I understand how `PolySolid` offers an advantage there.

sethrj:
@elliottbiondo I'm guessing the other one is due to infinities/negations messing up the intersections/unions.

And yeah with floating point errors (I should've said "precision") the difference is that, if using the Polyhedra, the JSON output would contain only *stuctural* information: the basic quantities pulled from Geant4, not computed quantities like the polygon points, which have greater variance due to the extra operations. Separating the two also allows "validation of import" to be more clearly separated from "validation of geometry construction".

elliottbiondo:
Okay cool, that should be a fairly easy modification. The downsides of this approach are that 1) we are essentially duplicating the effort `StackedExtrudedPolygon` rather than using it as directly, and 2) we are propagating the design choices of G4, which we may not agree with, into ORANGE-proper. I do agree that is it convenient that we don't have to deal with floating-point comparison within strings (I am surprised this hasn't come up before?). I will be surprised if this fixes the issue with the second CMS shape, as I don't think `construct_segments` is doing anything fundamentally different than `StackedExtrudedPolygon`.

sethrj:
> Okay cool, that should be a fairly easy modification. The downsides of this approach are that 1) we are essentially duplicating the effort `StackedExtrudedPolygon` rather than using it as directly

Hopefully there's an easy way to refactor most of the functionality so this isn't duplicated🤞 

> , and 2) we are propagating the design choices of G4, which we may not agree with, into ORANGE-proper. 

Most of our "objects" are based on Geant4 solids. Sometimes we simplify them (and then compose the result) and sometimes we genericize them (and provide the same functionality as multiple geant4 solids, like trap+trd+gentrap). There's room for both. I completely agree that this polyhedra isn't necessarily intuitive or useful outside Geant4 geometry, but building from an "inner + outer radius" allows us to harmonize the shape with our polycone, so the semantics of stacking and providing those constraints are the same. The "num sides" is the real bonkers choice, so perhaps we should clarify it's "number of edges along r_outer" somehow. The blasé characterization of the shape as a "polyhedra" is also goofy. Maybe we chould merge its functionality with the extruded polygon.

> I do agree that is it convenient that we don't have to deal with floating-point comparison within strings (I am surprised this hasn't come up before?).

The JSON comparator uses `SoftEqual` but this case happens to be one where the default (i.e., inherited from Nemesis/Draco) relative tolerance differing from the *absolute* tolerance causes problems. 1e-13 and 3e-13 are both within an absolute value of 1e-12 (which is the relative tolerance) but not within 1e-14. I have an open issue (in my head) to move the Orange tolerance (relative but with length scale that defaults to 1) up to where SoftEqual lives, and use a single value (i.e. the "bump distance") for SoftEqual.

> I will be surprised if this fixes the issue with the second CMS shape, as I don't think `construct_segments` is doing anything fundamentally different than `StackedExtrudedPolygon`.

The important thing about the segment construction, aside from omitting soft-zero z planes, is that it constructs one z segment at a time rather than the whole stack. I think that will allow it a better bounding box. I think while you're at it, you should add some tests of the extruded polygon with more than 2 z segments since testing looks a little sparse there.

elliottbiondo:
@sethrj this is getting pretty messy. The machinery for building a `Polysolid`, i.e., `construct_segments`, expects each segment to be an `IntersectRegion`. Of course `StackedExtrudedPolygon` (our only tool for doing CDT) is not an `IntersectRegion`. So even if I refactor out the stacking process from `StackedExtrudedPolygon`, it's not like I could drop it into `PolySolid`.

To me this looks more like a refactor to `StackedExtrudedPolygon` to support:
1) Excluded inner regions, especially cases where some segments (but not all) have zero-radii inner regions
2) Zero-height segments, which could come naturally from switching the loop order such that CDT is done as an inner loop within each z-segment.

sethrj:
@elliottbiondo I wasn't necessarily suggesting shoehorning into the existing polysolid infrastucture; it seems like there's less in common between the polycone and polyhedra. I think the segments and some of the code there could still be useful but I leave it to you.

My main goals are (1) avoid code duplication and (2) construct the tightest and smallest bounding boxes for the CSG nodes.  That means doing axial segments on the outside, polygon on the inside.

elliottbiondo:
All of this can be accomplished by changing:
```
    StackedExtrudedPolygon(std::string&& label,
                           VecReal2&& polygon,
                           VecReal3&& polyline,
                           VecReal&& scaling);
```
to:
```
    StackedExtrudedPolygon(std::string&& label,
                           VecReal2&& polygon,
                           VecReal3&& polyline,
                           VecReal&& inner_scaling);
                           VecReal&& outer_scaling);
```
 
and then reordering the `StackedExtrudedPolygon` construction process (which should incidentally fix the zero-height segment issue).

The `polyline` + `inner_` and `outer_scaling` handles all `G4ExtrudedSolid` and `G4Polyhedra` use cases. In other words, the remaining work is basically `StackedExtrudedPolygon` work, and the code in this MR is going to be nearly unchanged. Reordering `StackedExtrudedPolygon` is going to a heavy lift (probably rewrite 500+ lines of code/detailed doc strings/tests) and should probably be a separate MR.

sethrj:
@elliottbiondo Fair enough! I think with that approach the polyprism will be basically useless; could you confirm and, if so, add a note to the class that it's slated for removal?

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: That doesn't seem to be working within initializer lists for some reason. When I do:

```
using constants::sqrt_two;
```

I get:

```
/Users/veb/opt/celeritas/celeritas/test/orange/g4org/SolidConverter.test.cc:765:17: error: no matching constructor for initialization of 'const celeritas::Array<double, 3>'
  765 |                 {sqrt_two*15, sqrt_two*15, -9.},
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/veb/opt/celeritas/celeritas/src/corecel/cont/Array.hh:74:15: note: candidate constructor template not viable: no known conversion from 'Constant' to 'double' for 1st argument
   74 |     CELER_CEF Array(T first, Us... rest) : d_{first, static_cast<T>(rest)...}
```

I don't want to static_cast each time so I went with:

```
auto sqrt_two = real_type{constants::sqrt_two};
```

elliottbiondo:
@sethrj this should be ready for another look. Once this is in I will start a new MR for `StackedExtrudedPolygon`.

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -734,47 +746,187 @@ TEST_F(SolidConverterTest, polycone)
...
+        this->build_and_test(
+            G4Polyhedra(
+                "tri", 90 * deg, 360 * deg, 3, std::size(z2), z2, rmin2, rmax2),
+            R"json({"_type":"shape","interior":{"_type":"extrudedpolygon","bot_line_segment_point":[0.0,0.0,1.0],"bot_scaling_factor":1.9999999999999998,"polygon":[[0.0,1.0],[-0.8660254037844389,-0.49999999999999956],[0.8660254037844385,-0.5000000000000001]],"top_line_segment_point":[0.0,0.0,5.0],"top_scaling_factor":1.9999999999999998},"label":"tri"})json");
+    }
+
+    // CMS TESTS
+    // \TODO: Add support for duplicate z planes in StackedExtrudedPolygon
+    // to support these cases
+#if 0
```
sethrj: It would be a little better in this case to uncomment it and wrap the `build_and_test` in an `EXPECT_THROW(..., RuntimeError);` with the TODO above it

sethrj (approved):
I think this is good. One last request is that for two more related tests:
1. A simple stacked extruded polygon that has three segments, and the top segment is displaced completely outside the radial extents of the bottom one (so if it's a box of halfwidth 1, then the polyline could be x={0, .75, 1.5}) to check the bounding box construction
2. Take one of your polyhedra that has at least two levels and an inner+outer, and add a test to the "subtraction" case in `CsgObjects.test.cc` to check the CSG tree construction

Thanks!

elliottbiondo:
(Oh and yes, I can confirm that the `PolyPrism` stuff can be deleted. It will not be useful for anything once `StackedExtrudedPolygon` is refactored.)

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -907,26 +906,41 @@ TEST_F(SolidConverterTest, polyhedra)
...
+            celeritas::DebugError);
```
sethrj: DebugError is for programming errors (assertion failures, precondition violations, bad assumptions) whereas RuntimeError is for things like malformed user input and system status that we can't know about until runtime
```suggestion
            celeritas::RuntimeError);
```

elliottbiondo: This hits a programming error that I am yet to debug, hence the `DebugError`

sethrj: I thought it was a shortcoming that resulted in a `CELER_VALIDATE` error throwing? Since DebugError throws should (with very very few exceptions) be disabled in release builds, any GoogleTest that checks for a DebugError needs to have an `if (CELERITAS_DEBUG)` wrapping it.

Also the reason I said RuntimeError was that I was seeing one in the test output, so there must be a secondary failure as well:
```
/home/runner/work/celeritas/celeritas/test/orange/g4org/SolidConverter.test.cc:942
Expected: this->build_and_test( G4Polyhedra("HGCalHEAbsorber110x7f1fff5a7880", 0 * deg, 360 * deg, 36, std::size(z2), z2, rmin2, rmax2), R"json({"_type":"all","daughters":[{"_type":"stackedextrudedpolygon","polygon":[[1.0,0.0],[0.984807753012208,0.17364817766693033],[0.9396926207859084,0.3420201433256687],[0.8660254037844387,0.5],[0.766044443118978,0.6427876096865393],[0.6427876096865393,0.766044443118978],[0.5,0.8660254037844386],[0.3420201433256689,0.9396926207859083],[0.17364817766693044,0.984807753012208],[0.0,1.0],[-0.17364817766693044,0.984807753012208],[-0.34202014332566855,0.9396926207859084],[-0.4999999999999999,0.8660254037844387],[-0.6427876096865393,0.766044443118978],[-0.7660444431189778,0.6427876096865396],[-0.8660254037844385,0.5000000000000001],[-0.9396926207859083,0.3420201433256689],[-0.984807753012208,0.17364817766693044],[-1.0,0.0],[-0.984807753012208,-0.17364817766693044],[-0.9396926207859083,-0.3420201433256689],[-0.8660254037844389,-0.49999999999999956],[-0.7660444431189783,-0.642787609686539],[-0.6427876096865396,-0.7660444431189778],[-0.5000000000000001,-0.8660254037844385],[-0.3420201433256689,-0.9396926207859083],[-0.17364817766693044,-0.984807753012208],[0.0,-1.0],[0.17364817766692975,-0.9848077530122081],[0.3420201433256682,-0.9396926207859085],[0.49999999999999956,-0.8660254037844389],[0.642787609686539,-0.7660444431189783],[0.7660444431189778,-0.6427876096865396],[0.8660254037844385,-0.5000000000000001],[0.9396926207859083,-0.3420201433256689],[0.984807753012208,-0.17364817766693044]],"polyline":[[0.0,0.0,-2.075],[0.0,0.0,2.0740000000000203],[0.0,0.0,2.07400000000012],[0.0,0.0,2.07499999999995]],"scaling":[199.16630529221047,204.42053956048494,204.42053956048494,204.42180594618483]},{"_type":"negated","daughter":{"_type":"stackedextrudedpolygon","polygon":[[1.0,0.0],[0.984807753012208,0.17364817766693033],[0.9396926207859084,0.3420201433256687],[0.8660254037844387,0.5],[0.766044443118978,0.6427876096865393],[0.6427876096865393,0.766044443118978],[0.5,0.8660254037844386],[0.3420201433256689,0.9396926207859083],[0.17364817766693044,0.984807753012208],[0.0,1.0],[-0.17364817766693044,0.984807753012208],[-0.34202014332566855,0.9396926207859084],[-0.4999999999999999,0.8660254037844387],[-0.6427876096865393,0.766044443118978],[-0.7660444431189778,0.6427876096865396],[-0.8660254037844385,0.5000000000000001],[-0.9396926207859083,0.3420201433256689],[-0.984807753012208,0.17364817766693044],[-1.0,0.0],[-0.984807753012208,-0.17364817766693044],[-0.9396926207859083,-0.3420201433256689],[-0.8660254037844389,-0.49999999999999956],[-0.7660444431189783,-0.642787609686539],[-0.6427876096865396,-0.7660444431189778],[-0.5000000000000001,-0.8660254037844385],[-0.3420201433256689,-0.9396926207859083],[-0.17364817766693044,-0.984807753012208],[0.0,-1.0],[0.17364817766692975,-0.9848077530122081],[0.3420201433256682,-0.9396926207859085],[0.49999999999999956,-0.8660254037844389],[0.642787609686539,-0.7660444431189783],[0.7660444431189778,-0.6427876096865396],[0.8660254037844385,-0.5000000000000001],[0.9396926207859083,-0.3420201433256689],[0.984807753012208,-0.17364817766693044]],"polyline":[[0.0,0.0,-2.075],[0.0,0.0,2.0740000000000203],[0.0,0.0,2.07400000000012],[0.0,0.0,2.07499999999995]],"scaling":[34.99315953676109,34.99315953676109,42.019898399564525,42.019898399564525]},"label":""}],"label":"HGCalHEAbsorber110x7f1fff5a7880"})json") throws an exception of type celeritas::DebugError.
  Actual: it throws celeritas::RuntimeError with description "runtime error: z coordinates must be strictly increasing
/home/runner/work/celeritas/celeritas/src/orange/orangeinp/StackedExtrudedPolygon.cc:103: 'std::adjacent_find(polyline_.begin(), polyline_.end(), [](auto const& a, auto const& b) { return a[Z] >= b[Z]; }) == polyline_.end()' failed".
```

sethrj:
Judging by the test output we also seem to be introducing some additional imprecision, since OrangeGeantTest now has a small discrepancy:

```
[ RUN      ] PolyhedraTest.trace
/home/runner/work/celeritas/celeritas/test/geocel/GeoTests.cc:836: Failure
Expected: (ref) == (result), but
Values in: halfway_safeties
 Expected: ref
2 of 9 elements differ
by 9.9999999747524271e-07 relative error or 9.9999999747524271e-07 absolute error
i        ref    ACTUALDifference
1       0.99   0.989999 -1.20413e-06
7       0.99   0.989999 -1.20413e-06
```

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: Both CMS tests throw a `RuntimeError` for having zero-height `z` segments. The first one goes away if you bump, which what I was originally doing, but I took it out because 1) bumping is hack anyway, 2) I converted to `EXPECT_THROW`. The second CMS test hits a `CELER_ASSERT` even with bumping. I will wrap the second with `if (CELERITAS_DEBUG)`.

elliottbiondo:
(FYI I haven't done the `CsgObjects.test.cc` test yet)

elliottbiondo:
Unfortunately, using a more robust cross product implementation did not fix the `GeoTests` issue, even when doing the whole thing in double before converting back to `float`. I bumped the tolerance by 25%, though I fully recognize that this is very unsatisfying because I don't actually know the reason.

elliottbiondo:
@sethrj pending CI this is RFAL

PR #2061, opened 2025-10-20
Add a trivial surface interaction model

hhollenb:
This trivial interaction model returns a surface interaction independent of the rest of the surface physics state. Each surface can has one mode associated with it that it will apply to all incident photons. The three modes are: absorb, transmit, or back-scatter covering the absorb / refract / reflect interactions respectively.

While unrealistic, this model is useful for debugging and testing of surface physics.


sethrj (approved):
Perfect! Great commit size too 😉

amandalund (approved):
Great, thanks @hhollenb!

PR #2062, opened 2025-10-20
Create .github/.codecov.yml, loosen test requirement

pcanal:


.github/.codecov.yml
```diff
@@ -0,0 +1,44 @@
...
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: false
+
+ignore:
+#  - "test/**/*"
+#  - "tests/**/*"
+#  - "example/**/*"
+#  - "examples/**/*"
```
sethrj: Can this dead code be deleted?

sethrj (approved):
Thanks @pcanal !

Since `.github` is already invisible, and `.github/codecov.yml` is acceptable, could we name it that? Also, can you add to the header:
- A link to https://docs.codecov.com/docs/codecovyml-reference
- A note that validation can be performed with `curl -X POST --data-binary @codecov.yml https://codecov.io/validate`

And for follow-on work it would be good to validate the codecov yaml early in the CI process since it's fast.

pcanal:
> Since .github is already invisible, and .github/codecov.yml is acceptable, could we name it that? 

Sure

> A link to [docs.codecov.com/docs/codecovyml-reference](https://docs.codecov.com/docs/codecovyml-reference)

I added it ... but just a note that it is already listed in the first page of the link that was already there.

> And for follow-on work it would be good to validate the codecov yaml early in the CI process since it's fast.

ok.

Issue #2057, opened 2025-10-20
Reconstruct G4NavHistory from ORANGE state with volume instance stack


sethrj:
To correctly map to `G4NavigationHistory`, we need the volume instance ID (aka G4 physical volume) at every level in the geometry hierarchy. Since multiple levels can be inlined into a single universe, simply mapping local volume ID to volume instance ID isn't sufficient to determine or reconstruct this hierarchy. We need to determine the local parent volume ID (if any) for every local volume

Obstacles:
- The LevelId and depth we use isn't consistent across our geometry implementations; the ORANGE "universe" is should be an `ImplLevelId`
- Determining the depth of the volume stack cannot be done from the level ID alone (unless inlining is disabled). We have to track the "local depth", i.e. how many nested volumes are in the current universe
- May need to have the multi-level volume instance vector be bottom-up, not top-down, and stop when the first ID is false, rather than requiring `volume_instance(Span)` be correctly sized

Related issues:
- Extends #1749 
- Required #1353, #1815

PR #2063, opened 2025-10-21
Add UnivLevel and rename Level to VolumeLevel

sethrj:
This PR improves nomenclature and type safety to avoid confusion between canonical volume depth and ORANGE universe depth:
- Renames the "level" states in the ORANGE track view to "univ_level" with a corresponding `UnivLevelId` type.
- Renames "level" to "volume_level" in the G4/VG track views, that returns the canonical volume depth. (A placeholder ORANGE function here will be implemented in a follow-on PR.)
- Changes `max_depth` (one more than the deepest DepthId) to `num_X_levels` consistently in the geometry classes and in the volume params.
- Changes `volume_instance_depth` in the HitProcessor to `num_volume_levels`
- **Changes the ordinality of the VolumeParams "depth"**: previous `depth = 0` is now `num_volume_levels = 1`
- Rename UniverseId -> UnivId for brevity and consistency with UnivLevelId, and then universe -> univ in many places


sethrj:
Please take a look at this @elliottbiondo when you have a moment!

Also @pcanal is the codecov failure due to a lack of ORANGE coverage tests, or something else?

pcanal:
@sethrj Codecov claims that half the new lines in [src/orange/OrangeTrackView.hh](https://app.codecov.io/gh/celeritas-project/celeritas/pull/2063?src=pr&el=tree&filepath=src%2Forange%2FOrangeTrackView.hh&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=celeritas-project#diff-c3JjL29yYW5nZS9PcmFuZ2VUcmFja1ZpZXcuaGg=) are not being tested.   Some of them are `CELER_ENSURE` (probably need to 'void' out this macro for the Code Coverage build) and a function that is intentionally not used (`NOT_IMPLEMENTED`) and then some of the partial I need to investigate and/or suppress.

src/geocel/Types.hh
```diff
@@ -37,7 +39,7 @@ using SquareMatrixReal3 = SquareMatrix<real_type, 3>;
 //---------------------------------------------------------------------------//
 
 //! Type-safe "level", i.e., depth of embedded unit/scene/volume
-using LevelId = OpaqueId<struct Level_>;
+using LevelId = OpaqueId<struct Level_, std::uint_least8_t>;
```
elliottbiondo: 256 is probably conservative enough, but you might consider adding some DBCs during construction to make sure this doesn't get overstepped.

src/orange/OrangeTrackView.hh
```diff
@@ -506,9 +507,18 @@ CELER_FUNCTION VolumeInstanceId OrangeTrackView::volume_instance_id() const
...
+    return get_vol_inst(lev);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * The depth in the canonical volume graph.
+ */
+CELER_FUNCTION LevelId OrangeTrackView::depth() const
+{
+    CELER_NOT_IMPLEMENTED("canonical depth");
```
elliottbiondo: This will never be implemented, since this a G4-specific metric, correct? If it's just there for the template interface it should be marked as such.

elliottbiondo (changes requested):
As I understand it there are two concepts here:

1) "Depth": the nesting depth of G4Solids,
2) "Level": the nesting depth of ORANGE universes,

which are equivalent in the case of no inlining in the g4orange conversion process. I think using two generic words ("depth" and "level") to differentiate between these two specific concepts may lead to confusion. I think something "SolidLevel" and "UnivLevel", i.e. specifying what we kind of level (or depth) we are talking about, would be better.

Additional comments:
- Using `ImplLevelId` instead of `LevelId` seems superfluous if only ORANGE geometries have levels. What about `LocalVolId`s?
- `OrangeData::.scalars.max_depth` and `detail::DepthCalculator` probably needs to updated if  `depth` remains a G4-specific concept.

src/geocel/Types.hh
sethrj: Great point. I thought about that but forgot to implement it.

src/orange/OrangeTrackView.hh
sethrj: Actually no, OrangeTrackView will have to have all the volume structure metadata available on device: see https://github.com/celeritas-project/celeritas/issues/2057 . This is what I'm working toward.

sethrj:
>  Codecov claims that half the new lines in [src/orange/OrangeTrackView.hh](https://app.codecov.io/gh/celeritas-project/celeritas/pull/2063?src=pr&el=tree&filepath=src%2Forange%2FOrangeTrackView.hh&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=celeritas-project#diff-c3JjL29yYW5nZS9PcmFuZ2VUcmFja1ZpZXcuaGg=) are not being tested. Some of them are `CELER_ENSURE` (probably need to 'void' out this macro for the Code Coverage build) and a function that is intentionally not used (`NOT_IMPLEMENTED`) and then some of the partial I need to investigate and/or suppress.

@pcanal 
LevelStateAccessor is the really confusing one. Is the problem that `FORCEINLINE` means codecov never "sees" the line being touched even though it is?

sethrj:
> As I understand it there are two concepts here:
> 
> 1. "Depth": the nesting depth of G4Solids,

I'm trying to use more graph theoretical terms for the volume hierarchy: the "world" has depth zero, its daughter volumes (which correspond to G4LogicalVolume) are nodes of depth 1, etc.

> 2. "Level": the nesting depth of ORANGE universes,

This is what I've traditionally called it in ORANGE, based on the Lava/MCNP `lev`, but depth would probably be better here too.

> which are equivalent in the case of no inlining in the g4orange conversion process.

Correct.

> I think using two generic words ("depth" and "level") to differentiate between these two specific concepts may lead to confusion. I think something "SolidLevel" and "UnivLevel", i.e. specifying what we kind of level (or depth) we are talking about, would be better.

Agreed; so maybe part of the confusion is I should rename `LevelId` `DepthId`? Then `LevelId` in ORANGE (which refers to universe) becomes `ImplDepthId`?

> Additional comments:
> 
> * Using `ImplLevelId` instead of `LevelId` seems superfluous if only ORANGE geometries have levels. What about `LocalVolId`s?
> * `OrangeData::.scalars.max_depth` and `detail::DepthCalculator` probably needs to updated if  `depth` remains a G4-specific concept.

Yes, the "depth" here was also a problem I realized as I tried to unify all the different volume representations. There, "max depth" is really "maximum value of LevelId + 1" which is why I called it thus. I should just call it `universe_depth` for orange and `volume_depth` for the graph structure.

sethrj:
I'm working on a massive name change for depth/max depth/etc. I'll do that first, revert the size change, update the PR title, and then follow on with a separate PR to reduce the int size and add appropriate checks.

elliottbiondo:
> Agreed; so maybe part of the confusion is I should rename LevelId DepthId? Then LevelId in ORANGE (which refers to universe) becomes ImplDepthId?

> I should just call it `universe_depth` for orange and `volume_depth` for the graph structure.

If you go with `universe_depth`/`volume_depth` is there still a need for the `Impl` notation?

sethrj:
No, I'll be removing that as well.

pcanal:
> LevelStateAccessor is the really confusing one. Is the problem that FORCEINLINE means codecov never "sees" the line being touched even though it is?

That is a good point (`FORCEINLINE`), I am currently investigating in more details how codecov gets its info with a view on tuning the build to be more code-coverage friendly and/or tweaks the codecov options.

sethrj:
In the meantime @pcanal I'm deferring those changes to the track view.

@elliottbiondo  I think I've got all the names changed and working now, including a new type-safe UnivLevelId. Please take another look. All changes related to ID sizes, and some other minor ORANGE changes, have been reverted for now.

src/orange/LevelStateAccessor.hh
```diff
@@ -7,14 +7,19 @@
...
 #include "OrangeTypes.hh"
 
 namespace celeritas
 {
 //---------------------------------------------------------------------------//
 /*!
- * Access the 2D fields (i.e., {track slot, level}) of OrangeStateData.
+ * Access the 2D fields (i.e., {track slot, udepth}) of OrangeStateData.
+ *
+ * \internal The class name is from the legacy nomenclature of "level" meaning
```
elliottbiondo: I don't think we should keep it this way long term. Add a \todo to rename to `UnivDepthStateAccessor`

src/orange/detail/DepthCalculator.hh
```diff
@@ -20,6 +20,8 @@ namespace detail
 //---------------------------------------------------------------------------//
 /*!
  * Calculate the maximum number of levels deep in a geometry.
+ *
+ * \todo Replace with VolumeVisitor? Rename UnivDepthCalculator?
```
elliottbiondo, line 24: `UnivDepthCalculator`, if anything

elliottbiondo:
Is your intent that on it's own, "depth" always means volume depth? While I would prefer the more emphatic "`vol_depth`", I think this is reasonable, as this is the terminology used in Geant4 (e.g., `G4NavigationHistory.hh`).

sethrj:
Shit I guess `depth` is also prone to off-by-one errors... it *sometimes* is the maximum number of nodes, *sometimes* the number of errors. I think we can define `volume_depth` as the "depth in terms of volumes [nodes]", so number of nodes along any path. That's also an argument for namign it volume_depth in VolumeParams :)

src/geocel/VolumeParams.cc
```diff
@@ -16,16 +16,16 @@ namespace celeritas
 namespace
 {
 //---------------------------------------------------------------------------//
-LevelId::size_type calc_depth(VolumeParams const& params)
+int calc_depth(VolumeParams const& params)
```
elliottbiondo: Just curious why you changed this. In `PolyPrism` you use `int num_sides` as well, so it seems there is a pattern I am missing.

src/orange/LevelStateAccessor.hh
elliottbiondo: Is there a reason other than scope creep not to change this?

src/orange/OrangeTrackView.hh
```diff
@@ -102,7 +102,9 @@ class OrangeTrackView
     inline CELER_FUNCTION VolumeId volume_id() const;
     // Get the canonical volume instance ID in the current impl volume
     inline CELER_FUNCTION VolumeInstanceId volume_instance_id() const;
-    // Get the volume instance ID for all levels
+    // The depth in the canonical volume graph
```
elliottbiondo: Add one line to the top-level doc string explaining that principal "canonical volume graph" is the graph of G4Solids in the `g4org` case.

elliottbiondo:
I don't think `depth` eliminates the off-by-one confusion either. Europeans and Americans disagree on what floor of the building is the "first floor", and `depth` has the same issue: does a single global universe have a depth of 0 or 1? The principle we are trying to communicate is cardinality, and I think `num` is probably the best way to do it. I think we'll have to agree to disagree.

elliottbiondo (approved):

src/orange/OrangeTrackView.hh
elliottbiondo: Maybe it isn't the right place for this. Looks like I didn't review `VolId`/`ImplVolId`, so I am not really familiar with this.

sethrj: Yeah, that is all explained in the VolumeParams and extensively in the geometry docs; see file:///Users/seth/Code/celeritas/build/doc/html/implementation/geometry.html

src/orange/LevelStateAccessor.hh
sethrj: Yeah, I think there's a note somewhere in here that this class is actually used by SCALE orange so it would be a breaking change. (But I might've already broken it already...) I could still do it and add a deprecated alias since we've already creeped the scope so much already 🙃

src/geocel/VolumeParams.cc
sethrj: This was compatibility with the volume params recursive function. There's probably a better heuristic for deciding int versus unsigned, but since using 'int' to tell the compiler that a value can't overflow, I use it for unbounded loops.

src/orange/detail/DepthCalculator.hh
sethrj: I think we could probably eliminate the class by using the VolumeVisitor which abstracts the DFS/BFS recursions, so I'd rather just not touch it

src/orange/OrangeTrackView.hh
sethrj: And FWIW the canonical/depth/etc will be further documented in the follow-on PR that makes a virtual interface for testing.

src/orange/LevelStateAccessor.hh
sethrj: Since it would require keeping a stub file too, let's just avoid the scope creep.

sethrj:
Thanks @elliottbiondo for all the feedback and discussion; I think ultimately you are right about the improved clarity.

This change set was very painful. What I thought would be a trivial change ended up requiring many iterations of find/replace/compile/fix etc.  I think the lesson is to try to plan out in advance what name changes should be performed... that is on me, so thanks for bearing with me.

PR #2064, opened 2025-10-22
Update pre-commit hooks and add ruff

sethrj:
This adds a python formatter (ruff) for the python scripts we have scattered around, and it bumps slightly the clang-format version. I added but disabled the ruff linter: it caught an unused code error, but it has some false positives (e.g., sphinx conf.py) that I don't want to deal with right now.


.pre-commit-config.yaml
```diff
@@ -7,30 +7,41 @@ repos:
...
     hooks:
     - id: codespell
 
-  # TODO: https://github.com/celeritas-project/celeritas/issues/1763
+  # TODO: licensing
+  # see https://github.com/celeritas-project/celeritas/issues/1763
 
 ci:
+  autofix_commit_msg: |
+    Apply pre-commit hooks: please install pre-commit locally!
```
pcanal: If I understand correctly, the default is: 
```
'[pre-commit.ci] auto fixes [...]'
```
The new message may (or may not) have lost the 'information' that some fixes were needed and have been automatically applied.

pcanal (approved):
LGTM.

.pre-commit-config.yaml
pcanal: Before
Screenshot 2025-10-22 at 19 34 46
and after
Screenshot 2025-10-22 at 19 39 08

sethrj: I think the second one does provide more information? The fact that's pre-commit.ci is obvious from the icon and the link. The commit message is what it does in the imperative-style commit message that is supposed to be git standard ("Apply pre-commit hooks") and then it goes on to tell you what to do (please install pre-commit locally). Do you find the second message more confusing? Is there a way to improve on it?

pcanal: I don't have a strong opinion and you can leave as is.  Here is a suggestion:
```suggestion
    Apply pre-commit hooks' fixes. (Please install pre-commit locally to avoid this commit!)
```

sethrj:
Thanks for the feedback @pcanal , I think the commit message and associated instructions are quite clear now 😄

PR #2067, opened 2025-10-22
Enable celeritas to build with CUDA 13

### LINKED ISSUE #2066: Celeritas doesn't build with CUDA 13
Because CUDA 13 moved some properties, like memory clock rate, out of the cudaDeviceProperties structure, celeritas won't compile the Device.cc file. I will submit a PR to fix this soon.

LSchwiebert:
Due to changes in where CUDA version 13 stores certain information, like the memory clock rate, these values need to be retrieved from a different location. (More information is available in the CUDA 13 release notes.) This patch allows the code to find this information when using CUDA 13 or later and retains the previous method with earlier versions of CUDA or with HIP.

The patch has been tested on CUDA 12.6 and CUDA 13 with gcc 11.4.0. All tests are passing, so this should be sufficient to enable celeritas to work with CUDA 13.


sethrj (approved):
Perfect! Thank you, and hopefully the next time I update the CI we'll have at least one CUDA 13 runner.

By the way, we generally do not require an issue corresponding to a pull request. I think our rule of thumb is to create one if it's going to require multiple PRs, or if it's an issue we can't work on immediately.

Issue #1058, opened 2023-12-10
Automate tricky input parameters


sethrj:
- [ ] Automatically resize track initializer vector when the capacity is exceeded (remove manual `initializer_capacity` input)
- [ ] Keep track of events so that `max_events` isn't needed
- [x] ... or possibly use "per stream events" so that we can just set max=1 for Geant4 integration

PR #2068, opened 2025-10-22
Replace thrust calls with cub towards reducing CPU synchronization

LSchwiebert:
CUDA [cub](https://docs.nvidia.com/cuda/cub/index.html) calls store all results on the GPU, unlike [Thrust](https://nvidia.github.io/cccl/thrust/) which returns some of the results to the CPU with a synchronous copy. This patch replaces the thrust calls with cub calls, so the results are available on the GPU, deferring synchronization with the CPU.

To minimize changes in this PR, the new code copies the GPU output, such as the number of vacancies, back to the CPU before the end of the action. **Follow-on work** will eliminate the synchronization by writing results such as `CoreStateCounters` directly to GPU  and refactoring the code that requires those results to use GPU memory or update asynchronously.

The code here produces results that match Thrust calls but using CUB calls. **The sort order changes from previous results** because the partition function in cub behaves differently from the thrust `stable_partition` function. The Thrust function puts the tracks in stable order for both the ones that are selected and the ones that aren't. The CUB function puts the tracks in stable order for the ones that are selected, but puts the ones that aren't selected in reverse order. It's consistent though—always reverse order.

CUB also requires two calls: the first determines the working memory needed for the CUB function, and the second performs the function using the asynchronously allocated memory. As long as the size of the inputs, such as number of tracks, don't change, this memory could be allocated once in a preprocessing step, reused for each call, and deallocated during a finalization step.


src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +54,73 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+    StreamT stream = s.get();
+
+    void* d_temp_storage = nullptr;
+    size_t temp_storage_bytes = 0;
+    // *** For testing with existing code for consistency
+    size_type* d_num_vacancies = nullptr;
+    // *** Allocate temporary storage
+    // *** For testing with existing code for consistency
+    d_num_vacancies
+        = static_cast<size_type*>(s.malloc_async(sizeof(size_type)));
```
sethrj: We have a wrapper for memory management and type safety that I think should be equivalent (@esseivaju added the async capability and can perhaps chime in)

```suggestion
    DeviceVector<size_type> num_vacancies{1, stream_id};
```

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +54,73 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
         thrust::make_counting_iterator<size_type>(0), TransformType{});
     auto result = device_pointer_cast(vacancies.data());
-    auto end = thrust::copy_if(thrust_execute_on(stream_id),
-                               start,
-                               start + vacancies.size(),
-                               device_pointer_cast(status.data()),
-                               result,
-                               IsVacant{});
+    auto flags = device_pointer_cast(status.data());
+    cub::DeviceSelect::FlaggedIf(d_temp_storage,
```
sethrj: Passing `nullptr` directly would improve the readability.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +54,73 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+                                 start,
+                                 flags,
+                                 result,
+                                 d_num_vacancies,
+                                 vacancies.size(),
+                                 IsVacant{},
+                                 stream);
+
+    // Allocate temporary storage
+    d_temp_storage = s.malloc_async(temp_storage_bytes);
```
sethrj: Since `temp_storage_bytes` is a stack value, doesn't this introduce a synchronization even if the previous kernel and the following allocation are async?

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Agreed. I'll change those calls.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Sounds good. As long as we know that the number of tracks (size of the vacancy array, not the total number of tracks) won't change during a run (or we know the maximum number of tracks in advance), it would be better to allocate and deallocate this working space storage just once instead of each time.

LSchwiebert: For device level cub calls, the documentation explicitly states that it is thread-safe to reuse this memory between calls. We would, of course, want to allocated different values for each call or if it can be done in the same place, allocate the maximum needed for the cub calls.

sethrj: We'd *like* to make the vacancy/secondary size flexible (https://github.com/celeritas-project/celeritas/issues/1058) but haven't yet. In our current implementation, we can strictly bound its maximum growth rate (factor of two, usually less), and we could grow it asynchronously as needed.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Good question. I hadn't thought about that before. It turns out that it doesn't call the kernel when the first parameter = nullptr. It just determines, on the cpu, how much memory is needed for the working space. It always requests at least one byte, even when it doesn't need any. Then it writes this value into the second parameter and returns without running the kernel. So, this variation of the function call doesn't run on the gpu.

To verify this, I looked into the cub open source. The project is archived, but I'm confident this aspect hasn't changed. Searching for the relevant Invoke function is the most straightforward way to verify this.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
amandalund: I think even when we support reallocating the initializer buffer, the vacancy size will stay constant since it's determined by the number of track slots.

LSchwiebert: Another option is to add a variable that tracks the maximum size. If the value becomes larger than that, we can recompute how much space is needed and reallocate if necessary. I think for many of the cub functions, the working space is independent of the problem size. I can provide a pointer to some other code where I did that, if it helps.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Thanks. I originally had the device_pointer_cast() and then changed it. I'll change them back.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
sethrj: Neat!

LSchwiebert:
Maybe this is the wrong place to ask, but I don't want to clutter Issues with questions that aren't useful. If you prefer just asking these sorts of things as Issues, I'll do that in the future.

Would there be value in specializing the DeviceVector class to handle type void? Basically, just pass the count (the first parameter) to DeviceAllocator as the allocation size instead of count * sizeof(T), since count * sizeof(void) == 0. It would allow straightforward use of DeviceVector<size_type> for the d_temp_storage parameter. Not sure if there would be other uses though, so maybe you prefer not to add it...

An alternative would be to test if sizeof(T) == 0, but that seems clunky to me.

If yes, I'm happy to create the PR.

sethrj:
I'm not sure I understand what a vector of void would do; you can't increment or dereference a void pointer right?

LSchwiebert:
Yeah. I just tried it and it won't build because it's not trivially-copyable. Obviously.

The cub functions are looking for a void pointer. I wonder if I can get it to work with just a char * instead. I'll try that...

sethrj:
@LSchwiebert Won't any pointer decay into a void pointer? I think you can just directly pass in a char*. Or does it need a pointer to a void pointer?

LSchwiebert:
It's complaining but I'll try a few things. I agree it should work.

LSchwiebert:
I think I figured out what the problem is. I need to pass the data() field, not the device vector.

sethrj:
Aha yep. Just lik you can't pass `std::vector<int>` into a function taking `int*`: you have to use `vec.data()`.

LSchwiebert:
I tried making it a vector of unsigned char, since I want a specific number of bytes, but it's complaining about a loss of precision in going from long int to short int or something to that effect. So, I'll just set it aside for now and focus on other things. I'm sure I can figure it out with some more thought. Anyway, thanks for the suggestions/feedback. I'm sure it's headed in the right direction now.

LSchwiebert:
The reason the CI is failing is that the hipcub header files aren't on the include search path. The hipcub package needs to be installed separately, just like other ROCm packages like hipthrust. Will add the necessary changes to this PR.

LSchwiebert:
@sethrj I changed the Dockerfile to add hipcub-dev as a requirement and I added hipcub to the external packages in the ubuntu-rocm6.yml file. Those are both in the docker dev folder, not the docker ci folder. Of course, it's still failing the test.

So, I think the dev builds should include hipcub now, but not sure how to get this working for the CI builds. It looks like it just uses the existing docker container, so might need to build a new docker container. If so, a couple comments: (1) The latest version of ROCm is 7.0.2 but there are no changes to hipcub between the current container ROCm version (6.3.1) and the latest release. (2) There is an old filename in the comment on line 47 of the build.sh script in the /scripts/docker folder.

BTW: I don't think it's urgent to resolve this, as the PR isn't finished yet.

sethrj:
Thanks @LSchwiebert ! I'll see about trying to update the docker scripts to the latest spack, vecgeom, etc. Then hopefully the cub addition will work 🤞

LSchwiebert:
If you want, I can open a PR (or issue if you prefer) for updating the ROCm docker image and move my changes there. And, if you want, update it to ROCm 7.0.2. I guess the PR should be for both docker images, but I'll just change that one.

LSchwiebert:
@sethrj You most likely want the ROCm docker image to be for ROCm 7.1.0, which was just released. To release thrust::stable_partition with cub library calls, we need DeviceTransform::transform, which is available starting with ROCm 7.1.0 and CUDA 12.9. Otherwise, we have to use the thrust::transform(), which won't be gpu resident, I'm afraid. (It might be interesting to profile it though.) The code should build correctly either way. I'll push the revised file today, even though I'm still confirming the results are correct.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -6,19 +6,35 @@
...
 #include "corecel/sys/Device.hh"
 #include "corecel/sys/ScopedProfiling.hh"
 #include "corecel/sys/Stream.hh"
 #include "corecel/sys/Thrust.device.hh"
 
+#if CELERITAS_USE_CUDA
+using namespace cub;
+#elif CELERITAS_USE_HIP
+using namespace hipcub;
+#endif
```
sethrj, line 39: Instead of `using namespace` (which obscures what functions in the file are in use), try to target the exact classes being used, or use a namespace alias (`namespace cub = hipcub;`)

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
 
     ScopedProfiling profile_this{"copy-if-vacant"};
 
+    // cub functions expect a cudaStream_t pointer for the stream
+    using StreamT = CELER_DEVICE_API_SYMBOL(Stream_t);
+    StreamT stream = device().stream(stream_id).get();
+
+    // *** For testing with existing code for consistency
+    DeviceVector<size_type> num_vacancies{1, stream_id};
+    // *** For testing with existing code for consistency
```
sethrj: Are the comments/code supposed to be left in?

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+                            vacancies.size(),
+                            IsVacant{},
+                            stream);
 
-    return end - result;
+    // Allocate temporary storage
+    DeviceAllocation temp_storage(temp_storage_bytes, stream_id);
+
+    DeviceSelect::FlaggedIf(temp_storage.data(),
+                            temp_storage_bytes,
```
sethrj: This is an "inout" (mutable reference) variable I suppose, so we can't pass an rvalue `.size()`?

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
-                               device_pointer_cast(status.data()),
-                               result,
-                               IsVacant{});
-    CELER_DEVICE_API_CALL(PeekAtLastError());
+    DeviceSelect::FlaggedIf(nullptr,
+                            temp_storage_bytes,
+                            start,
+                            flags,
+                            result,
+                            num_vacancies.data(),
```
sethrj: For the "sizing" kernel, do the data pointers have to be valid or can they be nullptr too?

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+                            vacancies.size(),
+                            IsVacant{},
+                            stream);
+#elif CELERITAS_USE_HIP
+    auto start = device_pointer_cast(status.data());
+    DeviceVector<unsigned char> flags{status.size(), stream_id};
+    auto data = device_pointer_cast(vacancies.data());
+#    if CELERITAS_USE_HIP && HIPCUB_VERSION >= 400100
```
sethrj: One thing we do to improve documentation for version-dependent bugs is to do at the top of the file (or where apporpriate)
```c++
#if CELERITAS_USE_HIP && HIPCUB_VERSION < 400100
#define CELER_BUGGY_HIPCUB_TRANSFORM 1
#else
#define CELER_BUGGY_HIPCUB_TRANSFORM 0
#endif
```

Then the code itself becomes the documentation:

```suggestion
#    if CELER_BUGGY_HIPCUB_TRANSFORM
```

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+    {
+        auto cub_error_code = DeviceTransform::Transform(
+            start, flags.data(), status.size(), IsVacant{}, stream);
+        // HIP is particular about checking return codes from hipCUB functions,
```
sethrj: What do you mean about "particular"?  Also I think it should always be OK to `CELER_DEVICE_API_CALL(PeekAtLastError())`: maybe call that once before the function (to make sure nothing else has queued an error), then just call `CELER_DEVICE_API_CALL(PeekAtLastError());` and don't worry about trying to catch the error code

src/celeritas/optical/action/detail/TrackInitAlgorithms.hh
```diff
@@ -28,7 +28,7 @@ using TrackStatusRef = StateCollection<TrackStatus, Ownership::reference, M>;
 //! Whether the track slot is vacant
 struct IsVacant
 {
-    CELER_FUNCTION bool operator()(TrackStatus status) const
+    CELER_FUNCTION bool operator()(TrackStatus const& status) const
```
sethrj: TrackStatus should be small and trivially copiable ; does the HIP API require this change?

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -6,46 +6,137 @@
...
 {
 //---------------------------------------------------------------------------//
 /*!
- * Remove all elements in the vacancy vector that were flagged as active
- * tracks.
+ * Create a functor to recognize specific tracks.
  */
+
+//
+//
```
sethrj:
```suggestion
```

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -6,46 +6,137 @@
...
+template<class T>
+struct NotEqual
+{
+    int compare_;
+
+    NotEqual<T>(int compare) : compare_(compare) {}
```
sethrj: The default constructor should do the same (unless you need to use `NotEqual(1)` instead of `NotEqual{}`
```suggestion
```

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -93,16 +209,99 @@ void partition_initializers(
...
     ScopedProfiling profile_this{"partition-initializers"};
 
-    // Partition the indices based on the track initializer charge
-    auto start = device_pointer_cast(init.indices.data());
-    auto end = start + count;
+    // Understandably, celeritas doesn't like creating zero-byte vectors. Since
+    // cub needs some vectors, trying to allocate these leads to a failed
+    // assertion. So, just return. No need to partition zero tracks.
+    if (count == 0)
+        return;
```
sethrj: To ensure consistency between host/device, it might be better to move the check to the caller's responsibility, and put a `CELER_EXPECT(count !=0)` conditional on the function.

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -93,16 +209,99 @@ void partition_initializers(
...
     auto stencil = static_cast<TrackInitializer*>(init.initializers.data())
                    + counters.num_initializers - count;
-    thrust::stable_partition(
-        thrust_execute_on(stream_id),
-        start,
-        end,
-        IsNeutralStencil{params.ptr<MemSpace::native>(), stencil});
+    DeviceVector<unsigned char> flags{count, stream_id};
+    // DeviceTransform added in cub 2.8/hipcub 4.1, else fall back to thrust
+#if CELERITAS_USE_CUDA && CUB_VERSION >= 200800
```
sethrj: I think a `CELER_CUB_HAS_TRANSFORM` would be good here too

test/celeritas/track/TrackSort.test.cc
```diff
@@ -660,24 +660,24 @@ TEST_F(PartitionDataTest, TEST_IF_CELER_DEVICE(step_device))
...
+            "_NNNNNN_NNNNNNNNNNNNNNNNN_NNNCNNNN_NNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNN_NNNNNN_NNNNNNNNNNNNNNCNNN_N_NNC_NNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "N_NNNNNNNNNNNNNNNNNNNNNNN_NNN__NNNCNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNNNNNNNNNNNN_NNNNNNNNNNNNNN_NNNNN_NNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NN_NNNNNNNNNNNNNNN_NNNNNN_NN_NNNNNNNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNNNNNN_NNNNNNNNNNNNNNNNNNNNNNNN_NNNN_CCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNNNNNNNNNNNNNN_NNNNNNNNNNNNNNNNNNNNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "N_NNNNNNNNNNNNNC_NNN_NNNNNNNNNNNNNNNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNNNNNNNN_NNNN_CNNNNNN_NNN_NNNNN_NNNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
+            "NNNNNN_NN_NNNNNCNNNNNN_NNN_N_N_NCNNNNNCCCCCCCCCCCCCCCCCCCCCCCCCC",
```
sethrj, line 680: Do you understand why these changed? Is it reproducible across the different implementations (HIP/CUB) and versions?

sethrj (commented):
Thanks @LSchwiebert ! A few comments, and also unfortunately since Frontier is still running ROCM 6, we might need to make sure we can support it. So I think I should add at least one of the older ROCm version builds back to the CI 🤔

amandalund:
I'm running into an error building this branch (CUDA 12.3):
```console
[39/275] Building CUDA object src/celeritas/CMakeFiles/celeritas_objects.dir/optical/action/detail/TrackInitAlgorithms.cu.o
FAILED: src/celeritas/CMakeFiles/celeritas_objects.dir/optical/action/detail/TrackInitAlgorithms.cu.o 
/soft/compilers/cuda/cuda-12.3.0/bin/nvcc -forward-unknown-to-host-compiler -DVECCORE_ENABLE_CUDA -I/home/alund/celeritas_project/celeritas/src -I/home/alund/celeritas_project/celeritas/build-reldeb/include -isystem /soft/compilers/cuda/cuda-12.3.0/targets/x86_64-linux/include -isystem /home/alund/spack/var/spack/environments/celeritas/.spack-env/view/include -isystem /home/alund/spack/opt/spack/linux-opensuse15-zen2/gcc-12.2.0/veccore-0.8.2-uadwij55dwz44cetb5e5ooyospfdkzl5/include -Werror all-warnings -O2 -g -DNDEBUG -std=c++17 "--generate-code=arch=compute_70,code=[compute_70,sm_70]" "--generate-code=arch=compute_80,code=[compute_80,sm_80]" "--generate-code=arch=compute_90,code=[compute_90,sm_90]" -Xcompiler=-fPIC -Xnvlink --suppress-stack-size-warning -Xcompiler= -MD -MT src/celeritas/CMakeFiles/celeritas_objects.dir/optical/action/detail/TrackInitAlgorithms.cu.o -MF src/celeritas/CMakeFiles/celeritas_objects.dir/optical/action/detail/TrackInitAlgorithms.cu.o.d -x cu -rdc=true -c /home/alund/celeritas_project/celeritas/src/celeritas/optical/action/detail/TrackInitAlgorithms.cu -o src/celeritas/CMakeFiles/celeritas_objects.dir/optical/action/detail/TrackInitAlgorithms.cu.o
/home/alund/celeritas_project/celeritas/src/celeritas/optical/action/detail/TrackInitAlgorithms.cu(88): error: class "cub::CUB_200200_700_800_900_NS::DeviceSelect" has no member "FlaggedIf"
      DeviceSelect::FlaggedIf(nullptr,
                    ^

/home/alund/celeritas_project/celeritas/src/celeritas/optical/action/detail/TrackInitAlgorithms.cu(101): error: class "cub::CUB_200200_700_800_900_NS::DeviceSelect" has no member "FlaggedIf"
      DeviceSelect::FlaggedIf(temp_storage.data(),
                    ^

2 errors detected in the compilation of "/home/alund/celeritas_project/celeritas/src/celeritas/optical/action/detail/TrackInitAlgorithms.cu".
```

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: I will remove those from this PR. They are there more as placeholders for the next step of having these counters on the GPU.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: I didn't think about using an alias. I'll do that. Otherwise, I have to have two versions of the code for every call. One for CUDA and one for ROCm, since they use different namespaces.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: I got an error message when I tried that. It said it had to be type size_t. It's only an input here (when not passing nullptr). I will try using .size() there, but not optimistic. I can also use a static_cast if you prefer that.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Maybe they can be nullptr for some. For instance, for the output variable like what's on line 93. For others, I think it may need an actual variable as the working space could be data-type dependent. I've never seen any cub code where they used different variables for the two calls, other than the nullptr on the first argument, though.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Sounds good. I'll change those.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Ahh. I could have been more direct here. It won't compile.

It gives two warnings that are escalated to errors. One for ignoring the returned variable on a [[nodiscard]] function and another for not using the return code. I didn't test it, but I think if I don't check the error code, it will likely still give me a -Wunused-variable error. I suppose another option is to tell the compiler to ignore these warnings. Note that cub doesn't complain about ignoring the returned variable.

src/celeritas/optical/action/detail/TrackInitAlgorithms.hh
LSchwiebert: I changed it when I was trying to get hipCUB to work, but that didn't help. Technically, CUDA cub says the argument should be a const reference, but then passes a non-const int value in the example. So, I will change it back.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Will do. Some tests will definitely fail though, because I found this only because some tests failed. However, it doesn't make a lot of sense to call these functions with no tracks. I believe it was being called with a count of zero at the start of the  event.

test/celeritas/track/TrackSort.test.cc
LSchwiebert: It's happening because the partition function in cub behaves differently from the thrust stable_partition function. The thrust function puts the tracks in stable order for both the ones that are selected and the ones that aren't. The cub function puts the tracks in stable order for the ones that are selected, but puts the ones that aren't selected in reverse order. It's consistent though -- always reverse order.

I didn't test with HIP, since I don't have access to that hardware, but I would expect that HIP would produce the same result. I think I tried CUDA 12.6 and 13.0, but if it didn't match, it would be a bug in the CUB implementation. I verified that the ordering matches between cub and thrust, modulo this difference. However, at some point the results start to diverge. I think that's due to the random decisions being made on different tracks, but I didn't verify that.

An option would be to sort the ones that aren't selected, but that seems like unnecessary extra computation.

LSchwiebert:
That means that CUDA 12.3 doesn't support that function. I'll figure out a work-around. At a minimum, I'll resolve this by not calling this function when using an older version of CUDA.

It turns out that this function was added in CUDA 12.6 and ROCm 6.4. It isn't working correctly in ROCm 7.1.0 anyway, at least it's not working as the cub implementation does. I'm going to change the CUDA function to use the same functions as the workaround for ROCm 7.1. That means that the DeviceTransform function will use thrust instead of cub, but there are only a few releases between when FlaggedIf was added (CUDA 12.6) and DeviceTransform was added (CUDA 12.8).

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: I'll add a default constructor. For this case, it needs to have the value of -1. It really needs to be the null value for a TrackSlotId, but I couldn't get that to work. Of course, I'm sure there's some way to do it.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
sethrj: No, if there's a version that sets the argument here, it's probably requiring an lvalue

src/celeritas/track/detail/TrackInitAlgorithms.cu
amandalund: Yeah, right now it's called during the "warm up" stage when no tracks are active:
https://github.com/celeritas-project/celeritas/blob/ae92d08fb1dcd94659887c1deee2360d12ba6376/src/celeritas/global/Stepper.cc#L101-L110

so we should probably skip the warmup here:
https://github.com/celeritas-project/celeritas/blob/ae92d08fb1dcd94659887c1deee2360d12ba6376/src/celeritas/track/InitializeTracksAction.cc#L64

test/celeritas/track/TrackSort.test.cc
sethrj: Perfect! I've added this documentation to the pull request description.

src/celeritas/track/detail/TrackInitAlgorithms.cu
sethrj: OK that's an important detail. Does it have to be an `int` instead of `TrackSlotId::size_type`? you can use `TrackSlotId{}.unchecked_get()` for the null value.

LSchwiebert:
I believe all the feedback has been addressed except for the code not working with CUDA 12.3. That change is pending.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
sethrj: Gotcha, I think instead of "particular" mentioning the `[[nodiscard]]` would be good. (I think it's bad practice for them to put nodiscard on a non-pure function that returns error codes, but that's just me 😉 )

There is also a `CELER_DISCARD` macro that's useful for this situation.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -53,18 +69,101 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+
+    // HIP is particular about checking return codes from hipCUB functions, so
+    // check for an error from the call and proceed accordingly
+    if (cub_error_code)
+        CELER_DEVICE_API_CALL(PeekAtLastError());
+#endif
+    // *** For testing with existing code for consistency
+    // *** Replace with the appropriate GPU counter
+    auto num = ItemCopier<size_type>{stream_id}(num_vacancies.data());
+    // *** For testing with existing code for consistency
```
sethrj: More testing lines here

sethrj (commented):
Thanks @LSchwiebert ! I see a couple of the earlier comments that might've gotten auto-hidden: there's some semi-duplicate code in `optical/...` that needs a refresh.

I also think we should ensure compatibility with Frontier's software stack, which I think is still 6.x.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
LSchwiebert: Sorry about that. I should have mentioned in my comment that I didn't push the updated version of that file because I'm still working on one part. When I made my previous comment, I wasn't thinking about these changes.

Once I get the #defines sorted out, the code should be compatible with 6.x, but it will still require some reasonable version of hipcub to build. CUDA started including cub in release 11, so I don't think we need to worry about that, but would you like me to add code to revert to the original thrust functions if hipcub is missing. (I think I can check that by testing if HIPCUB_VERSION is defined.)

sethrj: We can add a `CELERITAS_HAVE_HIPCUB` like `CELERITAS_HAVE_ROCTX`, defined in the top-level cmakelists. I'll see if we can test on JLSE/EXCL (where we're going to do performance profiling) to determine whether they have hipcub.

LSchwiebert: I'll add that to the cmakelist file. I'm thinking to also add something like this to the code as well in those two source files:

#ifdef CELER_USE_HIP && !CELERITAS_HAVE_HIPCUB
#define CELER_USE_THRUST 1
#else
#define CELER_USE_THRUST 0
#endif

That would handle the case of running CUDA or the case of running HIP without hipcub. Mentioning in case you have a cleaner solution in mind, rather than iterating over it.

sethrj: sounds perfect to me!

LSchwiebert: Using CELER_DISCARD is breaking the cub code. For example, in the global_Stepper test, when it tries to run the SiimpleComptonTest.device, it's failing with the message: internal assertion failed: ptr" thrown in the test body.

That's the place where it's trying to do an malloc_async() on line 50 of DeviceAllocation.cc

So, I'm reverting that for now. If you have any suggestions, happy to give them a try.

LSchwiebert:
@sethrj @amandalund I pushed all the files I've updated. The code should build, but the cub code is not generating correct output for the copy_if_vacant function in optical/action/detail/TrackInitAlgorithms.cu. You can check that your comments have been addressed, but please don't merge.

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -6,25 +6,69 @@
...
+#    include <hipcub/hipcub_version.hpp>
+#endif
+// DeviceTransform is unavailable in older versions of cub/hipcub, so fall back
+// to using thrust::transform instead
+#if CELERITAS_USE_CUDA && CUB_VERSION >= 200800
+#    define CELER_CUB_HAS_TRANSFORM 1
+#else
+#    define CELER_CUB_HAS_TRANSFORM 0
+#endif
+#if CELERITAS_USE_HIP && HIPCUB_VERSION >= 400100
```
sethrj: Some of these conditionals could be simplified a bit by using the fact that undefined macro symbols are set to zero in the preprocessor (or for clarity you could set them to zero yourself).

src/celeritas/optical/action/detail/TrackInitAlgorithms.cu
```diff
@@ -65,6 +109,54 @@ size_type copy_if_vacant(TrackStatusRef<MemSpace::device> const& status,
...
+
+    DeviceVector<size_type> num_vacancies{1, stream_id};
+
+    auto start = device_pointer_cast(status.data());
+    DeviceVector<unsigned char> flags{status.size(), stream_id};
+#    if CELER_CUB_HAS_TRANSFORM || CELER_HIPCUB_HAS_TRANSFORM
+    // HIP defines hipCUB functions as [[nodiscard]], but we defer error checks
+    {
+        auto cub_error_code = cub::DeviceTransform::Transform(
+            start, flags.data(), status.size(), IsVacant{}, stream);
```
sethrj: You may also need 
```suggestion
            start, flags.data(), status.size(), IsVacant{}, stream);
        CELER_DISCARD(cub_error_code);
```
if the compiler flags have `-Wno-unused`

CMakeLists.txt
```diff
@@ -340,10 +340,15 @@ elseif(CELERITAS_USE_HIP)
...
     )
   endif()
-  find_library(ROCTX_LIBRARY roctx64)
+  # Include the default ROCm installation path in the search
+  find_library(ROCTX_LIBRARY roctx64 PATHS /opt/rocm/lib)
   if(ROCTX_LIBRARY)
     set(CELERITAS_HAVE_ROCTX ON)
   endif()
+  find_file(HIPCUB hipcub_version.hpp PATHS /opt/rocm/include/hipcub)
+  if(NOT HIPCUB STREQUAL "HIPCUB-NOTFOUND")
```
sethrj: I think this should also work:
```suggestion
  if(HIPCUB)
```

Also this and the next two lines (and this change could be made for the ROCTX flag too!) could be:
```cmake
celeritas_set_onoff(CELERITAS_HAVE_HIPCUB ${HIPCUB})
```

sethrj (approved):
This is great @LSchwiebert ! The cleaned comments and backward compatibility are the icing on the cake. Maybe @amandalund could give it a quick look as well?

LSchwiebert:
@sethrj Please don't merge yet. I have to figure out why the current CUB implementation isn't producing correct results. So, it will fail some tests even though it builds. I just wanted to push everything so you could see that the other comments were addressed and get feedback on the rest.

sethrj:
@LSchwiebert I'm going to "convert to draft" since you don't want it to be merged yet; please hit the "ready to review" link once you've got the kinks ironed out! :)

LSchwiebert:
[like]  Loren Schwiebert reacted to your message:
________________________________

LSchwiebert:
@amandalund The cub commands are all working now, at least I think so...

amandalund:
Thanks @LSchwiebert, I'll take a look!

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -6,26 +6,75 @@
...
 {
+//---------------------------------------------------------------------------//
+/*!
+ * Whether the track slot is being used
+ */
+struct NotNull
+{
+    CELER_FUNCTION bool operator()(TrackSlotId a) const noexcept
+    {
+        return a.get() != TrackSlotId{}.unchecked_get();
```
amandalund:
```suggestion
        return static_cast<bool>(a);
```
is equivalent, and calling `a.get()` on an invalid track slot ID will trigger an assertion failure with the opaque message:
```consol
/home/alund/celeritas_project/celeritas/src/corecel/OpaqueId.hh:116:
celeritas: internal assertion failed: *this
```

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -6,26 +6,75 @@
 //---------------------------------------------------------------------------//
 #include "TrackInitAlgorithms.hh"
 
+// CUDA has included CUB since CUDA 11, but ROCm does not include hipCUB by
+// default, so test for the availability of hipCUB and use thrust instead if
+// it's unavailable. And some further checks for newer CUB/hipCUB functions.
+#if CELERITAS_USE_HIP && !CELERITAS_HAVE_HIPCUB
```
amandalund: I wonder if we should move some of this to a separate header file so it's easier to maintain since there's some duplication in the optical `TrackInitAlgorithms`

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -104,6 +217,77 @@ void partition_initializers(
...
+                      stencil + count,
+                      flags.data(),
+                      IsNeutral{params.ptr<MemSpace::native>()});
+#    endif
+    // CUB doesn't support in-place partitioning, so create a new variable,
+    // initial, of the same type and copy the current data in the init.indices
+    // object. Use initial for the input data and overwrite init.indices with
+    // the partitioned data, as expected from an in-place algorithm.
+    StateCollection<size_type, Ownership::value, MemSpace::device> initial;
+    initial = init.indices;
```
amandalund: I think we could also use a thrust `counting_iterator` here since at this point the indices are always sequential from zero.

amandalund (approved):
This is really nice work, thanks @LSchwiebert!

Curious if you've done any performance comparisons with the updated algorithms: I did a quick test on a simple benchmark and saw a couple percent slowdown overall, which looks like it's coming from the track partitioning (~10% slowdown in the initialize tracks action). Not a big deal since it's relatively small and ultimately this should improve performance by reducing synchronization, but it might still be helpful to identify/understand any changes at this intermediate stage.

I think we'll also need to revert [d1cd6f0](https://github.com/celeritas-project/celeritas/pull/2068/commits/d1cd6f03ebc63b9aa597840548dfa6cfcb307fdd).

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Yes. That would make sense. How about corecel/Macros.hh? The only tricky part is we have to load the CUB or hipCUB version header file to determine what version we have for some of the checks, but if it's OK to put that there (after checking we are using CUDA/HIP), I'll go ahead.

LSchwiebert:
I didn't do any performance analysis, but I agree it would be a good idea to do that at this point to have a benchmark going forward and also to understand how it's performing with these changes. What would be some good tests to use for this purpose? Or, more generally, how do you recommend doing the performance testing? What version of CUDA did you use for your performance runs?

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Nice! That may have been the reason for the performance degradation.

amandalund: I reran with the updates and it looks like this did bring the time back down!

amandalund:
We have a suite of test problems that we regularly use for analyzing performance (https://github.com/celeritas-project/regression), but for right now that would probably be overkill. I was using CUDA 12.3: looks like we're back on track in terms of performance anyway, but if you wanted to do a sanity check with a newer CUDA this is the simple benchmark I used:

```json
{
"geometry_file": "/home/alund/celeritas_project/celeritas/test/geocel/data/testem3-flat.gdml",
"primary_options": {
  "seed": 0,
  "direction": [1, 0, 0],
  "energy": 10000,
  "num_events": 16,
  "pdg": 11,
  "position": [-22, 0, 0],
  "primaries_per_event": 1000
},
"physics_options": {
  "coulomb_scattering": false,
  "eloss_fluctuation": false,
  "em_bins_per_decade": 56,
  "lpm": true,
  "msc": "urban",
  "physics": "em_basic"
},
"field": [0.0, 0.0, 0.0],
"num_track_slots": 1048576,
"initializer_capacity": 16777216,
"seed": 20220904,
"merge_events": true,
"write_track_counts": false,
"write_step_times": false,
"track_order": "init_charge",
"action_times": false,
"use_device": true
}
```

Run with `./bin/celer-sim inp.json > out.json`. Setting `track_order` to `init_charge` enables the track partitioning, and `none` disables it. Enabling `action_times` will synchronize the stream between kernel launches and output accumulated timing results for each action.

LSchwiebert:
Thanks for the info. I'll give it a try as I'm curious about how newer CUDA versions will perform, since they are using different CUB functions.

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -46,6 +100,42 @@ size_type remove_if_alive(
...
+    // Run selection
+    cub_error_code = cub::DeviceSelect::If(temp_storage.data(),
+                                           temp_storage_bytes,
+                                           data,
+                                           num_not_active.data(),
+                                           vacancies.size(),
+                                           NotNull{},
+                                           stream);
+    CELER_DISCARD(cub_error_code);
+    auto num = ItemCopier<size_type>{stream_id}(num_not_active.data());
```
esseivaju: The copy here is asynchronous. We're not copying back to pinned memory, so cuda might execute the copy sycnhronously, but shouldn't we still have a stream synchronize here to guarantee that the host isn't accessing the memory before the device transfer is done?

LSchwiebert: You mean executing the `return num;` statement before the code on line on 134 completes? Looking at it carefully, I notice some inconsistency in the code, including for the original thrust calls. For instance, see the thrust code here. Would just returning the value instead of assigning it to a variable resolve this? I don't think that resolves your concern though.

Maybe I should mention that this is just a stopgap solution. The long term plan is to keep these counters on the GPU until the end of the step, instead of copying them back to the CPU at the end of each function. Of the other hand, we don't want to have a race condition in the code until that's ready.

```
    auto stop = thrust::exclusive_scan(thrust_execute_on(stream_id),
                                       data,
                                       data + counts.size(),
                                       data,
                                       size_type(0));
    CELER_DEVICE_API_CALL(PeekAtLastError());

    // Copy the last element (accumulated total) back to host
    return ItemCopier<size_type>{stream_id}(stop.get() - 1);
```

esseivaju: I was thinking this:
```c++
    auto num = ItemCopier<size_type>{stream_id}(num_not_active.data());

    CELER_DEVICE_API_CALL(PeekAtLastError());
    CELER_DEVICE_API_CALL(StreamSynchronize(stream));
    return num;
```
PeekAtLastError would check the return code of the async call, but I don't think it will force any synchronization and wait on the asynchronous operation to complete. 

So, the explicit stream synchronization should guarantee that cuda will complete the async memcpy before returning the value to the caller. Without that synchronization, I don't think that cuda provides any guarantees on the memory being ready.

LSchwiebert: OK. Got your point now. Sorry for missing it before. Anyway, thanks for finding this. I'm changing it from:

```
    auto num = ItemCopier<size_type>{stream_id}(num_not_active.data());

    CELER_DEVICE_API_CALL(PeekAtLastError());
    return num;

```
to

```
    CELER_DEVICE_API_CALL(PeekAtLastError());

    return ItemCopier<size_type>{stream_id}(num_not_active.data());

```

The only reason I was assigning to a variable was to check the value for debugging and I forgot to change it back. This is how the code works elsewhere, but we can discuss further if you think it's adequate resolve your concern.

amandalund: @LSchwiebert you're right that we were already doing this elsewhere in the code, but I don't think that was correct. We should update the code here and in the other places you pointed out to explicitly synchronize the stream between copying and returning the value as @esseivaju suggests.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: I've added all the #defines for the CUB/hipCUB libraries to corecel/Macros.hh and streamlined the code accordingly as much as I could.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Understood. I changed these to what @esseivaju suggested for both the new CUB/hipCUB calls and the original thrust calls.

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -74,7 +154,43 @@ size_type exclusive_scan_counts(
     CELER_DEVICE_API_CALL(PeekAtLastError());
 
     // Copy the last element (accumulated total) back to host
-    return ItemCopier<size_type>{stream_id}(stop.get() - 1);
+    auto result = ItemCopier<size_type>{stream_id}(stop.get() - 1);
+
+    // CUDA/HIP functions expect a cudaStream_t pointer for the stream
+    using StreamT = CELER_DEVICE_API_SYMBOL(Stream_t);
+    StreamT stream = device().stream(stream_id).get();
+    CELER_DEVICE_API_CALL(StreamSynchronize(stream));
```
amandalund: To simplify things a bit, you could add above line 146:
```c++
auto stream = device().stream(stream_id);
```
Then this becomes:
```suggestion
    stream.sync();
```
and similarly below you can use `stream.get()` and `stream.sync()` where needed.

LSchwiebert: Sounds good. Updated with this change.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Yes. This is much cleaner.

amandalund (approved):
Thanks again @LSchwiebert! @sethrj do you want to take a final look?

sethrj:
Sounds like you have all done a very thorough check so I'm good to merge! Thanks everyone!! 🎉

sethrj:
@LSchwiebert the last thing I think is to ensure all the comments are resolved; could you double check that none are left open accidentally?

LSchwiebert:
I double checked all of the pending conversations. I made all the requested changes, so I updated them to Resolved. I'm going to go ahead and merge. Thanks everyone for a thorough review. The PR is significantly better than before the feedback.

PR #2069, opened 2025-10-22
Define a GeoTrackInterface class for documentation and testing

sethrj:
To test the upcoming volume reconstruction in #2057, it would be nice to build a test once and run it for all geometries, using accessors like "volume depth".

To that end, this PR adds a new abstract base class for wrapping geometry track views that is used as the primary source for documentation. A concrete templated class for each geometry track view `WrappedGeoTrackView` lives inside the `test` directory so that unit tests can be compiled  once and run on any geometry type. Moreover, the `CheckedGeoTrackView` now also has the same interface, so we can add additional tests and checking without overhead and ensure uniform usage of the geometry track views across the code base.

I used the interface to refactor the `test_detailed_tracking` tests to be compiled into the .cc file using the track interface. I would've switched them to the checked view, but there's a failure in Geant4 for one case that needs further investigation.

This will cause more conflicts with #1720.


test/geocel/GenericGeoTestInterface.cc
```diff
@@ -8,23 +8,285 @@
...
+        CELER_LOG(warning) << "Surface normal checking is disabled for "
+                           << this->gdml_basename() << " using "
+                           << this->geometry_type();
+        result.disable_surface_normal();
+    }
+
+#if 0
+    // FIXME: Geant4 solidstest fails at one point
+    CheckedGeoTrackView geo{this->make_geo_track_view_interface()};
+    geo.check_normal(check_surface_normal);
```
sethrj, line 56: This is the only real to-do item from this PR.

test/geocel/CheckedGeoTrackView.cc
```diff
@@ -0,0 +1,227 @@
...
+
+        // Check for tangent crossing warning
+        if (soft_zero(dot_product(t_->dir(), post_norm)))
+        {
+            CELER_LOG(warning)
+                << "Crossed into " << t_->impl_volume_id().get()
+                << " at a tangent; traveling along " << repr(t_->dir())
+                << ", normal is " << repr(post_norm);
+        }
+    }
```
sethrj, line 222: This check was moved from the GenericGeo `track` function. The one difference is that since the CheckedGeoTrackView doesn't have a reference to the geo/volume params (and since it's CPU only we could actually do that) we print impl volume IDs rather than volume names if it fails.

test/geocel/GenericGeoTestInterface.hh
```diff
@@ -32,54 +34,80 @@ struct GenericGeoVolumeStackResult;
...
 
     //! Get the label for this geometry: Geant4, VecGeom, ORANGE
     virtual std::string_view geometry_type() const = 0;
 
     //! Access the geometry interface
     virtual GeoParamsInterface const& geometry_interface() const = 0;
 
-    // Get the basename or unique geometry key
-    virtual std::string_view gdml_basename() const = 0;
+    //! Create a track view (TODO: replace geo test base view)
```
sethrj, line 72: I'm not sure what this to-do means 😅 I'll remove it

test/geocel/GenericGeoTestInterface.hh
```diff
@@ -32,54 +34,80 @@ struct GenericGeoVolumeStackResult;
...
-    //! Generate a track
-    virtual TrackingResult track(Real3 const& pos_cm, Real3 const& dir) = 0;
+    //// TESTS ////
 
-    //! Get the safety tolerance (defaults to SoftEq tol) for tracking result
-    virtual GenericGeoTrackingTolerance tracking_tol() const;
+    // Track until exiting the geometry
+    TrackingResult track(Real3 const& pos_cm, Real3 const& dir);
+    // Obtain the "touchable history" at a point
+    VolumeStackResult volume_stack(Real3 const& pos_cm);
```
sethrj, line 57: These used to be in the derived class but since they're no longer templated (they call the wrapped "make geo track view") I moved them here.

test/celeritas/field/LinearPropagator.test.cc
```diff
@@ -25,35 +25,19 @@ namespace test
 // TEST HARNESS
 //---------------------------------------------------------------------------//
 
-template<class HP>
-class LinearPropagatorTest : public AllGeoTypedTestBase<HP>
+class LinearPropagatorTest : public CoreGeoTestBase
```
sethrj, line 28: This change means that we are only testing one geometry type per build, rather than all three. I think that can be relaxed later but is fine for the time being.

elliottbiondo:
@sethrj is this still ready to review?

sethrj:
Yessir thank you!

test/geocel/CheckedTrackInterface.hh
```diff
@@ -0,0 +1,40 @@
...
+namespace test
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Interface for accessing checking/instrumentation data from track views.
+ *
+ * This interface provides access to diagnostic counters and utilities for
+ * track views that perform validation and checking during navigation.
+ */
+class CheckedTrackInterface
```
amandalund: Where is this used?

amandalund (approved):
Looks good to me!

test/geocel/CheckedTrackInterface.hh
sethrj: Good catch! It is not used at all.

PR #2071, opened 2025-10-23
Fix opaque ID arithmetic for small integers

sethrj:
Integer promotion causes subtraction of small integers to become larger integers, which then can't be implicitly cast to small integers again.


pcanal (commented):
LGTM

amandalund (approved):
Thanks!

PR #2072, opened 2025-10-23
Use short integers for DepthId, UnivDepthId

sethrj:


sethrj:
I'll reopen this when we have a chance to test for performance.

PR #2073, opened 2025-10-23
Re-enable code coverage 

pcanal:
By actually using the file generated by gcovr ...  Without this file, there is basically no usage report.

Also stop displaying information about the `app` and `test` directory as they are not meant to be 100% covered.  In the test the failing code path are by definition not taken ... It is more debatable for the app .. their coverage would depend on whether we have enough tests to cover all their configurations.


.github/codecov.yml
```diff
@@ -37,6 +37,8 @@ comment:
   require_changes: false
 
 ignore:
+  - "test"
+  - "app"
```
sethrj, line 41: Ignoring the test directory is fine because there are so many false positives (though eliminating dead code would be nice, and we have a lot of complex test harnesses). The app directory, however, is something we should definitely test for coverage: hopefully as we consolidate code into the main libraries there will be less to test.

sethrj (approved):
I'm getting 400 errors from codecov but if you can check it, and re-enable app, I'd say this is G2G

pcanal:
> I'm getting 400 errors from codecov 

What do you mean? (I don't see/find any errors)

pcanal:
> and re-enable app, 

I would/will re-enable the app in the PR that make it take into account running the installed apps.

sethrj:
Sorry @pcanal that was very unclear from me. I meant that I'm unable to view the errors on codecov due to server error codes: HTTP error code 400.

pcanal:
Since the codecov web error are unrelated to this PR, I will merge it to be able to renable the coverage reports and move forward.

.github/codecov.yml
pcanal: Since the codecov web error are unrelated to this PR, I will merge it to be able to renable the coverage reports and move forward.

`app` will be re-added when they are fully taken into account.

PR #2078, opened 2025-10-27
Add data() getter functions for the DeviceAllocation class

LSchwiebert:
This PR allows access to the DeviceAllocation class data member in the same way that the DeviceVector class does. This is useful for low-level device memory allocations and access, such as what CUDA cub needs for working memory.


sethrj (approved):
Thanks!

Issue #2079, opened 2025-10-28
Fix reentrant multiple boundary crossings


sethrj:
Originally, ORANGE assumed straight-line propagation and did not account for direction changes on boundaries. However, both field propagation *and* multiple scattering can change the direction so that the track is reentrant *after a geometry-limited step* (aka after a call to `move_to_boundary` that sets up a subsequent `cross_boundary` call in the BoundaryAction kernel).

The workaround in place is for the ORANGE track view to store an extra `BoundaryResult boundary` state that is reentrant/exiting. When reentrant, `find_next_step` returns zero distance (I'm not sure where this is needed or if it's correct???) and the call to `cross_boundary` merely resets the `boundary` flag to "exiting".

*However* something is broken whenever we cross into a new volume and *then* change the direction back into the original. This is causing errors for #1512 . Instead of changing volumes when `set_dir` is called to move back into the original volume, no crossing happens.

Some thoughts:
- May need to distinguish `exiting` from `exited`
- Should eliminate `geo_normal` that changes sign based on underlying surface sense
- Remember that multiple `set_dir` may be called between `move_to_boundary` and `cross_boundary`, and same *after* `cross_boundary`
- May want to check consistency between normal direction and surface sense (and consider edge cases?)
- Maybe this reentrant capability should sit at a higher level than the geo crossing since it's only applicable to some particle types/use cases?

I'll wait on #2069 to start debugging so that I can take advantage of the multiple-geometry comparisons.

sethrj:
Fixed by #2081

PR #2081, opened 2025-10-28
Fix reentrant boundary crossing via surface physics

hhollenb:
To address #2079 I changed the `BoundaryResult` in ORANGE to be more symmetric with respect to whether a track is moving into or off of a surface (name subject to change, just wrote the first thing that came to mind). 

Whenever a track is on a boundary, its direction determines whether it is `moving_into` or `moving_off` of the boundary. Calling `move_to_boundary` will automatically set the boundary result to be `moving_into`. A call to `set_dir` while on the boundary will flip `into` / `off` depending on relative sign change of the dot product with the normal (same as previous behavior). Calling `cross_boundary` then swaps the sense only if the track is `moving_into` the boundary and does nothing if `moving_off`.

The main difference from the previous behavior is that repeated calls to `cross_boundary` won't change anything. They need to be interspersed with `set_dir` going into the opposite volume, i.e. `cross_boundary` moves into the volume along the track's current direction.


test/geocel/GeoTests.cc
```diff
@@ -403,8 +397,10 @@ void FourLevelsGeoTest::test_detailed_tracking() const
         else if (test_->geometry_type() == "ORANGE")
         {
             // Should be able to relocate back and forth
+            geo.set_dir(Real3{1, 0, 0});
             geo.cross_boundary();
             EXPECT_EQ("Shape1", test_->volume_name(geo));
+            geo.set_dir(Real3{-1, 0, 0});
```
sethrj, line 414: I think this reveals the conceptual problem that cropped up. Originally my intent was "cross boundary -> flip surface sign and relocate" so it's purely logical. Originally we had "move to boundary -> next thing will be cross boundary". But with MSC and field propagation, we had to be able to detect when a change in direction meant you actually weren't going to cross the boundary after all. And to detect that I added the code inside `set_dir` to ignore the next `cross_boundary` (or to redo it if the direction crosses the surface normal *again*). Now with surface physics we have to do it but while initially heading out the other side.

I think the way we've changed it to now, toward vs away, is correct. When crossing a boundary without changing direction on it, we don't have to do special checking of the direction or surface normal. (This applies to gammas.) But when we do have to know whether we're truly exiting, the normal calculation will tell us.

src/orange/OrangeTrackView.hh
```diff
@@ -689,17 +690,17 @@ CELER_FUNCTION void OrangeTrackView::cross_boundary()
     CELER_EXPECT(this->is_on_boundary());
     CELER_EXPECT(!this->has_next_step());
 
-    if (CELER_UNLIKELY(this->boundary() == BoundaryResult::reentrant))
+    if (CELER_UNLIKELY(this->boundary() == BoundaryResult::moving_off))
     {
         // Direction changed while on boundary leading to no change in
         // volume/surface. This is logically equivalent to a reflection.
-        this->boundary(BoundaryResult::exiting);
+        // this->boundary(BoundaryResult::exiting);
```
sethrj: I think we are missing a test case for this: field can change from exiting to reentering, then msc can change from reentering back to exiting. We can change direction relative to the normal multiple times on a boundary.

src/orange/OrangeTypes.hh
```diff
@@ -226,8 +226,8 @@ enum class SurfaceState : bool
  */
 enum class BoundaryResult : bool
 {
-    reentrant = false,
-    exiting = true
+    moving_off = false,
+    moving_into = true
```
sethrj: If you look at the documentation above, the enum name and values are not being used for their intended purpose. The idea was to return this as part of a propagation or set_dir or something, and then downstream logic would determine whether or not to call cross_boundary. That may not be possible because we can do multiple direction changes on the boundary (field causes a track to re-enter, then msc direction change means it leaves again).

I'm not quite sure of the semantics you're trying to describe: "on and exiting, on and entering" while making the default "off and entering" equal to "on and exiting"?

test/geocel/GeoTests.cc
```diff
@@ -403,8 +397,10 @@ void FourLevelsGeoTest::test_detailed_tracking() const
         else if (test_->geometry_type() == "ORANGE")
```
sethrj: If we're changing the direction now, maybe vecgeom will also work? So let's see what happens when you delete the `if` here.

sethrj (commented):
I think this is on the right track but am not quite sure of the distinction between `moving_off/moving_into` versus `exiting/entering`. 😕

test/geocel/GeoTests.cc
hhollenb: My question then would be what's responsible for checking direction before crossing a boundary. My understanding is that `cross_boundary` is always called after the along step actions which makes the geo track view responsible for storing directional state relative to the surface normal and not crossing when appropriate. We could instead make the boundary action responsible for checking along step results and not crossing, which would simplify the geometry logic.

sethrj: Yeah, I actually think that would be better 🤔 but let's make that a follow-on to this patch so that you can get back to physics validation. Can you add a test for the double-change below, update the names/docs, and then I'll subsequently try to improve the directional treatment?

sethrj:
@hhollenb I saw a couple of my comments from the last review didn't get addressed/answered/resolved; did you see them/want to discuss?

sethrj (approved):
Thanks @hhollenb !

Issue #1967, opened 2025-09-17
Asynchronously copy primaries on push


sethrj:
The primary push mechanism with the stepper takes a Span, resizes a collection, and copies (with a stream, but still usually synchronously since the span points to unpinned memory) primaries to the GPU. This causes a substantial slowdown when generating lots of primaries from Geant4 that have relatively low energy (see 
 https://github.com/celeritas-project/celeritas/discussions/1941 ). Maybe we want a double buffer/ring buffer that uses pinned memory with head/tail pointers, copies a chunk at a time (perhaps uses `cudaLaunchHostFunc` to update a host pointer to the last successfully copied block after the async copy), and blocks if head == tail + 1 (buffer full). We could make this class reusable for primaries and hits.

amandalund:
The optical offload will have a similar push mechanism for copying generator distribution data to the GPU.

PR #2082, opened 2025-10-28
Refactor and extend optical setup and generation for standalone loop support

amandalund:
This makes some changes to the optical code that will be needed to run a standalone optical stepping loop either using a primary generator or through the accel offloading interface:

- Construction of the optical core params is separated from the optical collector construction in the problem setup
- Optical params are made accessible through `SharedParams`
- Cherenkov and scintillation params are made accessible through optical params
- An `inp` option is added for the optical generation mechanism
- An `insert` method is added to the generator action to copy distribution data to device
- The optical primary generator test is simplified and expanded to include the `GeneratorAction`


src/celeritas/optical/CoreParams.cc
```diff
@@ -50,6 +52,15 @@ build_params_refs(CoreParams::Input const& p, CoreScalars const& scalars)
     ref.surface = get_ref<M>(*p.surface);
     ref.surface_physics = get_ref<M>(*p.surface_physics);
     ref.rng = get_ref<M>(*p.rng);
+    // TODO: Get detectors ref
```
sethrj, line 55: Good catch 😅

src/celeritas/optical/Transporter.cc
```diff
@@ -38,18 +38,17 @@ Transporter::Transporter(Input&& inp)
...
-}
-
-//---------------------------------------------------------------------------//
-/*!
- * Transport all pending optical tracks on the device.
- */
-void Transporter::operator()(CoreStateDevice& state) const
-{
-    this->transport_impl(state);
+    if (auto* s = dynamic_cast<CoreStateHost*>(&state))
```
sethrj, line 43: What's the reason for this change? Since we know the derived type in most cases (at least in the optical offload we certainly do) this means downcasting and upcasting again, violating (to an admittedly tiny extent) the "only pay for what you use" principle. Is this because expect to be launching GPU states from a CPU celeritas or vice versa?

src/celeritas/optical/gen/GeneratorAction.cc
```diff
@@ -184,24 +229,25 @@ void GeneratorAction::step_impl(CoreParams const& params,
...
     {
-        auto cherenkov = data_.cherenkov ? data_.cherenkov->host_ref()
-                                         : HostCRef<CherenkovData>{};
-        auto scint = data_.scintillation ? data_.scintillation->host_ref()
-                                         : HostCRef<ScintillationData>{};
+        auto cherenkov = params.cherenkov() ? params.cherenkov()->host_ref()
+                                            : HostCRef<CherenkovData>{};
+        auto scintillation = params.scintillation()
+                                 ? params.scintillation()->host_ref()
+                                 : HostCRef<ScintillationData>{};
```
sethrj: Since the `optical::CoreParamsData` now has cherenkov/scintillation, and they're empty rather than null pointers if disabled, I think you could just do `params.ref<MemSpace::native>().cherenkov()` etc. here and in the .cu file.

src/celeritas/optical/gen/GeneratorAction.cc
```diff
@@ -105,11 +103,28 @@ auto GeneratorAction::create_state(MemSpace m, StreamId id, size_type) const
...
+        return make_state<MemSpace::device>(id, initial_capacity_);
+    }
+    CELER_ASSERT_UNREACHABLE();
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Add user-provided host distribution data.
+ */
+void GeneratorAction::insert(CoreStateBase& state, SpanConstData data) const
```
sethrj, line 119: We'll have to add a note to #1967 to revisit for this case too.

src/celeritas/optical/gen/GeneratorData.hh
```diff
@@ -152,6 +155,14 @@ struct GeneratorState : public GeneratorStateBase
 {
     CollectionStateStore<GeneratorStateData, M> store;
 
+    //! Access valid range of distributions
+    auto distributions()
+    {
+        return this->store.ref()
+            .distributions[ItemRange<GeneratorDistributionData>(
+                ItemId<GeneratorDistributionData>(this->counters.buffer_size))];
```
sethrj, line 163: A comment about the return type would be helpful: I guess it's auto because it's a span for host data, an LdgSpan for device?

sethrj (approved):
Excellent work as always! Just a couple small points.

src/celeritas/optical/gen/GeneratorAction.cc
amandalund: Oh good point, that simplifies things!

src/celeritas/optical/gen/GeneratorData.hh
amandalund: It's auto because I copied it from here :)

https://github.com/celeritas-project/celeritas/blob/aa12691062218ff278315706f747c6af428e4f7d/src/celeritas/track/ExtendFromPrimariesAction.hh#L111-L115

src/celeritas/optical/Transporter.cc
amandalund: Right now the accel optical offload class I'm working on has a `std::shared_ptr<optical::CoreStateBase>` (like the generator test), unlike the optical launch which is a stepping action. But that class may evolve (and alternatively we could downcast it in `Flush`)...

src/celeritas/optical/gen/GeneratorData.hh
sethrj: Mea culpa 😅

src/celeritas/optical/Transporter.cc
sethrj: One dynamic cast per transport is negligible so if it makes the downstream code easier, then do it; it's just kind of different from whatever else we've done before.

amandalund: Ok I'll leave it as is for now, and we can reevaluate when we add the local offload and standalone Celeritas support.

hhollenb (approved):
Looks good, great work!

PR #2083, opened 2025-10-29
Fix run manager construction in integration test base

amandalund:
I just noticed the integration unit tests were failing with Geant4 v10:

```console
C++ exception with description "/home/runner/work/celeritas/celeritas/test/accel/IntegrationTestBase.cc:294:
celeritas: internal assertion failed: rm" thrown in the test body.
```

This wasn't obvious in the CI since we continue on error for older Geant4 versions.


sethrj (approved):
Oops... maybe now that we actually have some integration tests, we should update the CI to only test `accel/` and require the tests to pass.

PR #2085, opened 2025-10-29
Fix construction of CMS polyhedra with discontinuities

### LINKED ISSUE #2046: Fix and implement last couple of missing ORANGE shape conversions

elliottbiondo:
This MR enacts two changes to get the CMS polyhedra cases to pass:

1) `StackedExtrudedPolygon` is modified so that, when extruded along the polyline, polyline segments with zero length are simply skipped. This allows for the creation of shapes where the `r(z)` shape of a polyhedron is discontinuous.

2) In `ExtrudedPolygon` the tolerance passed to `filter_collinear_points` is changed from the `BumpCalculator` tolerance, which is both absolute and relative, to a pure absolute tolerance. In the CMS test, a 36-gon has some (probably 18) of its points filtered out because using `BumpCalculator` with its large size created a tolerance that was far too aggressive. Note that we are already using a pure absolute tolerance for `RevolvedPolygon`.

Additionally, `G4Polyhedra` allows for cases where, the inner radius is only sometimes zero, creating discontinuous hollow regions. I added support to SEP for cases that look like:

```
    __________
   |__________|
        __
       |__|
```

But not cases where the radius is zero on exactly one end, i.e. it comes to a point:

```
    __________
   |__________|
        _
       /_\
```

This latter cases is not actually much harder. For now I added a DBC to catch it. I can do it for this MR, but I kind of feel like we should see it the wild before bothering to implement.


src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -93,14 +93,14 @@ StackedExtrudedPolygon::StackedExtrudedPolygon(std::string&& label,
...
+    // Validate that z coordinates are weakly increasing
     CELER_VALIDATE(std::adjacent_find(polyline_.begin(),
                                       polyline_.end(),
                                       [](auto const& a, auto const& b) {
-                                          return a[Z] >= b[Z];
+                                          return a[Z] > b[Z];
                                       })
                        == polyline_.end(),
-                   << "z coordinates must be strictly increasing");
+                   << "z coordinates must be weakly increasing");
```
sethrj: Is that common nomenclature? I would just write "nondecreasing".

sethrj (approved):

src/orange/orangeinp/StackedExtrudedPolygon.cc
elliottbiondo: Yes "weakly" and "strictly" are regularly used terms for the monotonicity of functions, but I agree "nondescreasing"is more clear.

sethrj: I've heard and used "strictly" before but not "weakly".

elliottbiondo:
This is still draft until I 1) write a good SEP test, 2) figure out the other CMS issue

src/orange/orangeinp/StackedExtrudedPolygon.cc
elliottbiondo: FWIW:

"To avoid ambiguity, the terms weakly monotone, weakly increasing and weakly decreasing are often used to refer to non-strict monotonicity."

https://en.wikipedia.org/wiki/Monotonic_function

elliottbiondo:
@sethrj RTR

src/orange/orangeinp/StackedExtrudedPolygon.cc
sethrj: Yeah I saw that, wasn't doubting you 😄

src/orange/orangeinp/IntersectRegion.cc
```diff
@@ -666,13 +666,10 @@ ExtrudedPolygon::ExtrudedPolygon(ExtrudedPolygon::VecReal2 const& polygon,
...
 
-    // Store only non-collinear points
-    Real3 const extents{
-        x_range_[1] - x_range_[0], y_range_[1] - y_range_[0], 0};
-    real_type abs_tol = ::celeritas::detail::BumpCalculator(
-        Tolerance<>::from_default())(extents);
-
-    polygon_ = detail::filter_collinear_points(polygon, abs_tol);
+    // Store only non-collinear points. Use an absolute tolerance; otherwise,
+    // for example, an arbitrilarly large regular dodecagon becomes a hexagon
```
sethrj:
```suggestion
    // for example, an arbitrarily large regular dodecagon becomes a hexagon
```
guess codespell doesn't catch them all 😉

src/orange/orangeinp/StackedExtrudedPolygon.cc
```diff
@@ -187,18 +187,30 @@ StackedExtrudedPolygon::make_stack(detail::VolumeBuilder& vb,
...
     {
-        // Create the ExtrudedPolygon for this segment
-        ExtrudedPolygon shape{polygon,
-                              {polyline_[i], scaling_[i]},
-                              {polyline_[i + 1], scaling_[i + 1]}};
-
-        // Build this segment with unique label
-        nodes[i] = build_intersect_region(
-            vb, label_, this->make_segment_ext(si, i), shape);
+        CELER_VALIDATE(soft_zero(scaling_[0]) == soft_zero(scaling_[1])
```
sethrj: Should this be 
```suggestion
        CELER_VALIDATE(soft_zero(scaling_[i]) == soft_zero(scaling_[i + 1])
```

test/orange/g4org/SolidConverter.test.cc
```diff
@@ -826,6 +826,25 @@ TEST_F(SolidConverterTest, polyhedra)
...
+        static double const rmax[] = {2, 2, 2, 2, 2, 2};
+        auto sqrt_two = real_type{constants::sqrt_two};
+
+        // Full diamond shape, no interior
+        this->build_and_test(
+            G4Polyhedra(
+                "polyhedra", 0 * deg, 360 * deg, 4, std::size(z), z, rmin, rmax),
+            R"json({"_type":"all","daughters":[{"_type":"stackedextrudedpolygon","polygon":[[1.0,0.0],[0.0,1.0],[-1.0,0.0],[0.0,-1.0]],"polyline":[[0.0,0.0,0.0],[0.0,0.0,0.10000000149011612],[0.0,0.0,0.10000000149011612],[0.0,0.0,0.20000000298023224],[0.0,0.0,0.20000000298023224],[0.0,0.0,0.30000001192092896]],"scaling":[0.2828427255153656,0.2828427255153656,0.2828427255153656,0.2828427255153656,0.2828427255153656,0.2828427255153656]},{"_type":"negated","daughter":{"_type":"stackedextrudedpolygon","polygon":[[1.0,0.0],[0.0,1.0],[-1.0,0.0],[0.0,-1.0]],"polyline":[[0.0,0.0,0.0],[0.0,0.0,0.10000000149011612],[0.0,0.0,0.10000000149011612],[0.0,0.0,0.20000000298023224],[0.0,0.0,0.20000000298023224],[0.0,0.0,0.30000001192092896]],"scaling":[0.1414213627576828,0.1414213627576828,0.0,0.0,0.1414213627576828,0.1414213627576828]},"label":""}],"label":"polyhedra"})json",
```
sethrj: wait, why do the polylines only have 7 digits of precision? (Did you do this on a single precision build)?  Also I don't think `sqrt_two` is used, and if you're passing it as input to geant4 you should use `double`

sethrj (approved):
Small comments to check out, looks good otherwise!

test/orange/g4org/SolidConverter.test.cc
elliottbiondo: Oops, yeah, single precision build. Will fix all.

elliottbiondo: Yes one sqrt_two isn't used. Another one is, and it's `real_type` because it's not going to Geant4.

elliottbiondo: Hold on, what has 7 digits of precision? I just rebuilt with `double` and I am getting the same strings.

sethrj: Your polyline z points are `0.10000000149011612`, `0.20000000298023224` , etc. when they should be numerically `0.1`, `0.2`, etc... that looks like 1 + epsilon_float to me... or am I missing something?

elliottbiondo: Fixed. Though I rebuilt with float but apparently didn't

PR #2087, opened 2025-10-29
Add tests for unusual boundary crossing scenarios

sethrj:
These new tests demonstrate the failure in #2079 (and are partially borrowed by #2081). There are two important scenarios for direction changes on a boundary:
1. After moving to boundary but before exiting, the direction changes so that the track is pointed back inside. Then `cross_boundary` is called and should be ignored. This happens with MSC and fields for nearly tangent tracks.
2. After crossing a boundary but before moving off it, the direction changes to point back into the old volume, and `cross_boundary` should return to the old volume. This happens in optical surface reflection.

The only geometry that seems to behave is VecGeom, which might be because it "bumps" everything along the forward direction. ORANGE and Geant4 both fail to cross the boundary in case 2, and Geant4 fails to stay in the same volume in case 1.


test/geocel/GeoTests.cc
```diff
@@ -2257,6 +2240,155 @@ void TwoBoxesGeoTest::test_detailed_tracking() const
...
+        // FIXME: reentrant fails!
+        EXPECT_EQ("world", test_->volume_name(geo));
+        GTEST_SKIP() << "Unexpected failure to cross volume";
+    }
+    EXPECT_EQ("inner", test_->volume_name(geo));
+
+    // Find the next boundary and make sure that nearer distances aren't
+    // accepted
+    next = geo.find_next_step();
+    EXPECT_SOFT_EQ(7.0710678118654755, to_cm(next.distance));
```
hhollenb: Maybe change this to `10 * dx` as well?

hhollenb (approved):
Tests look good!

test/geocel/GeoTests.cc
```diff
@@ -2257,6 +2240,155 @@ void TwoBoxesGeoTest::test_detailed_tracking() const
...
+    EXPECT_EQ("world", test_->volume_name(geo));
+
+    // Cross back into previous volume (-; -,-)
+    if (CELERITAS_DEBUG && test_->geometry_type() == "Geant4")
+    {
+        // FIXME: can't cross boundary and return
+        EXPECT_THROW(geo.cross_boundary(), DebugError);
+        GTEST_SKIP() << "Consecutive boundary crossing fails for G4";
+    }
+    EXPECT_TRUE(geo.is_on_boundary());
```
amandalund, line 2306: Do we still need to cross back for orange/vecgeom?

amandalund (approved):
Good tests!

test/geocel/GeoTests.cc
```diff
@@ -2257,6 +2240,155 @@ void TwoBoxesGeoTest::test_detailed_tracking() const
...
+        // FIXME: Geant4 changes volumes :(
+        EXPECT_EQ("world", test_->volume_name(geo));
+        GTEST_SKIP() << "Unexpected boundary crossing";
+    }
+    EXPECT_EQ("inner", test_->volume_name(geo));
+
+    // Find the next boundary and make sure that nearer distances aren't
+    // accepted
+    next = geo.find_next_step();
+    EXPECT_SOFT_EQ(10.0 * dx, to_cm(next.distance));
```
amandalund, line 2396: Also I'm seeing a failure with VecGeom here:
```console
89: [ RUN      ] TwoBoxesVgdmlTest.tangent
89: /home/alund/celeritas_project/celeritas/test/geocel/GeoTests.cc:2366: Failure
89: Expected equality of these values:
89:   "inner"
89:   test_->volume_name(geo)
89:     Which is: "world"
89: 
89: /home/alund/celeritas_project/celeritas/test/geocel/GeoTests.cc:2371: Failure
89: Value of: to_cm(next.distance)
89:   Actual: 1e-08
89: Expected: 10.0 * dx
89: Which is: 7.0710678118654746
89: (Relative error -0.99999999858578648 exceeds tolerance 9.9999999999999998e-13)
```

test/geocel/GeoTests.cc
sethrj: Oops! Meant to add that `else` but 🐿️

test/geocel/GeoTests.cc
sethrj: This was caused by the omission above

amandalund: Hmm I was seeing this in tangent test though.

sethrj: Yep turns out that even the CI sees different behaviors. Yay! I've added a note and allowed both until we figure out what's going on.

PR #2088, opened 2025-10-29
Use opaque ID rather than integer for physics offset

sethrj:
Related to #2072, this fixes the slight misuse of `size_type` for physics action<->model translation by using a type-safe action ID rather than a size offset.


amandalund (approved):
Nice, thanks!

PR #2089, opened 2025-10-29
Enable checking of integration tests for older Geant4 versions

sethrj:
Following on to #2083, this removes the "continue on error" for testing g4 version < 11.0, and simply chooses to test the high level integration. Since some capabilities (e.g., tracking manager) aren't supported for older versions, the test logic is flipped with [WILL_FAIL](https://cmake.org/cmake/help/latest/prop_test/WILL_FAIL.html) to show that the tests indeed fail.


amandalund (approved):
Thanks @sethrj!

PR #2091, opened 2025-10-31
Upgrade CI docker image to ROCm 7.1

LSchwiebert:
This PR makes the changes necessary to add hipcub to the ROCm image and move from ROCm 6.3.1 to ROCm 7.1.0. The main reason for moving to a new version is that ROCm 7.1.0 is needed to replace the thrust::DeviceTransform() call to a hipcub::DeviceTransform() call, which is needed to move to a gpu-resident implementation.

Note 1: In the Dockerfile, the web links on lines 24 & 25 are correct. There is no separate link for 7.1 for line 25 -- only latest. Not ideal, so mentioned here.
Note 2: Not sure if everything should be 7.1.0 instead of 7.1, for instance in the yml file.


sethrj:
Thanks Loren! I'm going to try to building this today and upgrading the CI image accordingly.

sethrj (approved):
There's definitely some follow-on work with upgrading the spack toolchain 😅 in a perfect world we could use something like https://github.com/eic/containers

Issue #1746, opened 2025-04-17
Add full-capability G4Track/Step reconstruction


sethrj:
Create a data structure that contains *all* the attributes needed to fully reconstruct a G4Track for sending to Geant4 via `G4StackManager::PushOneTrack` (#1894), or `G4VSensitiveDetector::Hit` (#581, #1397), or `G4VProcess::PostStepDoIt` (#1898).

Required for:
- Improved hit processing (track data over entire step): EM hits #581 and optical #1397 
- Extra physics (track data over most of a step): #1898 
- Region-based return to Geant4 (track state at end of step): #2084
- Potential "tail" execution when little work remains in an event [^1]: apt-sim/AdePT#360
- Stepping action (track data over entire step): #1745

[^1]: We usually think of this in terms of "few tracks in the tracking loop" or "many steps taken" which are proxies for "small total amount of energy" which is a proxy for "not much work left".

Plan:
- Define step point data structure with all necessary data: maybe could just reuse `StepStateDataImpl` in `celeritas/user/StepData` ?
- Defer volume instance stack reconstruction until ORANGE capability is in place and path-to-scalar (i.e., volume instance stack to UniqueVolumeInstanceId) mapping is set up
- Move reconstruction parts from HitProcessor etc. into new classes for reconstructing StepPoint, Track, and Step:
  * Track reconstructor emits a unique_ptr<G4Track> that can be stacked or used as a "per-particle-type" object for detector callbacks
  * Touchable updater gets moved into step point converter
- Identify additional attributes that need to be copied from GPU

Implementation details:
- Collection groups: step state, persistent state, track state
  * Two step states and one persistent state live in track state
  * Step state has position, direction, volume, etc.: key track attributes
  * Persistent state has (particle ID? primary ID? step counters?)
  * Do not worry about making data configurable like we do for hit processor: assume we copy everything
- Managing the data:
  * Before pre-step, swap pre- and post- array pointers [^2], and then record newly initialized particles into the array
  * Also at pre-step, record persistent state from newly initialized particles
  * Async copy out of persistent and pre-step immediately after kernel
  * At post-step, record all updated states before track is destroyed/replaced by secondary
  * Async copy out of post-step state
- Helper classes to convert step points, persistent state, etc.

[...]

PR #1474, opened 2024-10-29
Anchor vtables to fix broken dynamic cast in tests

sethrj:
This follow-on to https://github.com/celeritas-project/celeritas/pull/1436 is to fix another manifestation of the bug caused by https://github.com/llvm/llvm-project/commit/9d525bf94b255df89587db955b5fa2d3c03c2c3e : in optimized mode, llvm fails to `dynamic_cast` classes that are declared `final` but have weak vtables. The weird part of fixing this instance was that to make the vtable strong, I had to anchor the `GeoParamsInterface` virtual destructor because that was in a separate library. (The GeantGeoParams and VecgeomParams worked because the inline virtual destructor was in the same library, so the final class had a "strong" vtable symbol.)

Before:
```console
$ ninja orange && nm lib/liborange.dylib | c++filt | grep 'vtable for celeritas::.*Orange.*' | cut -b 18- | sort
S vtable for celeritas::OrangeParamsOutput
S vtable for celeritas::RaytraceImager<celeritas::OrangeParams>
s vtable for celeritas::CollectionMirror<celeritas::OrangeParamsData>
s vtable for celeritas::OrangeParams
$ ninja geocel && nm lib/libgeocel.dylib | c++filt | grep 'vtable for celeritas::.*Params.*' | cut -b 18- | sort
S vtable for celeritas::GeantGeoParams
S vtable for celeritas::GeoParamsOutput
S vtable for celeritas::RaytraceImager<celeritas::GeantGeoParams>
S vtable for celeritas::RaytraceImager<celeritas::VecgeomParams>
S vtable for celeritas::VecgeomParams
S vtable for celeritas::VecgeomParamsOutput
s vtable for celeritas::CollectionMirror<celeritas::ImageParamsData>
s vtable for celeritas::ImageParams
```

After (note `celeritas::OrangeParams` changes from `s` to `S`:
```console
$ ninja orange && nm lib/liborange.dylib | c++filt | grep 'vtable for celeritas::.*Orange.*' | cut -b 18- | sort
S vtable for celeritas::CollectionMirror<celeritas::OrangeParamsData>
S vtable for celeritas::OrangeParams
S vtable for celeritas::OrangeParamsOutput
S vtable for celeritas::ParamsDataInterface<celeritas::OrangeParamsData>
S vtable for celeritas::RaytraceImager<celeritas::OrangeParams>
[...]

PR #2092, opened 2025-10-31
Support offloading optical distributions to Celeritas standalone optical loop

amandalund:
To support sending optical distributions from Geant4 to a Celeritas optical stepping loop, this adds an interface class for offloading through accel and a `LocalOpticalOffload` for offloading optical distribution data. This works very much like the `LocalTransporter`, except `Push` will buffer a distribution rather than a track. It will take a little more work to cleanly incorporate this into our existing integration classes, but I've don't just enough here to add a user action integration test with optical distribution offloading through a stepping action.


src/accel/LocalOffloadInterface.cc
```diff
@@ -0,0 +1,114 @@
...
+        G4Event const* event = event_manager_->GetConstCurrentEvent();
+        CELER_ASSERT(event);
+        if (event_id_ != id_cast<UniqueEventId>(event->GetEventID()))
+        {
+            // The event ID has changed: reseed it
+            this->InitializeEvent(event->GetEventID());
+        }
+    }
+    CELER_ENSURE(event_id_);
+};
```
sethrj:
```suggestion
}
```

src/accel/LocalOffloadInterface.cc
```diff
@@ -0,0 +1,114 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Validate the thread ID and threading model.
+ */
+void LocalOffloadBase::validate_threading(size_type num_streams) const
```
sethrj: Looks like this method can also be static? Maybe it should be a helper function or performed inside the integration base at the point where we know the thread ID is set for a worker. I think we can assume anyone using the low-level interface will know not to combine openmp with geant4 MT.

src/accel/LocalOffloadInterface.hh
```diff
@@ -0,0 +1,93 @@
...
+    LocalOffloadInterface() = default;
+    CELER_DEFAULT_COPY_MOVE(LocalOffloadInterface);
+    //!@}
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Manage common functionality for offloading EM tracks and optical data.
+ */
+class LocalOffloadBase : public LocalOffloadInterface
```
sethrj: This might be better a separate `detail/EventIdManager.hh` (or something) and used with composition: `InitializeEvent` is the only thing that it overrides, and that's the only functionality it has.

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,186 @@
...
+                       && std::holds_alternative<inp::OpticalOffloadGenerator>(
+                           *options.optical_generator),
+                   << "invalid optical photon generation mechanism for local "
+                      "optical offload");
+
+    // Check the thread ID and MT model
+    this->validate_threading(params.Params()->max_streams());
+
+    // Save a pointer to the generator action
+    CELER_ASSERT(params.optical_params()->gen_reg()->size() == 1);
```
sethrj, line 54: I think an extra `params.optical_params()` assertion (maybe after the validate?) would be good?

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,186 @@
...
+    generate_ = std::dynamic_pointer_cast<optical::GeneratorAction const>(
+        params.optical_params()->gen_reg()->at(GeneratorId(0)));
+    CELER_ASSERT(generate_);
+
+    // Number of photons to buffer before offloading
+    auto const& capacity = *options.optical_capacity;
+    auto_flush_ = capacity.primaries;
+
+    StreamId stream_id{static_cast<size_type>(get_geant_thread_id())};
```
sethrj:
```suggestion
    auto stream_id = id_cast<StreamId>(get_geant_thread_id());
```

src/accel/LocalOffloadInterface.cc
```diff
@@ -0,0 +1,114 @@
...
+
+    event_id_ = id_cast<UniqueEventId>(id);
+
+    if (!(G4Threading::IsMultithreadedApplication()
+          && G4MTRunManager::SeedOncePerCommunication()))
+    {
+        // Since Geant4 schedules events dynamically, reseed the Celeritas RNGs
+        // using the Geant4 event ID for reproducibility. This guarantees that
+        // an event can be reproduced given the event ID.
+        this->Reseed(event_id_.get());
```
sethrj: I forgot this was here when I commented previously: what if you make the interface for this function and below:
```c++
// Return new event ID, or null ID if the same event is already running
[[nodiscard]]  UniqueEventId update(G4Event*);
// Update from run manager, returning a new event ID if updated
[[nodiscard]]  UniqueEventId update();
```

and then the caller can do:
```c++
if (auto new_id = local_event_.update(g4event))
{
    state_.reseed(...);
}
```
for the explicit "initialize event" or 

```c++
if (auto new_id = local_event_.update())
{
    state_.reseed(...);
}
```
for the "flush".

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,186 @@
...
+        state_ = std::make_shared<optical::CoreState<MemSpace::host>>(
+            *params.optical_params(), stream_id, capacity.tracks);
+    }
+
+    // Allocate auxiliary data
+    if (params.Params()->aux_reg())
+    {
+        state_->aux() = std::make_shared<AuxStateVec>(
+            *params.Params()->aux_reg(), memspace, stream_id, capacity.tracks);
+    }
```
sethrj, line 83: I feel like this should be in the optical core params ... and here are we passing the core registry to the optical state?

src/accel/detail/IntegrationSingleton.hh
```diff
@@ -91,6 +99,10 @@ class IntegrationSingleton
     real_type stop_timer() { return get_time_(); }
 
   private:
+    //// TYPES ////
+
+    using VariantOffload = std::variant<LocalTransporter, LocalOpticalOffload>;
```
sethrj: It's a little weird to have a variant of two virtual classes that share the same interface. And even weirder to have an optional of that variant (usually you'd use `variant<std::monostate, ...>`.

Why not `std::unique_ptr<LocalOffloadInterface>` plus dynamic-cast?

src/accel/LocalOffloadInterface.hh
```diff
@@ -0,0 +1,93 @@
...
+{
+  public:
+    // Initialize with options and core shared data
+    virtual void Initialize(SetupOptions const&, SharedParams&) = 0;
+
+    // Set the event ID and reseed the Celeritas RNG at the start of an event
+    virtual void InitializeEvent(int) = 0;
+
+    // Reseed the RNG states
+    virtual void Reseed(size_type) = 0;
```
sethrj: Is reseed intended to be called by the user or just part of the base interface? (If we refactor the InterfaceBase as suggested it can go away; if it needs to stay but is not intended to be aclled by the user, then just make it protected.)

src/accel/LocalTransporter.cc
```diff
@@ -164,48 +156,22 @@ LocalTransporter::LocalTransporter(SetupOptions const& options,
...
-#endif
-        }
-    }
+    // Check the thread ID and MT model
+    this->validate_threading(params.Params()->max_streams());
 
     // Create hit processor on the local thread so that it's deallocated when
     // this object is destroyed
-    StreamId stream_id{static_cast<size_type>(thread_id)};
+    StreamId stream_id{static_cast<size_type>(get_geant_thread_id())};
```
sethrj: Also `id_cast`

sethrj (commented):
Working through this, will resume this evening 😄

src/accel/LocalOpticalOffload.cc
amandalund: Feel like the aux registry should be in the optical core params? And right, currently there's only one `AuxParamsRegistry`/`AuxStateVec` that's used by both core and optical.

src/accel/detail/IntegrationSingleton.hh
amandalund: Yeah I was struggling a little with how to incorporate the optical offload here while maintaining backward compatibility. I can make it a `unique_ptr` to the interface (though I guess that means we may be downcasting every time we `Push`).

src/accel/LocalOffloadInterface.cc
amandalund: We'll still need to implement `LocalOffloadInterface::InitializeEvent(int)`... maybe I'll just duplicate this code for now and we can come back to it.

src/accel/LocalOffloadInterface.hh
amandalund: Removed this.

test/accel/UserActionIntegration.test.cc
```diff
@@ -83,6 +94,190 @@ TEST_F(LarSphere, run)
...
+            // TODO: Why are some steps of zero length generating photons?
+            CELER_LOG(warning)
+                << "Skipping " << num_cherenkov << " Cherenkov photons and "
+                << num_scintillation << " scintillation photons for "
+                << step->GetTrack()->GetDefinition()->GetParticleName()
+                << " with step length=" << step->GetStepLength() << " and "
+                << post_step->GetProcessDefinedStep()->GetProcessName()
+                << "-defined step";
+            return;
+        }
```
amandalund:
```console
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 99 Cherenkov photons and 917 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 87 Cherenkov photons and 853 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 54 Cherenkov photons and 1204 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 7 Cherenkov photons and 701 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 0 Cherenkov photons and 353 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 0 Cherenkov photons and 216 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 0 Cherenkov photons and 17 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 0 Cherenkov photons and 15 scintillation photons
384: debug@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:183: Generating 0 Cherenkov photons and 696 scintillation photons
384: warning@/home/alund/celeritas_project/celeritas/test/accel/UserActionIntegration.test.cc:141: Skipping 0 Cherenkov photons and 696 scintillation photons for e- with step length=0 and NoProcess-defined step
```

Looks like we should skip in this case, which can happen at the last step (when track is at rest) where the `NoProcess` is (from the Geant4 history):

> G4NoProcess : introduced a new class (proposed and written by K. Genser)
  representing an empty process, which is not assigned to any particle,
  and used only in the method G4SteppingManager::InvokeAtRestDoItProcs
  for stable ions at rest (to avoid that Radioactive Decay appears as
  process defining their last step, which can mislead users).

src/accel/LocalOpticalOffload.cc
sethrj: Yes and aha, I'm not sure I designed it with that in mind but I don't see offhand why it wouldn't work 🤔

src/accel/detail/IntegrationSingleton.hh
sethrj: Given the number of virtual calls that happen in Geant4 per step, I think we can afford to not worry about it :)

test/accel/UserActionIntegration.test.cc
sethrj: Sounds like a bug in Geant4 to me (@whokion could probably confirm?) and I have no idea what purpose NoProcess serves even after that description... why would radioactive decay apply to stable ions at rest? (Or is it implying that G4 needs at least *a* process for at-rest tracks, and nothing applies in that case, so they made up a dummy process?)

src/accel/LocalOpticalOffload.cc
amandalund: Yeah, I think currently the generator data even _needs_ to be in the core aux state for the `OffloadAction` to be able to push distributions to it (and none of that data is state-sized). For #1746 though it sounds like we'll need a separate optical aux.

sethrj: We actually might not 🤔 For #1746 we'll need a separate optical step action to process hits, and each instance of the step processor will get its own aux ID I suppose?

test/accel/UserActionIntegration.test.cc
whokion: Yes, we should skip `NoProcess` for optical processes - I believe that it was introduced for the muon decay to the electron from the muonic atom to support the mu2e experiment.  From the user action of my prototype code is also check this `NoProcess`.

```
/*!
 * At the end of stepping, determine whether to use Celeritas to transport it.
 *
 * If the step is one of optical generation processes, we create an associated
 * celeritas optical distribution and push it to a Celeritas local transport (which
 * stacks the distribution on its buffer and potentially flushes it).
 */
void SteppingAction::UserSteppingAction(G4Step const* step)
{
    CELER_EXPECT(step);

    G4double edep = step->GetTotalEnergyDeposit();
    auto* proc_defined_step = step->GetPostStepPoint()->GetProcessDefinedStep();
    if(edep <= 0 || proc_defined_step->GetProcessName() == "NoProcess")
    {
        return;
    }
    ....
```

src/accel/SetupOptions.hh
```diff
@@ -217,7 +223,7 @@ struct SetupOptions
     //! Do not use Celeritas physics for the given Geant4 process names
     VecString ignore_processes;
     //! Only offload a subset of particles
-    VecG4PD offload_particles;
+    std::optional<VecG4PD> offload_particles;
```
sethrj, line 236: Can you add documentation to explain the difference between a false optional and an empty vector, and if/how it complements the CELER_DISABLE/KILL options?

src/accel/SetupOptions.cc
```diff
@@ -257,8 +262,12 @@ inp::FrameworkInput to_inp(SetupOptions const& so)
...
 
     auto includes_muon = [&so]() -> bool {
-        return std::any_of(so.offload_particles.begin(),
-                           so.offload_particles.end(),
+        if (!so.offload_particles)
+        {
+            return false;
+        }
+        return std::any_of(so.offload_particles->begin(),
+                           so.offload_particles->end(),
```
sethrj, line 270: More to consider for https://github.com/celeritas-project/celeritas/issues/1964 ...

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,220 @@
...
+
+    if (num_photons_ >= auto_flush_)
+    {
+        this->Flush();
+    }
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Transport the buffered tracks and all secondaries produced.
```
whokion:
```suggestion
 * Generate and transport optical photons from the buffered distribution data.
```

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,220 @@
...
+            // The event ID has changed: reseed it
+            this->InitializeEvent(event->GetEventID());
+        }
+    }
+    CELER_ASSERT(event_id_);
+
+    if (celeritas::device())
+    {
+        CELER_LOG_LOCAL(debug)
+            << "Transporting " << num_photons_ << " photons from event "
```
whokion:
```suggestion
            << "Transporting " << num_photons_ << " optical photons from event "
```

src/accel/LocalOpticalOffload.cc
```diff
@@ -0,0 +1,220 @@
...
+    // Check the thread ID and MT model
+    validate_geant_threading(params.Params()->max_streams());
+
+    // Save a pointer to the generator action
+    CELER_ASSERT(params.optical_params()->gen_reg()->size() == 1);
+    generate_ = std::dynamic_pointer_cast<optical::GeneratorAction const>(
+        params.optical_params()->gen_reg()->at(GeneratorId(0)));
+    CELER_ASSERT(generate_);
+
+    // Number of photons to buffer before offloading
```
whokion:
```suggestion
    // Number of optical photons to buffer before offloading
```

whokion (commented):
Very good!  

I have the following error from my test  
```
C++ exception with description "/Users/syjun/celeritas/src/celeritas/test/celeritas/optical/Generator.test.cc:112:
celeritas: internal assertion failed: state" thrown in the test body.
```
Is it known or am I doing something wrong?

src/accel/SetupOptions.hh
```diff
@@ -165,6 +169,8 @@ struct SetupOptions
 
     //! Capacity for storing optical photon state
     std::optional<inp::OpticalStateCapacity> optical_capacity;
+    //! Optical photon generation mechanism
+    std::optional<inp::OpticalGenerator> optical_generator;
```
sethrj, line 183: Should we make this a single optional struct that has both inputs? The second one being an optional variant is still odd. Or should we wait until we complete the full `inp` setup mechanism for accel?

src/accel/detail/IntegrationSingleton.hh
```diff
@@ -105,6 +116,12 @@ class IntegrationSingleton
     // Only this class can construct
     IntegrationSingleton();
 
+    // Static thread-local Celeritas state data
+    static UPOffload& offload();
```
sethrj: I think this name could be a bit confusing with the two static thread-local (and non-static interface?!) methods above...

src/accel/detail/IntegrationSingleton.cc
```diff
@@ -259,12 +298,16 @@ void IntegrationSingleton::finalize_local_transporter()
...
             CELER_VALIDATE(lt,
                            << "local thread "
                            << G4Threading::G4GetThreadId() + 1
                            << " cannot be finalized more than once");
-            params_.timer()->RecordActionTime(lt.GetActionTime());
+            if (!this->optical_offload())
+            {
+                params_.timer()->RecordActionTime(
+                    IntegrationSingleton::local_transporter().GetActionTime());
+            }
```
sethrj, line 310: Should we reset the UP after calling `Finalize`? (I don't think we really need to unless we have one `run` that uses optical offload and another that uses EM offload, but no one will ever do that and/or we shouldn't support it.)

test/accel/UserActionIntegration.test.cc
```diff
@@ -83,6 +94,190 @@ TEST_F(LarSphere, run)
     cout << "second run done" << endl;
 }
 
+//---------------------------------------------------------------------------//
+// LAR SPHERE WITH OPTICAL OFFLOAD
+//---------------------------------------------------------------------------//
+class LSOOSteppingAction final : public G4UserSteppingAction
+{
+    void UserSteppingAction(G4Step const* step) final
```
sethrj: Maybe un-inline this long method to improve readability

amandalund:
Thanks @whokion! I'm not sure why you'd be hitting that error... I didn't see it in the CI, and I double checked that I'm not seeing it locally in the device test either.

sethrj (approved):
It's a bit messy but hopefully the continuing work after this PR will clean it up?

src/accel/detail/IntegrationSingleton.cc
amandalund: Doesn't hurt! (though I agree that kind of mixed mode isn't something we should support)

amandalund: Actually then I'm seeing failures in the tracking manager integration tests with multiple runs:
```console
365: error@/home/alund/celeritas_project/celeritas/src/accel/TrackingManagerIntegration.cc:58: Particle "e-"@0x1eec40f0 (PDG = 11): tracking manager does not have the expected local transporter
365: error@/home/alund/celeritas_project/celeritas/src/accel/TrackingManagerIntegration.cc:58: Particle "e+"@0x1eec42d0 (PDG = -11): tracking manager does not have the expected local transporter
365: error@/home/alund/celeritas_project/celeritas/src/accel/TrackingManagerIntegration.cc:58: Particle "gamma"@0x1eec3f10 (PDG = 22): tracking manager does not have the expected local transporter
```

sethrj: K leave it then :)

src/accel/SetupOptions.hh
amandalund: I think anything that moves us closer to the final accel `inp` setup is a good idea, and a separate struct for the optical options makes sense since these are already `inp`-like. Maybe we can improve the optical options in a follow-on, since there's still more work to be done supporting the optical offload in our integration interface.

sethrj:
> Very good!
> 
> I have the following error from my test
> 
> ```
> C++ exception with description "/Users/syjun/celeritas/src/celeritas/test/celeritas/optical/Generator.test.cc:112:
> celeritas: internal assertion failed: state" thrown in the test body.
> ```
> 
> Is it known or am I doing something wrong?

@whokion I just saw @mrguilima with the same error:  https://github.com/celeritas-project/celeritas/pull/1474#issuecomment-3482194243 .

test/accel/UserActionIntegration.test.cc
```diff
@@ -83,6 +94,191 @@ TEST_F(LarSphere, run)
...
+    }
+
+    auto* pre_step = step->GetPreStepPoint();
+    auto* post_step = step->GetPostStepPoint();
+    CELER_ASSERT(pre_step && post_step);
+
+    // Create distribution and push to Celeritas
+    // TODO: Get optical material ID
+    // TODO: Does the post-step speed account for only continuous energy
+    // loss or continuous+discrete?
```
sethrj, line 237: I think continuous plus discrete! @Rashika-Gupta @stognini this ties into our conversation today. We could probably use a new issue for how to accurately describe the scintillation/cherenkov due to along-step energy loss versus end-of-step energy deposition due to tracking cuts or (for example) the hard-coded 100 eV Klein-Nishina electron production threshold.

sethrj (approved):
Thanks @amandalund ! Updates look good.

sethrj:
#1860

test/accel/UserActionIntegration.test.cc
sethrj: See #2152, #2160

PR #2094, opened 2025-11-03
Avoid optical state dynamic casts in test

sethrj:
This works around a weird crash reported by @mrguilima and @whokion (https://github.com/celeritas-project/celeritas/pull/2092#pullrequestreview-3411605606) who are building with RelWithDebInfo on their mac laptops using the latest xcode (apple clang 17 I presume). We worked around the failure on @mrguilima 's laptop by changing the build type to Debug. I cannot reproduce the failure  on my machine.

I believe the error is related to #1474 and caused by the recent change in #2082. Unfortunately it's not quite the same because the base-est class destructor is indeed anchored, but it *could* be because `CoreStateBase`'s implicit inline virtual destructor is not?

The "fix" itself should be 35a30c8760ec9972c2aa21a4d7d9e28bba6d2ac2, but 0a03b723c0bfe11fb9294d852b717e4f0bc4c9a2 and ee1a4da6fefb5917159ffefa595f85a869416fd9 both eliminate the need for dynamic casts altogether, so I have two reasons to think it'll be fixed :)


amandalund (approved):
https://github.com/celeritas-project/celeritas/pull/2094/commits/81ad28df990eeba47667b1bb4b64961e5aafaf3e and https://github.com/celeritas-project/celeritas/commit/0a03b723c0bfe11fb9294d852b717e4f0bc4c9a2 are also done in #2092 :)

whokion (approved):
Indeed, this MR cleared the error from the `optical_Generator` test.

sethrj:
Oops sorry @amandalund for the duplication! I remembered there was indeed cleanup mixed into the new functionality 😅

PR #2096, opened 2025-11-04
Reconstruct canonical volumes directly from ORANGE track view

### LINKED ISSUE #2057: Reconstruct G4NavHistory from ORANGE state with volume instance stack

sethrj:
This adds the necessary functionality to reconstruct a Geant4 "navigation history" (i.e. unique instance ID/path in the volume graph) from the converted and constructed ORANGE geometry, including inlined volumes. 

The summary below is included in an update to the documentation in this PR and is [rendered on my fork](https://sethrj.github.io/celeritas/user/implementation/geometry/orange/translation.html) for the time being.

---

There are several steps that have to be taken:
- During the conversion of local+physical volumes to "unit protos" and volumes (material placements, background, daughters), when placing children inside a volume that's already instantiated/placed in the current proto, we inform the child that it's inside a parent material. This creates a new "local parent" entry in the unit proto input describing the connectivity, even if the volumes may live side-by-side as far as the proto/implementation is concerned. Because we don't know the "volume ID" when building the proto input, we actually save the "proto material ID" (which is a volume placement with all the children subtracted out).
- During proto build, the local material IDs are remapped to volume IDs and saved into the unit input as a "local parent map".
- During the flattening of the geometry in the UnitInserter, we construct two vectors for each universe. One is the "local parent volume", which is (if present) is the local impl volume ID of the canonical volume that encloses the  "local volume level", the volume level relative to the universe itself for inlining universes.
- During execution, we sum the local volume levels over all universes to get the total volume level. We jump backward along the local parents to get the full volume stack.

Consider the following geometry with volumes world, A–D, X, and Y (note that some have multiple placements, and one of X has a rotation as well):
inlining-physical

This is represented by the following volume hierarchy (the same as VolumeParams) and in ORANGE will be converted to three universes  U, V, and W, even though the geometry has four volume levels.  Leaf volumes are "inlined" into the same universe as the parent, and one-off enclosures like D also are usually inlined.
inlining-hierarchy

The resulting ORANGE geometry has three universes. Each universe has a few "local impl volumes": the top impl volume is a true *volume* since it can be replicated multiple times, and the other volumes are *volume instances* since they're specific transformed placements in the geometry. The arrows in this graph represent the "local volume parent", and the "local volume level" is listed on the far left.
inlining-levels

An ORANGE state has a stack of `[{universe, impl volume}, ...]`. The following shows the correspondence between the volume path (which, when using volume instances rather than volumes, defines the "unique instance") and the ORANGE state, along with the "local levels" calculated from the ORANGE state that sum to the actual volume level.

Volume path | Volume level | ORANGE state | Local levels
-- | -- | -- | --
World→X→Y→C | 3 | {U, X}, {V, Y}, {W, C} | {1, 1, 1}
World | 0 | {U, World} | {0}
World→A→D | 2 | {U, D} | {2}
World→X→B | 2 | {U, X}, {V, B} | {1, 1}
World→X | 1 | {U, X} | {1}

Closes #2057

**FOLLOW UP: quick PR to delete navi touchable updater**


sethrj:
@elliottbiondo this is ready for review! I think I'll add the PR's documentation into the online user manual in a follow-on PR.

elliottbiondo:
I think you mean:

```
During the conversion of logical+physical volumes...
```

src/orange/OrangeTrackView.hh
```diff
@@ -488,19 +505,51 @@ CELER_FUNCTION VolumeLevelId OrangeTrackView::volume_level() const
...
+    // Start writing backward from end of levels array
+    CELER_ASSERT(levels.size() > 0);
+    unsigned rdst = levels.size();
+
+    // Loop over universes, local to global
+    UnivLevelId const ulev{this->univ_level()};
+    auto ui = this->make_univ_indexer();
+    TrackerVisitor visit_tracker{params_};
+
+    for (auto neg_ulev : range(ulev.unchecked_get() + 1))
```
elliottbiondo:
```suggestion
    for (auto ulev : range<int>(ulev.unchecked_get() + 1).step(-1))
```

src/orange/OrangeTrackView.hh
```diff
@@ -488,19 +505,51 @@ CELER_FUNCTION VolumeLevelId OrangeTrackView::volume_level() const
...
     CELER_EXPECT(!this->is_outside());
     CELER_EXPECT(this->univ_level() < levels.size());
 
-    // To guard against errors and enable unit tests, we first make sure we're
-    // not going off the end. (If we are at the global level without correct
-    // instance information, then this will just return a null ID.)
-    CELER_NOT_IMPLEMENTED("canonical volume instance");
+    // Start writing backward from end of levels array
+    CELER_ASSERT(levels.size() > 0);
+    unsigned rdst = levels.size();
```
elliottbiondo: It's not clear to me what this stands for? Why not `level_idx`?

src/orange/OrangeTrackView.hh
```diff
@@ -488,19 +505,51 @@ CELER_FUNCTION VolumeLevelId OrangeTrackView::volume_level() const
...
+                    [lv_id](auto&& t) { return t.local_parent(lv_id); }, univ);
+            }
+            else
+            {
+                // No volume instance at this level
+                break;
+            }
+        } while (lv_id);
+    }
+    // Input should have been resized to exactly match number of nested level
```
elliottbiondo: levels

src/orange/OrangeData.hh
```diff
@@ -68,7 +71,7 @@ struct OrangeParamsScalars
 
 //---------------------------------------------------------------------------//
 /*!
- * Data for a single volume definition.
+ * Data for a single local implementation volume definition.
```
elliottbiondo: Sanity check: "local" is used only for implementation volumes, right?

src/orange/OrangeData.hh
```diff
@@ -395,6 +417,7 @@ struct OrangeParamsData
     Items<LocalSurfaceId> local_surface_ids;
     Items<LocalVolumeId> local_volume_ids;
     Items<RealId> real_ids;
+    Items<vol_level_uint> vd_uints;
```
elliottbiondo: `vl_uints`

src/orange/OrangeData.hh
```diff
@@ -436,12 +459,14 @@ struct OrangeParamsData
...
         volume_ids = other.volume_ids;
         volume_instance_ids = other.volume_instance_ids;
+        parent_impl_volumes = other.parent_impl_volumes;
 
         bih_tree_data = other.bih_tree_data;
 
         local_surface_ids = other.local_surface_ids;
         local_volume_ids = other.local_volume_ids;
         real_ids = other.real_ids;
+        vd_uints = other.vd_uints;
```
elliottbiondo: Same

elliottbiondo: Same

src/orange/detail/UnitInserter.cc
```diff
@@ -259,6 +260,82 @@ std::string to_string(VolumeInput::VariantLabel const& vlabel)
                       vlabel);
 }
 
+//---------------------------------------------------------------------------//
```
elliottbiondo, line 264: ```// Make a vector of local parents for each volume```

src/orange/detail/UnitInserter.cc
```diff
@@ -259,6 +260,82 @@ std::string to_string(VolumeInput::VariantLabel const& vlabel)
...
+    {
+        CELER_ASSERT(child < num_volumes);
+        CELER_ASSERT(parent < num_volumes);
+        local_parents[child.get()] = parent;
+    }
+
+    return local_parents;
+}
+
+//---------------------------------------------------------------------------//
```
elliottbiondo, line 291: ```// Make a vector of local levels for each volume```

src/orange/detail/UnitInserter.cc
```diff
@@ -259,6 +260,82 @@ std::string to_string(VolumeInput::VariantLabel const& vlabel)
...
+std::vector<vol_level_uint>
+make_local_level_vec(std::vector<LocalVolumeId> const& local_parents)
+{
+    constexpr vol_level_uint not_visited{static_cast<vol_level_uint>(-1)};
+    std::vector<vol_level_uint> local_vol_level(local_parents.size(),
+                                                not_visited);
+    std::vector<LocalVolumeId> stack;
+
+    // Fill local levels
```
elliottbiondo: ... using a depth-first search starting with each local volume

src/orange/detail/UnitInserter.hh
```diff
@@ -56,6 +56,7 @@ class UnitInserter
     DedupeCollectionBuilder<LocalSurfaceId> local_surface_ids_;
     DedupeCollectionBuilder<LocalVolumeId> local_volume_ids_;
     DedupeCollectionBuilder<OpaqueId<real_type>> real_ids_;
+    DedupeCollectionBuilder<vol_level_uint> vd_uints_;
```
elliottbiondo: I have been thinking that `vd` was the now-unused "volume depth", but now I am not sure...

test/orange/OrangeGeant.test.cc
```diff
@@ -89,6 +93,90 @@ class MultiLevelTest
 {
 };
 
+// Test the stack/volume points to see what universe and local volume they
+// translate to
+TEST_F(MultiLevelTest, univ_depths)
```
elliottbiondo: univ_levels

src/orange/g4org/ProtoConstructor.cc
```diff
@@ -47,13 +47,14 @@ GeoMatId background_fill(GeoMatId mat)
...
 /*!
- * Construct a proto-universe from a physical volume.
+ * Construct a proto-universe from a logical volume.
  *
  * We can use logical volume for the structure, but we need to associate the
  * world physical volume ID.
  */
 auto ProtoConstructor::operator()(LogicalVolume const& lv) -> SPUnitProto
 {
+    ++depth_;
```
elliottbiondo, line 57: `level` here and throughout

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -693,5 +698,41 @@ CELER_FUNCTION DaughterId SimpleUnitTracker::daughter(LocalVolumeId vol) const
...
+ */
+CELER_FUNCTION auto SimpleUnitTracker::local_vol_level(LocalVolumeId vol) const
+    -> vol_level_uint
+{
+    CELER_EXPECT(unit_record_.local_vol_level.empty()
+                 || vol < unit_record_.volumes.size());
+    if (unit_record_.local_vol_level.empty())
+        return 0;
+
+    OpaqueId<vol_level_uint> depth_ptr = unit_record_.local_vol_level[vol];
```
elliottbiondo: level

amandalund:
Sorry for the delay in reviewing this @sethrj, at the top of my list!

src/orange/OrangeData.hh
```diff
@@ -436,12 +459,14 @@ struct OrangeParamsData
 
         volume_ids = other.volume_ids;
         volume_instance_ids = other.volume_instance_ids;
+        parent_impl_volumes = other.parent_impl_volumes;
```
elliottbiondo: These are NOT local volumes, i.e., this is the global impl volumes?

src/orange/OrangeData.hh
```diff
@@ -382,11 +405,10 @@ struct OrangeParamsData
     Items<RectArrayRecord> rect_arrays;
     Items<TransformRecord> transforms;
 
-    // Optional map of ORANGE internal volume ID -> Celeritas volume ID
+    // Map of ORANGE internal volume ID -> canonical IDs and structure
     ImplVolumeItems<VolumeId> volume_ids;
     ImplVolumeItems<VolumeInstanceId> volume_instance_ids;
-    // TODO: for reconstructing hierarchy:
-    // ImplVolumeItems<ImplVolumeId> parent_impl_volumes;
+    ImplVolumeItems<ImplVolumeId> parent_impl_volumes;
```
elliottbiondo: The 408 comment does not apply to this line

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -73,6 +72,12 @@ class SimpleUnitTracker
     // DaughterId of universe embedded in a given volume
     inline CELER_FUNCTION DaughterId daughter(LocalVolumeId vol) const;
 
+    // Canonical volume depth relative to this universe
+    inline CELER_FUNCTION vol_level_uint local_vol_level(LocalVolumeId vol) const;
+
+    // Canonical parent volume placement in the local universe, if any
+    inline CELER_FUNCTION LocalVolumeId local_parent(LocalVolumeId vol) const;
```
elliottbiondo, line 83: I am confused: `LocalVolumeId` is being used as to describe a canonical volume here, but an impl volume places like the BIH?

elliottbiondo (changes requested):
Thanks again for the diagrams. These make the concepts straightforward, but there are places in the code where variable/type names don's make it clear what concept is being referred to. The biggest issue is canonical vs impl volumes, especially for example, when talking about parent volumes, where we care about both. Can go through this together Monday?

src/orange/OrangeData.hh
sethrj: Correct!

src/orange/OrangeData.hh
sethrj: I'll rephrase the comment; the idea is that these fields are used only for reconstructing canonical volume information.

src/orange/OrangeTrackView.hh
sethrj: It's meant to be like a reverse iterator (rend); and when I first devised the algorithm I wasn't expecting `levels` to be sized correctly based on `depth`. But since we require the for now, I can update it to be the level.

src/orange/OrangeTrackView.hh
sethrj: Unfortunately the reverse `range` code doesn't work well with opaque IDs and signed integers... I'll see what I can do

src/orange/OrangeData.hh
sethrj: Actually this field is unused; originally I was going to put all the impl volumes into a single array, but now I'm putting local volumes into a per-universe array.

src/orange/detail/UnitInserter.hh
sethrj: Yep, you're right, fixed

src/orange/g4org/ProtoConstructor.cc
sethrj: It's another overload here 😅 this one is about construction depth strictly for debug output. It seems like the depth here is neither volume nor volume instance but a combination of both...

src/orange/univ/SimpleUnitTracker.hh
sethrj: This is the local volume ID for the "canonical" parent. I've improved the documentation.

test/orange/OrangeGeant.test.cc
sethrj: Not quite; this is a test common to all the geant4 tests that compares the canonical volume instance stack.

src/orange/univ/SimpleUnitTracker.hh
elliottbiondo: When we first started doing multi-level tracking, we created `LocalVolumeId` in contrast with `VolumeId` to avoid mixing them up. I am wondering if a similar distinction between `Impl` and `Canonical` is worth making; I would think these two types of volumes are even more prone to mixing up.

src/orange/OrangeData.hh
elliottbiondo: FWIW elsewhere in the code you have: 

> Create a vector (indexed by local volume ID) of local canonical parents.

src/orange/univ/SimpleUnitTracker.hh
sethrj: `LocalVolumeId` is a local `ImplVolumeId`; all `Volume*Id` are canonical (and I strongly oppose the hypothetical name `CanonicalVolumeUniqueInstanceId`). If we do anything, I would say we rename `LocalVolumeId` -> `LocalImplVolId`.

sethrj: The distinction you're talking about, BTW, was started in #1839 (where we just used a type alias to document intent) and finished in #1903.

doc/implementation/geometry/orange/construction.rst
```diff
@@ -0,0 +1,137 @@
...
+   IntersectRegion,
+   which is an implementation that builds the underlying surfaces and bounding
+   boxes. Shapes should be as simple as possible, aligned along and
+   usually centered on the *z* axis.
+Solid
+   A shape that's hollowed out and/or has a slice removed. It is equivalent to
+   a CSG operation on two shapes of the same type and an azimuthal wedge.
+PolySolid
+   A union of transformed solids along the *z* axis, which can also be hollowed
+   and sliced azimuthally.
```
elliottbiondo, line 60:
```
RevolvedPolygon
    A convex or concave polygon, revolved around the *z* axis, which can be sliced azimuthally.
```

(mea culpa)

src/orange/OrangeData.hh
```diff
@@ -217,18 +221,38 @@ struct TransformRecord
 //---------------------------------------------------------------------------//
 /*!
  * Scalar data for a single "unit" of volumes defined by surfaces.
+ *
+ * If using a HEP geometry implementation and this unit represents multiple
+ * parent/child levels of the geometry hierarchy (via "inlining" during
+ * construction; see \c g4org::ProtoConstructor):
+ * - Local parent is a map of local to local volume IDs: the value is another
```
elliottbiondo: Specify that both the key and value refer to canonical volumes

src/orange/detail/UnitInserter.cc
```diff
@@ -259,6 +261,98 @@ std::string to_string(VolumeInput::VariantLabel const& vlabel)
...
+    }
+
+    return local_parents;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Determine relative canonical volume levels of each local volume.
+ *
+ * Use a depth-first search to fill an array, indexed by local volumes, of
```
elliottbiondo:
```suggestion
 * Use a depth-first search to fill an array, indexed by local impl volumes, of
```

src/orange/detail/UnitInserter.cc
```diff
@@ -259,6 +261,98 @@ std::string to_string(VolumeInput::VariantLabel const& vlabel)
                       vlabel);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Create a vector (indexed by local volume ID) of local canonical parents.
```
elliottbiondo, line 266:
```suggestion
 * Create a vector (indexed by local impl volume ID) of local canonical parents.
```

src/orange/orangeinp/UnitProto.cc
```diff
@@ -356,6 +359,29 @@ void UnitProto::build(ProtoBuilder& input) const
     // nodes for the region, because we can't know which ones have the
     // user-supplied volume names
     auto vol_iter = result.volumes.begin();
+    // Local volume ID of the first local 'material' placement: after exterior,
```
elliottbiondo:
```suggestion
    // Local impl volume ID of the first local 'material' placement: after exterior,
```

elliottbiondo (approved):
The comments really helped---thanks! Just a few more things.

src/orange/OrangeData.hh
sethrj: They refer to local impl volumes, but the "parent" aspect refers to the canonical volumes they represent

Issue #2015, opened 2025-10-07
Asynchronously run a step without waiting for thrust algorithms


sethrj:
Currently our core step algorithms have several "device synchronize" due to thrust algorithms:
- [ ] Determining the number of active tracks (celeritas/track/detail/TrackInitAlgorithms.cu)
- [ ] Determining the location of vacancies to fill (celeritas/track/detail/TrackInitAlgorithms.cu)

And in the optical loop similarly:
- [ ] Counting the number of pending tracks (celeritas/optical/gen/detail/OffloadAlgorithms.cu)
- [ ] Removing empty pending distributions (celeritas/optical/gen/detail/OffloadAlgorithms.cu)
- [ ] Compacting the TrackSlotIds of the inactive tracks (celeritas/optical/action/detail/TrackInitAlgorithms.cu)
- [ ] Partitioning those vacancies among the generators (celeritas/optical/gen/detail/GeneratorAlgorithms.cu)

This is in *addition* to the step collection/hit algorithms.

All of these will have to be replaced with algorithms that can write to a device memory location, and dependencies on that data will have to remain on device (rather than be managed on CPU). Quantities such as the number of active/alive tracks can be asynchronously copied to host memory and picked up at the start of the next step (or deferred to later if we're ok with running an empty step).

@LSchwiebert

PR #2097, opened 2025-11-04
Move core step counter updates to primary generator

sethrj:
Needed for https://github.com/celeritas-project/celeritas/issues/2015 but just want to check first :)

### DIFF
--- a/src/celeritas/global/Stepper.cc
+++ b/src/celeritas/global/Stepper.cc
@@ -131,7 +131,6 @@ auto Stepper<M>::operator()() -> result_type
 {
     ScopedProfiling profile_this{"step"};
     auto& counters = state_->counters();
-    counters.num_generated = 0;
     actions_->step(*params_, *state_);
 
     // Get the number of track initializers and active tracks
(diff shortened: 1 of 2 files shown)


src/celeritas/global/Stepper.cc
```diff
@@ -131,7 +131,6 @@ auto Stepper<M>::operator()() -> result_type
 {
     ScopedProfiling profile_this{"step"};
     auto& counters = state_->counters();
-    counters.num_generated = 0;
```
amandalund, line 134: With these changes we're only clearing `num_generated` when primaries are inserted, correct? I think we might need to reset that elsewhere or we could have a nonzero number generated when we're stepping on existing tracks.

sethrj: You're right, I should have inserted the num_generated into the `step_impl`: that should get called every time, right?

amandalund: I think every step where we _don't_ insert primaries?

sethrj: Maybe I need to add a test of the reported step counts. Nothing is failing and it seems like it should :\

amandalund: Yep, probably a good idea, it doesn't look like we test that currently.

sethrj:
Closing since I didn't get around to adding the tests before @LSchwiebert started working on that part of the code.

PR #342, opened 2022-01-12
Overhaul CMake options and replace scripts with CMakePresets

sethrj:
* Refactor CMake code for setting up dependencies:
    - Toggle all packages based on availability (searching for package `X` quietly if `CELERITAS_USE_X` isn't defined)
    - Minimize errors for missing packages when requesting demo
    - Use generator expressions rather than variables for safety (python executable in the generator script)
    - Add `FindX` modules and move vecgeom code there (replacing the `celeritas_find_package` function)
    - Add MakeRulesOverride for default compiler-dependent flags (`-g -G`, color diagnostics, warnings, etc)
    - Enable documentation based on availability of Doxygen, not with a separate option flag
* Remove external subdirectory and git submodules. We're not using them in CI or in our development so they just complicate things. If needed we can use [https://cmake.org/cmake/help/latest/module/FetchContent.html](FetchContent) to download and build googletest/json as part of the configure/build process.
* Replace build scripts with [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html): I've been wanting to try out this feature for a while, but it's a new standard and cross-platform way of setting configure/build/test options. A set of standard options lives in `CMakePresets.json` at the top level, and you can define `CMakeUserPresets.json` to extend those. A small new helper script symlinks system-based CMakeUserPresets from `scripts/cmake-presets`.

PR #2098, opened 2025-11-04
Do not set `-Wall -Wextra -pedantic` by default

sethrj:
The [CMAKE_USER_MAKE_RULES_OVERRIDE](https://cmake.org/cmake/help/latest/variable/CMAKE_USER_MAKE_RULES_OVERRIDE.html) was an old and basically unsupported way to provide default options for compilers via `CMAKE_<LANG>_FLAGS_INIT`. Now that we have cmake presets and CI, we should let users decide their default flags.


cmake/CeleritasMakeRulesOverride.cmake
```diff
@@ -1,40 +0,0 @@
...
-and
-https://cmake.org/cmake/help/latest/variable/CMAKE_USER_MAKE_RULES_OVERRIDE.html
-
-#]=======================================================================]
-
-# Enable lots of warnings for GCC and Clang by default
-foreach(_lang C CXX)
-  set(_id "${CMAKE_${_lang}_COMPILER_ID}")
-  if(_id STREQUAL "GNU" OR _id MATCHES "Clang$")
-    string(APPEND CMAKE_${_lang}_FLAGS_INIT "-Wall -Wextra -pedantic")
```
pcanal, line 22: Should we mention the dropping (no longer the default for Celeritas build) of those options? (Even if they are in all the presets, they would be 'gone' from a non-preset build).

pcanal:
> Now that we have cmake presets, it's no longer necessary.

Ironically the use of `MakeRulesOverride` was introduced in ` Overhaul CMake options and replace scripts with CMakePresets (#342)` :)

pcanal (approved):
LGTM.

cmake/CeleritasMakeRulesOverride.cmake
sethrj: Yeah we should find an appropriate place to mention it. Maybe this PR title since that shows up in the release notes? Ideas?

pcanal: If it does show in the release note, that would be sufficient I think.

PR #2099, opened 2025-11-07
Tweak optical offload settings and construction

amandalund:
A few small changes following on to #2092:
- Construct a single `optical::Transporter` (and `ActionGroups`)  per params instead of per stream for `LocalOpticalOffload`
- Move the optical setup options to a separate struct
- Log accumulated optical loop stats in `LocalOpticalOffload::Finalize()`
- Save optical `max_step_iters` to `inp::Problem`

### DIFF
--- a/test/accel/IntegrationTestBase.cc
+++ b/test/accel/IntegrationTestBase.cc
@@ -552,12 +552,12 @@ SetupOptions OpNoviceIntegrationMixin::make_setup_options()
 {
     auto result = Base::make_setup_options();
     result.sd.enabled = false;
-    result.optical_capacity = [] {
-        inp::OpticalStateCapacity cap;
-        cap.tracks = 32768;
-        cap.generators = 32768 * 8;
-        cap.primaries = cap.generators;
-        return cap;
+    result.optical = [] {
+        OpticalSetupOptions opt;
+        opt.capacity.tracks = 32768;
+        opt.capacity.generators = 32768 * 8;
+        opt.capacity.primaries = opt.capacity.generators;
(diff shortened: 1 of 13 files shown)


test/accel/IntegrationTestBase.cc
```diff
@@ -552,12 +552,12 @@ SetupOptions OpNoviceIntegrationMixin::make_setup_options()
...
-        cap.generators = 32768 * 8;
-        cap.primaries = cap.generators;
-        return cap;
+    result.optical = [] {
+        OpticalSetupOptions opt;
+        opt.capacity.tracks = 32768;
+        opt.capacity.generators = 32768 * 8;
+        opt.capacity.primaries = opt.capacity.generators;
+        return opt;
     }();
```
sethrj: Or I think this would work too (for a smaller diff 😉 ):
```suggestion
    result.optical.emplace([] {
        inp::OpticalStateCapacity cap;
        cap.tracks = 32768;
        cap.generators = 32768 * 8;
        cap.primaries = cap.generators;
        return cap;
}();
```

sethrj (approved):
Looks great! Only have one trivial suggestion you can ignore. Maybe before 0.7 once we have the capability in place, we can figure out a more intuitive way to activate the different optical generator types.

amandalund:
> Maybe before 0.7 once we have the capability in place, we can figure out a more intuitive way to activate the different optical generator types.

For sure! The setup is a bit of a mess right now (just enough to test that things are working), but once we have all the functionality it should be clearer how to improve that.

PR #2100, opened 2025-11-08
Add grid-based sampling for scintillation spectra

### LINKED ISSUE #2047: Switch to grid-based sampling for scintillation spectra
Following on to #1974, no experiments or users have the custom Celeritas scintillation mean/stdev distributions: they all use grid-based inputs in energy space. So let's use those, but still convert to wavelength space.

- [ ] Establish whether the user inputs are probability per bin, or a spectrum, etc. (add a test with nonuniform energy points): look at geant4 source code
- [ ] Convert probability/energy into CDF vs wavelength
- [ ] For lambda/sigma Gaussian inputs in unit tests (good for validation) use `erf` as integrated normal distribution to determine PDF/CDF
- [ ] Replace gaussian sampler in scintillation with the appropriate sampler
- [ ] For validation, have a nonuniform spectrum and bin the primary photon energies uniformly and compare between celeritas and g4 and see if they match expected spectrum

Rashika-Gupta:
Added support for grid-based inputs (energy, intensity) in `ImportScintComponent`
Built CDFs from input spectra and stored them in `ScintillationData`
Updated `ScintillationGenerator` to sample photons using either Gaussian or grid-based spectra
Preserved backward compatibility with existing Gaussian-only inputs


test/celeritas/grid/UniformLogGridCalculator.test.cc
```diff
@@ -89,7 +89,7 @@ TEST_F(UniformLogGridCalculatorTest, spline)
     }
     else
     {
-        EXPECT_SOFT_EQ(60.439491271972656, calc(Energy(99.99)));
+        EXPECT_SOFT_EQ(60.433685302734375, calc(Energy(99.99)));
```
sethrj: The CI still seems to have the old result, and you didn't do anything to change the underlying code... is this test failing on your local machine? What configuration have you got set up?

test/celeritas/optical/Generator.test.cc
```diff
@@ -241,8 +241,8 @@ TEST_F(LArSphereGeneratorTest, generator)
     if (reference_configuration)
     {
         EXPECT_EQ(51226, gen.num_generated);
-        EXPECT_EQ(54319, result.steps);
-        EXPECT_EQ(15, result.step_iters);
+        EXPECT_EQ(53429, result.steps);
+        EXPECT_EQ(14, result.step_iters);
```
sethrj, line 245: This shouldn't have changed since the lar sphere doesn't use a scintillation spectrum, it uses gaussian. Right?

test/celeritas/grid/UniformLogGridCalculator.test.cc
Rashika-Gupta: Yes, that was the case. It was failing on my machine under `float` precision. 
I updated this and reverted in the commit below.

sethrj: Interesting, when you come by today let's take a look at your build configuration and see why it's different; we may need to loosen the tolerance for this case.

test/celeritas/optical/Generator.test.cc
Rashika-Gupta: Looking at `lar-sphere.gdml` , one of the scintillation component `SC3` is tabular as shown [here](https://github.com/celeritas-project/celeritas/blob/7ce7a34566458e39fd74fd2d503843d7d7420e21/test/geocel/data/lar-sphere.gdml#L28). Also discussed in PR #1722

src/celeritas/ext/GeantImporter.cc
```diff
@@ -222,32 +222,46 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
-        }
-
         // Rise time is not defined for particle type in Geant4
         get(&comp.rise_time, "RISETIME", ImportUnits::time);
         get(&comp.fall_time, "TIMECONSTANT", ImportUnits::time);
 
-        if (any_found)
+        using ScintOption
+            = std::variant<ImportGaussianScintComponent, inp::Grid>;
+        ScintOption scint_data{ImportGaussianScintComponent{}};
```
sethrj: I suggest this change:
```suggestion
        // Spectrum for this component may be missing (default), gaussian (Celeritas-only), or grid
        using ScintSpectrumComponent
            = std::variant<std::monostate, ImportGaussianScintComponent, inp::Grid>;
        ScintSpectrumComponent component;
```
The default variant ("monostate", sort of like a built-in `std::optional`) eliminates the need for `any_found`, so you should delete that afterward.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -222,32 +222,46 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
+                    any_found = true;
+                }
+                else if (get(prop, label, ImportUnits::len))
+                {
+                    CELER_LOG(warning)
+                        << "Deprecated property name " << prefix << label
+                        << ": use CELER_" << prefix << label;
+                }
+            }
+            if (gauss.lambda_mean == 0)
```
sethrj: This function is getting long and deeply nested. I think you should have a function
```c++
ImportGaussianScintComponent load_gauss_scint(std::string const& prefix, detail::GeantMaterialPropertyGetter& get_property);
```
that tries to load both `...LAMBDAMEAN` and `...LAMBDASIGMA` (and prints an error message if exactly one is present), and then, after you check for the grid (*without* an `else`):
```c++
ImportGaussianScintComponent gauss;
if (gauss = load_gauss_scint("CELER_" + prefix))
{
    CELER_VALIDATE(std::holds_alternative<std::monostate>(scint_data),
         << "conflicting/redundant scintillation properties for " << prefix");
    scint_data = std::move(gauss);
}
if (gauss = load_gauss_scint(prefix))
{
    CELER_VALIDATE(std::holds_alternative<std::monostate>(scint_data),
         << "conflicting/redundant scintillation properties for " << prefix");
    CELER_LOG(warning) << "Deprecated property prefix " << prefix << "LAMBDA: use CELER_" << prefix;
    scint_data = std::move(gauss);
}
```

This handles the case of user overspecifying the scintillation.

sethrj: We should be able to entirely delete the `gauss.lambda_mean == 0` condition,, as well as MomentCalculator.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -286,19 +300,26 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
+        {
+            if (auto* g = std::get_if<ImportGaussianScintComponent>(&scint_data))
+                comp.gauss = *g;
+            else if (auto* gr = std::get_if<inp::Grid>(&scint_data))
+                comp.spectrum = *gr;
+            // Note that the user may be missing some properties: in that
+            // case (if Geant4 didn't warn/error/die already) then we will
+            // rely on the downstream code to validate.
             components.push_back(std::move(comp));
         }
```
sethrj: The `if` here means the whole component is ignored if the user doesn't specify a spetctrum, whicih contradicts the internal comment. We should perhaps always add it ? (in that case, remove the `any_found` condition and unindent the code)

src/celeritas/io/ImportOpticalMaterial.hh
```diff
@@ -22,19 +22,30 @@ namespace celeritas
...
-    double yield_frac{};  //!< Fraction of total scintillation yield
     double lambda_mean{};  //!< Mean wavelength [len]
     double lambda_sigma{};  //!< Standard deviation of wavelength [len]
+    explicit operator bool() const
+    {
+        return (lambda_mean > 0 && lambda_sigma > 0);
+    }
+};
+
+struct ImportScintComponent
```
sethrj, line 39: Add documentation about the choice between the two spectrum desriptions

src/celeritas/optical/gen/detail/MatScintSpecInserter.hh
```diff
@@ -92,13 +93,38 @@ auto MatScintSpecInserter::operator()(ImportMaterialScintSpectrum const& mat)
         CELER_VALIDATE(comp.yield_frac > 0,
                        << "invalid yield=" << comp.yield_frac);
 
+        NonuniformGridInserter insert_energy_cdf(&data_->reals,
+                                                 &data_->energy_cdfs);
```
sethrj: All builders/inserters for an `Inserter` class should be *member data* rather than *local*, because they include logic for deduplicating the grids. `NonuniformGridInserter insert_energy_cdf_` will replace `Data* data_`.

src/celeritas/optical/gen/detail/MatScintSpecInserter.hh
```diff
@@ -92,13 +93,38 @@ auto MatScintSpecInserter::operator()(ImportMaterialScintSpectrum const& mat)
...
+
         ScintRecord scint;
-        scint.lambda_mean = comp.lambda_mean;
-        scint.lambda_sigma = comp.lambda_sigma;
         scint.rise_time = comp.rise_time;
         scint.fall_time = comp.fall_time;
+        if (comp.spectrum)
+        {
```
sethrj, line 104: You might want to add some validation of the user input here (monotonic positive x values, nonnegative y values)

src/celeritas/optical/gen/ScintillationData.hh
```diff
@@ -128,6 +129,8 @@ struct ScintillationData
     //! Material-dependent scintillation spectrum data [OptMatId]
     OptMatItems<MatScintSpectrum> materials;
 
+    // Cumulative probability of emission as a function of energy
```
sethrj:
```suggestion
    // Cumulative probability of emission as a function of energy [MeV]
```

src/celeritas/optical/gen/ScintillationGenerator.hh
```diff
@@ -81,6 +81,7 @@ class ScintillationGenerator
 
     GeneratorDistributionData const& dist_;
     NativeCRef<ScintillationData> const& shared_;
+    // NonuniformGridCalculator calc_cdf_;
```
sethrj: Delete dead code

test/celeritas/optical/Scintillation.test.cc
```diff
@@ -434,6 +470,50 @@ TEST_F(MaterialScintillationTest, stress_test)
...
+            EXPECT_SOFT_EQ(cdf_grid.front(), calc_energy(0));
+            EXPECT_SOFT_EQ(cdf_grid.back(), calc_energy(1));
+
+            std::vector<real_type> energy;
+            Rng rng;
+            for ([[maybe_unused]] auto i : range(4))
+            {
+                energy.push_back(calc_energy(generate_canonical(rng)));
+            }
+            // real_type energy = calc_energy(generate_canonical(rng));
```
sethrj:
```suggestion
```

sethrj (approved):
This is great! Good testing too. I have mainly a couple suggestions for improving the construction.

src/celeritas/ext/GeantImporter.cc
Rashika-Gupta: During testing I found that some components contain no meaningful data (e.g., no spectrum and all timing/yield values default to zero), which causes a validation failure such as invalid `fall_time=0` triggered through `MatScintInserter`. Hence, adding minimal guard so we push components only if either some component was provided.

src/celeritas/ext/GeantImporter.cc
```diff
@@ -194,6 +194,33 @@ auto& optical_particles_map()
     return map;
 }
 
+//---------------------------------------------------------------------------//
+//! Custom-defined scintillation properties approximating to Gaussian
+//! distribution
```
sethrj, line 199: What do you mean approximating?

src/celeritas/ext/GeantImporter.cc
```diff
@@ -222,82 +249,64 @@ fill_vec_import_scint_comp(detail::GeantMaterialPropertyGetter& get_property,
...
+                std::holds_alternative<std::monostate>(spectrum_component),
+                << "conflicting/redundant scintillation properties "
+                   "for "
+                << prefix);
+            CELER_LOG(warning) << "Deprecated property prefix " << prefix
+                               << "LAMBDA: use CELER_" << prefix;
+            spectrum_component = std::move(gauss);
+        }
+        if (get_property(&grid, name, {ImportUnits::mev, ImportUnits::unitless}))
+        {
```
sethrj, line 285: This should also have a VALIDATE I think?

test/celeritas/optical/OpticalCollector.test.cc
```diff
@@ -468,8 +468,8 @@ TEST_F(LArSphereOffloadTest, host_generate_small)
 
     if (CELERITAS_REAL_TYPE == CELERITAS_REAL_TYPE_DOUBLE)
     {
-        constexpr unsigned int expected_steps = using_surface_vg ? 109 : 121;
-        constexpr unsigned int expected_step_iters = using_surface_vg ? 4 : 5;
+        constexpr unsigned int expected_steps = using_surface_vg ? 109 : 116;
+        constexpr unsigned int expected_step_iters = 4;
```
sethrj, line 472: The next time we hit a similar failure @amandalund @Rashika-Gupta @hhollenb we should just add `&& !using_surface_vg` to the conditional since this is a meaningless regression check. The extra validation for the vecgeom surface model here doesn't tell us anything useful.

sethrj (approved):

test/celeritas/optical/OpticalCollector.test.cc
Rashika-Gupta: Would you like me to change it here and add the `if` condition ,or keep it as it is ?

src/celeritas/ext/GeantImporter.cc
Rashika-Gupta: Adding it in `MatScintSpecInserter` because, that is where all the validation of the scintillation data is done, before inserting it.

test/celeritas/optical/OpticalCollector.test.cc
sethrj: Since it's passing for now, just keep it :)

src/celeritas/ext/GeantImporter.cc
sethrj: Makes sense!

sethrj:
@Rashika-Gupta could you merge upstream/develop and mark as "resolved" all the comments you've addressed? Thanks!

PR #2101, opened 2025-11-08
Support timing optical actions

amandalund:
This adds a class that will create aux state data to accumulate action times (related to #1556) and uses it to record the times in the optical transporter. I've also changed the timing results from `real_type` to `double` in a couple places. If this looks ok I can update the core action timers to use this as a follow-on.


src/celeritas/optical/Transporter.cc
```diff
@@ -73,22 +80,31 @@ void Transporter::transport_impl(CoreState<M>& state) const
...
             ScopedProfiling profile_this{action->label()};
-            action->step(*params_, state);
+            Stopwatch get_time;
+            action->step(*this->params(), state);
+            if (accum_time)
+            {
+                if (M == MemSpace::device)
+                {
+                    device().stream(state.stream_id()).sync();
+                }
```
sethrj, line 90: A nice little to-do (maybe for @LSchwiebert ?) would be for the action time to always report the CPU time (including internal synchronizations with CUDA) and to have a second vector when doing a GPU run that uses the CUDA event API to record device times.

src/celeritas/user/ActionTimes.hh
```diff
@@ -0,0 +1,74 @@
...
+#include "corecel/data/AuxStateVec.hh"
+
+namespace celeritas
+{
+class ActionRegistry;
+struct ActionTimesState;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Manage state data for accumulating action times.
```
sethrj, line 25: Another couple sentences here would definitely help those with less context than you and me :)

src/celeritas/optical/Transporter.cc
```diff
@@ -100,6 +116,19 @@ void Transporter::transport_impl(CoreState<M>& state) const
     ++state.accum().flushes;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Get the accumulated action times.
+ */
+auto Transporter::action_times(AuxStateVec const& aux) const -> MapStrDbl
```
sethrj: I think renaming to `get_action_times` would be helpful to indicate that it's not a pure accessor; same with the `action_times` method being called (which I at first thought was returning a cheap reference).

src/celeritas/user/ActionTimes.hh
```diff
@@ -0,0 +1,74 @@
...
+ */
+class ActionTimes : public AuxParamsInterface
+{
+  public:
+    //!@{
+    //! \name Type aliases
+    using SPActionRegistry = std::shared_ptr<ActionRegistry>;
```
sethrj, line 38: Since `ActionTimes` isn't an action and therefore isn't necessarily stored in the ActionRegistry, this *shouldn't* have to be a weak pointer (which was the case for the core registry), but I think it would be safer to make this a weak pointer here, especially since it's only used once per run.

src/celeritas/user/ActionTimes.cc
```diff
@@ -0,0 +1,66 @@
...
+    return std::make_unique<ActionTimesState>();
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Access the state.
+ */
+ActionTimesState const& ActionTimes::state(AuxStateVec const& aux) const
+{
+    return dynamic_cast<ActionTimesState const&>(aux.at(aux_id_));
```
sethrj: In `AuxStateVec.hh` there's a `get` method intended to reduce boilerplate like this and with more debugging help:

```suggestion
    return get<ActionTimesState>(aux,aux_id_);
```

sethrj (commented):
I'll finish reviewing later this morning, but great work so far and way to actually get code into the main branch, unlike my old abortive attempts https://github.com/sethrj/celeritas/commits/action-times/ 😛

src/celeritas/user/ActionTimes.cc
amandalund: Oh right, forgot about that one!

src/celeritas/setup/Problem.cc
```diff
@@ -377,6 +379,10 @@ auto build_optical_offload(
     oc_inp.buffer_capacity = ceil_div(cap.generators, num_streams);
     oc_inp.auto_flush = ceil_div(cap.primaries, num_streams);
     oc_inp.max_step_iters = p.tracking.limits.optical_step_iters;
+    if (p.control.device_debug)
+    {
+        oc_inp.action_times = p.control.device_debug->sync_stream;
+    }
```
sethrj: Don't we also want the ability to store action_times to be stored when running host?

src/celeritas/setup/Problem.cc
```diff
@@ -689,6 +695,17 @@ ProblemLoaded problem(inp::Problem const& p, ImportData const& imported)
...
+                    if (p.control.device_debug
+                        && p.control.device_debug->sync_stream)
+                    {
+                        // Create aux data to accumulate optical action times
+                        AuxParamsRegistry& aux = *core_params->aux_reg();
+                        inp.action_times = std::make_shared<ActionTimes>(
+                            aux.next_id(),
+                            optical_params->action_reg(),
+                            "optical-action-times");
+                        aux.insert(inp.action_times);
```
sethrj: I think a `make_and_insert` method (given aux, action reg) would make this more foolproof

sethrj (approved):
Thanks @amandalund!

src/celeritas/setup/Problem.cc
amandalund: OH good point, we definitely do. Should we always store CPU action times?

src/celeritas/user/ActionTimes.hh
```diff
@@ -0,0 +1,93 @@
...
+ * Accumulated action times on each thread.
+ *
+ * \todo Always report CPU times and add a second map for device runs that uses
+ * the CUDA event API to record GPU times.
+ */
+struct ActionTimesState : public AuxStateInterface
+{
+    using MapIdDbl = std::unordered_map<ActionId, double>;
+
+    MapIdDbl accum_time;
```
sethrj: I think incrementing times on CPU is good, but if you're going to be updating every action at every step on CPU, use an array (the action IDs aren't going to be very sparse; certainly less than the overhead of a map). The final `get_action_times` can filter out zeros.

src/celeritas/setup/Problem.cc
sethrj: Yeah I think that's a good idea.

sethrj (approved):

sethrj:
Thanks @amandalund ! Sorry I forgot to follow up with your comment till this morning.

amandalund:
No worries, thanks!

PR #2102, opened 2025-11-09
Use dot instead of mermaid

sethrj:
Full circle from #1988, mmdc and sphinx-mermaid were causing too much trouble and required a whole installation of chromium to be downloaded to the CI (yuck). This just uses nice clean simple `dot`.


sethrj:
Quick documentation change for you @pcanal :) It decreases the doc build time on CI by 20%

doc/CMakeLists.txt
```diff
@@ -247,28 +246,20 @@ file(GLOB _DOXYGEN_SOURCE
...
   "${PROJECT_SOURCE_DIR}/src/celeritas/optical/gen/*.hh"
-  "${PROJECT_SOURCE_DIR}/src/celeritas/optical/gen/*.cc"
   "${PROJECT_SOURCE_DIR}/src/celeritas/optical/interactor/*.hh"
   "${PROJECT_SOURCE_DIR}/src/celeritas/optical/model/*.hh"
-  "${PROJECT_SOURCE_DIR}/src/celeritas/optical/model/*.cc"
   "${PROJECT_SOURCE_DIR}/src/celeritas/optical/surface/*.hh"
-  "${PROJECT_SOURCE_DIR}/src/celeritas/optical/surface/*.cc"
   "${PROJECT_SOURCE_DIR}/src/celeritas/optical/surface/model/*.hh"
-  "${PROJECT_SOURCE_DIR}/src/celeritas/optical/surface/model/*.cc"
   "${PROJECT_SOURCE_DIR}/test/*.hh"
```
pcanal, line 262: Should we also remove/hide the test headers?

doc/CMakeLists.txt
```diff
@@ -282,29 +273,28 @@ doxygen_add_docs(doxygen_xml
...
 
-# Note that *all* mermaid files must be present at the same time to prevent
+# Note that *all* dot files must be present at the same time to prevent
 # errors when doing separate HTML/PDF builds: the doctrees directory caches the
 # file listing and will fail if only one image type is generated
-set(_mermaid_output)
-set(_png_fallback "${CMAKE_CURRENT_SOURCE_DIR}/_static/mermaid-not-found.png")
-set(_output_base "${CMAKE_CURRENT_SOURCE_DIR}/_static/mermaid")
+set(_dot_output)
+set(_png_fallback "${CMAKE_CURRENT_SOURCE_DIR}/_static/dot-not-found.png")
```
pcanal, line 280: Does it file already exist in the repository (it does not appear to be there)?

doc/CMakeLists.txt
```diff
@@ -282,29 +273,28 @@ doxygen_add_docs(doxygen_xml
...
+set(_output_base "${CMAKE_CURRENT_SOURCE_DIR}/_static/dot")
```
pcanal, line 281: (Admittedly that is already the case before), it feels 'odd' to have the output base locate int he source directory.

doc/CMakeLists.txt
```diff
@@ -282,29 +273,28 @@ doxygen_add_docs(doxygen_xml
...
 
-macro(_celer_build_mermaid inp)
+macro(_celer_build_dot inp)
   get_filename_component(basename "${inp}" NAME_WE)
-  if(MMDC_EXECUTABLE)
-    # Generate SVG and PDF for HTML, PDF respectively
-    foreach(_fmt pdf svg)
+  if(DOXYGEN_DOT_EXECUTABLE)
```
pcanal, line 285: (minor) Why the link between `Doxygen` and `Dot`?

CMakeLists.txt
```diff
@@ -537,12 +542,14 @@ if(CELERITAS_BUILD_DOCS)
...
-    message(STATUS "Could NOT find mermaid generator `mmdc`")
+  if(NOT DOXYGEN_DOT_EXECUTABLE)
+    find_program(DOXYGEN_DOT_EXECUTABLE dot)
+  endif()
+  if(DOXYGEN_DOT_EXECUTABLE AND NOT _prev_dot)
+    message(STATUS "Found dot: ${DOXYGEN_DOT_EXECUTABLE}")
+  elseif(_prev_dot STREQUAL "")
+    # Note: use string comparison, not boolean, to avoid matching old "NOTFOUND"
+    message(STATUS "Could NOT find graphviz generator `dot`")
   endif()
```
pcanal, line 553: If I read this correctly: 
When `DOXYGEN_DOT_EXECUTABLE` is not found now but was found before (i.e. `_prev_dot` is set), there is no message **AND** `DOXYGEN_DOT_EXECUTABLE` is an un-usable state.  I.e. do we need something like:
```suggestion
      set(DOXYGEN_DOT_EXECUTABLE "${_prev_dot}")
  endif()
```
?

pcanal (approved):
LGTM (although I did not verify the dot syntax).

doc/CMakeLists.txt
sethrj: Ah the user docs do include an overview of testing, includign the Test class and some of the custom macros.

doc/CMakeLists.txt
sethrj: Ah thank you, I knew there was somethign I forgot to do 😄

doc/CMakeLists.txt
sethrj: It does, but as far as I can tell there is no way around it :(

doc/CMakeLists.txt
sethrj: CMake's FindDoxygen automatically finds `dot`, and it includes some extra logic about where to look; so we might as well piggyback off it.

CMakeLists.txt
sethrj: > When DOXYGEN_DOT_EXECUTABLE is not found now but was found before

This condition shouldn't be possible; once a cache variable is set, `find_executable` doesn't check or change it. I just tried by changing `DOXYGEN_DOT_EXECUTABLE` to something bogus, and the variable remained set to the same value rather than empty.

pcanal: Then `prev_dot` is solely recorded whether it had been found before ...

doc/CMakeLists.txt
pcanal: Which product (`CMake`, `dot`, ...) is imposing this? (i.e. why/where
```
set(_output_base "${CMAKE_CURRENT_BINARY_DIR}/_static/dot")
```
fails)?

CMakeLists.txt
sethrj: Well, it encodes an "optional boolean": not defined, false value, or true value.

doc/CMakeLists.txt
sethrj: Sphinx requires it; you have to give a path relative to the source directory for images ☹️

CMakeLists.txt
pcanal: (I remember you prefer restricting the nesting, so this might be moot):

Would this be equivalent with 'simpler' conditional?
```
  if(_prev_dot STREQUAL "")  # Note: use string comparison, not boolean, to avoid matching old "NOTFOUND"
	  if(DOXYGEN_DOT_EXECUTABLE)
	    message(STATUS "Found dot: ${DOXYGEN_DOT_EXECUTABLE}")
	  else   
	    message(STATUS "Could NOT find graphviz generator `dot`")
	  endif()
  endif()
```

doc/CMakeLists.txt
pcanal: Thanks.

(insert picture of a Sphinx being dizzy trying to answer its own question)

CMakeLists.txt
sethrj: I think so? Stupid CMake booleans, I can't remember if you can match a copied "unset" variable with `STREQUAL ""`...

Issue #2103, opened 2025-11-09
Directly offload to optical TrackingManager for validation/testing


sethrj:
Add support for using the existing tracking manager (and integration) to offload optical photons directly from Geant4 into the optical tracking loop.

Suggested approach:
- Add an optical primary generator[^1] that stores a vector of initial states (position, direction, wavelength) as aux data, and the "generator" instantiates tracks from those. @hhollenb will create this by adapting from his unit test.
- Subclass the `OffloadInterface` [^2] with another interface class `TrackOffloadInterface`, maybe rename `LocalOpticalOffload` to `LocalOpticalGenOffload`, add `LocalOpticalTrackOffload` that subclasses `TrackOffloadInterface`
- Update `celeritas::TrackingManager` to take a `TrackOffloadInterface` rather than the EM-specific one
- Update `TrackingManagerConstructor` so that instead of its `LocalTransporterFromThread` function returning `integration.local_transporter()`, it will return `integration.optical_offload` if the `opts.offload_particles` is set to optical photons
- For now the optical params setup can create generators for optical, cherenkov, and the "primary"; but as an optimization we should only create for the modes that are supported (done in the constructor of the different OffloadInterface subclasses)

I will definitely need @amandalund's view of whether this is sensible and fits into our structure without too much hassle, and I will get #2099 and company reviewed pronto.

[^1]: Our names for this are really confusing right now because of how the primaries/initialzers/generators/optical generators have evolved. I think "generator" is what we should name an object/action that creates tracks/initializers. I hesitate to use "primary" at all because an optical photon could have many "primaries" (suppose that we have a photon that came from a wavelength shifted photon, which came from a scintillation photon, which came from an EM track secondary in Celeritas, which came from an EM particle offloaded from Geant4, which came from an actual hadronic primary.) The current "optical primary generator" should perhaps be a "general distribution generator" or something.

[^2]: Speaking of names, this might be a good time to rename `LocalTransporter` to `LocalOffload` (and perhaps alias to `LocalEmOffload` for the time being since we don't support hadrons)

amandalund:
I think this is the right approach. A couple thoughts:
  - It makes sense for `LocalTransporter` and `LocalOpticalTrackOffload` to inherit from a `TrackOffloadInterface` since they should be used in the same way. I expect the implementation of `LocalOpticalTrackOffload` will be nearly identical to `LocalOpticalOffload` though.
[...]

PR #2104, opened 2025-11-10
Add direct optical photon generator

hhollenb:
For issue #2103 

This PR adds a generator for directly initializing optical photons from track initializers. For simplicity only one photon is generated per initializer.


src/celeritas/optical/gen/DirectGeneratorAction.cc
```diff
@@ -0,0 +1,237 @@
...
+    {
+        // Reallocate with enough capacity
+        aux_state.store = CollectionStateStore<DirectGeneratorStateData, M>{
+            state.stream_id(), static_cast<size_type>(data.size())};
+    }
+
+    // Update counters and copy distributions to aux state storage
+    aux_state.counters.buffer_size = data.size();
+    aux_state.counters.num_pending = data.size();
+    state.counters().num_pending = data.size();
```
amandalund: I don't think it should matter here because

1. I expect this will only be called at the start of the stepping loop, and
2. I think the only other generator we'd use concurrently with this one is WLS, which we don't have yet,

but since this is number of pending photons over all the generators:
```suggestion
    state.counters().num_pending += data.size();
```
That's my bad since I did this in the PrimaryGeneratorAction: would you mind updating that one too?

amandalund (approved):
Very nice! Thanks @hhollenb.

sethrj:
I'll defer review of this to @amandalund , but @Rashika-Gupta can you please take a look and ask any questions you might have?

hhollenb:
For the generators test, there's some failures when using the Geant4 geometry, which I think is expected since surface normals don't work correctly. Not sure why it's passing for some of the other tests (maybe larger state sizes hide the fluctuations?) but I can disable it when using Geant4 geometry or override the surface physics to just absorb and ignore the surface normals.

sethrj:
@hhollenb Just have the test ignore the failing bits for the G4 geometry, which is mostly for validation. Our layer on top the G4Navigator doesn't seem to play nicely with "surface crossings" because G4 doesn't think of them that way.

test/celeritas/optical/Generator.test.cc
```diff
@@ -168,6 +169,46 @@ TEST_F(LArSphereGeneratorTest, primary_generator)
...
+                                  Real3{0, 0, 0},
+                                  Real3{1, 0, 0},
+                                  Real3{0, 1, 0},
+                                  0,
+                                  ImplVolumeId{0}});
+    auto generate = optical::DirectGeneratorAction::make_and_insert(
+        *this->core(), *this->optical_params());
+
+    this->build_state<MemSpace::host>(32);
+    this->build_transporter();
```
amandalund: The transporter should be built before the state now so the aux data is correctly constructed:
```suggestion
    this->build_transporter();
    this->build_state<MemSpace::host>(32);
```

Issue #1898, opened 2025-08-11
Support leptonuclear and "extra" EM physics


sethrj:
Celeritas doesn't currently support the models and processes implemented as "extra" physics. This shows up in #1756 as fewer steps with slightly less energy deposition, and it is non-negligible in ATLAS. 

Suggested work plan for rpogressive capability:

- [ ] Log a warning if *any* processes that apply to Celeritas-managed particle types are ignored
- [ ] Load energy, material-dependent cross sections from processes that aren't implemented and combine into an "absorption" process to locally deposit energy (preserving total cross section of the particle type)
- [ ] Instead of just absorbing, reconstruct a G4Track from the current state (see https://github.com/celeritas-project/celeritas/issues/1894), get the interaction length for the G4 "extra" processes that apply, sample proportionally based on those, and call "DoIt" on the sampled process (?) and complete the rest of the G4 stepping loop on CPU, including hits and stacking G4 secondaries
-  ~~Implement the process in Celeritas but offload the secondary to Geant4~~

| Particle Type | Process | Model/Implementation | Notes |
|---------------|---------|---------------------|-------|
| **Gamma (γ)** | Photonuclear | `G4HadronInelasticProcess` | Uses `G4GammaNuclearXS` |
| | | `G4LowEGammaNuclearModel` | Low energy (< 200 MeV) |
| | | `G4CascadeInterface` (Bertini) | Moderate energies |
| | | `G4TheoFSGenerator` + `G4QGSModel` | High energies with string fragmentation |
| **Electron (e⁻)** | Electronuclear | `G4ElectronNuclearProcess` | Uses `G4ElectroVDNuclearModel` |
| **Positron (e⁺)** | Positronuclear | `G4PositronNuclearProcess` | Uses `G4ElectroVDNuclearModel` |
| **Muon (μ⁺/μ⁻)** | Muon-nuclear | `G4MuonNuclearProcess` | Uses `G4MuonVDNuclearModel` |

whokion:
The implementation of the associated lepto-nuclear models (including sampling of final states) should be outside the current scope, as it requires a substantial development effort. For example, adding the gamma–nuclear interaction of a Bertini-like cascade model also requires implementing intra-nuclear interactions for all other hadrons and the following de-excitation models — the same applies to QGS.
[...]

PR #2105, opened 2025-11-11
Add initial skeleton implementation of the gamma-nuclear process

whokion:
## Description

This MR introduces the initial set of gamma-nuclear onloading code, including:
- skeleton implementation of the gamma-nuclear process, model, and related classes and updates
- low-energy gamma-nuclear cross-section calculator
- basic unit tests and reference data


sethrj:
Tagging #1898.

Thanks @whokion ! What made you decide that a tabulated cross section approach (i.e., what G4HepEm does) is insufficiently accurate?

whokion:
@sethrj It will uses a hybrid approach: tabulated cross sections at low energies (below 130 MeV, on a non-uniform linear grid, this MR) and on-the-fly calculations at high energies (above 150 MeV, based on parameterized data in a logarithmic grid, to be added in the next MR). The electro-nuclear cross section also follow a different grid style of parameterized calculation. Since the cross section data are built during model initialization, table lookups using Geant4-equivalent grids may be computationally inexpensive for step competition. We can revisit the data organization later, but for now, I plan to implement the Geant4-equivalent version first for validation, assuming it introduces minimal performance overhead.

sethrj:
@whokion I was less concerned with performance overhead than simplicity; the grid approach should let us calculate the cross sections with a simple call to Geant4, once, during setup. It would require no extra file readers, micro/macro xs calculators, etc. Ultimately I agree the in-house calculation is the ideal approach, but you've already skipped several steps in the proposed plan from #1898, which risks delaying the capability. 😕

whokion:
This approach is essentially the same as the pre-built table method — we still need to read the data once and compute the cross section on the fly (using the default XS calculator). I don’t think merging all different data styles or binning schemes into a single uniform structure, especially for hadronic data, is a good idea for the sake of simplicity.  If you still prefer to use the EM-like lambda (xs) table for the gamma- and lepto-nuclear processes, I would need @stognini help to import them directly from Geant4, since I don’t have a clear idea of how to combine them effectively across all valid energy ranges without oversimplifying the structure.

Can you clarify what I have skipped several steps in the proposed plan from https://github.com/celeritas-project/celeritas/issues/1898 ?

Also, should I worry about the error from widows related to the data file path,
`Failed to open test data file at 'D:/a/celeritas/celeritas/test/celeritas/data/gamma-nucl/'` ?
Should I move data files to the `data` directory?

sethrj:
> This approach is essentially the same as the pre-built table method — we still need to read the data once and compute the cross section on the fly (using the default XS calculator).

Perhaps I'm not understanding what you or Mihaly are doing: the hadronic tables have a function `hadGNucXSDataStore->ComputeCrossSection(dyGamma, g4MatCut->GetMaterial())` for calculating the total gamma-nuclear cross section, used [here](https://github.com/mnovak42/g4hepem/blob/0b1c7ecea142e5fa0867d39faaaf7f360f7fce24/G4HepEm/G4HepEmInit/src/G4HepEmGammaTableBuilder.cc#L123). With the per-material cross section tabulated, we don't need any on-the-fly: we add the interpolated gamma-nuclear XS into the total while tracking, and if we hit that, then we send the track back to the CPU to sample the element, isotope, etc.

> I don’t think merging all different data styles or binning schemes into a single uniform structure, especially for hadronic data, is a good idea for the sake of simplicity. If you still prefer to use the EM-like lambda (xs) table for the gamma- and lepto-nuclear processes, I would need @stognini help to import them directly from Geant4, since I don’t have a clear idea of how to combine them effectively across all valid energy ranges without oversimplifying the structure.

Judging by what g4hepem gets away with, I think the structure can be quite simplified (this is why I asked about the cross section visualization). For HEP applications these processes might be unimportant enough to get away with pretty large errors in the cross section.

> Can you clarify what I have skipped several steps in the proposed plan from #1898 ?

- The first simple step was identifying the existence of "extra" (including gamma-nuclear) processes that apply to e-/e+/gamma, and logging them would tell us that our tests are set up correctly to include the processes, and tell our users that the physics is wrong until they get implemented. I expected that to be a separate small PR.
- The second step was to pull the EM->hadronic cross sections from `G4CrossSectionDataStore` and use it to adjust the total cross section. It would add the process but return "absorption" instead of a new "onload" type. The interactor would look like the one you've got, but we don't need to select the element on GPU nor to do any on-the-fly calculation.

> Also, should I worry about the error from widows related to the data file path, `Failed to open test data file at 'D:/a/celeritas/celeritas/test/celeritas/data/gamma-nucl/'` ? Should I move data files to the `data` directory?

This indicates `test_data_path` isn't being used as it's supposed to: the second argument is `filename` not `subdirectory`. I guess on UNIX-y systems, you can "open" the directory name and it's reported as good, but windows says no, `subdir/` is not a file and it will not open it. Since your reader function requires a directory name, you'll need to `#include "celeritas_test_config.h"` and use `path = celeritas_source_dir + "/test/celeritas/data/gamma-nucl/"`

whokion:
@sethrj Sorry for the belated response - I am still in the middle of wrapping up some other tasks, but I expect to follow up in more closely in the next couple of days. As a quick note on your comments and proposal: I am a bit confused about what our targeted reference is, G4HepEM or Geant4? My intended implementation plan is to follow Geant4, since it handles cross sections in general at the element level rather than the material level for hadronic physics. That is at least my understanding, but I can double-check what the correct approach should be for this case.  Regarding onloading (particles), we do not send back e+/e- for lepto-nuclear interactions to CPU, but still processing them on GPU. We only return photons to the CPU from both gamma-nuclear and lepto-nuclear interations. The electro-nuclear interactor updates the final state of the e-/e+ on GPU and generates a secondary (virtual) photon that will be returned to the CPU for the post gamma-nuclear interaction (by [LEND]/bertini/QGS models). This avoids special handling for these electron/positron stepping further on the CPU (i.e. having two segmented tracks, the first in GPU and the subsequent second in CPU from the same primary). The onloaded photon from either gamma-nuclear or electro-nuclear is effectively captured by the nucleus in the same way via an inelastic interaction and produces many soft photons, which we were missing in earlier Celeritas validation.

Anyway, I will follow up with more details later, but want to make sure that we are thinking along the same track. Personally, I would like to avoid any shortcuts that compromise physics quality, aside from reorganizing cross section calculations or implementing the electron/photon interactors more efficiently.

sethrj:
Thanks @whokion , it's not urgent.

>  As a quick note on your comments and proposal: I am a bit confused about what our targeted reference is, G4HepEM or Geant4? My intended implementation plan is to follow Geant4, since it handles cross sections in general at the element level rather than the material level for hadronic physics. That is at least my understanding, but I can double-check what the correct approach should be for this case. 

Our immediate target is to get the minimum level of capability needed to get the LHC experiments validated. Mihaly has been able to convince everyone that G4HepEm is "accelerated Geant4 physics", so we too can give ourselves a lot of leeway. I think since the beginning I've argued that we don't need to do things exactly the same as Geant4, especially since we're targeting GPU execution. Unless we're running depleted hot reactor cores with dozens of nuclides at dozens of temperatures, it may make more sense for us to pre-calculate a grid of material cross sections. I think that's especially true if we're not going to do element-based physics on the GPU for those processes.

> Regarding onloading (particles), we do not send back e+/e- for lepto-nuclear interactions to CPU, but still processing them on GPU. We only return photons to the CPU from both gamma-nuclear and lepto-nuclear interations. The electro-nuclear interactor updates the final state of the e-/e+ on GPU and generates a secondary (virtual) photon that will be returned to the CPU for the post gamma-nuclear interaction (by [LEND]/bertini/QGS models). This avoids special handling for these electron/positron stepping further on the CPU (i.e. having two segmented tracks, the first in GPU and the subsequent second in CPU from the same primary). The onloaded photon from either gamma-nuclear or electro-nuclear is effectively captured by the nucleus in the same way via an inelastic interaction and produces many soft photons, which we were missing in earlier Celeritas validation.

Ahhh ok, I clearly didn't understand your comment from https://github.com/celeritas-project/celeritas/issues/1898#issuecomment-3398330492 . In that case your approach makes a lot of sense! Is it possible to use the existing Geant4 mechanics to do *just* the final state sampling?

> Anyway, I will follow up with more details later, but want to make sure that we are thinking along the same track. Personally, I would like to avoid any shortcuts that compromise physics quality, aside from reorganizing cross section calculations or implementing the electron/photon interactors more efficiently.

I really appreciate you taking the time to explain this to me. I think at this stage it's OK for us to take shortcuts if the error is small or washes out in the final physics solution. But if you think we can get this approach done in the next month or so, let's do it!

whokion:
For a quick follow-up: 
1) Gamma-nuclear cross sections (PARTICLEXS, <130 MeV) in this MR is built from a mixture of evaluated photon libraries and is sensitive to each element’s binning structure. Pre-building tables per material (i.e., total cross section over a material with its own binning and energy points) can distort the original structure. Using the smallest bin size across all elements may preserve the original values, but this still needs to be validated. We can revisit that approach after the first implementation and potentially provide options for both approaches before finalizing the choice. I could spend several weeks validating the material-level table approach across a selected set of materials, but we cannot realistically study all combinations in priori. Also, we need an element-level cross-section calculator anyway for the electro-nuclear interactor, so keeping a consistent structure is a reasonable starting point. For parameterized cross-section calculations, I can consider using the total cross section table approach, but that requires deciding on an appropriate binning to retain the resonance structure - may be another multi-week study. My concern is avoiding any questionable bias without knowing the level of accuracy with a short cut. 

2)  We can definitely reuse the existing Geant4 `ApplyYourself` of gamma-nuclear models or even call the `PostStepDoIt` of the inelastic hadron process directly. This needs further investigation. 

3) Moving forward with the current approach will take less time (at least for me), allowing us to skip detailed cross-section-structure studies before producing a working implementation. Since the electro-nuclear process will follow a similar code structure, I believe end-of-year (or at least end of January) is a reasonable target for delivering the first version with some initial validation, considering the coming holiday season and my other ongoing commitments. Of course, it would be very helpful @stognini can provide imported material cross-section tables (with a choice of number of bins and at the different energy ranges with different interpolations) for the total cross section approach if it is the implementation choice.

src/celeritas/em/data/GammaNuclearData.hh
```diff
@@ -0,0 +1,90 @@
...
+    ParticleId gamma_id;
+
+    //! Model's maximum energy limit [MeV]
+    static CELER_CONSTEXPR_FUNCTION units::MevEnergy max_valid_energy()
+    {
+        return units::MevEnergy{5e+4};
+    }
+
+    //! Maximum energy limit for PARTICLEXS/gamma cross sections [MeV]
+    static CELER_CONSTEXPR_FUNCTION units::MevEnergy max_low_energy()
```
amandalund, line 35: This name is a little vague... maybe something like `tabulated_max_energy` or `otf_xs_threshold`?

src/celeritas/em/process/GammaNuclearProcess.hh
```diff
@@ -0,0 +1,67 @@
...
+    bool applies_at_rest() const final { return false; }
+
+    // Name of the process
+    std::string_view label() const final;
+
+  private:
+    SPConstParticles particles_;
+    SPConstMaterials materials_;
+    ReadData load_data_;
+    ParticleId gamma_id_;
```
amandalund: What is `gamma_id_` needed for?

src/celeritas/em/xs/GammaNuclearMicroXsCalculator.hh
```diff
@@ -0,0 +1,80 @@
...
+    CELER_EXPECT(el_id < shared_.micro_xs.size());
+
+    // Get element cross section data
+    NonuniformGridRecord grid = shared_.micro_xs[el_id];
+
+    // Calculate micro cross section at the given energy
+    NonuniformGridCalculator calc_xs(grid, shared_.reals);
+    real_type result = calc_xs(inc_energy_);
+
+    return BarnXs{result};
```
amandalund, line 76: I'm guessing this class will be expanded to include the on-the-fly calculation for higher energies?

test/celeritas/em/GammaNuclear.test.cc
```diff
@@ -0,0 +1,146 @@
...
+TEST_F(GammaNuclearTest, micro_xs)
+{
+    // Calculate the gamma-nuclear microscopic (element) cross section
+    using XsCalculator = GammaNuclearMicroXsCalculator;
+
+    // Set the target element: Pb
+    ElementId el_id{2};
+
+    // Check the size of the element cross section data (G4PARTICLEXS4.1)
+    //    GammaNuclearRef shared = model_->host_ref();
```
amandalund:
```suggestion
```

src/celeritas/phys/Interaction.hh
```diff
@@ -36,6 +36,7 @@ struct Interaction
         absorbed,  //!< Absorbed or transformed to another particle type
         unchanged,  //!< No state change, no secondaries
         failed,  //!< Ran out of memory during sampling
+        onloaded,  //!< Transfer interaction to the host
```
amandalund: I think this should go before `unchanged`: the `changed()` method below returns true if `action < Action::unchanged`, and we use this to return early from the `InteractionApplier`.

src/celeritas/em/executor/GammaNuclearExecutor.hh
```diff
@@ -0,0 +1,61 @@
...
+        elcomp_id = select_el(rng);
+        CELER_ASSERT(elcomp_id);
+        track.physics_step().element(elcomp_id);
+    }
+
+    // Construct the interactor
+    GammaNuclearInteractor interact(params, particle);
+
+    // Execute the interactor
+    return interact();
```
amandalund, line 57: Is the interactor going to do more in the future, or will it just return `Interaction::from_onloaded()`? If the latter, could we just return the interaction here?

amandalund (approved):
Thanks @whokion, looks like a good start!

src/celeritas/em/data/GammaNuclearData.hh
whokion: More parameters will be added when the high-energy cross sections are implemented. I’ll rename this accordingly once the final structure is defined.

src/celeritas/em/process/GammaNuclearProcess.hh
whokion: Okay, I removed it for now - I though that the input particle should be check as photon in the process level.

src/celeritas/phys/Interaction.hh
whokion: Move it before unchanged - I did not know that the ordering matters.

src/celeritas/em/executor/GammaNuclearExecutor.hh
whokion: Okay, I’ll keep this in mind and follow your suggestion if I do not add anything to the interactor (likely) when the next MR is ready for final review.  Thanks.

whokion:
@amandalund Thanks, as always, for your review! This is a rather large MR, but I hope it will be straightforward to review, as the additions generally follow the standard structure for physics processes.

src/celeritas/em/xs/GammaNuclearMicroXsCalculator.hh
whokion: Yes, this code will be expanded when cross section calculation for the high energy region is added.

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//---------------------------------------------------------------------------//
+#if !CELER_USE_DEVICE
+void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
+{
+    CELER_NOT_CONFIGURED("CUDA OR HIP");
+}
+#endif
+
```
stognini, line 100:
```suggestion
//!@}
```

whokion: @stognini Thanks. Will add this in the next MR as the branch is merged.

PR #2106, opened 2025-11-12
Update build script with best practices

sethrj:
This applies some fixes via chatgpt that I recognized were necessary after discussions with @pcanal on #2035 .

## Shebang and shell options
Changed `#!/bin/sh -e` to a plain shebang with an explicit `set -e` for full POSIX compliance.  
Many systems ignore or misinterpret flags after `#!/bin/sh`.

## Quoting and pathname safety
Quoted all variable expansions (`"$0"`, `"$src"`, `"$dst"`, etc.) in command arguments and messages to prevent breakage on spaces, glob characters, or special symbols.

## Safe handling of set -e
Rewrote command substitutions (e.g., `git_dir=$(...)`) so they do not cause the script to exit under `set -e` when commands fail.  
Used `if var=$(command)` style instead of checking `$?` afterward.

## Use of command -v instead of which
Replaced all uses of `which` and `hash` with the POSIX-compliant `command -v` for command existence and path lookup.

## Numeric validation in ccache usage
Added validation of numeric fields before performing arithmetic to prevent runtime errors and extraneous stderr output if `ccache --print-stats` fails or returns malformed data.

## Argument quoting in cmake and ctest calls
Quoted `$CMAKE_PRESET` and argument lists (`"$@"`) to handle presets or paths containing spaces safely.

## Consistent output formatting
Replaced non-portable `echo` usage with `printf '%s\n' ...` for predictable, POSIX-standard output behavior across different shells.

## Removed readlink dependency
~Eliminated non-POSIX use of `readlink`.~
Use a POSIX test for symlink, but print the target with `readlink`

## Fix HOSTNAME

The HOSTNAME variable isn't always exported to secondary processes; so `uname -n` is used instead.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -43,7 +43,7 @@ log() {
 fancy_hostname() {
   sys=${LMOD_SYSTEM_NAME}
   if [ -z "${sys}" ]; then
-    sys=$(uname -n)
+    sys="$(uname -n)"
     # Trim login/compute from head of string
     case "${sys}" in
       login[0-9]*.*|compute[0-9]*.*) sys=${sys#*.} ;;
@@ -62,7 +62,7 @@ ln_presets() {
 
   # Return early if it exists
   if [ -L "${dst}" ]; then
-    src=$(readlink "${dst}" 2>/dev/null || printf "<unknown>")
+    src="$(readlink "${dst}" 2>/dev/null || printf "<unknown>")"
     log debug "CMake preset already exists: ${dst} -> ${src}"
     return
   elif [ -e "${dst}" ]; then
@@ -145,12 +145,12 @@ OLD_CMAKE="$(command -v cmake 2>/dev/null || printf '')"
 OLD_PRE_COMMIT="$(command -v pre-commit 2>/dev/null || printf '')"
 
 # Load environment paths
-_env_script="scripts/env/${SYSTEM_NAME}.sh"
-if [ -f "${_env_script}" ]; then
-  log info "Sourcing environment script at ${_env_script}"
-  . "${_env_script}"
+ENV_SCRIPT="scripts/env/${SYSTEM_NAME}.sh"
+if [ -f "${ENV_SCRIPT}" ]; then
+  log info "Sourcing environment script at ${ENV_SCRIPT}"
+  . "${ENV_SCRIPT}"
 else
-  log debug "No environment script exists at ${_env_script}"
+  log debug "No environment script exists at ${ENV_SCRIPT}"
 fi
 
 NEW_CMAKE="$(command -v cmake 2>/dev/null || printf 'cmake unavailable')"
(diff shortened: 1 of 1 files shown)


pcanal:
> Added die() helper ....

This is no longer in the PR.

scripts/build.sh
```diff
@@ -50,42 +52,47 @@ fancy_hostname() {
...
   src="scripts/cmake-presets/$1.json"
   dst="CMakeUserPresets.json"
 
-  # Return early if they exist
-  if [ -e "${dst}" ]; then
-    src=$(readlink ${dst} || echo "<not a symlink>")
-    log debug "CMake preset already exists: ${dst} -> ${src}"
+  # Return early if it exists
+  if [ -L "$dst" ]; then
+    src=$(readlink "$dst" 2>/dev/null || printf "<unknown>")
```
pcanal: This seems to conflict with the PR description: "Removed readlink dependency"

scripts/build.sh
```diff
@@ -17,30 +17,32 @@
...
 
-  printf "\033[${color}m%s:\033[0m %s\n" "$level" "$message" >&2
+  printf "\033[%sm%s:\033[0m %s\n" "$color" "$level" "$message" >&2
 }
 
 # Get the hostname, trying to account for being on a compute node or cluster
 fancy_hostname() {
   sys=${LMOD_SYSTEM_NAME}
-  if [ -z "${sys}" ]; then
+  if [ -z "$sys" ]; then
```
pcanal: This may or may not be the right choice (I have been migrating the other direction for some reasons) but this does not seems described in the PR description.

sethrj: Yeah chatgpt did this and I guess I assumed there was a good reason for it? But looks like it was just for "cleanness":
```
When braces are optional

$var is fine if:
	•	Surrounded by whitespace, punctuation, or quotes:
echo "$var", cd "$dir", log info "$msg".

Rationale for change

Removing superfluous braces:
	•	Improves readability and reduces visual noise.
	•	Keeps the style consistent with idiomatic POSIX shell scripts (as in the POSIX sh and dash man pages, and most portable system scripts).
	•	Avoids suggesting unnecessary parameter expansion features.
```

So it's I guess "more idiomatic" vs "more internally consistent". Thoughts?

pcanal:
> $var is fine if: Surrounded by whitespace, punctuation, or quotes:

This is fair/fine. 

> So it's I guess "more idiomatic" vs "more internally consistent".

The rest of the code in that file uses `${var}` but without quotes, so I would have lean towards using `${var}` everywhere.

> Thoughts?

It is really a matter of preferences.  Both are fine.  The lowest effort is to just add a note in the PR description/commit (along the line of https://github.com/celeritas-project/celeritas/pull/2106#discussion_r2519308558)

pcanal (approved):
LGTM (see comments)

scripts/build.sh
sethrj: Quite so, I put it back in.

PR #2108, opened 2025-11-13
Use a single `ActionSequence` across all states

amandalund:
The updates the `ActionSequence` to use aux data for accumulating action times and moves to a single `ActionSequence` per params rather than one per `Stepper` (part of #1556).

### DIFF
--- a/src/celeritas/global/ActionSequence.cc
+++ b/src/celeritas/global/ActionSequence.cc
@@ -9,6 +9,7 @@
 #include <algorithm>
 #include <type_traits>
 #include <utility>
+#include <vector>
 
 #include "corecel/DeviceRuntimeApi.hh"
 
@@ -36,9 +37,6 @@ namespace celeritas
 ActionSequence::ActionSequence(ActionRegistry const& reg, Options options)
     : actions_{reg}, options_{std::move(options)}
 {
-    // Initialize timing
-    accum_time_.resize(actions_.step().size());
-
     // Get status checker if available
     for (auto const& brun_sp : actions_.begin_run())
     {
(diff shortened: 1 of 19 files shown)


src/celeritas/setup/Problem.cc
```diff
@@ -732,6 +736,19 @@ ProblemLoaded problem(inp::Problem const& p, ImportData const& imported)
                           "physics");
     }
 
+    // Construct the action sequence
```
amandalund, line 739: Should all the actions have been added to the registry by this point?

sethrj: Technically we allow actions until the states/transporter are created, but I think we ought to restrict it to during setup only (user callbacks will create actions). So let's add an assertion (maybe runtime in the action registry) that the action count hasn't changed since startup.

src/celeritas/global/ActionSequence.cc
```diff
@@ -76,13 +72,19 @@ void ActionSequence::begin_run(CoreParams const& params, CoreState<M>& state)
 template<MemSpace M>
 void ActionSequence::step(CoreParams const& params, CoreState<M>& state)
 {
-    // Save the stream for synchronizing after each step for timing
-    // NOTE: instead of synchronizing the stream we could add
-    // device timers to reduce the performance
+    // Save a pointer to aux data and the stream for synchronizing after each
+    // step for timing
+    // NOTE: instead of synchronizing the stream we could add device timers to
+    // reduce the performance
```
sethrj:
```suggestion
    // reduce the performance impact
```
legacy Seth 🐿️?

sethrj (commented):
LGTM thanks!

src/celeritas/setup/Problem.cc
amandalund: Ah yes I wasn't sure about those, that sounds good.

src/celeritas/global/ActionSequence.cc
amandalund: Stop trying to sabotage us!

sethrj (approved):
Awesome!

PR #2109, opened 2025-11-13
Reverse build order of protos

elliottbiondo:
This MR reverses the order in which protos are built such that child protos are built before their parents. This is the first step towards support for union boundaries as outlined in #1260.


elliottbiondo:
@sethrj this is ready for a first look. As best as I can tell we don't have any unit tests that actually resulted in initial guesses for bboxes affecting the final result 🤷

sethrj:
@elliottbiondo Could you try running the `models/analysis/orange-complexity.ipynb` notebook from the `benchmarks` repo? Hopefully it's clear enough how to run, and under `Load results` you can see the relevant failures with:
```py
failures = ~results.failure.isna()
results.failure[failures].xs('nav', level='optname').to_dict()
```

elliottbiondo:
@sethrj since this does not cause any additional failures in `models/analysis/orange-complexity.ipynb` relative to develop, I think you can continue your review.

src/orange/orangeinp/detail/ProtoBuilder.cc
```diff
@@ -72,9 +72,12 @@ void ProtoBuilder::save_json(JsonPimpl&& jp) const
...
 void ProtoBuilder::insert(VariantUniverseInput&& unit)
 {
-    CELER_EXPECT(inp_->universes.size() < protos_.size());
+    inp_->universes[current_uid_.get()] = std::move(unit);
 
-    inp_->universes.emplace_back(std::move(unit));
+    if (!this->is_global_universe())
+    {
+        --current_uid_;
+    }
```
sethrj, line 62: This makes me uneasy:
- It feels "surprising" to decrement a counter by inserting (I should've named it `push_back`)
- There's an `if` statement that applies exactly 1 out of N times if called as expected
- The behavior is inconsistent after the counter reaches zero (the global universe gets overwritten instead of anything detectable changing or an assertion firing)

I think you should take the usual "count backward with unisgned ints" approach where you initialize with the expected size, set `UniverseId::index_type num_inserted_{0}`, return `UniverseId{size_ - num_inserted_ - 1}` for `current_universe`, and increment the inserted count every time.  Either of `current_uid` or `insert` could raise an exception if it's past the number of expected.

src/orange/orangeinp/UnitProto.cc
```diff
@@ -436,7 +403,7 @@ void UnitProto::build(ProtoBuilder& input) const
         // parent-reference-frame bbox
         auto local_bbox = apply_transform(calc_inverse(iter->second.transform),
                                           result.volumes[vol_id.get()].bbox);
-        input.expand_bbox(iter->second.univ_id, bump_bbox(local_bbox));
+        pb.expand_bbox(iter->second.univ_id, bump_bbox(local_bbox));
```
sethrj: Do these bboxes even matter now? You might be able to delete all that related code.

src/orange/orangeinp/UnitProto.cc
```diff
@@ -225,16 +222,13 @@ auto UnitProto::daughters() const -> VecProto
  * Construction is done from highest masking precedence to lowest (reverse
  * zorder): exterior, then holes, then arrays, then media.
  */
-void UnitProto::build(ProtoBuilder& input) const
+void UnitProto::build(ProtoBuilder& pb) const
```
sethrj, line 228: FWIW the name was because it was oiginally passing directly the OrangeInput and pushing back onto the list of universes.

sethrj (approved):

src/orange/orangeinp/detail/ProtoBuilder.cc
elliottbiondo: I agree that incrementing `num_inserted` is the better option. However, I do think the `if` statement is still necessary, otherwise, after the last universe is inserted, this breaks:

```
UnivId ProtoBuilder::current_uid() const
{
    CELER_EXPECT(num_univs_inserted_ < num_univs_);
    return UnivId{num_univs_ - num_univs_inserted_ - 1};
}
```

It looks like `current_uid` isn't actually used anywhere after the final insertion, but this also feels surprising to me.

elliottbiondo:

src/orange/orangeinp/detail/ProtoBuilder.cc
sethrj: I see, I guess that is surprising but at least it's a clear failure? Maybe this all is just showing we're not doing the update in the wrong place: maybe the `InputBuilder` should just be manually setting the `UnitBuilder`'s state before each call to `build`.

sethrj (approved):
Thanks @elliottbiondo , looks like there's an unused value failure in https://github.com/celeritas-project/celeritas/actions/runs/19837033592/job/56836697612?pr=2109 ; please resolve and merge when you're satisfied!

src/orange/orangeinp/detail/ProtoBuilder.cc
elliottbiondo: That would mean that client code would be directly changing the internal state. Then it would be possible for `ProtoBuilder` to "lose its place," i.e., have `current_id` not point to the next available spot in the `universes` array.

I am having trouble coming up with a concise suggestion to fix this. Fundamentally I think it would be more straightforward for `void UnitProto::build(ProtoBuilder& pb) const` to instead return `VariantUniverseInput`, which would mean that `InputBuilder` would track the universe id, eliminating state from `ProtoBuilder`. We could actually eliminate the `ProtoBuilder` entirely with:

 `VariantUniverseInput UnitProto::build(ProtoMap const& protos, Options const& opts) const`

I know that one of the points of `ProtoBuilder` is facilitate a common interface with SCALE but we could achieve this by simply bundling arguments into a struct.

elliottbiondo: Also `UnitProto::build` is way to too long, and I think eliminating `ProtoBuilder` could be a first step in refactoring it.

sethrj: There's lots that can use refactoring 😅 but I agree this ProtoBuilder helper was not as elegant as I'd hoped.

sethrj:
@elliottbiondo one last error: https://github.com/celeritas-project/celeritas/actions/runs/19857710621/job/56901327530?pr=2109

recommended practice is to zero-initialize stuff in the header, and variables that aren't zero-initialized in the constructor

PR #2110, opened 2025-11-14
CudaRdcUtils v13 --no-undefined fix, add missing cudart

pcanal:
Add missing cudart for non-cuda library that depends on a cuda library.

Avoid no-undefined flag when constructing middle library

Also explicitly allow undefined symbols for rdc library (in case `—no-undefined` becomes the default for linker)

The cuda_rdc middle library has by definition many unresolved symbols. Those symbols are generated by nvlink whose result is only included in the final library.

If the user request `-Wl,--no-undefined` to be applied/used when generating a shared library, we must remove it for the middle library generation (we could not find a flag that took priority/reverted the `no-undefined` if/when it is passed to the linker).

This is required for building within the LHCb context where we have:

 ```
 set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--as-needed -Wl,--no-undefined")
 ```

This patch means that for CMake project that create any RDC libraries, and relies on setting `--no-undefined` via `CMAKE_SHARED_LINKER_FLAGS[_INIT]` then `--no-undefined` will be disabled globally (as opposed to 'just' for the CUDA rdc
 middle library).

Instead of modifying `CMAKE_SHARED_LINKER_FLAGS`, when it is available we can use `-z undefs`

### DIFF
--- a/cmake/external/CudaRdcUtils.cmake
+++ b/cmake/external/CudaRdcUtils.cmake
@@ -812,7 +812,7 @@ function(cuda_rdc_use_middle_lib_in_property target property)
   set(_new_values)
   foreach(_lib IN LISTS _target_libs)
     set(_newlib ${_lib})
-    # Simplistic treatement of `$<LINK_ONLY:...>`
+    # Simplistic treatment of `$<LINK_ONLY:...>`
     string(REGEX REPLACE "\\\$<LINK_ONLY:(.*)>" "\\1" _stripped_lib ${_lib})
     if(_stripped_lib STREQUAL _lib)
       set(_stripped_lib)
@@ -921,7 +921,7 @@ function(cuda_rdc_check_cuda_runtime OUTVAR library)
     #   You have used file(GET_RUNTIME_DEPENDENCIES) in project mode.  This is
     #     probably not what you intended to do.
     # On the other hand, if the library is using (relocatable) CUDA code and
-    # the shared run-time library and we don't have the scafolding libraries
+    # the shared run-time library and we don't have the scaffolding libraries
     # (shared/static/final) then this won't work well. i.e. if we were to detect this
     # case we probably need to 'error out'.
     get_target_property(_cuda_middle_library ${library} CUDA_RDC_MIDDLE_LIBRARY)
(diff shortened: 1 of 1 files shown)


cmake/external/CudaRdcUtils.cmake
```diff
@@ -844,7 +921,7 @@ function(cuda_rdc_check_cuda_runtime OUTVAR library)
     #   You have used file(GET_RUNTIME_DEPENDENCIES) in project mode.  This is
     #     probably not what you intended to do.
     # On the other hand, if the library is using (relocatable) CUDA code and
-    # the shared run-time library and we don't have the scaffolding libraries
+    # the shared run-time library and we don't have the scafolding libraries
```
sethrj: This plus line 815 should be reverted 🙃 I guess you didn't backport some of our pre-commit's comments to CudaRdc 😉

cmake/external/CudaRdcUtils.cmake
```diff
@@ -460,6 +528,15 @@ function(cuda_rdc_add_library target)
     EXPORT_PROPERTIES "CUDA_RUNTIME_LIBRARY;CUDA_RDC_LIBRARY_TYPE;CUDA_RDC_FINAL_LIBRARY;CUDA_RDC_MIDDLE_LIBRARY;CUDA_RDC_STATIC_LIBRARY"
   )
 
+  if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
sethrj: Are you sure this only applies to linux? Wouldn't it also technically depend on the linker (ld versus gold vs ...)?

sethrj (approved):

cmake/external/CudaRdcUtils.cmake
pcanal: It is indeed more about the linker ...

cmake/external/CudaRdcUtils.cmake
pcanal: Fixed.

cmake/external/CudaRdcUtils.cmake
pcanal: Removed.

sethrj:
Thanks @pcanal ! Hopefully the "Linux" check is redundant anyway, since macOS doesn't have CUDA, and God help us all if we ever try to compile with VecGeom on windows...

PR #2115, opened 2025-11-17
Use type aliases for vecgeom and simplify detail headers

sethrj:
This is a next step toward simplifying and refactoring the various vecgeom navigators.
- Define a type helper for switching back and forth between cxx/cuda namespace based on the Memspace value
- Combine some of the detail headers (traits, version) into a VecgeomTypes
- Use more consistent aliases for  vecgeom types: `VgLogVol`, `VgPlacedVol`, `VgNavState`, ...
- Define types for integer/reals of concern: `vg_real_type  -> vecgeom::Precision`, `VgSurfaceInt -> long`


src/geocel/vg/detail/BVHNavigator.hh
```diff
@@ -102,7 +99,6 @@ class BVHNavigator
             path.Pop();
             return LocatePointIn(currentmother, transformed, path, false);
```
mrguilima, line 106:
```suggestion
            LocatePointIn(currentmother, transformed, path, false);
```

src/geocel/vg/detail/BVHNavigator.hh
```diff
@@ -33,30 +33,29 @@ namespace detail
...
+    static constexpr vg_real_type kBoundaryPush = 10 * vecgeom::kTolerance;
 
-    CELER_FUNCTION static VPlacedVolumePtr_t
-    LocatePointIn(vecgeom::VPlacedVolume const* vol,
-                  Vector3D const& point,
-                  vecgeom::NavigationState& path,
+    CELER_FUNCTION static void
+    LocatePointIn(VgPlacedVol const* vol,
+                  VgReal3 const& point,
+                  Nav& path,
```
mrguilima: Document path as output / side effect?

src/geocel/vg/detail/BVHNavigator.hh
```diff
@@ -33,30 +33,29 @@ namespace detail
...
-                  vecgeom::VPlacedVolume const* exclude = nullptr)
+                  VgPlacedVol const* exclude = nullptr)
     {
         if (top)
         {
-            assert(vol != nullptr);
+            CELER_ASSERT(vol != nullptr);
             if (!vol->UnplacedContains(point))
-                return nullptr;
+                return;
```
mrguilima, line 57: Clear (output) path before returning.

mrguilima (approved):
Looks great!

src/geocel/vg/detail/BVHNavigator.hh
mrguilima: A minor comment: the Nav alias feels a bit vague. How about NavState?

sethrj: These have always been side effects, and I've updated the name to NavState.

src/geocel/vg/detail/BVHNavigator.hh
sethrj: I didn't change the behavior on this PR; I think this only used by the `VecgeomTrackView::operator=` case.

src/geocel/vg/detail/BVHNavigator.hh
sethrj: It also needs to return; this just ensures the return value of the functions are consistent (i.e., one can't be void if the other isn't).

src/geocel/vg/detail/BVHNavigator.hh
mrguilima: I know, but in the spirit of improving the code, since this function has no control over the nav state argument, which could be anything as provided by the callee, it would be probably sensible to make sure it is reset to null_pointer in this case, when one knows the point is outside of "top" volume.

Issue #2111, opened 2025-11-15
Optimize vecgeom usage


sethrj:
The new vecgeom 2 solid navigator in #1720 is 3–5⨉ slower than v1.0 because it does a full relocation at each distance calculation, and those relocations are discarded most of the time during tracking and especially during field propagation. There are still inconsistencies with bounding boxes and surface crossings. Additionally the Celeritas field propagation approach is different than that in VecGeom/Geant4 since we converge with direct distance calculations rather than use the safety distance as we approach a boundary.

The vecgeom state is also odd, because the "NavState" objects have two internal states, one of which is pretty much unused (and essentially inaccessible in vecgeom 1) in addition to a boundary state. Ideally we would only store and use one state.

PR #2116, opened 2025-11-18
Remove hidden state from VecGeom

### LINKED ISSUE #1921: Support surface normals in VecGeom

sethrj:
As part of #2111 I wanted to eliminate the it-could-almost-be-useful `fLastExited` path that lives inside the VecGeom 2 NavState but is inaccessible from VG1 and not present in the NavPath implementation. We need *something* like it to fix #1921 because the normal may live on the parent or the child (before *or* after). So to avoid keeping multiple exiting/entering states, since we already have "next" and "current", I added some wrappers that look like a navstate but operate on references to the state/boundary values. (Storage is also easier and more efficient with these since boundary bools live side by side.)

Lo and behold, several of the vecgeom tests no longer needed the exception "is solid && vecgeom 2" added in #1720. The *only* difference is that we discard `fLastExited` after every VecGeom navigation call. This indicates the new logic for excluding volumes previously entered/exited is interfering with how we do tracking.

The next step after this will be merging the solid BVH navigators to avoid the "relocate on find distance".

---

## Implementation notes

- `CELER_VGNAV` is defined based on vecgeom configuration to `CELER_VGNAV_PATH`, `CELER_VGNAV_TUPLE`, or `CELER_VGNAV_INDEX`. The "path" selection is unfortunately deployed on all vg1 CPU and has special treatment for its state allocation.
- Vecgeom surface navigation does not need the 'next boundary' flag since it already has 'next surface'. We use that instead, but for the moment to keep the interface consistent still allocate a 'next boundary'.
- `VgNavStateWrapper` has the interface of a VG nav state, but it lacks the "last exited" state, and it simply forwards calls to the stored references to state/boundary. This lets us use it to do things like Push/Pop
- For compatibility with other vecgeom functions that require the whole big state, there's a `ScopedVgNavState` that allows the full states to be saved on the stack and reconstruct the pointed-to data on destruction (discarding fLastExited in the process).
- Use Collection instead of `VecgeomNavCollection` for all types except for the navigation path.
- A new VgBoundary (off/on) stores whether the track is on the boundary; we might want to move this to `geocel/Types` and use it across geometry engines
- Use constexpr instead of macros where possible, including by holding a null pointer to the next surface int (used only for surface geometry)


mrguilima:
Regarding the NavIndexPath option for CPU, since it is not compatible with GPU, it gets automatically "upgraded" out of the *Path option when GPU is enabled.

src/geocel/CMakeLists.txt
```diff
@@ -75,10 +75,14 @@ if(CELERITAS_USE_VecGeom)
   list(APPEND SOURCES
     vg/VecgeomParams.cc
     vg/VecgeomParamsOutput.cc
-    vg/detail/VecgeomNavCollection.cc
     vg/RaytraceImager.cc
     vg/SafetyImager.cc
   )
+  if(VecGeom_NAV STREQUAL "path")
```
mrguilima: I don't see a point in preserving the ability to run with VecGeom_NAV in "path" mode, as the performance is probably better in the NavIndex mode.

src/geocel/vg/detail/VecgeomNavCollection.hh
```diff
@@ -78,104 +88,29 @@ struct VecgeomNavCollection<Ownership::reference, MemSpace::host>
...
-        return pool_view.IsValid();
+        CELER_EXPECT(*this);
+        CELER_EXPECT(tid < nav_state.size());
+        return *nav_state[tid.unchecked_get()];
     }
+
+    //! Number of elements
+    size_type size() const { return nav_state.size(); }
+
+    //! True if the collection is assigned/valiid
```
mrguilima:
```suggestion
    //! True if the collection is assigned/valid
```

sethrj:
@mrguilima This is ready for review.

mrguilima (approved):
Looks great!
Thanks Seth!

sethrj:
Ugh, bad news: as of this commit, vecgeom 2 fails 

with config:
```json
{"_category":"system","_label":"build","config":{"build_type":"relwithdebinfo,shared","core_geo":"VecGeom","core_rng":"xorwow","debug":true,"geant4":"global-dynamic,multithreaded,shared,gdml","gpu_architectures":"90","hostname":"hudson","openmp":"disabled","real_type":"double","units":"CGS","use":["covfie","cuda","geant4","hepmc3","root","vecgeom"],"vecgeom":"bvh_single,tuple,Scalar","versions":{"CLHEP":"2.4.7.1","CUDA":"12.9.86","G4VG":"1.0.5","Geant4":"11.3.2","HepMC3":"3.03.01","ROOT":"6.36.04","Thrust":"","VecGeom":"2.0.0-dev.7","covfie":"0.15.3"}},"version":"0.7.0-dev.204+71736034a"}
```
before:
```
The following tests FAILED:
	232 - celeritas/geo/Geometry:SimpleCms* (Failed)        gpu unit
	234 - celeritas/geo/Geometry:ThreeSpheres* (Failed)     gpu unit
```
after (with a patch to fix missing vecgeom IWYU):
```
The following tests FAILED:
	 87 - geocel/vg/Vecgeom:SimpleCmsVgdmlTest.* (Subprocess aborted) gpu unit
	232 - celeritas/geo/Geometry:SimpleCms* (Failed)        gpu unit
	234 - celeritas/geo/Geometry:ThreeSpheres* (Failed)     gpu unit
	317 - celeritas/track/TrackSort (Failed)                gpu unit
	323 - celeritas/user/SlotDiagnostic (Failed)            gpu unit
	354 - accel/TrackingManagerIntegration:LarSphere.run:gpu:mt (Failed) gpu unit
	355 - accel/TrackingManagerIntegration:LarSphere.run:gpu:task (Failed) gpu unit
```

Looks like this is because of my hacky (incorrect+illegal) use of NavStateTuple as POD.

PR #2118, opened 2025-11-18
Fix VecGeom 2 solids behavior by clearing last exited state

sethrj:
This is the core change of #2116 to fix discrepancies between VecGeom 1 and 2 for the solids model. I think the way we query and move across boundaries is different enough to AdePT's that the recent VecGeom changes broke our existing method by preventing "reentry" into a volume recently queried.

@agheata I think it could improve the consistency and usage of the navigation "state" to split out the two embedded states and the boundary state, and *just* have a single nav tuple/index represent the state. (This is what I've partly done in #2116.) There's also an increasing amount of unused code in VecGeom's master, perhaps a description of what's intended to be used and what's to be discarded would help? We can discuss at the next vecgeom meeting if needed.

### DIFF
--- a/src/geocel/vg/detail/SolidsNavigator.hh
+++ b/src/geocel/vg/detail/SolidsNavigator.hh
@@ -77,6 +77,7 @@ class SolidsNavigator
         real_type step = navigator->ComputeStepAndPropagatedState(
             glpos, gldir, step_limit, curr, next);
         curr.SetLastExited({});
+        next.SetLastExited({});
 
         return step;
     }
--- a/src/geocel/vg/detail/SolidsNavigator.hh
+++ b/src/geocel/vg/detail/SolidsNavigator.hh
@@ -76,6 +76,7 @@ class SolidsNavigator
         auto* navigator = curr_volume->GetNavigator();
         real_type step = navigator->ComputeStepAndPropagatedState(
             glpos, gldir, step_limit, curr, next);
+        curr.SetLastExited({});
 
         return step;
     }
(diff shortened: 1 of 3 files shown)


agheata:
> @agheata I think it could improve the consistency and usage of the navigation "state" to split out the two embedded states and the boundary state, and _just_ have a single nav tuple/index represent the state. (This is what I've partly done in #2116.) There's also an increasing amount of unused code in VecGeom's master, perhaps a description of what's intended to be used and what's to be discarded would help? We can discuss at the next vecgeom meeting if needed.

Indeed, the use of fLastExited - although useful for discarding exiting paths - is quite brittle when it comes to dealing with back scattering. We also had to override the value in AdePT https://github.com/apt-sim/AdePT/blob/3f9b7759ae44c89a66309f24ea33bf6a633c6d8a/include/AdePT/kernels/electrons.cuh#L445-L451 after relocation. I would be in favor of splitting to a pure state and passing last_exited and boundary state flag separately, giving the user the choice to use or not to use. The current form was mainly dictated by compatibility with the previous NavSatePath. I also agree that some interfaces are not used and need to be removed, while the existing ones need to be much better documented. I guess this will be part of the cleanup for next year. I discussed with Severin, and we will put together in the coming weeks a detailed plan of action that we are going to share. We should consider which may be some priority pre-V2 items. We are not really bound to stamp v2 now; we can still go on with rc/dev[N] for a while, if we agree on what features must be in the V2 release.

mrguilima (approved):
Awesome!  Thanks @sethrj

PR #2120, opened 2025-11-18
Add helper classes for managing aux data

sethrj:
For #1746 .


Rashika-Gupta (approved):

src/corecel/data/AuxParams.hh
```diff
@@ -0,0 +1,122 @@
...
+    // Access the *state* ref (const)
+    template<MemSpace M>
+    inline StateRefT<M> const& ref(AuxStateVec const& v) const;
+
+    // Access the *state* ref (mutable)
+    template<MemSpace M>
+    inline StateRefT<M>& ref(AuxStateVec& v) const;
+
+    //! Allow access host/device *params* ref
+    using ParamsDataInterface<P>::ref;
```
sethrj, line 72: Is this too much overloading? Should I rename the first two functions `state_ref`?

sethrj:
@amandalund Would you be able to give me any feedback on this and hit merge if you think it's ok?

amandalund:
Will do!

src/celeritas/track/StatusChecker.cc
```diff
@@ -117,20 +107,20 @@ void StatusChecker::step(ActionId prev_action,
                          CoreState<M>& state) const
 {
     CELER_EXPECT(data_);
+    CELER_EXPECT(this->data_);
```
amandalund:
```suggestion
```

src/corecel/data/AuxParams.hh
amandalund: The signature makes it clear those methods access the state, but I agree naming them just `ref` inside a class named `Params` could be ambiguous.

amandalund (approved):
Looks good!

sethrj:
Thanks @amandalund , I've also removed the `Data` part of the class names from the params/state template classes, since we should be more consistent about using Data exclusively for collection groups.

PR #2121, opened 2025-11-19
Switch to infix tracking to accelerate ORANGE point-in-volume (#2121)

### LINKED ISSUE #1282: Implement short-circuit volume evaluation using infix notation

esseivaju:
Draft MR to track failing tests when infix tracking is on


esseivaju:
I think the remaining failing tests are sort of expected from hard-coded values assuming postfix logic.

@codex review

src/orange/univ/detail/InfixEvaluator.hh
```diff
@@ -109,8 +109,12 @@ CELER_FUNCTION bool InfixEvaluator::operator()(F&& eval_sense) const
...
             // negation of a sub-expression is not supported
             CELER_ASSUME(i + 1 < logic_.size());
-            CELER_EXPECT(!logic::is_operator_token(logic_[i + 1]));
-            result = !static_cast<bool>(eval_sense(FaceId{logic_[++i]}));
+            CELER_EXPECT(!logic::is_operator_token(logic_[i + 1])
+                         || logic_[i + 1] == logic::ltrue);
+            auto const& operand = logic_[++i];
+            result = operand == logic::ltrue
+                         ? false
+                         : !static_cast<bool>(eval_sense(FaceId{operand}));
```
sethrj: I think this is a bit clearer and simpler since the result is negating the next operand:
```suggestion
            result = !((operand == logic::ltrue) || static_cast<bool>(eval_sense(FaceId{operand})));
```

Would it be better to include a literal`lfalse` and use that?

Also (I thought we already had this) but maybe a `to_bool(Sense s)`...

sethrj (approved):
Excellent!

sethrj:
I think the last step is, in the `OrangeParams` constructor, compare the `input.logic` against `orange_tracking_logic` and, if different, apply the postfix-to-infix conversion to all volumes?

elliottbiondo (approved):

esseivaju:
> I think the last step is, in the `OrangeParams` constructor, compare the `input.logic` against `orange_tracking_logic` and, if different, apply the postfix-to-infix conversion to all volumes?

I think that should already be there? We're calling [convert_logic](https://github.com/esseivaju/celeritas/blob/d2f04f34a3e531fc5e4dc35aaa2b6cf75dc150f7/src/orange/OrangeParams.cc#L200) in the constructor, and [loop over all the volumes](https://github.com/esseivaju/celeritas/blob/d2f04f34a3e531fc5e4dc35aaa2b6cf75dc150f7/src/orange/detail/LogicUtils.cc#L492-L507)

sethrj:
Oh nice! sorry I forgot we already did that :)

sethrj:
@esseivaju To reduce churn in the unit tests, we could have the input builder use `LogicNotation` as a construction option, remove the `remove_negated_join` option from UnitProto, and replace https://github.com/celeritas-project/celeritas/blob/605aec4f4022d16dc369106f6bbd561a1f6b7ef7/src/orange/orangeinp/UnitProto.cc#L600-L601 with code in `build(ProtoBuilder& pb)` that performs the simplification on `csg_unit` if the ProtoBuilder requests infix notation.

esseivaju:
> @esseivaju To reduce churn in the unit tests, we could have the input builder use `LogicNotation` as a construction option, remove the `remove_negated_join` option from UnitProto, and replace
> 
> https://github.com/celeritas-project/celeritas/blob/605aec4f4022d16dc369106f6bbd561a1f6b7ef7/src/orange/orangeinp/UnitProto.cc#L600-L601
> 
> with code in `build(ProtoBuilder& pb)` that performs the simplification on `csg_unit` if the ProtoBuilder requests infix notation.

Wouldn't that cause some consistency issue? If `orange_tracking_logic()` is set to infix, we have to perform the simplification regardless of the input configuration, or issue an error.

sethrj:
> Wouldn't that cause some consistency issue? If `orange_tracking_logic()` is set to infix, we have to perform the simplification regardless of the input configuration, or issue an error.

If `orange_tracking_logic` is set to infix, and the input builder is set to generate postfix for testing, the OrangeParams code will convert the postfix to infix. Right?

sethrj:
Thanks for bringing this home @esseivaju ! 🎉

esseivaju:
> > Wouldn't that cause some consistency issue? If `orange_tracking_logic()` is set to infix, we have to perform the simplification regardless of the input configuration, or issue an error.
> 
> If `orange_tracking_logic` is set to infix, and the input builder is set to generate postfix for testing, the OrangeParams code will convert the postfix to infix. Right?

Right, the params make sure that the logic notation matches the logic evaluation used in geometry navigation.

PR #2122, opened 2025-11-20
Make optical primary generator distributions configurable

amandalund:
Previously our optical primary generator was hardcoded to sample a point source of mono-energetic isotropic photons. This updates the generator and data to support runtime-configurable distribution types from user input. I've also added an input option for a Gaussian energy distribution.


src/celeritas/inp/Events.hh
```diff
@@ -65,45 +65,46 @@ struct MonoenergeticDistribution
     MevEnergy energy;
 };
 
+//! Generate primaries with Gaussian-distributed energy
+struct GaussianDistribution
+{
+    real_type mean;
+    real_type stddev;
```
sethrj: Add default initializers {0},{1} to match the default GaussianDistribution.

src/celeritas/optical/gen/PrimaryGenerator.hh
```diff
@@ -39,7 +37,7 @@ class PrimaryGenerator
     //// DATA ////
 
     PrimaryDistributionData const& data_;
-    IsotropicDistribution<real_type> sample_angle_;
+    IsotropicDistribution<real_type> sample_polarization_;
```
sethrj: We should add a note to our separable user distribution that the sampled light is unpolarized.

src/celeritas/optical/gen/PrimaryGeneratorAction.hh
```diff
@@ -33,6 +33,8 @@ class CoreStateBase;
  *
  * This reproducibly samples and initializes optical photons directly in track
  * slots.
+ *
+ * \todo Integrate this action into a standalone stepping loop.
```
sethrj: Another \todo: use our compile-time polymorphism methods to allow sampling on device.

test/celeritas/optical/Generator.test.cc
```diff
@@ -136,17 +136,17 @@ TEST_F(LArSphereGeneratorTest, primary_generator)
 {
     // Create primary generator action
     inp::OpticalPrimaryGenerator inp;
-    inp.num_events = 1;
-    inp.primaries_per_event = 65536;
-    inp.energy.energy = units::MevEnergy{1e-5};
+    inp.primaries = 65536;
```
sethrj, line 139: I guess in source code we could also write this as `ipow<16>(2)` 😉

sethrj (approved):
Looks great!

src/celeritas/optical/gen/PrimaryGenerator.hh
```diff
@@ -68,16 +66,17 @@ CELER_FUNCTION optical::TrackInitializer
...
-    result.position = data_.position;
-    result.direction = sample_angle_(rng);
+    result.energy = data_.sample_energy(rng);
+    result.position = data_.sample_position(rng);
+    result.direction = data_.sample_direction(rng);
     do
     {
         result.polarization = make_unit_vector(
-            make_orthogonal(sample_angle_(rng), result.direction));
+            make_orthogonal(sample_polarization_(rng), result.direction));
```
hhollenb: I believe the polarization can be sampled directly from
```
result.polarization = ExitingDirectionSampler{0, result.direction}(rng);
```

hhollenb (approved):
Looks great! Will definitely be a help for upcoming validation.

test/celeritas/optical/Generator.test.cc
amandalund: You want to write 0x10000, don't you

src/celeritas/optical/gen/PrimaryGenerator.hh
amandalund: Nice!

test/celeritas/optical/Generator.test.cc
sethrj: Or `1 << 16` to be even more confusing 😉

src/celeritas/optical/gen/PrimaryGeneratorData.hh
```diff
@@ -0,0 +1,224 @@
...
+struct EnergySampler
+{
+    union DistributionData
+    {
+        inp::MonoenergeticDistribution monoenergetic;
+        inp::GaussianDistribution gaussian;
+
+        //! Default constructor (uninitialized state)
+        CELER_FORCEINLINE_FUNCTION DistributionData() {}
+    };
```
sethrj: Sorry I didn't even look at this file! (I think github hid it for being too large.) I had forgotten you made the original generator on device (unlike the CPU version which is host only). I think you may indeed have to take the approach of #2123 ... I don't think it's a heavy lift though? We can discuss at today's meeting

src/corecel/random/distribution/DistributionTypeTraits.hh
```diff
@@ -0,0 +1,144 @@
...
+CELER_DISTRIB_TRAITS(isotropic, IsotropicDistribution<real_type>);
+CELER_DISTRIB_TRAITS(uniform_box, UniformBoxDistribution<real_type>);
+
+#undef CELER_DISTRIB_TRAITS
+
+//---------------------------------------------------------------------------//
+/*!
+ * Expand a macro to a switch statement over all possible distribution types.
+ */
+struct DistributionVisitor
```
sethrj: I'd put this into a separate file, separating the types/data from the runtime/execution.

src/corecel/Types.hh
```diff
@@ -79,6 +79,25 @@ enum class UnitSystem
...
+//---------------------------------------------------------------------------//
+//! Distributions returning a length three array of real types
+enum class ThreedDistributionType
+{
+    delta,
+    isotropic,
+    uniform_box,
+    size_
+};
+
```
sethrj: I wonder when we should come up with some rules about the granularity of `Types.hh` files: I know I shouldn't be worried about it, but it kind of bothers me that adding a new 1D distribution will cause the whole codebase to recompile.

src/corecel/random/distribution/DistributionTypeTraits.hh
```diff
@@ -0,0 +1,144 @@
...
+    template<>                                                          \
+    struct OnedDistributionTypeTraits<OnedDistributionType::ENUM_VALUE> \
+        : public EnumToClass<OnedDistributionType,                      \
+                             OnedDistributionType::ENUM_VALUE,          \
+                             CLS>                                       \
+    {                                                                   \
+    }
+
+CELER_DISTRIB_TRAITS(delta, DeltaDistribution<real_type>);
+CELER_DISTRIB_TRAITS(normal, NormalDistribution<real_type>);
```
sethrj, line 38:
```suggestion
#define CELER_DISTRIB_TRAITS(ENUM_VALUE, CLS, RECORD)                   \
    template<>                                                          \
    struct OnedDistributionTypeTraits<OnedDistributionType::ENUM_VALUE> \
        : public EnumToClass<OnedDistributionType,                      \
                             OnedDistributionType::ENUM_VALUE,          \
                             CLS>                                       \
    {                                                                   \
        using record_type = CLS##Record;                                \
    }

CELER_DISTRIB_TRAITS(delta, DeltaDistribution<real_type>, DeltaDistributionRecord<real_type>);
CELER_DISTRIB_TRAITS(normal, NormalDistribution<real_type>, NormalDistributionRecord);
```

Adding the types here will remove the need for the "detail::Builder" helper in the visitor.

src/corecel/random/data/DistributionData.hh
```diff
@@ -0,0 +1,131 @@
...
+ */
+template<Ownership W, MemSpace M>
+struct DistributionParamsData
+{
+    //// TYPES ////
+
+    template<class T>
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using OnedDistributionItems = Collection<T, W, M, OnedDistributionId>;
```
sethrj, line 68: I don't think we should be using `OnedDistributionId` for this opaque ID type, since it implies that the ID to access a `delta_real` is that ID, but it's really a separate (usually smaller) value.

src/corecel/random/distribution/DistributionTypeTraits.hh
```diff
@@ -0,0 +1,144 @@
...
+            build(params.FIELD[ItemId<RECORD>(idx)]))
+    switch (type)
+    {
+        CELER_DISTRIB_CASE(
+            delta, delta_real, DeltaDistributionRecord<real_type>);
+        CELER_DISTRIB_CASE(normal, normal, NormalDistributionRecord);
+        default:
+            CELER_ASSERT_UNREACHABLE();
+    }
+#undef CELER_DISTRIB_CASE
```
sethrj: I'm not sure whether this ends up being more or less readable, but you could do:

```suggestion
    using DistTypeEnum = OnedDistributionType;

#define CELER_DISTRIB_CASE(ENUM_VALUE, FIELD)                                    \
    case DistTypeEnum::ENUM_VALUE: {                                             \
        return this->impl<OnedDistributionTypeTraits<DistTypeEnum::ENUM_VALUE>>( \
            celeritas::forward<F>(func), params.FIELD, idx);

    switch (type)
    {
        CELER_DISTRIB_CASE(delta, delta_real);
        CELER_DISTRIB_CASE(normal, normal);
        default:
            CELER_ASSERT_UNREACHABLE();
    }
#undef CELER_DISTRIB_CASE
```

with a helper function that can be used for both the 1D and 3D distributions:
```c++
    template<class TraitsT, class F>
    decltype(auto) impl(F&& func,
                        Collection<TraitsT::record_type,
                                   Ownership::const_reference,
                                   MemSpace::native> const& records,
                        size_type idx)
    {
        using IdT = ItemId<typename TraitsT::record_type>;
        using TypeT = typename TraitsT::type;
        return celeritas::forward<F>(func)(TypeT{records[id_cast<IdT>(idx)]});
    }
```

sethrj (commented):
Double checking the type trait stuff but here's some initial comments!

sethrj:
Could you also move https://github.com/celeritas-project/celeritas/pull/2122/commits/953ab1ee9f9500b7c4ccbb8385d8a8b4f5e90079 to a little separate PR? It's a significant enough change.

src/corecel/Types.hh
amandalund: It was rather annoying to rebuild everything when I tweaked those enums, and I don't really think they belong at this level. We do have Types.hh in the optical and field subdirectories: should we add one to corecel/random?

sethrj: Yeah let's do that.

src/corecel/random/distribution/DistributionTypeTraits.hh
amandalund: Yeah it is more readable without the class name in the macro.

src/corecel/random/data/DistributionData.hh
amandalund: Not sure I follow... we shouldn't be accessing the types/indices with `OnedDistributionId`?

sethrj: Oh I'm sorry, I misread what this was used for. It's correct as is: types and indices should indeed be OnedDistributionItems; but delta_real and normal should not. (This is what you've got right now.)

src/celeritas/optical/gen/PrimaryGenerator.hh
```diff
@@ -67,17 +71,20 @@ template<class Generator>
...
+    DistributionVisitor visit{params_};
+
     optical::TrackInitializer result;
-    result.energy = data_.energy;
-    result.position = data_.position;
-    result.direction = sample_angle_(rng);
+    result.energy = units::MevEnergy{
+        visit([&rng](auto&& d) { return d(rng); }, data_.energy)};
+    result.position = visit([&rng](auto&& d) { return d(rng); }, data_.shape);
+    result.direction = visit([&rng](auto&& d) { return d(rng); }, data_.angle);
```
sethrj: If we do this in more places, it might be worth making a helper:
```c++
template<class Visitor, class Dist, class Engine>
CELER_FUNCTION decltype(auto) sample_with(Visitor&& visit, Dist&& dist, Engine& rng)
{
    return celeritas::forward<Visitor>(visit)(
        [&rng](auto&& d) -> decltype(auto) { return d(rng); },
        celeritas::forward<Dist>(dist));
}
```
then this becomes

```suggestion
result.energy = units::MevEnergy{sample_with(visit, data_.energy, rng)};
result.position  = sample_with(visit, data_.shape, rng);
result.direction = sample_with(visit, data_.angle, rng);
```

and we could use the same for the 3D distributions etc.

To support `std::visit` we could define (in `corecel/cont/VariantUtils.hh`):
```c++
struct StdVisitor
{
    template<class F, class... Vs>
    decltype(auto) operator()(F&& f, Vs&&... vs) const
    {
        return std::visit(std::forward<F>(f),
                          std::forward<Vs>(vs)...);
    }
};
```

src/corecel/math/ArrayUtils.hh
```diff
@@ -85,6 +85,12 @@ template<class T>
 [[nodiscard]] inline CELER_FUNCTION Array<T, 3>
 rotate(Array<T, 3> const& dir, Array<T, 3> const& rot);
 
+//---------------------------------------------------------------------------//
+// Convert an array from type \c T2 to \c T1
+template<class T1, class T2, size_type N>
+[[nodiscard]] inline CELER_FUNCTION Array<T1, N>
+array_cast(Array<T2, N> const& x);
```
sethrj, line 92: Nice!

src/corecel/random/distribution/DistributionVisitor.hh
```diff
@@ -0,0 +1,121 @@
...
+    CELER_EXPECT(id < params_.oned_types.size());
+
+    ODT type = params_.oned_types[id];
+    size_type idx = params_.oned_indices[id];
+
+#define CELER_DISTRIB_CASE(ENUM, FIELD)                                     \
+    case ODT::ENUM: {                                                       \
+        using TypeT = typename OnedDistributionTypeTraits<ODT::ENUM>::type; \
+        return celeritas::forward<F>(func)(                                 \
+            TypeT{params_.FIELD[ItemId<TypeT::RecordT>(idx)]});             \
```
sethrj, line 78: I think you might need a `typename` here as well

src/corecel/random/Types.hh
```diff
@@ -0,0 +1,57 @@
...
+    uniform_box,
+    size_
+};
+
+//---------------------------------------------------------------------------//
+// TYPE ALIASES
+//---------------------------------------------------------------------------//
+
+//! Three-dimensional cartesian coordinates
+using Real3 = Array<real_type, 3>;
```
sethrj: Will this conflict with the types elsewhere? Perhaps we should just define this in `corecel/Types.hh` by forward-declaring `Array`, since we use this type all over the place.

src/celeritas/inp/Events.hh
```diff
@@ -20,18 +21,16 @@ namespace celeritas
 namespace inp
 {
 //---------------------------------------------------------------------------//
-//! Generate at a single point
-struct PointDistribution
-{
-    Real3 pos{0, 0, 0};  // [length]
-};
+//! Generate at a single energy value
```
sethrj: Since we've lost the Quantity annotation, make sure to document the units as `MeV`. In the future we can add our "runtime" units annotation.

src/corecel/inp/Distributions.hh
```diff
@@ -0,0 +1,50 @@
...
+//---------------------------------------------------------------------------//
+//! "Sample" the given value
+template<class T>
+struct DeltaDistribution
+{
+    T value{};
+};
+
+//---------------------------------------------------------------------------//
+//! Generate primaries with Gaussian-distributed energy
```
sethrj:
```suggestion
//! Sample from a Gaussian (normal) distribution
```
since we could use this for wavelength, etc.

src/corecel/inp/Distributions.hh
```diff
@@ -0,0 +1,50 @@
...
+//! Sample a point uniformly on the surface of a unit sphere
+struct IsotropicDistribution
+{
+};
+
+//---------------------------------------------------------------------------//
+//! Sample uniformly in a box
+struct UniformBoxDistribution
+{
+    using Real3 = Array<double, 3>;
```
sethrj, line 42: This should no longer be necessary with the top-level type alias right?

test/accel/UserActionIntegration.test.cc
```diff
@@ -135,12 +136,12 @@ class LarSphereOpticalOffload : public LarSphere
...
 {
-    using MevEnergy = Quantity<units::Mev, double>;
     auto result = LarSphereIntegrationMixin::make_primary_input();
 
-    result.shape = inp::PointDistribution{from_cm({0.1, 0.1, 0})};
+    result.shape
+        = inp::PointDistribution{array_cast<double>(from_cm({0.1, 0.1, 0}))};
     result.primaries_per_event = 1;
-    result.energy = inp::MonoenergeticDistribution{MevEnergy{1}};
+    result.energy = inp::MonoenergeticDistribution{1};
```
sethrj: To help document we could do:
```suggestion
    result.energy = inp::MonoenergeticDistribution{MevEnergy{1}.value()};
```
or (to avoid the unused type alias)
```suggestion
    result.energy = inp::MonoenergeticDistribution{1}; // [MeV]
```

sethrj (approved):
Excellent work!

src/corecel/inp/Distributions.hh
amandalund: I kept this one as `double` since it's input.

src/corecel/random/distribution/DistributionVisitor.hh
amandalund: I need it for the template argument?

PR #2127, opened 2025-11-24
Remove temporary sense storage

esseivaju:
We've already been using on-the-fly sense evaluation for a while. The temporary sense storage was still being allocated in the state data the even though it's unused. Remove the storage, and  sense calculators depending on it.


src/orange/SenseUtils.hh
```diff
@@ -112,62 +112,4 @@ inline char const* to_cstring(SignedSense s)
...
-    CELER_CONSTEXPR_FUNCTION bool is_assigned() const
-    {
-        return sense_[is_assigned_bit];
-    }
-
-    //! Clear the sense value
-    CELER_CONSTEXPR_FUNCTION void reset() { sense_.reset(); }
-
-  private:
-    Bitset<2> sense_;
```
sethrj, line 170: is Bitset used anywhere now? We spent far too much time on that thing 😪

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -562,7 +562,7 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
 
     // NOTE: if switching to the "eager" SenseCalculator, this must be moved
     // inside the loop, since it recalculates senses only on construction.
-    detail::LazySenseCalculator calc_sense{
+    detail::SenseCalculator calc_sense{
```
sethrj: I think the note can be deleted above. Also I think changing the class name confuses the PR since the diff makes it look like the SenseCalculator changed implementation 🙁 Generally we should try to have pure renaming done separately from changes.

src/orange/univ/detail/SenseCalculator.hh
```diff
@@ -28,22 +28,23 @@ namespace detail
...
  * volume's face with respect to the given position. This class is used to
- * lazily calculate sense during evaluation of a logic expression, contrary to
- * CachedLazySenseCalculator, this class does not cache the calculated sense:
- * potentially recomputing the same sense value multiple time. The advantage
- * is that we do not need to access global memory to store the cached sense.
+ * lazily calculate sense during evaluation of a logic expression.
+ * This class does not cache the calculated sense, that is senses are
+ * recomputed every time it is requested. The advantage of not caching senses
+ * is that we eliminate global memory usage. With infix evaluation, and
+ * short-circuiting logic, the cost of recomputing senses should be small.
```
sethrj: With the latest changes this should be "on-the-fly" rather than "lazy", which usually implies caching.

sethrj (approved):
Thanks @esseivaju !

src/orange/SenseUtils.hh
esseivaju: Nop, no use as of now 😅

src/orange/univ/SimpleUnitTracker.hh
esseivaju: Yeah it really messed up the diff... I've reverted that commit

PR #2129, opened 2025-11-25
Fix spurious warning when loading scintillation properties and update device tests

amandalund:
- Fixes the warning emitted even when the correct prefix was used:
```console
src/celeritas/ext/GeantImporter.cc:280: warning: Deprecated property prefix SCINTILLATIONLAMBDA: use CELER_SCINTILLATION
src/celeritas/ext/GeantImporter.cc:280: warning: Deprecated property prefix SCINTILLATIONLAMBDA: use CELER_SCINTILLATION
```
- Validates that either both or neither of the mean and sigma are present
- Updates the device tests for #2100 (which we always forget to do :))

### DIFF
--- a/src/celeritas/ext/GeantImporter.cc
+++ b/src/celeritas/ext/GeantImporter.cc
@@ -202,22 +202,17 @@ load_gauss_scint(std::string const& prefix,
                  detail::GeantMaterialPropertyGetter& get_property,
                  int comp_idx)
 {
-    ImportGaussianScintComponent gauss;
-    bool found_mean = false;
+    ImportGaussianScintComponent gauss{};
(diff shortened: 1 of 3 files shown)


sethrj:
I think @Rashika-Gupta has an account on ExCL now, so she should be able to ssh into `hudson` and run tests on future PRs 😄 

Also as we fill out our optical capabilities we might consider deleting the "count the number of total steps" tests since it might be a bit too sensitive?

sethrj (approved):
Sorry for not caching this in review!

Rashika-Gupta (commented):
Thank you Amanda :)

PR #2131, opened 2025-11-25
Improve geometry track view testing

sethrj:
Part of #2111 : I'm adding better error checking to the vecgeom track view, and this improves testing for the checked geo track view as part of it. It enables the use of the CheckedGeoTrackView for the tracking tests, which required deleting the `find_safety` calculation inside the `move_inside` check. (This fixed the G4 failure alluded to in the `track(` function.)


src/corecel/Assert.hh
```diff
@@ -84,12 +84,6 @@
                   << "failed to open '" << filename
                   << "' (should contain relaxation data)");
  * \endcode
- *
- * An always-on debug-type assertion without a detailed message can be
- * constructed by omitting the stream (but leaving the comma):
- * \code
-    CELER_VALIDATE(file_stream,);
- * \endcode
```
sethrj, line 92: It turns out that doing this causes warnings, so I deleted the text here.

test/geocel/vg/Vecgeom.test.cu
```diff
@@ -52,7 +52,7 @@ __global__ void vgg_test_kernel(DeviceCRef<VecgeomParamsData> const params,
         ids[tid.get() * max_segments + seg]
             = (geo.is_outside()
                    ? -2
-                   : static_cast<int>(geo.impl_volume_id().unchecked_get()));
+                   : static_cast<int>(geo.volume_id().unchecked_get()));
```
sethrj, line 56: it turns out that the volume and impl volume IDs are the same for this problem, regardless of being built by vgdml or g4vg.

mrguilima (approved):
Looks great!
Thanks, @sethrj

test/geocel/GenericGeoTestInterface.cc
```diff
@@ -26,206 +31,144 @@ namespace celeritas
...
-        }
-    }
+    // Length scale and description
+    auto const unit_length = this->unit_length();
+    // Convert from Celeritas native unit system to unit test's internal system
+    auto from_native_length
+        = [scale = unit_length.value](auto&& v) { return v / scale; };
+    auto const tol = this->tracking_tol();
 
+    int remaining_steps = 250;
```
elliottbiondo: Move this back to the top

elliottbiondo (approved):

PR #2133, opened 2025-11-29
Remove bumps and relocation-on-find from vecgeom tracker

sethrj:
**Blocked by #2131 .**

This removes the "bumps" used in the different vecgeom navigators and adds error checking for finding valid distances and actually crossing boundaries. Now tracks that used to get stuck will fail gracefully. It makes the behavior of the different vecgeom versions much more consistent.

Follow-on work will add counters for how many tracks failed. I will also do performance testing before and after this changeset.


sethrj:
@esseivaju @mrguilima These are the changes I was talking about today. It looks like the missing piece was bumping greater than single-precision epsilon, which is what the vecgeom BVH requires.

sethrj:
This is going to be rewritten using the new geo_status to get the edge cases correct.

PR #2137, opened 2025-12-01
Fix uninitialized hit counter in `HitProcessor`

amandalund:
Likely why @rahmans1 was seeing an unexpectedly large number of hits compared to the number of steps.


stognini (approved):
Thanks!

PR #2139, opened 2025-12-03
Reduce ORANGE state size for models with background volumes

elliottbiondo:
This MR fixes a bug that causes `OrangeStateData` to allocate significantly more runtime scratch space than necessary, which has been observed to cause perfectly doable simulations to run out of memory.

`OrangeStateData` stores, among other things, three fields that serve as scratch space during intersection operations: `temp_face`, `temp_distance`, and `temp_isect`, which are sized to be `max_intersections`*`num_tracks`.  This sizing allows for intersection checks to store all potential intersections with a volume, i.e., intersections with all constituent surfaces. The distance to intersection is then found by sorting these intersections by distance and finding the first one that results in a volume sense change upon crossing.

Because any track could encounter an intersection operation with any volume, `max_intersections` must be the global maximum over all volumes. However, prior to this MR, "background" volumes were erroneously included in this maximum. "Background" volumes are volumes that are implicitly defined as any space that does not exist within another volume within a given universe. This is useful in cases where a universe contains, for example, many complex components surrounded by air. Technically, the background volume has the outer boundaries of all of its daughters as constituent surfaces, which causes `max_intersections` within background volumes to be high---potentially orders of magnitude higher than any other volume in the geometry. However, intersection operations within background volumes *do not* check for intersections with all of these constituent surfaces at once. Instead, a Bounding Interval Hierarchy (BIH) is used to test for intersections with daughter volumes one at a time (with acceleration). Thus, `max_intersections` only needs to be as large as the `max_intersections` of the largest non-background volume.

This is accomplished by simply setting `max_intersections` to zero for background volumes. This approach is appropriate because `max_intersections` is only used as a global quantity for allocating data. While `max_intersections` is stored on each `VolumeRecord`, this is only used for DBC checks. A value of zero means that DBC checks will fail when erroneously trying to use the standard (non-BIH) intersection algorithm on a background volume.

### DIFF
--- a/src/orange/detail/UnitInserter.cc
+++ b/src/orange/detail/UnitInserter.cc
@@ -552,15 +552,26 @@ VolumeRecord UnitInserter::insert_volume(SurfacesRecord const& surf_record,
     auto params_cref = make_const_ref(*orange_data_);
     LocalSurfaceVisitor visit_surface(params_cref, surf_record);
 
-    // Mark as 'simple safety' if all the surfaces are simple
+    // Mark this volume as "simple safety" if all of its constituent surfaces
+    // support it, even in the case where the volume is implicit
     bool simple_safety = true;
-    size_type max_intersections = 0;
-
     for (LocalSurfaceId sid : v.faces)
     {
         simple_safety = simple_safety
                         && visit_surface(SimpleSafetyGetter{}, sid);
-        max_intersections += visit_surface(NumIntersectionGetter{}, sid);
+    }
+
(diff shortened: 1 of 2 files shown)


sethrj:
Before:
old
After this plus on-the-fly sense state and #2127 :
new

This seems to have sped up the more complex geometry (on my laptop CPU) by a factor of ten or more! All the detectors still seem to raytrace correctly. Nice work figuring this out as such a small change now that we have BIH.

sethrj (approved):
Awesome!

PR #2141, opened 2025-12-03
Support RANLUX++ as core RNG

sethrj:
This reduces the sample size of the ranlux++ generator to 32 bits and uses the Celeritas `GenerateCanonical` (rather than the host-only `std` version) to concatenate the samples for 64-bit sampling requirements.

### DIFF
--- a/src/corecel/random/engine/RanluxppRngEngine.hh
+++ b/src/corecel/random/engine/RanluxppRngEngine.hh
@@ -14,8 +14,11 @@
 
 #include "corecel/Assert.hh"
 #include "corecel/random/data/RanluxppRngData.hh"
-#include "corecel/random/engine/detail/RanluxppImpl.hh"
+#include "corecel/random/distribution/GenerateCanonical.hh"
+#include "corecel/random/distribution/detail/GenerateCanonical32.hh"
 #include "corecel/sys/ThreadId.hh"
+
+#include "detail/RanluxppImpl.hh"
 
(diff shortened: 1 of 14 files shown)


src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -114,6 +105,29 @@ class RanluxppRngEngine
     RanluxppRngState* state_;
 };
 
+//---------------------------------------------------------------------------//
+/*!
+ * Specialization of GenerateCanonical for XorwowRngEngine.
```
davidsgr: Change to "Specialization of GenerateCanonical for RanluxppRngEngine.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -78,33 +80,22 @@ class RanluxppRngEngine
...
     {
-        return celeritas::numeric_limits<RanluxppUInt>::max();
+        return celeritas::numeric_limits<result_type>::max();
     }
 
     // Initialize state with the given seed.
     inline CELER_FUNCTION RanluxppRngEngine&
     operator=(Initializer_t const& init);
 
     // Generate a double-precision random number
```
amandalund:
```suggestion
    // Generate a 32-bit random number
```

amandalund (approved):
Looks good!

PR #2142, opened 2025-12-03
Only include CUDA RDC utils when using CUDA-enabled VecGeom

sethrj:
Avoid potentially confusing configuration messages and cache variables by only including and distributing the CUDA RDC macros when VecGeom CUDA is available.


sethrj:
Draft: fails with HIP builds due to cuda_rdc_get_sources_and_options

pcanal:
> Avoid potentially confusing configuration messages and cache variables

Those are gone with https://github.com/celeritas-project/celeritas/pull/2143

sethrj:
#2143 is good enough.

PR #2147, opened 2025-12-05
Set up A100 environment on ExCL

sethrj:
This adds scripts for the somewhat dated but still functional `milan0` system, which has two A100s. ROOT is not yet available on the system: only the non-hidden presets defined in `milan0.json` should work.

Hint for those working on multiple systems sharing the same home directory over NFS: use `git worktree add ../celeritas-$(hostname -s)`  to create a local checkout that shares the same git directory, allowing you to `git push` in one and `git pop` in the other (for example).


scripts/cmake-presets/milan0.json
```diff
@@ -0,0 +1,96 @@
...
+        "CMAKE_CXX_EXTENSIONS": {"type": "BOOL", "value": "OFF"},
+        "CMAKE_CUDA_FLAGS": "-Werror all-warnings",
+        "CMAKE_CUDA_HOST_COMPILER": "/usr/bin/c++",
+        "CMAKE_CUDA_ARCHITECTURES": "80",
+        "CMAKE_INSTALL_PREFIX": "/scratch/$env{USER}/install/celeritas-${presetName}",
+        "CMAKE_EXPORT_COMPILE_COMMANDS": {"type": "BOOL", "value": "ON"}
+      }
+    },
+    {
+      "name": ".orange",
```
LSchwiebert, line 29: Is this missing "hidden": "true",

If so, then it probably is needed in hudson.json, too.

LSchwiebert:
On the release build, the following three tests failed because the CPU and GPU results differ. The reldeb build is fine. It's unlikely to be due to your patch, though.

        234 - celeritas/geo/Geometry:SimpleCms* (Failed)        gpu unit
        235 - celeritas/geo/Geometry:TestEm3* (Failed)          gpu unit
        236 - celeritas/geo/Geometry:ThreeSpheres* (Failed)     gpu unit

None of the -orange builds are working. They are all generating errors like the following:

/usr/bin/ld: CMakeFiles/orange_final.dir/cmake_device_link.o: in function `__cudaRegisterLinkedBinary_8f27958b_8_SOA3D_cu_a998c327_3337491':
/tmp/tmpxft_00098591_00000000-3_cmake_device_link.reg.c:79: undefined reference to `__fatbinwrap_8f27958b_8_SOA3D_cu_a998c327_3337491'

The default and full builds aren't working, but I assume that's expected.

sethrj:
@LSchwiebert Oops thanks for that! ~~Looks like we're pulling in G4VG when we shouldn't be, and that's causing CUDA RDC to activate 😬~~ An ordering issue meant that VecGeom was still on, which leads to conflicts with the -WL,-z,-defs.

LSchwiebert:
The release build is still failing these three tests because the CPU and GPU are producing different results, but I'm approving the PR because I don't think that's causing these failures.

        232 - celeritas/geo/Geometry:SimpleCms* (Failed)        gpu unit
        233 - celeritas/geo/Geometry:TestEm3* (Failed)          gpu unit
        234 - celeritas/geo/Geometry:ThreeSpheres* (Failed)     gpu unit

The remaining builds are working, except for full and default.

LSchwiebert (approved):
The various builds are all working, except as noted in my comment.

sethrj:
@Rashika-Gupta or @stognini or @elliottbiondo : I updated the environment so that it stores caches in `/scratch/${USER}` for better performance; it works for me but I'd appreciate another check besides Loren :)

PR #2148, opened 2025-12-05
Fix document warnings related CI infrastructure

sethrj:
Warnings in the sphinx documentation were accidentally disabled in (I think) #1988 because warnings are now shown during doctree generation rather than html construction. I fixed the couple errors in the docs that have accumulated during that time.

Related, I've removed a now-unnecessary workaround for https://github.com/actions/checkout/issues/2041 , and I've added code to get the build size and clean the build directory after it's last used, since we now seem to see intermittent failures on the CUDA builds https://github.com/celeritas-project/celeritas/actions/runs/19948559972/job/57204118695 .

To fix the size overrun errors (nearly 5GB for the debug/reldeb) I've disabled CUBIN compilation for the reldeb case and turned off DEVICE_DEBUG for the debug case.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -137,7 +137,7 @@ jobs:
         run: |
           echo "::group::Install"
           cmake --install .
-          echo "::endgroup::
+          echo "::endgroup::"
           cmake --build . --target clean
       - name: Check installation
         if: >-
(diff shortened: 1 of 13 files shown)


pcanal:
> I've removed a now-unnecessary workaround for https://github.com/actions/checkout/issues/2041

Why is it now unnecessary? (The linked issue is not explicitly marked resolved).

sethrj:
@pcanal The issue I posted was primarily because of a bug in Git that was present for one minor patch version that seems to have been quickly fixed and deployed to the latest github runners.

pcanal (approved):
LGTM.

PR #2149, opened 2025-12-05
Add CHIPS gamma-nuclear cross sections 

whokion:
This MR adds gamma–nuclear cross sections (CHIPS) in the high-energy region.


src/celeritas/g4/EmExtraPhysicsHelper.hh
```diff
@@ -0,0 +1,41 @@
...
+#include <memory>
+
+class G4GammaNuclearXS;
+
+namespace celeritas
+{
+
+//---------------------------------------------------------------------------//
+/*!
+ * A helper class to interface with Geant4 cross sections.
```
sethrj: I don't understand the purpose or interface of this class. Are you expecting to add more methods to it? Is it supposed to have the same interface as a particular Geant4 class? Will it be used outside of the GammaNuclearModel data construction?

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +112,64 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    size_type nbin_high = 250;
+    size_type nbin_ultra = 50;
+
+    result.x.resize(nbin_high + nbin_ultra);
+    result.y.resize(nbin_high + nbin_ultra);
+
+    // Bin sizes in logarithmic energy scale
+    double const de_high = (std::log(emid) - std::log(emin)) / (nbin_high - 1);
+    double const de_ultra = (std::log(emax) - std::log(emid))
+                            / (nbin_ultra - 1);
```
sethrj: I think you should be able to include `corecel/grid/VectorUtils.hh` and use:

```c++
// Concatenate two log-energy scales
xs.x = geomspace(emin, emid, nbin_high - 1);
xs.x.pop_back();
auto ultra = geomspace(emid, ehigh, nbin_ultra - 1);
xs.x.insert(xs.x.end(), ultra.begin(), ultra.end());
```

I think the terms "mid", "high", and "ultra" could use a consistency check.

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +112,64 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    for (size_type i = 0; i < nbin_high + nbin_ultra; ++i)
+    {
+        double energy
+            = (i < nbin_high)
+                  ? std::exp(std::log(emin) + de_high * i)
+                  : std::exp(std::log(emid) + de_ultra * (i - nbin_high + 1));
+
+        result.x[i] = energy;
+        xs.value()
+            = helper_->GammaNuclearElementXS(energy, atomic_number.get());
```
sethrj:
```suggestion
        MmSqMicroXs xs{helper_->GammaNuclearElementXS(energy, atomic_number.get())};
```
but if you're going to make a wrapper to Geant4's helper, you should ensure it has units.

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +112,64 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+//---------------------------------------------------------------------------//
+/*!
+ * Build CHIPS gamma-nuclear element cross sections using G4GammaNuclearXS
+ * above the upper energy limit of G4PARTICLEXS/gamma (IAEA) data. The cross
+ * sections are derived from the parameterization developed by M. V. Kossov
+ * (CERN/ITEP Moscow) in the high energy region (106 MeV < E < 50 GeV) and
+ * from a Reggeon-based parameterization in the ultra-high-energy region
+ * (E > 50 GeV). G4GammaNuclearXS uses CHIPS (G4PhotoNuclearCrossSection)
+ * above 150 MeV and performs linear interpolation between the upper energy
+ * limit of G4PARTICLEXS/gamma (IAEA) data and 150 MeV.
```
sethrj: Our user/methods manual pulls in documentation only from the class documentation (in the `.hh` file): should we move this to the class docs where it's more visible? Does the Kossov citation have a reference?

src/celeritas/em/model/GammaNuclearModel.hh
```diff
@@ -31,6 +34,10 @@ class GammaNuclearModel final : public Model, public StaticConcreteAction
     //!@{
     using MevEnergy = units::MevEnergy;
     using ReadData = std::function<inp::Grid(AtomicNumber)>;
+    using result_type = inp::Grid;
```
sethrj: This isn't the result type of the model itself; if you want to define a helper class `ChipsXsCalc`, that would have an `operator()(...) -> result_type`.

src/celeritas/em/model/GammaNuclearModel.hh
```diff
@@ -31,6 +34,10 @@ class GammaNuclearModel final : public Model, public StaticConcreteAction
...
+    using BarnXs = units::BarnXs;
+    using MmSqMicroXs
+        = Quantity<UnitProduct<units::Millimeter, units::Millimeter>, double>;
     using HostRef = HostCRef<GammaNuclearData>;
```
sethrj: These aren't part of the class's public interface, so they should be constrained to where they're used.

src/celeritas/em/data/GammaNuclearData.hh
```diff
@@ -61,16 +55,19 @@ struct GammaNuclearData
     // Scalar data
     GammaNuclearScalars scalars;
 
-    // Microscopic (element) cross section data (G4PARTICLEXS/gamma/inelZ)
-    ElementItems<NonuniformGridRecord> micro_xs;
+    // Microscopic cross sections using G4PARTICLEXS/gamma nuclear (IAEA) data
+    ElementItems<NonuniformGridRecord> micro_xs_iaea;
+    Items<real_type> reals_iaea;
```
sethrj: You should be able to use the same `reals` backend array for both grid types.

sethrj (commented):
Thanks @whokion .

whokion:
@sethrj Thanks a lot for your review and comments - will take care of them accordingly. Also, I find that there is a link error.  Do you have any quick idea to resolve it or am I doing something wrong?  Thanks again.

src/celeritas/g4/EmExtraPhysicsHelper.hh
whokion: Yes, I plan to add additional methods for other cross sections used in the ElectroNuclear and potentially the MuonNuclear models. Nevertheless, I am open to adopting an alternative approach if there is a better way to provide a similar wrapper to corresponding Geant4 methods.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: Nice utility!  I use it accordingly.  Thanks.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: Updated.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: I move the description to the header and add a citation for the CHIPS gamma-nuclear cross section parameterization in the doc  - assume that I do not need to add the paper itself to zotero"

src/celeritas/em/model/GammaNuclearModel.hh
whokion: Deleted unncessary type definitions.

src/celeritas/em/model/GammaNuclearModel.hh
whokion: Move it to the cc.

src/celeritas/em/data/GammaNuclearData.hh
whokion: Used a common backend array for both cross section grids.

src/celeritas/em/model/GammaNuclearModel.cc
sethrj: Zotero exports the bibliography entries into our references, so we do need it there, so I've added it myself. Thanks!

sethrj:
@whokion Thanks for the quick update. Regarding the link error: we support Geant4-free builds for fast CI testing and mini-apps on HPC platforms, so you'll need to add `#if !CELERITAS_USE_G4 ...` methods at the end of the `EmExtraPhysicsHelper.hh` file as is done with our other wrappers:
```c++
#if !CELERITAS_USE_GEANT4
inline GeantImporter::GeantImporter()
{
    CELER_NOT_CONFIGURED("Geant4");
}

inline GeantImporter::GeantImporter(GeantSetup&&)
{
    CELER_NOT_CONFIGURED("Geant4");
}

inline ImportData GeantImporter::operator()(DataSelection const&)
{
    CELER_ASSERT_UNREACHABLE();
}
#endif
```

src/celeritas/g4/EmExtraPhysicsHelper.hh
sethrj: Now that you've mentioned linking errors, I understand the purpose of this class: you'd probably tried to call the methods directly earlier and found that you couldn't include G4 headers directly from the `em/model` directory. So we do indeed need a G4 XS interface class.

Could you document (for now) that this is why the class is needed?

src/celeritas/g4/EmExtraPhysicsHelper.hh
```diff
@@ -0,0 +1,41 @@
...
+ */
+class EmExtraPhysicsHelper
+{
+  public:
+    // Construct EM extra physics helper
+    EmExtraPhysicsHelper();
+
+    // Calculate gamma-nuclear element cross section
+    double GammaNuclearElementXS(double energy, int z);
```
sethrj: Since this is an interface class and not meant to plug in to existing Geant4 code, please use style and unit conventions from Celeritas:
```c++
XsMmSq calc_gamma_nuclear_xs(AtomicNumber z, MevEnergy energy) const;
```

Also, if you're going to reuse this for multiple XS types, it could be changed in the future to simply hold a `G4VCrossSectionDataSet` shared pointer, and be initialized with `"GammaNuclearXS"` or `"PhotoNuclearXS"`  so that it can use `G4CrossSectionDataSetRegistry::Instance()->GetCrossSectionDataSet(name)` as the physics list constructors do.

And if you *don't* need `max_high_energy` below, then you could make this class a function-like `DatasetMicroXsCalculator` (or whatever) with an `operator()`.

src/celeritas/g4/EmExtraPhysicsHelper.hh
```diff
@@ -0,0 +1,41 @@
...
+
+    // The maximum high energy of G4PhotoNuclearCrossSection
+    static constexpr double max_high_energy()
+    {
+        return 5e+4;  // clhep::MeV
```
sethrj: Is there any way to get this programmatically from Geant4? Or should this live in GammaNuclearModel?

src/celeritas/em/model/GammaNuclearModel.hh
```diff
@@ -62,13 +77,19 @@ class GammaNuclearModel final : public Model, public StaticConcreteAction
 
   private:
     //// DATA ////
+    std::shared_ptr<EmExtraPhysicsHelper> helper_;
```
sethrj, line 81: This doesn't need to be class data; it should be constructed temporarily at load time.

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +111,51 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    result.x.resize(nbin_total);
+    result.y.resize(nbin_total);
+
+    result.x = geomspace(emin, emid, nbin_total - nbin_ultra);
+    result.x.pop_back();
+    auto ultra = geomspace(emid, emax, nbin_ultra + 1);
+    result.x.insert(result.x.end(), ultra.begin(), ultra.end());
+
+    // Tabulate the cross section from emin to emax
+    Quantity<UnitProduct<units::Millimeter, units::Millimeter>, double> xs;
```
sethrj: You can make this a type alias inside the `DatasetMicroXsCalculator`.

sethrj (commented):
Looking good @whokion !

src/celeritas/g4/EmExtraPhysicsHelper.hh
whokion: I added a short class description and will expand it as I add more methods.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: Thanks!

src/celeritas/g4/EmExtraPhysicsHelper.hh
whokion: I updated the method as suggested. Thanks for the additional recommendations! I will consider and take into account the suggestion when adding ElectroNuclearCrossSection and other features in following MRs.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: Move it to EmExtraPhysicsHelper for now.

src/celeritas/em/model/GammaNuclearModel.hh
whokion: Sorry that I did not fully get this suggestion.  Thanks for clarification.

src/celeritas/g4/EmExtraPhysicsHelper.hh
whokion: There is no G4 interface for this - it was temporarily moved to the `GammaNuclearModel::calc_chips_xs` method as it is no longer needed when the high energy micro cross section is calcuated.

whokion:
@sethrj Can you consider to merge this MR?

src/celeritas/em/data/GammaNuclearData.hh
```diff
@@ -61,16 +55,18 @@ struct GammaNuclearData
...
 
+    // Microscopic cross sections using parameterized CHIPS data at high energy
+    ElementItems<NonuniformGridRecord> xs_chips;
+
     //! Whether the data are assigned
     explicit CELER_FUNCTION operator bool() const
     {
-        return scalars && !micro_xs.empty() && !reals.empty();
+        return scalars && !xs_iaea.empty() && !reals.empty()
+               && !xs_chips.empty();
```
amandalund:
```suggestion
               && xs_chips.size() == xs_iaea.size();
```

src/celeritas/g4/EmExtraPhysicsHelper.cc
```diff
@@ -0,0 +1,36 @@
...
+/*!
+ * Calculate the gamma-nuclear element cross section using G4GammaNuclearXS.
+ */
+auto EmExtraPhysicsHelper::calc_gamma_nuclear_xs(AtomicNumber z,
+                                                 MevEnergy energy) const
+    -> MmSqXs
+{
+    MmSqXs xs;
+    xs.value() = gn_xs_->ElementCrossSection(energy.value(), z.get());
+    return xs;
```
amandalund:
```suggestion
    return MmSqXs{gn_xs_->ElementCrossSection(energy.value(), z.get())};
```

test/celeritas/em/GammaNuclear.test.cc
amandalund, line 77: Looks like because of the `EmExtraPhysicsHelper` you'll have to add `_needs_geant4` for this test.

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +111,49 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    result.x = geomspace(emin, emid, nbin_total - nbin_ultra);
+    result.x.pop_back();
+    auto ultra = geomspace(emid, emax, nbin_ultra + 1);
+    result.x.insert(result.x.end(), ultra.begin(), ultra.end());
+
+    // Tabulate the cross section from emin to emax
+    EmExtraPhysicsHelper::MmSqXs xs;
+    for (size_type i = 0; i < nbin_total; ++i)
+    {
+        xs = helper_->calc_gamma_nuclear_xs(z, MevEnergy{result.x[i]});
```
amandalund: Since it's only used inside the loop, to simplify scope:
```suggestion 
    for (size_type i = 0; i < nbin_total; ++i)
    {
        auto xs = helper_->calc_gamma_nuclear_xs(z, MevEnergy{result.x[i]});
```

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +111,49 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    // to 50 GeV) and the calculations used in G4PhotoNuclearCrossSection,
+    // which can be made configurable if needed (TODO). Note that the linear
+    // interpolation between the upper limit of the IAEA cross-section data
+    // and 150 MeV, as used in G4GammaNuclearXS, is also included in this
+    // tabulation.
+    double const emid = 5e+4;
+    size_type nbin_total = 300;
+    size_type nbin_ultra = 50;
+
+    result.x.resize(nbin_total);
```
amandalund: The resize is redundant since `result.x` is overwritten below:
```suggestion
```

src/celeritas/em/model/GammaNuclearModel.cc
```diff
@@ -98,5 +111,49 @@ void GammaNuclearModel::step(CoreParams const&, CoreStateDevice&) const
...
+    // Tabulate cross sections using separate parameterizations for the high
+    // energy region (emin < E < 50 GeV) and the ultra high energy region up
+    // to the maximum valid energy (emax). The numbers of bins are chosen to
+    // adequately capture both the parameterized points (224 bins from 106 MeV
+    // to 50 GeV) and the calculations used in G4PhotoNuclearCrossSection,
+    // which can be made configurable if needed (TODO). Note that the linear
+    // interpolation between the upper limit of the IAEA cross-section data
+    // and 150 MeV, as used in G4GammaNuclearXS, is also included in this
+    // tabulation.
+    double const emid = 5e+4;
```
amandalund, line 137: An extra comment for this variable (including units) might improve clarity.

amandalund (approved):
Nice @whokion, thanks for the addition!

src/celeritas/em/data/GammaNuclearData.hh
whokion: Change as suggested.

src/celeritas/g4/EmExtraPhysicsHelper.cc
whokion: Simplified the code as suggested.

test/celeritas/em/GammaNuclear.test.cc
whokion: Added ${_needs_geant4} for GammaNuclearTest.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: SImplified the code as suggested.

src/celeritas/em/model/GammaNuclearModel.cc
whokion: Added a breief description for emid.

whokion:
@amandalund Thanks for your review and comments!

amandalund:
@whokion not sure why the tests are crashing... have there been changes in `G4GammaNuclearXS::ElementCrossSection()` that might cause older versions  to segfault or produce different results?

whokion:
@amandalund  Yes, it appears that there are several issues related to the use of G4GammaNuclearXS: 1) It was introduced in Geant4 10.7, so it cannot support earlier Geant4 versions, 2) The public interface G4GammaNuclearXS::ElementCrossSection(energy, z) was added in Geant4 11.0, which may be replaced with G4GammaNuclearXS::GetElementCrossSection(energy, z, material) to support 10.7, 3) The low-energy gamma–nuclear cross-section data have been updated several times recently. In the unit tests, we use copied data files from Geant4 11.3.2 (G4PARTICLEXS4.1), which may differ in format from older data sets. I still do not understand why it causes failures with Geant4 versions 11.0, 11.1, and 11.2 as the maximum value of the low-energy cross section (emid in celeritas::GammaNuclearModel) is automatically extracted from the data and should define the starting point of the transition region. Although I will investigate this further, it seems that the current approach may be problematic for supporting older Geant4 versions and may require an alternative strategy.

sethrj:
@whokion it's ok to disable this for g4<11 since the LHC experiments are the only ones that need it. Sorry for the delay also, I'm on vacation till Tuesday :)

whokion:
So far, I could not reproduce the test failure with Geant4 11.0.p04.

whokion:
@sethrj I have addressed all CI build test failures except for three tests with Geant4 releases prior to 11.0, in which `G4GammaNuclearXS` was not supported. Please let me know if there is anything else that needs to be addressed before merging. Thanks.

sethrj:
@whokion Are you planning to address the failures yourself (disabling code and skipping tests) or do you need a hand to help get the CI passing?

whokion:
@sethrj Your help will be appreciated - I do not know how to skip them.  Thanks.

sethrj:
@whokion Please see 047538775f59ba1e7b7d9c0f8bace62e2eec2b4f for reference 😄 Note that `CELERITAS_GEANT4_VERSION` is zero if geant4 is not enabled, and the value is hex-encoded (major, minor, patch) with 2 hex digits each. (This helps avoid the problem that version 010000 is actually *octal*)

PR #2113, opened 2025-11-17
Add CODEOWNERS and update roles

sethrj:
This adds a preliminary set of "code owners" (i.e. experts/responsible reviewers) and a brief discussion of its meaning. It also updates the roles to reflect our current practices, and removing "maintainers" (which didn't work as expected: being able to approve a PR required the same "write" privilege as being able to hit the "merge" button). It also proposes the "affiliate" role to replace "student".

@celeritas-project/core Please add code suggestions to add or remove yourself, adding new directories as needed.

PR #2150, opened 2025-12-05
Simplify CODEOWNERS

sethrj:
#2113 has led to too many people being requested by default as codeowners. This reduces the numbers of reviewers requested.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,9 +4,6 @@
 # Note that later lines take precedence over earlier ones.
 #-----------------------------------------------------------------------------#
 
-# Default all changes to ping the code lead if not overridden below
-* @celeritas-project/code-lead
-
 #-------------------------------------#
 # Topics
 #-------------------------------------#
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
pcanal, line 14: github reports:
```
Unknown owner on line 15: make sure the team @celeritas-project/core-advisor exists, is publicly visible, and has write access to the repository
```

pcanal: But @celeritas-project/core-advisor seems to exist :( ...

pcanal:
> has led to too many people being requested by default as codeowners

Which PR has too many (eg. #2149 is seemingly 'correct' except for the code-lead being added because of the missing coverage for the test directories)?

.github/CODEOWNERS
```diff
@@ -1,12 +1,9 @@
-#-----------------------------------------------------------------------------#
+#-----------------------------*- codeowners -*--------------------------------#
 # CELERITAS PROJECT CODEOWNERS
 # See discussion in /doc/development/administration.rst .
 # Note that later lines take precedence over earlier ones.
 #-----------------------------------------------------------------------------#
 
-# Default all changes to ping the code lead if not overridden below
-* @celeritas-project/code-lead
```
pcanal, line 8: Not having this is fine per se as it 'only' means that for PR that contains only files that are not matching any patterns, there wont be any auto assignment.

pcanal (approved):
LGTM (although I would call this a simplification but rather an extension/correction to cover all top level directories in addition to `src`)

.github/CODEOWNERS
sethrj: It's not quite that simple: *every PR* that contains files not matching any other reviews gets an automatic review from code-lead (which is redundant with me and distracts from things that really need the "lead" attention).

pcanal: Side note `example` and `app` and most of `doc` and `scripts` do not have any owners yet as they have a different sub-directory structure.

sethrj:
> > has led to too many people being requested by default as codeowners
> 
> Which PR has too many (eg. #2149 is seemingly 'correct' except for the code-lead being added because of the missing coverage for the test directories)?

I think @amandalund ended up deselecting a bunch of irrelevant people for a recent refactor: some of my pulls have shown up with like six people . I wish we could distinguish "give people a heads up that this code is being touched" from "please take a look at this and give me feedback" .

.github/CODEOWNERS
sethrj: where does it report this?

pcanal: A few of places including: https://github.com/celeritas-project/celeritas/blob/develop/.github/CODEOWNERS
and [`11d286e` (#2150)](https://github.com/celeritas-project/celeritas/pull/2150/commits/11d286edc6bd405129ccc9d27eaac5aae89b99ce)

.github/CODEOWNERS
pcanal: > very PR that contains files not matching any other reviews

And/or is it an indication that those files don't have a owner yet :)

amandalund:
Yes, I didn't need a review from half the team in #2146 for a minor reorganization... but since it touched files in many different places maybe it's difficult to avoid.

.github/CODEOWNERS
pcanal: > (which is redundant with me and distracts from things that really need the "lead" attention)

Indeed the system is missing a 'ping me only if nobody else is pinged' ...

pcanal:
> I didn't need a review from half the team in https://github.com/celeritas-project/celeritas/pull/2146 for a minor reorganization...

True but conversely they (maybe) needed to be made aware of the change ... (i.e. Seth's distinction between `please review` vs `please take note`).

.github/CODEOWNERS
sethrj: Aha thanks. It's because core-advisor only has "triage" not "write" privileges. I'll remove.

sethrj:
I think "reviewers" should be people who are requested to review the code, since at least one must do so. A secondary function ('pinging' people who have worked recently on the given files) could be implemented with a bot tagging people who have git history with the files (and we could maybe filter on PRs that aren't 'minor').

PR #2151, opened 2025-12-05
Hide known warnings and reduce verbosity related to Geant4 geometry and tracker

sethrj:
This reduces some of the verbosity in the output from the Geant geometry track view wrapper, which I've been using to validate tracking test cases.


amandalund (approved):
Thanks!

PR #2152, opened 2025-12-06
Use correct energy when generating Cherenkov and scintillation photons

amandalund:
This addresses the incorrect sampling of Cherenkov and scintillation photons when the track undergoes a collision at the end of the step. Because the distribution data was offloaded at the user-post step point, energy lost in the discrete interaction was incorrectly included in the energy loss used to sample the number of optical photons and photon times.

For Cherenkov the only change needed to fix this was to move the offload action from `user_post` to `pre_post`. Scintillation was a bit more complicated because we need the pre-step quantities, the speed following the slowing down but before the discrete interaction, _and_ the energy deposition following the discrete interaction. I've kept the offload action for scintillation as user-post so we can get the post-interaction energy deposition and added a second step gather action to collect the pre-post data (by templating the `OffloadGatherAction` on `StepActionOrder`).

I updated the scintillation photon position sampling to use the fraction of energy deposited locally at the collision site to determine whether to generate photons at the endpoint rather than basing it off whether the particle is charged, but reverted it for now since I wasn't sure it was an improvement (https://github.com/celeritas-project/celeritas/commit/6cfab205482457749c68e6d63a557236396d45ff).


src/celeritas/global/CoreState.hh
```diff
@@ -202,6 +203,8 @@ auto CoreState<M>::aux_data(AuxId auxid) -> StateRef<S>&
     CELER_EXPECT(auxid < aux_state_->size());
 
     // TODO: use "checked static cast" for better runtime performance
+    CELER_LOG(info) << "Accessing aux state with type "
```
sethrj: I assume this is a debug statement accidentally left in?

sethrj (approved):
Looks good @amandalund , thanks! It would be nice if we could have some more granular test like AlongStep.test.cc or Stepper.

src/celeritas/global/CoreState.hh
amandalund: I added it because I can't reproduce these test failures locally :(

amandalund:
Ok @sethrj this is working now.

> I updated the scintillation photon position sampling to use the fraction of energy deposited locally at the collision site to determine whether to generate photons at the endpoint rather than basing it off whether the particle is charged, but reverted it for now since I wasn't sure it was an improvement (https://github.com/celeritas-project/celeritas/commit/6cfab205482457749c68e6d63a557236396d45ff).

Should I leave the scintillation sampling as is?

src/celeritas/optical/gen/OffloadAction.cc
```diff
@@ -122,16 +123,22 @@ template<GeneratorType G>
 void OffloadAction<G>::offload(CoreParams const& core_params,
                                CoreStateHost& core_state) const
 {
-    auto& step = core_state.aux_data<OffloadStepStateData>(data_.step_id);
+    auto& pre_step
+        = core_state.aux_data<PreTraitsT::template Data>(data_.pre_step_id);
```
sethrj, line 127: Did you figure out why this was needed? 😕

sethrj (approved):
Thanks! Regarding 6cfab205482457749c68e6d63a557236396d45ff , it seems like the physically correct thing to do, but it deserves a highlight of its own, so perhaps make it a new PR?

src/celeritas/optical/gen/OffloadAction.cc
amandalund: The dynamic cast was failing when I used different aliases in the offload gather and offload actions:
```console
C++ exception with description "runtime error: failed to access aux state with ID: 3, stored type: N9celeritas8AuxStateINS_6detail19OffloadGatherTraitsILNS_15StepActionOrderE4EE4DataELNS_8MemSpaceE0EEE, expected type: N9celeritas8AuxStateINS_19OffloadPreStateDataELNS_8MemSpaceE0EEE
/home/runner/work/celeritas/celeritas/src/celeritas/global/CoreState.hh:212: 'state' failed" thrown in the test body.
```

amandalund:
Sounds good, I can make that change in a follow on.

src/celeritas/optical/gen/OffloadAction.cc
sethrj: That's weird... compiler bug or misunderstanding on our part? I definitely know I've seen some "surprising" limitations with templates-of-templates.

amandalund: [this is an older post](https://stackoverflow.com/questions/30650073/c-dynamic-downcasting-to-class-template-having-template-template-parameter-bei), but it suggests that the dynamic cast returning a null pointer was the correct behavior.

PR #2138, opened 2025-12-02
Add ExCL MI300A environment

sethrj:
This adds a build environment for [the Faraday system](https://docs.excl.ornl.gov/system-overview/faraday) on ExCL. I've successfully built on the MI300X (though covfie seems to have errors @esseivaju !): all tests pass.

PR #2153, opened 2025-12-09
Support builds on fnal-dev-sl7 on scisoftbuild01

sethrj:
This fixes a couple of minor build errors needed to build Celeritas against the larsoft toolchain. It adds a somewhat complicated `scisoftbuild01.sh` that loads the basic environments for running inside an apptainer:
- Sets up DUNE UPS with `setup_dune.sh`
- Sets up larsoft development with `setup larsoft ...`
- Creates an MRB dev environment with `mrb newDev`
- Activates the environement with `. .../setup`
- Clones larg4 with `mrb g`
- Loads CMake and other larg4 dependencies with `mrbsetenv`

```sh
# Set up scratch directory on bare metal
. scripts/env/scisoftbuild01.sh
# Start apptainer, forwarding necessary directories
/cvmfs/oasis.opensciencegrid.org/mis/apptainer/current/bin/apptainer \
  shell --shell=/bin/bash \
  -B /cvmfs,$SCRATCHDIR,$HOME,$XDG_RUNTIME_DIR,/opt,/etc/hostname,/etc/hosts,/etc/krb5.conf  \
  --ipc --pid  \
  /cvmfs/singularity.opensciencegrid.org/fermilab/fnal-dev-sl7:latest
```


src/geocel/ScopedGeantLogger.cc
```diff
@@ -202,7 +202,7 @@ enum class SGLState
 
 SGLState& sgl_state()
 {
-    G4ThreadLocal SGLState result{SGLState::inactive};
+    static G4ThreadLocal SGLState result{SGLState::inactive};
```
pcanal, line 205: What was the error (message) without the `static`?

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+  fi
+fi
+
+# Install larg4
+if [ -n "${MRB_SOURCE}" ]; then
+  _pkg=larg4
+  if ! [ -d "${MRB_SOURCE}/${_pkg}" ]; then
+    celerlog info "Installing ${_pkg}"
+    mrb g ${_pkg}
+  fi
```
stognini:
```suggestion
# Checkout arbitrary package so that mrbsetenv can be fully executed
if [ -n "${MRB_SOURCE}" ]; then
  _pkg=larg4
  if ! [ -d "${MRB_SOURCE}/${_pkg}" ]; then
    celerlog info "Cloning ${_pkg} via mrb"
    mrb g ${_pkg}
  fi
```

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+
+  # Now that a package exists in MRB source, cmake and dependencies can load
+  celerlog info "Activating MRB environment"
+  mrbsetenv
```
stognini, line 106: Can this be anywhere, or is the script at this point inside the `build_[x]` folder?

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+    chmod 700 "${_scratch}" || return $?
+  fi
+  unset _scratch
+done
+
+# Set up larsoft if running inside an apptainer
+if [ -n "${MRB_PROJECT}" ]; then
+  LARSCRATCHDIR="${SCRATCHDIR}/${MRB_PROJECT}"
+  if ! [ -d "${LARSCRATCHDIR}" ]; then
+    celerlog info "Setting up MRB in ${LARSCRATCHDIR}..."
```
pcanal:
```suggestion
    celerlog info "Creating MRB dev area in ${LARSCRATCHDIR}..."
```

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+      mrb newDev
+    )
+    celerlog debug "MRB environment created"
+  fi
+  _setup_filename="${LARSCRATCHDIR}/localProducts_${MRB_PROJECT}_${MRB_PROJECT_VERSION}_${MRB_QUALS//:/_}/setup"
+  if ! [ -f "${_setup_filename}" ]; then
+    celerlog warn "Expected setup file at ${_setup_filename}: MRB may not have been set up correctly"
+    . "${LARSCRATCHDIR}/localProducts_${MRB_PROJECT}*/setup"
+  else
+    . "${LARSCRATCHDIR}/localProducts_${MRB_PROJECT}_${MRB_PROJECT_VERSION}_${MRB_QUALS//:/_}/setup"
```
pcanal:
```suggestion
    . "${_setup_filename}"
```

pcanal:
>  It adds a somewhat complicated scisoftbuild01.sh that loads the basic environments for running inside an apptainer:

Should the bullets be copy/pasted in script's header as summary/documentation?

> # Set up scratch directory on bare metal

Shouldn't this be somewhere in the repository?  Either in a `.md` or maybe the new script's header?

src/geocel/ScopedGeantLogger.cc
sethrj: Because `G4ThreadLocal` is compiled out to `` in their non-MT build, the warning is that we're returning a reference to a local variable (which it is!)

sethrj: I could make a separate "fix non-MT geant4" PR if you think it deserves a special callout.

scripts/env/scisoftbuild01.sh
sethrj: Is it arbitrary though? I figured larg4 would have the same dependencies as celeritas

scripts/env/scisoftbuild01.sh
sethrj: Yep it can be called from any directory; it seems to use `MRB_DIR` which is set by sourcing the `.../setup` script.

src/geocel/ScopedGeantLogger.cc
pcanal: It is not needed, but this is a bug in G4 :) (They also have `G4ThreadLocalStatic` ... )
[ `it` == `separate PR` ]

scripts/env/scisoftbuild01.sh
stognini: I think the dependencies are already in the environment from the larsoft/dune setup itself. Both `larg4` and `celeritas` are just using UPS products. So I may be wrong, but my understanding is that `mrbsetenv` is just returning early if there are no packages in `srcs/`, and thus does not complete the environment setup (such as adding `cmake` to your path).

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+#-----------------------------------------------------------------------------#
+
+# Reduce I/O metadata overhead by avoiding language translation lookups
+export LC_ALL=C
+
+# Allow running from user rc setup outside of build.sh environment
+if ! command -v celerlog >/dev/null 2>&1; then
+  celerlog() {
+    printf "%s: %s\n" "\$1" "\$2" >&2
```
amandalund:
```suggestion
    printf "%s: %s\n" "$1" "$2" >&2
```

scripts/env/scisoftbuild01.sh
```diff
@@ -0,0 +1,111 @@
...
+  }
+fi
+if [ -z "${SYSTEM_NAME}" ]; then
+  celerlog debug "Set SYSTEM_NAME=${SYSTEM_NAME}"
+  SYSTEM_NAME=$(uname -s)
```
amandalund:
```suggestion
  SYSTEM_NAME=$(uname -s)
  celerlog debug "Set SYSTEM_NAME=${SYSTEM_NAME}"
```

scripts/env/scisoftbuild01.sh
sethrj: Weird, wonder how that happened (this was an error in #2138).

stognini (approved):
Thanks!

Issue #2124, opened 2025-11-20
Support reproducibility independent of thread count and asynchronous ordering


sethrj:
## Current capability

At present, reproducibility in Celeritas requires:
1. keeping the same number of threads across runs,
2. sorting "track initializers" from secondaries (daughter particles) so that they are produced in an ordered slot (or queued for later injection), and
3. preventing new primaries (source particles) from being added during the transport solve.

Additionally, supporting multiple Geant4 CPU threads and framework-selectable events requires us to additionally call `rng_reseed` on each per-CPU state vector at each new event.

## Goal

We want to be able to share a single state vector across multiple Geant4 CPU threads, interleaving EM/optical primaries offloaded from multiple CPU threads (*assume* the same ordering of primaries coming from each CPU thread, regardless of how Celeritas interacts with it). After this and all the subtasks on #2014 are complete, then we can truly have an asynchronous optimized transporter.

There is a side benefit to having RNG states that only depend on the parent: a requirement by #1349 that killing a track prematurely for efficiency has minimal impact on the invidual (not statistical!) behavior of the particles in the rest of the simulation. Currently that's not true because killing tracks causes all the subsequently tracked particles to have different RNG states.

## Implementation plan

1. Implement ranluxpp because it's been deemed "high quality" and can be compared against adept results. (#2141)
2. Add "spawning" from an RNG state. From what I can tell, [the ADePT spawning code](https://github.com/apt-sim/AdePT/blob/3f9b7759ae44c89a66309f24ea33bf6a633c6d8a/include/AdePT/copcore/Ranluxpp.h#L170-L180) takes two successive RNG states in RANLUX space, XORs them, and assumes that the result is well distributed. (With 2^576 RNG states, and ~4 doubles per RNG state, this means we're likely never going to repeat if we're getting a truly well distributed sample. But I feel like we need a mathematician to analyze the correlation between two sequential RNG states and whether xoring them is actually going to be well distributed.) It's unclear if this will work as well for a "shuffling" RNG engine like XORWOW that already has XOR built in to its requirements.
[...]

PR #2155, opened 2025-12-12
Add rng branching

davidsgr:
## Description

Addresses Task 2 in #2124 and adds the capability of spawning a new, decorrelated RNG from an existing RNG.


src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -222,6 +226,47 @@ CELER_FUNCTION auto RanluxppRngEngine::operator()() -> result_type
...
+    CELER_EXPECT(tid < state.state.size());
+
+    RanluxppRngEngine new_rng(params_, state, tid);
+    RanluxppRngState& new_state = *new_rng.state_;
+
+    // Copy the current state into the new state
+    for (auto i : celeritas::range(9))
+    {
+        new_state.value.number[i] = (*state_).value.number[i];
+    }
```
amandalund: We should be able to assign this directly:
```suggestion
    new_state.value.number = state_->value.number;
```

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -222,6 +226,47 @@ CELER_FUNCTION auto RanluxppRngEngine::operator()() -> result_type
...
+ * Branching is performed in two steps.  First, the state of the new RNG
+ * (\f$x^{\prime}\f$) is initialized as
+ * \f[
+ *      x^{i,\prime}_j = x^i_j ^ x^{i+1}_j \, .
+ * \f]
+ * Second, to decorrelate the new RNG from this RNG, the new RNG is
+ * advanced forward to the next block
+ */
+CELER_FUNCTION RanluxppRngEngine RanluxppRngEngine::branch(StateRef const& state,
+                                                           TrackSlotId tid)
```
amandalund: I'm not quite sure I understand how this will be used in practice: my understanding was when a new secondary is produced we'd branch a new RNG from the parent track's state and store the new RNG state in the `TrackInitializer`. If so:
  -  For an RNG engine that's been constructed in a given track slot, which `TrackSlotId` does it use in the `branch()` call?
  - More of an implementation detail, should we be returning a `RanluxppRngEngine`, or just the state that will be stored in the initializer?

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -222,6 +226,47 @@ CELER_FUNCTION auto RanluxppRngEngine::operator()() -> result_type
...
+    RanluxppRngState& new_state = *new_rng.state_;
+
+    // Copy the current state into the new state
+    for (auto i : celeritas::range(9))
+    {
+        new_state.value.number[i] = (*state_).value.number[i];
+    }
+
+    // Advance the RNG
+    this->advance();
```
amandalund: It looks like in AdePT the new RNG state copies the original state (including `position`) after advancing the original, while here it's copied before advancing. The new state here is also XORed with the original state after it's been advanced (vs. the state before advancing in AdePT). Is this any different in practice?

amandalund (commented):
Thanks @davidsgr! I haven't given any consideration to branching with xorwow yet, but here's a couple initial questions/comments.

src/corecel/random/engine/RanluxppRngEngine.hh
sethrj: Agreed, we need to return a new state that will be later used to initialize a track slot. (For Ranlux++ this will be a *subset* of the Ranluxpp state: just the 9x64 bit array. The state position will always be reset zero.) So the function should have a signature like `RngStateInitializer branch()` where `RngStateInitializer` is `RanluxppNumber` for ranluxpp and `XorwowState` (six uints) for xorwow.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -222,6 +226,47 @@ CELER_FUNCTION auto RanluxppRngEngine::operator()() -> result_type
     return bits & 0xffffffffu;
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Branch a new RNG from this RNG
```
sethrj: Minor documentation note: the first line of an extended doc block like this *requires* a trailing period to become the "brief" summary, and the single-line `//!` (or non-doxygenified summary `//` ) do *not* need a period.

Also, since user documentation only includes the class definition's docs, the block here will only show up in the developer docs. So for important technical details like this, please move to the class definition. You can add a `\par Branching` to have it render a formatted block about the capability.

src/corecel/random/engine/RanluxppRngEngine.hh
sethrj: I think it's the same thing: for state `s`, if `A` is the advance operator, then you get `(s, s') <- A (s XOR A s), A s`. The position is reset so doesn't matter, one of the states is advanced, and one of the states is the XOR of the before+after states (order doesn't matter) and then advanced.

test/corecel/random/RanluxppRngEngine.test.cc
```diff
@@ -337,6 +337,67 @@ TEST_F(RanluxppRngEngineTest, jump)
...
+    rng = RanluxppInitializer{12345, 0, 0};
+
+    // Initialize second RNG
+    RanluxppRngEngine branched_rng = rng.branch(states.ref(), TrackSlotId{1});
+
+    // Get references to the state
+    auto& ref_rng_state = states.ref().state[TrackSlotId{2}];
+    auto& ref_branched_rng_state = states.ref().state[TrackSlotId{3}];
+
+    // Create a third RNG, and branch it's state manually
```
sethrj:
```suggestion
    // Create a third RNG, and branch its state manually
```

src/corecel/random/engine/XorwowRngEngine.hh
```diff
@@ -198,6 +202,42 @@ CELER_FUNCTION void XorwowRngEngine::discard(ull_int count)
...
+
+    // Copy the state into the new state
+    for (auto i : celeritas::range(new_state.xorstate.size()))
+    {
+        new_state.xorstate[i] = state_->xorstate[i];
+    }
+    new_state.weylstate = state_->weylstate;
+
+    // Advance this RNG
+    this->discard(1);
```
sethrj: Since discarding 1 merely rotates 3 out of the 4 states, we should probably have at least 4 discard to get a new state.

sethrj (approved):
Looks good! Next step is adding a new type of "RngStateInitializer" to the track initializer data.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -83,21 +84,39 @@ class RanluxppRngEngine
...
+     *
+     * \par Branching
+     * Branching is performed in two steps.  First, the state of the new RNG
+     * (\f$x^{\prime}\f$) is initialized as
+     * \f[
+     *      x^{i,\prime}_j = x^i_j ^ x^{i+1}_j \, .
+     * \f]
+     * Second, to decorrelate the new RNG from this RNG, the new RNG is
+     * advanced forward to the next block
+     */
```
sethrj: Detailed documentation should be part of the declaration, not definition: move further down in the file.

sethrj (approved):
Excellent! Thanks @davidsgr .

sethrj:
So to recap: the ranluxpp generator is a transformation between an LCG and a "RANLUX" state, with the state advancement happening in its LCG form, the bit extraction (and xoring) happening in its RANLUX form. The ranlux state requires an extra ~9 bits as overflow compared to LCG.

One thing I don't understand (and maybe the developers who thought "branch" would work didn't either) is why the carry bits in the ranlux state are copied rather than xored or zeroed... it seems to me they should also be xored since they're part of the 'state'.

sethrj:
Reminder @davidsgr that there are a couple comments that still should be addressed before we merge.

src/corecel/random/engine/RanluxppRngEngine.hh
```diff
@@ -222,16 +241,56 @@ CELER_FUNCTION auto RanluxppRngEngine::operator()() -> result_type
...
+ * advanced forward to the next block
+ */
+CELER_FUNCTION RanluxppRngEngine::RngStateInitializer_t
+RanluxppRngEngine::branch()
+{
+    // Create a new state
+    RanluxppRngState new_state = *state_;
+
+    // Copy the current state into the new state
+    new_state.value.number = (*state_).value.number;
```
amandalund: Is this redundant with the copy above?

src/corecel/random/data/XorwowRngData.hh
```diff
@@ -82,6 +82,16 @@ struct XorwowRngInitializer
...
 
+//---------------------------------------------------------------------------//
+/*!
+ * Initializes an RNG state.
+ */
+struct XorwowRngStateInitializer
+{
+    Array<XorwowUInt, 5> xorstate;  //! x, y, z, w, v
+    XorwowUInt weylstate;
+};
```
amandalund: Or since this is equivalent to the state:
```suggestion
using XorwowRngStateInitializer = XorwowState;
```
and `state_ = state_init` in the `XorwowRngEngine::operator=`

amandalund (approved):

src/corecel/random/engine/RanluxppRngEngine.hh
sethrj: (Also minor style note, this should be written `state_->value.number`)

PR #2156, opened 2025-12-13
Cleanup build script output

esseivaju:
Minor tweaks to the build script to cleanup output if the user doesn't have an environment script.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -258,7 +258,7 @@ if ${needs_env}; then
   rc_file="$(get_shell_rc_file)"
   if install_shell_env "${rc_file}" "${ENV_SCRIPT}"; then
     needs_env=false
-  elif [ -n "${ENV_SCRIPT}" ]; then
+  else
     log warning "Please manually add the following to ${rc_file}:"
     printf ' . %s\n' "${ENV_SCRIPT}" >&2
   fi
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -305,7 +305,7 @@ else
   log error "build failed: check configuration and build errors above"
 fi
 
-if ${needs_env}; then
+if ${needs_env} && [ -n "${ENV_SCRIPT}" ]; then
```
sethrj: When would `needs_env` be true if there's no env script?

scripts/build.sh
```diff
@@ -258,7 +258,7 @@ if ${needs_env}; then
   rc_file="$(get_shell_rc_file)"
   if install_shell_env "${rc_file}" "${ENV_SCRIPT}"; then
     needs_env=false
-  else
+  elif [ -n "${ENV_SCRIPT}" ]; then
```
sethrj: This shouldn't be necessary because `needs_env` should be false. I assume you saw this because of my mistaken pre-commit vs cmake mixup?

sethrj (commented):
I think it may be just the first line that needs fixing. You might just wrap all the `needs_env` inside a `if [ -n  ${ENV_SCRIPT} ]`

scripts/build.sh
esseivaju: yep, saw that only because of the cmake/pre-commit mixup. The other lines should be fine

sethrj (approved):
Thanks!

PR #2157, opened 2025-12-14
Fix compile error with VecGeom `v2.0.0-rc.8` and Cuda

esseivaju:
`NavIndex_t` moved to a new namespace. Use type alias to pick the correct type depending the VecGeom version


sethrj (approved):
Thanks!

mrguilima (approved):
LGTM
Thanks!

sethrj (approved):
I missed these: must've only tested it with C++. Thanks!

PR #2158, opened 2025-12-15
Add Toroid surface

osanstrong:
## Add z aligned, elliptical toroid surface.

Adds the surface class Toroid, representing an elliptical toroid with ellipse radii a and b, revolved at major radius r around a z-aligned axis at an origin o:
```
       ___   _________   ___
     /  |  \           /     \
    /   b   \         /       \
   |    |    |       |         |
   |-a--+    |   o-----r--+    |
   |         |       |         |
    \       /         \       /
     \     /           \     /
       ⁻⁻⁻   ⁻⁻⁻⁻⁻⁻⁻⁻⁻   ⁻⁻⁻
```
This PR does not fully integrate the Toroid surface, only adding the Toroid class (using prior Ferrari quartic solver implementation), an accompanying test file, and required SurfaceIO details.


osanstrong:
Update: both questions resolved
~~Two immediate questions: I wrote this referencing #1295 , but #1295 used `CELER_NOT_IMPLEMENTED` for `surface_type()` to avoid fully integrating the surface all at once. I considered using this, but wasn't able to replicate the use without it throwing compile-time errors (the function `surface_type()` is elsewhere defined using `CELER_CONSTEXPR_FUNCTION` but it seems like `CELER_NOT_IMPLEMENTED` is not constexpr itself. at least from compiler error below, so I'm not sure how the initial Involute implementation used that)~~
```
[2/4] Building CXX object test/orange/CMakeFiles/orange_surf_Toroid.dir/surf/Toroid.test.cc.o
FAILED: test/orange/CMakeFiles/orange_surf_Toroid.dir/surf/Toroid.test.cc.o
/usr/bin/c++ -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -I/home/oss/ORNL/celeritas/celeritas/src -I/home/oss/ORNL/celeritas/celeritas/build/include -I/home/oss/ORNL/celeritas/celeritas/test -I/home/oss/ORNL/celeritas/celeritas/build/test -isystem /home/oss/spack/var/spack/environments/celeritas/.spack-env/view/include -g -std=c++17 -fPIE -MD -MT test/orange/CMakeFiles/orange_surf_Toroid.dir/surf/Toroid.test.cc.o -MF test/orange/CMakeFiles/orange_surf_Toroid.dir/surf/Toroid.test.cc.o.d -o test/orange/CMakeFiles/orange_surf_Toroid.dir/surf/Toroid.test.cc.o -c /home/oss/ORNL/celeritas/celeritas/test/orange/surf/Toroid.test.cc
In file included from /home/oss/ORNL/celeritas/celeritas/src/corecel/cont/detail/SpanImpl.hh:12,
                 from /home/oss/ORNL/celeritas/celeritas/src/corecel/cont/Span.hh:16,
                 from /home/oss/ORNL/celeritas/celeritas/src/corecel/cont/SpanIO.hh:12,
                 from /home/oss/ORNL/celeritas/celeritas/src/corecel/cont/ArrayIO.hh:14,
                 from /home/oss/ORNL/celeritas/celeritas/src/orange/surf/Toroid.hh:12,
                 from /home/oss/ORNL/celeritas/celeritas/src/orange/surf/Toroid.cc:7,
                 from /home/oss/ORNL/celeritas/celeritas/test/orange/surf/Toroid.test.cc:7:
/home/oss/ORNL/celeritas/celeritas/src/orange/surf/Toroid.hh: In static member function ‘static constexpr celeritas::SurfaceType celeritas::Toroid::surface_type()’:
/home/oss/ORNL/celeritas/celeritas/src/corecel/Assert.hh:186:41: error: call to non-‘constexpr’ function ‘void celeritas::throw_runtime_error(celeritas::RuntimeErrorDetails&&)’
  186 |         ::celeritas::throw_runtime_error({         \
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
  187 |             WHICH,                                 \
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  188 |             WHAT,                                  \
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  189 |             COND,                                  \
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  190 |             __FILE__,                              \
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  191 |             __LINE__,                              \
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  192 |         })
      |         ~~
/home/oss/ORNL/celeritas/celeritas/src/corecel/Assert.hh:245:5: note: in expansion of macro ‘CELER_RUNTIME_THROW’
  245 |     CELER_RUNTIME_THROW(::celeritas::RuntimeError::not_impl_err_str, WHAT, {})
      |     ^~~~~~~~~~~~~~~~~~~
/home/oss/ORNL/celeritas/celeritas/src/orange/surf/Toroid.hh:63:9: note: in expansion of macro ‘CELER_NOT_IMPLEMENTED’
   63 |         CELER_NOT_IMPLEMENTED("runtime toroid");
      |         ^~~~~~~~~~~~~~~~~~~~~
/home/oss/ORNL/celeritas/celeritas/src/corecel/Assert.hh:382:19: note: ‘void celeritas::throw_runtime_error(celeritas::RuntimeErrorDetails&&)’ declared here
  382 | [[noreturn]] void throw_runtime_error(RuntimeErrorDetails&&);
      |                   ^~~~~~~~~~~~~~~~~~~
ninja: build stopped: cannot make progress due to previous errors.
```
~~As for the second question, I'm still new to linking & header files: how are the surfaces set up so you only need to include `<name>.hh`? (E.g. `SimpleQuadric.test.cc`: `#include "orange/surf/SimpleQuadric.hh"`). When I set up my test file, the compiler couldn't find the constructor definition in `Toroid.cc` if I included `#include "orange/surf/Toroid.hh"`, and I had to instead include `#include "orange/surf/Toroid.cc"`, but this feels like a cop-out solution/I'm missing something, since the other surfaces don't do that.~~

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+CELER_FUNCTION real_type Toroid::sq(real_type val)
+{
+    return ipow<2>(val);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Shorthand for subtracting vector b from vector a
+ */
+CELER_FUNCTION Real3 Toroid::sub(Real3 const& a, Real3 const& b)
```
sethrj: This and `sq` can be CELER_CONSTEXPR_FUNCTION; you can also include `ArrayOperator.hh` (which we discourage by default because it's easy to write surprisingly inefficient expressions like `(x = a - b + c * d)`) and just use `a - b`.

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate outward facing normal at a position on or close to the surface.
+ */
+CELER_FUNCTION auto Toroid::calc_normal(Real3 const& pos) const -> Real3
+{
+    auto [x0, y0, z0] = sub(pos, origin_);
+
+    real_type d = std::sqrt(sq(x0) + sq(y0));
```
sethrj: Use `hypot` from corecel/math/Algorithms.hh
```suggestion
    real_type d = hypot(x0, y0);
```

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+                                     Real3 const& dir,
+                                     SurfaceState on_surface) const -> Real5
+{
+    auto [x0, y0, z0] = sub(pos, origin_);
+    auto [ax, ay, az] = make_unit_vector(dir);
+
+    // Intermediate terms
+    real_type p = sq(a_) / sq(b_);
+    real_type A0 = 4 * sq(r_);
+    real_type B0 = sq(r_) - sq(a_);
```
sethrj: We try to keep to the variable formatting standards even for math equations (no uppercase), and it looks like you might even be able to easily eliminate these two variables since they're only used once (q, u)

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+
+    real_type f = 1 - sq(az);
+    real_type g = f + p * sq(az);
+    real_type l = 2 * (x0 * ax + y0 * ay);
```
sethrj: `l` is also dangerous since it looks like `1` and `I` 🙃

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+    auto [x0, y0, z0] = sub(pos, origin_);
+
+    real_type val = (sq(sq(x0) + sq(y0) + sq(z0 * a_ / b_) + (sq(r_) - sq(a_)))
+                     - (4 * sq(r_)) * (sq(x0) + sq(y0)));
+    if (val < 0)
+        return SignedSense::inside;
+    else if (val > 0)
+        return SignedSense::outside;
+    else
+        return SignedSense::on;
```
sethrj: From `orange/SenseUtils.hh`, use `real_to_sense`
```suggestion
    return real_to_sense(val);
```

sethrj (commented):
This is really amazing work @osanstrong ! I have a couple of comments so far from a quick review, but I think what you've got is practically perfect.

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+ * |         |       |         |
+ *  \       /         \       /
+ *   \     /           \     /
+ *     ⁻⁻⁻   ⁻⁻⁻⁻⁻⁻⁻⁻⁻   ⁻⁻⁻
+ *
+ * This torus can be defined with the following quartic equation:
+ * \f[
+ *   (x^2 + y^2 + p*y^2 + B_0) - A_0 * (x^2 + y^2) = 0
+ * \f]
+ * where \f[p = a^2/b^2, A_0 = 4*r^2, and B_0 = (r^2-a^2)\f].
```
elliottbiondo: For inline LaTeX use `\f$ ... \f$`

src/orange/surf/Toroid.cc
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file orange/surf/Toroid.cc
+//---------------------------------------------------------------------------//
+#include "Toroid.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/**
```
elliottbiondo:
```suggestion
/*
```

src/orange/surf/Toroid.cc
```diff
@@ -0,0 +1,34 @@
...
+ * Construct toroid from origin point and radii
```
elliottbiondo:
```suggestion
 * Construct toroid from origin point and radii.
```

src/orange/surf/Toroid.cc
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * \param origin 3d origin of the toroid.
```
elliottbiondo, line 15:
```suggestion
 * \param Origin 3d origin of the toroid.
```

src/orange/surf/Toroid.cc
```diff
@@ -0,0 +1,34 @@
...
+Toroid::Toroid(Real3 const& origin,
+               real_type major_radius,
+               real_type ellipse_xy_radius,
+               real_type ellipse_z_radius)
+    : origin_(origin)
+    , r_(major_radius)
+    , a_(ellipse_xy_radius)
+    , b_(ellipse_z_radius)
+{
+    CELER_EXPECT(major_radius > 0);
```
elliottbiondo: It doesn't matter too much, but usually we validate the member data rather than the arguments

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Z-aligned Elliptical Toroid.
+ *
+ * An elliptical toroid is a shape created by revolving an axis-aligned ellipse
+ * around a central axis. This shape can be used in everything from pipe bends
+ * to tokamaks in fusion reactors. Possesses a major radius r, and ellipse
+ * radii a and b, as shown in the below diagram:
+ *     ___   _________   ___
```
elliottbiondo: Put ASCII art in a \verbatim block

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+    CELER_FUNCTION real_type major_radius() const { return r_; }
+
+    //! Radius of revolved ellipse along xy plane
+    CELER_FUNCTION real_type ellipse_xy_radius() const { return a_; }
+
+    //! Radius of revolved ellipse along z axis
+    CELER_FUNCTION real_type ellipse_z_radius() const { return b_; }
+
+    //! View of data for type-deleted storage
+    CELER_FUNCTION StorageSpan data() const { return {&origin_[0], 4}; }
```
elliottbiondo: The length should be 6: i.e., 3 for origin_ then r_, a_, b_.

@sethrj this is non-standard C++ and not actually guaranteed to work. Seems like a short function for manually serializing these would be better.

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+
+    return roots;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Calculate the coefficients of the polynomial corresponding to the given
+ * ray's intersections with the toroid.
+ *
+ * Written referencing Graphics Gems II.
```
elliottbiondo: Add the citation key `arvo_graphics-gems_1995`

test/orange/surf/Toroid.test.cc
```diff
@@ -0,0 +1,179 @@
...
+
+    {
+        // Reconstruction from data
+        SCOPED_TRACE("reconstructed");
+        Toroid recon{tor.data()};
+        check_props(tor);
+    }
+}
+
+Real3 add(Real3 const a, Real3 const b)
```
elliottbiondo: Use `ArrayOperator.hh`

elliottbiondo (dismissed):
Looks great @osanstrong! Just a couple of things.

src/orange/surf/Toroid.hh
sethrj: Let's also add this to the user docs

src/orange/surf/Toroid.hh
sethrj: I think it's legal and we use it for the other shapes, but I agree an insert mechanism like the recent work with @amandalund is better. But for now let's stick with this.

elliottbiondo: Yeah it's fine here. For posterity, I don't think it's legal in the general case; see the layout section here:

https://en.cppreference.com/w/cpp/language/data_members.html

"Alignment requirements may necessitate padding between members, or after the last member of a class."

Since these members are are all standard-sized POD, specifically doubles, the compiler is not going to add padding, but it strictly speaking doing so would be legal.

src/orange/surf/Toroid.cc
sethrj: The reason I validate data is it'll also catch the error or forgetting to initialize it :)

src/orange/surf/Toroid.hh
sethrj: Well it seems you're right; this is an old class and I had just been working in C so perhaps I misunderstood a C rule. Okeydoke, I'll add a construction utility to the to-do list 🙃

src/orange/surf/Toroid.cc
sethrj: No, `origin` is the name of the parameter, which should be lowercase

src/orange/surf/Toroid.cc
sethrj:
```suggestion
/*!
```

src/orange/surf/Toroid.hh
osanstrong: Done; I didn't see a DOI associated with the book, should I put anything else like an ISBN, or leave it as `\citep{arvo_graphics-gems_1995}`?

src/orange/surf/Toroid.hh
osanstrong: I think I get it but would you be able to elaborate on that example of an inefficient expression using `ArrayOperator.hh`? That part I didn't quite get. Do you mean like someone would write a variable `x = a - b + c * d`, but only ever use it as `y = x - c * d`?

src/orange/surf/Toroid.hh
osanstrong: Is adding to user docs something in scope for this pr, or would we make a separate issue for that?

(unless it's already been put in there between then and now; I didn't see it on the website but I haven't exactly done the deepest of docs dives so I might have missed it.)

src/orange/surf/Toroid.hh
osanstrong: Good catch! I think I missed that one because the loading/unloading both used 6 members so it technically still worked.

So are we sticking with the same raw data constructor as the other shapes for now?

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,271 @@
...
+    CELER_FUNCTION real_type major_radius() const { return r_; }
+
+    //! Radius of revolved ellipse along xy plane
+    CELER_FUNCTION real_type ellipse_xy_radius() const { return a_; }
+
+    //! Radius of revolved ellipse along z axis
+    CELER_FUNCTION real_type ellipse_z_radius() const { return b_; }
+
+    //! View of data for type-deleted storage
+    CELER_FUNCTION StorageSpan data() const { return {&origin_[0], 6}; }
```
pcanal:
```suggestion
#include <cstddef>
....
    CELER_FUNCTION StorageSpan data() const {
       static_assert((offsetof(Toroid, b_) - offsetof(Toroid, origin_)) == sizeof(float)*5);
       return {&origin_[0], 6};
    }
```
If you want to be on the safe side.
See a [live](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:61,endLineNumber:9,positionColumn:61,positionLineNumber:9,selectionStartColumn:61,selectionStartLineNumber:9,startColumn:61,startLineNumber:9),source:'%23include+%3Ccstddef%3E%0A%0Astruct+Example+%7B%0A++++float+a%3B%0A++++float+b%3B%0A++++float+c%3B%0A++++float+*function()+%7B%0A++++++++offsetof(Example,+c)%3B%0A++++++++static_assert((offsetof(Example,+c)-offsetof(Example,+a))+%3D%3D+sizeof(float)*2)%3B%0A++++++++return+%26a%3B%0A++++%7D%0A%7D%3B%0A%0A%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:53.80150899593732,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:output,i:(compilerName:'x86-64+gcc+13.3',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+19.1.0+(assertions)+(Compiler+%231)',t:'0')),k:46.19849100406268,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang191assert,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:fmt,ver:'1200')),options:'-std%3Dc%2B%2B20+-O3+-Wshadow+-Wall',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+19.1.0+(assertions)+(Editor+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:46.19849100406268,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) example

osanstrong: Thanks for the example!

src/orange/surf/Toroid.hh
elliottbiondo: I think:

```
\citep{arvo_graphics-gems_1995, https://shop.elsevier.com/books/graphics-gems-ii/arvo/978-0-08-050754-5}
```

would be good. That's the official site for the book and the URL includes the ISBN

src/orange/surf/Toroid.hh
elliottbiondo: Yep that's fine

src/orange/surf/Toroid.hh
elliottbiondo: @sethrj in the user doc, doxygen pulls the class-level doc strings for each `IntersectRegion`. This is the doc string for the surface. I think he should hold on until he creates an `IntersectRegion`.

src/orange/surf/Toroid.hh
elliottbiondo: I believe what @sethrj is saying that if you have `N` vectors:

```
v1, v2, ... vN
```

and you want to calculate

```
x = v1 + v2 + ... + vN
```

using `ArrayOperators.hh`, under the hood there will be `N` for loops executed in serial (if you look at the code in `ArrayOperators.hh` this will be apparent). The first for loop will add `v1` and `v2`, then the second for loop will add `v3`, etc. Of course, operation can be done in single for loop:

```
for i : [1, N]
    x += v1[i] + v2[i] + ... _ vN[i]
```

Thus, be careful when chaining array `operator` expressions.

src/orange/surf/Toroid.hh
sethrj: Yeah, @elliottbiondo and I talked about this; what I did for the ORANGE surfaces was an early version of the GPU-compatible variant stuff; but I didn't realize that this way of generically exporting the data was illegal.

sethrj: Also I don't think that's much safer since it depends on the ordering of the variables; you'd be better off checking `sizeof(*this) == 6 * sizeof(real_type)` and `offset_of(Toroid, origin_) == 0`. But we should skip it since we're not even doing this for the other surfaces and to do it correctly we would rewrite it.

src/orange/surf/Toroid.hh
sethrj: I updated this (moved it to the class docs where it's visible from the user docs) to have the correct citation key to match our `doc/_static/zotero.bib`; it would be nice if there were an ISBN equivalent of doi.org but I think also the default `cite` is just as good (the URL is only for convenience when browsing the doxygen or in a code browser).

src/orange/surf/Toroid.hh
sethrj: Didn't see this was the extension of a previous comment https://github.com/celeritas-project/celeritas/pull/2158#discussion_r2620409435 🙃

src/orange/surf/Toroid.hh
sethrj: I thought the ORANGE docs also included surfaces but apparently I'm wrong! Let's defer then.

src/orange/surf/Toroid.hh
sethrj: Worse than that, it's technically creating a bunch of intermediate results, so unless the compiler is really smart, you'll increase the register usage of the statement by a factor of N :)

src/orange/surf/Toroid.hh
osanstrong: So nix the safety check for now, since this data export method needs to be rethought anyways?

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+    real_type b_;
+
+    //// HELPER FUNCTIONS ////
+
+    // Calculate the coefficients of the polynomial for ray intersection
+    inline CELER_FUNCTION Real5 calc_intersection_polynomial(
+        Real3 const& pos, Real3 const& dir, SurfaceState on_surface) const;
+
+    // Shorthand to square a number
+    static CELER_CONSTEXPR_FUNCTION real_type sq(real_type val);
```
elliottbiondo, line 145: Normally I would put something like this in an anonymous namespace, but I think it's fine here.

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,276 @@
...
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Z-aligned Elliptical Toroid.
+ *
+ * An elliptical toroid is a shape created by revolving an axis-aligned ellipse
+ * around a central axis. This shape can be used in everything from pipe bends
+ * to tokamaks in fusion reactors. Possesses a major radius \em r, and ellipse
```
elliottbiondo:
```suggestion
 * to tokamaks in fusion reactors. It possesses a major radius \em r, and ellipse
```

elliottbiondo (approved):
I did a final review. Very polished! One minor fix, otherwise it should be good to go, unless @sethrj wants `sq` in an anonymous namespace

src/orange/surf/Toroid.hh
sethrj: yeah since this is a header file we can't do that. I think this acceptable; in the future if everyone likes it we can move to the `math.hh`.

elliottbiondo: Oh good call about the header

sethrj:
Thanks for getting this through @osanstrong !! Great work.

src/orange/surf/Toroid.hh
```diff
@@ -0,0 +1,275 @@
...
+    -> Intersections
+{
+    Real5 abcde = calc_intersection_polynomial(pos, dir, on_surface);
+    FerrariSolver solve{};  // Default tolerance
+    Intersections roots;
+
+    if (on_surface == SurfaceState::on)
+    {
+        auto [a, b, c, d, e] = abcde;
+        roots = solve(Real4{a, b, c, d});
```
sethrj, line 205: It's too late for this PR but I just checked the [coverage report](https://app.codecov.io/gh/celeritas-project/celeritas/pull/2158?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=celeritas-project) and this isn't called in the tests: even though it's mostly redundant with the solver testing, a small test to make sure (e.g.) that we're not discarding the wrong polynomial value would be a good idea. You can go ahead and add that as part of the next capability update for this.

PR #2159, opened 2025-12-15
Support different CODATA measured constants with `CELERITAS_CODATA` config option

sethrj:
A new cmake option allows us to choose between the 2018 value (existing, default) of the CODATA experimental constants, the new 2022 values, or the 2006 (used by CLHEP, see https://github.com/celeritas-project/celeritas/pull/1309#discussion_r1672818600) values. An `inline` namespace allows us to just use `celeritas::constants::foo` in the regular case but specifically choose a sub-namespace if a certain class wants to use it for reproducibility.

This also fixes the documentation for units and constants: a weird doxygen bug meant that omitting the `INCLUDE_PATH` prevented preprocessing from affecting the content of the files, leading to the electron constant showing up as 1 (for CLHEP) rather than `1.602176634e-19 C`.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -37,13 +37,12 @@ jobs:
         image: ["rocky-cuda", "ubuntu-rocm"]
         exclude:
           - geometry: "vecgeom"
+            buildtype: "debug" # Build is too large
+          - geometry: "vecgeom"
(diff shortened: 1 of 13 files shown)


pcanal (approved):
LGTM

amandalund (approved):
Thanks @sethrj!

PR #2160, opened 2025-12-15
Use local energy deposition to determine if scintillation photons should be sampled at collision site

amandalund:
Following on to #2152, this updates the scintillation photon position sampling to use the fraction of energy deposited locally at the collision site to determine whether to generate photons at the endpoint rather than basing it off whether the particle is neutral.


src/celeritas/optical/gen/ScintillationGenerator.hh
```diff
@@ -210,8 +209,21 @@ CELER_FUNCTION TrackInitializer ScintillationGenerator::operator()(Generator& rn
...
-    // Sample position: endpoint (collision site) if neutral, else uniform
-    real_type u = is_neutral_ ? 1 : UniformRealDist{}(rng);
+    // Sample the position
+    real_type u = [&rng, &p = dist_.continuous_edep_fraction] {
+        // The number of photons generated along the step (continuous energy
+        // loss) and at the interaction site (local energy deposition) is
+        // proportional to their respective energy contributions. If both
+        // components are present, sample where to generate using the fraction
+        // of the energy deposited along the step.
+        if (p == 1 || (p != 0 && BernoulliDistribution{p}(rng)))
```
sethrj, line 224: Is this merely an optimization of
```suggestion
        if (BernoulliDistribution{p}(rng)))
```
or would there be an error if we did it that way?

sethrj (approved):
Nice, thanks! Certainly more detail/accuracy than the competitors 🙃

src/celeritas/optical/gen/ScintillationGenerator.hh
amandalund: Yep, it  just saves some rng samples in the usual case where p is 0 or 1 and keeps the test results from changing. And I should document that :)

PR #2161, opened 2025-12-16
Add an option to kill optical tracks after a specified number of steps

amandalund:
This adds a `max_steps` option for the optical stepping loop. I've added an optical `SimParams` and moved the `max_step_iters` there as well.  `max_step_iters` is now required to be greater than zero (this was previously allowed solely for testing purposes in the `OpticalLaunchAction`).

Also tagging #2150 since this is about to request reviews from 9 people.


sethrj (approved):
Nice, thanks @amandalund ! I think we can restrict pull the CODEOWNERS to "executor" type stuff (i.e. changes to the physics), and/or remove the long list of people requested for `accel`...

amandalund:
Agreed, I think narrowing down the `accel` list or making it a bit more fine-grained in terms of who owns what could help.

PR #2167, opened 2025-12-17
Simplify codeowners again

sethrj:
This removes the list of people who all get asked to review anytime something in @accel changes. Now only interactors ping responsible folks; other classes likely to be frequently refactored are less likely to touch a code owner.


pcanal (approved):
Fair enough.

amandalund (approved):
Should help, thanks!

PR #2169, opened 2025-12-18
Remove branching in CylMapField evaluation

esseivaju:
This is just a small tweak to reduce divergence in `CylMapField`

### DIFF
--- a/src/celeritas/field/CylMapField.hh
+++ b/src/celeritas/field/CylMapField.hh
@@ -90,18 +90,9 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
 
     // Convert Cartesian to cylindrical coordinates
     real_type r = hypot(pos[0], pos[1]);
-    // Ensure phi is in [0, 2\f$\pi\f$)
+    // Wrap turn from [-1/2; 1/2] into [0, 1)
     auto phi = atan2turn<real_type>(pos[1], pos[0]);
-    if (phi < zero_quantity())
-    {
-        phi.value() += 1;
-    }
-    if (CELER_UNLIKELY(phi.value() == 1))
-    {
-        // Make sure phi is in [0, 1). If phi is a negative value smaller
-        // than machine epsilon, adding 1 will result in phi equal to 1
-        phi.value() = 0;
-    }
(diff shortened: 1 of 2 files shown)


src/celeritas/field/CylMapField.hh
```diff
@@ -90,18 +90,9 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
...
-    {
-        phi.value() += 1;
-    }
-    if (CELER_UNLIKELY(phi.value() == 1))
-    {
-        // Make sure phi is in [0, 1). If phi is a negative value smaller
-        // than machine epsilon, adding 1 will result in phi equal to 1
-        phi.value() = 0;
-    }
+    phi.value() -= floor(phi.value());
```
sethrj, line 95: Won't this likely have just as many branches of the above? Have you compared the ptx output?

src/celeritas/field/CylMapField.hh
```diff
@@ -143,15 +134,9 @@ CELER_FUNCTION auto CylMapField::operator()(Real3 const& pos) const -> Real3
...
-    real_type sin_phi = 0;
-
-    if (r != 0)
-    {
-        cos_phi = pos[0] / r;
-        sin_phi = pos[1] / r;
-    }
+    real_type sin_phi;
+    real_type cos_phi;
+    sincos(phi, &sin_phi, &cos_phi);
```
sethrj, line 139: Wait, was this wrong before? Did we forget to call `sincos`?? 😅

esseivaju: I think it was correct, we were doing the manual calculation $cos \phi=x/r$

src/celeritas/field/CylMapField.hh
esseivaju: I didn't check I can have a look tomorrow. I'd hope the intrinsic can do this more efficiently

esseivaju: PTX does seem to have [instructions for rounding](https://docs.nvidia.com/cuda/parallel-thread-execution/#rounding-modifiers), `cvt.rm` / `cvt.rmi`

sethrj (approved):
Thanks!

PR #2170, opened 2025-12-19
Add an initial support for electro-nuclear process and model 

whokion:
This MR adds electro-nuclear cross sections using G4ElectroNuclearCrossSection, along with an initial skeleton to support the electro-nuclear process and model. The interactor will be implemented in a follow-up MR.


whokion:
@amandalund, @sethrj This MR is relatively large, but it closely mirrors the previous gamma-nuclear MR and consists of physics boilerplate codes. Follow-up MRs will add the details of the interactor along with additional helper codes, and onloading pipeline for Geant4 post step actions. I would also like to keep the cross-section codes as they are for now to estabilish a baseline for earlier test and validation, but plan to consolidate or simplify them as much as possible later. Thanks for your review, and sorry for another big MR.

src/celeritas/em/interactor/ElectroNuclearInteractor.hh
```diff
@@ -0,0 +1,64 @@
...
+    inline CELER_FUNCTION
+    ElectroNuclearInteractor(NativeCRef<ElectroNuclearData> const& shared,
+                             ParticleTrackView const& particle);
+
+    // Sample an interaction
+    inline CELER_FUNCTION Interaction operator()();
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
```
stognini, line 41:
```suggestion
// INLINE DEFINITIONS
//---------------------------------------------------------------------------//
```

src/celeritas/em/model/ElectroNuclearModel.cc
```diff
@@ -0,0 +1,157 @@
...
+    // Save IDs
+    data.scalars.electron_id = particles.find(pdg::electron());
+    data.scalars.positron_id = particles.find(pdg::positron());
+
+    CELER_VALIDATE(data.scalars.electron_id && data.scalars.positron_id,
+                   << "missing particles (required for " << this->description()
+                   << ")");
+
+    // Save particle properties
+    CELER_EXPECT(data.scalars);
```
stognini: I'd drop this, since it is doing the same thing as the validate.

amandalund:
Thanks @whokion for this addition! Yes, the size should be mostly due to the physics boilerplate, so no worries :). I am off today but will take a look before the holidays!

whokion:
Thanks @amandalund ! No problem and enjoy your timeoff!

src/celeritas/em/interactor/ElectroNuclearInteractor.hh
whokion: Yes, this code is still imcomplete.  Anyway, I added the line - also in `GammaNuclearInteractor`.

src/celeritas/em/model/ElectroNuclearModel.cc
whokion: Remove the check - also in `GammaNuclearModel.cc`.

src/celeritas/em/model/ElectroNuclearModel.cc
```diff
@@ -0,0 +1,154 @@
...
+    CELER_EXPECT(z);
+
+    inp::Grid result;
+
+    // Tabulate cross sections using separate parameterizations for the high
+    // energy region (emin < E < 50 GeV) and the ultra high energy region up
+    // to the maximum valid energy (emax). The numbers of bins are chosen to
+    // adequately capture both parameterized points (336 bins from 2.0612 MeV
+    // to 50 GeV) and calculations used in G4ElectroNuclearCrossSection, which
+    // can be made configurable if needed (TODO).
```
stognini: Feel like this could be moved to the top documentation.

src/celeritas/em/executor/ElectroNuclearExecutor.hh
```diff
@@ -0,0 +1,61 @@
...
+ */
+CELER_FUNCTION Interaction
+ElectroNuclearExecutor::operator()(CoreTrackView const& track)
+{
+    auto particle = track.particle();
+
+    // Select a target element
+    auto material = track.material().material_record();
+    auto elcomp_id = track.physics_step().element();
+    if (!elcomp_id)
```
stognini, line 41: So sometimes the element is already selected, and sometimes you need to select the element on the fly? Wouldn't it be always on the fly?

amandalund: This is so in the unlikely case we need to redo the interaction (e.g. we ran out of secondary memory) we don't alter the rng stream.

stognini: Ohh I see. Thanks!

sethrj: This was thought to be an important issue at the beginning of the project... but what I would do now would be to have a single 2x-state-sized buffer for post-interaction secondaries+state changes, plus a "generator" that would allow things like atomic cascades.

src/celeritas/phys/PhysicsTrackView.hh
```diff
@@ -458,11 +459,17 @@ CELER_FUNCTION real_type PhysicsTrackView::calc_xs(ParticleProcessId ppid,
...
         else if (model_id == params_.hardwired.ids.eplusgg)
         {
             result = EPlusGGMacroXsCalculator(params_.hardwired.eplusgg,
                                               material)(energy);
         }
+        else if (model_id == params_.hardwired.ids.electro_vd)
+        {
+            result = MacroXsCalculator<ElectroNuclearMicroXsCalculator>(
+                params_.hardwired.electro_vd, material)(energy);
+        }
```
sethrj, line 472: This and many of the other models are just tabulated nonuniform cross sections; we should be able to significantly simplify the code someday.

sethrj (commented):
Looks great @whokion !

src/celeritas/em/process/ElectroNuclearProcess.hh
```diff
@@ -0,0 +1,62 @@
...
+    VecModel build_models(ActionIdIter start_id) const final;
+
+    // Get the interaction cross sections for the given energy range
+    XsGrid macro_xs(Applicability range) const final;
+
+    // Get the energy loss for the given energy range
+    EnergyLossGrid energy_loss(Applicability range) const final;
+
+    //! Whether the integral method can be used to sample interaction length
+    bool supports_integral_xs() const final { return false; }
```
amandalund, line 48: This is false even though the incident particle is an electron or positron?

src/celeritas/em/xs/ElectroNuclearMicroXsCalculator.hh
```diff
@@ -0,0 +1,83 @@
...
+    inline CELER_FUNCTION
+    ElectroNuclearMicroXsCalculator(ParamsRef const& shared, Energy energy);
+
+    // Compute cross section
+    inline CELER_FUNCTION BarnXs operator()(ElementId el_id) const;
+
+  private:
+    // Shared cross section data
+    ParamsRef const& data_;
+    // Incident photon energy
```
amandalund:
```suggestion
    // Incident particle energy
```

src/celeritas/em/executor/ElectroNuclearExecutor.hh
amandalund: This also shouldn't be necessary currently since we're not allocating any secondaries in the interactor, but I assume the follow-on PR will?

src/celeritas/em/interactor/ElectroNuclearInteractor.hh
```diff
@@ -0,0 +1,65 @@
...
+                 || particle.particle_id() == shared.scalars.positron_id);
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Onload the electro-nuclear interaction.
+ */
+CELER_FUNCTION Interaction ElectroNuclearInteractor::operator()()
+{
+    return Interaction::from_onloaded();
```
amandalund, line 61: Will we have a way to distinguish this type of onloading (where the incident particle continues to be tracked but the secondary is onloaded) from gamma-nuclear (where the incident particle is onloaded)?

amandalund (approved):
Thanks @whokion!

src/celeritas/em/model/ElectroNuclearModel.cc
whokion: Move cross-section tabulation documentation to the method description.

src/celeritas/em/process/ElectroNuclearProcess.hh
whokion: This is a good question, but it would be better to set it to false. ElectroNuclearProcess is technically a hadronic inelastic process in Geant4, not a G4VEMProcess. Even though it is a energy loss process, the integral treatment may not be adequate, especially in energy regions with many resonances. To remain consistent with Geant4 and to take a conservative approach, I would set it to false.

src/celeritas/em/xs/ElectroNuclearMicroXsCalculator.hh
whokion: Fixed.

src/celeritas/em/interactor/ElectroNuclearInteractor.hh
whokion: Yes, I agree. This will depend on how we handle onloaded particles, and a proper status flag should be introduced as needed in one of the follow-up MRs.

whokion:
@sethrj I guess that I have addressed all review comments for now and will follow up on the remaining related suggestions in subsequent MRs. Thanks again for your review and for considering merging this at this stage.

PR #2171, opened 2025-12-19
Fix ExCL env script

elliottbiondo:
This MR fixes a small bug that prevented environment scripts from loading the correct, node-specific spack environment.


elliottbiondo:
@sethrj small one RTR

sethrj (approved):

PR #2172, opened 2025-12-20
Add a `celer-config` app and milan2 config

sethrj:
This adds a small `celer-config` app for printing out the system configuration with only a corecel dependency (as opposed to celer-sim which needs all of celeritas). The config output is printed when building with the build.sh script if there is a test failure.


elliottbiondo (approved):
I am surprised you are adding an env for the V100 nodes?

amandalund (approved):

sethrj:
@elliottbiondo The ExCL admins said we might want to use mostly that node for our basic build/test, and it'll still be nice to have an alternative to wildstyle.

PR #2173, opened 2025-12-22
Refactor along-step energy loss applier

sethrj:
I'm starting to look at #538 and one of the first steps is simplifying some of the code that's there. This reduces some of the code duplication especially about tracking cuts. The goal is to make an `EnergyLossTrackExecutor` struct that is templated on an energy loss calculator which contains GPU data.


src/celeritas/alongstep/detail/FluctELoss.hh
```diff
@@ -96,81 +78,65 @@ CELER_FUNCTION bool FluctELoss::is_applicable(CoreTrackView const& track) const
...
-        && (particle.energy() - eloss <= phys.particle_scalars().lowest_energy))
+    if (eloss >= particle.energy())
     {
-        // Deposit all energy when we end below the tracking cut
-        return particle.energy();
+        // Sampled energy loss can be greater than actual remaining energy
+        // because the range calculation is based on the *mean* energy
+        // loss. To fix this, we would need to sample the range from a
+        // distribution as well.
+        if (CELER_UNLIKELY(track.geometry().is_on_boundary()))
```
amandalund: Is it unlikely for the track to be on a boundary?

amandalund (approved):
Looks good, thanks @sethrj!

src/celeritas/alongstep/detail/FluctELoss.hh
sethrj: If we fix the range calculation (integrating de/dx down to the tracking cut rather than zero), and if we don't have energy fluctuations, then it should be basically impossible. It should still be unusual, but I don't have any idea how "unikely" it is. I'll remove the decorator.

Issue #1508, opened 2024-11-15
Add muon-catalyzed fusion physics


stognini:
# Physical processes

## Necessary for a proof-of-principle

- Muonic atom capture (G4/NK)
  - Formation of $d\_\mu$ and $t\_\mu$
- Muonic molecule formation (G4/NK)
  - Formation of $(dd)\_\mu$ , $(dt)\_\mu$ , or $(tt)\_\mu$
- MuCF (NK)
  - Fusion of of $(dd)\_\mu$ , $(dt)\_\mu$ , and $(tt)\_\mu$
- Muon decay (G4)
- Muon basic EM physics
- Neutron, alpha, proton, $^3\text{He}$, and $^4\text{H}$ transport

## Good to have

- Muonic atom decay (G4)
- Isotopic transfer (NK)
  - Muon tends to be transferred to higher Z atoms
- Muon atom stripping (NK)
- Alpha deexcitation cascade
- Spin flip
- Muon, alpha, neutron (G4)
  - Energy-loss
  - Ionization
  - Scattering

# TODO

- [x] Verify changes in the [NK-Labs Geant4](https://github.com/stognini/nklabs-geant4)
  - [x] Test changes on Geant4 v11+
- [x] Add NK Labs physics list to `utils/celer-geant`
- [ ] Add muon decay to Celeritas' stepper ([PR 1456](https://github.com/celeritas-project/celeritas/pull/1456) and [PR 1991](https://github.com/celeritas-project/celeritas/pull/1991))
- Add muCF in Celeritas
  - [x] Expand `SetupOptions` to include muCF
  - [x] Expand `inp::` data to include muCF physics ([PR 2125](https://github.com/celeritas-project/celeritas/pull/2125))
  - [x] Expand `GeantImporter` to initialize muCF physics ([PR 2125](https://github.com/celeritas-project/celeritas/pull/2125)) 
  - [ ] Expand `ProcessBuilder` to include muCF process
  - [ ] Add neutron, even if it does not have physics attached to it
  - [ ] Address muonic atom and its decay in Celeritas
  - [ ] Add at-rest competition between processes
  - [ ] Add process/model class
    - [x] Draft full skeleton as an at-rest execution and define host/device data struct ([PR 2211](https://github.com/celeritas-project/celeritas/pull/2211))
    - [ ] Add tabulated data in `inp::` and construct device data ([PR 2240](https://github.com/celeritas-project/celeritas/pull/2240))
    - MuCF interactors
[...]

PR #2174, opened 2025-12-29
Add muon-catalyzed fusion process/model skeletons

stognini:
This PR adds the basic structure for the muon-catalyzed fusion process. See issue [#1508](https://github.com/celeritas-project/celeritas/issues/1508) for details.

 It is an at rest process, and it applies only to d-t mixtures. The model `executor` is responsible for taking a muon at rest and:
- Create the muonic atom
- Apply any secondary effect (e.g. isotopic transfer, where the muon transfers to a heavier atom)
- Create the muonic molecule
- Verify if fusion happens before the muon decays
- Call appropriate interactor (decay or fusion)


src/celeritas/inp/MucfPhysicsData.hh
```diff
@@ -0,0 +1,117 @@
...
+{
+namespace inp
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Static muon energy CDF data for muon-catalyzed fusion.
+ *
+ * \todo This data may be loaded from an external file in the future.
+ */
+static Grid mucf_muon_energy_cdf()
```
sethrj: These should probably be inlined directly into `MucfPhysics::from_default`. At any rate, they should live inside that cc file.

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+
+    //! Check whether all particles are assigned
+    CELER_FUNCTION explicit operator bool() const
+    {
+        return mu_minus && neutron && proton && alpha && helium3
+               && muonic_hydrogen && muonic_alpha && muonic_triton
+               && muonic_helium3;
+    }
+
+    //! Assign from ParticleParams (anonymous free function in the model.cc?)
```
sethrj: Correct, this should be in the model.cc file.

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+    static MucfParticles from_params(ParticleParams const& particles)
+    {
+#define MP_ADD(MEMBER)                               \
+    pids.MEMBER = particles.find(pdg::MEMBER());     \
+    if (!pids.MEMBER)                                \
+    {                                                \
+        missing.push_back({#MEMBER, pdg::MEMBER()}); \
+    }
```
sethrj: Since this doesn't require any code-to-text macro preprocessing (which similar JSON code does, `"X" = X`) I think it would improve readability to make this a lambda function.

```suggestion
```

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+
+        using PairStrPdg = std::pair<std::string, PDGNumber>;
+        std::vector<PairStrPdg> missing;
+        MucfParticles pids;
+
+        MP_ADD(mu_minus);
+        MP_ADD(neutron);
+        MP_ADD(proton);
+        MP_ADD(alpha);
```
sethrj:
```suggestion


   auto set_pid = [&](ParticleId& pid, PdgNumber num) {
...
    };
        set_pid(pids.mu_minus, pdg::pids.mu_minus());
...
```

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+        MP_ADD(muonic_triton);
+        MP_ADD(muonic_helium3);
+#endif
+        CELER_VALIDATE(
+            missing.empty(),
+            << "missing particles required for muon-catalyzed fusion: "
+            << join(missing.begin(), missing.end(), ", ", [](PairStrPdg const& p) {
+                   return p.first + " (PDG "
+                          + std::to_string(p.second.unchecked_get()) + ')';
+               }));
```
sethrj:
```suggestion
            << join_stream(missing.begin(), missing.end(), ", ", [](std::ostream& os, PairStrPdg const& p) {
                   os << p.first << " (PDG " << p.second << ')';
               }));
```
(include `corecel/OpaqueIdIO.hh`)

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+template<Ownership W, MemSpace M>
+struct DTMixMucfData
+{
+    template<class T>
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using MatCycleItems = Collection<T, W, M, MuCfMatCompId>;
+    template<class T>
+    using MaterialItems = Collection<T, W, M, MuCfMatCompId>;
+    using Grid = NonuniformGridRecord;
```
sethrj: Since the grid and the grid record are different concepts maybe:
```suggestion
    using GridRecord = NonuniformGridRecord;
```

Same thing with ParticleIds vs Particles. I think it's still OK to keep the names as "grid" and "particle" but the type name should reflect the type.

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,156 @@
...
+ */
+template<Ownership W, MemSpace M>
+struct DTMixMucfData
+{
+    template<class T>
+    using Items = Collection<T, W, M>;
+    template<class T>
+    using MatCycleItems = Collection<T, W, M, MuCfMatCompId>;
+    template<class T>
+    using MaterialItems = Collection<T, W, M, MuCfMatCompId>;
```
sethrj: These are the same type, but it looks like one is mat*molecule*spin whereas one is mat? They should have different IDs if the IDs have different semantics.

src/celeritas/mucf/interactor/DDMucfInteractor.hh
```diff
@@ -0,0 +1,152 @@
...
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct with shared and state data.
+ */
+DDMucfInteractor::DDMucfInteractor(NativeCRef<DTMixMucfData> const& data,
+                                   Channel const channel,
```
sethrj, line 79:
```suggestion
                                   Channel channel,
```

src/celeritas/mucf/interactor/DDMucfInteractor.hh
```diff
@@ -0,0 +1,152 @@
...
+    result.secondaries = this->sample_secondaries(secondaries, rng);
+
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Return number of secondaries from each fusion channel.
+ */
+CELER_FUNCTION size_type DDMucfInteractor::num_secondaries() const
```
sethrj: This should be a static enum array that lives in the mucf data.

src/celeritas/Types.hh
```diff
@@ -237,7 +243,10 @@ enum class MucfMuonicAtom
 };
 
 //---------------------------------------------------------------------------//
-//! Muon-catalyzed fusion molecules
+/*!
+ * Muonic molecule selection from material data. This is *not* intended to be
+ * used by the transport loop.
+ */
```
sethrj: I thought we decided to move all these mucf definitions to a different file (e.g. celeritas/mucf/Types.hh) ? They're very specialized.

src/celeritas/phys/PDGNumber.hh
```diff
@@ -146,6 +146,10 @@ CELER_DEFINE_PDGNUMBER(anti_he3, -1000020030)
 CELER_DEFINE_PDGNUMBER(alpha, 1000020040)
 CELER_DEFINE_PDGNUMBER(anti_alpha, -1000020040)
 
+// Muonic nuclei
```
sethrj, line 149: Are these numbers standardized or in a "reserved" space? Are they specific to Celeritas or used in Geant4 as well?

src/celeritas/mucf/model/DTMixMucfModel.hh
```diff
@@ -0,0 +1,75 @@
...
+    // Construct from model ID and other necessary data
+    DTMixMucfModel(ActionId id,
+                   ParticleParams const& particles,
+                   MaterialParams const& materials);
+
+    // Particle types and energy ranges that this model applies to
+    SetApplicability applicability() const final;
+
+    // Get the microscopic cross sections for the given particle and material
+    XsTable micro_xs(Applicability) const final;
```
sethrj, line 54: Does this PR depend on #1991? We don't have a proper "at rest" right now.

src/celeritas/mucf/process/MucfProcess.cc
```diff
@@ -0,0 +1,66 @@
...
+/*!
+ * Construct from host data.
+ */
+MucfProcess::MucfProcess(SPConstParticles particles, SPConstMaterials materials)
+    : particles_(particles), materials_(materials)
+{
+    //! \todo Fix ImportProcessClass
+    CELER_EXPECT(particles_);
+    CELER_EXPECT(materials_);
+    CELER_EXPECT(particles_->find(pdg::mu_minus()));
```
sethrj: This will change a runtime error into a debug error (it will be hit before the later validate on the particle type) so I would skip this.

sethrj (changes requested):
I appreciate the detail classes for the context of long term planning, but I think they should be deleted since they're not used and not tested: having them there is an easy way to introduce zombie code into the codebase.

src/celeritas/mucf/data/DTMixMucfData.hh
stognini: Wait, but without the macro I lose the ability to store the particle name as a string in the lambda function (we don't have a `to_string` to convert our PDGs to their names to populate the `first` element of the `PairStrPdg`). Am I missing something? 

We can always just print out the PDG, but aside from the basic elementary particles, the PDG numbering scheme gets very cumbersome very fast (that is why heavier particles, like a deuteron, have wildly large PDG numbers).

src/celeritas/mucf/data/DTMixMucfData.hh
stognini: Good catch. I did not clean this up while deciding how to organize the data. I'll keep `MaterialItems` for both, since the main index is still the material component: `matcompid_to_matid` allows us to know which component is tied to which `MaterialId`, so from the material id we retrieve all cached material information (for now only cycle time, but this will be equivalent for the other time data, like spin flip).

src/celeritas/phys/PDGNumber.hh
stognini: These are standard PDG codes. In G4, they are calculated on the fly for specific cases. E.g. when a muonic atom is created (whenever there is sticking, like a muonic alpha after the fusion happens), the secondary's `G4ParticleDefinition` comes from `G4IonTable::GetMuonicAtom(int Z, int A)`.

src/celeritas/mucf/model/DTMixMucfModel.hh
stognini: It does. We need to select which at rest process happens (decay or mucf) during the stepping loop.
Having said that, since a lot happens in the mucf executor, _if_ mucf is selected, we may still return a decay.

See comments in the executor:
```cpp
52:    // Muon decay may compete against other "actions" in this executor

95:   // Check if muon decays before fusion happens
```

src/celeritas/mucf/interactor/DDMucfInteractor.hh
stognini: I pondered about it when drafting the data, and since each molecule (dd, dt, tt) has its own set of channels, keeping the channel list as part of the interactor seemed easier to maintain and follow.

So in your suggestion, do you think we should have hardcoded in the `DTMixMucfData.hh` all channel enums (such as an `enum DDFusionChannel` and so on), and one `EnumArray` for each (e.g. `EnumArray<DDFusionChannel, size_type> dd_num_secondaries;`) in the `struct DTMixMucfData`?

src/celeritas/Types.hh
stognini: Agreed. Moved everything to a new `celeritas/mucf/Types.hh`.

src/celeritas/mucf/data/DTMixMucfData.hh
sethrj: Oh you're right, I did miss that. The macro makes more sense now :)

sethrj: PS the two previous discussions are resolved; but you should still use `join_stream` to simplify the `join` operation.

stognini:
> I appreciate the detail classes for the context of long term planning, but I think they should be deleted since they're not used and not tested

Agreed that they are not tested yet, but they are actively being called in the executor. Do you prefer me to make a todo in the executor and remove them until a future PR?

sethrj:
> Agreed that they are not tested yet, but they are actively being called in the executor. Do you prefer me to make a todo in the executor and remove them until a future PR?

Oh I didn't catch that. I'll take another look at your comments later this afternoon.

src/celeritas/inp/MucfPhysics.cc
```diff
@@ -6,12 +6,107 @@
...
+/*!
+ * Static cycle rate data for muon-catalyzed fusion.
+ *
+ * \todo This data may be loaded from an external file in the future.
+ */
+static std::vector<MucfCycleRate> mucf_cycle_rates()
+{
+    //! \todo Implement
+    return {};
+};
```
sethrj: Unwanted semicolon; also please put all these in an anonymous namespace rather than indivdual `static` (which is the C way of doing the equivalent)

src/celeritas/inp/MucfPhysics.cc
```diff
@@ -10,6 +10,103 @@ namespace celeritas
...
+             0.6866449511400652,
+             0.7309446254071662,
+             0.7778501628664496,
+             0.8104234527687297,
+             0.8403908794788275,
+             0.8618892508143323,
+             0.8814332247557004,
+             0.8970684039087949,
+             0.903583061889251,
+             1.0};
```
sethrj: Add trailng comma for these long indented lists: 
```suggestion
             1.0,
             };
```

src/celeritas/inp/MucfPhysics.cc
```diff
@@ -10,6 +10,103 @@ namespace celeritas
...
+{
+    //! \todo Implement
+    return {};
+};
+
+//---------------------------------------------------------------------------//
+/*!
+ * Static spin-flip data for muon-catalyzed fusion.
+ *
+ * \todo This data may be loaded from an external file in the future.
```
sethrj: You can leave off these todos; the functions are all just used for defaulting and can be overridden in the future.

src/celeritas/inp/MucfPhysics.hh
```diff
@@ -68,51 +68,60 @@ struct MucfCycleRate
...
+ * combinations being the \c vector<MucfAtomSpinFlipRate> in \c MucfPhysics
+ * .
  *
  * \note These grids are host-only, with only the final spin flip rate per
- * state (which is just a \c real_type ) for each combination being needed in
- * the stepping loop. This is because these rates are material dependent, and
- * thus can be cached at model construction.
+ * state (which is just a \c real_type ) for each combination being needed
+ * in the stepping loop. This is because these rates are material
+ * dependent, and thus can be cached at model construction.
```
sethrj: vscode is doing some unnecessary reindentation I think :(

src/celeritas/inp/MucfPhysics.hh
```diff
@@ -68,51 +68,60 @@ struct MucfCycleRate
...
  * Minimum requirements for muon-catalyzed fusion:
- * - Muon energy CDF data, required for sampling the outgoing muCF muon, and
+ * - Muon energy CDF data, required for sampling the outgoing muCF muon,
+ * and
  * - Mean cycle rate data for dd, dt, and tt muonic molecules.
  *
- * Muonic atom transfer and muonic atom spin flip are secondary effects and not
- * required for muCF to function.
+ * Muonic atom transfer and muonic atom spin flip are secondary effects and
+ * not required for muCF to function.
```
sethrj: Here too :(

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,133 @@
...
+    //! Material-dependent data calculated at model construction
+    //! \c PhysMatId indexed by \c MuCfMatCompId
+    MaterialItems<PhysMatId> matcompid_to_matid;
+    //! Cycle times per material: [mat_comp_id][muonic_molecule][spin_index]
+    MaterialItems<CycleTimesArray> cycle_times;  //!< In [s]
+    //! \todo Add mean atom spin flip times
+    //! \todo Add mean atom transfer times
+    //!@}
+
+    //! \todo Check whether the data are assigned
```
sethrj:
```suggestion
    //! Check whether the data are assigned
```

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,133 @@
...
+    explicit CELER_FUNCTION operator bool() const
+    {
+        //! \todo Finalize implementation
```
sethrj: ?

src/celeritas/mucf/data/DTMixMucfData.hh
```diff
@@ -0,0 +1,133 @@
...
+        {
+            if (auto const comp_id = MuCfMatCompId{i};
+                this->matcompid_to_matid[comp_id] == matid)
+            {
+                return comp_id;
+            }
+        }
+        // Component ID not found
+        return MuCfMatCompId{};
+    }
```
sethrj, line 93: This should be in a view, it should be labeled `find_` , and note that since we only expect a very few materials to be mucf mat compounds we can do a linear search.

src/celeritas/mucf/interactor/DDMucfInteractor.hh
sethrj: I forgot this was  just for DD and that `Channel` was local to the interactor. I think it's ok then, but you shoudl still replace with an EnumArray rather than a `case`.

src/celeritas/mucf/model/DTMixMucfModel.cc
```diff
@@ -0,0 +1,195 @@
...
+        // Store material ID and calculated data
+        //! \todo Store mean atom spin flip and transfer times
+        vec_physmatid.push_back(PhysMatId{matid});
+        vec_cycle_times.push_back(mat_calculator.cycle_times());
+    }
+
+    make_builder(&host_data.matcompid_to_matid)
+        .insert_back(vec_physmatid.begin(), vec_physmatid.end());
+    make_builder(&host_data.cycle_times)
+        .insert_back(vec_cycle_times.begin(), vec_cycle_times.end());
```
sethrj: You can skip the intermediate vectors and build directly (also try to use the same names as the data):
```suggestion
    CollectionBuilder cycle_times{&host_data.cycle_times};
    CollectionBuilder matcompid_to_matid{&host_data.matcompid_to_matid};
    for (auto const& matid : range(materials.num_materials()))
    {
        auto const& mat_view = materials.get(PhysMatId{matid});

        // Construct calculator for this material
        detail::DTMixMaterialCalculator mat_calculator(mat_view);
        if (!mat_calculator)
        {
            // Skip non-dt mixture materials
            continue;
        }

        // Store material ID and calculated data
        //! \todo Store mean atom spin flip and transfer times
        matcompid_to_matid.push_back(PhysMatId{matid});
        cycle_times.push_back(mat_calculator.cycle_times());
    }
```

Also you're not really doing a "material component" here, you're doing a "FusionMatId" (or equivalent): it's a separate numbering that's based on the physics materials and what supports fusion.

src/celeritas/mucf/model/DTMixMucfModel.cc
```diff
@@ -0,0 +1,195 @@
...
+    // Calculate and cache quantities for all materials with dt mixtures
+    VecCycleTimes vec_cycle_times;
+    VecMatId vec_physmatid;
+    for (auto const& matid : range(materials.num_materials()))
+    {
+        auto const& mat_view = materials.get(PhysMatId{matid});
+
+        // Construct calculator for this material
+        detail::DTMixMaterialCalculator mat_calculator(mat_view);
+        if (!mat_calculator)
```
sethrj: The operator bool isn't semantically very consistent since this is behaving like a helper rather than a function.

src/celeritas/mucf/model/detail/DTMixMaterialCalculator.hh
```diff
@@ -0,0 +1,142 @@
...
+
+    static constexpr units::AmuMass deuterium()
+    {
+        return units::AmuMass{2.014101777844};
+    }
+
+    static constexpr units::AmuMass tritium()
+    {
+        return units::AmuMass{3.016049281320};
+    }
```
sethrj: For now, specify these in the `inp` data and provide defaults. Don't hardcode them.

src/celeritas/mucf/model/detail/DTMixMaterialCalculator.hh
```diff
@@ -0,0 +1,142 @@
...
+    //}
+
+    // Liquid hydrogen density (LHD) unit [1/cm^3]
+    static constexpr auto liquid_hydrogen_density()
+    {
+        return units::InvCcDensity{4.25e22};
+    }
```
sethrj: This should also be input data.

src/celeritas/mucf/model/detail/DTMixMaterialCalculator.hh
```diff
@@ -0,0 +1,142 @@
...
+       auto mat_view = materials.material_view(PhysMatId{matid});
+       DTMixMaterialCalculator material_calculator(mat_view);
+       if (material_calculator)
+       {
+           // Valid d-t mixture material; Store data
+       }
+   }
+ * \endcode
+ */
+class DTMixMaterialCalculator
```
sethrj: I don't think this class structure makes sense. Maybe you could rebuild as an inserter (e.g., `optical/gen/detail/MatScintSpecInserter.hh`, `em/model/detail/SBTableInserter.hh`, ...) and each call to operator() returns a `FusionMatId`?

sethrj (commented):
I would prefer to have at least some level of testing (even if it's running through a bunch of null-ops) here. I see now that the executor is being used, but it would be very helpful to have a code coverage tool tell us what is actually being called and tested.

Maybe it's possible to set up a test that creates a state vector and the model (manually setting all the particle attributes that are needed), and calls the `step` operator for the mucf model?

src/celeritas/mucf/data/DTMixMucfData.hh
stognini: Yep, I was fine with the loop since we will likely have only 1 material. Maybe 2.

I also considered a view, but seemed an overkill for a single function. It originally was a small lambda inside the executor. Do you think the lambda is okay, or should I go ahead and make the view?

src/celeritas/mucf/model/detail/DTMixMaterialCalculator.hh
stognini: Let me rethink it and see if I can make it more like the inserter.

For me it made sense, because it would calculate and cache multiple quantities per material (cycle times, spin flip times, atom transfer times). And since the vast majority of materials are not a d-t mixture, it seemed reasonable to hide that verification in the calculator: If I create the calculator with a given `MaterialView`, the calculator would tell me via its operator bool if that is a valid fusion material---and if so, that meant I could add all those cached quantities to the model data for the executor.

src/celeritas/mucf/model/DTMixMucfModel.cc
stognini: Agreed that I can improve the name. Maybe on the lines of mucf material record id? Since it's just the contiguous internal storing of material cached data?

sethrj: I think `MucfMatId` (and you can keep it in the mucf type file) is exactly what you're describing: for optical and physics materials, the mat ID corresponds exactly to a material record.

src/celeritas/mucf/model/detail/DTMixMaterialCalculator.hh
sethrj: But all these calculations are happening "locally", within a single invocation. There's no need to make a class, precalculate a bunch of stuff, and then make another piece of code insert those through accessors. With an inserter class, you'd do everything locally within `MucfMatId operator()(MatId)`, likely with some private helper functions to calculate these "cached" quantities.

sethrj (approved):
Good enough. I do think we should avoid "skeleton" PRs. I know this is a fairly complex piece of machinery, but in the future let's build up one piece at a time: no skeletons in the code. It's hard to tell if a piece of construction code is correct or optimal if it's not used.

src/celeritas/mucf/model/DTMixMucfModel.hh
```diff
@@ -0,0 +1,75 @@
...
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/mucf/model/DTMixMucfModel.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "celeritas/mat/MaterialParams.hh"
+#include "celeritas/mucf/data/DTMixMucfData.hh"
+#include "celeritas/mucf/executor/DTMixMucfExecutor.hh"  // IWYU pragma: associated
+#include "celeritas/phys/InteractionApplier.hh"  // IWYU pragma: associated
```
sethrj: Not sure these are needed; executor should be in .cc/cu for now

src/celeritas/mucf/process/MucfProcess.hh
```diff
@@ -0,0 +1,60 @@
...
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/mucf/process/MucfProcess.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include "celeritas/Types.hh"
+#include "celeritas/mat/MaterialParams.hh"
+#include "celeritas/phys/ParticleParams.hh"
```
sethrj: Forward declare these; I think they're breaking the CUDA builds.

src/celeritas/mucf/process/MucfProcess.hh
```diff
@@ -0,0 +1,60 @@
...
+    VecModel build_models(ActionIdIter start_id) const final;
+
+    // Get the interaction cross sections for the given energy range
+    XsGrid macro_xs(Applicability) const final;
+
+    // Get the energy loss for the given energy range
+    EnergyLossGrid energy_loss(Applicability) const final;
+
+    //! Whether the integral method can be used to sample interaction length
+    bool supports_integral_xs() const final { return true; }  //! \todo Check
```
amandalund: If this process/model _only_ applies at rest I think this should be false.

amandalund: If this process/model only applies at rest I think this should be false.

stognini: Good catch. Thanks.

sethrj:
@stognini looks like you're missing some `CELER_FUNCTION` decorators:
```
  /__w/celeritas/celeritas/src/celeritas/mucf/executor/DTMixMucfExecutor.hh(56): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
        detail::DTMixMuonicAtomSelector form_atom;
                                        ^
  
  Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
  
  /__w/celeritas/celeritas/src/celeritas/mucf/executor/DTMixMucfExecutor.hh(69): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
        detail::DTMixMuonicMoleculeSelector form_muonic_molecule;
                                            ^
  
  /__w/celeritas/celeritas/src/celeritas/mucf/executor/DTMixMucfExecutor.hh(118): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
                DDMucfInteractor interact(
                                 ^
  
  /__w/celeritas/celeritas/src/celeritas/mucf/executor/DTMixMucfExecutor.hh(125): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
                DTMucfInteractor interact(
                                 ^
  
  /__w/celeritas/celeritas/src/celeritas/mucf/executor/DTMixMucfExecutor.hh(132): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
                TTMucfInteractor interact(
                                 ^
```

sethrj:
@amandalund @stognini are we ready to merge this? According to codecov, literally *none* of the new code is tested or even executed; it simply compiles. I really do not like that. It took three weeks to review and fix the *builds*, and we can't even know if it works yet.

Next time we have to put in a big feature, maybe we focus on hardcoding the main use case, then refactoring and refining as we support more fine-grained features. (E.g., start with supporting only DD fusion, one spin, etc., and at least test that it works.) I think right now our muon code has zero coverage at all.

stognini:
Agree that the lack of testing is a problem. This also is one single executor. E.g. a discrete EM interaction we were in general able to make the interactor and model in one or two PRs.

One option, if you'd like, is to revert the ordering I was planning. Keep this as draft, and:
- The next PR can be one of the interactors (which can use mock data for a test). And will use the bare minimum from this branch (since the big picture here is mostly set);
- Then spin selection;
- Then material calculation;
- Then we add the process into the stepping loop;

sethrj:
The interactor we had the ability to test individually, which was great. I don't want to spend any more time on muons than necessary, but I'd say the next follow-on should be a PR with an integration-level test of muon physics: either in `accel/` or the lower-level `stepper` test.

sethrj:
I'm going to merge this out of expedience to reduce the time spent mitigating code rot; @stognini 's next task is to add a unit test that builds the µCF model and compares low level interactor distributions against the G4 code. Then we'll add an integration test that uses muons.

PR #2176, opened 2026-01-01
Remove use of G4Backtrace

sethrj:
This was only enabled temporarily in geant4 10.7 (15 May 2021) and disabled subsequently by @drbenmorgan for Geant4 11.0.3 (2022-03-18). Since all its code is inlined, including the file means activating some of its components. Instead of touching the signals ourselves, rely on upstream users to set `GEANT4_BUILD_BUILTIN_BACKTRACE=OFF`.


stognini (approved):
Looks good.

PR #2177, opened 2026-01-03
Move global CoreStateCounters to the GPU

LSchwiebert:
This PR continues the effort to allow an entire step to run on the device without host/device synchronization. When building for the device, CoreState counters in src/celeritas/global are now maintained on the device instead of the host. Functions are added to allow these counters to be transferred between the device and the host. To complete this optimization, additional changes are needed, including:

1) Make the same change for the Core State Counters in the optical code (?)
2) Make the same change for the Core State Counters in the auxiliary state code (?)
3) Move the host code that gets or sets counter values to small device kernels


LSchwiebert:
I was unable to figure out how to determine whether the code was running on host or device in the TrackInit test, so I changed the sync_get_counters function to allow it to work for host or device code. I would prefer to revert that change.

LSchwiebert:
The code passes all tests when being built on a system with GPU support, but isn't building correctly when building without the GPU libraries. Please wait until that's fixed to review this PR. I'll delete this message when I push working code.

sethrj:
Awesome! Thanks @LSchwiebert .

LSchwiebert:
The issues that caused compilation and linking issues when building for host-only systems have been resolved. Any feedback/suggestions welcome.

src/celeritas/global/CoreState.cc
```diff
@@ -133,19 +150,84 @@ Range<ThreadId> CoreState<M>::get_action_range(ActionId action_id) const
...
+            = ItemCopier<CoreStateCounters>{stream_id()}(counters.get());
+        stream.sync();
+        return result;
+    }
+    else if constexpr (M == MemSpace::host)
+    {
+        CELER_ASSERT_UNREACHABLE();
+        return CoreStateCounters{};
+    }
+}
```
amandalund, line 159: Instead of using this method to access device counters and `counters()` for host counters, we could use `sync_get_counters()` for both. This should reduce the divergent code paths and make things less error prone and easier to maintain.
 
```suggestion
template<MemSpace M>
CoreStateCounters CoreState<M>::sync_get_counters() const
{
    auto* counters = static_cast<CoreStateCounters*>(this->ref().init.counters.data());
    CELER_ASSERT(counters);

    if constexpr (M == MemSpace::device)
    {   
        auto result = ItemCopier<CoreStateCounters>{this->stream_id()}(counters);
        device().stream(this->stream_id()).sync();
        return result;
    }
    return *counters;
}
```

(Also just return by value since there's not benefit to returning by const value).

src/celeritas/global/CoreState.cc
```diff
@@ -133,19 +150,84 @@ Range<ThreadId> CoreState<M>::get_action_range(ActionId action_id) const
...
+                   stream_id());
+        stream.sync();
+    }
+    else if constexpr (M == MemSpace::host)
+    {
+        CELER_ASSERT_UNREACHABLE();
+    }
+    return;
+}
+#endif
```
amandalund: Similarly here, we can use this method to assign both host and device counters (using the `Copier` instead of the lower-level `copy_bytes` and passing `host_counters` as const ref).

```suggestion
template<MemSpace M>
void CoreState<M>::sync_put_counters(CoreStateCounters const& host_counters)
{
    auto* counters = static_cast<CoreStateCounters*>(this->ref().init.counters.data());
    CELER_ASSERT(counters);

    Copier<CoreStateCounters, M> copy{{counters, 1}, this->stream_id()};
    copy(MemSpace::host, {&host_counters, 1});
    if constexpr (M == MemSpace::device)
    {
        device().stream(this->stream_id()).sync();
    }
}
```

src/celeritas/global/CoreState.cc
```diff
@@ -114,7 +120,18 @@ CoreState<M>::~CoreState()
...
+    {
+        auto& counters = this->counters();
+        num_active = counters.num_active;
+    }
+    else if constexpr (M == MemSpace::device)
+    {
+        auto counters = this->sync_get_counters();
+        num_active = counters.num_active;
+    }
+    CELER_EXPECT(!new_state || num_active == 0);
```
amandalund: Then for example this becomes:

```suggestion
    CELER_EXPECT(!new_state || this->sync_get_counters().num_active == 0);
```

and similar simplifications can be made in the constructor and `reset()`.

src/celeritas/global/Stepper.cc
```diff
@@ -107,14 +107,29 @@ Stepper<M>::~Stepper() = default;
...
+    {
+        auto counters = state_->sync_get_counters();
+        CELER_VALIDATE(counters.num_active == 0,
+                       << "cannot warm up when state has active tracks");
+
+        ScopedProfiling profile_this{"warmup"};
+        state_->warming_up(true);
+        ScopeExit on_exit_{[this] { state_->warming_up(false); }};
+        actions_->step(*params_, *state_);
+        CELER_ENSURE(counters.num_active == 0);
```
amandalund: We can simplify the changes to the `Stepper` as well. Also here I think we'll need to sync again after the step for the postcondition

src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -150,6 +150,34 @@ void OpticalLaunchAction::execute_impl(CoreParams const&,
     (*transport_)(state);
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Launch the optical tracking loop.
+ */
+void OpticalLaunchAction::execute_impl(CoreParams const&,
+                                       CoreStateDevice& core_state) const
```
amandalund: And we can revert to templating this (as well as the methods in the track initialization actions) on memspace instead of duplicating it for the device state.

src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -150,6 +150,34 @@ void OpticalLaunchAction::execute_impl(CoreParams const&,
...
+{
+    auto& state = get<optical::CoreState<MemSpace::device>>(core_state.aux(),
+                                                            this->aux_id());
+    CELER_ASSERT(state.size() > 0);
+
+    auto core_counters = core_state.sync_get_counters();
+    auto counters = state.counters();
```
amandalund:
```suggestion
    auto const& counters = state.counters();
```

src/celeritas/optical/detail/OpticalLaunchAction.cc
```diff
@@ -150,6 +150,34 @@ void OpticalLaunchAction::execute_impl(CoreParams const&,
...
+        || counters.num_pending == 0)
+    {
+        // Don't launch the optical loop if the number of pending tracks is
+        // below the threshold and the core stepping loop hasn't completed yet
+        return;
+    }
+
+    // Transport pending optical tracks
+    (*transport_)(state);
+    core_state.sync_put_counters(core_counters);
```
amandalund: I don't think we need to sync these at the end since they weren't modified?

amandalund (commented):
Looking good @LSchwiebert, thanks for all the work on this!

Because of the different code paths for host/device when accessing/syncing the counters there’s a lot of duplicate code now. I think the next step is to reduce the duplication as much as possible (and I think we should be able to eliminate most of it).

LSchwiebert:
@amandalund Thanks for the detailed feedback -- it's very helpful. I'll incorporate your suggestions and update the PR.

src/celeritas/global/CoreState.cc
LSchwiebert: Thanks. Yes, this is clear. I'll do this throughout. This will significantly reduce the code changes.

src/celeritas/global/CoreState.cc
LSchwiebert: I replaced the existing function with the new version. Thanks.

src/celeritas/global/CoreState.cc
LSchwiebert: I also made this change. Thanks.

src/celeritas/global/Stepper.cc
LSchwiebert: Simplified. And, yes, we do. Thanks for catching this.

src/celeritas/optical/detail/OpticalLaunchAction.cc
LSchwiebert: Done throughout.

src/celeritas/optical/detail/OpticalLaunchAction.cc
LSchwiebert: Updated this, too.

src/celeritas/optical/detail/OpticalLaunchAction.cc
LSchwiebert: Agreed. I removed that line.

sethrj (approved):
Kickass work! Thanks for getting this done and making it look so elegant 😁

src/celeritas/track/ExtendFromPrimariesAction.cu
```diff
@@ -24,11 +24,9 @@ void ExtendFromPrimariesAction::process_primaries(
     PrimaryStateData<MemSpace::device> const& pstate) const
 {
     auto primaries = pstate.primaries();
+    auto counters = state.sync_get_counters();
     detail::ProcessPrimariesExecutor execute_thread{
-        params.ptr<MemSpace::native>(),
-        state.ptr(),
-        state.counters(),
-        primaries};
+        params.ptr<MemSpace::native>(), state.ptr(), counters, primaries};
```
amandalund: We could access the counters directly from the state in the primaries executor instead of syncing and passing them in here.

src/celeritas/track/ExtendFromSecondariesAction.cu
```diff
@@ -64,7 +64,7 @@ void ExtendFromSecondariesAction::process_secondaries(
     launch(core_state,
            Executor{core_params.ptr<MemSpace::native>(),
                     core_state.ptr(),
-                    core_state.counters()});
+                    core_state.sync_get_counters()});
```
amandalund: Same thing here.

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -208,7 +214,7 @@ void partition_initializers(
     auto start = device_pointer_cast(init.indices.data());
     auto end = start + count;
     auto stencil = static_cast<TrackInitializer*>(init.initializers.data())
-                   + counters.num_initializers - count;
+                   + init.counters.num_initializers - count;
```
amandalund: Have you tried building with a configuration that requires thrust? It doesn't look like this is accessing the counters correctly.

src/celeritas/track/detail/TrackInitAlgorithms.cu
```diff
@@ -256,14 +264,15 @@ void partition_initializers(
     auto data = device_pointer_cast(init.indices.data());
     // Allocate storage for the number of neutral tracks (unused by celeritas)
     DeviceVector<size_type> num_neutral{1, stream_id};
```
amandalund: Can we remove this now?

```suggestion
```

amandalund (approved):
Awesome, great work @LSchwiebert!

src/celeritas/track/ExtendFromPrimariesAction.cu
LSchwiebert: Yes. Thanks. I'll change that.

src/celeritas/track/ExtendFromSecondariesAction.cu
LSchwiebert: Agreed. Thanks. Will do.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Oops. Sorry. Thanks for noticing that. It fell off my mental to do list. Yes, I'm pretty sure that's not going to work. I just put in some placeholder code to get it to compile and then forgot to go back and fix it. Will fix.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Yes. Will remove. Thanks.

src/celeritas/track/detail/TrackInitAlgorithms.cu
LSchwiebert: Fixed. I commented out some code for all three functions to use the thrust commands, so they are all working. I also replaced the low-level copy_bytes call with the Copier call.

PR #2179, opened 2026-01-04
Add NO_COLOR support and new env helper

sethrj:
This adds support for `NO_COLOR` (see https://no-color.org) and defines a new `getenv_flag_lazy` that will only evaluate the default value if none is provided. This will be used to check for the presence of a GPU by only calling CUDA if no environment variable is specified.


sethrj:
@pcanal Would you mind taking a peek at this? The lazy env stuff is to support the new `inp::System` setup.

src/corecel/sys/Environment.cc
```diff
@@ -14,16 +14,19 @@
...
 
+using BoolFunc = std::function<bool()>;
+
 namespace celeritas
 {
 namespace
 {
 //---------------------------------------------------------------------------//
-std::mutex& getenv_mutex()
+std::recursive_mutex& getenv_mutex()
```
pcanal, line 30: Is the motivation for this change (using a recursive mutex) described in the git log?  Either way, consider also adding it here.

pcanal (approved):
LGTM.  Thanks.

src/corecel/sys/Environment.cc
sethrj: Somewhat but good point, I'll add a note in the code here.

PR #2154, opened 2025-12-11
Add LArSoft plugin skeleton

sethrj:
This is the first step in the Celeritas side of #1548 . I have built and successfully run the new test on `fnal-dev-sl7:latest` using `scisoftbuild01` and the environment from #2153 with both release and reldeb configurations.

From start to finish, here's what's new:
- Add a few thin wrappers `FindX.cmake` for LArSoft dependencies: cetmodules, art, lardataobj
- Add `FindLArSoft.cmake` which wraps those scripts into a single big one with the correct ordering and name
- Add cmake option `CELERITAS_USE_LArSoft` and associated dependency output and logic
- Add a new library `larceler` that will include front-end components used to convert object models between LArSoft and Celeritas (e.g., create optical generators using edep objects)
- Add a small `inp` setup struct for building `LarStandaloneRunner`, which will wrap the standalone Celeritas optical solver for testing purposes
- Add a skeleton for `LarStandaloneRunner` that currently just loads the input GDML file
- Add a FHiCL config struct that and conversion to the `inp` setup struct
- Add a "class tool" plugin `LarCelerStandalone` that inherits from the work-in-progress `OpticalSimInterface`, defines Art parameters, and executes `LarStandaloneRunner`
- Add a test of `LarStandaloneRunner`
- Add initial documentation for LArSoft integration

TO DO (potentially on this PR?):
- Add CI build for CVMFS (@pcanal or @rahmans1 could help?)

Issue #1548, opened 2024-12-19
Integrate with LArSim via standalone plug-in


sethrj:
https://github.com/LArSoft/larsim is used by DUNE (#1344 ) and other neutrino physics-oriented detectors:
- ICARUS https://icarus.fnal.gov/
- SBND https://sbn-nd.fnal.gov/
- $\mu$ BooNE https://microboone.fnal.gov/

This represents "phase 1" of our integration plan: rather than running simultaneously with Geant4, we take the pre-processed hits (with the number of photons already calculated) and determine the detector responses in a standalone simultion.

## Execution plan

See google doc for more details.

- [x] LArSoft cmake and tool interface #2154 
- [x] Add optical propagation interface to larsoft https://github.com/nuRiceLab/larsim/pull/1
- [x] Refactor LarSoft PDFastSimPAR_module to call optical propagation module with plugin tool (@stognini )
- [x] Test Celeritas build alongside MRB and calling it as part of a larsoft/dune workflow once the `OpticalSimInterface` module is complete (@sethrj )
- [x] Update `LarStandaloneRunner` to build and run an optical Transporter (see #2197)
- [ ] Test that we can build the necessary optical physics *after* a geant4 run manager has been torn down
- [x] Collect and sort photon hits: per detector, sorted by time: use per-detector [OBTRHelper class](https://github.com/LArSoft/lardataobj/blob/develop/lardataobj/Simulation/OpDetBacktrackerRecord.h) for short term, be smart about it for long term

sethrj:
Note from our discussion Friday that LArSoft currently uses Geant4 10.6 but is encouraging its users to move to 11.4 .

sethrj:
Test system setup: #2153

stognini:
# Generating MC samples

With help from Aaron, this is the workflow to generate example Geant4 ROOT files with `art` energy deposition objects.

- `dunesw` `fcl` job files:
  - GENIE: `prodgenie_nu_dune10kt_1x2x6.fcl`
  - Geant4: `standard_g4_dune10kt_1x2x6.fcl`
- `fcl` files are part of the environment once `dunesw` is set up and thus can be called from anywhere

## Generate GENIE events
```sh
$ lar -c prodgenie_nu_dune10kt_1x2x6.fcl -n 10 -o genie_output.root
```

## Generate energy depositions from Geant4
```sh
[...]

PR #2180, opened 2026-01-06
Link against development LArSim

sethrj:
This adds a copy of the `IOpticalPropagation` file from https://github.com/nuRiceLab/larsim/pull/1 for testing against the release version of larsoft as an independent CMake build, while supporting linking against the nuRiceLab development version as part of `mrb`. It also addresses fixes from #2154.

I've verified that it builds inside MRB, but there's probably a few steps before it's fully "integrated": whereas larsim installs to `localProducts.../larsim/v10_09_01/slf7.x86_64.e26.prof/lib/`, the celeritas-as-mrb installs to `localProducts.../lib/`. We may need an mrb expert to tide us over till the new spack-based workflow is in place.

### DIFF
--- a/doc/usage/installation.rst
+++ b/doc/usage/installation.rst
@@ -233,7 +233,7 @@ It includes environment files for quickly getting started on systems including N
    $ cd celeritas
    $ ./scripts/build.sh base
 
-You can, of course, manually build as a
+You can, of course, build manually:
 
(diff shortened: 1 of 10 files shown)


stognini (approved):
Thanks!

sethrj:
Tagging #1548

Issue #2030, opened 2025-10-10
Replace SetupOptions with `inp::FrameworkInput`


sethrj:
The parameter setup that's evolved in `accel` is fragile and half-duplicated in several other places in the code. We want migrate to using the `inp` data structures.

UI integration:
- [ ] Add G4UI wrapper class that stores values into a JSON object (refactoring `SetupOptionMessenger` as needed)
- [ ] Add JSON converters for the key celeritas configuration classes (system, scoring,  control)

Input integration:
- [ ] Have `SetOptions` convert immediately to `inp`, have `SharedParams` take `FrameworkInput` (and for backward compatibility, leave setupoptions)
- [ ] Add a `SetInput` method to `IntegrationBase`
- [ ] At initialize, apply UI json options to stored input, then reset UI options

PR #2181, opened 2026-01-07
Fix InitializedValue semantics

sethrj:
It wasn't actually used in the main codebase but will be as part of the inp refactor (#2030) but the finalizer stuff for `InitializedValue` was inconsistent. With this change, the finalizer will be called on every non-default value that gets overwritten or destroyed, which is what one would expect for MPI finalize, memory allocation, etc.

### DIFF
--- a/src/corecel/cont/InitializedValue.hh
+++ b/src/corecel/cont/InitializedValue.hh
@@ -9,6 +9,8 @@
 #include <type_traits>
 #include <utility>
 
+#include "corecel/Macros.hh"
+
 namespace celeritas
 {
 namespace detail
@@ -17,7 +19,7 @@ namespace detail
 template<class T>
 struct DefaultFinalize
 {
-    void operator()(T&) const noexcept {}
+    CELER_FORCEINLINE void operator()(T&) const noexcept {}
 };
 //---------------------------------------------------------------------------//
 }  // namespace detail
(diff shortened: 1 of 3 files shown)


src/corecel/cont/InitializedValue.hh
```diff
@@ -73,50 +65,42 @@ class InitializedValue
     {
     }
 
-    ~InitializedValue() = default;
+    ~InitializedValue() noexcept(noexcept_finalize_)
+    {
+        if (value_ != T{})
```
pcanal: Why is the default value 'special'?   Is the semantic 'not user assigned' or really 'is the same value as the default'?

sethrj: This is meant for types where the default value means unassigned, uninitialized, etc: opaque IDs, pointers, sizes. I can update the class documentation to reflect this. The class is unfortunately a bit half-baked...

pcanal: >  I can update the class documentation to reflect this. 

Please :)

src/corecel/cont/InitializedValue.hh
```diff
@@ -73,50 +65,42 @@ class InitializedValue
...
+        {
+            Finalizer{}(value_);
```
esseivaju: should we add a `static_assert(std::is_default_constructible_v<Finalizer>)` in the class definition

esseivaju (approved):
Looks good to me! Just a minor comment

PR #2182, opened 2026-01-07
Refactor logger handles

sethrj:
Another part of #2030, this eliminates the dependence between the logger setup and MPI (and also eliminates confusing side effects about the environment variables). The MPI-based logging defaults will be reinstated soon.


pcanal (approved):
LGTM.

PR #2183, opened 2026-01-08
Document and add checks about tracking manager construction

sethrj:
It's an odd subtlety that a tracking manager gets created for the master thread in an MT run, even though it will never be sent any tracks. A minor change avoids accessing the static `LocalTransporter` on the master thread, and it documents some of the oddities.


amandalund (approved):
Thanks for the update and improved docs!

PR #2185, opened 2026-01-09
Centralize default capacity values

amandalund:
This moves the default capacity values, which were previously defined in accel/celer-sim/celer-g4, to `inp::StateCapacity`.


sethrj (approved):
Great! That's one less complicated thing to worry about.

esseivaju (approved):
Looks good thanks!

PR #2184, opened 2026-01-08
Disable debug assertions for code coverage tests

sethrj:
I realized the easy way to disable codecov false positives (https://github.com/celeritas-project/celeritas/issues/2042) for all the debug assertions is probably to just disable the debug assertions themselves so that they don't generate code. Validation errors will still generate code, but we often test those (and generally should).

Here are the comparisons between develop and three different versions: development which has assertions enabled; *discard* assertions (D); *ignore* assertions with gcov macros (I); and both discard *and* ignore (D&I).

```diff
@@                      Coverage Diff                    @@
##                          D       I     D&I            ##
##           develop  57756d3 a72cd6d 19af33f      +/-   ##
===========================================================
+ Coverage    85.10%   85.96%  85.96%  86.12%   +1.02%     
===========================================================
  Files         1297     1287    1287    1287      -10     
  Lines        45155    40719   43087   40706    -4449     
  Branches     17448    12638   12993   12986    -4462     
===========================================================
- Hits         38430    37039   37039   35059    -3371     
+ Misses        4719     5039    5039    4635      -84     
+ Partials      2006     1009    1009    1012     -994     
```

I think it's appropriate for the coverage to test the "release" build. A follow-on PR will improve always-on coverage for the data `operator bool` which previously, for the most part, have only been touched by `CELER_ASSERT(host_data)`.

PR #2186, opened 2026-01-09
Ignore `BUILTIN_X` when `CELERITAS_USE_X` is false

sethrj:
This fixes an issue where dependencies such as covfie would be downloaded even when unused. Related to those changes, the github ubuntu cmake presets have been broken apart for an increase in modularity.

Split off #2184.


pcanal (approved):
LGTM.

PR #2187, opened 2026-01-09
Simplify along-step propagation

sethrj:
Working toward error mitigation for geometry propagation, for ORANGE and performance and #538, this simplifies the propagation step so that we have just a single "track propagator" that's templated on the field type.


src/celeritas/alongstep/detail/PropagationApplier.hh
```diff
@@ -21,100 +18,44 @@ namespace celeritas
...
-    bool tracks_can_loop;
     Propagation p;
     {
 #if CELERITAS_DEBUG
         Real3 const orig_pos = track.geometry().pos();
 #endif
-        auto propagate = make_propagator(track);
-        p = propagate(sim.step_length());
-        tracks_can_loop = propagate.tracks_can_loop();
+        p = this->propagate(track);
```
esseivaju, line 58: `propagate` makes it look like this is a member function
```suggestion
        p = this->propagator(track);
```

src/celeritas/alongstep/detail/PropagationApplier.hh
```diff
@@ -21,100 +18,44 @@ namespace celeritas
...
-    {
-    }
-};
-
-template<class MP>
-struct PropagationApplier<MP, std::enable_if_t<kernel_max_blocks_min_warps<MP>>>
-    : public PropagationApplierBaseImpl<MP>
-{
-    static constexpr int max_block_size = MP::max_block_size;
-    static constexpr int min_warps_per_eu = MP::min_warps_per_eu;
```
esseivaju, line 63: Do we want to keep this functionality? [Kernel launch implementation](https://github.com/sethrj/celeritas/blob/dab9fffa5f1431ae0e07fba5503d9ad03d49244c/src/corecel/sys/detail/KernelLauncherImpl.device.hh#L44-L45) has traits to detect whether the function object has `max_block_size`, and `min_warps_per_eu` members.

src/celeritas/alongstep/detail/PropagationApplier.hh
```diff
@@ -21,100 +18,44 @@ namespace celeritas
...
-};
-
-template<class MP>
-struct PropagationApplier<MP, std::enable_if_t<kernel_max_blocks<MP>>>
-    : public PropagationApplierBaseImpl<MP>
+template<class TP>
+struct PropagationApplier
 {
-    static constexpr int max_block_size = MP::max_block_size;
+    inline CELER_FUNCTION void operator()(CoreTrackView& track);
```
esseivaju, line 33:
```suggestion
    static_assert(
        std::is_invocable_r_v<Propagation, TP, CoreTrackView&>,
        "TP must be callable as Propagation (*)(CoreTrackView&)");
    inline CELER_FUNCTION void operator()(CoreTrackView& track);
```

we also have a lot of templating in `PropagationApplier<Propagator<Field>>` which internally uses `Integrator<Equation<Field>>`, maybe we can add a few more assertions for those types to explicitly state requirements.

src/geocel/vg/detail/BVHNavigator.hh
```diff
@@ -49,7 +49,11 @@ class BVHNavigator
     using NavState = detail::VgNavStateWrapper;
 #endif
 
-    static constexpr vg_real_type kBoundaryPush = 10 * vecgeom::kTolerance;
+#ifdef VECGEOM_FLOAT_PRECISION
+    static constexpr vg_real_type kBoundaryPush = 10 * 1e-3f;
+#else
+    static constexpr vg_real_type kBoundaryPush = 10 * 1e-9;
+#endif
```
esseivaju, line 56: Has VecGeom removed `kTolerance` in a recent version?

esseivaju (commented):
Looks good to me, just a few comments.

src/celeritas/field/LinearPropagator.hh
```diff
@@ -77,6 +58,13 @@ CELER_FUNCTION auto LinearPropagator<GTV>::operator()(real_type dist)
     CELER_EXPECT(dist > 0);
 
     result_type result = geo_.find_next_step(dist);
+    if (CELER_UNLIKELY(geo_.failed()))
+    {
+        // Mark as looping: update momentum and bump
+        result.looping = true;
```
amandalund: Where is this used? It seems a bit odd to mark a track as looping with no field.

amandalund (approved):
Nice simplification!

src/celeritas/field/LinearPropagator.hh
sethrj: Oh shoot good catch, this was me trying to find a way around failures with the vecgeom relocation, from a separate branch.

src/geocel/vg/detail/BVHNavigator.hh
sethrj: No, the problem is that they declared them `const __constant__` in CUDA! That's a bad idea in general but it also means it's not constexpr :( This error might have been caught only by a newer CUDA version I tried to compile with.

src/celeritas/alongstep/detail/PropagationApplier.hh
sethrj: I want to restore it someday because it's a really cool capability, but right now we aren't using it anywhere (we're always using block size 256) and I'd like to simplify the code first. Later we can play with the warp size. The lock size also will probably depend on the geometry, field type, etc., which we currently aren't accounting for.

src/celeritas/alongstep/detail/PropagationApplier.hh
sethrj: Feature or bug? 🤔 In python you can't tell a callable member data from a function (by design, really). I still think it's still good to give verby names class instances that act like functions. To make it clearer that it's data, I could rename it `propagate_`?

src/celeritas/alongstep/detail/PropagationApplier.hh
sethrj: Once we can use C++20 I hope we can use Concepts for those...

src/celeritas/alongstep/detail/PropagationApplier.hh
esseivaju: Okay, I think I just like being able to easily tell between code/data with strong typing. I'd just leave it since it as is follows our current naming conventions (or `propagate_` with declaring `PropagationApplier` as a class).

src/celeritas/alongstep/detail/PropagationApplier.hh
esseivaju: Hopefully soon 😄

PR #2188, opened 2026-01-09
Fix ddceler build error due to IWYU 

sethrj:
A recent change caused ddceler (which isn't currently required to pass under CI) to break due to a missing logger include. I fixed that as well as a minor standards issue I didn't catch with the formatting of `CELER_VALIDATE`.


rahmans1 (approved):
LGTM.

PR #2135, opened 2025-12-01
Offload optical photons from Geant4 to optical track

Rashika-Gupta:
This PR introduces optical track offloading from Geant4 to Celeritas  #2103 

- `LocalOpticalOffload` is renamed to `LocalOpticalGenOffload` to distinguish generator-based optical offloading from track-based offloading.
- `TrackOffloadInterface` generalizes the tracking manager to support both EM and optical track offload implementations.
- `LocalOpticalTrackOffload` classes provide a dedicated pathway for buffering and  transporting (in the follow up PR) optical photon tracks.

PR #2189, opened 2026-01-10
Add optical problem input and setup

amandalund:
This adds an optical-only problem input and setup for cases where we only want to run the optical stepping loop in Celeritas. For now it's only used and tested in accel (when offloading optical distributions) but will be used standalone optical problems as well. I struggled a bit trying to figure out how best to set this up and could definitely use some feedback on how to improve it.


src/corecel/io/OutputRegistry.hh
```diff
@@ -40,6 +40,9 @@ class OutputRegistry
     // Add an interface for writing
     void insert(SPConstInterface);
 
+    // Add an interfaces for writing system diagnostics
+    void insert_system_diagnostics();
+
```
sethrj: Since this uses only public functions, I'd make it a free function

src/celeritas/setup/Problem.cc
```diff
@@ -755,6 +712,140 @@ ProblemLoaded problem(inp::Problem const& p, ImportData const& imported)
...
+    //! \todo Get sensitive detectors
+
+    CELER_ASSERT(pi);
+    auto params = std::make_shared<optical::CoreParams>(std::move(pi));
+
+    // Construct the optical generator
+    std::visit(Overload{
+                   [&](inp::OpticalEmGenerator) {
+                       CELER_VALIDATE(false,
+                                      << "OpticatEmGenerator cannot be used "
```
sethrj:
```suggestion
                                      << "OpticalEmGenerator cannot be used "
```
👓🐱

sethrj (approved):
Looks great! The `OpticalProblem` may warrant some refinement in the future but getting this in is a huge step in the right direction. Thanks!

(PS this will conflict with #2135 but since @Rashika-Gupta has been out I think we can work with her on merging that when she returns.)

PR #2191, opened 2026-01-12
Fix edge cases with optical setup and runtime

sethrj:
This adds a high-level coverage test for optical physics with surfaces, originally targeting #1928 , but using a test input that discovered two edge cases we didn't account for:

- Total cross section can be exactly zero in vacuum
- Not all optical materials have to be used in the geometry (we could turn this into a warning instead of a subtle informational message)

### DIFF
--- a/test/accel/TrackingManagerIntegration.test.cc
+++ b/test/accel/TrackingManagerIntegration.test.cc
@@ -14,11 +14,13 @@
 #include <G4UserTrackingAction.hh>
 #include <G4VModularPhysicsList.hh>
 
+#include "corecel/cont/Array.hh"
 #include "corecel/io/Logger.hh"
 #include "geocel/GeantUtils.hh"
 #include "geocel/UnitUtils.hh"
 #include "celeritas/ext/GeantParticleView.hh"
 #include "celeritas/global/CoreState.hh"
+#include "celeritas/inp/Events.hh"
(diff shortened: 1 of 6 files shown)


test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -526,6 +514,127 @@ TEST_F(OpNoviceOptical, run)
...
+                << accum_stats.flushes << " flushes";
+            EXPECT_GT(accum_stats.steps, 0);
+            EXPECT_GT(accum_stats.step_iters, 0);
+            EXPECT_GT(accum_stats.flushes, 0);
+
+            auto& aux_state = local_transporter.GetState().aux();
+            auto counts = optical_collector->buffer_counts(aux_state);
+            EXPECT_EQ(0, counts.buffer_size);  //!< Pending generators
+            EXPECT_EQ(0, counts.num_pending);  //!< Photons pending generation
+            EXPECT_EQ(0, counts.num_generated);  //!< Photons generated
```
amandalund, line 615: I always forget that `num_generated` _should_ be zero here (counters get reset once all tracks have been generated).

amandalund (approved):
Nice, thanks!

sethrj:
Shit, I didn't update with #2189 before merging.

PR #2193, opened 2026-01-15
Add documentation and debug statements about Geant4 logger redirection

sethrj:
This adds some additional documentation about debugging in Celeritas.  It also makes the `ScopedGeantLogger` code slightly safer by not changing the class data based on whether `USE_G4` is defined.


esseivaju (approved):
LGTM!

PR #2195, opened 2026-01-16
Add copilot agent instructions

sethrj:
Copilot volunteered to scan the codebase and create a set of high-level agent instructions to summarize the repo. I modified them a bit (c1c1f0229d7138b98987a07966933f447621fd9e) to reflect the more important cases.


AGENTS.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Quick Start
+
+Use a standard cmake workflow:
+```bash
+cmake --preset=base
+cmake --build --preset=base
+ctest --preset=base
+```
+
```
esseivaju, line 31:
```suggestion

If you run into environment errors when building Celeritas, you can try to use `scripts/build.sh base` to setup the user environment on supported systems in `scripts/env` 
```

AGENTS.md
esseivaju, line 1: We could consider having this as an `AGENTS.md` file telling the agent to read these instructions.

esseivaju (approved):
LGTM!

pcanal (approved):
LGTM.  It can even be helpful for new comers to read :)

AGENTS.md
sethrj: I'm not super familiar with such workflows but putting at the top level sounds useful for humans too, as @pcanal sqid.

PR #743, opened 2023-05-02
Access params/state through global rather than constant memory

sethrj:
For CUDA kernels, this now uses the *on-device pointers* to host and state data, rather than passing the pointers themselves. The first two refactoring commits pass `Ptr` instead of `Ref` (i.e. a single pointer rather than a copy of the whole `struct` of pointers) to the action kernels and along-step kernels. The occupancy decreases:
```diff
    "occupancy": {
-    "along_step_apply_mean_eloss": 0.625,
+    "along_step_apply_mean_eloss": 0.5,
-    "discrete_select": 0.5,
+    "discrete_select": 0.375,
-    "pre_step": 0.5,
+    "pre_step": 0.375,
    },
```
but there's no observable performance difference.

The final commit (passing pointers to the `interact` function) **changes the answers** for `cms2018+field+msc-vecgeom-gpu`. There's no obvious explanation why this would happen. I added an always-on assertion to the `trackinit` (see #730) to see if the existing code is somehow breaking, but it did not fire... so I can't say whether the old result or the new result (or neither) is correct.

Issue #639, opened 2023-02-10
Debug assertion failure: `is_soft_unit_vector(init.dir)`


sethrj:
@amandalund discovered that ORANGE geometries are now failing after #625 fixed positron annihilation: secondaries being created *only on GPU* trigger an assertion:
```
/home/s3j/.local/src/celeritas/src/orange/OrangeTrackView.hh:236:
celeritas: internal assertion failed: is_soft_unit_vector(init.dir)
```

This shows up in basically all the regression problems that use ORANGE. It's not clear why we don't see this in our `app/demo-loop` test.

Because v0.2.0 doesn't include positron annihilation (whoops!) we don't see the problem. The `backports/v0.1` branch, which was before the recent multi-universe work on ORANGE but includes the annihilation fixes, does *not* show the problem.

amandalund:
Actually I do also see it in the `app/demo-loop` GPU test (with debug on and reldeb build).

sethrj:
I wonder if it's something to do with the "let's help our debugging" implementation when resizing states:
```c++
template<class T, MemSpace M, class I>
void CollectionBuilder<T, M, I>::resize(size_type size)
{
    CELER_EXPECT(this->storage().empty());
    this->storage() = StorageT(size);
    if constexpr (CELERITAS_DEBUG && M == MemSpace::host)
    {
        // Fill with invalid values to help with debugging on host
        detail::fill_invalid(&col_);
    }
}
```

Either way, we clearly need to add a reldeb configuration to our CI after this...

sethrj:
Necessary conditions via @amandalund :
-`CELERITAS_DEBUG=true`
- release or reldeb build (some optimizations)
- GPU only, not CPU
- positron annihilation enabled

directions coming out of interaction are good; but when trying to create secondaries the directions have been zeroed out

amandalund:
One more observation: running with a single thread on the GPU it fails with a different assertion:
```console
/home/alund/celeritas_project/celeritas/src/corecel/data/Collection.hh:367:
celeritas: internal assertion failed: i < this->size()
```
[...]

PR #2197, opened 2026-01-17
Add standalone optical runner

amandalund:
This adds a standalone optical runner for setting up an optical-only problem in Celeritas and transporting photons generated through either a primary generator, track initializers, or Cherenkov/scintillation distributions.

Draft for now because there are still some todos/questions related to setup and testing.


src/celeritas/optical/gen/PrimaryGenerator.hh
```diff
@@ -79,7 +80,8 @@ PrimaryGenerator::operator()(Generator& rng)
     result.direction = sample_with(visit, data_.angle, rng);
     do
     {
-        result.polarization = ExitingDirectionSampler{0, result.direction}(rng);
+        result.polarization = make_unit_vector(
+            make_orthogonal(sample_polarization_(rng), result.direction));
```
amandalund: Using the `ExitingDirectionSampler` to sample polarization was leading to a huge number of iterations in the rejection loop in some cases when the direction was nearly +z. For example, it took over 2 billion iterations with `direction = make_unit_vector(Real3{1e-3, -1e-3, 1})`.

We should probably add a more robust helper function for sampling polarization that can be used elsewhere in the code as a separate PR.

test/celeritas/optical/Runner.test.cc
```diff
@@ -0,0 +1,96 @@
...
+// TEST FIXTURES
+//---------------------------------------------------------------------------//
+
+class LArSphereRunnerTest : public Test
+{
+  public:
+    void SetUp() override
+    {
+        // TODO: How should we disable the device for tests? (This won't work
+        // because the device is activated in test_main)
```
amandalund: This test is essentially one of the `Generator` tests refactored to use `inp`/`setup` to build the problem instead of the `GlobalTestBase`. Ideally I could instead refactor all of those to use the optical runner this way, but we'd need to get around the "running Geant4 more than once per execution" issue.

I'm also not sure how we should enable/disable the device for these tests.

src/celeritas/setup/StandaloneInput.cc
```diff
@@ -102,6 +102,60 @@ StandaloneLoaded standalone_input(inp::StandaloneInput& si)
...
+    // loaded during setup, or are some a combination of both?
+
+    // Manually enable Cherenkov and scintillation if set in input since only
+    // optical photon processes are imported from Geant4
+    if (std::holds_alternative<inp::OpticalOffloadGenerator>(
+            si.problem.generator))
+    {
+        si.problem.physics.cherenkov = si.geant_setup.cherenkov.enable;
+        si.problem.physics.scintillation = si.geant_setup.scintillation.enable;
+    }
```
amandalund, line 149: Right now I guess it makes most sense to enable them in `GeantPhysicsOptions`?

test/celeritas/optical/Runner.test.cc
sethrj: Maybe we use something like the `celeritas_g4_add_tests` and `celeritas_add_integration_tests` cmake macro/function (see `test/accel/CMakeLists`) and just pass the environment variable to enable/disable? We'd have one test executable but only one combination of problem/memspace per launch.

src/celeritas/optical/gen/PrimaryGenerator.hh
sethrj: That sounds like a good task! It looks like there's already a `todo` there; so a helper function documenting the explosion in required samples would be a nice first task. (Good for a student and/or some AI coding.)

test/celeritas/optical/Runner.test.cc
amandalund: Ok I tried hacking together something that seems to be working, but please feel free to clean it up :)

src/celeritas/optical/Runner.cc
```diff
@@ -0,0 +1,145 @@
...
+    CELER_ASSERT(this->params()->aux_reg());
+    state_->aux() = std::make_shared<AuxStateVec>(
+        *this->params()->aux_reg(), memspace, stream_id, num_tracks);
+
+    std::move(osi) = {};
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Transport tracks generated with a primary generator
```
sethrj: Missing some trailing periods here and in the other strings.

src/celeritas/optical/Runner.hh
```diff
@@ -0,0 +1,83 @@
...
+    explicit Runner(inp::OpticalStandaloneInput&&);
+
+    // Transport tracks generated with a primary generator
+    Result operator()();
+
+    // Transport tracks generated directly from track initializers
+    Result operator()(DirectGeneratorAction::SpanConstData);
+
+    // Transport tracks generated through scintillation or Cherenkov
+    Result operator()(GeneratorAction::SpanConstData);
```
sethrj: I think type aliases for these would improve readability.

src/celeritas/setup/System.cc
```diff
@@ -29,9 +30,16 @@ void system(inp::System const& sys)
...
+        auto inserted = env.insert(kv);
+        if (!inserted)
+        {
+            CELER_LOG(warning)
+                << "Failed to set environment variable " << kv.first << ": '"
+                << kv.second << "' because it is already set to " << kv.first
+                << ": '" << env[kv.first] << "'";
+        }
     }
+    CELER_LOG(debug) << "Celeritas environment variables: " << env;
```
sethrj, line 42: I have a WIP branch to help clean this mess 😅

sethrj (approved):
A little hacky, as you say, but it gets us to the next milestone. Thanks!

src/celeritas/setup/StandaloneInput.cc
sethrj: Yep. The goal is for `physics.cherenkov` be a variant:
- `Automatic = std::monostate`: set from Geant4 physics
- `Disabled`: prevent import and do not create celeritas process
- `CherenkovProcess`: actual physics input data

amandalund:
@sethrj any idea why the tests are crashing with the Geant4 navigator?

I'm also still seeing the celeritas/geo_Geometry device test failures, and I'm not sure what's going on there or what I might have changed here that could affect those tests :(

Copied from slack:

> I'm seeing some weird failures in the celeritas/geo_Geometry device tests (with a vecgeom build but not with orange):
> ```console
> alund@gpu07:~/celeritas_project/celeritas/build-reldeb$ ./test/celeritas/geo_Geometry --gtest_filter=SimpleCms*
> Celeritas version 0.7.0-dev.266+optical-runner.2aca7d452
> Note: Google Test filter = SimpleCms*
> [==========] Running 2 tests from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 2 tests from SimpleCmsTest
> [ RUN      ] SimpleCmsTest.avg_path
> info: Loading Geant4 geometry from GDML at /home/alund/celeritas_project/celeritas/test/geocel/data/simple-cms.gdml
> status: Loading VecGeom geometry from in-memory Geant4 geometry
> status: Converting Geant4 geometry
> status: Initializing tracking information
> info: CUDA stack size was changed from 1024 to 8192; restoring to original values
> status: Constructing metadata
> status: Running heuristic test on host
> info: Creating 1 device streams
> status: Running heuristic test on device
> /home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> /home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> /home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> ...
> ```
> It looks like the failure is happening when sampling a direction from an isotropic distribution in HeuristicGeoExecutor.  Adding some debugging there
> ```diff
> diff --git a/src/corecel/random/distribution/IsotropicDistribution.haha > b/src/corecel/random/distribution/IsotropicDistribution.hh
> index e2085a460..d19ac79bc 100644
> --- a/src/corecel/random/distribution/IsotropicDistribution.hh
> +++ b/src/corecel/random/distribution/IsotropicDistribution.hh
> @@ -73,6 +73,11 @@ CELER_FUNCTION auto IsotropicDistribution<RealType>::operator()(Generator& rng)
>      -> result_type
>  {
>      real_type const costheta = sample_costheta_(rng);
> +    if (!(costheta >= -1 && costheta <= 1))
> +    {
> +        printf("costheta=%.16e\n", costheta);
> +    }
> +    CELER_EXPECT(costheta >= -1 && costheta <= 1);
>      real_type const phi = sample_phi_(rng);
>      return from_spherical(costheta, phi);
>  }
> ```
> gives this output from the test:
> ```console
> status: Running heuristic test on device
> costheta=3.0976430986012091e+00
> costheta=3.4996892896420899e+00
> costheta=5.6417703130975561e+00
> costheta=2.3823207097853132e+00
> ...
> /home/alund/celeritas_project/celeritas/src/corecel/random/distribution/IsotropicDistribution.hh:80:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> /home/alund/celeritas_project/celeritas/src/corecel/random/distribution/IsotropicDistribution.hh:80:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> /home/alund/celeritas_project/celeritas/src/corecel/random/distribution/IsotropicDistribution.hh:80:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> /home/alund/celeritas_project/celeritas/src/corecel/random/distribution/IsotropicDistribution.hh:80:
> celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
> ```

sethrj:
Oh yes, there's something weird about how the geo state is allocated for g4 geometry. Stupid thread local allocators. You'll note for the accel tests that I've had to disable it. 😔

sethrj:
Sorry I forgot about the slack message, I'll take a look... ~~but maybe some NaN?~~ nope I clearly didn't read closely enough

amandalund:
A few more data points:
- The failure does not occur in debug mode but does in reldeb and release
- I still see it with a single GPU thread
- The value that's returned in `UniformRealDistribution::operator()` is within the correct bounds, but when that value is checked in `IsotropicDistribution::operator()` it's not

src/celeritas/setup/StandaloneInput.cc
```diff
@@ -102,6 +102,60 @@ StandaloneLoaded standalone_input(inp::StandaloneInput& si)
...
+    pfg.data_selection.processes = GeantImportDataSelection::optical;
+    setup::physics_from(pfg, imported);
+
+    // Copy optical physics from import data
+    si.problem.physics = imported.optical_physics;
+
+    // TODO: Where should these processes be enabled by the user?
+    // inp::OpticalPhysics?  inp::GeantSetup (GeantPhysicsOptions)?
+    // inp::PhysicsFromGeant? Which is meant to be user input and which is
+    // loaded during setup, or are some a combination of both?
```
hhollenb, line 140: Yeah this is a sticking point I still have a bit with the current input system (and any actually refactoring can probably be deferred until after we get some integration results). I personally think it makes sense for the `inp::OpticalPhysics` to be entirely separated from Geant, and have user applications populate it. If they want to import data / defaults from Geant4, then helper classes can return a `inp::OpticalPhysics` struct with values pulled from Geant and then user apps have full control to check and modify those values. Then there's a kind of unambiguous "truth" in `inp::OpticalPhysics` and we don't have to juggle every possible way users might want to import data from Geant.

hhollenb (approved):
Looks great, thank you! I'll go ahead and pull and try running a standalone program to see if I encounter any issues as well.

src/celeritas/setup/StandaloneInput.cc
sethrj: Not sure if you saw https://github.com/celeritas-project/celeritas/pull/2197#discussion_r2717274888 but I think the default should be to mirror as best we can what the Geant4 application has selected, and allow users to override or customize as needed. Requiring users to populate the physics a second time is asking for trouble.

sethrj:
> A few more data points:
> 
> * The failure does not occur in debug mode but does in reldeb and release
> * I still see it with a single GPU thread
> * The value that's returned in `UniformRealDistribution::operator()` is within the correct bounds, but when that value is checked in `IsotropicDistribution::operator()` it's not

A few thoughts so far @amandalund:
- This is another one of those bizarre "your kernel argument size is too large" errors like we saw before #743
- It's possible I accidentally broke our vecgeom on device in #2116  ; try reverting that? (though this will also change what's in constant memory since the state data is different and will be hard to separate from the above suggestion)
- We've seen similar when there are unwitting duplicate symbols in the code

Can you post the output from `./bin/celer-config`?

amandalund:
Yeah, it does kind of have the flavor of #639 unfortunately. 

I tried reverting #2116 (hoping I resolved the conflicts reasonably) and it did get rid of those failures, though as you said, that doesn't necessarily mean it's the culprit.

Output from celer-config:
```json
{
 "config": {
  "build_type": "relwithdebinfo,shared",
  "constants": "CODATA 2018",
  "core_geo": "VecGeom",
  "core_rng": "xorwow",
  "debug": true,
  "geant4": "global-dynamic,multithreaded,shared,gdml",
  "gpu_architectures": "70;80;90",
  "hostname": "gpu07",
  "openmp": "event",
  "real_type": "double",
  "units": "CGS",
  "use": [
   "covfie",
   "cuda",
   "geant4",
   "hepmc3",
   "openmp",
   "root",
   "vecgeom"
  ],
  "vecgeom": "index,Scalar",
  "versions": {
   "CLHEP": "2.4.7.1",
   "CUDA": "12.3.52",
   "G4VG": "1.0.4",
   "Geant4": "11.3.0",
   "HepMC3": "3.03.00",
   "ROOT": "6.32.08",
   "Thrust": "2.2.0.0",
   "VecGeom": "1.2.10",
   "covfie": "0.15.3+builtin"
  }
 },
 "version": "0.7.0-dev.266+optical-runner.2aca7d452"
}
```

sethrj:
I'll see if i can reproduce it and try moving it to global memory.

sethrj:
OK I reproduced it on milan0 (A100) on your branch. I'll see if moving the params+state to global memory helps.

amandalund:
Awesome, thanks!

sethrj:
Moving to global memory doesn't fix it. I'm 90% sure this is a compiler bug with RDC that's being triggered because we're generating an IsotropicDistribution CUDA function into the `celeritas` library whereas before we weren't. I'll see if I can reproduce on develop ...

amandalund:
Added the patch @sethrj suggested to word around the failures for now.

PR #2198, opened 2026-01-17
Simplify errors when initialization fails

sethrj:
If an error happens during Celeritas setup via one of the integration managers, a cascade of other failures can occur in debug mode. This suppresses those other messages so that the original failure is more apparent to the user.

As a bonus, this also refactors and substantially simplifies the setup performed in `BeginOfRunAction` (and to a lesser extend, `EndOfRunAction`) by moving them into the integration singleton.

Finally, it fixes an accidental omission of the "set GPU on worker thread" that I only noticed due to code coverage (thanks @pcanal !) because we had unwittingly guarded a "setup shared params" call in the integration base.


src/accel/detail/IntegrationSingleton.hh
```diff
@@ -110,6 +98,7 @@ class IntegrationSingleton
     std::unique_ptr<SetupOptionsMessenger> messenger_;
     Stopwatch get_time_;
     bool have_created_logger_{false};
+    bool failed_setup_{false};
```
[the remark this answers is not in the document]
sethrj: In Geant4 the master execution always precedes the workers, so there is no race condition.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -137,10 +143,11 @@ class TMITestBase : virtual public IntegrationTestBase
 };
 
 //---------------------------------------------------------------------------//
-// LAR SPHERE
+// LAR SPHERE (EM only)
 //---------------------------------------------------------------------------//
 class LarSphere : public LarSphereIntegrationMixin, public TMITestBase
 {
+  public:
```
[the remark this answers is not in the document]
sethrj: Wrong; the default access specifier is public.

cmake/CeleritasG4Tests.cmake
```diff
@@ -20,14 +20,16 @@ run manager.
...
     <target>
-    [PREFIX string]      # before target in test name
-    [SUFFIX string]      # after target in test name
+    [NAME string]        # test name (default: target)
     [ARGS arg [...]]     # passed to command line
     [LABELS label [...]] # tagged in CTestFile
+    [RMTYPE [serial] [mt] [task]] # run manager
+    [OFFLOAD [cpu] [gpu] [g4] [ko]] # offload
     [DISABLE]            # display but don't run
     [WILL_FAIL]          # expect the test to exit with a failure
```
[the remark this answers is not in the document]
sethrj: The documentation was incorrect previously.

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -170,6 +177,35 @@ class LarSphere : public LarSphereIntegrationMixin, public TMITestBase
         EXPECT_DOUBLE_EQ((event_id == 1 ? 10.0 : 1.0),
                          step->GetTrack()->GetWeight());
     }
+
+    //! Check wrapped RuntimeError caught by GeantExceptionHandler
+    void caught_g4_runtime_error(RuntimeError const& e) override
+    {
+        if (!catch_exceptions_)
+        {
+            return TMITestBase::caught_g4_runtime_error(e);
```
pcanal, line 187:
```suggestion
            throw TMITestBase::caught_g4_runtime_error(e);
```
i.e. what is the meaning of `!catch_exceptions_`?

src/accel/detail/IntegrationSingleton.cc
```diff
@@ -390,6 +328,92 @@ void IntegrationSingleton::update_logger()
     }
 }
 
+//---------------------------------------------------------------------------//
+/*!
+ * Initialize shared params implementation.
+ */
+void IntegrationSingleton::initialize_master_impl()
```
pcanal, line 336:
```suggestion
void IntegrationSingleton::initialize_controller_impl()
```
or `manager` to better match `worker`, other alternative `main`, `Primary`

pcanal (approved):
LGTM

src/accel/detail/IntegrationSingleton.cc
sethrj: Yeah I guess there i was using g4 nomenclature (master) but will change to manager.

test/accel/TrackingManagerIntegration.test.cc
sethrj: Perhaps "handle and check" is a better name. The function call is definitely *not* supposed to be a throw; it lets the base class handle the exception with inheritance, which will log the exception and mark the test as failed.

pcanal: Then `catch_exceptions_atch` is indeed a misnomer and would recommend renaming it.

amandalund (approved):
Nice, thanks @sethrj!

PR #2200, opened 2026-01-22
Extract splitmix64 rng

davidsgr:
In preparation for reseeding an RNG every time a track is created, this PR extracts the SplitMix64 RNG from within the Xorwow RNG to make it an independent class.

### DIFF
--- a/src/corecel/random/engine/SplitMix64.hh
+++ b/src/corecel/random/engine/SplitMix64.hh
@@ -39,7 +39,7 @@ class SplitMix64
 /*!
  * Construct the SplitMix64 engine with the given seed.
  */
-SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
+CELER_FUNCTION SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
 {
     /* * */
 }
--- a/src/corecel/random/engine/SplitMix64.hh
+++ b/src/corecel/random/engine/SplitMix64.hh
@@ -23,7 +23,7 @@ class SplitMix64
 {
   public:
     // Construct the SplitMix64 class with the given seed
-    inline explicit SplitMix64(std::uint64_t seed);
+    inline CELER_FUNCTION explicit SplitMix64(std::uint64_t seed);
 
     // Produce a random number
     inline CELER_FUNCTION std::uint64_t operator()();
@@ -37,7 +37,7 @@ class SplitMix64
 // INLINE DEFINITIONS
 //---------------------------------------------------------------------------//
 /*!
- * Construct the SplitMix64 engine with the given seed
+ * Construct the SplitMix64 engine with the given seed.
  */
 SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
 {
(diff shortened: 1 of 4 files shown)


src/corecel/random/engine/SplitMix64.hh
```diff
@@ -0,0 +1,62 @@
...
+/*!
+ * \brief RNG for initializing the state for other RNGs
+ *
+ * See https://prng.di.unimi.it for a details of the SplitMix64 engine
+ */
+class SplitMix64
+{
+  public:
+    // Construct the SplitMix64 class with the given seed
+    inline explicit SplitMix64(std::uint64_t seed);
```
sethrj: This will need to be CELER_FUNCTION too.

src/corecel/random/engine/SplitMix64.hh
```diff
@@ -0,0 +1,62 @@
...
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct the SplitMix64 engine with the given seed
+ */
+SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
+{
+    /* * */
```
sethrj:
```suggestion
 * Construct the SplitMix64 engine with the given seed.
 */
SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
{
```

src/corecel/random/engine/SplitMix64.hh
davidsgr: Fixed.

src/corecel/random/engine/SplitMix64.hh
davidsgr: Fixed.

test/corecel/random/SplitMix64.test.cc
```diff
@@ -0,0 +1,41 @@
...
+TEST(SplitMix64Test, TEST_IF_CELER_DEVICE(device))
+{
+    celeritas::SplitMix64 sm(12345);
+
+    EXPECT_EQ(2454886589211414944ul, sm());
+    EXPECT_EQ(3778200017661327597ul, sm());
+    EXPECT_EQ(2205171434679333405ul, sm());
+    EXPECT_EQ(3248800117070709450ul, sm());
+    EXPECT_EQ(9350289611492784363ul, sm());
+}
```
amandalund: As is this test won't actually use the device since the data is still allocated on the host.

amandalund (approved):

test/corecel/random/SplitMix64.test.cc
davidsgr: Removed this test since this class has no device-specific code.

src/corecel/random/engine/SplitMix64.hh
```diff
@@ -0,0 +1,62 @@
...
+    std::uint64_t state_;
+};
+
+//---------------------------------------------------------------------------//
+// INLINE DEFINITIONS
+//---------------------------------------------------------------------------//
+/*!
+ * Construct the SplitMix64 engine with the given seed.
+ */
+SplitMix64::SplitMix64(std::uint64_t seed) : state_(seed)
```
sethrj: Also missing host device decoration. Interestingly nvcc seems to be ok with it

sethrj (approved):
Thanks!

src/corecel/random/engine/SplitMix64.hh
davidsgr: Fixed.

davidsgr:
This should be ready.  Failing tests appear unrelated.

sethrj:
Yeah I really need to redo our CI images so that it doesn't fail when trying to use intel hardware :(

PR #2201, opened 2026-01-22
Rename SDParams and move to geocel

sethrj:
This renames `celeritas/user/SDParams` to `geocel/DetectorParams` to correspond to DetectorId and the expected use of `track.detectors()` to get the view. It should be part of the "geometry model" rather than the physics since it just describes the mapping of geometry to user detectors. To allow testing in geocel I made the class inherit from GeantGeoTestBase, and used the opportunity to do some minor work to allow compatibility with single-precision builds, simplifying the CMakeLists. I also moved a couple of the geant4-only geometry tests into the universal GeoTests.

This also renames GeoTrackView to CoreGeoTrackView for consistency.

See #1397, #1750


test/geocel/GenericGeoTestBase.t.hh
```diff
@@ -95,7 +95,17 @@ auto GenericGeoTestBase<HP>::geometry() const -> SPConstGeo const&
...
-    return std::make_unique<WrappedGeoTrack>(this->make_geo_track_view());
+    if constexpr (std::is_same_v<real_type,
+                                 typename TraitsT::TrackView::real_type>)
+    {
+        return std::make_unique<WrappedGeoTrack>(this->make_geo_track_view());
+    }
+    else
+    {
+        CELER_NOT_CONFIGURED(
+            "geometry track view compatible with current real_type");
```
stognini, line 98:
```suggestion
            "geometry track view not compatible with current real_type");
```

sethrj: The `CELER_NOT_CONFIGURED` argument is what *needs to be*, not what *is*. This isn't exactly the most correct way to use the macro though.

amandalund (approved):

PR #2202, opened 2026-01-23
Debug and fix CUDA RDC linking issue in heuristic geo

sethrj:
For Reasons Unknown, adding an `IsotropicDistribution` as class data to the PrimaryGenerator causes UB-like failures downstream in the celeritas/geo/Geometry heuristic test (see #2197). This occurs even when the heuristic test accesses all data from global memory rather than const. Two things seemed to fix it:
- moving `IsotropicDistribution{}` from class data in `PrimaryGenerator` to be a local variable in the templated `PrimaryGenerator::operator()`
- linking the heuristic geo test CUDA into libtestcel_celeritas rather than directly into the geo_Geometry test (2359b6514b15cbd6b0889789b765e1fde2e8a1ff).

So this PR does three small but related things:
- Adds more diagnostics to the BuildOutput (namely, device debug and VecGeom cuda flags)
- Moves the heuristic geo executor data to global memory rather than constant memory (i.e. passed by value as part of the executor in function arguments)
- Links the heuristic test .cu file into the celeritas test library rather than directly into the test (which fixes the failure, though with CUDA RDC you never know whether it's a permanent fix)

```json
{
 "config": {
  "build_type": "relwithdebinfo,shared",
  "constants": "CODATA 2018",
  "core_geo": "VecGeom",
  "core_rng": "xorwow",
  "debug": [
   "host",
   "device"
  ],
  "geant4": "global-dynamic,multithreaded,shared,gdml",
  "gpu_architectures": "80",
  "hostname": "milan0",
  "openmp": "disabled",
  "real_type": "double",
  "units": "CGS",
  "use": [
   "covfie",
   "cuda",
   "geant4",
   "hepmc3",
   "vecgeom"
  ],
  "vecgeom": "index,Scalar,cuda=80",
  "versions": {
   "CLHEP": "2.4.7.1",
   "CUDA": "12.9.86",
   "G4VG": "1.0.5",
   "Geant4": "11.3.2",
   "HepMC3": "3.03.01",
   "Thrust": "",
   "VecGeom": "1.2.11",
   "covfie": "0.15.3"
  }
 },
 "version": "0.7.0-dev.279+heuristic-sample-failure.4e15c13b6"
}
```


amandalund (approved):
Thanks for digging into this @sethrj and for the quick fix!

PR #2077, opened 2025-10-27
Add SD params and output on boundary crossing

lebuller:
This MR adds detector params in the core params and passes through to the optical core params. The `detector` data is added to the core track data and a sensitive detector view was added to the core track view. In addition, the detector params class was updated to accept an `input::detectors` at construction instead of a volume id vector.

PR #2203, opened 2026-01-24
Add SD mapping params to core and optical data

sethrj:
This substantially refactors and finishes some work @lebuller started in #2077. It constructs detector params similar to surface/volume params, and it passes them into the core and optical data. These will replace the hit processor and similar.

See #1750.


hhollenb (approved):
Looks pretty straightforward from the optical side!

amandalund (approved):
Looks good!

PR #2204, opened 2026-01-25
Add primary IDs to optical simulation

sethrj:
LArSim's detector output requires the detector hits to be correlated to Geant4 primary particles. This means that, like with #1837, we need to send a primary ID to the device and keep it with the created track.


src/celeritas/inp/Events.hh
```diff
@@ -93,11 +93,14 @@ struct CorePrimaryGenerator : PrimaryGenerator
...
  *
  * \note The sampled optical photon primaries are unpolarized.
+ * \sa celeritas::optical::PrimaryGeneratorAction
  */
 struct OpticalPrimaryGenerator : PrimaryGenerator
 {
     //! Total number of primaries
     size_type primaries{};
+    //! ID to assign all primaries
+    PrimaryId primary_id;
```
amandalund: I'm not sure it makes sense to store a primary ID in the primary generator data: this is used for the "particle gun" type generation, so any particles created from these distributions _are_ the primaries and aren't correlated with any Geant4 parent tracks.

sethrj: That's true. I thought it could help with testing but you're right that it doesn't really make sense.

src/celeritas/optical/gen/PrimaryGeneratorData.hh
```diff
@@ -20,6 +22,7 @@ namespace optical
 struct PrimaryDistributionData
 {
     size_type num_photons{};
+    PrimaryId primary;  //!< For correlating to G4 tracks (test use)
```
amandalund: I think we can remove this too.

amandalund (approved):
Thanks!

PR #2205, opened 2026-01-25
Add optical physics AGENTS.md

sethrj:


src/celeritas/optical/AGENTS.md
```diff
@@ -0,0 +1,191 @@
...
+### 4. Primary Generation (`PrimaryGenerator`)
+- User-configurable distributions (energy, angle, shape)
+- For standalone testing without Geant4
+- Can specify primary ID explicitly (defaults to invalid)
+- **Path**: `gen/PrimaryGenerator.{hh,cc}`, `gen/PrimaryGeneratorAction.{hh,cc}`
+
+### 5. Geant4 Offload
+- Photons generated from main stepping loop distributions
+- Stored in `GeneratorDistributionData` with primary ID
+- Actions: `GeneratorAction`, `OffloadAction`, `OffloadGatherAction`
```
amandalund: This one seems a bit redundant with Cherenkov and scintillation. It could also be a little confusing because the "offload" actions listed here are offloading data from the Celeritas EM loop to the optical loop rather than from Geant4 to Celeritas.

We could also list the "direct generator" (generating photons from buffered track initializers).

sethrj: Yes, that's the one I was thinking of.

esseivaju (approved):

PR #2206, opened 2026-01-26
Rename and slightly refactor G4 track reconstruction helper

sethrj:
This renames the `TrackProcessor` to `GeantTrackReconstruction`, which better reflects its job of saving and restoring track data rather than "processing" tracks as we process hits. I tried to adjust names to better reflect that we're irreversibly capturing data from the input track, and the track that we reconstruct is not "new"'d but is a reference to an owned and *temporarily* valid track. Finally, I adjusted the ownership semantics of the G4Step to reflect that it's shared between the track reconstructor and hit procsesor.

### DIFF
--- a/src/celeritas/ext/GeantTrackReconstruction.hh
+++ b/src/celeritas/ext/GeantTrackReconstruction.hh
@@ -45,7 +45,7 @@ class GeantTrackReconstruction
     CELER_DEFAULT_MOVE_DELETE_COPY(GeantTrackReconstruction);
 
     // Clear G4Track reconstruction data
-    void end_event();
+    void clear();
 
     // Register mapping from Celeritas PrimaryID to Geant4 track ID
     [[nodiscard]] PrimaryId acquire(G4Track&);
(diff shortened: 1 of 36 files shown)


src/celeritas/ext/GeantTrackReconstruction.hh
```diff
@@ -20,55 +20,50 @@ class G4VUserTrackInformation;
...
-    explicit TrackProcessor(VecParticle const&);
+    GeantTrackReconstruction(VecParticle const&, SPStep);
 
-    ~TrackProcessor();
-    CELER_DEFAULT_MOVE_DELETE_COPY(TrackProcessor);
+    ~GeantTrackReconstruction();
+    CELER_DEFAULT_MOVE_DELETE_COPY(GeantTrackReconstruction);
 
     // Clear G4Track reconstruction data
     void end_event();
```
esseivaju: Maybe `end_event` should be renamed as we're calling it each time `LocalTransporter::Flush` is called, which doesn't always coincide with end of event.

esseivaju (approved):
Looks good! Just one minor comment

src/celeritas/ext/GeantTrackReconstruction.hh
sethrj: Hmmm so as implemented we could have multiple "primaries" that share the same primary ID, but there's just been a Flush in between? I guess `clear` would be suitable?

esseivaju: Right, we can have primaries sharing the same ID after the Flush. If we don't want that behavior I suppose we'd need to differentiate between an end-of-event flush, and a buffer full flush. Otherwise, renaming it to `clear` sounds good

PR #2207, opened 2026-01-26
Add track offload interface base class

Rashika-Gupta:
This PR:
- renames the `LocalOpticalOffload` to `LocalOpticalGenOffload` 
- introduces interface class `TrackOffloadInterface` for pushing `G4Track`
- unifies "local offload" interface construction in integration singleton
- adds additional error checking to `TrackingManagerConstructor`


sethrj:
@amandalund This is ready for you to take a look now. I think I may have screwed up when suggesting an interface hierarchy for the LocalTransporter code, since we'd previously assumed that we could have a long-lived pointer to the object and only "initialize" it at the beginning of the run. Depending on how messy the follow-on work is, we might want to restore the local transporter as a standalone object, but have a private set of classes that implement the different kinds of offloading (since we could have both EM and optical and maybe even optical gen as well).

test/accel/TrackingManagerIntegration.test.cc
```diff
@@ -404,7 +404,8 @@ void LarSphereOptical::EndOfRunAction(G4Run const* run)
     auto& integration = detail::IntegrationSingleton::instance();
     if (integration.mode() == OffloadMode::enabled)
     {
-        auto& local_transporter = integration.local_transporter();
+        auto& local_transporter
+            = dynamic_cast<LocalTransporter&>(integration.local_offload());
```
amandalund, line 408: Why do we access `local_offload` instead of `local_track_offload` here? Because it'd require a second dynamic cast?

amandalund (approved):
Yeah, I think we'll still have some work to do to clean this up, it's getting a little complicated with all our different integration methods and offload modes... but that can be later on once we have most of the functionality in place.

test/accel/TrackingManagerIntegration.test.cc
sethrj: Yeah it's pretty much equivalent, I guess it seemed more intuitive to me to get the base class and downcast to the point that we need it. It does indeed need cleaning...

PR #2208, opened 2026-01-26
Move Optical Core State Counters to Device

LSchwiebert:
This makes the same change as PR #2177 except that the optical core state counters are moved to the GPU. This is the next step toward removing/reducing the GPU synchronization when running an event.


LSchwiebert:
I accidentally submitted this as a draft PR. Now that I switched it, it still won't run the build-docker step. Not sure how to fix that.

Also, I noticed that one of the tests is intermittently failing. It is accel/TrackManagerIntegration.test.cc in the void OpticalSurfaces::EndOfRunAction(G4Run const* run) function on lines 687-689, where the accumulators are all 0 and should be >0.

If it's something I broke with this PR, I will check it out. Probably, but I'm not sure if that's the case.

amandalund:
If you mark it as ready for review and then push a commit (an empty one is fine) that should trigger the CI.

sethrj (approved):
Nice, thanks!

src/celeritas/optical/Transporter.cc
```diff
@@ -93,6 +93,9 @@ void Transporter::transport_impl(CoreState<M>& state) const
             }
         }
 
+        // No longer have a reference to the counters, so need to retrieve the
+        // updated values
+        counters = state.sync_get_counters();
```
amandalund, line 98: Do you know how we eventually plan to get around the D->H transfer and sync here as well as in the core stepper?

amandalund (approved):
Thanks @LSchwiebert, looks great!

src/celeritas/optical/Transporter.cc
sethrj: For the core stepper, I expect we won't be doing a full solve per invocation but something closer to a single step: we could return immediately without fetching counters and then sync during the flush (or at the start of the next "step" call).

For optical, we can make this an async copy and then sync on the previous loop's counter. If that is zero, then we just did an extra loop with zero optical tracks, but it'll let us completely hide the latency.

LSchwiebert: OK. Good. I was in the process of writing a long speculative reply about some things we could try, but seems like you already have an idea. In any case, I agree this is the most challenging place to solve this.

For the core stepper, I also think it would be OK to invoke a small kernel to update the values, since we don't have any conditionals on the CPU side that are checking the values, like there are in the function you highlighted.

amandalund: Sounds like a plan!

esseivaju: I suppose that would also require changing how we buffer primaries? For the core stepper, we'll flush at the end of event, or when the buffer is full. We'll probably want to transport tracks before the buffer is full

Issue #1894, opened 2025-08-06
Implement Geant4 step/track onload


esseivaju:
In Geant4, it might be necessary to "onload" tracks back to the CPU/Geant4. Reasons include:

- Unsupported detector geometry regions
- Use CPU fast simulation, or other custom physics processes
- Low number of active tracks on GPU

Geant4 tracking algorithm is detailed in the "Book For Toolkit Developers, section 2.4" ([html version](https://geant4-userdoc.web.cern.ch/UsersGuides/ForToolkitDeveloper/html/OOAnalysisDesign/Tracking/tracking.html)). In short, `G4Track` are transported by a "tracking manager', either the tracking manager associated to the particle definition, or a default thread-local tracking manager if there are none. Tracking manager have a single function call, `HandOverOneTrack(G4track*)` that is responsible to transport one track to completion through the stepping manager.

To offload a track back to Geant4, Celeritas can reconstruct `G4Track` at the end of a step, and push the reconstructed track back to Geant4 depending on dynamic track properties. There are a few inconvenient with this approach:

- Once the Geant4 tracking manager transports a track, there is no easy way to offload that track back to GPU. We'd probably need a custom stepping action for that
- For Celeritas secondaries that need to be pushed to Geant4, we need to assign them a new track ID as we reconstruct the `G4Track`. Because Celeritas doesn't see all Geant4 track, we don't know which ids are available to use

An alternative would be to extend the Celeritas tracking manager to support hybrid CPU/GPU transport instead of pushing the track back to Geant4:

- This would required that Celeritas is initialized to transport tracks on either CPU or GPU, I'm not sure if already support that, or if that can be easily implemented
- Celeritas Tracking manager can use the Geant4 stepping manager (or G4HepEM?) to transport a step (which will take care of invoking fast simulation), at the end of the step we can decide if the track needs to be sent back to the GPU

sethrj:
Note that `G4StackManager::PushOneTrack` assumes ownership of the track pointer, so the new track reconstructor should return a `unique_ptr`.

PR #2209, opened 2026-01-27
Define views for Geant4 step/track integration

sethrj:
This expands #1954 to include Geant4 track views (and step/step point views) that will allow better code reuse for the optical primary creation (#2103) and hits (#1397), as well as step/track reconstruction (#1894).

Refactored with Copilot's assistance.


src/celeritas/ext/GeantStepView.cc
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file celeritas/ext/GeantStepView.cc
+//---------------------------------------------------------------------------//
+#include "GeantStepView.hh"
+
+namespace celeritas
+{
+//---------------------------------------------------------------------------//
+
```
stognini, line 78: Was there anything that you did not want to be inlined?

sethrj: Sorry yes I noticed that during self review and still have to fix the docs.

sethrj:
@pcanal Regarding our performance discussion, it looks like clang (and presumably gcc) is able to completely optimize,  yielding the same assembly code for
```c++
auto y_position(G4Track const& t)
{
    GeantTrackViewConst view{t};
    return native_value_from(view.pos()[1]);
}
```
as with
```c++
double y_position_direct(G4Track const& t)
{
    return t.GetPosition()[1] * units::millimeter;
}
```

sethrj:
@pcanal see https://github.com/sethrj/testsnippets/tree/master/_celeritas_code/track-view

sethrj:
@pcanal Could you please review this one too? Thanks.

pcanal:
@sethrj Did you get a chance to run the extra checks (varying optimization level) we mentioned?

sethrj:
@pcanal `-O2` is identical, `-O1` is identical for the view code (but G4String included via G4Version loses some optimizations), and `-O0` is unreadable because of all the other static initialization crap that gets activated (like `CLHEP::Z_HAT4`).

pcanal (approved):
LGTM (The CI failure 'seem' unrelated)

PR #2210, opened 2026-01-27
Add JSON serialization for distribution inputs

amandalund:
This add `to_json` and `from_json` functions for the distribution and optical primary generator inputs.

### DIFF
--- a/src/celeritas/inp/EventsIO.json.cc
+++ b/src/celeritas/inp/EventsIO.json.cc
@@ -38,7 +38,7 @@ void from_json(nlohmann::json const& j, EnergyDistribution& v)
 {
     EIO_LOAD_VARIANT(delta, MonoenergeticDistribution);
     EIO_LOAD_VARIANT(normal, NormalDistribution);
-    CELER_ASSERT_UNREACHABLE();
+    CELER_VALIDATE(false, << "invalid EnergyDistribution input");
 }
 
 void to_json(nlohmann::json& j, ShapeDistribution const& v)
@@ -50,7 +50,7 @@ void from_json(nlohmann::json const& j, ShapeDistribution& v)
 {
     EIO_LOAD_VARIANT(delta, PointDistribution);
     EIO_LOAD_VARIANT(uniform_box, UniformBoxDistribution);
-    CELER_ASSERT_UNREACHABLE();
+    CELER_VALIDATE(false, << "invalid ShapeDistribution input");
 }
 
 void to_json(nlohmann::json& j, AngleDistribution const& v)
@@ -62,7 +62,7 @@ void from_json(nlohmann::json const& j, AngleDistribution& v)
 {
     EIO_LOAD_VARIANT(delta, MonodirectionalDistribution);
     EIO_LOAD_VARIANT(isotropic, IsotropicDistribution);
-    CELER_ASSERT_UNREACHABLE();
+    CELER_VALIDATE(false, << "invalid AngleDistribution input");
 }
 
 void to_json(nlohmann::json& j, OpticalPrimaryGenerator const& v)
--- a/src/celeritas/inp/EventsIO.json.cc
+++ b/src/celeritas/inp/EventsIO.json.cc
@@ -19,14 +19,14 @@ namespace inp
 //!@{
 //! I/O routines for JSON
 
-#define EIO_LOAD_VARIANT(NAME, TYPE)                    \
-    do                                                  \
-    {                                                   \
-        if (auto iter = j.find(#NAME); iter != j.end()) \
-        {                                               \
-            v = j.get<TYPE>();                          \
-            return;                                     \
-        }                                               \
+#define EIO_LOAD_VARIANT(NAME, TYPE) \
+    do                               \
+    {                                \
+        if (j.at("_type") == #NAME)  \
+        {                            \
+            v = j.get<TYPE>();       \
+            return;                  \
+        }                            \
     } while (0)
 
 void to_json(nlohmann::json& j, EnergyDistribution const& v)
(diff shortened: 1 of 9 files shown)


src/corecel/inp/DistributionsIO.json.cc
```diff
@@ -0,0 +1,82 @@
...
+    j = nlohmann::json{
+        {"normal", {{"mean", v.mean}, {"stddev", v.stddev}}},
+    };
+}
+
+void from_json(nlohmann::json const& j, NormalDistribution& v)
+{
+    auto const& params = j.at("normal");
+    params.at("mean").get_to(v.mean);
+    params.at("stddev").get_to(v.stddev);
```
sethrj:
```suggestion
    CELER_JSON_LOAD_REQUIRED(params, v, mean);
    CELER_JSON_LOAD_REQUIRED(params, v, stddev);
```

src/corecel/inp/DistributionsIO.json.cc
```diff
@@ -0,0 +1,82 @@
...
+template<class T>
+void from_json(nlohmann::json const& j, DeltaDistribution<T>& v)
+{
+    j.at("delta").get_to(v.value);
+}
+
+void to_json(nlohmann::json& j, NormalDistribution const& v)
+{
+    j = nlohmann::json{
+        {"normal", {{"mean", v.mean}, {"stddev", v.stddev}}},
```
sethrj:
```suggestion
        {"normal", {CELER_JSON_PAIR(v, mean), CELER_JSON_PAIR(v, stddev)}},
```

src/celeritas/inp/EventsIO.json.cc
```diff
@@ -0,0 +1,92 @@
...
+{
+    j = std::visit([](auto const& d) { return nlohmann::json(d); }, v);
+}
+
+void from_json(nlohmann::json const& j, EnergyDistribution& v)
+{
+    EIO_LOAD_VARIANT(delta, MonoenergeticDistribution);
+    EIO_LOAD_VARIANT(normal, NormalDistribution);
+    CELER_ASSERT_UNREACHABLE();
+}
```
sethrj: It seems that the common way of defining JSON variant ("anyOf") types for schema is to have a [discriminator field](https://redocly.com/learn/openapi/discriminator#when-to-use-the-openapi-discriminator) such as `$type` (C#), `_typename` (ROOT), and to have the other attributes live at the same level. That also prevents the ambiguity that currently exists of `{"delta": {...}, "normal": {...}}`. Is it ok to change to that?

We should also document/codify this for reuse. Perhaps a function in JsonUtils:
```c++
/*!
 * Construct a key/value pair for JSON polymorphism.
 */
inline std::pair<std::string, std::string> json_type_pair(std::string&& s)
{
    return {"_type", std::move(s)};
}
```
so that `to_json` would look like:
```c++
j = {json_type_pair("normal"), CELER_JSON_PAIR(v, mean), CELER_JSON_PAIR(v, stddev)};
```

Thoughts?

sethrj: Actually this is definitely what we want: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions

src/corecel/inp/DistributionsIO.json.cc
amandalund: Yeah, I left it out because the macro seemed more opaque/not any more concise, but I guess it's safer (e.g. prevents typos like `{"normal", {{"mean", v.mean}, {"stddev", v.mean}}}`).

src/celeritas/inp/EventsIO.json.cc
amandalund: Sounds good!

src/celeritas/inp/EventsIO.json.cc
```diff
@@ -0,0 +1,92 @@
...
+
+#define EIO_LOAD_VARIANT(NAME, TYPE) \
+    do                               \
+    {                                \
+        if (j.at("_type") == #NAME)  \
+        {                            \
+            v = j.get<TYPE>();       \
+            return;                  \
+        }                            \
+    } while (0)
```
sethrj, line 30: We could also probably turn this into a regular function:
```c++
bool try_load_variant<T>(std::string const& typename, nlohmann::json const& j);
```
so it would be:
```c++
try_load_variant<MonoenergeticDistribution>("delta", j)
|| try_load_variant<NormalDistribution>("normal", j)
|| failed_to_load_variant(j);
```
which could later be expanded through template metaprogramming if we wanted.

src/celeritas/inp/EventsIO.json.cc
```diff
@@ -0,0 +1,92 @@
...
+void to_json(nlohmann::json& j, EnergyDistribution const& v)
+{
+    j = std::visit([](auto const& d) { return nlohmann::json(d); }, v);
+}
+
+void from_json(nlohmann::json const& j, EnergyDistribution& v)
+{
+    EIO_LOAD_VARIANT(delta, MonoenergeticDistribution);
+    EIO_LOAD_VARIANT(normal, NormalDistribution);
+    CELER_ASSERT_UNREACHABLE();
```
sethrj: This should probably be a validate since it could happen from malformed user input.

sethrj (approved):
Thanks! You can ignore these comments and keep moving forward.

src/celeritas/inp/EventsIO.json.cc
amandalund: Good point.

src/celeritas/inp/EventsIO.json.cc
amandalund: Will consider for future improvements :)

PR #2211, opened 2026-01-27
Add MuCF interactor for muonic dt molecules

stognini:
This PR implements the muon-catalyzed fusion interactor for $(dt)_\mu$ molecules.

The interactor follows the same implementation as it is currently done by Acceleron. There is an oversimplification in the sampling mechanism that leads to a conservation of momentum but not energy, and it is noted in the interactor's documentation. This interactor is one component of #1508 .


stognini:
# Comparison of Celeritas vs. Geant4 (Acceleron)

## Sampling of $(dt)_\mu \rightarrow n + \mu + \alpha$

As noted in the `interactor`'s documentation, energy is not correctly conserved.
This may not be a concern yet, as the most important quantities---the neutron and muon energies---come from experimental data, and the variation in the total kinetic energy is about 3%.

The discrepancies seen in the muon energy are a consequence of the data interpolation, described in the next section.

dt_neutron_energy

dt_muon_energy

dt_alpha_energy

dt_total_energy


## Discrepancy in the cubic spline interpolation

Acceleron's interpolation is constructed via
```cpp
G4DataInterpolation(muonEnergyCDF,
                    muonEnergy,
                    numMuonEnergyPoints,
                    muonEnergy[1] / muonEnergyCDF[1],
                    0.0);
```
where `muonEnergy` and `muonEnergyCDF` is the same grid defined in `celeritas::inp::MucfPhysics::muon_energy_cdf`. Here the first/last derivatives are set, and are likely causing the discrepancies. The plot below compares the original CDF data with 100 interpolations from Geant4 (using `G4DataInterpolation`) and Celeritas (using `NonuniformGridCalculator`).

The discrepancy for muons with higher energies in the sampled CDF by Geant4 is the reason we see an excess of muons sampled with a high energy in the previous section. Since the $\alpha$'s momentum is determined via $\sum \vec{p_i} = 0$, the excess in high energy muons also leads to the discrepancy in the $\alpha$ kinetic energy plot. 

muon-energy-cdf-interpolation

src/celeritas/inp/MucfPhysics.cc
```diff
@@ -46,30 +46,28 @@ Grid mucf_muon_energy_cdf()
...
+             0.007953772477101693,
+             0.009194596305637525,
+             0.010849180562221111,
+             0.012353474314071864,
+             0.014045888515617822,
+             0.015650634617544647,
+             0.01738079707555165,
+             0.019111008546659452,
+             0.019976130619913615,
+             0.08};
```
sethrj:
```suggestion
             0.08,
             };
```
and reformat

test/celeritas/mucf/MucfInteractorHostTestBase.cc
```diff
@@ -0,0 +1,150 @@
...
+{
+    using constants::stable_decay_constant;
+    using AtomicMassNumber = AtomicNumber;
+    using units::AmuMass;
+    using units::ElementaryCharge;
+    using units::MevMass;
+    using units::Second;
+    using InvSecond = RealQuantity<UnitInverse<Second>>;
+
+    // Particle masses (from PDG)
```
sethrj: Which PDG? (they change from year to year; data provenance is good) 😉

amandalund:
Interesting, nice analysis of the discrepancies @stognini! Were you using linear interpolation in Celeritas? Do you know why they chose to use spline interpolation and clamped boundary conditions? It doesn't look like the best choice for the data...

src/celeritas/mucf/interactor/DTMucfInteractor.hh
```diff
@@ -88,35 +107,111 @@ template<class Engine>
...
+ * Calculate kinetic energy given a momentum magnitude and particle mass via
+ * \f$ K = \sqrt{p^2 - m^2} - m\f$ .
+ */
+CELER_FUNCTION units::MevEnergy
+DTMucfInteractor::calc_kinetic_energy(Real3 momentum_vec,
+                                      units::MevMass mass) const
+{
+    real_type momentum_mag = std::sqrt(ipow<2>(momentum_vec[0])
+                                       + ipow<2>(momentum_vec[1])
+                                       + ipow<2>(momentum_vec[2]));
```
sethrj: pass momentum_vec by reference, and use:
```suggestion
    real_type momentum_mag = celeritas::norm(momentum_vec);
```

src/celeritas/mucf/interactor/DTMucfInteractor.hh
```diff
@@ -88,35 +107,111 @@ template<class Engine>
...
+        case Channel::muonicalpha_neutron: {
+            // Muonic alpha: Equal and opposite momentum to neutron
+            sec[muonicalpha_idx].particle_id = data_.particle_ids.muonic_alpha;
+            sec[muonicalpha_idx].energy = sec[neutron_idx].energy;
+            for (auto i : range(3))
+            {
+                sec[muonicalpha_idx].direction[i]
+                    = -sec[neutron_idx].direction[i];
+            }
+            break;
```
sethrj, line 174: We could definitely use some kinematic helper functions ... but that can be a refactor for another day.

stognini:
> Were you using linear interpolation in Celeritas?

We are using cubic spline (which is the same used by Acceleron). On our side, that is defined in `MucfPhysics.cc::22`:
```cpp
cdf.interpolation.type = InterpolationType::cubic_spline;
```

src/celeritas/mucf/interactor/DTMucfInteractor.hh
```diff
@@ -88,35 +107,111 @@ template<class Engine>
...
 
-    // Kill primary and generate secondaries
-    Interaction result = Interaction::from_absorption();
-    result.secondaries = this->sample_secondaries(secondaries, rng);
+    size_type const neutron_idx{0};  // Both channels
+    size_type const muon_idx{1}, alpha_idx{2};  // Channel::alpha_muon_neutron
+    size_type const muonicalpha_idx{1};  // Channel::muonicalpha_neutron
+
+    IsotropicDistribution sample_isotropic;
+    NonuniformGridCalculator sample_muon_energy(data_.muon_energy_cdf,
```
sethrj, line 124: I'd add a note about the grid having a range of [0,1) and domain in MeV since that's a bit unusual.

sethrj (approved):
LGTM, just a couple of comments.

amandalund:
Aha gotcha, so we're using not-a-knot BCs (or really Geant4's not quite right not-a-knot).

It looks like the `G4DataInterpolation` class only support natural and clamped boundary conditions? And we don't currently support clamped.

test/celeritas/mucf/MucfInteractorHostTestBase.cc
stognini: Fair point. I was not too worried since it is mock data, but I'll add the exact PDG reference.

src/celeritas/mucf/interactor/DTMucfInteractor.hh
stognini: Beyond generic kinematic helpers, I also believe that some free functions shared among the 3 interactors will appear once I add DD and TT interactors and figure out their similarities.

test/celeritas/mucf/MucfInteractorHostTestBase.cc
sethrj: Oh yeah, I didn't notice. Still it's nice to at least put the year in case people in the future go "what, why are these different?"

src/celeritas/mucf/interactor/DTMucfInteractor.hh
sethrj: Perfect.

stognini:
The zero derivative at the last point is likely a mistake in Acceleron's implementation, according to Ara.

sethrj (approved):

PR #2212, opened 2026-01-28
Add JSON serialization for optical standalone input

amandalund:
This adds to/from json functions for everything required for the `OpticalStandaloneInput`. For a few of the inputs (control, tracking, system) I went ahead and added everything, but for the rest I added just what was needed for standalone optical.


src/corecel/io/JsonUtils.json.hh
```diff
@@ -39,6 +39,26 @@
...
+        if (iter != (OBJ).end() && !iter->is_null()) \
+        {                                            \
+            (STRUCT).NAME.emplace();                 \
+            iter->get_to(*(STRUCT).NAME);            \
+        }                                            \
+        else                                         \
+        {                                            \
+            (STRUCT).NAME.reset();                   \
+        }                                            \
+    } while (0)
```
sethrj: This would be better as a combination macro + templated function (macro passes in obj, #name, struct.name).

sethrj (approved):
I hope you're using copilot to help convert these 😅 thanks for the changes! 

And I wonder if, to avoid the proliferation of files and since we usually have to load all of the IO functions at once,  it would've been better to do one big JSON IO per subdirectory/directory. Too late now though 😅

pcanal:
> And I wonder if, to avoid the proliferation of files and since we usually have to load all of the IO functions at once,

I also wonder to what extend we could auto-generate those since we have access to an engine that can parse the header and give us a view to the input (Cling) [I assume that the json output from regular ROOT I/O is not working well for these use cases]

sethrj:
ROOT doesn't support JSON input does it? I thought it was output only. And as you know the constraints on ROOT (requires shared libs, is a bitch to build and keep working) mean we don't want to rely on it for critical I/O pathways.

PR #2213, opened 2026-01-28
Add grid documentation and fix detail build

sethrj:


amandalund (approved):

src/corecel/random/engine/detail/RanluxppImpl.hh
```diff
@@ -430,7 +430,7 @@ CELER_FUNCTION RanluxppArray18 multiply_9x9(RanluxppArray9 const& in1,
  * \p out is guaranteed to be smaller than the modulus.
```
mrguilima, line 430: Noticed the "out" variable name is actually "result".  I suggest rewriting the sentence:
- The result in \p out is guaranteed...
+ The output in \p result is guaranteed...

PR #2214, opened 2026-01-28
Fix `GeantOpticalPhysicsOptions` JSON serialization 

amandalund:
Some of the optical physics options (Cherenkov, scintillation, boundary, WLS) write a null json object if the process is disabled. When loading `GeantOpticalPhysicsOptions` from json, `CELER_JSON_LOAD_OPTION` was used to load the individual processes. However, this macro omits the field if it's null, which meant the default (activated) options were being used instead of the intended null (deactivated) ones.

### DIFF
--- a/src/celeritas/ext/GeantOpticalPhysicsOptionsIO.json.cc
+++ b/src/celeritas/ext/GeantOpticalPhysicsOptionsIO.json.cc
@@ -215,7 +215,14 @@ void from_json(nlohmann::json const& j, GeantOpticalPhysicsOptions& options)
         return;
     }
 
-#define GOPO_LOAD_OPTION(NAME) CELER_JSON_LOAD_OPTION(j, options, NAME)
+#define GOPO_LOAD_OPTION(NAME)                          \
+    do                                                  \
+    {                                                   \
+        if (auto iter = j.find(#NAME); iter != j.end()) \
+        {                                               \
+            iter->get_to(options.NAME);                 \
+        }                                               \
+    } while (0)
     check_format(j, format_str);
     GOPO_LOAD_OPTION(cherenkov);
     GOPO_LOAD_OPTION(scintillation);
--- a/test/JsonUtils.hh
+++ b/test/JsonUtils.hh
@@ -1,0 +1,31 @@
+//------------------------------- -*- C++ -*- -------------------------------//
+// Copyright Celeritas contributors: see top-level COPYRIGHT file for details
+// SPDX-License-Identifier: (Apache-2.0 OR MIT)
+//---------------------------------------------------------------------------//
+//! \file JsonUtils.hh
+//---------------------------------------------------------------------------//
+#pragma once
+
+#include <nlohmann/json.hpp>
+
+#include "TestMacros.hh"
+
+namespace celeritas
+{
+namespace test
+{
+//---------------------------------------------------------------------------//
+//! Verify JSON round-trip serialization.
+template<class T>
+inline void verify_json_round_trip(T const& input, char const* expected)
+{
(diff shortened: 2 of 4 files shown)


src/celeritas/ext/GeantOpticalPhysicsOptionsIO.json.cc
```diff
@@ -215,7 +215,14 @@ void from_json(nlohmann::json const& j, GeantOpticalPhysicsOptions& options)
         return;
     }
 
-#define GOPO_LOAD_OPTION(NAME) CELER_JSON_LOAD_OPTION(j, options, NAME)
+#define GOPO_LOAD_OPTION(NAME)                          \
```
sethrj, line 218: Now that you point it out, I think `CELER_JSON_LOAD_OPTION` should be changed to match this behavior:
- no key: default (this would be an "error" if it's required)
- `null` value: use `get_to`, deserializing the class with a null value (this is what your updated code does and is the most sensible since the class's `from_json` can specify whether `null` means disable)
- `{}` value: also use `get_to` (but the class would use presumably interpret this as "default values for everything")

We'd probably need to also adjust `CELER_JSON_SAVE_WHEN` to match the behavior. The idea was that it's good to know what default values are being used, and what parameters are being omitted.

sethrj (approved):
This explains those warnings I was seeing 😅

src/celeritas/ext/GeantOpticalPhysicsOptionsIO.json.cc
amandalund: I think that makes sense.

I'm not sure how to update `CELER_JSON_PAIR_OPTION` and `CELER_JSON_PAIR_WHEN` accordingly, except to remove them and replace with `CELER_JSON_SAVE_WHEN`.

sethrj: We can leave it for another day

amandalund: Eh yeah it looks like we'd have to update several tests (mainly orange json results), so another day, another PR.

amandalund:
Some bad news: I'm seeing the error from #2197, #2202 show up again, this time in the optical primary generator device test:
```console
[ RUN      ] LArSphereGeneratorTest.primary
status: Initializing Geant4 run manager
status: Initializing Geant4 geometry and physics list
status: Building Geant4 physics tables
info: Loading Geant4 geometry from GDML at /home/alund/celeritas_project/celeritas/test/geocel/data/lar-sphere.gdml
status: Initializing problem
status: Loading VecGeom geometry from in-memory Geant4 geometry
info: CUDA stack size was changed from 1024 to 8192; restoring to original values
status: Constructing metadata
/home/alund/celeritas_project/celeritas/src/celeritas/setup/Problem.cc:807: warning: Problem contains optical physics without any geometry surface definitions: default physics will be used for all surfaces
info: Creating 1 device streams
info: Constructed 1 optical materials with 1 present in the geometry
status: Celeritas optical setup complete
status: Initialized Celeritas optical state
/home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
/home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
/home/alund/celeritas_project/celeritas/src/corecel/math/ArrayUtils.hh:254:
celeritas: internal assertion failed: costheta >= -1 && costheta <= 1
```
(Also on develop)

PR #2215, opened 2026-01-28
Add optical sensitive detectors

hhollenb:
Add optical sensitive detectors for scoring photon hits for #1397 

As a simple first implementation of scoring hits, the same detectors from `DetectorParams` is used to define the detector regions. The `DetectorExecutor` kernel is responsible for copying hits for every track into the state buffer, with tracks not in detector regions marking their hits with an invalid detector ID. All hits are copied into pinned memory on the host, which is then cleaned up and passed to a user callback function.

This isn't an optimized / complete implementation by any means, but I've added some tests to help check future optimizations are correct.

I ran into some collisions with naming `ScoringParams` and `DetectorParams` so I just wrote something quickly and I'm seeking suggestions for what to rename things to.


src/celeritas/optical/CoreTrackData.hh
```diff
@@ -103,6 +104,7 @@ struct CoreStateData
     ParticleStateData<W, M> particle;
     PhysicsStateData<W, M> physics;
     RngStateData<W, M> rng;
+    DetectorStateData<W, M> scoring;
```
sethrj: This is confusing...

src/celeritas/optical/detector/ScoringTrackView.hh
```diff
@@ -0,0 +1,100 @@
...
+//---------------------------------------------------------------------------//
+/*!
+ * Set the hit data for this track.
+ *
+ * Should have a valid detector ID to indicate it is a valid hit.
+ */
+CELER_FUNCTION void ScoringTrackView::score_hit(DetectorHit hit)
+{
+    CELER_EXPECT(hit.detector);
+    this->hit() = std::move(hit);
```
sethrj: Since move can't be used on device (and the data is all trivial) let's just send the argument by const reference and make a copy here.

src/celeritas/optical/detector/DetectorData.hh
```diff
@@ -0,0 +1,139 @@
...
+template<Ownership W, MemSpace M>
+struct DetectorStateData
+{
+    //!@{
+    //! \name Type aliases
+    template<class T>
+    using StateItems = StateCollection<T, W, M>;
+    //!@}
+
+    StateItems<DetectorHit> all_track_hits;
```
sethrj:
```suggestion
    StateItems<DetectorHit> detector_hits;
```
for consistency with other state data objects in the code

src/celeritas/optical/detector/ScoringParams.hh
```diff
@@ -0,0 +1,50 @@
...
+class DetectorAction;
+//---------------------------------------------------------------------------//
+/*!
+ * Manages user callback for optical detectors.
+ *
+ * Constructs the \c DetectorAction used by sensitive detectors to send hits
+ * back to the user provided callback function. If the callback function is not
+ * provided, then no \c DetectorAction is created.
+ */
+class ScoringParams final
```
sethrj: We don't mark as final any of the other classes that don't inherit from an interface

src/celeritas/optical/action/DetectorAction.hh
```diff
@@ -0,0 +1,62 @@
...
+ * Record sensitive detector data for optical photons at the end of every step.
+ *
+ * The \c DetectorExecutor is responsible for copying hit data for every photon
+ * into the state buffer at the end of every step on a kernel level. Even if a
+ * track was not in a detector, it is still copied into the state buffer with
+ * an invalid detector ID. All hits are copied into pinned memory on the host,
+ * where invalid hits are erased. A span of only valid hits is then passed into
+ * the user provided callback function.
+ */
+class DetectorAction final : public OpticalStepActionInterface,
```
sethrj, line 33: I think to be consistent these should all be ScoringAction/Data/Executor... we might want to defer renaming to a different PR though, or at least lay out what the 

I was also confused initially because DetectorAction is just an implementation detail of ScoringParams, and the *creation* of those is what does all the work (kind of like the optical offload). The action/executor definitely need to be in a detail namespace.

sethrj (approved):
Sorry this took me a minute to review. I have just a couple of thoughts, most of them with naming. Perhaps we could do a quick hackathon today/tomorrow to get the names  all lined up across the codebase?

src/celeritas/optical/action/DetectorAction.hh
hhollenb: Looking back on it, I think scoring is a bad name cause that is usually used for the simpler tallying of things like energy deposits. Maybe a DetectorHitParams or UserDetectorParams is better for indicating that it's meant to deal with the actual hits and callbacks?

sethrj: Don't think too hard about it; we could easily end up in a hell of combining detector, sensitive, hit, callback, user, score, tally, ...

I think "sensitive detector" is a special case of a "scoring region": the scoring region is specifically a set of logical volumes, and the scoring action is a general "callback" capability that gets the entire step.

An arguably scoring regions are just a specialization of a "user stepping action"...

src/celeritas/inp/Scoring.hh
```diff
@@ -148,6 +154,22 @@ struct Scoring
...
+//---------------------------------------------------------------------------//
+/*!
+ * Enable detector callback for hits in optical physics simulations.
+ */
+struct OpticalScoring
+{
+    //!@{
+    //! \name Type aliases
+    using HitCallbackFunc
+        = std::function<void(Span<optical::DetectorHit> const&)>;
```
sethrj: Should be a "sized reference" to const hits

```suggestion
        = std::function<void(Span<optical::DetectorHit const>)>;
```

src/celeritas/inp/Scoring.hh
```diff
@@ -148,6 +154,22 @@ struct Scoring
...
+    //!@}
+
+    //! Hit callback function for optical detectors
+    std::optional<HitCallbackFunc> detector_callback;
```
sethrj:
```suggestion
    HitCallbackFunc callback;
```

sethrj (commented):
To do per our discussion:
- Eliminate SCoringParams
- DetectorAction is constructed with callback functions, if and only if a non-null function is given
- OpticalScoring -> OpticalDetector
- ScoringTrackView internlals go directly into Executor

src/celeritas/optical/CoreTrackData.hh
sethrj: TODO: rename detector

sethrj:
Thanks again!!

src/celeritas/optical/detector/DetectorData.cu
```diff
@@ -0,0 +1,48 @@
...
+    DetectorStateData<Ownership::reference, MemSpace::device> const& state,
+    StreamId stream_id)
+{
+    CELER_EXPECT(output);
+    CELER_EXPECT(stream_id);
+
+    size_type num_tracks = state.detector_hits.size();
+
+    // Copy all track hits from device
+    output->hits.resize(num_tracks);
```
amandalund: Since this is pinned memory and there might be some overhead to allocating maybe we should pre-allocate it at the start of the run?

src/celeritas/optical/detector/DetectorData.cu
```diff
@@ -0,0 +1,48 @@
...
+    Copier<DetectorHit, MemSpace::host> copy{{output->hits.data(), num_tracks},
+                                             stream_id};
+    copy(MemSpace::device, {state.detector_hits.data().get(), num_tracks});
+
+    // Synchronize to ensure all data is transferred before continuing
+    CELER_DEVICE_API_CALL(
+        StreamSynchronize(celeritas::device().stream(stream_id).get()));
```
amandalund:
```suggestion
    celeritas::device().stream(stream_id).sync();
```

amandalund:
```suggestion
    celeritas::device().stream(stream_id).sync();
```

amandalund (approved):
Awesome @hhollenb!

src/celeritas/optical/detector/DetectorData.cu
sethrj: It should only allocate when it *increases* since it's a `std::vector` backed by the pinned allocator. We're also going to postpone optimization on this one: we'll want to do it in conjunction with the async refactoring.

amandalund: But since we construct a new `DetectorHitOutput hit_results` in `DetectorAction::process_hits_impl()` currently won't it be reallocated at each step iteration?

hhollenb: Yeah I think it does get reallocated every step iteration. Would it then make sense to set it as a data member in `DetectorAction` and then reuse that in the `process_hits_impl`?

hhollenb: Making it a data member runs into a problem because the step methods are all const qualified, so it can't modify the DetectorHitOutput. As a temporary optimization maybe making it static inside the method?

amandalund: Or maybe store a shared_ptr?

sethrj: It needs to be stream-local though, right? It should be "aux data" ideally... and I think static may cause crashes during destruction due to the pinned allocation.

sethrj: For now we're not syncing, we just want to get the data back to the CPU. Just  make  it a regular synchronous std::vector that we allocate locally on every hit. We'll refactor to an async algorithm+storage+copy to follow on.

sethrj:
I incorporated the "optimize later" to the copying and simplified a bit: I'll create a new issue that tries to describe the requirements and possible implementations for a dynamic+async container. Since this is also a main stepping action, I updated the file names to be in line with the other optical stuff.

sethrj:
@amandalund @hhollenb If you could take a quick look at https://github.com/celeritas-project/celeritas/pull/2215/changes/20f81c531faa395eed8d32f4abca0f4cf6de6980..3d6cc0bfb7178b2b75bd584d8e3c18e3a9e216f4, this should be g2g.

amandalund (approved):

PR #2216, opened 2026-01-29
Add muCF atom and spin selectors

stognini:
This PR adds the capability of selecting a muonic atom and its spin based on the number fraction of d and t isotopes present in a dt mixture, which are two of the components described in #1508 .


src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
```diff
@@ -0,0 +1,99 @@
...
+    {
+        result = MucfMuonicAtom::deuterium;
+    }
+    else
+    {
+        result = MucfMuonicAtom::tritium;
+    }
+
+    CELER_ENSURE(result < MucfMuonicAtom::size_);
+    return result;
```
amandalund: I think this could be simplified using the Bernoulli distribution and calculating `deuterium_probability_` (or even the distribution itself) in the constructor:

```suggestion
    return BernoulliDistribution(deuterium_probability_)(rng)
               ? MucfMuonicAtom::deuterium
               : MucfMuonicAtom::tritium;
```

And similarly in the `MuonicAtomSpinSelector`.

stognini: Good point. Also good point about the `deuterium_probability_`. Better to calculate that in the constructor.

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
```diff
@@ -0,0 +1,89 @@
...
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Select a muonic atom given the material information.
+ *
+ * This class uses the \f$ q_\text{1S} \f$ formula [ \todo
+ * https://doi.org/10.1134/1.1926428 ]
```
sethrj: What's this to-do mean?

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
```diff
@@ -0,0 +1,89 @@
...
+/*!
+ * Select a muonic atom.
+ */
+template<class Engine>
+CELER_FUNCTION MucfMuonicAtom MuonicAtomSelector::operator()(Engine& rng)
+{
+    return BernoulliDistribution(deuterium_probability_)(rng)
+               ? MucfMuonicAtom::deuterium
+               : MucfMuonicAtom::tritium;
+}
```
sethrj, line 88: This class seems a little simple: are you planning to expand it in the future? If not, it could be better represented as a "model" (or just two linear parameters) for calculating the probability.

src/celeritas/mucf/executor/detail/MuonicAtomSpinSelector.hh
```diff
@@ -45,18 +61,30 @@ MuonicAtomSpinSelector::MuonicAtomSpinSelector(MucfMuonicAtom atom)
...
 {
-    //! \todo switch on atom_
-    CELER_ASSERT_UNREACHABLE();
+    switch (atom_)
+    {
+        case MucfMuonicAtom::deuterium:
+            return BernoulliDistribution(this->deuterium_spin_probability())(
+                       rng)
+                       ? 3  // Spin 3/2
+                       : 1;  // Spin 1/2
```
sethrj: You could define `using HalfSpinInt = Quantity<HalfSpin, short int>` that uses `struct HalfSpin` with a value `Constant{0.5}`; so it would store the half-spin exactly or convert to a `Constant` when casting. Alternatively, you can use an enum for spin if you're switching over exact values.

test/celeritas/CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ celeritas_add_test(decay/MuDecay.test.cc ${_needs_double})
 #-----------------------------------------------------------------------------#
 # MuCF
 
-# Interactors
+celeritas_add_test(mucf/MuonicAtomSelector.test.cc ${_needs_double})
```
sethrj: Instead of `needs_double` (which we just added for old tests when I added float precision) prefer to run the code but skip testing the results: `if constexpr (CELERITAS_REAL_TYPE ...)`

test/celeritas/mucf/MuonicAtomSelector.test.cc
```diff
@@ -0,0 +1,243 @@
...
+        return deuterium_frac * q1s;
+    }
+
+    real_type calc_sigma(real_type num_samples, real_type success_prob)
+    {
+        return std::sqrt(num_samples * success_prob * (1 - success_prob));
+    }
+
+  private:
+    Engine rng_;
```
sethrj: For unit test you can just make the data protected and access it directly, rather than adding accessors.

test/celeritas/mucf/MuonicAtomSelector.test.cc
```diff
@@ -0,0 +1,243 @@
...
+//---------------------------------------------------------------------------//
+// TEST HARNESS
+//---------------------------------------------------------------------------//
+
+class MuonicAtomSelectorTest : public Test
+{
+  protected:
+    using Engine = ::celeritas::test::DiagnosticRngEngine<std::mt19937>;
+
+    void SetUp() override { rng_.reset_count(); }
```
sethrj: The rng count is automatically set to zero.

sethrj (approved):
I think it may be a bit overcomplicated, but it looks good.

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
stognini: Oh, that was for adding the reference and citing it instead of just the pasting the link itself.

Edit: Added the references to zotero.bib.

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
stognini: I don't know if there are any other ways to select the atom, so I do not have a plan on expanding it. My reasoning is mainly on making each of the "tasks" in the mucf executor a functor. So that if other mucf models are added in the future, or the executor is reorganized, or even if the theory/hardcoded data used by each functor changes, it is very easy to update, reuse, and move code.

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
```diff
@@ -0,0 +1,89 @@
...
+#include "celeritas/Types.hh"
+#include "celeritas/mucf/Types.hh"
+
+namespace celeritas
+{
+namespace detail
+{
+//---------------------------------------------------------------------------//
+/*!
+ * Select a muonic atom given the material information.
```
sethrj: The class should make it clear that this is a conditional probability *given that DT fusion will happen* (so I assume the embedded formula is intimately tied to the cross section?) Otherwise it's weird that the probability goes to 1 as the fraction of T goes to zero.

src/celeritas/mucf/executor/detail/MuonicAtomSelector.hh
sethrj: For something this small, I guess I'm saying it could be almost as effective to make a function `calc_deuterium_probability`.

src/celeritas/UnitTypes.hh
```diff
@@ -133,6 +133,13 @@ struct Mol
 
 //!@}
 
+//! Spin (in units of hbar/2)
+struct HalfSpin
+{
+    static CELER_CONSTEXPR_FUNCTION Constant value() { return Constant{0.5}; }
```
sethrj: I was thinking this struct should be in mucf types, but I like it here instead. However we need to use the actual value:
```suggestion
    static CELER_CONSTEXPR_FUNCTION Constant value() { return constants::hbar_planck / 2; }
```

test/celeritas/mucf/DTMucfInteractor.test.cc
```diff
@@ -104,30 +104,34 @@ class DTMucfInteractorTest : public MucfInteractorHostTestBase
...
+                    sec[0].energy, host_data.particle_masses.neutron);
+                auto const muon_p_mag = this->calc_momentum(
+                    sec[1].energy, host_data.particle_masses.mu_minus);
+
+                Real3 alpha_momentum, total_momentum;
+                for (auto i : range(3))
+                {
+                    real_type neutron_momentum_i = sec[0].direction[i]
+                                                   * neutron_p_mag;
+                    real_type muon_momentum_i = +sec[1].direction[i]
```
sethrj: Weird and possibly unintentional use of the unary `+` 🙃

test/celeritas/mucf/MuonicAtomSelector.test.cc
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+//---------------------------------------------------------------------------//
+TEST_F(MuonicAtomSelectorTest, muonic_atom_asymmetric_mixture)
+{
+    // 70/30 mixture
+    real_type const d_frac = 0.7;
+    real_type const t_frac = 0.3;
+    MuonicAtomSelector select_atom(d_frac, t_frac);
```
sethrj: Shouldn't you also have a case where there's an impurity e.g. H2 (protium)? That was the part that confused me.

sethrj (approved):

PR #2217, opened 2026-01-29
Warn and replace unphysical values in MFP grid

amandalund:
In the optical surface integration test I ran into some cases where the photon's mean free path (and step length) was zero. This occurred because the user-provided absorption MFP was zero at the last energy grid point and the photon energy was outside the grid. To avoid this without introducing special-case logic in the stepping loop, this change warns if the MFP grid contains nonpositive values and replaces them with epsilon.


sethrj (approved):
Nice, thanks!

PR #2218, opened 2026-01-30
Add generic G4 field adapter and add field testing

sethrj:
This reduces a major coverage gap: previously our cyl/cart map fields were untested. I've identified some shortcomings that we can discuss:
- the r/phi/z decomposition of the magnetic field in the CylMap case is either wrong or unsuitable for representing linear fields (which seems bad)
- the cartesian map field crashes when given points outside its range of validity

The main change outside of documentation is to simplify the celeritas-geant4 field conversion with a new simple wrapper.


src/celeritas/field/CartMapField.covfie.hh
```diff
@@ -20,6 +20,9 @@ namespace celeritas
 //---------------------------------------------------------------------------//
 /*!
  * Interpolate a magnetic field vector on an x/y/z grid.
+ *
+ * NOTE: values outside the grid result in invalid extrapolation or a
+ * segmentation fault.
```
esseivaju: I think this is not true in cuda builds. Texture memory will clamp any out of bounds access

src/accel/CartMapMagneticField.hh
```diff
@@ -28,41 +27,26 @@ struct CartMapFieldGridParams
...
+// Generate field input with user-defined uniform grid from global field
+CartMapFieldInput MakeCartMapFieldInput(CartMapFieldGridParams const& params);
 
 //---------------------------------------------------------------------------//
-/*!
- * A user magnetic field equivalent to celeritas::CartMapField.
- */
-class CartMapMagneticField : public G4MagneticField
+//! On-the-fly field calculation with covfie using Celeritas data+units
+struct CartAdapterField
```
esseivaju: Maybe we can say that the reason we need the adapter is to hide the covfie dependency from downstream users?

test/accel/CartMapMagneticField.test.cc
```diff
@@ -0,0 +1,138 @@
...
+    {
+        // TODO: Checking outside the field's domain gives wrong errors or
+        // causes C assertions to fail
+        pos = {-1 * cm, 0.1 * cm, -0.1 * cm};
+        EXPECT_VEC_NEAR(
+            (Dbl3{0, 0, 0}), this->calc_field(cart_field, pos), tol);
+        pos = {1 * cm, 0.1 * cm, -0.1 * cm};
+        EXPECT_VEC_NEAR(
+            (Dbl3{0, 0, 0}), this->calc_field(cart_field, pos), tol);
+    }
```
esseivaju, line 133: I think the output of the first assertion is expected as covfie will clamp the values for out of bounds access. I am not sure why the 2nd assertion is triggering an error in covfie though...

test/accel/CylMapMagneticField.test.cc
```diff
@@ -0,0 +1,139 @@
...
+    this->check_field(cyl_field, pos, tol);
+
+    // Check elsewhere inside cylindrical volume
+    if (false)
+    {
+        // FIXME: this field value is wrong!
+        // expected X=-0.3T, actual X=1.2T
+        pos = {0.5 * cm, -1.6 * cm, 2.5 * cm};
+        this->check_field(cyl_field, pos, tol);
+    }
```
esseivaju, line 126: mmh, weird that only X differs...

esseivaju (commented):
Thanks! Nice refactor, gotta figure out what's going wrong.... 

Regarding covfie, and clamping of out-of-bounds values, is that acceptable or should it match other fields which zero out out-of-bounds?

src/accel/CylMapMagneticField.cc
esseivaju, line 44: I think the erroneous values in the cylindrical field come from here. We're calculating phi from the bfield values instead of x/y position. 😅

src/celeritas/field/CartMapField.covfie.hh
sethrj: Ah, interesting. I think I wrote this incorrectly: it seems to clamp on the low side but throws an assertion on the high side.

test/accel/CylMapMagneticField.test.cc
sethrj: It may be something with the atan2 in the r/phi/z cylindrical coordinate conversion? Frankly I forgot that we were interpolating the field *itself* in a cyl coordinate system; perhaps we should just store the x/y/z field on the cyl grid coordinates?

src/accel/CylMapMagneticField.cc
sethrj: Yeah I thought that might've been fishy... after this gets merged you can follow up with a fix? 😄

esseivaju: I'll follow-up with a fix

esseivaju (approved):

PR #2219, opened 2026-01-31
Add standalone optical app

amandalund:
This adds an app for running standalone optical photon simulations, similar to celer-sim but with only the optical data and stepping loop present.

Waiting on #2220 and #2221


hhollenb (approved):
Looks good, especially with getting the optical lar-sphere tests passing with the new app!

app/celer-optical/celer-optical.cc
```diff
@@ -0,0 +1,246 @@
...
+    {
+        os = &outf;
+        outf.open(output_filename);
+        CELER_VALIDATE(outf,
+                       << "failed to open output file at '" << output_filename
+                       << "'");
+        msg << "'" << output_filename << "'";
+    }
+    CELER_ASSERT(os);
+
```
sethrj: Replace with `corecel/io/FileOrConsole.hh`

sethrj (approved):
Nice (albeit expedient), thanks!

PR #2221, opened 2026-02-01
Fix optical photon import for older Geant4 versions

amandalund:
The fixes a couple errors I ran into in #2219 with only optical physics enabled.

In the Geant importer we were filtering particles _before_ correcting the optical photon PDG number for older Geant4 versions, meaning when only optical physics was enabled no particles were imported. During the process import we were then trying to get the Geant4 optical photon particle definition using the modified PDG.

We were also hitting a Geant4 error when trying to set the Geant4 scintillation process properties, even when scintillation was disabled:
```console
/home/runner/work/celeritas/celeritas/build/bin/celer-optical: critical: Geant4 error: both electron and proton should exist
G4EmSaturation::InitialiseBirksCoefficient: 'em0001' failed
```
so those are now only set if the process is enabled.


sethrj (approved):
Thanks!

PR #2222, opened 2026-02-01
Fix construction of CylMapMagneticField from G4MagneticField

esseivaju:
Follow-up to #2218 fixing erroneous values in CylMapField caused by incorrectly calculating phi during cartesian to cylindrical vector conversion


esseivaju:
@codex review

sethrj (approved):
Thanks!

PR #2223, opened 2026-02-01
Add helpers for array type, precision, and quantity conversion

sethrj:
The existing `convert_{to/from}_geant` functions are rather confusing because they do three things:
1. Convert the type from geant4 to/from celeritas types
2. Static-cast the precision of scalars/arrays if using single-precision Celeritas
3. Apply unit conversion factors but without using quantities.

To avoid confusion and errors i.e. the ones I originally *thought* were present in #2218, this defines new helper functions:
1. `to_array` and `to_g4vector` for type conversion
2. renames `array_cast` to `static_array_cast` to clarify what exactly is being cast (similar to the standard library)
3. `make_array_quantity` and related function for creating `Array<Quantity<...>>` that allows more expressive and type-safe conversion.


app/celer-g4/SensitiveDetector.cc
```diff
@@ -89,8 +89,8 @@ bool SensitiveDetector::ProcessHits(G4Step* g4step, G4TouchableHistory*)
     // reloaded
     hit.volume = log_vol->GetInstanceID();
     hit.copy_num = phys_vol->GetCopyNo();
-    hit.energy_dep = convert_from_geant(edep, CLHEP::MeV);
-    hit.time = convert_from_geant(pre_step->GetGlobalTime(), CLHEP::s);
+    hit.energy_dep = units::ClhepEnergy{edep}.value();
+    hit.time = native_value_from(units::ClhepTime{pre_step->GetGlobalTime()});
```
amandalund, line 93: Maybe the former should also be `native_value_from` (just to make  it clearer to developers that it's converting units) even though we know the energy units are both MeV.

src/accel/CartMapMagneticField.cc
```diff
@@ -79,7 +80,9 @@ MakeCartMapFieldInput(G4Field const& field,
     auto field_converter = [](Array<G4double, 3> const& bfield,
                               Array<G4double, 4> const&,
                               real_type cur_bfield[3]) {
-        auto bfield_native = convert_from_geant(bfield.data(), clhep_field);
+        using ClhepField = Quantity<units::ClhepUnitBField, double>;
```
amandalund: Use `ClhepField` from Quantities.hh?

src/corecel/math/ArrayQuantity.hh
```diff
@@ -0,0 +1,115 @@
...
+        result[i] = native_value_to<Q>(value[i]);
+    }
+    return result;
+}
+
+//---------------------------------------------------------------------------//
+/*!
+ * Convert an array of quantities to native values.
+ *
+ * This applies native_value_from element-wise to each component.
```
amandalund:
```suggestion
 * Get the values from an array of quantities.
 *
 * This applies value_as element-wise to each component.
```

src/geocel/detail/LengthQuantities.hh
```diff
@@ -0,0 +1,55 @@
...
+};
+
+//!@}
+
+//---------------------------------------------------------------------------//
+//!@{
+//! \name Length quantity aliases
+
+using CmLength = RealQuantity<Centimeter>;
+using ClhepLength = Quantity<Millimeter, double>;
```
amandalund, line 50: We also have a `ClhepLength` alias in `celeritas::units`

src/geocel/g4/Convert.hh
```diff
@@ -6,35 +6,54 @@
...
+//! Convert via a quantity to native Geant4 types/units
+template<class Q, class T>
+G4ThreeVector convert_to_geant(Array<T, 3> const& v)
+{
+    return to_g4vector(value_as<Q>(native_value_to<Q>(v)));
+}
 
 //---------------------------------------------------------------------------//
-// FREE FUNCTIONS
+//! Convert via a quantity to native Geant4 types/units
```
amandalund: Update description here and `convert_from_geant` below

amandalund (approved):
Looks good, thanks!

src/accel/CartMapMagneticField.cc
sethrj: Not the same: real_type vs Geant4 value (double ).

src/geocel/detail/LengthQuantities.hh
sethrj: Yea, the separate namespace here (lengthunits) is to allow inclusion in geocel: that was introduced a while back. I *thought* there was a good reason to not make the `celeritas/Units` directly alias `lengthunits` but I can check.

src/accel/CartMapMagneticField.cc
amandalund: The ones in Quantities are double too, were they meant to be real_type?

sethrj:
I'm going to move this to draft until we have a moment to breathe... don't want to step on the other work going on.

sethrj:
Once #2278 is in I'll pull in develop and update new usage as discussed.

app/celer-g4/SensitiveDetector.cc
sethrj: Actually it's not doing a conversion; if we used native_value_from it would convert from MeV to erg (if CGS).

PR #2224, opened 2026-02-02
Modify ORANGE tracking assertions for production-level problems

elliottbiondo:
This MR relaxes a DBC check in `SimpleUnitTracker::complex_intersect` to become a warning, issued when running a debug build on host. This particular DBC check was being hit at a rate of 1 in 2.2e5 calls when running the ATLAS lower end cap model. This MR will allow transport to proceed on this model, and other production-level models with the same issue (presumably a small overlap).

### DIFF
--- a/src/orange/univ/SimpleUnitTracker.hh
+++ b/src/orange/univ/SimpleUnitTracker.hh
@@ -572,22 +572,18 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
     // Current senses should put us inside the volume
     LogicEvaluator is_inside(vol.logic());
 
-    // Return early if the current sense already matches the target sense
-    // (a tracking error)
-    if (is_inside(calc_sense) == (target_sense == Sense::inside))
-    {
+    // Log a warning if the current sense already matches the target sense
 #if !CELER_DEVICE_COMPILE
-        if constexpr (CELERITAS_DEBUG)
+    if constexpr (CELERITAS_DEBUG)
+    {
+        if (is_inside(calc_sense) == (target_sense == Sense::inside))
         {
-            {
-                CELER_LOG_LOCAL(warning)
-                    << "Calculated surface sense at position " << repr(pos)
-                    << " already matches target sense";
-            }
+            CELER_LOG_LOCAL(warning)
+                << "Calculated surface sense at position " << repr(pos)
+                << " already matches target sense";
         }
-#endif
-        return {};
     }
+#endif
 
     // previous isect distance for move delta
     real_type previous_distance{0};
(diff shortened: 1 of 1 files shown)


elliottbiondo:
@sethrj RTR

src/orange/univ/SimpleUnitTracker.hh
```diff
@@ -566,7 +571,23 @@ SimpleUnitTracker::complex_intersect(LocalState const& state,
     // Calculate local senses, taking current face into account
     // Current senses should put us inside the volume
     LogicEvaluator is_inside(vol.logic());
-    CELER_ASSERT(is_inside(calc_sense) != (target_sense == Sense::inside));
+
+    // Return early if the current sense already matches the target sense
+    // (a tracking error)
```
sethrj: Did you compare the total number of steps with/without identifying an error? I.e., if we don't bother checking the sense, maybe the track will "naturally" restore balance again after a surface crossing or two?  The condition here indicates that a tracking error *already happened*, so it's possible that the next local crossing will put us to rights? 🤔

elliottbiondo: The lower end cap model runs whether we return early here or continue. I adopted the philosophy of returning immediately once we detect a failure to avoid hitting other DBCs downstream (even though that doesn't happen in this case). I agree that allowing tracking to continue could result in more steps, but whether or not the correct behavior depends on the overall goal of the simulation.

Hold on I will check to see if we actually do get more steps without early return.

elliottbiondo: Interestingly early return actually give more tracks:

With early return: 17,431,021
Without early return:   17,420,143

though the effect is minimal, as expected. I don't think there is a satisfactory answer here (without digging deeper in the underlying issues with this geom). I am happy to live it as is or change it.

elliottbiondo:
@sethrj RFAL

src/orange/univ/SimpleUnitTracker.hh
sethrj: We need to keep in mind that "return early" means "act like there's no intersection in this universe" which means "the last volume we (erroneously!) thought we were in continues until the nearest boundary in the global level". Returning early is *not* marking the track as errored.

elliottbiondo: I agree that not returning early give the possibility of remediating the track. Below is a care where it doesn't work, and we totally miss V2. However, if the V2 internal surface was oriented so to that the track hits it, then things would work out. 
sense_error

elliottbiondo: Changing to "no early return", per offline discussion, in order to ensure same behavior across build types and architecture.

elliottbiondo:
@sethrj this should be good-to-go pending CI

PR #2225, opened 2026-02-03
Clamp out-of-bound indices in `CartMapField`

esseivaju:
Follow-up to #2222. This add a `covfie::clamp` layer to host `CartMapField` to clamp out-of-bound access. 

The clamp layer is placed before linear interpolation because covfie’s linear backend [casts](https://github.com/acts-project/covfie/blob/4447a6076c3d5d486ab9a4492196db4877363c36/lib/core/covfie/core/backend/transformer/linear.hpp#L234-L236) the interpolation coordinate to an unsigned index. The [affine transform](https://github.com/celeritas-project/celeritas/blob/3331decb02742422a8268dc938ad8f842659ab64/src/celeritas/field/CartMapFieldParams.covfie.cc#L69-L83) only guarantees valid index coordinates for positions inside the field’s world bounds. For negative out-of-bounds positions, the cast would wrap to a positive index and trigger invalid access. Clamping first keeps the coordinate in-range and prevents that wraparound.

This makes the GPU/behavior consistent with each other.

### DIFF
--- a/src/celeritas/field/CartMapField.covfie.hh
+++ b/src/celeritas/field/CartMapField.covfie.hh
@@ -64,9 +64,9 @@ CartMapField::CartMapField(ParamsRef const& shared) : field_{shared.get_view()}
 /*!
  * Calculate the magnetic field vector for the given position.
  *
- * This does a 3-D interpolation on the input grid and reconstructs the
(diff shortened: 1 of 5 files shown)


sethrj (approved):
Awesome!

Issue #2014, opened 2025-10-07
Make core stepping algorithm asynchronous


sethrj:
*Derived from https://github.com/celeritas-project/docs/issues/17#issuecomment-2985727260 and associated dsicussion*

Ideally we could launch an entire step without having to wait on a CUDA return result, so that we can do useful work on the CPU in the meanwhile. There are several aspects of this can be tackled more or less independently, so we can add code to asynchronously...
- [ ] Copy primaries to the GPU (precondition for the first step): #1967
- [ ] Copy hits back to the CPU

sethrj:
@LSchwiebert @amandalund @esseivaju @whokion This is more or less what I'm envisioning for the overall async behavior. When we push primaries, we can buffer them and convert to initializers while we process lagged (i.e., from previous step, dotted lines) hit data, projectiles ($\gamma$, $e^{\pm}$ and $n$) which undergo nuclear interactions , and onloaded tracks. Then we can launch a step if no step is currently running (for efficiency we can also delay launching until we have some threshold of active tracks).

Finally at the end of the event, the tracking manages and fast processes are flushed, and then we can step/process until there are any stacked g4 tracks, or until we complete all the lagged processing and inactive tracks.

<img width="4733" height="4366" alt="Image" src="https://github.com/user-attachments/assets/2624cda6-e570-49b0-8b4b-a37c98989774" />
Here, dashed lines indicate "lagged" data (from a previous step), and the red-bordered boxes indicate that we need to "flush" them.

whokion:
@sethrj Very nice!  I assume that `ProcessDoIt` represents post-step actions (mainly for particles undergoing nuclear interactions, e.g., gamma-nuclear, electro-nuclear, and potentially neutron inelastic or capture), while on-loaded tracks correspond to region-dependent track segments or hadrons produced by decays(?). Should these also be connected to the TrackingManager (e.g. with a dotted line), since ProcessDoIt produces secondary particles and on-loaded tracks will be reprocessed by Geant4 or is it implicity implied?

sethrj:
Yes, exactly. I think those are more implicitly connected: I was assuming they're pushed to the G4 stacking manager, and then the regular stepping/event mechanics will automatically pop them off after the end of the step (and all the hit processing callbacks should happen either inside the start of the track, or before the end of a step).

whokion:
[...]

Issue #1958, opened 2025-09-12
Optimally share GPU memory resources


sethrj:
Currently the number of track slots has to be set manually, and increasing the number of Geant4 CPU threads (i.e., streams) offloading to Celeritas decreases the GPU performance because the track slots are independent for each stream.
- [ ] Share a single GPU state among many CPU streams/threads (would replace https://github.com/celeritas-project/celeritas/issues/1233, since each event would correspond to a separate CPU thread)
  - [ ] Spawn daughter RNGs based only on parent and seed data
  - [ ] Make stepping loop fully CPU-asynchronous (no/limited thrust, see #1877)
  - [ ] Use pinned-memory flag to communicate whether a new GPU step should be launched
- [ ] Automatically determine GPU memory availability, memory requirements per track slot, initializer requirements per track slot, and allocate based on that, or try expanding until memory limits hit
  - [ ] Auto-expand initializers: https://github.com/celeritas-project/celeritas/issues/1058
- [ ] Use async memory pools for temporary data that only need to be alive for part of a stepping loop (e.g., surface normals, interactions, hits)
- [ ] Make configurable by CPU memory usage limits + GPU memory usage limits?

sethrj:
See #1877

sethrj:
@davidsgr This is the base issue for the RNG work we're doing; I'll split that part off into a sub-issue.

PR #2226, opened 2026-02-03
Add abstract interface to CUDA/HIP events

sethrj:
This adds a platform-portable interface for waiting until the device is finished with a particular task. This will be needed for:
1. Asynchronously initializing primaries, launching the stepping loop, and processing hits: it allows us to check without blocking that memory transfers are complete. (#2014)
2. Dynamically resizing initializers (#1958): events will inform whether it's safe to swap pointers and start the stepping loop.

I also changed the Stream so that it can be constructed with CUDA/HIP disabled, in which case operations are fully synchronous (but memory allocation is disabled).


scripts/env/faraday.sh
```diff
@@ -10,11 +10,11 @@ if ! command -v load_system_env >/dev/null 2>&1; then
 fi
 
 # From modules/rocmmod
-export ROCM_PATH="/opt/rocm-7.0.1"
+export ROCM_PATH="/opt/rocm-7.2.0"
```
LSchwiebert, line 13: I think you can get away with just using "/opt/rocm" as this points to the same path. And it looks like they make only the latest installed version available anyway.

(I also just submitted a PR on the faraday docs to update the ROCm version and then they upgraded again. Lol.)

src/corecel/sys/Stream.hh
```diff
@@ -86,14 +91,17 @@ class Stream
     // Access the thrust resource allocator associated with the stream
     ResourceT& memory_resource();
 
-    // Synchronize this stream
+    // Block host execution until stream operations are all done
```
LSchwiebert: In the earlier code, the comment reads:

"Block host execution until stream operations are all complete."

If you want to be consistent/pedantic. Otherwise, ignore.

LSchwiebert:
I made a couple small comments -- both optional. I'll try running this on faraday and a system with a CUDA GPU once all the tests are passing.

scripts/env/faraday.sh
sethrj: I emailed Steve and he said they were going to try not deleting the old installations going forward: and since the ABIs for the entire toolchain are likely incompatible going from one version to the next, I think it's safer to explicitly encode the version here. That way, instead of weird link errors or subdirectories going missing, we get a very obvious "that compiler doesn't exist" which includes the original version we expected.

sethrj:
Thanks @LSchwiebert . I did check for success on faraday and milan0, but it looks like the failing case is when CUDA is enabled but no device is present (which copilot actually pointed out...)

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,95 @@
...
+  // Then do CPU work until the kernel or CPU is done
+  while (!cpu_work_done() && !my_kernel.ready())
+  {
+      cpu_work();
+  }
+ * \endcode
+ * Use \c my_kernel.sync() before the kernel launch to wait on the previous
+ * kernel launch before going again.
+ */
+class DeviceEvent
```
esseivaju, line 61: I think the class name is a bit misleading. CUDA events are really associated with a stream rather than device. Maybe `StreamEvent`, or `DeviceStreamEvent` are more accurate.

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,95 @@
...
+//---------------------------------------------------------------------------//
+class Stream;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Minimal wrapper around a CUDA/HIP event for synchronization.
+ *
+ * Events provide a mechanism for querying the status of asynchronous
+ * operations on GPU streams and synchronizing between host and device. This
+ * class creates events with timing disabled for minimal overhead.
```
esseivaju: We can also use event to synchronize between two streams. This will probably be needed in #2014 to synchronize between the data/compute streams.

We should extend API to support calling `cudaStreamWaitEvent`, e.g. `void stream_sync(Stream& other)`, or retrieve the underlying `cudaEvent_t` and synchronize through our `Stream` wrapper. [[ref](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/cuda-graphs.html#cross-stream-dependencies-and-events)]

src/corecel/sys/DeviceEvent.cu
```diff
@@ -0,0 +1,134 @@
...
+ * Construct a device event for the given stream.
+ *
+ * The stream pointer is stored internally so that \c record() can be called
+ * without passing the stream.
+ */
+DeviceEvent::DeviceEvent(Stream const& stream)
+{
+    EventT event;
+    CELER_DEVICE_API_CALL(EventCreateWithFlags(
+        &event, CELER_DEVICE_API_SYMBOL(EventDisableTiming)));
```
esseivaju: Probably fine for now, but we might want to consider setting the [cudaEventBlockingSync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g7b317e07ff385d85aa656204b971a042) flag. The doc says that best performance are obtained when it is not set, however, not setting the flag will result in a busy wait. Setting that flag yields back the CPU to other threads. In a realistic workload, the CPU will have plenty of work to do.

sethrj: It looks like that applies to the calling thread (i.e. our Geant4 thread) rather than the CUDA management thread, so I think the option would make sense only if we're oversubscribing threads. So maybe we should make it a configuration/env option?

src/corecel/sys/DeviceEvent.hh
sethrj: I think StreamEvent could work. Mainly I wanted to avoid ambiguity with HEP "Event".

sethrj:
@esseivaju I added the multi-stream stuff but it's not behaving as I expect... also I think I'm going to change the interfaces to more closely match the cuda (not storing a stream).
```
[ RUN      ] DeviceEventTest.device
/home/s3j/Code/celeritas-milan/test/corecel/sys/DeviceEvent.test.cc:142: Failure
Value of: e.ready()
  Actual: true
Expected: false

/home/s3j/Code/celeritas-milan/test/corecel/sys/DeviceEvent.test.cc:143: Failure
Value of: e2.ready()
  Actual: true
Expected: false

/home/s3j/Code/celeritas-milan/test/corecel/sys/DeviceEvent.test.cc:164: Failure
  new_g_value
    Which is: 3
  g_value
    Which is: 0

[  FAILED  ] DeviceEventTest.device (150 ms)
```

esseivaju:
Fixed the test, I think you swapped the two streams, and misplaced the e2 record so the sync wasn't waiting on the event being complete

esseivaju:
I think it makes sense to match the CUDA API and not store the stream as you suggest. It's only used in `record`, so it can just be passed as an argument.

src/corecel/sys/DeviceEvent.cu
esseivaju: Right, in most situation we wouldn't be oversubscribing so we can leave it as is for now, can be followed-up later with a env option (similar to other low-level device options which are also configured through env var).

sethrj:
@esseivaju @LSchwiebert I've refactored this to keep the device event independent from the stream: I added a "wait" method to the stream. I've also changed most of the functions to be clear about when host functionality isn't available; we can still "wait" on a null (no-device) event, but now we can't make any changes to the stream like launching functions.

src/corecel/sys/DeviceEvent.cc
```diff
@@ -0,0 +1,165 @@
...
+/*!
+ * Internal implementation holding the native CUDA/HIP event handle.
+ */
+struct DeviceEvent::Impl
+{
+    EventT event{nullptr};
+};
+
+//---------------------------------------------------------------------------//
+//! Destroy the event.
```
esseivaju, line 32:
```suggestion
/*!
 * Destroy the event.
 */
```

src/corecel/sys/DeviceEvent.cc
```diff
@@ -0,0 +1,165 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * Wait for the event to complete.
+ *
+ * This blocks the calling thread until all operations recorded before this
+ * event have finished executing on the device. Use this to synchronize the
+ * host with device operations.
+ *
+ * If the event is uninitialized
```
esseivaju:
```suggestion
 * If the event is uninitialized, this is a no-op
```

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,122 @@
...
+
+#include <memory>
+
+#include "corecel/Assert.hh"  // IWYU pragma: keep
+#include "corecel/Macros.hh"
+#include "corecel/sys/ThreadId.hh"
+
+#if CELER_DEVICE_SOURCE
+#    include "corecel/DeviceRuntimeApi.hh"
+#endif
```
esseivaju, line 17: Is there a reason that this needs an include guard? It seems like it would be fine to include from any cc file?

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,122 @@
...
+class Stream;
+
+//---------------------------------------------------------------------------//
+/*!
+ * Minimal wrapper around a CUDA/HIP event for synchronization.
+ *
+ * Events provide a mechanism for querying the status of asynchronous
+ * operations on GPU streams and synchronizing between host and device, and
+ * synchronizing between streams. This particular event is associated with a
+ * single stream.
```
esseivaju:
```suggestion
 * synchronizing between streams.
```

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * \par States
+ * - \b Constructed: when built with an active device, the instance
+ *   evaluates to \c true and manages an event object.
+ * - \b Null: when constructed with a nullptr, or when \c moved from, the class
+ *   instance is \c false. It does not manage an event nor does it associate
+ *   with a stream. The \c sync function is a null-op (the event
+ *   is always \c ready ), and \c record cannot be called.
```
esseivaju:
```suggestion
 *   instance is \c false. It does not manage an event. 
 *   The \c sync function is a null-op (the event is always
 *   \c ready ), and \c record cannot be called.
```

src/corecel/sys/DeviceEvent.hh
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * If no device is enabled (or Celeritas is compiled without CUDA/HIP support),
+ * only the null state is available.
+ *
+ * \par Example:
+ * \code
+  // Setup:
+  DeviceEvent my_kernel;
```
esseivaju:
```suggestion
  DeviceEvent my_kernel(celeritas::device());
```

src/corecel/sys/Stream.cc
```diff
@@ -58,29 +61,48 @@ struct Stream::Impl
...
     }
 }
 
 //---------------------------------------------------------------------------//
 /*!
- * Construct by creating a stream.
+ * Construct by creating a stream with the active device context.
+ *
+ * \pre A device must be active and configured.
+ * \deprecated This constructor is ambiguous: remnove it.
```
esseivaju:
```suggestion
 * \deprecated This constructor is ambiguous: remove it.
```

esseivaju (approved):
Looks good! Just a few mostly doc-related comments.

src/corecel/sys/DeviceEvent.hh
sethrj: It would propagate the requirement of the DeviceApi includes to basically everything downstream. This is a kind of hacky way of doing it; I think we can improve in the future by removing stream construction from the device.

esseivaju (approved):
Nice addition, thanks!

PR #2229, opened 2026-02-04
Add per track reseeding

davidsgr:
Enables the random number generator to be reseeding for every new track primary or secondary.  This is enabled through a conditional set at configuration.


davidsgr:
@sethrj This is ready for review.  I'm not sure if I added a compiler flag correctly, so take a look

src/accel/LocalTransporter.cc
```diff
@@ -293,6 +293,10 @@ void LocalTransporter::Push(G4Track& g4track)
      */
     track.event_id = EventId{0};
 
+    // Add track id and initialize step counter
+    track.geant_track_id = g4track.GetTrackID();
+    track.geant_step_count = g4track.GetCurrentStepNumber();
```
amandalund: Won't the step number always be zero for tracks we offload to Celeritas?

src/celeritas/phys/PrimaryGenerator.cc
```diff
@@ -183,6 +183,8 @@ auto PrimaryGenerator::operator()() -> result_type
         p.direction = sample_dir_(rng_);
         p.time = 0;
         p.event_id = EventId{event_count_};
+        p.primary_id = PrimaryId{event_count_ * primaries_per_event_ + i};
+        p.geant_track_id = p.primary_id.get();
```
amandalund: Since both the `primary_id` and `geant_track_id` are for mapping to a Geant4 track, it would make more sense for them to be invalid here because when a primary generator is used no tracks originate from Geant4.

Or if we need the `geant_track_id` to be set for initializing the RNG when inserting primaries, maybe we should name it something else? Wouldn't we also need to set the "step number" field? I think we'd also need to set them in our various `EventReader`s.

src/celeritas/track/detail/ProcessPrimariesExecutor.hh
```diff
@@ -45,6 +49,23 @@ struct ProcessPrimariesExecutor
...
+    // Fill the RNG state initializer for the Xorwow engine
+    inline CELER_FUNCTION void fillRngStateInitializer(
+        unsigned int seed,
+        unsigned int event_id,
+        unsigned int geant_track_id,
+        unsigned int geant_step_id,
+        XorwowRngEngine::RngStateInitializer_t& rng_init) const;
+
+    // Fill the RNG state initializer for the Ranluxpp engine
+    void fillRngStateInitializer(
```
amandalund:
```suggestion
    inline CELER_FUNCTION void fillRngStateInitializer(
```

src/celeritas/track/detail/ProcessSecondariesExecutor.hh
```diff
@@ -84,8 +84,11 @@ ProcessSecondariesExecutor::operator()(TrackSlotId tid) const
     // initialized in this slot
     TrackId const track_id{sim.track_id()};
 
-    for (auto const& secondary : track.physics_step().secondaries())
+    for (unsigned int secondary_idx :
+         celeritas::range(track.physics_step().secondaries().size()))
     {
+        auto const& secondary
+            = track.physics_step().secondaries()[secondary_idx];
```
amandalund: Can revert if we don't need the secondary index:

```suggestion
    for (auto const& secondary : track.physics_step().secondaries())
{
```

src/accel/LocalTransporter.cc
sethrj: Not when we do region based offload, but I guess they might be guaranteed to be unique track IDs?

davidsgr: So is that fine?  If we have unique track ids that should be sufficient to give us unique RNG states.

src/celeritas/phys/PrimaryGenerator.cc
davidsgr: I guess the step count should be set to zero, since these primaries are all at the first step.  I don't *think* that should be problem, since all of the primaries will have a unique (`geant_track_id`, `event_id`) and therefore a unique RNG.  

I agree actually that since primaries can be created in ways other than from Geant that this field should be renamed.  I named it that to distinguish from the Celeritas track id.  The primaries are basically all source particles, correct?  What about `source_track_id` and `source_step_count`?

src/accel/LocalTransporter.cc
amandalund: @sethrj you mean might not be guaranteed to be unique track IDs for region-based offload? I could see having repeated track IDs if tracks get offloaded, onloaded, and then offloaded again, and in that case the step count combined with the track ID should give us a unique state. For our current offload method it doesn't add anything though.

src/celeritas/phys/PrimaryGenerator.cc
davidsgr: Also, I'll take a look at the event readers

src/accel/LocalTransporter.cc
sethrj: Yes exactly. But then I thought a bit more: when we do the fast region offload, we're almost certainly going to kill the old track and inject a new track via stacking action, so it'll become a unique track ID.

src/celeritas/track/detail/ProcessPrimariesExecutor.hh
sethrj: and use celeritas naming conventions: `fill_rng_state_initializer`

src/accel/LocalTransporter.cc
amandalund: That makes sense. So should we remove the step count field?

sethrj: Yep, indeed.

CMakeLists.txt
```diff
@@ -137,6 +137,15 @@ celeritas_setup_option(CELERITAS_CORE_RNG hipRAND CELERITAS_USE_HIP)
 celeritas_define_options(CELERITAS_CORE_RNG
   "Celeritas runtime random number generator")
 
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# CELERITAS_RESEED
+# Random number generator selection
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+celeritas_setup_option(CELERITAS_RESEED trackslot)
+celeritas_setup_option(CELERITAS_RESEED track)
+celeritas_define_options(CELERITAS_RESEED
```
pcanal, line 146: Did we add documentation for the user to find and/or understand these options?  Do we have a documentation (at the code or user level) of what is the algorithm used for the reseeding?

src/accel/LocalTransporter.cc
```diff
@@ -227,10 +227,15 @@ void LocalTransporter::InitializeEvent(int id)
     if (!(G4Threading::IsMultithreadedApplication()
           && G4MTRunManager::SeedOncePerCommunication()))
     {
-        // Since Geant4 schedules events dynamically, reseed the Celeritas RNGs
-        // using the Geant4 event ID for reproducibility. This guarantees that
-        // an event can be reproduced given the event ID.
-        step_->reseed(event_id_);
+        if constexpr (CELERITAS_RESEED == CELERITAS_RESEED_TRACKSLOT)
+        {
```
sethrj: Switching this with the non-constexpr `if` above would allow the compiler to skip those checks.

CMakeLists.txt
sethrj: Good point. @davidsgr could you please add to the `Configuration options` section of `doc/usage/installation.rst`?

src/celeritas/track/TrackInitData.hh
```diff
@@ -69,6 +70,7 @@ struct TrackInitializer
     SimTrackInitializer sim;
     GeoTrackInitializer geo;
     ParticleTrackInitializer particle;
+    RngStateInitializer rng;
```
sethrj, line 73: We might want to guard this with `#if CELERITAS_RESEED == CELERITAS_RESEED_TRACKSLOT` to eliminate any potential performance impact from the extra memory/register usage. There would be downstream effects (e.g. also having to guard the places where it was used instead of using `if constexpr`)

src/corecel/random/engine/InitializeRngState.hh
```diff
@@ -0,0 +1,79 @@
...
+ * Fill a Ranluxpp state initializer given a seed, event id, and primary id
+ */
+CELER_FUNCTION void
+initialize_rng_state(unsigned int seed,
+                     unsigned int event_id,
+                     unsigned int primary_id,
+                     RanluxppRngEngine::RngStateInitializer_t& rng_init)
+{
+    // Initialize SplitMix64 with the seed XORed with the track id
+    SplitMix64 rng(seed ^ event_id ^ primary_id);
```
sethrj: Is the idea to hash the three integers to provide more randomness to the starting state? (Good idea.) Not sure two xors are going to be a very effective way to hash though?

src/corecel/random/data/RanluxppRngData.hh
```diff
@@ -36,6 +36,10 @@ struct RanluxppRngParamsDataImpl
     RanluxppArray9 advance_sequence;
 
     //// FUNCTIONS ////
+
+    //! Get the seed as a 64-bit unsigned integer
+    std::uint64_t get_seed() const { return seed; }
```
sethrj: Since these are structs, they do not need (and shouldn't have) accessors. They're part of the "data-oriented" paradigm. Just access the seed directly.

sethrj (approved):
Looks good! Thanks greg.

sethrj:
@davidsgr not sure if you saw the feedback above!

davidsgr:
> @davidsgr not sure if you saw the feedback above!

I did.  Sorry for the delay, I've been in a conference/meeting all week.  I'm planning on getting hopefully today or tomorrow.

src/accel/LocalTransporter.cc
davidsgr: Step count field removed

src/celeritas/track/TrackInitData.hh
davidsgr: Is this something you want to do now, or do you want to wait until we've studied performance impacts?

src/corecel/random/engine/InitializeRngState.hh
davidsgr: More randomness, yes, but in particular I didn't want any part of the Ranluxpp state to be repeated over and over again during a simulation.

Since `seed` is set once at the beginning by the user, what I wanted to avoid is for the first three values in the Ranluxpp state to always be the same.  I basically wanted SplitMix64 to have a unique seed for every (`seed`/`event_id`/`primary_id`) combination, so that all of the values in the state would be random.

At the same time, I wanted something computationally cheap.  This is what I came up with, but I'm open to suggestions.  Another possibility is to do something like a proper hash combine: https://www.boost.org/doc/libs/1_36_0/doc/html/hash/combine.html

CMakeLists.txt
davidsgr: Added.  Please review.

src/accel/LocalTransporter.cc
davidsgr: Done.

src/corecel/random/data/RanluxppRngData.hh
davidsgr: Changed.

src/celeritas/track/TrackInitData.hh
sethrj: How about we leave it for now and see if @elliottbiondo can check before/after numbers.

src/corecel/random/engine/InitializeRngState.hh
sethrj: There's actually a very large overlap between hashing and PRNGs: hash combine with FNV is:
```
set state ← magic number
for d in data:
  state ← state ^ d
  state ← state * magic prime
return state
```

So what give the maximum spread is something like:
```
set splitmix_state ← seed
for d in (event, primary):
  splitmix_state ← advance(splitmix_state)
  splitmix_state ← splitmix_state ^ d

for i in rng_ints.size():
  splitmix_state ← advance(splitmix_state)
  rng_ints.size[i] ← splitmix_state
  splitmix_state ← splitmix_state ^ (i % 2 == 0 ? event : primary)
```

I think that would guarantee (except for 'hash' collisions) that every integer in the resulting state is unique by (event,primary) tuple and completely different for a change to `seed`.

davidsgr: Ok, take another look

davidsgr:
@sethrj I think this is ready for another look.

PR #2230, opened 2026-02-04
Improve testing for field and geo view

sethrj:
This moves some changes to the base behavior out of #2133 to better illustrate the changes in that PR.


test/celeritas/CMakeLists.txt
```diff
@@ -248,7 +248,8 @@ endif()
 #-----------------------------------------------------------------------------#
 # Field
 
-if(CELERITAS_USE_VecGeom)
+if(CELERITAS_CORE_GEO STREQUAL "VecGeom"
+ AND (CELERITAS_USE_CUDA OR VecGeom_VERSION VERSION_GREATER_EQUAL "2.0"))
```
mrguilima, line 252: Can't the VG 2.0 solids-based geometry be rebuilt?

mrguilima (commented):
A late comment, just in case...

test/celeritas/CMakeLists.txt
sethrj: Alas no...

PR #2231, opened 2026-02-05
Avoid integer overflow in `DeMorganSimplifier` for large geometries

amandalund:
@elliottbiondo was seeing a seg fault during orange construction with ATLAS geometry, which looks like an integer overflow (in DeMorganSimplifier::Matrix2D::operator()):
```console
row: 10658, col: 10659, extent_: 98162, idx: 1046221255, data_.size(): 1045843652, sizeof(size_type): 4
```
We probably shouldn't be using a dense matrix if it requires O(1e10) elements, so this uses a set instead for the parents matrix.


esseivaju (approved):
Thanks for this fix! I wasn't expecting `CsgTree` with this many nodes 😅 I suspect we'll be able to update the other `std::vector` as well in a follow-up but at least it's not quadratic scaling...

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -141,7 +135,7 @@ class DeMorganSimplifier
 
     //! Parents matrix. For nodes n1, n2, if n1 * tree_.size() + n2 is set, it
     //! means that n2 is a parent of n1
```
esseivaju:
```suggestion
//! Parents matrix. If the pair {e1, e2} exists, e2 is parent of e1
```

src/orange/orangeinp/detail/DeMorganSimplifier.cc
```diff
@@ -25,37 +25,6 @@ namespace orangeinp
...
- * Access the element at the given index.
- */
-std::vector<bool>::reference
-DeMorganSimplifier::Matrix2D::operator[](indices index)
-{
-    auto& [row, col] = index;
-    CELER_EXPECT(row < extent_ && col < extent_);
-    return data_[row.get() * extent_ + col.get()];
-}
-
```
sethrj, line 49: It would be a little cleaner to have this return a `Matrix2D::ReferenceImpl` with a custom `operator=(bool)` that inserts/removes the element and `operator bool` that returns whether the element exists. That would allow us to change the implementation again in the future. Probably not worth the effort since that would be a detail of a detail of a detail..... :)

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -89,25 +102,6 @@ class DeMorganSimplifier
...
-
-        // Create the matrix view with the given extent size
-        explicit Matrix2D(size_type) noexcept;
-        // Access the element at the given index
-        std::vector<bool>::reference operator[](indices);
-        // The extent along one dimension
-        size_type extent() const noexcept;
-
-      private:
-        // TODO: sparse storage
```
sethrj, line 106: See https://github.com/celeritas-project/celeritas/pull/1289#discussion_r1724959752 😉

src/orange/orangeinp/detail/DeMorganSimplifier.hh
```diff
@@ -6,6 +6,7 @@
 //---------------------------------------------------------------------------//
 #pragma once
 
+#include <unordered_set>
```
sethrj, line 9: Also include `corecel/math/HashUtils.hh` (currently transiently included through CSG tree I think)

sethrj (approved):
Thanks! We shoudl've added a warning/error about massively large CSG trees.

Also it would be nice to add a "performance test" either at a low level or high level: a programmatic 10x10x10 grid of boxes (could implement with a parameterized gdml input) would hopefully let us see the memory/runtime impact of this.

sethrj:
I think a lot of the complexity arises from our workaround in https://github.com/celeritas-project/celeritas/issues/1260 : this is @elliottbiondo 's next major task and I think it'll help a lot.

elliottbiondo (approved):
Thanks for this!!! I can confirm that with this patch, the ATLAS lower end cap model runs without error.

